PR #2, opened 2025-05-12
Add CI for docgen

eric-wieser:


.github/workflows/blank.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Build project
+        run: lake build
+
+      - name: Save olean cache
+        run: |
+          lake exe cache pack
+          ls ~/.cache
+        
+      - name: Set up olean cache
```
mo271:
```suggestion
      - name: Save olean cache
```
or "Set down olean cache"? Just to distinguish from the other identically named step.

PR #6, opened 2025-05-13
Fix some latex formatting

mo271:
drive-by: correct classification of erdos_168.variants.limit_exists

### DIFF
--- a/FormalConjectures/Books/UniformDistributionOfSequences/Equidistribution.lean
+++ b/FormalConjectures/Books/UniformDistributionOfSequences/Equidistribution.lean
@@ -20,7 +20,7 @@ import FormalConjectures.Util.ProblemImports
 
 Corollary 4.2 of Chapter 1 states that the sequence $(x^n), n = 1, 2, ... ,$ is equidistributed modulo 1 for
 almost all x > 1. And a little bit further down:
-"one does not know whether sequences such as $(e^n)$, $(π^n)$, or even $((frac 3 2)^n)$"
+"one does not know whether sequences such as $(e^n)$, $(π^n)$, or even $((\frac 3 2)^n)$"
 are equidistributed modulo 1 or not.
 
 *References:*


FormalConjectures/Books/UniformDistributionOfSequences/Equidistribution.lean
```diff
@@ -20,7 +20,7 @@ import FormalConjectures.Util.ProblemImports
 
 Corollary 4.2 of Chapter 1 states that the sequence $(x^n), n = 1, 2, ... ,$ is equidistributed modulo 1 for
 almost all x > 1. And a little bit further down:
-"one does not know whether sequences such as $(e^n)$, $(π^n7)$, or even $((3/2)^n)$"
+"one does not know whether sequences such as $(e^n)$, $(π^n)$, or even $((frac 3 2)^n)$"
```
eric-wieser:
```suggestion
"one does not know whether sequences such as $(e^n)$, $(π^n)$, or even $(\frac{3}{2}^n)$"
```

FormalConjectures/Wikipedia/FermatCatalanConjecture.lean
```diff
@@ -45,18 +45,18 @@ def FermatCatalanConjecture : Prop :=
 
 /--
 The **Fermat–Catalan conjecture** states that the equation
-`a^m + b^n = c^k` has only finitely many solutions `(a,b,c,m,n,k)` with distinct triplets of values
-`(a^m, b^n, c^k)` where `a, b, c` are positive coprime integers and `m, n, k` are positive integers satisfying
-`1/m + 1/n + 1/k < 1`.
+$a^m + b^n = c^k$ has only finitely many solutions $(a,b,c,m,n,k)$ with distinct triplets of values
+$(a^m, b^n, c^k)$ where $a, b, c$ are positive coprime integers and $m, n, k$ are positive integers satisfying
+$\frac 1 m + \frac 1 n + \frac 1 k < 1$.
```
eric-wieser, line 50: Does this work? I didn't think latex used space separated arguments. 

Test: $\frac 1 n$.

mo271: It even works without the space (but looks a bit ugly): $\frac1n$

FormalConjectures/Books/UniformDistributionOfSequences/Equidistribution.lean
mo271: done (without the extra curly braces)

PR #8, opened 2025-05-13
ci: correct target in `Prepare Documentation`

mo271:


mo271:
Should it be enough to run `lake update` without any arguments here?

eric-wieser:
Why did this work before?

PR #10, opened 2025-05-15
clarify solution for  Erdős 198

### LINKED ISSUE #4: Update Erdős 198
with the new information that was added after contacting Thomas Bloom.

mo271:
The answer on the website was updated from "yes" to "no" by Thomas Bloom after we reported the counter example that AlphaProof found.

Fixes #4

### DIFF
--- a/FormalConjectures/ErdosProblems/198.lean
+++ b/FormalConjectures/ErdosProblems/198.lean
@@ -21,11 +21,11 @@ import FormalConjectures.Util.ProblemImports
 
 *Reference:* [erdosproblems.com/198](https://www.erdosproblems.com/198)
 -/
-open Function Set
+open Function Set Nat
 
 /-- Let $V$ be a vector space over the rationals and let $k$ be a fixed
-positive integer. Then there is a set $X_k ⊆ Y$ such that $X_k$, meets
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/198.lean
```diff
@@ -46,42 +46,41 @@ lemma baumgartner_headline (V : Type*) [AddCommGroup V] [Module ℚ V] :
...
 -/
 @[category research solved, AMS 5, AMS 11]
-theorem erdos_198
-    (A : ℕ →o ℕ)
-    (hA : IsSidon A) :
-    ∃ᵉ (a ∈ (range A)ᶜ) (d ≠ 0), ∀ n : ℕ, a + n • d ∈ (range A)ᶜ := by
+theorem erdos_198.variant_concrete :  ∃ A : ℕ →o ℕ,
+    (∀ n, A n = n ! + n) ∧
+    IsSidon A ∧ (∀ Y, IsAPOfLength Y ⊤ → ((range A) ∩ Y).Nonempty) := by
   sorry
```
mo271, line 86: Perhaps it would be a good idea to add the cleaned-up proof of this here?

PR #14, opened 2025-05-16
fix misformalization: add missing factor 1/2

mo271:
Without this missing factor, the statement is false for p = 1/6 and k = 1, i.e. m = 2.


FormalConjectures/Arxiv/0911.2077/Conjecture6_3.lean
```diff
@@ -47,4 +47,4 @@ theorem arxiv.id0911_2077.conjecture6_3
   (σ : ℝ) (h_σ : σ = (p * (1 - p)).sqrt) :
   ((PMF.binomial (.ofReal p : ℝ≥0∞) hp' (2 * k)).toMeasure (Set.Ici k)).toReal
  ≥ 1 - Φ ((1 / 2 - p) * sqrt (2 * k : ℝ≥0) / σ)
-    + ((2 * k).choose k) * σ.rpow (2 * k) := sorry
+    + (1 / 2) * ((2 * k).choose k) * σ.rpow (2 * k) := sorry
```
eric-wieser: Why are we using `.rpow` here?

mo271: we shoudn't, replaced with hPow.

PR #15, opened 2025-05-17
remove category_stats command

mo271:


mo271:
Currently when everything is in the environment this looks like this:
```
Current problem subject statistics:
 Number theory: 266
 Combinatorics: 100
 Special functions: 29
 Field theory and polynomials: 17
 Algebraic geometry: 14
 Functions of a complex variable: 6
 Associative rings and algebras: 6
 Commutative algebra: 6
 General algebraic systems: 5
 Geometry: 3
 Linear and multilinear algebra; matrix theory: 3
 Probability theory and stochastic processes: 1
 Order, lattices, ordered algebraic structures: 1
 Group theory and generalizations: 1
 Real functions: 1
 General topology: 1
 Approximations and expansions: 1
 Sequences, series, summability: 1
```

FormalConjectures/Util/Attributes.lean
```diff
@@ -315,9 +318,37 @@ elab "#category_stats" : command => do
...
+
+  let sortedCounts := counts.toArray.qsort (lt := fun (_, c1) (_, c2) => c2 < c1)
+
+  Elab.Command.liftCoreM do
+    let mut statLines : Array String := #[]
+    for (subject, count) in sortedCounts do
+        -- Ensure we only print non-zero counts
+        if count > 0 then
+            let desc ←  subject.getDesc
+            statLines := statLines.push s!" {desc}: {count}"
```
eric-wieser: Nit: can you make this indent 2 spaces, and do the same for the other stat command?

mo271: not adding this code anymore...

mo271:
Yeah, let's get rid of this code duplication. I will add only what it needed in #16

PR #16, opened 2025-05-18
add subject stats to docs welcome page

mo271:
and restructure the welcome page a bit

### DIFF
--- a/docbuild/scripts/overwrite_index.lean
+++ b/docbuild/scripts/overwrite_index.lean
@@ -109,15 +109,12 @@ unsafe def main (args : List String) : IO Unit := do
 overwrites the contents of the `main` tag of a html `file` with a weclome page including stats."
   let inputHtmlContent ← IO.FS.readFile file
 
-  let (categoryStats, subjectStats) ← runWithImports do
-    let catStats ← getCategoryStatsMarkdown
-    let subjStats ← getSubjectStatsMarkdown
-    return (catStats, subjStats)
-    -- Or more concisely, as you suggested:
-    -- return (← getCategoryStatsMarkdown, ← getSubjectStatsMarkdown)
+  runWithImports do
+    let categoryStats ← getCategoryStatsMarkdown
+    let subjectStats ← getSubjectStatsMarkdown
 
-  let markdownBody :=
-    s!"# Welcome to the *Formal Conjectures* Documentation!
+    let markdownBody :=
+      s!"# Welcome to the *Formal Conjectures* Documentation!
 
 Check out the main
 [Formal Conjectures GitHub repository](https://github.com/google-deepmind/formal-conjectures)
@@ -137,7 +134,7 @@ classifications, please refer to the
 
 ## Subject Category Statistics
 {subjectStats}"
-  IO.println markdownBody
-  let .some newBody := MD4Lean.renderHtml (parserFlags := MD4Lean.MD_FLAG_TABLES ) markdownBody | throw <| .userError "Parsing failed"
-  let finalHtml ← replaceTag "main" inputHtmlContent newBody
-  IO.FS.writeFile file finalHtml
+    IO.println markdownBody
+    let .some newBody := MD4Lean.renderHtml (parserFlags := MD4Lean.MD_FLAG_TABLES ) markdownBody | throwError "Parsing failed"
+    let finalHtml ← replaceTag "main" inputHtmlContent newBody
+    IO.FS.writeFile file finalHtml


mo271:
is based on #15

mo271:
preview here:
https://github.com/google-deepmind/formal-conjectures/actions/runs/15096578262/attempts/1#summary-42432034236
Although  not exactly: the awkward linebreaks do not show up in when generating the index.html and looking at it locally.

FormalConjectures/Util/Attributes.lean
```diff
@@ -315,9 +318,37 @@ elab "#category_stats" : command => do
...
-  Lean.logInfo ("Current benchmark stats:\n" ++ out)
+  Lean.logInfo ("Benchmark statistics:\n" ++ out)
+
+/-- Prints the number of problems per AMS subject (when non-zero).
 
--- TODO(lezeau): add a `#subject_stats` command that does
--- prints the number of problems per subject (when non-zero)
+Note that this will depend on what declarations are present in the
+environment at the place where the command is called. -/
+elab "#subject_stats" : command =>  do
```
eric-wieser: The rest of the PR didn't need this, right? I'd be inclined to drop it.

mo271: dropping it in #15

mo271:
now not based on #15 anymore

docbuild/scripts/overwrite_index.lean
```diff
@@ -65,7 +85,7 @@ def replaceTag (tag : String) (inputHtmlContent : String) (newContent : String)
   return finalHtml
 
 
-unsafe def fetchStatsMarkdown : IO String := do
+unsafe def fetchStatsMarkdown (markdownGenerator : CoreM String ) : IO String := do
```
eric-wieser:
```suggestion
unsafe def fetchStatsMarkdown (markdownGenerator : CoreM String) : IO String := do
```

docbuild/scripts/overwrite_index.lean
```diff
@@ -77,23 +97,39 @@ unsafe def fetchStatsMarkdown : IO String := do
...
-
 unsafe def main (args : List String) : IO Unit := do
   let .some file := args.get? 0
     | IO.println "Usage: stats <file>
 overwrites the contents of the `main` tag of a html `file` with a weclome page including stats."
-
   let inputHtmlContent ← IO.FS.readFile file
-  let statsString ← fetchStatsMarkdown
+  let categoryStats ← fetchStatsMarkdown getCategoryStatsMarkdown
+  let subjectStats ← fetchStatsMarkdown getSubjectStatsMarkdown
```
eric-wieser: I think this is likely to go wrong; you should only call  `withImportModules` once.

Perhaps
```suggestion
  let (categoryStats, subjectStats) ← fetchStatsMarkdown do
    return (← getCategoryStatsMarkdown, ← getSubjectStatsMarkdown)
```

docbuild/scripts/overwrite_index.lean
eric-wieser: Though probably you want to give this an arbitrary type parameter, and call it `runWithImports` or similar.

mo271:
https://github.com/google-deepmind/formal-conjectures/actions/runs/15226936516/attempts/1#summary-42830124443

docbuild/scripts/overwrite_index.lean
mo271: I gave what you suggested a try and it seems work

docbuild/scripts/overwrite_index.lean
```diff
@@ -64,8 +84,11 @@ def replaceTag (tag : String) (inputHtmlContent : String) (newContent : String)
   let finalHtml := htmlPrefix ++ newContent ++ htmlSuffix
   return finalHtml
 
-
-unsafe def fetchStatsMarkdown : IO String := do
+/--
+Runs a `CoreM α` action in an environment where specified modules are imported.
```
eric-wieser:
```suggestion
Runs a `CoreM α` action in an environment where all FormalConjectures modules are imported.
```

docbuild/scripts/overwrite_index.lean
```diff
@@ -77,23 +100,43 @@ unsafe def fetchStatsMarkdown : IO String := do
...
     | IO.println "Usage: stats <file>
 overwrites the contents of the `main` tag of a html `file` with a weclome page including stats."
-
   let inputHtmlContent ← IO.FS.readFile file
-  let statsString ← fetchStatsMarkdown
+
+  let (categoryStats, subjectStats) ← runWithImports do
+    let catStats ← getCategoryStatsMarkdown
+    let subjStats ← getSubjectStatsMarkdown
+    return (catStats, subjStats)
```
eric-wieser: You could also just do all the work inside this do block, to avoid packing and unpacking the results.

FormalConjectures/Util/Attributes/AMS.lean
```diff
@@ -170,6 +170,11 @@ def numToAMSName (n : Nat) : MetaM Name := do
   unless !(← Lean.hasConst nm) do return nm
   throwError "Out of bounds"
 
+def AMS.getDesc (a : AMS) : CoreM String := do
+  let .const n [] := Lean.toExpr a | throwError "this shouldn't happen"
+  let .some doc := ← Lean.findDocString? (← getEnv) n | throwError s!"please add docstring to all AMS categories"
```
eric-wieser:
```suggestion
  let .some doc := ← Lean.findDocString? (← getEnv) n | throwError m!"{.ofConstName n} is missing a docstring"
```

docbuild/scripts/overwrite_index.lean
mo271: done

PR #20, opened 2025-05-26
change names starting with numbers

mo271:
Without this and the fix https://github.com/leanprover/doc-gen4/pull/282, which will only come when we bump to v4.20, there will be broken links in the doc-gen4 build.


mo271:
closing in favor of #21

PR #26, opened 2025-05-27
chore: Clarify terminology for solved research problems

Paul-Lez:


FormalConjectures/Util/Attributes.lean
```diff
@@ -92,7 +95,7 @@ inductive ProblemStatus
   /-- Indicates that a mathematical problem is still open. -/
   | open
   /-- Indicates that a mathematical problem is already solved,
-  i.e. there is a pusblished (informal) proof. -/
+  i.e. there is a published widely accepted by experts (informal) proof. -/
```
mo271:
```suggestion
  i.e., there is a published (informal) proof that is widely accepted by experts. -/
```

PR #30, opened 2025-05-27
feat: add formalization of Erdos 488

callesonne:


FormalConjectures/ErdosProblems/488.lean
```diff
@@ -0,0 +1,38 @@
+/-
+Copyright 2025 Google LLC
```
mo271:
```suggestion
Copyright 2025 The Formal Conjectures Authors.
```

FormalConjectures/ErdosProblems/488.lean
```diff
@@ -0,0 +1,38 @@
...
+open Classical
+
+/--
+Let $A$ be a finite set and
+$$B = \{n \ge 1 : a \nmid n \text{ for all } a \in A\}.$$
+Is it true that, for every $m > n \ge \max(A)$,
+$$\frac{|B \cap [1, m]|}{m} < 2 \frac{|B \cap [1, n]|}{n}?$$
+-/
+@[category research solved, AMS 5, AMS 11]
+theorem erdos_488 (A : Finset ℕ) (B : Set ℕ) (hB : B = {n ≥ 1 | ∀ a ∈ A, ¬ a ∣ n}) :
```
Paul-Lez: The problem is stated as a question so should be formalised using `answer(sorry)`.

FormalConjectures/ErdosProblems/488.lean
```diff
@@ -0,0 +1,38 @@
...
+
+open Classical
+
+/--
+Let $A$ be a finite set and
+$$B = \{n \ge 1 : a \nmid n \text{ for all } a \in A\}.$$
+Is it true that, for every $m > n \ge \max(A)$,
+$$\frac{|B \cap [1, m]|}{m} < 2 \frac{|B \cap [1, n]|}{n}?$$
+-/
+@[category research solved, AMS 5, AMS 11]
```
Paul-Lez: This should work
```suggestion
@[category research solved, AMS 5 11]
```

Paul-Lez (approved):

FormalConjectures/ErdosProblems/488.lean
harahu: I know this PR is merged, but I was a bit surprised by this. I agree that it is interesting to state the original question by Erdos. But since the problem is solved and an answer to the question exists, is it worth additionally stating either the answer to the question or the more specific theorem providing this answer? 

If this is only food for auto-formalization tasks anyway, one might as well milk the problem page for all interesting statements that could be formalized.

Paul-Lez: Good question! I typed the previous comment a little too quickly: what I meant to say is that we should be using `answer(False)`  here since the question has been answered to the negative; and `answer(sorry)` is indeed reserved for questions that have not been answered to the positive or the negative.

Paul-Lez: Fixed in #148 - thanks for flagging this:)

PR #32, opened 2025-05-27
fix: erdos 418

b-mehta:


FormalConjectures/ErdosProblems/418.lean
```diff
@@ -20,15 +20,16 @@ import FormalConjectures.Util.ProblemImports
...
 -/
 open scoped ArithmeticFunction
 
 /--
 Are there infinitely many integers not of the form $n - \phi(n)$?
 -/
 @[category research solved, AMS 11]
-theorem erdos_418 :
-    { (n - n.totient : ℕ) | n }ᶜ |>.Infinite :=
+theorem erdos_418 : { (n - n.totient : ℕ) | n }ᶜ.Infinite :=
```
b-mehta, line 32: style change

FormalConjectures/ErdosProblems/418.lean
```diff
@@ -37,11 +38,9 @@ written as $n - \phi(n)$.
 -/
 @[category research solved, AMS 11]
 theorem erdos_418.variants.conditional
-    (goldbach : ∀ (n : ℕ) (_ : 2 < n) (_ : Even n),
-      ∃ p q, Prime p ∧ Prime q ∧ n = p + q)
-    (m : ℕ)
-    (h : Odd m) :
-    ∃ n, m + n.totient = n :=
+    (goldbach : ∀ (n : ℕ), 2 < n → Even n → ∃ p q, Prime p ∧ Prime q ∧ n = p + q)
```
b-mehta, line 41: style change

FormalConjectures/ErdosProblems/418.lean
```diff
@@ -52,7 +51,7 @@ $\sigma(n) - n$.
 -/
 @[category research solved, AMS 11]
 theorem erdos_418.variants.sigma :
-    ∃ (S : Set ℕ) (hS : ∀ α, S.HasDensity α → 0 < α),
+    ∃ (S : Set ℕ) (hS : ∃ α, 0 < α ∧ S.HasDensity α),
```
b-mehta: Correctness change: the old formulation would have allowed for S to be a set which doesn't have a density (these things exist) as well. The new formulation insists that S has a density and that density is positive.

FormalConjectures/ErdosProblems/418.lean
```diff
@@ -74,6 +73,6 @@ not of the form $n - \phi(n)$.
 -/
 @[category research open, AMS 11]
 theorem erdos_418.variants.density :
-    ∃ (S : Set ℕ) (hS : ∀ α, S.HasDensity α → 0 < α),
+    ∃ (S : Set ℕ) (hS : ∃ α, 0 < α ∧ S.HasDensity α),
```
b-mehta: Correctness change: the old formulation would have allowed for S to be a set which doesn't have a density (these things exist) as well. The new formulation insists that S has a density and that density is positive.

FormalConjectures/ErdosProblems/418.lean
callesonne:
```suggestion
    ∃ (S : Set ℕ) (hS : S.HasPosDensity),
```
This problem must have been formalized before I added this ^

FormalConjectures/ErdosProblems/418.lean
callesonne:
```suggestion
    ∃ (S : Set ℕ) (hS : S.HasPosDensity),
```

mo271:
(I just repushed, not chaning the authors but changing Calle's email to make the CLA bot happy)

callesonne (approved):
LGTM, thanks!

b-mehta:
I think AMS 11 seems like the wrong tag if it is field theory and polynomials...

Paul-Lez:
> I think AMS 11 seems like the wrong tag if it is field theory and polynomials...

I think this is an issue with the GitHub labels - 11 corresponds to Number Theory in the [classification](https://mathscinet.ams.org/mathscinet/msc/pdfs/classifications2020.pdf). I'll fix this now.

eric-wieser:
Thanks @b-mehta!

PR #35, opened 2025-05-27
doc: clarification on problems that require answers

Paul-Lez:


FormalConjectures/Util/Answer.lean
```diff
@@ -23,6 +23,11 @@ import Batteries.Lean.Expr
 # The `answer( )` elaborator
 
 This file provides syntax for marking up answers in a problem statement.
+
+Note: certain problems also require guessing an answer, and can be formalised
```
eric-wieser:
```suggestion
Note: certain problems also require providing an answer, and can be formalised
```

FormalConjectures/Util/Answer.lean
```diff
@@ -23,6 +23,11 @@ import Batteries.Lean.Expr
...
+using `answer(sorry)` as a placeholder. In that case,
+**solving the problem is not the same as providing an answer and a proof**:
```
eric-wieser: Perhaps something more like:

> while providing a proof simply requires finding any way to replace `:= sorry`,
> providing an _answer_ is **not** just finding any way to replace `answer(sorry)`;
> it requires evaluation of mathematical meaning,
> which is a job for human mathematicians, not Lean alone.

README.md
```diff
@@ -184,10 +184,25 @@ elaborator allows us to formulate the problem without deciding for an answer.
...
 theorem HadwigerNelsonProblem :
-    IsLeast { n : ℕ | ExistsColoring n } answer(sorry) :=
+    IsLeast {n : ℕ | ExistsColoring n} answer(sorry) :=
   sorry
 ```
 
+## Problems that require answers
+
+Note that providing a term inside the `answer( )` elaborator together with a proof that
+the statement is true *does not* mean that the problem has been solved. For example, a question
```
eric-wieser:
```suggestion
the statement is true *does not* by itself mean that the problem has been solved. For example, a question
```

FormalConjectures/Util/Answer.lean
```diff
@@ -23,6 +23,12 @@ import Batteries.Lean.Expr
 # The `answer( )` elaborator
 
 This file provides syntax for marking up answers in a problem statement.
+
+Note: certain problems also providing an answer, and can be formalised
+using `answer(sorry)` as a placeholder. While providing a proof simply requires
+finding any way to replace `:= sorry`, providing an answer is not just finding
+any way to replace answer(sorry): it requires evaluation of mathematical meaning,
+which is a job for human mathematicians, not Lean.
```
eric-wieser:
```suggestion
which is a job for human mathematicians, not Lean alone.
```
since I guess lean can tell you that `"I don't know"` is not a legal answer when a real is expected.

eric-wieser:
I'll let @mo271 give this a final review.

README.md
```diff
@@ -184,10 +184,25 @@ elaborator allows us to formulate the problem without deciding for an answer.
...
-    IsLeast { n : ℕ | ExistsColoring n } answer(sorry) :=
+    IsLeast {n : ℕ | ExistsColoring n} answer(sorry) :=
   sorry
 ```
 
+## Problems that require answers
+
+Note that providing a term inside the `answer( )` elaborator together with a proof that
+the statement is true *does not* by itself mean that the problem has been solved. For example, a question
+of the form "Find all natural numbers that satisfy the predicate $P$" might be formalised as
```
eric-wieser: For research questions, perhaps this should be
```suggestion
of the form "Which natural numbers satisfy the predicate $P$?" might be formalised as
```

PR #39, opened 2025-05-27
feat: Formalise erdos_12

Paul-Lez:


FormalConjectures/ErdosProblems/12.lean
```diff
@@ -0,0 +1,77 @@
...
+*Reference:* [erdosproblems.com/12](https://www.erdosproblems.com/12)
+-/
+
+open Classical
+
+-- TODO: add other statements from the file
+
+/--
+A set `A` is "good" if it is infinite and there are no distinct `a,b,c` in `A`
+such that `a∣(b+c)` and `b > a`, `c > a`.
```
callesonne:
```suggestion
such that `a ∣ (b+c)` and `b > a`, `c > a`.
```

FormalConjectures/ErdosProblems/12.lean
```diff
@@ -0,0 +1,77 @@
...
+-/
+abbrev IsGood (A : Set ℕ) : Prop := A.Infinite ∧
+  ∀ᵉ (a ∈ A) (b ∈ A) (c ∈ A), a ∣ b + c → a < b →
+  a < c → b = c
+
+/-- The set of $p ^ 2$ where $p \cong 3 \mod 4$ is prime is an example of a good set. -/
+@[category test, AMS 11]
```
callesonne: Is this really a test if you have not proven it? Maybe the label `undergraduate` is more suitable (although I have no idea how easy / hard this is, I imagine its applying quadratic reciprocity somehow so `undergraduate` seems like a good guess)

FormalConjectures/ErdosProblems/12.lean
```diff
@@ -0,0 +1,77 @@
...
+/--
+Let $A$ be an infinite set such that there are no distinct $a,b,c \in A$
+such that $a \mid (b+c)$ and $b,c > a$. Is it true that $∑_{n \in A} \frac{1}{n} < \infty$?
+-/
+@[category research open, AMS 11]
+theorem erdos_12.parts.iii :
+    (∀ (A : Set ℕ), IsGood A → Summable (fun (n : A) ↦ (1 / n : ℝ))) ↔ answer(sorry) := by
+  sorry
+
+/-- Erdős and Sárközy proved that such an $A$ must have density 0. -/
```
callesonne: Could you also add the reference here?

FormalConjectures/ErdosProblems/12.lean
```diff
@@ -0,0 +1,77 @@
...
+
+/-!
+# Erdős Problem 12
+
+*Reference:* [erdosproblems.com/12](https://www.erdosproblems.com/12)
+-/
+
+open Classical
+
+-- TODO: add other statements from the file
```
callesonne: Could you also create a corresponding issue stating this (unless you have already!)

FormalConjectures/ErdosProblems/12.lean
```diff
@@ -0,0 +1,77 @@
...
+  sorry
+
+/--
+Let $A$ be an infinite set such that there are no distinct $a,b,c \in A$
+such that $a \mid (b+c)$ and $b,c > a$. Does there exist some absolute constant $c > 0$
+such that there are always infinitely many $N$
+with $|A \cap \{1, \dotsc, N\}| < N^{1−c}$?
+-/
+@[category research open, AMS 11]
+theorem erdos_12.parts.ii : (∀ (A : Set ℕ), IsGood A → ∃ c > (0 : ℝ),
```
mo271: I read (perhaps superficially) the meaning "absolute constant" as not dependent on `A`, hence I would change the quantifier order. Not super sure though.

FormalConjectures/ErdosProblems/12.lean
b-mehta: it's not hard: you just need that -1 isn't a square mod p, so `undergraduate` seems reasonable

FormalConjectures/ErdosProblems/12.lean
```diff
@@ -0,0 +1,77 @@
...
+
+/--
+Let $A$ be an infinite set such that there are no distinct $a,b,c \in A$
+such that $a \mid (b+c)$ and $b,c > a$. Is there such an $A$ with
+$\liminf \frac{|A \cap \{1, \dotsc, N\}|}{N^{1/2}} > 0$ ?
+-/
+@[category research open, AMS 11]
+theorem erdos_12.parts.i : (∃ (A : Set ℕ), IsGood A ∧
+    (0 : ℝ) < Filter.atTop.liminf
+      (fun N => ((Finset.Icc 1 N).filter (· ∈ A)).card / (N : ℝ).sqrt)) ↔ answer(sorry) := by
```
b-mehta: i wonder if you want a def for this numerator

FormalConjectures/ErdosProblems/12.lean
b-mehta: i agree

Paul-Lez: Agreed! Swapped the quantifier order as suggested

FormalConjectures/ErdosProblems/12.lean
Paul-Lez: Added a def

FormalConjectures/ErdosProblems/12.lean
Paul-Lez: Added a comment to the Erdos Problem 12 issue

PR #54, opened 2025-05-27
chore: fix naming and missing hypothesis in mathoverflow problem

callesonne:
See issue #53

PR #55, opened 2025-05-27
feat: use `EReal` for sup in mathoverflow_347178

callesonne:
This depends on #54, whilst looking into the problem it seems more accurate to me if the main problem takes the sup in EReal, since it does not seem to me like the supremums will necessarily be bounded.

### DIFF
--- a/FormalConjectures/Mathoverflow/347178.lean
+++ b/FormalConjectures/Mathoverflow/347178.lean
@@ -31,8 +31,7 @@ $$\sup_{x\in {\bf R}^n}f(x)=\sup_{x\in {\bf R}^n}f(x+\nabla f(x))$$?
 @[category research open]
 theorem mathoverflow_347178 {n : ℕ} (hn : 2 ≤ n) (f : EuclideanSpace ℝ (Fin n) → ℝ)
     (hf : ContDiff ℝ 1 f) :
-    (BddAbove (range f) ↔ BddAbove (range (fun x ↦ f (x + gradient f x)))) ∧
-    (⨆ x, f x) = ⨆ x, f (x + gradient f x) := sorry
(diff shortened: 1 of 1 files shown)


FormalConjectures/Mathoverflow/347178.lean
```diff
@@ -31,16 +31,15 @@ $$\sup_{x\in {\bf R}^n}f(x)=\sup_{x\in {\bf R}^n}f(x+\nabla f(x))$$?
 @[category research open]
 theorem mathoverflow_347178 {n : ℕ} (hn : 2 ≤ n) (f : EuclideanSpace ℝ (Fin n) → ℝ)
     (hf : ContDiff ℝ 1 f) :
-    (BddAbove (range f) ↔ BddAbove (range (fun x ↦ f (x + gradient f x)))) ∧
-    (⨆ x, f x) = ⨆ x, f (x + gradient f x) := sorry
+    (⨆ x, (f x : EReal)) = ⨆ x, (f (x + gradient f x) : EReal) := sorry
```
eric-wieser: Is `iSup` on `EReal` well-behaved?

callesonne: I think so.
`instance : SupSet EReal := inferInstanceAs (SupSet (WithBot (WithTop ℝ)))`
and `WithBot a` gets a `SupSet` instance if `a` has one. And `SupSet (WithTop a)` is:
`⟨fun S =>
    if ⊤ ∈ S then ⊤ else if BddAbove ((fun (a : α) ↦ ↑a) ⁻¹' S : Set α) then
      ↑(sSup ((fun (a : α) ↦ (a : WithTop α)) ⁻¹' S : Set α)) else ⊤⟩`

eric-wieser (approved):

PR #56, opened 2025-05-27
Correct conditional version of Erdos 418

b-mehta:
After discussion with @TFBloom it seems that this was the intention: the argument suggested only seems to work for distinct primes. The slight strengthening is expected to be true with about the same certainty as the main conjecture itself, although it is technically still a strengthening.

Also fill in the sorry.

### DIFF
--- a/FormalConjectures/ErdosProblems/418.lean
+++ b/FormalConjectures/ErdosProblems/418.lean
@@ -35,6 +35,9 @@ theorem erdos_418 : { (n - n.totient : ℕ) | n }ᶜ.Infinite :=
 /--
 It follows from a slight strengthening of the Goldbach conjecture that every odd number can be
 written as $n - \phi(n)$.
+In particular, we assume that every even number greater than 6 can be written as the sum of two
+*distinct* primes, in contrast to the usual Goldbach conjecture that every even number greater than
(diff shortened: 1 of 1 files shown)


Paul-Lez (commented):
Nice! Would you mind specifying the precise formulation of the strengthening in the docstring? (LGTM modulo that)

FormalConjectures/ErdosProblems/418.lean
```diff
@@ -33,15 +33,30 @@ theorem erdos_418 : { (n - n.totient : ℕ) | n }ᶜ.Infinite :=
   sorry
 
 /--
-It follows from the Goldbach conjecture that every odd number can be
+It follows from a slight strengthening of the Goldbach conjecture that every odd number can be
 written as $n - \phi(n)$.
```
b-mehta, line 37:
```suggestion
It follows from a slight strengthening of the Goldbach conjecture that every odd number can be
written as $n - \phi(n)$.
In particular, we assume that every even number greater than 6 can be written as the sum of two
*distinct* primes, in contrast to the usual Goldbach conjecture that every even number greater than
2 can be written as the sum of two primes.
```

callesonne (approved):
Thanks!

PR #58, opened 2025-05-28
feat(Erdos/913): fill in a sorry

b-mehta:


eric-wieser:
This is perhaps on the borderline regarding this section of the readme:

> While the main goal is to collect conjecture statements, we appreciate the inclusion of very short proofs for solved items or counterexamples, especially if they are illuminating and testing the definitions. Lengthy proofs are outside the scope of this repository.

b-mehta:
That's fair, your call.

mo271:
I'd be in favor of including this for now, if the maintenance become too much of a burden, let's remove it again.

Thanks in any case, @b-mehta!

b-mehta:
I think I'm counted as an external maintainer, so the maintenance burden can be dropped on my shoulders :)

PR #72, opened 2025-05-28
fix: follow the answer(sorry) convention for Landau problems

Paul-Lez:


FormalConjectures/LandauProblems/LegendreConjecture.lean
```diff
@@ -26,6 +26,7 @@ import FormalConjectures.Util.ProblemImports
 Does there always exist at least one prime between consecutive perfect squares?
 -/
 @[category research open, AMS 11]
-theorem legendre_conjecture (n : ℕ) (hn : 1 ≤ n) :
-    ∃ p ∈ Set.Ioo (n^2) ((n+1)^2), Prime p := by
+theorem legendre_conjecture :
+    (∀ᵉ (n ≥ 1), ∃ p ∈ Set.Ioo (n^2) ((n+1)^2), Prime p)
```
mo271, line 30: does the `∀ᵉ` requiere the switch from `1 ≤ n` to `n ≥ 1`?

mo271 (approved):

FormalConjectures/LandauProblems/LegendreConjecture.lean
Paul-Lez: It does indeed, unfortunately.

PR #89, opened 2025-05-28
MO question: right local semirings

alreadydone:
[A question by myself](https://mathoverflow.net/questions/486451). The first part is implied to be solved in the literature, but no example or proof is given, and I can't reach the authors. The second part is new as far as I know.

### DIFF
--- a/FormalConjectures/Mathoverflow/486451.lean
+++ b/FormalConjectures/Mathoverflow/486451.lean
@@ -23,12 +23,15 @@ asked by user [*Junyan Xu*](https://mathoverflow.net/users/3332/junyan-xu)
 -/
 
 @[category research solved, AMS 16]
+/-- There exists a semiring with a unique left maximal ideal but more than one right maximal ideals. -/
 theorem exists_semiring_unique_left_maximal_not_unique_right_maximal :
     ∃ (R : Type) (_ : Semiring R), (∃! I : Ideal R, I.IsMaximal) ∧
       ∃ I J : Ideal Rᵐᵒᵖ, I.IsMaximal ∧ J.IsMaximal ∧ I ≠ J := by
   sorry
 
 @[category research open, AMS 16]
+/-- There exists a semiring with a unique left maximal ideal and a unique right maximal ideal
+which are not the same as sets. -/
 theorem exists_semiring_unique_left_right_maximal_ne :
     ∃ (R : Type) (_ : Semiring R), (∃! I : Ideal R, I.IsMaximal) ∧
       (∃! J : Ideal Rᵐᵒᵖ, J.IsMaximal) ∧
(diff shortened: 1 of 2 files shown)


FormalConjectures/Mathoverflow/486451.lean
```diff
@@ -0,0 +1,36 @@
...
+
+import FormalConjectures.Util.ProblemImports
+/-!
+# Mathoverflow 486451
+
+*Reference:* [mathoverflow/486451](https://mathoverflow.net/questions/486451)
+asked by user [*Junyan Xu*](https://mathoverflow.net/users/3332/junyan-xu)
+-/
+
+@[category research solved, AMS 16]
```
Paul-Lez, line 26: Would you mind adding docstrings with the informal statement for both your theorems?

Paul-Lez (commented):
Thanks for the PR! One minor comment for now - I'd do another pass to take a look at the maths later today

FormalConjectures/Mathoverflow/486451.lean
```diff
@@ -0,0 +1,39 @@
...
+/-- There exists a semiring with a unique left maximal ideal but more than one right maximal ideals. -/
+theorem exists_semiring_unique_left_maximal_not_unique_right_maximal :
+    ∃ (R : Type) (_ : Semiring R), (∃! I : Ideal R, I.IsMaximal) ∧
+      ∃ I J : Ideal Rᵐᵒᵖ, I.IsMaximal ∧ J.IsMaximal ∧ I ≠ J := by
+  sorry
+
+@[category research open, AMS 16]
+/-- There exists a semiring with a unique left maximal ideal and a unique right maximal ideal
```
Paul-Lez, line 32: This is looking good! 

Out of curiosity, is the expectation that such a semiring exists? If this is unclear, using the answer(sorry) formulation would make sense here.

alreadydone: Yeah I'm not so sure about this. Changed to answer(sorry).

FormalConjectures/Mathoverflow/486451.lean
```diff
@@ -0,0 +1,39 @@
+/-
+Copyright 2025 Google LLC
```
Paul-Lez:
```suggestion
Copyright 2025 The Formal Conjectures Authors
```
(and you can add yourself to the AUTHORS file :))

alreadydone: Sorting by first name is a bit unusual :)

Paul-Lez: Good point :D

mo271:
rebased so that the new copyright header check is there.

Thanks for the pull request, nice question!

PR #98, opened 2025-05-29
ci: add a copyright header check

mo271:
This works only on the .lean files.


scripts/check_copyright_header.sh
```diff
@@ -0,0 +1,92 @@
...
+
+# Use process substitution to avoid creating a subshell for the while loop.
+# This ensures that the `files_checked` variable is correctly incremented in the main script.
+while IFS= read -r -d $'\0' file; do
+  files_checked=$((files_checked+1))
+
+  # Use `grep` for Linux environments like GitHub Actions.
+  # (Locally on macOS, you would use `ggrep`).
+  if ! grep -qPz "$expected_pattern" "$file"; then
+    echo "❌ Error: The file '$file' does not start with a valid copyright header."
```
eric-wieser:
```suggestion
    echo "::error file=$file,line=1,col=1,title=No copyright header::The file '$file' does not start with a valid copyright header"
```

eric-wieser: This will result in an annotation in the diff view.

.github/workflows/check_copyright_header.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  check-copyright:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out repository code
+      uses: actions/checkout@v4
+
+    - name: Run copyright check script
+      run: |
+        chmod +x ./scripts/check_copyright_header.sh
```
eric-wieser: You should do this in git, not in this script.

mo271: removed (it already had the correct permissions...)

mo271:
will need a little change after #99, I don't care which one is mreged first

eric-wieser (approved):

PR #103, opened 2025-05-29
feat: main conjecture on fusible numbers

alreadydone:
A conjecture by myself in a research paper.


FormalConjectures/Paper/FusibleNumber.lean
```diff
@@ -0,0 +1,46 @@
...
+import FormalConjectures.Util.ProblemImports
+/-!
+# Main conjecture on fusible numbers
+
+*Reference:* [Fusible numbers and Peano Arithmetic](https://arxiv.org/abs/2003.14342),
+by Jeff Erickson, Gabriel Nivasch, and Junyan Xu.
+Logical Methods in Computer Science, Volume 18, Issue 3 (July 28, 2022) lmcs:8555
+-/
+
+inductive IsFusible : ℚ → Prop
```
Paul-Lez, line 39: Would you mind adding a docstring for this definition?

FormalConjectures/Paper/FusibleNumber.lean
```diff
@@ -0,0 +1,46 @@
...
+  | zero : IsFusible 0
+  | fuse (a b : ℚ) : IsFusible a → IsFusible b → |a - b| < 1 → IsFusible ((a + b + 1) / 2)
+
```
Paul-Lez, line 42: I think it could also be nice to add a few "test"/API statements about this definition, e.g. proving certain known numbers are fusible, and basic results (which can be left sorried out:))

Paul-Lez (commented):
Nice!

FormalConjectures/Paper/FusibleNumber.lean
mo271, line 1: Let's move it to the Arxiv folder.

FormalConjectures/Paper/FusibleNumber.lean
```diff
@@ -0,0 +1,46 @@
...
+This formalization differs from Conjecture 7.1 in the paper in four ways:
+(1) it is obtained from Conjecture 7.1 by plugging in `n + 1` into `n`, which simplifies the expressions
+  and removes the need to assume `n ≥ 1`;
+(2) the `n + 1`st successor `s^(n+1)(x)` is replaced by the explicit value `x + (2 - 1 / 2 ^ n) * m`;
+(3) instead of defining `y` to be the successor of `x`, we assert that there is no fusible number between `x` and `y`;
+(4) instead of using `∃ z, IsFusible z ∧ q = s^(n+1)(x) ~ z` we use the value of `z` determined by the equality,
+  namely `z = 2 * q - 1 - s^(n+1)(x)`, and it is easy to see `z ∈ [x + 1 - m / 2 ^ n, x + 1)` as required. -/
+@[category research open, AMS 05]
+theorem conj_7_1 (x y q : ℚ) (n : ℕ) (fus_x : IsFusible x) (fus_y : IsFusible y) (lt : x < y)
+    (nmem_Icc : ∀ z, IsFusible z → z ∉ Set.Icc x y) :
```
mo271: Something seems off here: isn't `nmem_Icc x fus_ x` the statement `x  ∉ Set.Icc x y`, which is always False, making this entire conjecture empty?
Should "instead of defining `y` to be the successor of `x`, we assert that there is no fusible number between `x` and `y`;" it be something like `(no_fus_between : ∀ z ∈ Set.Ioo x y, ¬ IsFusible z)`?

alreadydone: Good catch! It should indeed be Ioo, and switching order looks better too

FormalConjectures/Paper/FusibleNumber.lean
alreadydone: Should published papers be there?

mo271: ah, if it is published, then this place is fine, let's in addition to the arxiv link also add the doi link (
https://doi.org/10.46298/lmcs-18%283%3A6%292022) or direct journal link (https://lmcs.episciences.org/9850) perhaps?

mo271:
@alreadydone  I really like this contribution and it seems like almost all comments are addressed. Do you plan to finish it at some point?

mo271:
I think this is a great contribution, @alreadydone can we get it to state where we can submit it?

mo271 (approved):

PR #105, opened 2025-05-29
feat: add erdos 304

b-mehta:
Co-authored-by: bhavikmehta@google.com

### DIFF
--- a/FormalConjectures/ErdosProblems/304.lean
+++ b/FormalConjectures/ErdosProblems/304.lean
@@ -23,12 +23,112 @@ import FormalConjectures.Util.ProblemImports
 
 open Asymptotics Filter
 
+/--
+The set of `k` for which `a / b` can be expressed as a sum of `k` distinct unit fractions.
+-/
+def unitFractionExpressible (a b : ℕ) : Set ℕ :=
+  {k | ∃ s : Finset ℕ, s.card = k ∧ (∀ n ∈ s, n > 1) ∧ (a / b : ℚ) = ∑ n ∈ s, (n : ℚ)⁻¹}
+
+@[category API, simp]
+lemma zero_mem_unitFractionExpressible_iff {a b : ℕ} :
+    0 ∈ unitFractionExpressible a b ↔ a = 0 ∨ b = 0 := by
+  simp_all [unitFractionExpressible]
+
+@[category API]
+lemma unitFractionExpressible_of_zero {a b : ℕ} (h : a = 0 ∨ b = 0) :
+    unitFractionExpressible a b = {0} := by
+  simp only [Set.eq_singleton_iff_unique_mem, zero_mem_unitFractionExpressible_iff, *]
+  have : (a / b : ℚ) = 0 := by simpa
+  simp only [unitFractionExpressible, gt_iff_lt, CharP.cast_eq_zero, div_zero, Set.mem_setOf_eq,
+    forall_exists_index, and_imp, true_and, or_true, this]
+  rintro _ s rfl hs h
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/304.lean
```diff
@@ -0,0 +1,74 @@
...
+theorem erdos_304.variants.upper_1985 :
+      (fun b => (smallestCollectionTo b : ℝ)) =O[atTop]
+      (fun b => Real.sqrt (Real.log b)) :=
+  sorry
+
+/--
+Is it true that $$N(b) \ll \log \log b$$?
+-/
+@[category research open, AMS 11]
+theorem upper_bound :
```
eric-wieser, line 171: This should have `erdos_304` in the name I think.

b-mehta, line 171:
```suggestion
theorem erdos_304 :
```

FormalConjectures/ErdosProblems/304.lean
```diff
@@ -0,0 +1,74 @@
...
+      (fun b => (smallestCollectionTo b : ℝ)) =O[atTop]
+      (fun b => Real.sqrt (Real.log b)) :=
+  sorry
+
+/--
+Is it true that $$N(b) \ll \log \log b$$?
+-/
+@[category research open, AMS 11]
+theorem erdos_304 :
+    (fun b : ℕ => (sSup {smallestCollection a b | a ∈ Finset.Ico 1 b} : ℝ)) =O[atTop]
```
Paul-Lez: I think this also works?
```suggestion
    (fun b : ℕ => (smallestCollectionTo b : ℝ)) =O[atTop]
```

FormalConjectures/ErdosProblems/304.lean
```diff
@@ -0,0 +1,74 @@
...
+  sSup {smallestCollection a b | a ∈ Finset.Ico 1 b}
+
+/--
+In 1950, Erdős [Er50c] proved the upper bound $$N(b) \ll \log b / \log \log b$$.
+[Er50c] Erdős, P., Az ${1}/{x_1} + {1}/{x_2} + \ldots + {1}/{x_n} =A/B$ egyenlet eg\'{E}sz sz\'{A}m\'{u} megold\'{A}sairól. Mat. Lapok (1950), 192-210.
+-/
+@[category research solved, AMS 11]
+theorem erdos_304.variants.upper_1950 :
+    (fun b => (smallestCollectionTo b : ℝ)) =O[atTop]
+    (fun b => Real.log b / Real.log (Real.log b)) := by
```
Paul-Lez: I'd be tempted to indent this (and statements below) as
```suggestion
    (fun b => (smallestCollectionTo b : ℝ)) =O[atTop]
      (fun b => Real.log b / Real.log (Real.log b)) := by
```
What do you think?

Paul-Lez (commented):
Nice work! If it's not too tedious, I'd suggest writing some "test"/API statements for the two definitions you added (and these can be sorried out if too annoying to prove)

b-mehta:
I added a bunch of API, I can do more if you'd like?

b-mehta:
Thinking about it more, there's plenty of API I could add here and so I'll probably just add it directly to mathlib, with a TODO in this file to change to the mathlib version once that's merged, since the conjectures will work with just the basic definition. How does that sound?

Paul-Lez:
That sounds good! This PR looks basically ready to merge - I'll just take another look at the maths later and then approve it!

b-mehta:
In the meantime I'll remove the new API for the moment :)

Paul-Lez:
You're more than welcome to leave it here and then remove it later once the upstreaming is done (whatever works best for you;))

PR #108, opened 2025-05-29
add erdos 289

b-mehta:
Co-authored-by: bhavikmehta@google.com

### DIFF
--- a/FormalConjectures/ErdosProblems/289.lean
+++ b/FormalConjectures/ErdosProblems/289.lean
@@ -1,0 +1,37 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 289
+*Reference:* [erdosproblems.com/289](https://www.erdosproblems.com/289)
+-/
+
+open Asymptotics Filter Finset
+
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/289.lean
```diff
@@ -0,0 +1,37 @@
...
+$I_1, \dotsc, I_k \subset \mathbb{N}$ with $|I_i| \geq 2$ for $1 \leq i \leq k$ such that
+$$
+1 = \sum_{i=1}^k \sum_{n \in I_i} \frac{1}{n}.
+$$
+-/
+@[category research open, AMS 11]
+theorem erdos_289 :
+    (∀ᶠ k : ℕ in atTop, ∃ I : Fin k → Finset ℕ,
+      (∀ i, 2 ≤ #(I i) ∧ ∃ a b, I i = Finset.Icc a b) ∧
+      ∑ i, ∑ n ∈ I i, (n⁻¹ : ℚ) = 1) ↔ answer(sorry) :=
```
b-mehta:
```suggestion
      ∑ i, ∑ n ∈ I i, (n⁻¹ : ℚ) = 1) ↔ answer(sorry) := by
```

FormalConjectures/ErdosProblems/289.lean
```diff
@@ -0,0 +1,37 @@
...
+/-- Is it true that, for all sufficiently large $k$, there exists finite intervals
+$I_1, \dotsc, I_k \subset \mathbb{N}$ with $|I_i| \geq 2$ for $1 \leq i \leq k$ such that
+$$
+1 = \sum_{i=1}^k \sum_{n \in I_i} \frac{1}{n}.
+$$
+-/
+@[category research open, AMS 11]
+theorem erdos_289 :
+    (∀ᶠ k : ℕ in atTop, ∃ I : Fin k → Finset ℕ,
+      (∀ i, 2 ≤ #(I i) ∧ ∃ a b, I i = Finset.Icc a b) ∧
```
Paul-Lez: (haven't thought about it too much): don't we also want to make sure the intervals don't contain `0`?

Paul-Lez (approved):
One small comment, LGTM otherwise!

FormalConjectures/ErdosProblems/289.lean
b-mehta: I don't think this should impact the truth of the statement, but you're correct that that's probably the intended meaning. I'll amend.

PR #109, opened 2025-05-30
add erdos 678

b-mehta:


FormalConjectures/ErdosProblems/678.lean
```diff
@@ -0,0 +1,74 @@
...
+`lcmInterval n k`.
+-/
+def lcmInterval (n k : ℕ) : ℕ := (Finset.Ioc n (n + k)).lcm id
+
+/--
+The referee of [Er79] found the example $M(96, 7) > M(104, 8)$, showing that there are cases where
+$M(n, k) > M(m, k + 1)$ with $m \geq n + k$.
+[Er79] Erdős, Paul, Some unconventional problems in number theory. Math. Mag. (1979), 67-70.
+-/
+@[category API]
```
Paul-Lez: Here I'd be tempted to put `test` as the category, but judging by the docstring this is a useful (but easy) problem. What do you think of having `high_school` instead?

FormalConjectures/ErdosProblems/678.lean
```diff
@@ -0,0 +1,74 @@
...
+Let $k \geq 3$. Are there infinitely many $m, n$ with $m \geq n + k$ such that
+$$
+M(n, k) > M(m, k + 1)
+$$?
+The answer is yes, as proved in a strong form by Cambie [Ca24].
+[Ca24] S. Cambie, Resolution of an Erdős' problem on least common multiples. arXiv:2410.09138 (2024).
+-/
+@[category research solved, AMS 11]
+theorem erdos_687 {k : ℕ} (hk : 3 ≤ k) :
+    {(m, n) | n + k ≤ m ∧ lcmInterval m (k + 1) < lcmInterval n k}.Infinite := by
```
Paul-Lez: Small nitpick: the English version was stated by a question, so following our convention we would want an `answer(True)` here.

FormalConjectures/ErdosProblems/678.lean
b-mehta: I ended up reading this paper, I think test might be more accurate. I'll amend.

Paul-Lez: Sounds good!

Paul-Lez (approved):
Nice work, thanks!

Issue #92, opened 2025-05-29
Local connectivity of & density of hyperbolicity in the Mandelbrot set


peabrainiac:
These are two closely related conjectures in holomorphic dynamics, with the former implying the latter; see [here](https://en.wikipedia.org/wiki/Mandelbrot_set#Hyperbolic_components) on wikipedia, or [here](https://arxiv.org/abs/1709.09869) for a survey paper. I think they would be great fits for the repository as they are easy to state and quite important for their field.  While I don't have a reference at hand, I believe that both properties are also conjectured for the [multibrot sets](https://en.wikipedia.org/wiki/Multibrot_set) for $n>2$.

Paul-Lez:
Nice! I'd been thinking we should add something about the Mandelbrot set to this repo. Would you like to have a go at this? Otherwise I can assign myself:)

peabrainiac:
I'll give it a try. I was hesitant at first because of the CLA bit (haven't seen that in other repositories before), but upon closer reading it does seem reasonable.

PR #112, opened 2025-05-30
Local connectivity of the Mandelbrot and Multibrot sets

peabrainiac:
Adds the [MLC conjecture](https://en.wikipedia.org/wiki/Mandelbrot_set#Local_connectivity), stating that the Mandelbrot set is locally connected / that all Multibrot sets are. This is the first of the two conjectures from #92.

I'm somewhat unsure whether this should be placed in the wikipedia folder or in the folder for a paper; the linked wikipedia article mentions the conjecture only for the Mandelbrot set, while e.g. [this paper](https://arxiv.org/abs/math/9902155) talks about the conjecture for Multibrot sets too.

### DIFF
--- a/.github/ISSUE_TEMPLATE/misformlization_report.md
+++ b/.github/ISSUE_TEMPLATE/misformlization_report.md
@@ -22,5 +22,5 @@ If available a suggestion of how to fix this misformalization.
 Add any other context about the problem here.
 
 ### Choose either option
-[ ] I plan on working on this issue
-[ ] This issue is up for grabs: I would like to see this misformalization fixed by somebody else
+- [ ] I plan on working on this issue
+- [ ] This issue is up for grabs: I would like to see this misformalization fixed by somebody else
--- a/.github/ISSUE_TEMPLATE/new_conjecture_request.md
+++ b/.github/ISSUE_TEMPLATE/new_conjecture_request.md
@@ -19,5 +19,5 @@ Consider linking any open Mathlib issues / PRs here.
 * ams-XX
 
 ### Choose either option
-[ ] I plan on working on this conjecture
-[ ] This issue is up for grabs: I would like to see this conjecture added by somebody else
+- [ ] I plan on working on this conjecture
+- [ ] This issue is up for grabs: I would like to see this conjecture added by somebody else
--- a/.github/workflows/build-and-docs.yml
+++ b/.github/workflows/build-and-docs.yml
@@ -27,7 +27,7 @@ permissions:
   id-token: write
 
 jobs:
-  build_project:
+  build:
     runs-on: ubuntu-latest
     name: Build project
     steps:
@@ -108,11 +108,20 @@ jobs:
           lake exe overwrite_index ./.lake/build/doc/index.html | tee -a "$GITHUB_STEP_SUMMARY"
 
       - name: Upload docs
+        id: deployment
         uses: actions/upload-pages-artifact@v3
         with:
           path: docbuild/.lake/build/doc
 
+  # Deployment job
+  deploy:
+    if: github.ref == 'refs/heads/main'
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
       - name: Deploy to GitHub Pages
-        if: github.ref == 'refs/heads/main'
         id: deployment
         uses: actions/deploy-pages@v4
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,4 +1,4 @@
-# This is the list of Formal Conjecturs's significant contributors.
+# This is the list of Formal Conjectures' significant contributors.
 #
 # This does not necessarily list everyone who has contributed code,
 # especially since many employees of one corporation may be contributing.
@@ -7,8 +7,10 @@
 Google LLC
 
 # Please keep the list sorted alphabetically
+Bhavik Mehta
 Calle Sönne
 Eric Wieser
+Junyan Xu
 Moritz Firsching
 Mirek Olšák
 Paul Lezeau
--- a/FormalConjectures/Arxiv/1609.08688/sIncreasingrTuples.lean
+++ b/FormalConjectures/Arxiv/1609.08688/sIncreasingrTuples.lean
@@ -220,7 +220,7 @@ example : [![1, 1, 1]] ⊗₂ [![1, 1, 1]] = toLex [toLex ![(1, 1), (1, 1), (1, 
 large $m$ such that $F(m) \geq m^{\alpha}$. -/
 @[category research solved, AMS 5]
 theorem maximalLength_pow {n : ℕ} {e : ℝ} (hn : 1 < n) (h : F n = (n : ℝ) ^ e) :
-    ∀ᶠ m : ℕ in Filter.atTop, (m : ℝ) ^ e ≤ F m :=
+    ∀ᶠ m : ℕ in Filter.atTop, (m : ℝ) ^ e ≤ F m := by
   sorry
 
 /-- $F(n) \leq n^{3/2}$. -/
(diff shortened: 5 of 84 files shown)


peabrainiac:
Another thing I wasn't sure about - should this and the density of hyperbolicity conjecture (which I mentioned in the issue but haven't yet formalised) be put in one file (called e.g. `Mandelbrot.lean`) or separate ones? They're separate conjectures, but related in that one implies the other, in that both use the definitions from this file and in that they would reference the same paper if the paper is used instead of wikipedia as the reference here.

FormalConjectures/Wikipedia/MLC.lean
```diff
@@ -0,0 +1,85 @@
...
+set of all parameters `c : ℂ` for which `0` does not escape to infinity under repeated application
+of `z ↦ z ^ 2 + c`. -/
+abbrev mandelbrotSet := multibrotSet 2
+
+/- To test the definition, we show that the mandelbrot set is equivalently the set of all parameters
+`c` for which the orbit of `0` under `z ↦ z ^ 2 + c` does not leave the closed disk of radius two
+around the origin. -/
+@[category test]
+example : mandelbrotSet = {c | ∀ k, ‖(fun z ↦ z ^ 2 + c)^[k] 0‖ ≤ 2} := by
+  ext c; refine ⟨fun h k ↦ ?_, fun h h' ↦ ?_⟩ <;> dsimp [mandelbrotSet, multibrotSet] at h ⊢
```
Paul-Lez, line 51: Nice work! This seems like a useful result - what do you think of changing it to a `theorem` and giving it the `API` catgeory instead?
Does a similar result hold for general multibrot sets? (if so, feel free to included it, possibly with a sorried proof)

FormalConjectures/Wikipedia/MLC.lean
```diff
@@ -0,0 +1,85 @@
...
+/-- The MLC conjecture, stating that the mandelbrot set is locally connected. -/
+@[category research open, AMS 37]
+theorem MLC : LocallyConnectedSpace mandelbrotSet := by
+  sorry
+
+/-- A stronger version of the MLC conjecture, stating that all multibrots are locally connected.
+Note that we don't need to require `2 ≤ n` because the conjecture holds in the trivial cases `n = 0`
+and `n = 1` too. -/
+@[category research open, AMS 37]
+theorem MLC' n : LocallyConnectedSpace (multibrotSet n) := by
```
Paul-Lez: How about
```suggestion
theorem MLC_general_exponent (n : ℕ) : LocallyConnectedSpace (multibrotSet n) := by
```

Paul-Lez:
> Another thing I wasn't sure about - should this and the density of hyperbolicity conjecture (which I mentioned in the issue but haven't yet formalised) be put in one file (called e.g. `Mandelbrot.lean`) or separate ones? They're separate conjectures, but related in that one implies the other, in that both use the definitions from this file and in that they would reference the same paper if the paper is used instead of wikipedia as the reference here.

I think it's reasonable to have them in the same file given that they are very related. If more `Mandelbrot` gets added later, it may be worth moving the basic definitions to `ForMathlib`.

Paul-Lez (commented):
Out of curiosity, are there any (reasonable to formalize) partial results in this direction?

Paul-Lez:
I think the Wiki subdirectory is reasonable as a location for this file, but definitely worth linking the paper as a reference!

FormalConjectures/Wikipedia/MLC.lean
peabrainiac: I've changed it from an `example` to a `theorem` now and also generalised it to multibrots. The proof is almost the same on paper, but turned out to be trickier to formalise than I had hoped - in particular, I needed a lemma `le_add_pow` that I wasn't sure where to put. It should either be generalised & moved to `ForMathlib` or simply inlined.

peabrainiac: Eric Wieser pointed me towards a more efficient way to prove that helper lemma on discord, so I've inlined that proof and deleted the lemma.

Paul-Lez: Nice work!

FormalConjectures/Wikipedia/MLC.lean
```diff
@@ -0,0 +1,106 @@
...
+
+/-- The Mandelbrot set is the special case of the multibrot set for n = 2. In other words, it is the
+set of all parameters `c : ℂ` for which `0` does not escape to infinity under repeated application
+of `z ↦ z ^ 2 + c`. -/
+abbrev mandelbrotSet := multibrotSet 2
+
+/-- The `multibrotSet n` is equivalently the set of all parameters `c` for which the orbit of `0`
+under `z ↦ z ^ n + c` does not leave the closed disk of radius `2 ^ (n - 1)⁻¹` around the origin. -/
+@[category API]
+theorem multibrotSet_eq {n : ℕ} [Fact (2 ≤ n)] :
```
Paul-Lez: I'd be tempted to have this as 
```suggestion
theorem multibrotSet_eq {n : ℕ} (hn : 2 ≤ n) :
```
since that would for example make the application easier in the proof of `mandelbrotSet_eq`

Paul-Lez (approved):
Thanks for the great work - LGTM!

FormalConjectures/Wikipedia/MLC.lean
peabrainiac: My thought was that if more API around multibrot sets was to be developed, it would be useful to keep `2 ≤ n` around as a fact to avoid having to pass it explicitly everytime a lemma that needs it is invoked. The issue of having to provide a `Fact (2 ≤ 2)` instance to apply such lemmas to the Mandelbrot set could be solved by setting up a global `fact_le_rfl` instance, analogous to how [`Nat.fact_prime_two`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Nat/Prime/Defs.html#Nat.fact_prime_two) allows working over characteristic two without having to keep referencing the fact that two is prime. Of course though, in the context of this repository specifically that doesn't matter much - I can change it to just take in the assumption explicitly if you want.

b-mehta: I disagree: instances like `Fact (Nat.Prime _)` are for inserting those facts into the typeclass system so that things like `Field (ZMod _)` can work. `Fact` specifically should not be used to avoid repeatedly referencing a fact!

peabrainiac: If one wants to register topological properties of the Multibrot sets as instances, e.g. a `CompactSpace (multibrotSet n)` instance would require `2 ≤ n` to be available to the typeclass system too, so I think for an expansive API it'd still make sense. I see your point though, and will remove the use of `Fact` because we don't need it here.

b-mehta: Fair point!

peabrainiac:
> Out of curiosity, are there any (reasonable to formalize) partial results in this direction?

Afaik most partial results are of the form "the mandelbrot set is locally connected at all points with property P", with property P not being very easy to state. One known result that could easily be stated here is e.g. that the Multibrot sets and their complements are connected; it has been formalised before though by Geoffrey Irving [here](https://github.com/girving/ray/blob/main/Ray/Dynamics/Multibrot/Connected.lean#L76-L111). The difficulty of that would be either graduate or solved research problem, I'm not sure.

Paul-Lez:
> Afaik most partial results are of the form "the mandelbrot set is locally connected at all points with property P", with property P not being very easy to state. One known result that could easily be stated here is e.g. that the Multibrot sets and their complements are connected; it has been formalised before though by Geoffrey Irving [here](https://github.com/girving/ray/blob/main/Ray/Dynamics/Multibrot/Connected.lean#L76-L111). The difficulty of that would be either graduate or solved research problem, I'm not sure.

Ok good to know - in that case adding partial results is out of scope for this PR. Thanks again!

PR #113, opened 2025-05-30
chore: Move some Landau problems to the Wikipedia subdirectory

Paul-Lez:


callesonne:
I would even prefer if these conjectures went straight into wikipedia (and not under Landau problems). But approving nonetheless and you can do as you wish.

callesonne (approved):

eric-wieser:
Maybe worth noting that @teorth has linked specifically to the old filenames from Mathstodon, so I think it would be wise to leave them in place until the traffic from there dies down.

Paul-Lez:
Good point - I'll shelf this for now and maybe revisit it later.

callesonne:
@Paul-Lez @eric-wieser would this be worth revisiting now?

PR #118, opened 2025-05-31
Fix caching issues with the doc build

eric-wieser:
Currently, doc pages for deleted Lean files remain in the build directory, and so get uploaded anyway.

This is using a backport of https://github.com/leanprover/doc-gen4/pull/286

### DIFF
--- a/.github/workflows/build-and-docs.yml
+++ b/.github/workflows/build-and-docs.yml
@@ -113,10 +113,10 @@ jobs:
           rsync -a --files-from=out-files.txt --relative .lake/build/doc ./out
 
       - name: Inject stats into index.html
+        shell: bash
         run: |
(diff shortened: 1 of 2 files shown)


b-mehta:
What are the caching issues? If there's a new shiny way to do docs caching then it might be worth using in other repos: FLT and Carleson (and my active project repos) have doc caching which works somewhat well, but not perfectly.

eric-wieser:
The caching issue is that doc pages for deleted Lean files remain in the build directory, and so get uploaded anyway. Note that most of the work here is from a PR to doc-gen.

b-mehta:
Ahh interesting, I'd observed that too but didn't realise where it came from!

mo271 (approved):

Issue #104, opened 2025-05-29
Easy formalization targets


alreadydone:
The "solved" ones may be more appropriate for direct PR to Mathlib, if they are not related to an unsolved conjecture.

Feel free to suggest additional targets.

#### Number Theory
- [x] [existence of an odd perfect number](https://en.wikipedia.org/wiki/Perfect_number#Odd_perfect_numbers) (WIP: #122)
- [ ] [existence of infinitely many Mersenne primes](https://en.wikipedia.org/wiki/Mersenne_conjectures#Lenstra%E2%80%93Pomerance%E2%80%93Wagstaff_conjecture): #134 
- [x] conjectures on Fermat primes from Wikipedia: #135
- [ ] [infinitude of primes of the form $x^2+1$](https://mathoverflow.net/questions/151159/status-of-the-x2-1-problem)
- [x] [Schinzel's hypothesis H](https://en.wikipedia.org/wiki/Schinzel's_hypothesis_H) (a generalization of the [twin prime conjecture](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/LandauProblems/TwinPrimes.lean) and the first [Hardy–Littlewood conjecture](https://github.com/google-deepmind/formal-conjectures/blob/3091da8029c4b2bbe585ffb37f109d67a8913187/FormalConjectures/Wikipedia/HardyLittlewood.lean)): #136 
- [ ] [Catalan's conjecture](https://en.wikipedia.org/wiki/Catalan%27s_conjecture) (solved)
- [ ] [Fermat–Catalan conjecture](https://en.wikipedia.org/wiki/Fermat%E2%80%93Catalan_conjecture) (generalization of Catalan's and FLT)
- [ ] [Euler's sum of powers conjecture](https://en.wikipedia.org/wiki/Euler%27s_sum_of_powers_conjecture) (unknown for k ≥ 6, disproved for k = 4, 5)
- [x] [Beal conjecture](https://github.com/google-deepmind/formal-conjectures/blob/cd787dbe6edd33b8eb1045c20bf2bd38bc1d0416/FormalConjectures/Wikipedia/BealConjecture.lean)
- [x] [abc conjecture](https://github.com/google-deepmind/formal-conjectures/blob/491083e7b95a7c61d16d56144d5f5579f9c2d68b/FormalConjectures/Wikipedia/ABC.lean)
[...]

PR #119, opened 2025-05-31
ramanujan-petersson conjecture

seewoo5:
https://en.wikipedia.org/wiki/Ramanujan%E2%80%93Petersson_conjecture

One of the easy targets in #104.

### DIFF
--- a/FormalConjectures/Wikipedia/RamanujanPetersson.lean
+++ b/FormalConjectures/Wikipedia/RamanujanPetersson.lean
@@ -1,0 +1,34 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Ramanujan-Petersson Conjecture
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Ramanujan%E2%80%93Petersson_conjecture)
+-/
+
+open PowerSeries PowerSeries.WithPiTopology
+
+noncomputable def Δ : PowerSeries ℝ := X * ∏' (n : ℕ+), (1 - X ^ (n : ℕ)) ^ 24
+
+noncomputable def τ (n : ℕ) : ℝ := PowerSeries.coeff ℝ n Δ
+
+
+@[category research solved, AMS 11]
+theorem ramanujan_petersson : ∀ p : ℕ, Prime p → abs (τ p) ≤ 2 * p ^ ((11 : ℝ) / 2) := by
+  sorry


FormalConjectures/Wikipedia/RamanujanPetersson.lean
```diff
@@ -0,0 +1,34 @@
...
+# Ramanujan-Petersson Conjecture
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Ramanujan%E2%80%93Petersson_conjecture)
+-/
+
+open PowerSeries PowerSeries.WithPiTopology
+
+noncomputable def Δ : PowerSeries ℝ := X * ∏' (n : ℕ+), (1 - X ^ (n : ℕ)) ^ 24
+
+noncomputable def τ (n : ℕ) : ℝ := PowerSeries.coeff ℝ n Δ
```
alreadydone: Nice! I'm a bit worried that the `tprod` may not work as expected though. Can we compute the first few values of τ?

Also I think if we define τ in mathlib we'd like it to take values in ℤ.

shamEiNew:
In reference to https://github.com/google-deepmind/formal-conjectures/issues/120, a PR already exists

FormalConjectures/Wikipedia/RamanujanPetersson.lean
b-mehta: @seewoo5 Am I correct that this `tprod` is the one that we're using in a certain other project, and so we have some degree of certainty that it's correct? Or is it different here?

seewoo5: @b-mehta Yes exactly, although that one is an actual modular form as a function, and here I define as a formal power series with Pi topology. So the equation itself is correct.

@alreadydone I changed to $$\mathbb{Z}$$. Do you want to change it into some computable definition, or just verify that it is a correct definition? I'm pretty sure that the equation itself is correct.

alreadydone: No need of making it computable, just want to see `τ 1 = 1`, `τ 2 = -24` etc. Maybe leave them as sorries as well if it's not easy to prove them now. Proving these probably requires implementing a computable function that is equal to τ anyway.
If @b-mehta is sure the definition works I have no objection.

seewoo5: I can add things like
```lean
theorem τ_one : τ 1 = 1 := by sorry
theorem τ_two : τ 2 = -24 := by sorry
```
would this be helpful?

b-mehta: I'm not completely certain how to show that the product is multipliable; we have in another repo that it's multipliable if X is changed to e^(2pi i z). But I'm reasonably sure that it's correct.

seewoo5: If I understood correctly, [PiTopology](https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/PowerSeries/PiTopology.html) of power series is a coarsest topology on $$R[[X]]$$ where the coefficient maps are all continuous. If $$R = \mathbb{Z}$$ is endowed with discrete topology, then this is the same as $$X$$-adic topology, i.e. $$X^n \to 0$$ as $$n \to \infty$$. Then the product makes sense (i.e. it converges) since for each $$k$$, $$k$$-th coefficient is stabilized after multiplying sufficiently many terms (e.g. multiplying $$(1 - X^n)^{24} = 1 - 24 X^n + \cdots + X^{24n}$$ for $$n \ge k + 1$$ does not change the coefficient of $$X^k$$).

b-mehta:
```
lemma multipliable : Multipliable fun n : ℕ+ ↦ ((1 - X ^ (n : ℕ)) ^ 24 : PowerSeries ℤ) := by
  sorry

lemma τ_zero : τ 0 = 0 := by simp [τ, Δ]

lemma τ_one : τ 1 = 1 := by 
  obtain ⟨i, hi⟩ := by simpa using ((continuous_constantCoeff ℤ).tendsto _).comp multipliable.hasProd
  simp [τ, Δ, hi i]
```

FormalConjectures/Wikipedia/RamanujanPetersson.lean
```diff
@@ -0,0 +1,47 @@
...
+lemma multipliable : Multipliable fun n : ℕ+ ↦ ((1 - X ^ (n : ℕ)) ^ 24 : PowerSeries ℤ) := by
+  sorry
+
+lemma τ_zero : τ 0 = 0 := by simp [τ, Δ]
+
+lemma τ_one : τ 1 = 1 := by
+  obtain ⟨i, hi⟩ := by simpa using ((continuous_constantCoeff ℤ).tendsto _).comp multipliable.hasProd
+  simp [τ, Δ, hi i]
+
+lemma τ_two : τ 2 = -24 := by
```
b-mehta:
```suggestion
@[category test]
lemma τ_two : τ 2 = -24 := by
```

FormalConjectures/Wikipedia/RamanujanPetersson.lean
```diff
@@ -0,0 +1,47 @@
...
+-/
+
+open PowerSeries PowerSeries.WithPiTopology
+
+noncomputable def Δ : PowerSeries ℤ := X * ∏' (n : ℕ+), (1 - X ^ (n : ℕ)) ^ 24
+
+noncomputable def τ (n : ℕ) : ℤ := PowerSeries.coeff ℤ n Δ
+
+
+lemma multipliable : Multipliable fun n : ℕ+ ↦ ((1 - X ^ (n : ℕ)) ^ 24 : PowerSeries ℤ) := by
```
b-mehta:
```suggestion
@[category API]
lemma multipliable : Multipliable fun n : ℕ+ ↦ ((1 - X ^ (n : ℕ)) ^ 24 : PowerSeries ℤ) := by
```

FormalConjectures/Wikipedia/RamanujanPetersson.lean
```diff
@@ -0,0 +1,47 @@
...
+  sorry
+
+lemma τ_zero : τ 0 = 0 := by simp [τ, Δ]
```
b-mehta:
```suggestion
@[category test]
lemma τ_zero : τ 0 = 0 := by simp [τ, Δ]
```

FormalConjectures/Wikipedia/RamanujanPetersson.lean
```diff
@@ -0,0 +1,47 @@
...
+
+lemma τ_one : τ 1 = 1 := by
```
b-mehta:
```suggestion
@[category test]
lemma τ_one : τ 1 = 1 := by
```

b-mehta (commented):
LGTM

FormalConjectures/Wikipedia/RamanujanPetersson.lean
alreadydone: How about we define `τ n` as the `n-1`st coefficient of the finite product of the first `n-1` factors (without the X factor)? You can use `Polynomial` instead of `PowerSeries` and it might be easier to compute.

seewoo5: @alreadydone I thought about that too, but I lean toward using power series since I thought it is more close to the original formulation. But I don't have a strong objection for your polynomial version.

b-mehta: I agree that's possible, but my preference is for the current definition, and your idea can be used to compute. This one more closely aligns with the informal version!

seewoo5:
@alreadydone @b-mehta Also I'm thinking about renaming the file as `Ramanujan` instead of `RamanujanPetersson`, since I think it is good to add Lehmer's conjecture (mentioned in #104) here and former will be a better name for containing both in one file - what do you think? (I may make a separate PR for it)

alreadydone:
Maybe rename it to `RamanujanTau`? I agree it's a good idea to add a solved conjecture in the same file as a related open conjecture, since this repo is not primarily intended for solved conjectures, unlike what I thought before.

b-mehta:
@mo271 Do you have thoughts on the above? Mathematically, this lgtm

mo271:
> @mo271 Do you have thoughts on the above? Mathematically, this lgtm

LGTM! No strong opinion on the name and whether we can add more to the file later. When the name is not a a good fit as a consequence, we can just rename it, I suppose. 
So I am in favor of merging this as is and expand later.

mo271 (approved):

alreadydone (approved):

PR #125, opened 2025-05-31
Add Andrica's conjecture

Reklle:


FormalConjectures/Wikipedia/Andrica.lean
```diff
@@ -0,0 +1,30 @@
...
+# Andrica's conjecture
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Andrica%27s_conjecture)
+-/
+@[category research open, AMS 11]
+theorem andrica_conjecture :
+  ∀ (n : ℕ),
+    let prev := Nat.nth Prime n;
+    let next := Nat.nth Prime (n+1);
+    Real.sqrt next - Real.sqrt prev < 1 := by
```
alreadydone: I don't see why you'd like to use `Nat`, and I think it looks better if you use dot notations, i.e. `n.nth Prime` and `(n+1).nth Prime`.

Reklle: Сorrected. I'm just learning, so I don't know much about style yet

FormalConjectures/Wikipedia/Andrica.lean
```diff
@@ -0,0 +1,33 @@
...
+-/
+
+/--
+The inequality `√pₙ₊₁-√pₙ < 1` holds for all `n`.
+-/
+@[category research open, AMS 11]
+theorem andrica_conjecture (n : ℕ) :
+    let prev := n.nth Prime;
+    let next := (n+1).nth Prime;
+    next.sqrt - prev.sqrt < 1 := by
```
alreadydone: You should use `Real.sqrt` (as you originally did) instead of `Nat.sqrt`, which takes the floor, so `5.sqrt = 3` and `3.sqrt = 2`, making the conjecture false.
```suggestion
theorem andrica_conjecture (n : ℕ) :
    Real.sqrt ((n+1).nth Prime) - Real.sqrt (n.nth Prime) < 1 := by
```
(by the way, there's no need of `;` at the end of `let` declarations.

FormalConjectures/Wikipedia/Andrica.lean
alreadydone: sorry, I meant I don't see why you need to use `let`

Reklle: I think it looks more readable than in one line

FormalConjectures/Wikipedia/Andrica.lean
Reklle: Thank you, I will remember! I fixed it.

FormalConjectures/Wikipedia/Andrica.lean
```diff
@@ -0,0 +1,30 @@
...
+/-!
+# Andrica's conjecture
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Andrica%27s_conjecture)
+-/
+
+/--
+The inequality `√pₙ₊₁-√pₙ < 1` holds for all `n`.
+-/
+@[category research open, AMS 11]theorem andrica_conjecture (n : ℕ) :
```
callesonne:
```suggestion
@[category research open, AMS 11]
theorem andrica_conjecture (n : ℕ) :
```

callesonne (commented):
Thanks for contributing this! If you want, I would suggest adding the variant with `g_n < 2 \sqrt{p_n} + 1` from the wiki also.

FormalConjectures/Wikipedia/Andrica.lean
```diff
@@ -0,0 +1,31 @@
...
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Andrica's conjecture
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Andrica%27s_conjecture)
+-/
+
+/--
+The inequality `√pₙ₊₁-√pₙ < 1` holds for all `n`.
```
mo271: nit: could also use latex for the docstring here

mo271: nit: could also use latex for the docstring here, and perhaps also mention $p_n$ denotes the $n$-th prime

mo271 (approved):
Thanks! LGTM

FormalConjectures/Wikipedia/Andrica.lean
Reklle: Yeah, I like latex better too.

FormalConjectures/Wikipedia/Andrica.lean
```diff
@@ -0,0 +1,31 @@
...
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Andrica's conjecture
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Andrica%27s_conjecture)
+-/
+
+/--
+The inequality $\sqrt{p_{n+1}}-\sqrt{p_n} \le 1$ holds for all $n$, where $p_n$ is the nth prime number.
```
Paul-Lez:
```suggestion
The inequality $\sqrt{p_{n+1}}-\sqrt{p_n} < 1$ holds for all $n$, where $p_n$ is the nth prime number.
```

Paul-Lez (approved):
Thanks for the PR!

FormalConjectures/Wikipedia/Andrica.lean
Reklle: As I know, there is no prime gap function, and the variant looks like a simple consequence. Or should it be added?

Paul-Lez: In this case it would be fine to just inline the definition of the prime gap function. That being said, this does look like an easy reformulation, so this doesn't have to be added/can be added later (having reformulations is a great way to detect misformalisations for example!)

PR #126, opened 2025-05-31
Add Brocard's Conjecture

Reklle:


FormalConjectures/Wikipedia/Brocard.lean
```diff
@@ -0,0 +1,35 @@
...
+
+For every n ≥ 2, between the squares of the n-th and (n+1)-th primes,
+there are at least four prime numbers.
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Brocard%27s_conjecture)
+-/
+
+@[category research open, AMS 11]
+theorem brocard_conjecture :
+  ∀ n : ℕ, n ≥ 2 →
```
Paul-Lez:
```suggestion
theorem brocard_conjecture (n : ℕ) (hn : n ≥ 1)
```

FormalConjectures/Wikipedia/Brocard.lean
```diff
@@ -0,0 +1,35 @@
...
+    let prev := Nat.nth Prime n;
```
Paul-Lez:
```suggestion
    letI prev := n.nth Prime
```

FormalConjectures/Wikipedia/Brocard.lean
```diff
@@ -0,0 +1,35 @@
...
+    let next := Nat.nth Prime (n+1);
```
Paul-Lez:
```suggestion
    letI next := (n+1).nth Prime
```

FormalConjectures/Wikipedia/Brocard.lean
```diff
@@ -0,0 +1,35 @@
...
+    let primesInBetween := (Finset.Ioo (prev^2) (next^2)).filter Nat.Prime
```
Paul-Lez:
```suggestion
    letI primesInBetween := (Finset.Ioo (prev^2) (next^2)).filter Nat.Prime
```

FormalConjectures/Wikipedia/Brocard.lean
```diff
@@ -0,0 +1,35 @@
...
+    primesInBetween.card ≥ 4 := by
```
Paul-Lez: (small nit: to stick to the Mathlib convention)
```suggestion
    4 ≤ primesInBetween.card := by
```

FormalConjectures/Wikipedia/Brocard.lean
Paul-Lez: Note: the wiki page is indexing primes from `1` while we're indexing them from `0` so we should also shift the condition on the index `n`.

FormalConjectures/Wikipedia/Brocard.lean
```diff
@@ -0,0 +1,35 @@
...
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Brocard's Conjecture
+
+For every n ≥ 2, between the squares of the n-th and (n+1)-th primes,
+there are at least four prime numbers.
```
Paul-Lez: Usually, we'd keep this as the docstring for the main statement.

FormalConjectures/Wikipedia/Brocard.lean
Reklle: Why is letI needed here?
I'm new to Lean so I apologize for the huge amount of editing my code that was required.

alreadydone: `I` means "inline"; you can see from `#print brocard_conjecture` that the expressions `n.nth Prime` etc. are plugged into the statement.

alreadydone: In general if an expression only appears once, it probably doesn't need to be extracted as a `let`/`letI`.

Paul-Lez: I often use `letI` to simplify lines involving nested complicated expressions (I find it helps a lot with readability), but this is more a matter of personal preferences.

FormalConjectures/Wikipedia/Brocard.lean
```diff
@@ -0,0 +1,35 @@
...
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Brocard's Conjecture
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Brocard%27s_conjecture)
+-/
+
+/--
+For every `n ≥ 2`, between the squares of the n-th and (n+1)-th primes,
```
Paul-Lez:
```suggestion
For every `n ≥ 2`, between the squares of the `n`-th and `(n+1)`-th primes,
```

FormalConjectures/Wikipedia/Brocard.lean
Paul-Lez: Also @Reklle feel free to ask more Lean questions (both here and on the Zulip) - we're always happy to answer!

Paul-Lez (approved):
Thanks for the contribution:)

PR #68, opened 2025-05-28
feat: add answer(sorry) to all Erdos problems stated as questions.

callesonne:
See #29.

This fixes this for all open Erdos problems, and also for the solved ones up to 442.

Also fixes a potential misformalization / issue with 244, where previously it was assumed that the set had some density (which does not seem clear from the problem statement).

Also, when creating this branch I by accident also had some work in converting (a few) docstrings into Latex. I have not bothered to remove this, as this is a (very minor) good thing to do anyways.

PR #130, opened 2025-05-31
chore: fix statement of erdos 316

b-mehta:
The statement of Erdos 316 was mis-negated in #68. This PR fixes that, and to confirm the new statement is correct, proves the new statement.


callesonne (approved):
Thanks for catching this!

PR #142, opened 2025-06-02
chore: various style fixes in Wikipedia subdirectory

Paul-Lez:
This PR does the following
- Use  the `answer(sorry)` convention when relevant
- Reword certain docstrings/theorem statements
- A few other drive-by style fixes

### DIFF
--- a/FormalConjectures/Wikipedia/Hadamard.lean
+++ b/FormalConjectures/Wikipedia/Hadamard.lean
@@ -72,6 +72,11 @@ example (n : ℕ) (M : Matrix (Fin n) (Fin n) ℝ) : IsHadamard' M ↔ IsHadamar
       norm_num
   · sorry
 
+/- Note: the conjecture was originally formulated by
+Hadamard as a question: "For which values of $n=4k$ does
+a Hadamard matrix exist." However the expectation seems
+to be that all such matrices are Hadamard, and the
+formalisation has been written with this in mind. -/
 
 /--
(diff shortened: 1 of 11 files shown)


FormalConjectures/Wikipedia/Hadamard.lean
```diff
@@ -74,12 +74,17 @@ example (n : ℕ) (M : Matrix (Fin n) (Fin n) ℝ) : IsHadamard' M ↔ IsHadamar
 
 
 /--
-Hadamard asks for which values of $n = 4k$ exists such a Matrix.
+There exists a Hadamard matrix for all $n = 4k$.
```
mo271, line 82: In the source, the question is phrased by Hadamard as:
"Peut-on trouver de tels déterminants pour d'autres valeurs de `n`"

But I do think the general expectation for this conjecture is a bias towards this being true. However, we could use the answer(sorry) formalization here. 

Not strong opinion either way.

mo271 (approved):
LGTM, thanks!

FormalConjectures/Wikipedia/Hadamard.lean
Paul-Lez: Right I see! My motivation for this edit was that the previous formulation would sort of imply an answer(sorry) formulation as you say.

Paul-Lez: Added a comment about that in the file.

PR #145, opened 2025-06-03
Density of hyperbolicity

### LINKED ISSUE #92: Local connectivity of & density of hyperbolicity in the Mandelbrot set

peabrainiac:
Adds the [density of hyperbolicity](https://en.wikipedia.org/wiki/Mandelbrot_set#Hyperbolic_components) conjecture for Mandelbrot and Multibrot sets. Together with #112, this closes #92.


peabrainiac:
Two things I wasn't sure about:
- is `MLC.lean` still a good name, or should this be renamed to `Mandelbrot.lean`?
- another conjecture I'd be interested in formalising is [that the boundary of the Mandelbrot set has zero area](https://mathoverflow.net/questions/37229/area-of-the-boundary-of-the-mandelbrot-set). That should take only a couple of lines, but is afaik mostly unrelated to the conjectures we already have; can it still go into the same file and PR, or should it go into a different one?

Paul-Lez:
I would suggest renaming the file to `Mandelbrot.lean`, and feel free to include the other conjecture you mentioned there. If needed, we can always split the file later on.

FormalConjectures/Wikipedia/MLC.lean
```diff
@@ -103,3 +108,42 @@ and `n = 1` too. -/
 @[category research open, AMS 37]
 theorem MLC_general_exponent (n : ℕ) : LocallyConnectedSpace (multibrotSet n) := by
   sorry
+
+/-- We say that `z : ℂ` is part of an attracting cycle of period `n` of `f : ℂ → ℂ` if it is an
+`n`-periodic point (i.e. `f^[n] z = z`), `f^[n]` is differentiable at `z` and `‖deriv f^[n] z‖` is
+strictly less than one.-/
+@[category API]
```
Paul-Lez: We usually only on use the `category` attribute on theorem/lemma/example statements
```suggestion
```

FormalConjectures/Wikipedia/MLC.lean
```diff
@@ -103,3 +108,42 @@ and `n = 1` too. -/
...
+example (f : ℂ → ℂ) (z : ℂ) : ¬ IsAttractingCycle f 0 z := by
+  simp [IsAttractingCycle]
+
+/-- The density of hyperbolicity conjecture, stating that the set of all parameters `c` for which
+`fun z ↦ z ^ 2 - c` has an attracting cycle is dense in the Mandelbrot set. -/
+@[category research open, AMS 37]
+theorem density_of_hyperbolicity :
+    mandelbrotSet ⊆ closure {c | ∃ n z, IsAttractingCycle (fun z ↦ z ^ 2 - c) n z} := by
+  sorry
+
```
Paul-Lez, line 143: If you feel like doing a little extra work, I think a nice piece of API to add could be
```
{c | ∃ n z, IsAttractingCycle (fun z ↦ z ^ n - c) n z} ⊆ multibrotSet n
```

Paul-Lez (commented):
Great work, thanks!

FormalConjectures/Wikipedia/MLC.lean
```diff
@@ -18,6 +18,11 @@ import FormalConjectures.Util.ProblemImports
 
 /-!
 # Local connectivity of the Mandelbrot and Multibrot sets
+This file adds two conjectures about the Mandelbrot and Multibrot sets:
+- the *MLC conjecture*, stating that these sets are locally connected
+- the *density of hyperbolicity* conjecture, stating that parameters with no attracting cycles are
+  dense in the Mandelbrot and Multibrot sets.
```
alreadydone: It seems your formal statements do not include the "no".

FormalConjectures/Wikipedia/MLC.lean
```diff
@@ -103,3 +108,41 @@ and `n = 1` too. -/
...
+
+/-- We say that `z : ℂ` is part of an attracting cycle of period `n` of `f : ℂ → ℂ` if it is an
+`n`-periodic point (i.e. `f^[n] z = z`), `f^[n]` is differentiable at `z` and `‖deriv f^[n] z‖` is
+strictly less than one.-/
+def IsAttractingCycle (f : ℂ → ℂ) (n : ℕ) (z : ℂ) : Prop :=
+  f.IsPeriodicPt n z ∧ DifferentiableAt ℂ f^[n] z ∧ ‖deriv f^[n] z‖ < 1
+
+/-- For example, `0` is part of an attracting `2`-cycle of `z ↦ z ^ 2 - 1`. -/
+@[category test]
+example : IsAttractingCycle (fun z ↦ z ^ 2 - 1) 2 0 :=
```
alreadydone, line 122: I'd write 
```suggestion
example : IsAttractingCycle (· ^ 2 - 1) 2 0 :=
```
but it's personal preference and I'm not sure if it's more readable.

FormalConjectures/Wikipedia/MLC.lean
peabrainiac: I agree that this would be a nice piece of API. Unfortunately I don't think it's feasible here - showing that parameters for which 0 is part of an attracting cycle belong to the Mandelbrot set is trivial, but showing the same for parameters for which we just know that *some* attracting cycle exists requires a lemma by Fatou that says that every attracting cycle is approached by some critical point. It should definitely be added to Mathlib at some point, but I don't know if we even have all the prerequisites from complex analysis for it yet.

FormalConjectures/Wikipedia/MLC.lean
peabrainiac: Good catch - it is indeed that comment that is wrong.

FormalConjectures/Wikipedia/MLC.lean
peabrainiac: I prefer the notation using `fun` because it allows setting the variable name. The default appears to be `x`, which imo makes it look like a function of a real variable - I suppose it doesn't matter here because you don't get to see the name, but for statements with longer proofs it definitely helps to have all the variables named correctly, so I've just made it a habit to use `fun` throughout.

peabrainiac:
> I would suggest renaming the file to `Mandelbrot.lean`, and feel free to include the other conjecture you mentioned there. If needed, we can always split the file later on.

Did that now. I also included a version of that conjecture for general exponents, but couldn't find a reference for it - the linked mathotherflow question only talks about the Mandelbrot set. Let me know if I should remove that version again.

FormalConjectures/Wikipedia/MLC.lean
Paul-Lez: Oh interesting! I thought it would be a lot easier. In that case let's leave this out.

Paul-Lez (approved):
Thanks a lot for the PR, this looks great:)

FormalConjectures/Wikipedia/MLC.lean
alreadydone: If it's true, why not add it with sorried proof?

Issue #149, opened 2025-06-03
Casas-Alvero conjecture


mo271:
### What is the conjecture
Mathoverflow post with link to paper in answer: [Casas-Alvero conjecture](https://mathoverflow.net/questions/27851/polynomials-having-a-common-root-with-their-derivatives)

### Prerequisites needed
Everything needed should be in mathlib

### [AMS categories](https://github.com/google-deepmind/formal-conjectures/labels?q=ams-)

* ams-12
* ams-14

### Choose either option
- [x] @shamEiNew  plans to work on this conjecture
- [ ] This issue is up for grabs: I would like to see this conjecture added by somebody else

shamEiNew:
I would work on it.

alreadydone:
The conjecture is now [claimed](https://arxiv.org/abs/2501.09272) to be proven by [Soham Ghosh](https://sites.math.washington.edu/~soham13/), 3rd year PhD candidate at U Washington, student of Max Lieblich and Farbod Shokrieh. It's 13 pages but builds on top of an earlier 29-page [paper](https://arxiv.org/abs/2402.18717). Formalizing the proof would probably be a test of mathlib's algebraic geometry and commutative/homological algebra libraries.

mo271:
Cool, then perhaps best to link that arxiv paper.

shamEiNew:
Should this be moved to solved category with link to paper ?

alreadydone:
I wouldn't recommend that before the papers are peer-reviewed.

alreadydone:
This can now be closed, I suppose.

PR #150, opened 2025-06-03
Casas-Alvero Conjecture

shamEiNew:
Formalization of casas-alvero conjecture https://github.com/google-deepmind/formal-conjectures/issues/149

References:
[mathoverflow](https://mathoverflow.net/questions/27851/polynomials-having-a-common-root-with-their-derivatives)
[paper](https://arxiv.org/pdf/math/0605090)


FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,91 @@
...
+-/
+
+open Polynomial
+
+/--
+A polynomial P satisfies the Casas-Alvero property if it shares a root with each
+of its derivatives up to order d-1, where d is the degree of P.
+-/
+def HasCasasAlveroProp {R : Type} [Field R] (P : R[X]) : Prop :=
+  ∀ i ∈ Finset.range P.natDegree, ∃ α : R, IsRoot P α ∧ IsRoot (derivative^[i] P) α
```
alreadydone: This is a bit different from the official statement which says P shares a factor (not necessarily linear) with its first derivatives. However I think they're equivalent since you can consider the algebraic closure; if P=(X-a)^n with `a` in the algebraic closure then `a` is actually in R because the X^(n-1) coefficient is `na` and the characteristic is 0.

nit: It's better to use K or F to denote a field rather than R.

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,95 @@
...
+The Casas-Alvero conjecture fails in positive characteristic p for polynomials of degree d > p.
+This was shown by Graf von Bothmer, Labs, Schicho, and van de Woestijne.
+
+Reference: [The Casas-Alvero conjecture for infinitely many degrees](https://arxiv.org/pdf/math/0605090)
+-/
+@[category research solved, AMS 12]
+theorem casas_alvero.positive_char_counterexample {p : ℕ} (hp : p.Prime) :
+    ∃ K : Type, ∃ (_ : Field K) (_ : CharP K p) (d : ℕ) (hd : d > p) (P : K[X]),
+      Monic P ∧ P.natDegree = d ∧ HasCasasAlveroProp P ∧
+      ¬∃ α : K, P = (X - C α) ^ P.natDegree := by
```
alreadydone: If you want this to work in positive characteristics, you have to use [Polynomial.hasseDeriv](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Polynomial/HasseDeriv.html#Polynomial.hasseDeriv) in the definition of HasCasasAlveroProp.

This statement also has a redundant variable `d`; it could be stated simply as
```suggestion
theorem casas_alvero.positive_char_counterexample {p : ℕ} (hp : p.Prime) :
    ∃ K : Type, ∃ (_ : Field K) (_ : CharP K p) (hd : P.natDegree > p) (P : K[X]),
      Monic P ∧ HasCasasAlveroProp P ∧ ¬∃ α : K, P = (X - C α) ^ P.natDegree := by
```

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,95 @@
...
+-/
+
+open Polynomial
+
+/--
+A polynomial P satisfies the Casas-Alvero property if it shares a root with each
+of its derivatives up to order d-1, where d is the degree of P.
+-/
+def HasCasasAlveroProp {F : Type} [Field F] (P : F[X]) : Prop :=
+  ∀ i ∈ Finset.range P.natDegree, ∃ α : F, IsRoot P α ∧ IsRoot (derivative^[i] P) α
```
alreadydone: Maybe switch F to K here too, and extract the `variable`s `{K : Type*} [Field K]` (I'm not sure whether there's a convention in this repo to use `Type*` or `Type`; `Type` may be closer to ZFC which is the generally accepted foundation? But I expect most questions' answer (true or false) are independent of which universe the objects are in.) Might be a good idea to extract `[CharZero K] (P : K[X]) (hP : Monic P)` too (you'll need `include hP`).

It might be better to add a version using `¬ IsCoprime P (derivative^[i] P)`, and show the conjecture is equivalent using both definitions (as an API lemma).

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,138 @@
...
+def HasCasasAlveroProp {K : Type} [Field K] (P : K[X]) : Prop :=
+  ∀ i ∈ Finset.range P.natDegree, ∃ α : K, IsRoot P α ∧ IsRoot (hasseDeriv i P) α
+
+/--
+Alternative characterization of the Casas-Alvero property using non-coprimality with derivatives.
+A polynomial has this property if it is not coprime with any of its Hasse derivatives up to
+order d-1, where d is the degree of P.
+-/
+def HasCasasAlveroPropCoprime {K : Type} [Field K] (P : K[X]) : Prop :=
+  ∀ i ∈ Finset.range P.natDegree, ¬ IsCoprime P (hasseDeriv i P)
```
shamEiNew: additional definition `HasCasasAlveroPropCoprime`

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,138 @@
...
+
+The conjecture has been proven for:
+* Degrees d ≤ 8
+* Degrees of the form p^k where p is prime
+* Degrees of the form 2p^k where p is prime
+
+The conjecture is false in positive characteristic p for polynomials of degree d > p.
+
+The conjecture is now claimed to be proven in this paper:
+* [The Casas-Alvero conjecture for infinitely many degrees](https://arxiv.org/pdf/2505.18000)
```
Paul-Lez: I think this ins't the paper you meant to link?

FormalConjectures/Paper/CasasAlvero.lean
Paul-Lez: That's good to add! To avoid having too many definitions for the same concept, I would suggest inlining this one in the equivalence statement `casas_alvero_iff_coprime` below

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,138 @@
...
+-/
+def HasCasasAlveroPropCoprime {K : Type} [Field K] (P : K[X]) : Prop :=
+  ∀ i ∈ Finset.range P.natDegree, ¬ IsCoprime P (hasseDeriv i P)
+
+/--
+The two characterizations of the Casas-Alvero property are equivalent.
+-/
+@[category API, AMS 12]
+lemma casas_alvero_iff_coprime {K : Type} [Field K] (P : K[X]) (hP : Monic P) :
+  HasCasasAlveroProp P ↔ HasCasasAlveroPropCoprime P := by
```
Paul-Lez:
```suggestion
    HasCasasAlveroProp P ↔ HasCasasAlveroPropCoprime P := by
```

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,138 @@
...
+
+The conjecture is now claimed to be proven in this paper:
+* [The Casas-Alvero conjecture for infinitely many degrees](https://arxiv.org/pdf/2505.18000)
+
+-/
+
+open Polynomial
+
+/--
+A polynomial P satisfies the Casas-Alvero property if it shares a root with each
```
Paul-Lez: Small nitpick (same for other docstrings below): usually we either add backticks arounds the maths, or use Latex, since this makes the doc pages nicer.

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,138 @@
...
+lemma casas_alvero_conjecture_iff {K : Type} [Field K] [CharZero K] (P : K[X]) (hP : Monic P) :
+    (HasCasasAlveroProp P → ∃ α : K, P = (X - C α) ^ P.natDegree) ↔
+    (HasCasasAlveroPropCoprime P → ∃ α : K, P = (X - C α) ^ P.natDegree) := by
+  constructor
+  · intro h1 h2
+    apply h1
+    exact (casas_alvero_iff_coprime P hP).mpr h2
+  · intro h1 h2
+    apply h1
+    exact (casas_alvero_iff_coprime P hP).mp h2
```
Paul-Lez: Since the only difference between these is that we're using one definition of the Casas-Alvero property rather than another, I think these two statements are a little redundant (especially since we've already stated that the two definitions are equivalent!)

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,138 @@
...
+The Casas-Alvero conjecture holds for polynomials of prime power degree.
+This was proved by Graf von Bothmer, Labs, Schicho, and van de Woestijne.
+
+Reference: [The Casas-Alvero conjecture for infinitely many degrees](https://arxiv.org/pdf/math/0605090)
+-/
+@[category research solved, AMS 12]
+theorem casas_alvero.prime_power {K : Type} [Field K] [CharZero K]
+    (p k : ℕ) (hp : p.Prime) (P : K[X]) (hP : Monic P) (hd : P.natDegree = p^k) :
+    HasCasasAlveroProp P → ∃ α : K, P = (X - C α) ^ P.natDegree := by
+  sorry
```
Paul-Lez:
```suggestion
@[category research solved, AMS 12]
theorem casas_alvero.prime_power {K : Type} [Field K] [CharZero K]
    (p k : ℕ) (hp : p.Prime) (P : K[X]) (hP : Monic P) (hd : P.natDegree = p^k)
    (hP' : HasCasasAlveroProp P) : ∃ α : K, P = (X - C α) ^ P.natDegree := by
  sorry
```

FormalConjectures/Paper/CasasAlvero.lean
Paul-Lez: I think using `Type*` is reasonable. 

Re stating the two versions of the conjecture are equivalent, [I'm tempted to say](https://github.com/google-deepmind/formal-conjectures/pull/150#discussion_r2131829802) that having the theorem that the two definitions are equivalent should be enough.

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,104 @@
...
+  ∀ i ∈ Finset.range P.natDegree, ∃ α : K, IsRoot P α ∧ IsRoot (hasseDeriv i P) α
+
+/--
+The Casas-Alvero property is equivalent to the polynomial not being coprime with any of its
+Hasse derivatives up to order `d-1`.
+-/
+@[category API, AMS 12]
+lemma casas_alvero_iff_coprime {K : Type*} [Field K] (P : K[X]) (hP : Monic P) :
+  HasCasasAlveroProp P ↔ ∀ i ∈ Finset.range P.natDegree, ¬ IsCoprime P (hasseDeriv i P) := by
+  sorry
```
alreadydone: I think we need to assume K is CharZero and quantify over all fields (one implication is easy but the other requires passing to the algebraic closure); otherwise we might only prove this equivalence after we've proven the conjecture. Please check out my proofs of API lemmas [here](https://github.com/shamEiNew/formal-conjectures/compare/CasasAlveroConjecture...alreadydone:formal-conjectures:patch-2). There's one sorry remaining which you're welcome to fill!

alreadydone:
Filled the sorry and cleaned up a bit more [here](https://github.com/google-deepmind/formal-conjectures/commit/ee336f460775b174852534f4cb4f720f9f7b9e91). I think it's ready to go.

shamEiNew:
@alreadydone thanks for the update. I've merged those changes into this PR.

mo271:
@shamEiNew perhaps it is good to resolve some of the conversations above, if they are already addressed in the code, to make reviewing easier

shamEiNew:
> @shamEiNew perhaps it is good to resolve some of the conversations above, if they are already addressed in the code, to make reviewing easier

@mo271  I've resolved the conversation associated with all changes. If there are any other requirements, please let me know.

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,160 @@
...
+  refine ⟨?_, HasCasasAlveroPropᵣ.hasCasasAlveroProp _⟩
+  simp_rw [HasCasasAlveroProp, HasCasasAlveroPropᵣ,
+    isCoprime_iff_aeval_ne_zero_of_isAlgClosed K K, coe_aeval_eq_eval]
+  push_neg
+  exact (· · ·)
+
+universe u in
+/--
+The Casas-Alvero property is equivalent to the polynomial not being coprime with any of its
+Hasse derivatives up to order `d-1`.
```
alreadydone:
```suggestion
Whether we use `HasCasasAlveroProp` or `HasCasasAlveroPropᵣ` to state the Casas-Alvero conjecture,
we obtain equivalent statements.
```

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,160 @@
...
+-/
+@[category research solved, AMS 12]
+theorem casas_alvero.double_prime_power (p k : ℕ) (hp : p.Prime) (hd : P.natDegree = 2 * p^k) :
+    HasCasasAlveroProp P → ∃ α : K, P = (X - C α) ^ P.natDegree := by
+  sorry
+
+end conjecture
+
+/--
+The Casas-Alvero conjecture fails in positive characteristic `p` for polynomials of degree `d > p`.
```
alreadydone: I don't think it fails for every degree `d > p` ...
```suggestion
The Casas-Alvero conjecture fails in positive characteristic `p` for a polynomial of degree `p + 1`.
```

alreadydone (approved):
LGTM otherwise

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,160 @@
...
+A polynomial `P` satisfies the Casas-Alvero property if it shares a factor with each
+of its Hasse derivatives up to order `d-1`, where `d` is the degree of `P`.
+-/
+def HasCasasAlveroProp (P : K[X]) : Prop :=
+  ∀ i ∈ Finset.range P.natDegree, ¬ IsCoprime P (P.hasseDeriv i)
+
+/-- An stronger version of the Casas-Alvero property, which requires that the polynomial `P`
+shares a root with each of its Hasse derivatives up to order `deg P - 1`.
+The subscript `r` indicates "root" in the definition. -/
+def HasCasasAlveroPropᵣ (P : K[X]): Prop :=
```
Paul-Lez:
```suggestion
def HasCasasAlveroPropᵣ (P : K[X]) : Prop :=
```

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,160 @@
...
+variable {K L : Type*} [Field K] [Field L] {f : K →+* L}
+
+/--
+A polynomial `P` satisfies the Casas-Alvero property if it shares a factor with each
+of its Hasse derivatives up to order `d-1`, where `d` is the degree of `P`.
+-/
+def HasCasasAlveroProp (P : K[X]) : Prop :=
+  ∀ i ∈ Finset.range P.natDegree, ¬ IsCoprime P (P.hasseDeriv i)
+
+/-- An stronger version of the Casas-Alvero property, which requires that the polynomial `P`
```
Paul-Lez:
```suggestion
/-- A stronger version of the Casas-Alvero property, which requires that the polynomial `P`
```

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,160 @@
...
+variable [CharZero K] (P : K[X]) (hP : Monic P)
+include hP
+
+/--
+The Casas-Alvero conjecture states that in characteristic zero, if a monic polynomial `P`
+has the Casas-Alvero property, then `P = (X - α)ᵈ` for some `α`.
+-/
+@[category research open, AMS 12]
+theorem casas_alvero_conjecture :
+    HasCasasAlveroProp P → ∃ α : K, P = (X - C α) ^ P.natDegree := by
```
Paul-Lez:
```suggestion
theorem casas_alvero_conjecture (hP' : HasCasasAlveroProp P) :
    ∃ α : K, P = (X - C α) ^ P.natDegree := by
```

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,160 @@
...
+
+/--
+The Casas-Alvero conjecture holds for polynomials of prime power degree.
+This was proved by Graf von Bothmer, Labs, Schicho, and van de Woestijne.
+
+Reference: [The Casas-Alvero conjecture for infinitely many degrees](https://arxiv.org/pdf/math/0605090)
+-/
+@[category research solved, AMS 12]
+theorem casas_alvero.prime_power (p k : ℕ) (hp : p.Prime) (hd : P.natDegree = p^k) :
+    HasCasasAlveroProp P → ∃ α : K, P = (X - C α) ^ P.natDegree := by
```
Paul-Lez:
```suggestion
theorem casas_alvero.prime_power (p k : ℕ) (hp : p.Prime) (hd : P.natDegree = p^k)
    (hP' : HasCasasAlveroProp P) : ∃ α : K, P = (X - C α) ^ P.natDegree := by
```

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,160 @@
...
+
+/--
+The Casas-Alvero conjecture holds for polynomials of degree `2p^k` where `p` is prime.
+This was proved by Graf von Bothmer, Labs, Schicho, and van de Woestijne.
+
+Reference: [The Casas-Alvero conjecture for infinitely many degrees](https://arxiv.org/pdf/math/0605090)
+-/
+@[category research solved, AMS 12]
+theorem casas_alvero.double_prime_power (p k : ℕ) (hp : p.Prime) (hd : P.natDegree = 2 * p^k) :
+    HasCasasAlveroProp P → ∃ α : K, P = (X - C α) ^ P.natDegree := by
```
Paul-Lez:
```suggestion
theorem casas_alvero.double_prime_power (p k : ℕ) (hp : p.Prime) (hd : P.natDegree = 2 * p^k)
    (hP' : HasCasasAlveroProp P) : ∃ α : K, P = (X - C α) ^ P.natDegree := by
```

Paul-Lez (approved):
LGTM - thanks!

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,161 @@
...
+
+/--
+The Casas-Alvero conjecture holds for polynomials of prime power degree.
+This was proved by Graf von Bothmer, Labs, Schicho, and van de Woestijne.
+
+Reference: [The Casas-Alvero conjecture for infinitely many degrees](https://arxiv.org/pdf/math/0605090)
+-/
+@[category research solved, AMS 12]
+theorem casas_alvero.prime_power (p k : ℕ) (hp : p.Prime) (hd : P.natDegree = p^k)
+(hP' : HasCasasAlveroProp P) : ∃ α : K, P = (X - C α) ^ P.natDegree := by
```
alreadydone:
```suggestion
    (hP' : HasCasasAlveroProp P) : ∃ α : K, P = (X - C α) ^ P.natDegree := by
```

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,161 @@
...
+
+/-- A stronger version of the Casas-Alvero property, which requires that the polynomial `P`
+shares a root with each of its Hasse derivatives up to order `deg P - 1`.
+The subscript `r` indicates "root" in the definition. -/
+
+def HasCasasAlveroPropᵣ (P : K[X]) : Prop :=
+  ∀ i ∈ Finset.range P.natDegree, ∃ α : K, IsRoot P α ∧ IsRoot (P.hasseDeriv i) α
+
+@[category API, AMS 12]
+theorem HasCasasAlveroPropᵣ.hasCasasAlveroProp (P : K[X])
```
alreadydone:
```suggestion
theorem HasCasasAlveroPropᵣ.hasCasasAlveroProp {P : K[X]}
```

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,161 @@
...
+@[category API, AMS 12]
+theorem HasCasasAlveroProp.map_iff {P : K[X]} :
+    HasCasasAlveroProp (P.map f) ↔ HasCasasAlveroProp P := by
+  simp_rw [HasCasasAlveroProp]; congr!; · simp
+  rw [hasseDeriv_map, isCoprime_map]
+
+@[category API, AMS 12]
+theorem hasCasasAlveroProp_iffᵣ {P : K[X]} [IsAlgClosed K] :
+    HasCasasAlveroProp P ↔ HasCasasAlveroPropᵣ P := by
+  refine ⟨?_, HasCasasAlveroPropᵣ.hasCasasAlveroProp _⟩
```
callesonne:
```suggestion
  refine ⟨?_, HasCasasAlveroPropᵣ.hasCasasAlveroProp⟩
```
This needs to be updated given the previous suggestion

FormalConjectures/ForMathlib/Algebra/Polynomial/HasseDeriv.lean
```diff
@@ -0,0 +1,22 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import Mathlib.Algebra.Polynomial.HasseDeriv
+
+open scoped Polynomial in -- probably removable in the mathlib file
+theorem Polynomial.hasseDeriv_map {R S : Type*} [Semiring R] [Semiring S]
+    {f : R →+* S} {k : ℕ} {p : R[X]} : (p.map f).hasseDeriv k = (p.hasseDeriv k).map f := by
```
callesonne, line 21:
```suggestion
    (f : R →+* S) (k : ℕ) (p : R[X]) : (p.map f).hasseDeriv k = (p.hasseDeriv k).map f := by
```
These should be explicit arguments.

FormalConjectures/Paper/CasasAlvero.lean
```diff
@@ -0,0 +1,161 @@
...
+  intro i hi coprime
+  simp_rw [HasCasasAlveroPropᵣ, ← dvd_iff_isRoot] at hca
+  have ⟨α, hα, hαi⟩ := hca i hi
+  exact not_isUnit_X_sub_C _ (coprime.isUnit_of_dvd' hα hαi)
+
+@[category API, AMS 12]
+theorem HasCasasAlveroProp.map_iff {P : K[X]} :
+    HasCasasAlveroProp (P.map f) ↔ HasCasasAlveroProp P := by
+  simp_rw [HasCasasAlveroProp]; congr!; · simp
+  rw [hasseDeriv_map, isCoprime_map]
```
callesonne:
```suggestion
  simp_rw [HasCasasAlveroProp]
  congr!
  · exact natDegree_map _
  · rw [hasseDeriv_map _ _ _, isCoprime_map]
```
Updating `hasseDeriv_map` based on my other suggestion, then I cleaned up the proof a bit.

callesonne:
and thanks for your contribution! :)

FormalConjectures/Paper/CasasAlvero.lean
alreadydone: I don't think the _ _ _ is necessary :)

FormalConjectures/ForMathlib/Algebra/Polynomial/HasseDeriv.lean
alreadydone: I usually make arguments that appear in both sides of the equality implicit, but I understand that not everyone adopts this practice. It appears that the author(s) of the hassederiv file doesn't, so your change is reasonable.

callesonne: Ah okay I didn't consider this, so then leaving it as-is is fine as well. Then my suggestion below needs modifying slightly as well (to remove the `_ _ _`).

FormalConjectures/Paper/CasasAlvero.lean
callesonne: It is given the above suggestion, but yes currently there are no errors!

shamEiNew:
> and thanks for your contribution! :)

are there any further changes to be made?

callesonne:
> > and thanks for your contribution! :)
> 
> are there any further changes to be made?

Yes, see my reply above. Its optional if you want the `hasseDeriv` lemma to take implicit or explicit parameters, but I would still implement the second suggestion (you will need to modify it yourself if you prefer `hasseDeriv` to keep implicit parameters).

FormalConjectures/ForMathlib/Algebra/Polynomial/HasseDeriv.lean
alreadydone: I tried and the _ _ _ are unnecessary regardless of implicit/explicitness.

PR #153, opened 2025-06-03
Add Gromov's polynomial growth theorem

AlexKontorovich:

PR #154, opened 2025-06-04
Schur's theorem on Galois groups of truncated exponential polynomials

AlexKontorovich:
As requested by Nick Katz

### DIFF
--- a/FormalConjectures/Other/SchurTruncatedExponential.lean
+++ b/FormalConjectures/Other/SchurTruncatedExponential.lean
@@ -48,7 +48,7 @@ exponential polynomial over `ℚ`. Then for `n ≥ 2`:
 - If `n ≡ 0 (mod 4)`, the Galois group of `f_n` is isomorphic to the alternating group `A_n`
 - If `n ≢ 0 (mod 4)`, the Galois group of `f_n` is isomorphic to the symmetric group `S_n`
 -/
-@[category research solved, AMS 11]
+@[category research solved, AMS 12]
 theorem schur_truncatedExp_galoisGroup_equiv (n : ℕ) (hn : n ≥ 2) :
   let f := truncatedExp n
   if n % 4 = 0 then
--- a/FormalConjectures/Other/SchurTruncatedExponential.lean
+++ b/FormalConjectures/Other/SchurTruncatedExponential.lean
@@ -1,0 +1,60 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Schur's theorem on Galois groups of truncated exponential polynomials
+
+*Reference:* (https://math.stackexchange.com/questions/2814220)
+
+*Reference* (https://mathoverflow.net/questions/477077)
+-/
+
+/-
+Note: This was asked by Nick Katz. Quasi-autoformalized using Claude 4.0 Sonnet.
+-/
+
+open Polynomial
+
+/--
+The truncated exponential polynomial `truncatedExp n` is
+given by `∑_{j=0}^{n} x^j / j!` over `ℚ`, which is the
+`n`-th partial sum of the Taylor series for the exponential function `e^x`.
+Note: It's very important that the upper limit is `j=n`,
+not `j=n-1`!!
+-/
+noncomputable def truncatedExp (n : ℕ) : ℚ[X] :=
+  ∑ j ∈ Finset.range (n + 1), (1 / j.factorial : ℚ) • X ^ j
+
+/--
+**Schur's Theorem (1924):**
+Let `f_n(x) = ∑_{j=0}^{n-1} x^j/j!` be the `n`-th truncated
+exponential polynomial over `ℚ`. Then for `n ≥ 2`:
+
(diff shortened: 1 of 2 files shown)


FormalConjectures/Other/SchurTruncatedExponential.lean
```diff
@@ -0,0 +1,60 @@
...
+`n`-th partial sum of the Taylor series for the exponential function `e^x`.
+Note: It's very important that the upper limit is `j=n`,
+not `j=n-1`!!
+-/
+noncomputable def truncatedExp (n : ℕ) : ℚ[X] :=
+  ∑ j ∈ Finset.range (n + 1), (1 / j.factorial : ℚ) • X ^ j
+
+/--
+**Schur's Theorem (1924):**
+Let `f_n(x) = ∑_{j=0}^{n-1} x^j/j!` be the `n`-th truncated
```
AlexKontorovich:
```suggestion
Let `f_n(x) = ∑_{j=0}^n x^j/j!` be the `n`-th truncated
```

FormalConjectures/Other/SchurTruncatedExponential.lean
```diff
@@ -0,0 +1,60 @@
...
+**Schur's Theorem (1924):**
+Let `f_n(x) = ∑_{j=0}^n x^j/j!` be the `n`-th truncated
+exponential polynomial over `ℚ`. Then for `n ≥ 2`:
+
+- If `n ≡ 0 (mod 4)`, the Galois group of `f_n` is isomorphic to the alternating group `A_n`
+- If `n ≢ 0 (mod 4)`, the Galois group of `f_n` is isomorphic to the symmetric group `S_n`
+-/
+@[category research solved, AMS 12]
+theorem schur_truncatedExp_galoisGroup_equiv (n : ℕ) (hn : n ≥ 2) :
+  let f := truncatedExp n
```
callesonne:
```suggestion
  letI f := truncatedExp n
```
Putting `letI` here makes it so that the `let` does not appear in the goal statement during proof.

FormalConjectures/Other/SchurTruncatedExponential.lean
```diff
@@ -0,0 +1,60 @@
...
+-/
+
+open Polynomial
+
+/--
+The truncated exponential polynomial `truncatedExp n` is
+given by `∑_{j=0}^{n} x^j / j!` over `ℚ`, which is the
+`n`-th partial sum of the Taylor series for the exponential function `e^x`.
+Note: It's very important that the upper limit is `j=n`,
+not `j=n-1`!!
```
callesonne:
```suggestion
```
I would not keep this in, I'm guessing this refers to its usage in the theorem below, but there you (Claude?) have explicitly written `f_n(x) = ∑_{j=0}^n x^j/j!` anyways, so it should be clear which one is necessary.

FormalConjectures/Other/SchurTruncatedExponential.lean
```diff
@@ -0,0 +1,60 @@
...
+-/
+noncomputable def truncatedExp (n : ℕ) : ℚ[X] :=
+  ∑ j ∈ Finset.range (n + 1), (1 / j.factorial : ℚ) • X ^ j
```
callesonne:
```suggestion
open Polynomial

open scoped Nat

/--
The truncated exponential polynomial `truncatedExp n` is
given by `∑_{j=0}^{n} x^j / j!` over `ℚ`, which is the
`n`-th partial sum of the Taylor series for the exponential function `e^x`.
Note: It's very important that the upper limit is `j=n`,
not `j=n-1`!!
-/
noncomputable def truncatedExp (n : ℕ) : ℚ[X] :=
  ∑ j ∈ Finset.range (n + 1), (1 / j ! : ℚ) • X ^ j
```
using factorial notation

FormalConjectures/Other/SchurTruncatedExponential.lean
```diff
@@ -0,0 +1,58 @@
...
+
+*Reference* (https://mathoverflow.net/questions/477077)
+-/
+
+/-
+Note: This was asked by Nick Katz. Quasi-autoformalized using Claude 4.0 Sonnet.
+-/
+
+open Polynomial
+
```
AlexKontorovich, line 32:
```suggestion

open scoped Nat
```

FormalConjectures/Other/SchurTruncatedExponential.lean
```diff
@@ -0,0 +1,59 @@
...
+open scoped Nat
+/--
```
AlexKontorovich, line 35:
```suggestion

/--
```

FormalConjectures/Other/SchurTruncatedExponential.lean
```diff
@@ -0,0 +1,60 @@
...
+
+open scoped Nat
+
+/--
+The truncated exponential polynomial `truncatedExp n` is
+given by `∑_{j=0}^{n} x^j / j!` over `ℚ`, which is the
+`n`-th partial sum of the Taylor series for the exponential function `e^x`.
+-/
+noncomputable def truncatedExp (n : ℕ) : ℚ[X] :=
+  ∑ j ∈ Finset.range (n + 1), (1 / j.factorial : ℚ) • X ^ j
```
AlexKontorovich:
```suggestion
  ∑ j ∈ Finset.range (n + 1), (1 / j ! : ℚ) • X ^ j
```

mo271:
Many thanks @AlexKontorovich (and by proxy Nick Katz) for the contribution! The focus on the repo we target is *open* conjectures, with two exceptions: 
 - solved problems as variants of open conjectures and 
 - solved problems from problem collections. 
 I do think we can be lenient in that rule and this pr and #153 seem to be good contributions. Perhaps they could be made even better by including a related *open* problem? (If there is anything that comes to mind here...)

AlexKontorovich:
> Many thanks @AlexKontorovich (and by proxy Nick Katz) for the contribution! The focus on the repo we target is _open_ conjectures, with two exceptions:
> 
> * solved problems as variants of open conjectures and
> * solved problems from problem collections.
>   I do think we can be lenient in that rule and this pr and [Add Gromov's polynomial growth theorem #153](https://github.com/google-deepmind/formal-conjectures/pull/153) seem to be good contributions. Perhaps they could be made even better by including a related _open_ problem? (If there is anything that comes to mind here...)

Got it, more open problems. Will do, thanks!

PR #155, opened 2025-06-04
chore: switch to round in LittlewoodConjecture

alreadydone:
not sure if/how I should adjust the docstring.
also the theorem statement fits in 100 chars.


mo271:
I'd say it is fine to leave the docstring as is

PR #161, opened 2025-06-04
Fix: Bateman-Horn Conjecture

AlexKontorovich:
Thanks to Will Sawin for pointing out that both Claude and I missed division by the product of degrees in the constant!

### DIFF
--- a/FormalConjectures/Wikipedia/BatemanHornConjecture.lean
+++ b/FormalConjectures/Wikipedia/BatemanHornConjecture.lean
@@ -81,7 +81,7 @@ number of residue classes modulo p for which at least one polynomial vanishes.
 
 The compatibility condition ensures that for each prime p, there exists some integer n
 such that p does not divide the product f₁(n)·f₂(n)·...·fₖ(n), which guarantees the
-infinite product converges to a positive value.
+infinite product converges to a positive value. -/
 
 @[category research open, AMS 11 12]
 theorem bateman_horn_conjecture
--- a/FormalConjectures/Wikipedia/BatemanHornConjecture.lean
+++ b/FormalConjectures/Wikipedia/BatemanHornConjecture.lean
@@ -65,7 +65,7 @@ noncomputable def CountSimultaneousPrimes (polys : Finset ℤ[X]) (x : ℝ) : �
 
 /-- **The Bateman-Horn Conjecture**
 
-Given a finite collection of distinct irreducible polynomials f₁, f₂, ..., fₖ ∈ ℤ[X]
+Given a finite collection of distinct, irreducible, non-constant polynomials f₁, f₂, ..., fₖ ∈ ℤ[X]
 with positive leading coefficients that satisfy the compatibility condition, the number
 of positive integers n ≤ x for which all polynomials f₁(n), f₂(n), ..., fₖ(n) are
 simultaneously prime is asymptotic to:
--- a/FormalConjectures/Wikipedia/BatemanHornConjecture.lean
+++ b/FormalConjectures/Wikipedia/BatemanHornConjecture.lean
@@ -33,18 +33,26 @@ open Polynomial Asymptotics Filter Topology
 def IsIrreducibleWithPosLeading (f : ℤ[X]) : Prop :=
   Irreducible f ∧ 0 < f.leadingCoeff
 
-/-- The compatibility condition for a finite set `S` of polynomials in the Bateman-Horn conjecture. 
+/-- The compatibility condition for a finite set `S` of polynomials in the Bateman-Horn conjecture.
 This states that for all primes `p`, there exists an `n` such that `∏ f ∈ S, f.eval n` is non-zero modulo `p`. -/
 def SatisfiesCompatibilityCondition (polys : Finset ℤ[X]) : Prop :=
   ∀ p : ℕ, Nat.Prime p → ∃ n : ℤ, ¬↑p ∣ (polys.prod id).eval n
 
 /-- `OmegaP S p` counts the number of residue classes mod `p` where at least one polynomial in `S` vanishes. -/
 noncomputable def OmegaP (polys : Finset ℤ[X]) (p : ℕ) : ℕ :=
-  {n : ZMod p | ∃ f ∈ polys, (f.map (Int.castRingHom (ZMod p))).eval n = 0}.ncard 
+  {n : ZMod p | ∃ f ∈ polys, (f.map (Int.castRingHom (ZMod p))).eval n = 0}.ncard
 
-/-- The Bateman-Horn constant of a set of polynomials `S`. This is defined as the infinite product over all primes: 
-$$\prod_p (1 - \frac{1}{p}) ^ {|S|} (1 - \frac{\omega_p(S)}{p}$$ where $\omega_p(S)}{p}$ is the number of residue classes mod $p$ where at least one polynomial in $S$ vanishes. -/
+/-- The product of degrees of polynomials in a finite set. -/
+def DegreesProduct (polys : Finset ℤ[X]) : ℕ :=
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/BatemanHornConjecture.lean
```diff
@@ -57,22 +65,27 @@ noncomputable def CountSimultaneousPrimes (polys : Finset ℤ[X]) (x : ℝ) : 
...
+The compatibility condition ensures that for each prime p, there exists some integer n
+such that p does not divide the product f₁(n)·f₂(n)·...·fₖ(n), which guarantees the
+infinite product converges to a positive value.
 
-The compatibility condition ensures that for each prime p, there exists some integer n 
-such that p does not divide the product f₁(n)·f₂(n)·...·fₖ(n), which guarantees the 
-infinite product converges to a positive value. -/
+This conjecture generalizes both Dirichlet's theorem on primes in arithmetic progressions
+(single linear polynomial case) and the Hardy-Littlewood k-tuple conjecture (when all
+polynomials are of the form X + aᵢ). -/
```
AlexKontorovich:
```suggestion
```

FormalConjectures/Wikipedia/BatemanHornConjecture.lean
```diff
@@ -57,22 +65,24 @@ noncomputable def CountSimultaneousPrimes (polys : Finset ℤ[X]) (x : ℝ) : 
...
+    C = (1/D) · ∏ₚ (1 - 1/p)^(-k) · (1 - ωₚ/p)
 
-    C = ∏ₚ (1 - 1/p)^(-k) · (1 - ωₚ/p)
+Here D = ∏ᵢ deg(fᵢ) is the product of the degrees of the polynomials, and ωₚ is the
+number of residue classes modulo p for which at least one polynomial vanishes.
 
-Here ωₚ is the number of residue classes modulo p for which at least one polynomial vanishes.
+The compatibility condition ensures that for each prime p, there exists some integer n
+such that p does not divide the product f₁(n)·f₂(n)·...·fₖ(n), which guarantees the
+infinite product converges to a positive value.
```
AlexKontorovich:
```suggestion
infinite product converges to a positive value. -/
```

Paul-Lez:
Thanks!

Paul-Lez (approved):

FormalConjectures/Wikipedia/BatemanHornConjecture.lean
```diff
@@ -57,21 +65,22 @@ noncomputable def CountSimultaneousPrimes (polys : Finset ℤ[X]) (x : ℝ) : 
...
+Here D = ∏ᵢ deg(fᵢ) is the product of the degrees of the polynomials, and ωₚ is the
+number of residue classes modulo p for which at least one polynomial vanishes.
 
-The compatibility condition ensures that for each prime p, there exists some integer n 
-such that p does not divide the product f₁(n)·f₂(n)·...·fₖ(n), which guarantees the 
+The compatibility condition ensures that for each prime p, there exists some integer n
+such that p does not divide the product f₁(n)·f₂(n)·...·fₖ(n), which guarantees the
 infinite product converges to a positive value. -/
 @[category research open, AMS 11 12]
 theorem bateman_horn_conjecture
```
Paul-Lez, line 86: Actually, shouldn't we also assume that the polynomials are non-constant?

AlexKontorovich: Sure (though if they're constant, they're either always prime or never -- in which case the singular series vanishes... But then I'd have to think about whether Lean's asymptotic works the same way it does in natural language), better safe than sorry.

FormalConjectures/Wikipedia/BatemanHornConjecture.lean
```diff
@@ -33,18 +33,26 @@ open Polynomial Asymptotics Filter Topology
 def IsIrreducibleWithPosLeading (f : ℤ[X]) : Prop :=
   Irreducible f ∧ 0 < f.leadingCoeff
```
AlexKontorovich:
```suggestion
  Irreducible f ∧ 0 < f.leadingCoeff ∧ 0 < f.natDegree
```

FormalConjectures/Wikipedia/BatemanHornConjecture.lean
```diff
@@ -31,20 +31,28 @@ open Polynomial Asymptotics Filter Topology
 
 /-- Definition of irreducible polynomial with positive leading coefficient -/
```
AlexKontorovich:
```suggestion
/-- Definition of irreducible, non-constant polynomial with positive leading coefficient -/
```

FormalConjectures/Wikipedia/BatemanHornConjecture.lean
```diff
@@ -57,21 +65,22 @@ noncomputable def CountSimultaneousPrimes (polys : Finset ℤ[X]) (x : ℝ) : 
 
 /-- **The Bateman-Horn Conjecture**
 
-Given a finite collection of distinct irreducible polynomials f₁, f₂, ..., fₖ ∈ ℤ[X] 
-with positive leading coefficients that satisfy the compatibility condition, the number 
-of positive integers n ≤ x for which all polynomials f₁(n), f₂(n), ..., fₖ(n) are 
+Given a finite collection of distinct irreducible polynomials f₁, f₂, ..., fₖ ∈ ℤ[X]
```
AlexKontorovich:
```suggestion
Given a finite collection of distinct, irreducible, non-constant polynomials f₁, f₂, ..., fₖ ∈ ℤ[X]
```

FormalConjectures/Wikipedia/BatemanHornConjecture.lean
AlexKontorovich: Actually we dropped distinctness at some point. But it's needed (or else the asymptotic is wrong!). Boy it's hard to get formalized statements *right*!!...

FormalConjectures/Wikipedia/BatemanHornConjecture.lean
```diff
@@ -57,27 +65,29 @@ noncomputable def CountSimultaneousPrimes (polys : Finset ℤ[X]) (x : ℝ) : 
...
-such that p does not divide the product f₁(n)·f₂(n)·...·fₖ(n), which guarantees the 
+The compatibility condition ensures that for each prime p, there exists some integer n
+such that p does not divide the product f₁(n)·f₂(n)·...·fₖ(n), which guarantees the
 infinite product converges to a positive value. -/
 @[category research open, AMS 11 12]
 theorem bateman_horn_conjecture
     (polys : Finset ℤ[X])
     (h_nonempty : polys.Nonempty)
     (h_irreducible : ∀ f ∈ polys, IsIrreducibleWithPosLeading f)
+    (h_distinct : ∀ f ∈ polys, ∀ g ∈ polys, f ≠ g)
```
callesonne: I have not looked into the whole statement, but is this really what you want to say? This would imply that `polys` is empty (since it says in particular that `∀ f ∈ polys, f ≠ f`).

callesonne: If I'm not misunderstanding, I think you get distinctness for free in this formulation since you are using a `Finset`

AlexKontorovich: Whoops! Yes the set structure automatically assumes distinctness! (It's usually stated in terms of a sequence of polynomials...) So indeed it's not needed...

AlexKontorovich:
```suggestion
    (h_irreducible : ∀ f ∈ polys, IsIrreducibleWithPosLeading f)
```

PR #165, opened 2025-06-05
feat: add Erdos 541

### LINKED ISSUE #143: Erdos 541
### What is the conjecture
https://www.erdosproblems.com/541

### Prerequisites needed
None

* ams-11

### Choose either option

callesonne:
This closes #143

### DIFF
--- a/FormalConjectures/ErdosProblems/541.lean
+++ b/FormalConjectures/ErdosProblems/541.lean
@@ -1,0 +1,62 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 541
+
+*Reference:* [erdosproblems.com/541](https://www.erdosproblems.com/541)
+-/
+
+open Filter
+
+/--
+Let $a_1, \dots, a_p$ be (not necessarily distinct) residues modulo a prime $p$, such that there
+exists some $r$ so that if $S \subseteq [p]$ is non-empty and
+$$\sum_{i \in S} a_i \equiv 0 \pmod{p}$$
+then $|S| = r$.
+
+Must there be at most two distinct residues amongst the $a_i$?
+-/
+@[category research solved, AMS 11]
+theorem erdos_541 (p : ℕ) [Fact p.Prime] (a : Fin p → ZMod p)
+    (ha₀ : ∃ r, ∀ (S : Finset (Fin p)), S ≠ ∅ → ∑ i ∈ S, a i = 0 → S.card = r) :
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/541.lean
```diff
@@ -0,0 +1,62 @@
...
+
+/--
+Let $a_1, \dots, a_p$ be (not necessarily distinct) residues modulo a prime $p$, such that there
+exists some $r$ so that if $S \subseteq [p]$ is non-empty and
+$$\sum_{i \in S} a_i \equiv 0 \pmod{p}$$
+then $|S| = r$.
+
+Must there be at most two distinct residues amongst the $a_i$?
+-/
+@[category research solved, AMS 11]
```
Paul-Lez, line 35: Nit: this should be using the `answer( )` formalism.

FormalConjectures/ErdosProblems/541.lean
```diff
@@ -0,0 +1,62 @@
...
+      (Set.range a).ncard ≤ 2 := by
+  sorry
+
+/-- This was proved by Erdős and Szemerédi [ErSz76] for p sufficiently large.
+
+[ErSz76] Erd\H os, E. and Szemerédi, E., On a problem of Graham.
+Publ. Math. Debrecen (1976), 123--127.
+-/
+@[category research solved, AMS 11]
+theorem erdos_541.variants.large_primes : ∀ᶠ p in atTop, Fact p.Prime → ∀ a : Fin p → ZMod p,
```
Paul-Lez: Does this work?
```suggestion
theorem erdos_541.variants.large_primes : ∀ᶠ p in atTop, p.Prime → ∀ a : Fin p → ZMod p,
```

callesonne: Yes, but given that the statement is about `ZMod p` I think that its more useful to get `Fact p.Prime` instead.

FormalConjectures/ErdosProblems/541.lean
callesonne: Right, not even a nit, thanks!

FormalConjectures/ErdosProblems/541.lean
```diff
@@ -0,0 +1,62 @@
...
```
callesonne:
```suggestion
theorem erdos_541.variants.large_primes : ∀ᶠ p in atTop, [Fact p.Prime] → ∀ a : Fin p → ZMod p,
```

FormalConjectures/ErdosProblems/541.lean
Paul-Lez: I agree that it's useful within the proof, but I'm a little wary of adding the assumption that way in the statement, since it's easy to add a `Fact ...` instance within the proof. Generally I tend to only state things with `Fact` if that is required to make the statement compile :)

FormalConjectures/ErdosProblems/541.lean
```diff
@@ -0,0 +1,62 @@
...
+      (Set.range a).ncard ≤ 2 := by
+  sorry
+
+/-- This was proved by Erdős and Szemerédi [ErSz76] for p sufficiently large.
+
+[ErSz76] Erd\H os, E. and Szemerédi, E., On a problem of Graham.
+Publ. Math. Debrecen (1976), 123--127.
+-/
+@[category research solved, AMS 11]
+theorem erdos_541.variants.large_primes : ∀ᶠ p in atTop, [Fact p.Prime] → ∀ a : Fin p → ZMod p,
```
callesonne:
```suggestion
theorem erdos_541.variants.large_primes : ∀ᶠ p in atTop, p.Prime → ∀ a : Fin p → ZMod p,
```

PR #168, opened 2025-06-05
Add Modularity Conjecture 

CBirkbeck:
### What is the conjecture
State the Shimura--Taniyama--Weil conjecture  (also known as the modularity conjecture/theorem) for elliptic curves over the rational numbers. 

https://en.wikipedia.org/wiki/Modularity_theorem

This is a very hard, known theorem in the case of the rational number and many  other cases, but open in general (and much harder to state!). The version we state is stronger than what is needed for FLT.

* ams-11

### DIFF
--- a/FormalConjectures/Wikipedia/ModularityConjecture.lean
+++ b/FormalConjectures/Wikipedia/ModularityConjecture.lean
@@ -39,6 +39,7 @@ open scoped UpperHalfPlane Real ModularForm CongruenceSubgroup
 
 noncomputable section
 
+/--Then `n`-th Fourier coefficient of a modular forms (around the cusp at infinity). -/
 def modularFormAn (n : ℕ) {N : ℕ} {k : ℤ} (f : CuspForm (Gamma0 N) k) : ℂ :=
   (qExpansion N f).coeff ℂ n
 
@@ -49,6 +50,7 @@ conductor of the elliptic curve saves us.-/
 def ratRed (q : ℚ) (p : ℕ) : ZMod p :=
   (q.num : ZMod p) * (q.den : ZMod p)⁻¹
 
+/-- The set of points on an elliptic curve over `ZMod n`. -/
 def setOfPointsModN (E : WeierstrassCurve ℚ) [E.IsElliptic] (n : ℕ) :=
   {P : ZMod n × ZMod n |
     let ⟨x, y⟩ := P
--- a/FormalConjectures/Wikipedia/ModularityConjecture.lean
+++ b/FormalConjectures/Wikipedia/ModularityConjecture.lean
@@ -1,0 +1,82 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Modularity conjecture
+
+The **Modularity conjecture** (also know as the Shimura--Taniyama--Weil conjecture) states that
+every rational elliptic curve is modular, meaning that it can be
+associated with a modular form. We state the `a_p` version of the conjecture, which relates the
+coefficients of the modular form to the number of points on the elliptic curve over finite fields.
+
+Since we don't have the conductor of the elliptic curve, our definition of `a_p(E)` differs from
+that in the literature at primes of bad reduction. For this reason, we state the conjecture with the
+assumption that `p ∤ N`, in order to give an equivalent statement.
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Modularity_theorem)
+-/
+
+
+open Complex CongruenceSubgroup ModularFormClass
+open scoped UpperHalfPlane Real ModularForm CongruenceSubgroup
+
+noncomputable section
+
+def modularFormAn (n : ℕ) {N : ℕ} {k : ℤ} (f : CuspForm (Gamma0 N) k) : ℂ :=
+  (qExpansion N f).coeff ℂ n
+
+local notation:73 "a_[" n:0 "]" f:72 => modularFormAn n f
+
+/-We need to reduce a rational modulo `p`, in practice we wont be dividing by zero since the
+conductor of the elliptic curve saves us.-/
+def ratRed (q : ℚ) (p : ℕ) : ZMod p :=
+  (q.num : ZMod p) * (q.den : ZMod p)⁻¹
+
+def setOfPointsModN (E : WeierstrassCurve ℚ) [E.IsElliptic] (n : ℕ) :=
+  {P : ZMod n × ZMod n |
+    let ⟨x, y⟩ := P
+    y ^ 2 + ratRed E.a₁ n * x * y + ratRed E.a₃ n * y =
+      x ^ 3 + ratRed E.a₂ n * x ^ 2 + ratRed E.a₄ n * x + ratRed E.a₆ n}
+
+/-- The set of point `mod n` is finite.-/
+instance apFintype (E : WeierstrassCurve ℚ) [E.IsElliptic] (p : ℕ+) :
+    Fintype (setOfPointsModN E p) := by
+  rw [setOfPointsModN]
+  apply Subtype.fintype _
+
+/-Note that normally this is written as `p + 1 - #E(𝔽ₚ)`, but since we don't have a point at infinty
+on this affine curve we only have `p` -/
+def WeierstrassCurve.ap (E : WeierstrassCurve ℚ) [E.IsElliptic] (p : ℕ) : ℕ :=
+  p - Cardinal.toNat (Cardinal.mk (setOfPointsModN E p))
+
+/- Since we don't have Hecke operators yet, we define this via the q-expansion coefficients. -/
+def IsNormalisedEigenform {N : ℕ} {k : ℤ} (f : CuspForm (Gamma0 N) k) : Prop :=
+  a_[1]f = 1 ∧
+    ∀ (m n : ℕ), m.Coprime n →
+      a_[n * m]f = a_[n]f * a_[m]f ∧
+        ∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) ≠ 0 →
+          a_[p ^ r]f = a_[p]f * a_[p ^ (r - 1)]f - p ^ (k - 1) * a_[p ^ (r - 2)]f ∧
+            ∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) = 0 →
+              a_[p ^ r]f = (a_[p]f) ^ r
+
+def ModularityConjecture (E : WeierstrassCurve ℚ) [E.IsElliptic] : Prop :=
+  ∃ (N : ℕ+) (f : CuspForm (Gamma0 N) 2), IsNormalisedEigenform f ∧
+    ∀ (p : ℕ), p.Prime → (N : ZMod p) ≠ 0 → a_[p]f = E.ap p
+
+@[category research solved, AMS 11]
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/ModularityConjecture.lean
```diff
@@ -0,0 +1,85 @@
...
+/-- The set of point `mod n` is finite.-/
+instance apFintype (E : WeierstrassCurve ℚ) [E.IsElliptic] (p : ℕ+) :
+    Fintype (setOfPointsModN E p) := by
+  rw [setOfPointsModN]
+  apply Subtype.fintype _
+
+/-Note that normally this is written as `p + 1 - #E(𝔽ₚ)`, but since we don't have a point at infinty
+on this affine curve we only have `p` -/
+def WeierstrassCurve.ap (E : WeierstrassCurve ℚ) [E.IsElliptic] (p : ℕ) : ℕ :=
+  p - Cardinal.toNat (Cardinal.mk (setOfPointsModN E p))
```
seewoo5, line 69: Just for curiosity, is there a reason why this (or number of mod n points) not in mathlib yet?

CBirkbeck: I think its since we don't have conductors yet.

FormalConjectures/Wikipedia/ModularityConjecture.lean
```diff
@@ -0,0 +1,85 @@
...
+            ∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) = 0 →
+              a_[p ^ r]f = (a_[p]f) ^ r
+
+/--See [diamondshurman2005] theorem 8.8.1. -/
+def ModularityConjecture (E : WeierstrassCurve ℚ) [E.IsElliptic] : Prop :=
+  ∃ (N : ℕ+) (f : CuspForm (Gamma0 N) 2), IsNormalisedEigenform f ∧
+    ∀ (p : ℕ), p.Prime → (N : ZMod p) ≠ 0 → a_[p]f = E.ap p
+
+@[category research solved, AMS 11]
+theorem modularity_conjecture (E : WeierstrassCurve ℚ) [E.IsElliptic] : ModularityConjecture E := sorry
```
callesonne:
```suggestion
theorem modularity_conjecture (E : WeierstrassCurve ℚ) [E.IsElliptic] : ModularityConjecture E := by
  sorry
```

FormalConjectures/Wikipedia/ModularityConjecture.lean
```diff
@@ -0,0 +1,85 @@
...
+def IsNormalisedEigenform {N : ℕ} {k : ℤ} (f : CuspForm (Gamma0 N) k) : Prop :=
+  a_[1]f = 1 ∧
+    ∀ (m n : ℕ), m.Coprime n →
+      a_[n * m]f = a_[n]f * a_[m]f ∧
+        ∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) ≠ 0 →
+          a_[p ^ r]f = a_[p]f * a_[p ^ (r - 1)]f - p ^ (k - 1) * a_[p ^ (r - 2)]f ∧
+            ∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) = 0 →
+              a_[p ^ r]f = (a_[p]f) ^ r
+
+/--See [diamondshurman2005] theorem 8.8.1. -/
```
callesonne:
```suggestion
/-- See [diamondshurman2005] theorem 8.8.1. -/
```

FormalConjectures/Wikipedia/ModularityConjecture.lean
```diff
@@ -0,0 +1,85 @@
...
+      x ^ 3 + ratRed E.a₂ n * x ^ 2 + ratRed E.a₄ n * x + ratRed E.a₆ n}
+
+/-- The set of point `mod n` is finite.-/
+instance apFintype (E : WeierstrassCurve ℚ) [E.IsElliptic] (p : ℕ+) :
+    Fintype (setOfPointsModN E p) := by
+  rw [setOfPointsModN]
+  apply Subtype.fintype _
+
+/-Note that normally this is written as `p + 1 - #E(𝔽ₚ)`, but since we don't have a point at infinty
+on this affine curve we only have `p` -/
```
callesonne:
```suggestion
/-- Note that normally this is written as `p + 1 - #E(𝔽ₚ)`, but since we don't have a point at
infinty on this affine curve we only have `p` -/
```

FormalConjectures/Wikipedia/ModularityConjecture.lean
```diff
@@ -0,0 +1,85 @@
...
+def WeierstrassCurve.ap (E : WeierstrassCurve ℚ) [E.IsElliptic] (p : ℕ) : ℕ :=
+  p - Cardinal.toNat (Cardinal.mk (setOfPointsModN E p))
+
+/- Since we don't have Hecke operators yet, we define this via the q-expansion coefficients. -/
```
callesonne:
```suggestion
/-- Since we don't have Hecke operators yet, we define this via the q-expansion coefficients. -/
```

FormalConjectures/Wikipedia/ModularityConjecture.lean
```diff
@@ -0,0 +1,85 @@
...
+open scoped UpperHalfPlane Real ModularForm CongruenceSubgroup
+
+noncomputable section
+
+def modularFormAn (n : ℕ) {N : ℕ} {k : ℤ} (f : CuspForm (Gamma0 N) k) : ℂ :=
+  (qExpansion N f).coeff ℂ n
+
+local notation:73 "a_[" n:0 "]" f:72 => modularFormAn n f
+
+/-We need to reduce a rational modulo `p`, in practice we wont be dividing by zero since the
```
callesonne:
```suggestion
/-- We need to reduce a rational modulo `p`, in practice we wont be dividing by zero since the
```

FormalConjectures/Wikipedia/ModularityConjecture.lean
```diff
@@ -0,0 +1,85 @@
...
+  a_[1]f = 1 ∧
+    ∀ (m n : ℕ), m.Coprime n →
+      a_[n * m]f = a_[n]f * a_[m]f ∧
+        ∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) ≠ 0 →
+          a_[p ^ r]f = a_[p]f * a_[p ^ (r - 1)]f - p ^ (k - 1) * a_[p ^ (r - 2)]f ∧
+            ∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) = 0 →
+              a_[p ^ r]f = (a_[p]f) ^ r
+
+/--See [diamondshurman2005] theorem 8.8.1. -/
+def ModularityConjecture (E : WeierstrassCurve ℚ) [E.IsElliptic] : Prop :=
```
callesonne, line 81: In 8.8.1 they ask that the form is also a newform. Is this only to ensure that the corresponding elliptic curve also has conductor `N`?

FormalConjectures/Wikipedia/ModularityConjecture.lean
```diff
@@ -0,0 +1,85 @@
...
+
+/- Since we don't have Hecke operators yet, we define this via the q-expansion coefficients. -/
+def IsNormalisedEigenform {N : ℕ} {k : ℤ} (f : CuspForm (Gamma0 N) k) : Prop :=
+  a_[1]f = 1 ∧
+    ∀ (m n : ℕ), m.Coprime n →
+      a_[n * m]f = a_[n]f * a_[m]f ∧
+        ∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) ≠ 0 →
+          a_[p ^ r]f = a_[p]f * a_[p ^ (r - 1)]f - p ^ (k - 1) * a_[p ^ (r - 2)]f ∧
+            ∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) = 0 →
+              a_[p ^ r]f = (a_[p]f) ^ r
```
callesonne:
```suggestion
 a_[1]f = 1 ∧
  (∀ (m n : ℕ), m.Coprime n → a_[n * m]f = a_[n]f * a_[m]f) ∧
  (∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) ≠ 0 →
    a_[p ^ r]f = a_[p]f * a_[p ^ (r - 1)]f - p ^ (k - 1) * a_[p ^ (r - 2)]f) ∧
  ∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) = 0 → a_[p ^ r]f = (a_[p]f) ^ r
```
I'm not sure if it was parenthesised correctly before.

callesonne (commented):
Hi! I'm really happy to see this added :)

FormalConjectures/Wikipedia/ModularityConjecture.lean
CBirkbeck: Yeah essentially. Since we can't pin down the level then we don't need this.

CBirkbeck: Or in other words, if it wasn't new, then there exists a newform of smaller level N which has the same eigenvalues at all primes not dividing N (see 5.8.4 of loc. cit.) so you get your newform as in 8.8.1.

FormalConjectures/Wikipedia/ModularityConjecture.lean
```diff
@@ -0,0 +1,84 @@
...
+  apply Subtype.fintype _
+
+/-- Note that normally this is written as `p + 1 - #E(𝔽ₚ)`, but since we don't have a point at
+infinty on this affine curve we only have `p` -/
+def WeierstrassCurve.ap (E : WeierstrassCurve ℚ) [E.IsElliptic] (p : ℕ) : ℕ :=
+  p - Cardinal.toNat (Cardinal.mk (setOfPointsModN E p))
+
+/-- Since we don't have Hecke operators yet, we define this via the q-expansion coefficients. -/
+def IsNormalisedEigenform {N : ℕ} {k : ℤ} (f : CuspForm (Gamma0 N) k) : Prop :=
+ a_[1]f = 1 ∧
```
callesonne:
```suggestion
  a_[1]f = 1 ∧
```
my bad!

FormalConjectures/Wikipedia/ModularityConjecture.lean
```diff
@@ -0,0 +1,84 @@
...
+  (∀ (m n : ℕ), m.Coprime n → a_[n * m]f = a_[n]f * a_[m]f) ∧
+  (∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) ≠ 0 →
+    a_[p ^ r]f = a_[p]f * a_[p ^ (r - 1)]f - p ^ (k - 1) * a_[p ^ (r - 2)]f) ∧
+  ∀ (p r : ℕ), p.Prime → 2 ≤ r → (N : ZMod p) = 0 → a_[p ^ r]f = (a_[p]f) ^ r
```
callesonne, line 78: Do you have a reference for the fact that this is sufficient to determine that it's an eigenform? Is it supposed to follow from `5.3.1` of Diamond-Shurman (and if so could you add it to the docstring)? Thanks!

CBirkbeck: So yes, it follows from 5.8.5 (which maybe follows from 5.3.1, I'm not sure). But in any case I'll add the reference.

CBirkbeck:
I cant seem to remove the awaiting author label, am I doing something wrong?

callesonne:
> I cant seem to remove the awaiting author label, am I doing something wrong?

Oh right, this is probably a permissions issue, so you're not doing anything wrong :)

FormalConjectures/Wikipedia/ModularityConjecture.lean
```diff
@@ -0,0 +1,85 @@
...
+* [F. Diamond and J. Shurman, *A First Course in Modular Forms*][diamondshurman2005]
+
+-/
+
+
+open Complex CongruenceSubgroup ModularFormClass
+open scoped UpperHalfPlane Real ModularForm CongruenceSubgroup
+
+noncomputable section
+
```
callesonne, line 41: Actually, could you add comments to this and the other `def`?

FormalConjectures/Wikipedia/ModularityConjecture.lean
```diff
@@ -0,0 +1,87 @@
...
+/--Then `n`-th Fourier coefficient of a modular forms (around the cusp at infinity). -/
```
callesonne:
```suggestion
/-- The `n`-th Fourier coefficient of a modular form (around the cusp at infinity). -/
```

PR #173, opened 2025-06-06
fix: Kaplansky's conjectures

alreadydone:
Fix the statement of `counter_unit_conjecture`, which didn't include `K : Type` in the existential, so it's using the K in the variables and claiming the same field K could have any characteristic.

Expand the scope of `omit hG` so `G` doesn't get included in the later statements.

Prove implications between the counter statements.

### DIFF
--- a/FormalConjectures/Wikipedia/Kaplansky.lean
+++ b/FormalConjectures/Wikipedia/Kaplansky.lean
@@ -64,7 +64,7 @@ lemma IsTrivialUnit.isUnit {u : MonoidAlgebra K G} (h : IsTrivialUnit u) : IsUni
 /-! ## Counterexamples -/
 
 /--
-**The Promislow group** ⟨ a, b | b⁻¹a²ba², a⁻¹b²ab² ⟩
+**The Promislow group** `⟨ a, b | b⁻¹a²ba², a⁻¹b²ab² ⟩`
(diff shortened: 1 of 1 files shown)


Paul-Lez (approved):
Thanks!

FormalConjectures/Wikipedia/Kaplansky.lean
```diff
@@ -54,41 +54,17 @@ A unit in `K[G]` is trivial if it is exactly of the form `kg` where:
...
-    ∃ (G : Type) (_ : Group G) (_ : Monoid.IsTorsionFree G),
-    ∃ (_ : Field K) (_ :  CharP K p) (u : (MonoidAlgebra K G)ˣ), ¬IsTrivialUnit u.val := by
-  sorry
-
 /-! ## Counterexamples -/
 
 /--
-**The Promislow group**
-$\langle a, b | b^{-1}a^2 b a^2, a^{-1}b^2 a b^2 \rangle$
+**The Promislow group** ⟨ a, b | b⁻¹a²ba², a⁻¹b²ab² ⟩
```
Paul-Lez:
```suggestion
**The Promislow group** `⟨ a, b | b⁻¹a²ba², a⁻¹b²ab² ⟩`
```

PR #174, opened 2025-06-06
Add Gilbreath's conjecture

Reklle:


FormalConjectures/Wikipedia/Gilbreath.lean
```diff
@@ -0,0 +1,40 @@
...
+  | 0 => fun n => n.nth Prime
+  | k + 1 => let prev := GilbreathNthDerivative k
+    fun n => (prev (n + 1) - prev n).natAbs
+
+/--
+Gilbreath nth difference sequence begins with 1.
+-/
+@[category research open, AMS 11]
+theorem gilbreath_conjecture (n : ℕ+) :
+    GilbreathNthDerivative n 0 = 1 := by
```
alreadydone: Your last line apparently doesn't compile but this works (I checked the subtraction is done in `Int`):
```suggestion
noncomputable def gilbreathNthDerivative : ℕ → (ℕ → ℕ)
  | 0 => fun n ↦ n.nth Nat.Prime
  | k + 1 => let prev := gilbreathNthDerivative k
    fun n ↦ Int.natAbs (prev (n + 1) - prev n)

/--
Gilbreath nth difference sequence begins with 1.
-/
@[category research open, AMS 11]
theorem gilbreath_conjecture (n : ℕ+) :
    gilbreathNthDerivative n 0 = 1 := by
```

Reklle: Thanks!

FormalConjectures/Wikipedia/Gilbreath.lean
```diff
@@ -0,0 +1,40 @@
...
+
+/--
+**Gilbreath nth difference**, $d^n$
+Let $d^0(n) = p_n$ and $d^k(n) = |d^{k-1}(n+1) - d^{k-1}(n)|
+-/
+noncomputable def gilbreathNthDerivative : ℕ → (ℕ → ℕ)
+  | 0 => fun n ↦ n.nth Nat.Prime
+  | k + 1 => let prev := gilbreathNthDerivative k
+    fun n ↦ Int.natAbs (prev (n + 1) - prev n)
+
```
callesonne:
```suggestion
noncomputable def Gilbreath.d : ℕ → (ℕ → ℕ)
  | 0 => fun n ↦ (n+1).nth Nat.Prime - n.nth Nat.Prime
  | k + 1 => fun n ↦ ((d k (n + 1)) ^ (k + 1) - (d k n) ^ (k + 1) : ℤ).natAbs

open Gilbreath
```

FormalConjectures/Wikipedia/Gilbreath.lean
```diff
@@ -0,0 +1,40 @@
...
+/--
+Gilbreath nth difference sequence begins with 1.
+-/
+@[category research open, AMS 11]
+theorem gilbreath_conjecture (n : ℕ+) :
+    gilbreathNthDerivative n 0 = 1 := by
```
callesonne:
```suggestion
theorem gilbreath_conjecture (k : ℕ) : d k 0 = 1 := by
```
Note that in my suggested change above, the indexing would start at `0`.

FormalConjectures/Wikipedia/Gilbreath.lean
```diff
@@ -0,0 +1,40 @@
...
+**Gilbreath nth difference**, $d^n$
+Let $d^0(n) = p_n$ and $d^k(n) = |d^{k-1}(n+1) - d^{k-1}(n)|
+-/
+noncomputable def gilbreathNthDerivative : ℕ → (ℕ → ℕ)
+  | 0 => fun n ↦ n.nth Nat.Prime
+  | k + 1 => let prev := gilbreathNthDerivative k
+    fun n ↦ Int.natAbs (prev (n + 1) - prev n)
+
+/--
+Gilbreath nth difference sequence begins with 1.
```
callesonne: Could you be a bit more descriptive about the conjecture here?

FormalConjectures/Wikipedia/Gilbreath.lean
alreadydone: The rename seems good, but the superscript is an index (superscript `k` means the `k`th sequence).
Also, `d 0` according to your definition would be $d^1$ on Wikipedia. I think the original indexing is better.
```suggestion
noncomputable def Gilbreath.d : ℕ → (ℕ → ℕ)
  | 0 => fun n ↦ n.nth Nat.Prime
  | k + 1 => fun n ↦ Int.natAbs (d k (n + 1) - d k n)

open Gilbreath
```

Reklle: @alreadydone, thanks! But why is `Int.natAbs` preferable to `.natAbs` here, although `p.Prime` is preferable to `Prime p`?

alreadydone: `(prev (n + 1) - prev n).natAbs` doesn't compile for me, and it's because `prev (n + 1) - prev n` is in `Nat` but there is no `Nat.natAbs`. Does it compile for you?

Reklle: Indeed, the compiler complains about this. But `(d k (n + 1) - d k n : Int).natAbs` compiles, but I'm not sure how good the syntax is.

alreadydone: If you replace `Int` by `ℤ` then your version doesn't use more characters :) I think both version are equally valid; in my version since `Int.natAbs` takes an argument of type `Int`, there is no need of the explicit type ascription as in your version.

FormalConjectures/Wikipedia/Gilbreath.lean
```diff
@@ -0,0 +1,40 @@
...
+Let $d^0(n) = p_n$ and $d^k(n) = |d^{k-1}(n+1) - d^{k-1}(n)|
+-/
+noncomputable def Gilbreath.d : ℕ → (ℕ → ℕ)
+  | 0 => fun n ↦ n.nth Nat.Prime
+  | k + 1 => fun n ↦ Int.natAbs (d k (n + 1) - d k n)
+
+open Gilbreath
+
+/--
+For each integer $n \geq 1$ Gilbreath's nth difference sequence begins with 1, $d^n(0) = 1$.
```
callesonne:
```suggestion
**Gilbreath's conjecture**
Gilbreaths conjecture states that every term in the sequence $d^k_0$ for $k >0$ is equal to 1.
```

callesonne (approved):
Thanks, looks good to me modulo the comment I left above.

FormalConjectures/Wikipedia/Gilbreath.lean
callesonne: Thanks for fixing my mistake!

FormalConjectures/Wikipedia/Gilbreath.lean
```diff
@@ -0,0 +1,41 @@
...
+-/
+noncomputable def Gilbreath.d : ℕ → (ℕ → ℕ)
+  | 0 => fun n ↦ n.nth Nat.Prime
+  | k + 1 => fun n ↦ Int.natAbs (d k (n + 1) - d k n)
+
+open Gilbreath
+
+/--
+**Gilbreath's conjecture**
+Gilbreaths conjecture states that every term in the sequence $d^k_0$ for $k >0$ is equal to 1.
```
alreadydone:
```suggestion
Gilbreath's conjecture states that every term in the sequence $d^k_0$ for $k > 0$ is equal to 1.
```

alreadydone (approved):

mo271 (approved):

PR #179, opened 2025-06-06
feat: add upper density to `ForMathlib` and fix Erdos 244

### LINKED ISSUE #59: Erdos 244
The informal formulation of Erdos problem 244 doesn't quite seem to require that the density exists, only that the lower density is positive. Certainly Ding's paper on the problem (as referenced on the website) talks explicitly about the lower density. 
That said, this set seems well behaved enough for the density to exist, so perhaps this isn't a real problem.

callesonne:
This PR adds upper (and lower) density to `ForMathlib` to resolve #59. Doing this we are also able to add a variant to 298 that is more faithful to the statement in the paper solving the problem.

### DIFF
--- a/FormalConjectures/ErdosProblems/298.lean
+++ b/FormalConjectures/ErdosProblems/298.lean
@@ -21,7 +21,8 @@ import FormalConjectures.Util.ProblemImports
 
 *Reference:* [erdosproblems.com/298](https://www.erdosproblems.com/298)
 -/
-/--Does every set `A ⊆ N` of positive density contain some finite `S ⊂ A` such that `∑ n ∈ S, 1 / n = 1`?
+/-- Does every set `A ⊆ N` of positive density contain some finite `S ⊂ A` such that
+`∑ n ∈ S, 1 / n = 1`?
 
 The answer is yes, proved by Bloom [Bl21].
 
@@ -31,5 +32,13 @@ Note: The solution to this problem has been formalized in Lean 3 by T. Bloom and
 https://github.com/b-mehta/unit-fractions -/
 @[category research solved, AMS 11]
 theorem erdos_298 : (∀ (A : Set ℕ), 0 ∉ A → A.HasPosDensity →
+    ∃ (S : Finset ℕ), S.toSet ⊆ A ∧ ∑ n ∈ S, (1 / n : ℚ) = 1) ↔ answer(True) := by
+  sorry
+
+/--
+In [Bl21] it is proved under the weaker assumption that `A` onlu has positive upper density.
+-/
+@[category research solved, AMS 11]
+theorem erdos_298.variants.upper_density : (∀ (A : Set ℕ), 0 < A.UpperDensity →
     ∃ (S : Finset ℕ), S.toSet ⊆ A ∧ ∑ n ∈ S, (1 / n : ℚ) = 1) ↔ answer(True) := by
   sorry
--- a/FormalConjectures/ForMathlib/Data/Set/Density.lean
+++ b/FormalConjectures/ForMathlib/Data/Set/Density.lean
@@ -29,6 +29,8 @@ namespace Set
 Given a set `S` and an element `b` in an order `β`, where all intervals bounded above are finite,
 we define the partial density of `S` (relative to a set `A`) to be the proportion of elements in
 `{x ∈ A | x < b}` that lie in `S ∩ A`.
+
+This definition was inspired from https://github.com/b-mehta/unit-fractions
 -/
 noncomputable abbrev PartialDensity {β : Type*} [Preorder β] [LocallyFiniteOrderBot β]
     (S : Set β) (A : Set β := Set.univ) (b : β) : ℝ :=
--- a/FormalConjectures/ForMathlib/Data/Set/Density.lean
+++ b/FormalConjectures/ForMathlib/Data/Set/Density.lean
@@ -26,6 +26,33 @@ open scoped Topology
 namespace Set
 
 /--
+Given a set `S` and an element `b` in an order `β`, where all intervals bounded above are finite,
+we define the partial density of `S` (relative to a set `A`) to be the proportion of elements in
+`{x ∈ A | x < b}` that lie in `S ∩ A`.
+-/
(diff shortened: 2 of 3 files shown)


FormalConjectures/ErdosProblems/298.lean
```diff
@@ -33,3 +34,11 @@ https://github.com/b-mehta/unit-fractions -/
 theorem erdos_298 : (∀ (A : Set ℕ), 0 ∉ A → A.HasPosDensity →
     ∃ (S : Finset ℕ), S.toSet ⊆ A ∧ ∑ n ∈ S, (1 / n : ℚ) = 1) ↔ answer(True) := by
   sorry
+
+/--
+In [Bl21] it is proved under the weaker assumption that `A` onlu has positive upper density.
```
b-mehta:
```suggestion
In [Bl21] it is proved under the weaker assumption that `A` only has positive upper density.
```

b-mehta: Note also that the version we formalised was with positive upper density (and so upper density is in that repo)

FormalConjectures/ForMathlib/Data/Set/Density.lean
```diff
@@ -25,6 +25,35 @@ open scoped Topology
...
 namespace Set
 
+/--
+Given a set `S` and an element `b` in an order `β`, where all intervals bounded above are finite,
+we define the partial density of `S` (relative to a set `A`) to be the proportion of elements in
+`{x ∈ A | x < b}` that lie in `S ∩ A`.
+
+This definition was inspired from https://github.com/b-mehta/unit-fractions
+-/
+noncomputable abbrev PartialDensity {β : Type*} [Preorder β] [LocallyFiniteOrderBot β]
```
b-mehta:
```suggestion
noncomputable abbrev partialDensity {β : Type*} [Preorder β] [LocallyFiniteOrderBot β]
```

FormalConjectures/ForMathlib/Data/Set/Density.lean
```diff
@@ -25,6 +25,35 @@ open scoped Topology
...
+    (S : Set β) (A : Set β := Set.univ) (b : β) : ℝ :=
+  (S ∩ A ∩ Set.Iio b).ncard / (A ∩ Set.Iio b).ncard
+
+/--
+Given a set `S` in an order `β`, where all intervals bounded above are finite, we define the upper
+density of `S` (relative to a set `A`) to be the limsup of the partial densities of `S`
+(relative to `A`) for `b → ∞`.
+-/
+noncomputable def UpperDensity {β : Type*} [Preorder β] [LocallyFiniteOrderBot β]
```
b-mehta:
```suggestion
noncomputable def upperDensity {β : Type*} [Preorder β] [LocallyFiniteOrderBot β]
```

FormalConjectures/ForMathlib/Data/Set/Density.lean
```diff
@@ -25,6 +25,35 @@ open scoped Topology
...
+    (S : Set β) (A : Set β := Set.univ) : ℝ :=
+  atTop.limsup (fun (b : β) ↦ S.PartialDensity A b)
+
+/--
+Given a set `S` in an order `β`, where all intervals bounded above are finite, we define the upper
+density of `S` (relative to a set `A`) to be the liminf of the partial densities of `S`
+(relative to `A`) for `b → ∞`.
+-/
+noncomputable def LowerDensity {β : Type*} [Preorder β] [LocallyFiniteOrderBot β]
```
b-mehta:
```suggestion
noncomputable def lowerDensity {β : Type*} [Preorder β] [LocallyFiniteOrderBot β]
```

FormalConjectures/ForMathlib/Data/Set/Density.lean
```diff
@@ -25,6 +25,35 @@ open scoped Topology
...
+    (S : Set β) (A : Set β := Set.univ) : ℝ :=
+  atTop.liminf (fun (b : β) ↦ S.PartialDensity A b)
```
b-mehta: It would be nice to (eventually) have lemmas lowerDensity <= upperDensity, and lowerDensity = upperDensity implies HasDensity, but that can be todo for now :)

b-mehta (approved):

FormalConjectures/ForMathlib/Data/Set/Density.lean
callesonne: oh right... thank you! :)

FormalConjectures/ErdosProblems/298.lean
callesonne: Yes the definition `partialDensity` is inspired from there (if you didn't see already)!

b-mehta: I wrote the message about 10 seconds before I learnt that :)

FormalConjectures/ForMathlib/Data/Set/Density.lean
callesonne: Yeah those should definitely get added before upstreaming these definitions to Mathlib

FormalConjectures/ForMathlib/Data/Set/Density.lean
```diff
@@ -25,6 +25,35 @@ open scoped Topology
...
+(relative to `A`) for `b → ∞`.
+-/
+noncomputable def upperDensity {β : Type*} [Preorder β] [LocallyFiniteOrderBot β]
+    (S : Set β) (A : Set β := Set.univ) : ℝ :=
+  atTop.limsup (fun (b : β) ↦ S.partialDensity A b)
+
+/--
+Given a set `S` in an order `β`, where all intervals bounded above are finite, we define the upper
+density of `S` (relative to a set `A`) to be the liminf of the partial densities of `S`
+(relative to `A`) for `b → ∞`.
```
mo271:
```suggestion
Given a set `S` in an order `β`, where all intervals bounded above are finite, we define the lower
density of `S` (relative to a set `A`) to be the liminf of the partial densities of `S`
(relative to `A`) for `b → ∞`.
```

callesonne: Thanks!

Issue #151, opened 2025-06-03
[Mini-Roadmap] Irrationality


Reklle:
## Goal
Сomplete the list of open questions and conjectures on irrational and transcendental numbers.

## Plan
- [ ] Update `FormalConjectures/Wikipedia/Irrational.lean`
- [ ] Make a list for open questions on transcendence
- [ ] Add  [Schanuel's conjecture](https://en.wikipedia.org/wiki/Schanuel%27s_conjecture)
- [ ] Add  [Littlewood conjecture](https://en.wikipedia.org/wiki/Littlewood_conjecture)
- [ ] Add family of [exponentials conjectures](https://en.wikipedia.org/wiki/Littlewood_conjecture)

I will be glad to add more related conjectures to the list!

## Suggestions
- Allocate a separate folder for this topic `FormalConjectures/Irrational`
- Merge `Irrational.lean` and `RiemannZeta.lean`

Paul-Lez:
Hey @Reklle, thanks for adding this! I've marked Schanuel's conjecture and Littlewood's conjecture as completed since they're already in the repo. I notice the link for the family of exponential conjectures task points to the Littlewood conjecture page - is that intentional?

Reklle:
@Paul-Lez, thanks! Fixed the third link, it's was a typo.

alreadydone:
I think Littlewood's shouldn't be moved in as it doesn't fit the theme. It may be included in a folder on Diophantine approximation.

Paul-Lez:
Re file moves, the repository is currently structured by source rather than topic, and the proposed move would break that, which isn't ideal;). 

That being said, I think the suggestion of having a unified way to find all these related questions is a great idea and would be really helpful as a way of documenting what's currently stated in the repo! 
As an alternative in this direction, it could be really nice to have a place that references the various irrationality-related questions across the repository if you feel like doing that. I'll have a think about the best way to go about this and get back to you:)

Concerning merging `Irrational.lean` and `RiemannZeta.lean`, I would hold off doing that since the problems in those files are coming from different Wiki pages (the idea is to try if possible to stick to one page per file unless there are good reasons not to)

Reklle:
Do you think it's worth splitting [SchanuelsConjecture.lean](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/Wikipedia/SchanuelsConjecture.lean) into parts and extend them:
- `Schanuel.lean`
- `Transcendental.lean`
[...]

PR #185, opened 2025-06-08
Fix: AMS codes

Reklle:
While working on #151, I noticed that not all theorems had AMS code. Fixed.
`SchanuelsConjecture.lean` will be fixed along with the rest of its changes.


Paul-Lez (approved):
Thanks!

PR #211, opened 2025-06-11
feat: Copyright linter

Paul-Lez:
The goal of this linter is to give the user a warning locally rather than having the CI throw an error.


FormalConjectures/Util/Linters/CopyrightLinterTestInvalid.lean
mo271, line 1: Not sure if/how we want to get this file past our check in the ci. Perhaps best to check manually and then delete it?

Paul-Lez: Sounds reasonable! The erroneous test file does throw an error as expected locally so I'll delete it now.

Paul-Lez:
Closing this PR and opening a new one since I won't be able to merge it because GitHub Copilot hasn't signed a CLA

Issue #196, opened 2025-06-10
Erdős Problem 17: Infinitely primes, written as certain different of primes?


mo271:
### What is the conjecture
    
https://www.erdosproblems.com/17

  Are there infinitely many primes $p$ such that every even number $n\leq p-3$ can be written as a difference of primes $n=q_1-q_2$ where $q_1,q_2\leq p$?

The first prime without this property is 97. The sequence of such primes is [A038133](https://oeis.org/A038133) . These are called cluster primes.

Status: open

### Prerequisites needed
None.

priyankaranke:
Hi, I'd like to try adding this conjecture. Thank you!

PR #212, opened 2025-06-11
Formalize Erdos 17

priyankaranke:
Please see issue #196

### DIFF
--- a/FormalConjectures/ErdosProblems/17.lean
+++ b/FormalConjectures/ErdosProblems/17.lean
@@ -21,24 +21,24 @@ import FormalConjectures.Util.ProblemImports
 *Reference:* [erdosproblems.com/17](https://www.erdosproblems.com/17)
 -/
 
-/-- A prime $p$ is a cluster prime if every even integer
+/-- A prime $p$ is a cluster prime if every even natural number
 $n \le p - 3$ can be written as a difference of two primes
 $q_1 - q_2$ with $q_1, q_2 \le p$. -/
-def ClusterPrime (p : ℕ) : Prop :=
+def IsClusterPrime (p : ℕ) : Prop :=
   p.Prime ∧
-    ∀ {n : ℕ}, Even n → n ≤ p - 3 →
+    ∀ {n : ℕ}, Even n → n ≤ (p - 3 : ℤ) →
       ∃ q₁ q₂ : ℕ, q₁.Prime ∧ q₂.Prime ∧
-        q₁ ≤ p ∧ q₂ ≤ p ∧ n = q₁ - q₂
+        q₁ ≤ p ∧ q₂ ≤ p ∧ n = (q₁ - q₂ : ℤ)
 
 /-- **Erdős Problem 17.** Are there infinitely many cluster primes? -/
 @[category research open, AMS 11]
 theorem erdos_17 :
-    {p : ℕ | ClusterPrime p}.Infinite ↔ answer(sorry) := by
+    {p : ℕ | IsClusterPrime p}.Infinite ↔ answer(sorry) := by
   sorry
 
 /-- The counting function of cluster primes $\le x$. -/
 noncomputable def clusterPrimeCount (x : ℕ) : ℕ :=
-  Nat.card {p : ℕ | p ≤ x ∧ ClusterPrime p}
+  Nat.card {p : ℕ | p ≤ x ∧ IsClusterPrime p}
 
 /--
 In 1999 Blecksmith, Erdős, and Selfridge [BES99] proved the upper bound
@@ -63,5 +63,5 @@ theorem erdos_17.variants.upper_Elsholtz {c : ℝ} (hc : c < (1 / 8 : ℝ)) :
 
 /-- $97$ is the smallest prime that is not a cluster prime. -/
 @[category test, AMS 11]
-example : IsLeast {p : ℕ | p.Prime ∧ ¬ ClusterPrime p} 97 := by
+example : IsLeast {p : ℕ | p.Prime ∧ ¬ IsClusterPrime p} 97 := by
   sorry
--- a/FormalConjectures/ErdosProblems/17.lean
+++ b/FormalConjectures/ErdosProblems/17.lean
@@ -21,6 +21,10 @@ import FormalConjectures.Util.ProblemImports
 *Reference:* [erdosproblems.com/17](https://www.erdosproblems.com/17)
 -/
 
+open Filter
+
+open scoped Asymptotics
+
 /-- A prime $p$ is a cluster prime if every even natural number
 $n \le p - 3$ can be written as a difference of two primes
 $q_1 - q_2$ with $q_1, q_2 \le p$. -/
@@ -43,22 +47,28 @@ noncomputable def clusterPrimeCount (x : ℕ) : ℕ :=
 /--
 In 1999 Blecksmith, Erdős, and Selfridge [BES99] proved the upper bound
 $$\pi^{\mathcal{C}}(x) \ll_A x(\log x)^{-A}$$ for every real $A > 0$.
+
+[BES99] Blecksmith, Richard and Erd\H os, Paul and Selfridge, J. L., Cluster primes. Amer. Math. Monthly (1999), 43--48.
 -/
 @[category research solved, AMS 11]
 theorem erdos_17.variants.upper_BES {A : ℝ} (hA : 0 < A) :
-  ∃ C : ℝ, 0 < C ∧
-    ∀ x : ℕ, (clusterPrimeCount x : ℝ) ≤ C * x / (Real.log x) ^ A := by
-  sorry
+  (fun x ↦ (clusterPrimeCount x : ℝ)) =O[atTop] fun x ↦ x / (Real.log x) ^ A := by
+    sorry
 
 /--
 In 2003, Elsholtz [El03] refined the upper bound to
 $$\pi^{\mathcal{C}}(x) \ll x\,\exp\!\bigl(-c(\log\log x)^2\bigr)$$
-for every real $0 < c < 1/8$. -/
+for every real $0 < c < 1/8$.
+
+[El03] Elsholtz, Christian, On cluster primes. Acta Arith. (2003), 281--284.
+-/
 @[category research solved, AMS 11]
-theorem erdos_17.variants.upper_Elsholtz {c : ℝ} (hc : 0 < c ∧ c < (1/8 : ℝ)) :
+theorem erdos_17.variants.upper_Elsholtz :
   ∃ C : ℝ, 0 < C ∧
-    ∀ x : ℕ, (clusterPrimeCount x : ℝ) ≤
-      C * x * Real.exp (-c * (Real.log (Real.log x)) ^ 2) := by
+    ∀ c : ℝ, 0 < c → c < (1 / 8 : ℝ) →
+      ∀ᶠ x : ℕ in atTop,
+        (clusterPrimeCount x : ℝ) ≤
+          C * x * Real.exp (-c * (Real.log (Real.log x)) ^ 2) := by
   sorry
 
 /-- $97$ is the smallest prime that is not a cluster prime. -/
--- a/FormalConjectures/ErdosProblems/17.lean
+++ b/FormalConjectures/ErdosProblems/17.lean
@@ -46,8 +46,8 @@ $$\pi^{\mathcal{C}}(x) \ll_A x(\log x)^{-A}$$ for every real $A > 0$.
 -/
 @[category research solved, AMS 11]
 theorem erdos_17.variants.upper_BES {A : ℝ} (hA : 0 < A) :
-    ∃ C : ℝ, 0 < C ∧
-      ∀ x : ℕ, (clusterPrimeCount x : ℝ) ≤ C * x / (Real.log x) ^ A := by
+  ∃ C : ℝ, 0 < C ∧
+    ∀ x : ℕ, (clusterPrimeCount x : ℝ) ≤ C * x / (Real.log x) ^ A := by
   sorry
 
 /--
@@ -55,10 +55,10 @@ In 2003, Elsholtz [El03] refined the upper bound to
 $$\pi^{\mathcal{C}}(x) \ll x\,\exp\!\bigl(-c(\log\log x)^2\bigr)$$
 for every real $0 < c < 1/8$. -/
 @[category research solved, AMS 11]
-theorem erdos_17.variants.upper_Elsholtz {c : ℝ} (hc : c < (1 / 8 : ℝ)) :
-    ∃ C : ℝ, 0 < C ∧
-      ∀ x : ℕ, (clusterPrimeCount x : ℝ) ≤
-        C * x * Real.exp (-c * (Real.log (Real.log x)) ^ 2) := by
+theorem erdos_17.variants.upper_Elsholtz {c : ℝ} (hc : 0 < c ∧ c < (1/8 : ℝ)) :
+  ∃ C : ℝ, 0 < C ∧
+    ∀ x : ℕ, (clusterPrimeCount x : ℝ) ≤
+      C * x * Real.exp (-c * (Real.log (Real.log x)) ^ 2) := by
   sorry
 
 /-- $97$ is the smallest prime that is not a cluster prime. -/
--- a/FormalConjectures/ErdosProblems/17.lean
+++ b/FormalConjectures/ErdosProblems/17.lean
@@ -1,0 +1,45 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,45 @@
...
+@[category research open, AMS 11]
+theorem erdos_17 :
+    ({p : ℕ | Erdos17GoodPrime p}.Infinite) ↔ answer(sorry) := by
+  sorry
+
+/--
+The first prime **without** the Erdős 17 property is $97$.
+-/
+@[category test, AMS 11]
+example : Prime 97 ∧ ¬ Erdos17GoodPrime 97 ∧
```
Paul-Lez: This works but I'd suggest formalising this statement using `IsLeast` (since that's a more principled approach to formalising this type of claim)

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,45 @@
...
+  Prime p ∧
+    ∀ {n : ℕ}, Even n → n ≤ p - 3 →
+      ∃ q₁ q₂ : ℕ, Prime q₁ ∧ Prime q₂ ∧
+        q₁ ≤ p ∧ q₂ ≤ p ∧ n = q₁ - q₂
+
+/-- **Erdős Problem 17.** Are there infinitely many good primes?
+(The term *good* is defined above.) -/
+@[category research open, AMS 11]
+theorem erdos_17 :
+    ({p : ℕ | Erdos17GoodPrime p}.Infinite) ↔ answer(sorry) := by
```
Paul-Lez:
```suggestion
    {p : ℕ | Erdos17GoodPrime p}.Infinite ↔ answer(sorry) := by
```

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,45 @@
...
+/-!
+# Erdős Problem 17
+*Reference:* [erdosproblems.com/17](https://www.erdosproblems.com/17)
+-/
+
+/-- A prime $p$ is **good** (Erdős 17) if every even integer
+$n \le p - 3$ can be written as a difference of two primes
+$q_1 - q_2$ with $q_1, q_2 \le p$. -/
+def Erdos17GoodPrime (p : ℕ) : Prop :=
+  Prime p ∧
```
Paul-Lez: (somewhat a matter of personal preference): I think it's little nicer to use `Nat.Prime` rather than `Prime`
```suggestion
  p.Prime ∧
```

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,45 @@
...
+      ∃ q₁ q₂ : ℕ, Prime q₁ ∧ Prime q₂ ∧
+        q₁ ≤ p ∧ q₂ ≤ p ∧ n = q₁ - q₂
+
+/-- **Erdős Problem 17.** Are there infinitely many good primes?
+(The term *good* is defined above.) -/
+@[category research open, AMS 11]
+theorem erdos_17 :
+    ({p : ℕ | Erdos17GoodPrime p}.Infinite) ↔ answer(sorry) := by
+  sorry
+
```
Paul-Lez, line 40: The Erdős Problems webpage mentions two related asymptotic results - would you be up for adding these too?

FormalConjectures/ErdosProblems/17.lean
priyankaranke: Thank you! Yeah, absolutely, I'm new to Lean so definitely looking to learn more about what the most precise/principled approach is.

FormalConjectures/ErdosProblems/17.lean
mo271: (of course you could also leave this as TODO)

priyankaranke: Thank you both! I've attempted to formalize the two related results, please let me know if there's a better way to represent asymptotic bounds and the counting that I may have missed!

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,67 @@
...
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 17
+*Reference:* [erdosproblems.com/17](https://www.erdosproblems.com/17)
+-/
+
+/-- A prime $p$ is a cluster prime if every even integer
```
callesonne:
```suggestion
/-- A prime $p$ is a cluster prime if every even natural number
```

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,67 @@
...
+$n \le p - 3$ can be written as a difference of two primes
+$q_1 - q_2$ with $q_1, q_2 \le p$. -/
+def ClusterPrime (p : ℕ) : Prop :=
+  p.Prime ∧
+    ∀ {n : ℕ}, Even n → n ≤ p - 3 →
+      ∃ q₁ q₂ : ℕ, q₁.Prime ∧ q₂.Prime ∧
+        q₁ ≤ p ∧ q₂ ≤ p ∧ n = q₁ - q₂
```
callesonne:
```suggestion
    ∀ {n : ℕ}, Even n → n ≤ (p - 3 : ℤ) →
      ∃ q₁ q₂ : ℕ, q₁.Prime ∧ q₂.Prime ∧
        q₁ ≤ p ∧ q₂ ≤ p ∧ n = (q₁ - q₂ : ℤ)
```
Here we need to be careful, since by default these subtractions will be natural number subtractions. I don't think there is an issue with the cases when truncation happens here, but I think its still better to avoid, so that it is more faithful to the informal statement.

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,67 @@
...
+
+/-!
+# Erdős Problem 17
+*Reference:* [erdosproblems.com/17](https://www.erdosproblems.com/17)
+-/
+
+/-- A prime $p$ is a cluster prime if every even integer
+$n \le p - 3$ can be written as a difference of two primes
+$q_1 - q_2$ with $q_1, q_2 \le p$. -/
+def ClusterPrime (p : ℕ) : Prop :=
```
callesonne:
```suggestion
def IsClusterPrime (p : ℕ) : Prop :=
```
I would call this `IsClusterPrime` since it is a `Prop`

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,67 @@
...
+@[category research solved, AMS 11]
+theorem erdos_17.variants.upper_Elsholtz {c : ℝ} (hc : c < (1 / 8 : ℝ)) :
+    ∃ C : ℝ, 0 < C ∧
+      ∀ x : ℕ, (clusterPrimeCount x : ℝ) ≤
+        C * x * Real.exp (-c * (Real.log (Real.log x)) ^ 2) := by
+  sorry
+
+/-- $97$ is the smallest prime that is not a cluster prime. -/
+@[category test, AMS 11]
+example : IsLeast {p : ℕ | p.Prime ∧ ¬ ClusterPrime p} 97 := by
```
callesonne:
```suggestion
example : IsLeast {p : ℕ | p.Prime ∧ ¬ IsClusterPrime p} 97 := by
```

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,67 @@
...
+def ClusterPrime (p : ℕ) : Prop :=
+  p.Prime ∧
+    ∀ {n : ℕ}, Even n → n ≤ p - 3 →
+      ∃ q₁ q₂ : ℕ, q₁.Prime ∧ q₂.Prime ∧
+        q₁ ≤ p ∧ q₂ ≤ p ∧ n = q₁ - q₂
+
+/-- **Erdős Problem 17.** Are there infinitely many cluster primes? -/
+@[category research open, AMS 11]
+theorem erdos_17 :
+    {p : ℕ | ClusterPrime p}.Infinite ↔ answer(sorry) := by
```
callesonne:
```suggestion
    {p : ℕ | IsClusterPrime p}.Infinite ↔ answer(sorry) := by
```

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,67 @@
...
+  sorry
+
+/-- The counting function of cluster primes $\le x$. -/
+noncomputable def clusterPrimeCount (x : ℕ) : ℕ :=
+  Nat.card {p : ℕ | p ≤ x ∧ ClusterPrime p}
```
callesonne:
```suggestion
noncomputable def clusterPrimeCount (x : ℕ) : ℕ :=
  Nat.card {p : ℕ | p ≤ x ∧ IsClusterPrime p}
```

callesonne (commented):
Thanks for the constribution! Here are some partial comments

FormalConjectures/ErdosProblems/17.lean
priyankaranke: That makes sense to me, fixed!

FormalConjectures/ErdosProblems/17.lean
priyankaranke: Fixed, thank you!

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,67 @@
...
+  sorry
+
+/-- The counting function of cluster primes $\le x$. -/
+noncomputable def clusterPrimeCount (x : ℕ) : ℕ :=
+  Nat.card {p : ℕ | p ≤ x ∧ IsClusterPrime p}
+
+/--
+In 1999 Blecksmith, Erdős, and Selfridge [BES99] proved the upper bound
+$$\pi^{\mathcal{C}}(x) \ll_A x(\log x)^{-A}$$ for every real $A > 0$.
+-/
```
callesonne:
```suggestion
$$\pi^{\mathcal{C}}(x) \ll_A x(\log x)^{-A}$$ for every real $A > 0$.

[BES99] Blecksmith, Richard and Erd\H os, Paul and Selfridge, J. L., Cluster primes. Amer. Math. Monthly (1999), 43--48.
-/
```

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,67 @@
...
+
+/--
+In 2003, Elsholtz [El03] refined the upper bound to
+$$\pi^{\mathcal{C}}(x) \ll x\,\exp\!\bigl(-c(\log\log x)^2\bigr)$$
+for every real $0 < c < 1/8$. -/
+@[category research solved, AMS 11]
+theorem erdos_17.variants.upper_Elsholtz {c : ℝ} (hc : 0 < c ∧ c < (1/8 : ℝ)) :
+  ∃ C : ℝ, 0 < C ∧
+    ∀ x : ℕ, (clusterPrimeCount x : ℝ) ≤
+      C * x * Real.exp (-c * (Real.log (Real.log x)) ^ 2) := by
```
callesonne: Here I think `C` should not depend on `c`. So the statement should only quantify on `c` after `C` has been introduced.

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,67 @@
...
+theorem erdos_17.variants.upper_BES {A : ℝ} (hA : 0 < A) :
+  ∃ C : ℝ, 0 < C ∧
+    ∀ x : ℕ, (clusterPrimeCount x : ℝ) ≤ C * x / (Real.log x) ^ A := by
+  sorry
+
+/--
+In 2003, Elsholtz [El03] refined the upper bound to
+$$\pi^{\mathcal{C}}(x) \ll x\,\exp\!\bigl(-c(\log\log x)^2\bigr)$$
+for every real $0 < c < 1/8$. -/
+@[category research solved, AMS 11]
```
callesonne:
```suggestion
for every real $0 < c < 1/8$.

[El03] Elsholtz, Christian, On cluster primes. Acta Arith. (2003), 281--284.
-/
@[category research solved, AMS 11]
```

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,67 @@
...
+  Nat.card {p : ℕ | p ≤ x ∧ IsClusterPrime p}
+
+/--
+In 1999 Blecksmith, Erdős, and Selfridge [BES99] proved the upper bound
+$$\pi^{\mathcal{C}}(x) \ll_A x(\log x)^{-A}$$ for every real $A > 0$.
+-/
+@[category research solved, AMS 11]
+theorem erdos_17.variants.upper_BES {A : ℝ} (hA : 0 < A) :
+  ∃ C : ℝ, 0 < C ∧
+    ∀ x : ℕ, (clusterPrimeCount x : ℝ) ≤ C * x / (Real.log x) ^ A := by
```
callesonne:
```suggestion
theorem erdos_17.variants.upper_BES {A : ℝ} (hA : 0 < A) :
    (fun x ↦ (clusterPrimeCount x : ℝ)) =O[atTop] fun x ↦ x / (Real.log x) ^ A := by
  sorry
```
See main review comment.

callesonne (commented):
Some final comments, once these are addressed it's ready to go!

About the asymptotics, they are supposed to hold only for sufficiently large `x`. The general way of doing this is to use [Filter.Eventually](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/Filter/Defs.html#Filter.Eventually) (notation `∀ᶠ`). In your situation however you can use `IsBigO` to make it even cleaner, I've left a suggestion for how to do it for `erdos_17.variants.upper_BES`

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,67 @@
...
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 17
+*Reference:* [erdosproblems.com/17](https://www.erdosproblems.com/17)
+-/
+
```
callesonne, line 23:
```suggestion
-/

open Filter

open scoped Asymptotics
```
For the suggestion about using `IsBigO` (and corresponding notation) we need to open this.

FormalConjectures/ErdosProblems/17.lean
priyankaranke: This is much nicer notation, thank you so much! I've fixed this for `erdos_17.variants.upper_BES` as you suggested.

For `erdos_17.variants.upper_Elsholtz`, I'm a bit concerned that using `=O[atTop]` would still allow the implicit variable in `IsBigO` to vary with `c`, which we don't want to allow ([comment](https://github.com/google-deepmind/formal-conjectures/pull/212#discussion_r2148818228)). 

For that reason, I kept the big-O formulation explicit, but added `∀ᶠ` to make the statement hold only for sufficiently large x.

Please let me know if that statement needs more work, thanks again for your help!

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,77 @@
...
+[El03] Elsholtz, Christian, On cluster primes. Acta Arith. (2003), 281--284.
+-/
+@[category research solved, AMS 11]
+theorem erdos_17.variants.upper_Elsholtz :
+  ∃ C : ℝ, 0 < C ∧
+    ∀ c : ℝ, (0 < c ∧ c < (1 / 8 : ℝ)) →
+      ∀ᶠ x : ℕ in atTop,
+        (clusterPrimeCount x : ℝ) ≤
+          C * x * Real.exp (-c * (Real.log (Real.log x)) ^ 2) := by
```
callesonne:
```suggestion
theorem erdos_17.variants.upper_Elsholtz :
  ∃ C : ℝ, 0 < C ∧
    ∀ c ∈ Set.Ioo 0 (1 / 8),
      IsBigOWith C atTop (fun x ↦ (clusterPrimeCount x : ℝ))
        (fun x ↦ x * exp (-c * (log (log x)) ^ 2)) := by
  sorry
```

FormalConjectures/ErdosProblems/17.lean
callesonne: There is `IsBigOWith`, which I think solves this! Downside is that there is no nice notation for it, although Mathlib should probably add that at some point :)

FormalConjectures/ErdosProblems/17.lean
```diff
@@ -0,0 +1,77 @@
...
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 17
+*Reference:* [erdosproblems.com/17](https://www.erdosproblems.com/17)
+-/
+
+open Filter
+
+open scoped Asymptotics
```
callesonne:
```suggestion
open Filter Asymptotics Real
```
Right, you also need open Asymptotics to implement my comment! And I would open Real, so you can get rid of the `Real.exp` and `Real.log` to make everything a bit shorter :)

FormalConjectures/ErdosProblems/17.lean
priyankaranke: Thanks for the `IsBigOWith` suggestion, fixed!

priyankaranke:
👋 Hey folks, could you let me know what the next steps are here? I don't have merge access to `google-deepmind:main`. Also happy to make other changes to this PR as needed. Thanks again for all your help.

mo271 (approved):

mo271:
Thanks for the contribution!

PR #214, opened 2025-06-12
feat: Implement linter to check the copyright header of files

Paul-Lez:


FormalConjectures/Util/Linters/CopyrightLinterTest.lean
mo271, line 1: not sure we need this file at all.

FormalConjectures/Util/Linters/CopyrightLinter.lean
```diff
@@ -0,0 +1,54 @@
...
+limitations under the License.
+-/
+
+import Mathlib.Tactic.Linter.Header
+
+open Lean Elab Meta Command Syntax
+
+def correctCopyrightHeader : String :=
+"/-
+Copyright 2025 The Formal Conjectures Authors.
```
mo271, line 25: This will break down when people start using "2026" here next year?

Either fix now, or leave a TODO (which should ideally be addressed in the next 6 months)

mo271 (approved):

FormalConjectures/Util/Linters/CopyrightLinter.lean
Paul-Lez: Good point! Added a note

FormalConjectures/Util/Linters/CopyrightLinterTest.lean
Paul-Lez: I've removed it.

PR #215, opened 2025-06-12
Erdõs 36: The minimal overlap problem

### LINKED ISSUE #208: Erdős Problem 36: the minimum overlap problem
### What is the conjecture
    
https://www.erdosproblems.com/36

  Find the optimal constant $c>0$ such that the following holds. For all sufficiently large $N$, if $A\sqcup B=\{1,\ldots,2N\}$ is a partition into two equal parts, so that $\lvert A\rvert=\lvert B\rvert=N$, then there is some $x$ such that the number of solutions to $a-b=x$ with $a\in A$ and $b\in B$ is at least $cN$.

https://en.wikipedia.org/wiki/Minimum_overlap_problem

Status: open

### Prerequisites needed
None

### Choose either option

mo271:
closes #208


FormalConjectures/Wikipedia/MinimalOverlapProblem.lean
mo271, line 1: Not sure if we really want this file, but it would be an example of a conjecture that is part of two problem sets: Erdős and Wikipedia. Making this dummy file might it make finding the problem easier?

FormalConjectures/ErdosProblems/36.lean
```diff
@@ -0,0 +1,178 @@
...
+@[category research solved, AMS 11]
+theorem minimum_overlap.variants.lower.white_2022 : 0.379005 < atTop.liminf MinOverlapQuotient := by sorry
+
+
+/--
+An upper bound of $\frac 1 2$.
+See [Some remarks on number theory (in Hebrew)](https://users.renyi.hu/~p_erdos/1955-13.pdf) by *Paul Erdős*, Riveon Lematematika 9, p.45-48,1955
+-/
+@[category research solved, AMS 11]
+theorem minimum_overlap.variants.upper.erdos_1955 : ∃ o, (o =o[atTop] fun n => (1 : ℝ)) ∧ atTop.limsup (fun N => MinOverlapQuotient N - o N) < (1 : ℝ) / 2 := by sorry
```
mo271: Using the `o` here makes it more closely follow Wikipedia, but I feel like it might be possible to do without somehow?

FormalConjectures/ErdosProblems/36.lean
```diff
@@ -0,0 +1,178 @@
...
+*References:*
+ - [erdosproblems.com/36](https://www.erdosproblems.com/36)
+ - [Wikipedial: Minimum overlap problem](https://en.wikipedia.org/wiki/Minimum_overlap_problem)
+-/
+
+/--
+The number of solutions to the equation $a - b = k$, for $a \in A$ and $b \in B$.
+This represents the "overlap" between sets $A$ and $B$ for a given difference $k$.
+-/
+noncomputable def Overlap (A B : Finset ℤ) (k : ℤ) : ℕ := {((a : A), (b: B)) | (a : ℤ) - b = k}.ncard
```
Paul-Lez:
```suggestion
noncomputable def Overlap (A B : Finset ℤ) (k : ℤ) : ℕ := {((a : A), (b : B)) | (a : ℤ) - b = k}.ncard
```

FormalConjectures/ErdosProblems/36.lean
```diff
@@ -0,0 +1,178 @@
...
+    (_same_card : A.card = B.card)}
+
+/--
+This example calculates the value of $M 1$. The set is $\{1, 2\}$, so the only partition is
+$A = \{1\}, B = \{2\}$ (or vice versa). The possible differences are $1 - 2 = -1$ and $2 - 1 = 1$.
+The `Overlap` for $k=-1$ is 1 (if $A=\{1\}, B=\{2\}$) and for $k=1$ also 1 (if $A=\{2\}, B=\{1\}$ ).
+The `MaxOverlap` is $1$, since the `Overlap` is $0$ for other $k$.
+Thus, $M 1 = 1$.
+-/
+@[category test, AMS 11]
```
Paul-Lez: I'd also suggest the AMS 5 category for problems on this page since they have a strong additive combinatorics flavour

FormalConjectures/Wikipedia/MinimalOverlapProblem.lean
Paul-Lez: I think this is a reasonable solution!

FormalConjectures/ErdosProblems/36.lean
```diff
@@ -0,0 +1,178 @@
...
+-/
+@[category graduate, AMS 11]
+theorem minimum_overlap.variants.lower.erdos_1955 : (1 : ℝ) / 4 < atTop.liminf MinOverlapQuotient := by sorry
+
+/--
+A lower bound of $1 - frac{1}{\sqrt 2}$.
+Scherk (written communication), see [On the minimal overlap problem of Erdös](https://eudml.org/doc/206397) by *Leo Moser*, Аста Аrithmetica V, p. 117-119, 1959
+-/
+@[category research solved, AMS 11]
+theorem minimum_overlap.variants.lower.scherk_1955 : 1 - 2 ^ (-(1 : ℝ) / 2) < atTop.liminf MinOverlapQuotient := by sorry
```
Paul-Lez: I'd be tempted to write this as follows for readability (and same for some of the other bounds below)
```suggestion
theorem minimum_overlap.variants.lower.scherk_1955 : 1 - (√2)⁻¹ < atTop.liminf MinOverlapQuotient := by sorry
```

Paul-Lez (commented):
A few initial comments - will take another look later!

FormalConjectures/ErdosProblems/36.lean
```diff
@@ -0,0 +1,178 @@
...
+
+/--
+This example calculates the value of $M 1$. The set is $\{1, 2\}$, so the only partition is
+$A = \{1\}, B = \{2\}$ (or vice versa). The possible differences are $1 - 2 = -1$ and $2 - 1 = 1$.
+The `Overlap` for $k=-1$ is 1 (if $A=\{1\}, B=\{2\}$) and for $k=1$ also 1 (if $A=\{2\}, B=\{1\}$ ).
+The `MaxOverlap` is $1$, since the `Overlap` is $0$ for other $k$.
+Thus, $M 1 = 1$.
+-/
+@[category test, AMS 11]
+example : M 1 = 1 := by
```
mo271, line 61: https://gist.github.com/mo271/7aa479dbb3bb9b5a946f4e9da52cbd1d

I wrote a an ungolfed proof of this, but I think it is too long to be included (even when golfed to 20% of the size)

FormalConjectures/ErdosProblems/36.lean
mo271: yeah, not sure why wikipedia uses the weird exponents here, perhaps this would be good to fix at wikipedia...

FormalConjectures/ErdosProblems/36.lean
```diff
@@ -0,0 +1,178 @@
...
+*References:*
+ - [erdosproblems.com/36](https://www.erdosproblems.com/36)
+ - [Wikipedial: Minimum overlap problem](https://en.wikipedia.org/wiki/Minimum_overlap_problem)
+-/
+
+/--
+The number of solutions to the equation $a - b = k$, for $a \in A$ and $b \in B$.
+This represents the "overlap" between sets $A$ and $B$ for a given difference $k$.
+-/
+noncomputable def Overlap (A B : Finset ℤ) (k : ℤ) : ℕ := {((a : A), (b : B)) | (a : ℤ) - b = k}.ncard
```
callesonne:
```suggestion
noncomputable def Overlap (A B : Finset ℤ) (k : ℤ) : ℕ := {((a : A), (b : B)) | k = a - b}.ncard
```
this avoids having to specify the type

FormalConjectures/ErdosProblems/36.lean
```diff
@@ -0,0 +1,178 @@
...
+-/
+noncomputable def MaxOverlap (A B : Finset ℤ) : ℕ := iSup <| Overlap A B
+
+/--
+Let $A$ and $B$ be two complementary subsets, a splitting of the numbers $\{1, 2, \dots, 2n\}$,
+such that both have the same cardinality $n$.
+Define $M(n)$ to be the minimum `MaxOverlap` that can be achieved, ranging over all such partitions $(A, B)$.
+-/
+noncomputable def M (n : ℕ) : ℕ :=
+  sInf {MaxOverlap A B | (A : Finset ℤ) (B: Finset ℤ)
```
callesonne:
```suggestion
  sInf {MaxOverlap A B | (A : Finset ℤ) (B : Finset ℤ)
```

FormalConjectures/ErdosProblems/36.lean
```diff
@@ -0,0 +1,178 @@
...
+
+@[category test, AMS 11]
+example : M 5 = 3 := by
+  sorry
+
+/--
+The quotient of the minimum maximum overlap $M(N)$ by $N$. The central question of the
+minimum overlap problem is to determine the asymptotic behavior of this quotient as $N \to \infty$.
+-/
+noncomputable def MinOverlapQuotient N := (M N : ℝ) / N
```
callesonne:
```suggestion
noncomputable def MinOverlapQuotient (N : ℕ) := (M N : ℝ) / N
```
Not necessary but to me it makes it more readable

callesonne (commented):
Nice! I also think that the problem corresponding to the erdos problem should be named `erdos_36`, or alternatively add an alias for that theorem.

FormalConjectures/ErdosProblems/36.lean
mo271: done

mo271:
> Nice! I also think that the problem corresponding to the erdos problem should be named `erdos_36`, or alternatively add an alias for that theorem.

renamed to erdos_36

FormalConjectures/ErdosProblems/36.lean
```diff
@@ -0,0 +1,178 @@
...
+-/
+noncomputable def MinOverlapQuotient (N : ℕ) := (M N : ℝ) / N
+
+
+/--
+A lower bound of $\frac 1 4$.
+See [Some remarks on number theory (in Hebrew)](https://users.renyi.hu/~p_erdos/1955-13.pdf) by *Paul Erdős*, Riveon Lematematika 9, p.45-48,1955
+-/
+@[category graduate, AMS 5 11]
+theorem minimum_overlap.variants.lower.erdos_1955 : (1 : ℝ) / 4 < atTop.liminf MinOverlapQuotient := by sorry
```
callesonne:
```suggestion
theorem minimum_overlap.variants.lower.erdos_1955 :
    (1 : ℝ) / 4 < atTop.liminf MinOverlapQuotient := by sorry
```
All of these lines should be <100 characters (and similarly for the remaining theorems below!)

FormalConjectures/ErdosProblems/36.lean
callesonne: Looking at the reference, I can't see this result in the English summary (only that 1/4 is a lower bound). Also I am a bit confused by the `o` appearing here (and in wikipedia). They don't seem to be present in e.g. this paper https://arxiv.org/pdf/1609.08000 by Haugland

mo271: for the upper bound of 1/2, I added the example from the erdos problems website.
Also I removed the `o`s throughout.

FormalConjectures/ErdosProblems/36.lean
mo271: done

FormalConjectures/ErdosProblems/36.lean
```diff
@@ -0,0 +1,207 @@
...
+*References:*
+ - [erdosproblems.com/36](https://www.erdosproblems.com/36)
+ - [Wikipedial: Minimum overlap problem](https://en.wikipedia.org/wiki/Minimum_overlap_problem)
+-/
+
+/--
+The number of solutions to the equation $a - b = k$, for $a \in A$ and $b \in B$.
+This represents the "overlap" between sets $A$ and $B$ for a given difference $k$.
+-/
+noncomputable def Overlap (A B : Finset ℤ) (k : ℤ) : ℕ := {((a : A), (b : B)) | k = a - b}.ncard
```
Paul-Lez: Small nitpick: since we're assuming `A`, `B` are finsets, I'd be tempted to define this in terms of filtering the product, as I think this would be easier to work with.

Paul-Lez (approved):

FormalConjectures/ErdosProblems/36.lean
mo271: I agree that's nicer!

mo271: done

PR #233, opened 2025-06-14
Split SchanuelsConjecture.lean

Reklle:
As discussed in #151, split the `SchanuelsConjecture.lean` and added some new conjectures.


callesonne (commented):
Thanks for doing this! It would be nice if you could split this PR into two parts, one which only splits the files, and another which adds the new conjectures. This way it will be a lot easier to review (and the file splitting one can get merged pretty much immediately!)

Reklle:
Now only the conjectures form `SchanuelsConjecture.lean` remain. With minor stylistic edits.

FormalConjectures/Wikipedia/Exponentials.lean
```diff
@@ -0,0 +1,45 @@
...
+    (h1 : LinearIndependent ℚ x) (h2 : LinearIndependent ℚ y) :
+    ∃ i j : Fin 2, Transcendental ℚ (exp (x i * y j)) := by
+  sorry
+
+/--
+The four exponential conjecture would imply that for any irrational number $t$,
+at least one of the numbers $2^t$ and $3^t$ is transcendental.
+-/
+@[category research open, AMS 11]
+theorem exists_transcendental_of_two_pow_irrat_three_pow_irrat (t : ℝ) (h : Irrational t) :
```
callesonne: I think this has the wrong name

FormalConjectures/Wikipedia/Exponentials.lean
```diff
@@ -0,0 +1,45 @@
...
+    (h1 : LinearIndependent ℚ x) (h2 : LinearIndependent ℚ y) :
+    ∃ i j : Fin 2, Transcendental ℚ (exp (x i * y j)) := by
+  sorry
+
+/--
+The four exponential conjecture would imply that for any irrational number $t$,
+at least one of the numbers $2^t$ and $3^t$ is transcendental.
+-/
+@[category research open, AMS 11]
+theorem two_pow_three_pow_transcendental (t : ℝ) (h : Irrational t) :
```
callesonne, line 43: I would move this to `Transcendental.lean`, and call it `transcendental_two_pow_or_three_pow`

Reklle: I think so too! But this statement is on [page about four exponentials](https://en.wikipedia.org/wiki/Four_exponentials_conjecture), but it is not on [page about transcendental numbers)](https://en.wikipedia.org/wiki/Transcendental_number)

callesonne: Okay actually looking at it again, I agree with its current location. Thanks a lot!

PR #234, opened 2025-06-14
Add Firoozbakht's conjecture

Reklle:
`firoozbakht_conjecture_consequence` is one of the consequences from [simple puzzles](https://www.primepuzzles.net/conjectures/conj_030.htm).
As I checked, it is violated only for few first $n$

### DIFF
--- a/FormalConjectures/Wikipedia/Firoozbakht.lean
+++ b/FormalConjectures/Wikipedia/Firoozbakht.lean
@@ -42,7 +42,7 @@ theorem firoozbakht_conjecture (n : ℕ) :
   sorry
 
 /--
-The inequality $p_{n+1}-p_n < (\log p_n)^2-\log p_n$ holds for all prime numbers $p_n$.
+The inequality $p_{n+1}-p_n < (\log p_n)^2-\log p_n$ holds for all $n>4$.
 A consequence of Firuzbakht's conjecture.
 -/
 @[category research solved, AMS 11]


FormalConjectures/Wikipedia/Firoozbakht.lean
```diff
@@ -0,0 +1,51 @@
...
+
+*References:*
+- [Wikipedia](https://en.wikipedia.org/wiki/Firoozbakht%27s_conjecture)
+- [primepuzzles](https://www.primepuzzles.net/conjectures/conj_030.htm)
+-/
+
+open Real
+
+/--
+The sequence of $\sqrt[n]{p_n}$ where $p_n$ is a prime number.
```
callesonne:
```suggestion
The sequence of $\sqrt[n]{p_n}$ where $p_n$ is the n:th prime number.
```

FormalConjectures/Wikipedia/Firoozbakht.lean
```diff
@@ -0,0 +1,51 @@
...
+-/
+noncomputable def firoozbakhtSeq (n : ℕ) : ℝ :=
+  (n.nth Prime)^(1/(n+1))
```
callesonne:
```suggestion
  (n.nth Prime)^(1/(n + 1) : ℝ)
```
Be careful with division of natural numbers :) This one defaulted to natural number division, which rounds down, so it would always have been 0!

FormalConjectures/Wikipedia/Firoozbakht.lean
```diff
@@ -0,0 +1,51 @@
...
+@[category research open, AMS 11]
+theorem firoozbakht_conjecture (n : ℕ) :
+    firoozbakhtSeq (n+1) < firoozbakhtSeq n := by
+  sorry
+
+/--
+The inequality $p_{n+1}-p_n < (\log p_n)^2-\log p_n$ holds for all prime numbers $p_n$.
+A consequence of Firuzbakht's conjecture.
+-/
+@[category research open, AMS 11]
```
callesonne: I'm guessing this is not open, since you are allowed to assume firoozbakht_conjecture. Is that right?

callesonne: I am also not sure if it should be tagged `research solved`. Maybe `graduate solved` is more fitting.

Reklle: This consequence is based on the result from the publication that
$$\displaystyle\frac{x}{\log x - 1}<\pi(x)$$ for all $x \geq 3299$
Despite this, this statement is true for $p_n>7$. I have [verified](https://github.com/Reklle/conjectures-check/blob/main/src/conjectures/firoozbakht.rs) this on prime numbers up to $10^9$.
Therefore, this statement seems to me complex enough to be `research solved`

callesonne: Okay great, my bad!

mo271 (approved):
Thanks

PR #236, opened 2025-06-16
add Erdos 274

### LINKED ISSUE #46: Formalise Erdos Problem 274

callesonne:
closes #46

### DIFF
--- a/FormalConjectures/ErdosProblems/274.lean
+++ b/FormalConjectures/ErdosProblems/274.lean
@@ -26,6 +26,8 @@ import FormalConjectures.Util.ProblemImports
 
 open scoped Pointwise Cardinal
 
+-- TODO(callesonne): add already proved results from the wiki page
+
 /--
 If `G` is an abelian group then can there exist an exact covering of `G` by more than one cosets of
 different sizes? (i.e. each element is contained in exactly one of the cosets.)
@@ -50,7 +52,7 @@ theorem erdos_274.variants.nonabelian :
     ∃ (P : Partition (⊤ : Subgroup G)),
       1 < P.parts.ncard ∧
       (∀ A ∈ P.parts, ∃ᵉ (s : G) (H : Subgroup G), s • (H : Set G) = A) ∧
-      P.parts.Pairwise fun A B ↦ Nat.card A ≠ Nat.card B) ↔ answer(sorry) := by
+      P.parts.Pairwise fun A B ↦ #A ≠ #B) ↔ answer(sorry) := by
   sorry
 
 


FormalConjectures/ErdosProblems/274.lean
```diff
@@ -0,0 +1,69 @@
...
+/--
+If `G` is an abelian group then can there exist an exact covering of `G` by more than one cosets of
+different sizes? (i.e. each element is contained in exactly one of the cosets.)
+-/
+@[category research open, AMS 20]
+theorem erdos_274 :
+    (∀ (G : Type*), [CommGroup G] →
+    ∃ (P : Partition (⊤ : Subgroup G)), 1 < P.parts.ncard ∧
+      (∀ A ∈ P.parts, ∃ (s : G) (H : Subgroup G), s • (H : Set G) = A) ∧
+      P.parts.Pairwise fun A B ↦ Nat.card A ≠ Nat.card B) ↔ answer(sorry) := by
```
Paul-Lez: I think you're in trouble here since you're using `Nat.card` without any finitness assumptions on the subgroups.

FormalConjectures/ErdosProblems/274.lean
```diff
@@ -0,0 +1,69 @@
...
+@[category research open, AMS 20]
+theorem erdos_274.variants.nonabelian :
+    (∀ (G : Type*), [Group G] → [Fintype G] →
+    ∃ (P : Partition (⊤ : Subgroup G)),
+      1 < P.parts.ncard ∧
+      (∀ A ∈ P.parts, ∃ᵉ (s : G) (H : Subgroup G), s • (H : Set G) = A) ∧
+      P.parts.Pairwise fun A B ↦ Nat.card A ≠ Nat.card B) ↔ answer(sorry) := by
+  sorry
+
+
```
Paul-Lez, line 58: I think it would also be nice to add the two (already proved) theorems mentioned in the Wiki page!

FormalConjectures/ErdosProblems/274.lean
callesonne: I think not because groups always have size \ge 1. So size is infinite iff Nat.card = 0. But maybe I should still change it because this is a "trick".

Paul-Lez: I think here the issue is that the condition prevents you from having e.g two infinite subgroups of different cardinalities (since they both have Nat.card 0)

callesonne: Ahh for some reason I thought this problem was not really concerned about larger cardinalities (esp. since they use the ambiguous term "size") but maybe it is actually.

callesonne: Okay I just came back to this now. Do you think it's in the spirit of the problem to care about different sized infinite cardinalities? I'm mentioning this because if I choose to formulate this using cardinalities then we lose the `Subgroup.index` API, as I can't find a corresponding function valued in Cardinals.

callesonne: From reading the wiki page it seems like the problem should be stated only in the case of finite indices.

Paul-Lez: Sounds reasonable!

FormalConjectures/ErdosProblems/274.lean
```diff
@@ -0,0 +1,69 @@
...
+[Er97c] Erdős, Paul, Some of my favorite problems and results.
+The mathematics of Paul Erd\H{o}s, I (1997), 47-67.
+-/
+@[category research open, AMS 20]
+theorem erdos_274.variants.nonabelian :
+    (∀ (G : Type*), [Group G] → [Fintype G] →
+    ∃ (P : Partition (⊤ : Subgroup G)),
+      1 < P.parts.ncard ∧
+      (∀ A ∈ P.parts, ∃ᵉ (s : G) (H : Subgroup G), s • (H : Set G) = A) ∧
+      P.parts.Pairwise fun A B ↦ Nat.card A ≠ Nat.card B) ↔ answer(sorry) := by
```
Paul-Lez:
```suggestion
      P.parts.Pairwise fun A B ↦ #A ≠ #B) ↔ answer(sorry) := by
```
(for consistency with above, I think this should work?)

FormalConjectures/ErdosProblems/274.lean
Paul-Lez: Added a TODO for you @callesonne

PR #237, opened 2025-06-16
Erdos 624

### LINKED ISSUE #45: Formalise Erdos Problem 624

smmercuri:
Closes #45

### DIFF
--- a/FormalConjectures/ErdosProblems/624.lean
+++ b/FormalConjectures/ErdosProblems/624.lean
@@ -1,5 +1,5 @@
 /-
-Copyright 2025 Google LLC
+Copyright 2025 The Formal Conjectures Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/FormalConjectures/ErdosProblems/624.lean
+++ b/FormalConjectures/ErdosProblems/624.lean
@@ -1,0 +1,138 @@
+/-
+Copyright 2025 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 624
+
+*Reference:* [erdosproblems.com/624](https://www.erdosproblems.com/624)
+-/
+open Filter
+
+open scoped Topology Finset
+
+/-- Structure encoding functions on finite sets that map the powerset of the
+supplied finite set `X` to `X`. -/
+structure PowersetMapFor {α : Type*} (X : Finset α) where
+  toFun : Finset α → α
+  image : {toFun A | A ⊆ X} ⊆ X
+
+instance {α : Type*} (X : Finset α) : FunLike (PowersetMapFor X) (Finset α) α where
+  coe := PowersetMapFor.toFun
+  coe_injective' f₁ _ _ := by cases f₁; congr
+
+/-- Structure encoding functions on finite sets that, for any finite set `X`,
+map the powerset of `X` to `X`. -/
+structure PowersetMap (α : Type*) where
+  toFun : Finset α → Finset α → α
+  image : ∀ X, {toFun X A | A ⊆ X} ⊆ X
+
+instance {α : Type*} : FunLike (PowersetMap α) (Finset α) (Finset α → α) where
+  coe := PowersetMap.toFun
+  coe_injective' f₁ _ _ := by cases f₁; congr
+
+/-- `f : PowersetMap` is `SurjectiveAbove` a function `H : ℕ → ℕ` if its
+image on the powerset of `Y` is `X` for any `X Y` such that `H #X ≤ #Y`. -/
+def PowersetMap.SurjectiveAbove {α : Type*} (H : ℕ → ℕ)
+    (f : PowersetMap α) : Prop :=
+  ∀ X Y, Y ⊆ X → H #X ≤ #Y → {f X A | A ⊆ Y} = X
+
+/--
+Let $X$ be a finite set of size $n$ and $H(n)$ be such that there
+is a function $f : \{A : A \subseteq X\} \to X$ so that for every
+$Y \subseteq X$ with $|Y|\geq H(n)$ we have
+$$
+  \{f(A) : A \subseteq Y\} = X.
+$$
+Prove that
+$$
+  H(n) - \log_2 n \to\infty.
+$$
+-/
+@[category research open, AMS 5]
+theorem erdos_624 {α : Type*} {H : ℕ → ℕ} {f : PowersetMap α}
+    (hf : f.SurjectiveAbove H) :
+    Tendsto (fun n => H n - Real.logb 2 n) atTop atTop :=
+  sorry
+
+/--
+Erdős and Hajnal proved that
+$$
+  \log_2 n\leq H(n) < \log_2 n + (3 + o(1))\log_2\log_2 n.
+$$
+-/
+@[category research solved, AMS 5]
+theorem erdos_624.variants.interval {α : Type*} {H : ℕ → ℕ} {f : PowersetMap α}
+    (hf : f.SurjectiveAbove H) :
+    ∃ (o : ℕ → ℕ) (ho : Tendsto o atTop (𝓝 0)), ∀ (n : ℕ),
+        Real.logb 2 n ≤ H n ∧ H n < Real.logb 2 n + (3 + o n) * Real.logb 2 (Real.logb 2 n) :=
+  sorry
+
+/--
+The weaker statement that for $n = 2 ^ k$ we have $H(n)\geq k + 1$ was
+proved by Alon.
+-/
+@[category undergraduate, AMS 5]
+theorem erdos_624.variants.pow_two_weak {α : Type*} {H : ℕ → ℕ}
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/624.lean
```diff
@@ -0,0 +1,138 @@
...
+Erdős and Hajnal proved that
+$$
+  \log_2 n\leq H(n) < \log_2 n + (3 + o(1))\log_2\log_2 n.
+$$
+-/
+@[category research solved, AMS 5]
+theorem erdos_624.variants.interval {α : Type*} {H : ℕ → ℕ} {f : PowersetMap α}
+    (hf : f.SurjectiveAbove H) :
+    ∃ (o : ℕ → ℕ) (ho : Tendsto o atTop (𝓝 0)), ∀ (n : ℕ),
+        Real.logb 2 n ≤ H n ∧ H n < Real.logb 2 n + (3 + o n) * Real.logb 2 (Real.logb 2 n) :=
```
callesonne: I think we should `open Real` for this

FormalConjectures/ErdosProblems/624.lean
```diff
@@ -0,0 +1,138 @@
...
+/--
+Erdős and Hajnal proved that
+$$
+  \log_2 n\leq H(n) < \log_2 n + (3 + o(1))\log_2\log_2 n.
+$$
+-/
+@[category research solved, AMS 5]
+theorem erdos_624.variants.interval {α : Type*} {H : ℕ → ℕ} {f : PowersetMap α}
+    (hf : f.SurjectiveAbove H) :
+    ∃ (o : ℕ → ℕ) (ho : Tendsto o atTop (𝓝 0)), ∀ (n : ℕ),
```
callesonne: I would prefer to write `(ho : o =o[atTop] (1 : ℕ → ℝ))`.

callesonne (commented):
Generally this looks good, but I am not quite sure about these `Powerset` definitions (but maybe I can be convinced). How about introducing a definition `Finset.relPowerset Y X` which provides the powerset of `Y` as a subset of `X`. Then you could use `Set.surjOn` to describe `SurjectiveAbove`?

smmercuri:
> Generally this looks good, but I am not quite sure about these `Powerset` definitions (but maybe I can be convinced). How about introducing a definition `Finset.relPowerset Y X` which provides the powerset of `Y` as a subset of `X`. Then you could use `Set.surjOn` to describe `SurjectiveAbove`?

I'm not sure I follow the suggestion here. Currently, we could use `Set.SurjOn` in the definition of `SurjectiveAbove` as follows
```lean
def PowersetMap.SurjectiveAbove {α : Type*} (H : ℕ → ℕ)
    (f : PowersetMap α) : Prop :=
  ∀ X Y, Y ⊆ X → H #X ≤ #Y → Set.SurjOn (f X) Y.powerset X 
```
but the issue here is that this needs to hold for all `X, Y` related via some specified `H`, maybe this is the rel you mean? 

`PowersetMapFor` could just be replaced by functions `f : X.powerset →X` I suppose! Then `PowersetMap` is just the type `(X : Finset α) →X.powerset → X`. I think I was trying to avoid coercing `Finset` to `Type`, but maybe that's better than new definitions

smmercuri:
Oh wait, I think I see what you mean

callesonne:
> > Generally this looks good, but I am not quite sure about these `Powerset` definitions (but maybe I can be convinced). How about introducing a definition `Finset.relPowerset Y X` which provides the powerset of `Y` as a subset of `X`. Then you could use `Set.surjOn` to describe `SurjectiveAbove`?
> 
> I'm not sure I follow the suggestion here. Currently, we could use `Set.SurjOn` in the definition of `SurjectiveAbove` as follows
> 
> ```lean
> def PowersetMap.SurjectiveAbove {α : Type*} (H : ℕ → ℕ)
>     (f : PowersetMap α) : Prop :=
>   ∀ X Y, Y ⊆ X → H #X ≤ #Y → Set.SurjOn (f X) Y.powerset X 
> ```
> 
> but the issue here is that this needs to hold for all `X, Y` related via some specified `H`, maybe this is the rel you mean?
> 
> `PowersetMapFor` could just be replaced by functions `f : X.powerset →X` I suppose! Then `PowersetMap` is just the type `(X : Finset α) →X.powerset → X`. I think I was trying to avoid coercing `Finset` to `Type`, but maybe that's better than new definitions

By `relPowerset` I meant "relative powerset", maybe the naming was a bit confusing. I think avoiding these new definitions would be good even if it introduces coercions, but I maybe I have not understood how annoying these coercions are to deal with.

smmercuri:
> By `relPowerset` I meant "relative powerset", maybe the naming was a bit confusing. I think avoiding these new definitions would be good even if it introduces coercions, but I maybe I have not understood how annoying these coercions are to deal with.

OK I see what you mean now. I tried this as follows
```lean
theorem Finset.mem_powerset_of_subset {α : Type*} {X Y : Finset α} (h : Y ⊆ X) {S : Finset α} 
    (hS : S ∈ Y.powerset) : S ∈ X.powerset := by 
  simp only [mem_powerset] at hS ⊢ 
  exact hS.trans h

def Finset.toPowersetOfSubset {α : Type*} {X Y : Finset α} (h : Y ⊆ X) : Set X.powerset := 
    { ⟨S, mem_powerset_of_subset h hS⟩ | (S : Finset α) (hS : S ∈ Y.powerset) } 

def Finset.SurjectiveAbove {α : Type*} (H : ℕ → ℕ)
    (f : (X : Finset α) → X.powerset → X) : Prop :=
  ∀ X Y, (h : Y ⊆ X) → H #X ≤ #Y → Set.SurjOn (f X) (toPowersetOfSubset h) .univ 
```
which works, but I think it does make the formalisations in this file a bit less neat because I need to keep using  `toPowersetOfSubset` throughout to coerce `Y.powerset` to `X.powerset`. Maybe I'm missing a trick here! I guess it's a question of style and the trade-off of the new definitions is neater formalisations and less coercions, which I personally prefer

FormalConjectures/ErdosProblems/624.lean
```diff
@@ -0,0 +1,139 @@
...
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 624
+
+*Reference:* [erdosproblems.com/624](https://www.erdosproblems.com/624)
+-/
+open Filter
```
callesonne:
```suggestion
-/

open Filter
```

FormalConjectures/ErdosProblems/624.lean
```diff
@@ -0,0 +1,139 @@
...
+
+instance {α : Type*} : FunLike (PowersetMap α) (Finset α) (Finset α → α) where
+  coe := PowersetMap.toFun
+  coe_injective' f₁ _ _ := by cases f₁; congr
+
+/-- `f : PowersetMap` is `SurjectiveAbove` a function `H : ℕ → ℕ` if its
+image on the powerset of `Y` is `X` for any `X Y` such that `H #X ≤ #Y`. -/
+def PowersetMap.SurjectiveAbove {α : Type*} (H : ℕ → ℕ)
+    (f : PowersetMap α) : Prop :=
+  ∀ X Y, Y ⊆ X → H #X ≤ #Y → Set.SurjOn (f X) Y.powerset X
```
callesonne, line 53: I am not completely sure this is accurate, but the quantification on the website seems a bit tricky so I might be wrong. Shouldnt the statement be that for every `n` there exists a finite set `X` and a function `H` such that there is a map `f` that is `SurjectiveAbove H f`?

callesonne:
> > By `relPowerset` I meant "relative powerset", maybe the naming was a bit confusing. I think avoiding these new definitions would be good even if it introduces coercions, but I maybe I have not understood how annoying these coercions are to deal with.
> 
> OK I see what you mean now. I tried this as follows
> 
> ```lean
> theorem Finset.mem_powerset_of_subset {α : Type*} {X Y : Finset α} (h : Y ⊆ X) {S : Finset α} 
>     (hS : S ∈ Y.powerset) : S ∈ X.powerset := by 
>   simp only [mem_powerset] at hS ⊢ 
>   exact hS.trans h
> 
> def Finset.toPowersetOfSubset {α : Type*} {X Y : Finset α} (h : Y ⊆ X) : Set X.powerset := 
>     { ⟨S, mem_powerset_of_subset h hS⟩ | (S : Finset α) (hS : S ∈ Y.powerset) } 
> 
> def Finset.SurjectiveAbove {α : Type*} (H : ℕ → ℕ)
>     (f : (X : Finset α) → X.powerset → X) : Prop :=
>   ∀ X Y, (h : Y ⊆ X) → H #X ≤ #Y → Set.SurjOn (f X) (toPowersetOfSubset h) .univ 
> ```
> 
> which works, but I think it does make the formalisations in this file a bit less neat because I need to keep using `toPowersetOfSubset` throughout to coerce `Y.powerset` to `X.powerset`. Maybe I'm missing a trick here! I guess it's a question of style and the trade-off of the new definitions is neater formalisations and less coercions, which I personally prefer

I wouldn't view this as a coercion, but rather a separate definition (the powerset of `Y` as a subset of the powerset of `X`). I just feel like the current definitions will be quite awkward to use, since you have to define the maps for all finite sets.

Also, my suggestion might have been based on a misunderstanding of the quantification in the problem. If I am correct, I think the definition above should just be
```
def Finset.SurjectiveAbove {α : Type*} (H : ℕ → ℕ) (X : Finset α) 
     (f : X.powerset → X) : Prop :=
   ∀ Y, (h : Y ⊆ X) → H #X ≤ #Y → Set.SurjOn (f X) (toPowersetOfSubset h) .univ 
```
and maybe in this case my suggestion makes more sense, as we only need a map `f` for every `n` (and corresponding `X(n)`).

FormalConjectures/ErdosProblems/624.lean
smmercuri: Yeah I found this one tricky too. I think the issue here is that the final part of the statement considers the limit of `H` in `n`. I'm taking that to mean that `H` should satisfy this property for all `n`, and so `f` needs to be some family of maps `X.powerset → X` with this property, hence why it depends additionally on `Finset α`.

PR #238, opened 2025-06-16
feat(ErdosProblems): 5

### LINKED ISSUE #188: Erdős Problem 5: limit points of $(p_{n+1} - p_n)/\log n$
### What is the conjecture

https://www.erdosproblems.com/5

  Let $C\geq 0$. Is there an infinite sequence of $n_i$ such that
$$\lim_{i\to \infty}\frac{p_{n_i+1}-p_{n_i}}{\log n_i}=C?$$
Status: open

(also a list of related solved problems)

### Prerequisites needed
Should all be in mathlib

* ams-11

### Choose either option

Reklle:
closes #188
I don't know how correct it is, but I decided not to introduce a separate function.


FormalConjectures/ErdosProblems/5.lean
```diff
@@ -0,0 +1,52 @@
...
+
+open Real Filter Function
+open scoped Topology
+
+/--
+For every nonnegative real number x, there exists a strictly increasing sequence of indices (n_i)
+such that the limit of $(p_{n_{i+1}} - p_{n_i}) / log n_i$ equals $x$,
+where $p_k$ denotes the k-th prime number.
+
+Note: the condition `n 0 ≥ 2` ensures the logarithm is defined for all indices
```
smmercuri: I think this condition is not quite needed because `log` takes the junk value `0` for the remaining indices, and `StrictMono` will only allow `log (n 0)` and `log (n 1)` to take junk values. So these should not affect the final asymptotic assertion -- although I guess it can't hurt to leave it in

smmercuri:
Thanks for the contrib! Would you be able to add some of the additional solved results on the [webpage](https://www.erdosproblems.com/5) as variants as well, or add a TODO?

FormalConjectures/ErdosProblems/5.lean
smmercuri: Looking through the sources, it also looks like some of the more modern statements of this problem equivalently use $\log(p_n)$ as the denominator, rather than $\log(n)$, which would avoid this issue entirely

FormalConjectures/ErdosProblems/5.lean
```diff
@@ -0,0 +1,52 @@
...
+-/
+@[category research open, AMS 11]
+theorem erdos_5.finite_case (x : ℝ) (hx : 0 ≤ x) :
+    ∃ n : ℕ → ℕ, StrictMono n ∧ n 0 ≥ 2 ∧
+    Tendsto (fun i ↦ ((n (i+1)).nth Prime  - (n i).nth Prime : ℝ) / log (n i)) atTop (𝓝 x) := by
```
smmercuri: nit: the cast to real can be removed here and in `infinite_case`, this will be inferred from `Real.log` in the denom

Reklle: Yes, I see that! Thank you!

FormalConjectures/ErdosProblems/5.lean
```diff
@@ -0,0 +1,92 @@
...
+where $p_k$ denotes the k-th prime number.
+
+**Note:**
+This definition is equivalent to $(p_{n_{i+1}} - p_{n_i}) / log n_i$
+Since $p_n \sim n \log n$, then $\log n \sim \log p_n$
+-/
+@[category research solved, AMS 11]
+theorem infinite_case :
+    ∃ p : ℕ → ℕ, StrictMono p ∧ (∀ n : ℕ, (p n).Prime) ∧
+    Tendsto (fun i ↦ (p (i+1)  - p i) / log (p i)) atTop atTop := by
```
callesonne:
```suggestion
    Tendsto (fun i ↦ (p (i+1) - p i) / log (p i)) atTop atTop := by
```

FormalConjectures/ErdosProblems/5.lean
```diff
@@ -0,0 +1,92 @@
...
+-/
+
+open Real Filter Function Set
+open scoped Topology
+
+namespace erdos_5
+
+/--
+There exists a strictly increasing sequence of indices (n_i)
+such that the limit of $(p_{n_{i+1}} - p_{n_i}) / log p_{n_i}$ equals $\infty$,
```
callesonne: Is this really correct, from the website it looks like it should be $p_{n_i + 1} - p_{n_i}$ (and this is the reason they introduce this sequence `n_i`!)

smmercuri: Good catch, yeah from the website it looks like it should be $p_{n_i + 1}$ rather than $p_{n_{i + 1}}$

FormalConjectures/ErdosProblems/5.lean
```diff
@@ -0,0 +1,92 @@
...
+where $p_k$ denotes the k-th prime number.
+
+**Note:**
+This definition is equivalent to $(p_{n_{i+1}} - p_{n_i}) / log n_i$
+Since $p_n \sim n \log n$, then $\log n \sim \log p_n$
+-/
+@[category research solved, AMS 11]
+theorem infinite_case :
+    ∃ p : ℕ → ℕ, StrictMono p ∧ (∀ n : ℕ, (p n).Prime) ∧
+    Tendsto (fun i ↦ (p (i+1) - p i) / log (p i)) atTop atTop := by
```
smmercuri, line 42: Given the comment above this should be modified so that only successive prime differences are considered

mo271:
@Reklle it seems like some of the comments are already adresses, could you resolve those?

Reklle:
@mo271, done! Sorry for the long delay

FormalConjectures/ErdosProblems/5.lean
```diff
@@ -0,0 +1,92 @@
...
+where $p_k$ denotes the k-th prime number.
+
+**Note:**
+This definition is equivalent to $(p_{n_i+1} - p_{n_i}) / log n_i$
+Since $p_n \sim n \log n$, then $\log n \sim \log p_n$
+-/
+@[category research solved, AMS 11]
+theorem infinite_case :
+    ∃ p : ℕ → ℕ, StrictMono p ∧ (∀ n : ℕ, (p n).Prime) ∧
+    Tendsto (fun i ↦ (p (i+1) - p i) / log (p i)) atTop atTop := by
```
callesonne, line 42: This is still not correct, you need to also introduce the sequence `n_i`. The function `p` you have introduced is analogous to `p(i) = p_{n_i}` in the problem, so `p(i+1) = p_{n_{i+1}}` and not `p_{n_i + 1}`. I would introduce the sequence `n` and use `Nat.nth Prime` as you had before.

mo271:
@Reklle do you plan to continue working on this?

YaelDillies:
Closing due to inactivity. Feel free to reopen if/when you have time!

PR #242, opened 2025-06-16
Add Erdős 68

Reklle:


FormalConjectures/ErdosProblems/68.lean
```diff
@@ -0,0 +1,33 @@
...
+-/
+
+/--
+Is
+$$\sum_{n=2}^\infty \frac{1}{n!-1}$$
+irrational?
+-/
+@[category research open, AMS 11]
+theorem erdos_68 :
+    Irrational (∑' n : ℕ, 1 / ((n+2).factorial-1 : ℝ)) ↔ answer(sorry) := by
```
Paul-Lez: For readability
```suggestion
    Irrational (∑' n : ℕ, 1 / ((n + 2).factorial - 1 : ℝ)) ↔ answer(sorry) := by
```

FormalConjectures/ErdosProblems/68.lean
```diff
@@ -0,0 +1,33 @@
...
+  sorry
```
Paul-Lez, line 42: Would you mind also adding the alternate formula for this sum from https://www.erdosproblems.com/68?

Paul-Lez (commented):
Thanks for the PR!

FormalConjectures/ErdosProblems/68.lean
Reklle: Done! And I introduced `let f` because it seems to be needed in the proof.

PR #245, opened 2025-06-18
Erdős Problem 645

mo271:
Co-authored-by: Bhavik Mehta <bhavikmehta8@gmail.com>

### DIFF
--- a/FormalConjectures/ErdosProblems/645.lean
+++ b/FormalConjectures/ErdosProblems/645.lean
@@ -1,0 +1,33 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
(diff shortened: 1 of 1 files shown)


b-mehta (commented):
Do you want to add the proof too?

mo271:
> Do you want to add the proof too?

@b-mehta  No, I wasn't planning on doing so, although it could probably be golfed to be 10 lines or so

b-mehta:
Okay! I think 10 lines is a bit optimistic, there is a bit of actual content in those proofs, but I agree with your broader point that these could be golfed.

mo271:
yeah, also we don't want to include longer proofs in the repo, mainly due to the implied maintenance burden.

PR #249, opened 2025-06-18
Conjectures about ranks of elliptic curves

alreadydone:


alreadydone:
Thanks! I think AMS 11 14 is suitable for all of these.

FormalConjectures/Wikipedia/EllipticCurveRank.lean
```diff
@@ -0,0 +1,205 @@
...
+    233 * 241 ^ 2 * 4139 * 678146849364709860535420504397393 *
+    159788990966780131363155786084695062643236502969 *
+    4402149008473369392540402625019227412319473055901 := by
+  rw [elkiesKlagsbrun29, Δ, b₂, b₄, b₆, b₈]; norm_num
+
+@[category test, AMS 11 14]
+instance : elkiesKlagsbrun29.IsElliptic where
+  isUnit := by rw [Δ_elkiesKlagsbrun29]; norm_num
+
+@[category test, AMS 11 14]
```
Paul-Lez: `test` seems a little weird to me (or it the rank calculation particularly easy in that case?). Does this seem sensible?
```suggestion
@[category graduate, AMS 11 14]
```

FormalConjectures/Wikipedia/EllipticCurveRank.lean
```diff
@@ -0,0 +1,205 @@
...
+/-- From [PPVW2016], Section 8.2:
+"Our heuristic predicts (a) All but finitely many E ∈ ℰ satisfy rk E(ℚ) ≤ 21".
+In other words, there are only finitely many elliptic curves over ℚ (up to isomorphism)
+with rank greater than 21.
+Notice that this contradicts the previous conjecture. -/
+@[category research open, AMS 11 14]
+theorem finite_twentyone_lt_finrank : {E : RatEllipticCurve | 21 < E.rank}.Finite := by
+  sorry
+
+/-- [PPVW2016] 8.2(b). Note: ℰ_H should be ℰ_{≤H}, see Theorem 7.3.3. -/
```
Paul-Lez: For this one and the one below, I think it would also be worth having the informal statements in the docstring.

FormalConjectures/Wikipedia/EllipticCurveRank.lean
alreadydone: Good point; what I had in mind is that given the [list of points](https://web.math.pmf.unizg.hr/~duje/tors/rk29.html) one can show they are linearly independent using a computer algebra system (see [this](https://math.stackexchange.com/a/3022032/12932) or the [PARI/GP code](https://listserv.nodak.edu/cgi-bin/wa.exe?A2=NMBRTHRY;b9d018b1.2409&FT=&P=&H=&S=b) given by Elkies). However, finding the 29 points is a research-level problem, so maybe even `research solved` is appropriate.

mo271: let's to with `research solved`

Issue #230, opened 2025-06-13
Beaver Math Olympiad problems


tcosmo:
Apart from [Antihydra](https://wiki.bbchallenge.org/wiki/Antihydra), which was already added [to the list](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/Other/BusyBeaverAntihydra.lean), bbchallenge has reformulated other problems based on the nonhalting of specific Turing machines, we've called this set of problems the "Beaver Math Olympiad".

There are currently two other unsolved problems other than Antihydra:

https://wiki.bbchallenge.org/wiki/Beaver_Math_Olympiad

The mathematical reformulations are simple and you might find them appealing.

In #229, I proposed to go more general (at the expense of losing the nice reformulations (unless we use both and prove them equivalent)).

Paul-Lez:
I think these would be great additions to the Wiki subdirectory! If you're feeling like taking this on, I think having both the "simple" form and the Turing Machine reformulation and proving equivalence could be nice (e.g. that would provide us with a good stress test for the TM definition in the ForMathlib subdirectory)

tcosmo:
OK I'll take it then!

Just to confirm, Antihydra is in the `Other/` subdirectory, in which subdirectory do you prefer I add these?

EDIT:

**Also:** are we interested in the solved ones (they are hard exercises in some cases and have not been formalised).

Paul-Lez:
Ah actually good point, the `Other/` subdirectory is probably better! I think it would be reasonable to have everything in the same file as the Antihydra formalisation. 

Also, feel free to add the solved ones with the `@[category research solved]` attribute - I think this is sensible since they are hard and quite related to other unsolved problems!

tcosmo:
#251 is pretty much ready, statements have also been checked by some bbchallenge contributors ([mei](https://github.com/meithecatte), [Shawn Ligocki](https://github.com/sligocki)). Potentially expect minor updates of the documentation comments, PR can be merged before that if needed (i.e. it's not blocking).

vlad902:
#251 has merged, should this be closed?

franzhusch:
I will close this issue, as #251 was merged. Feel free to reopen, if there was a other reason for not having the issue closed.

PR #251, opened 2025-06-20
Beaver Math Olympiad problems

tcosmo:
This PR adds the current [Beaver Math Olympiad problems](https://wiki.bbchallenge.org/wiki/Beaver_Math_Olympiad#5._1RB0LD_1LC0RA_1RA1LB_1LA1LE_1RF0LC_---0RE_(bbch)) to the project (see #230).

A few notes:

- I have added verbose comments documenting each problem and crediting the bbchallenge contributors who found the associated TMs and mathematical reformulations
- I have used a lean parenthesisation style that is slightly more verbose than was originally done to help non-native lean speakers
- [BMO#4](https://wiki.bbchallenge.org/wiki/Beaver_Math_Olympiad#4._1RB3RB---1LB0LA_2LA4RA3LA4RB1LB_(bbch)) original formulation is based on a partial definition of `a` I have taken the liberty of making it total through the `else` branch
- In BMO#5 statement, I've used [global definition of function "f"](https://github.com/tcosmo/formal-conjectures/blob/579d1ba125ce8b61a4f20b5085af36ccc07f33b5/FormalConjectures/Other/BeaverMathOlympiad.lean#L128), I don't know if it is compatible with your preferred good practice, I will inline it if you prefer
- I added "The bbchallenge Collaboration (bbchallenge.org)" in the AUTHORS, let me know if this is against your authoring guidelines and what we should do instead

### DIFF
--- a/FormalConjectures/Other/BeaverMathOlympiad.lean
+++ b/FormalConjectures/Other/BeaverMathOlympiad.lean
@@ -86,7 +86,7 @@ theorem beaver_math_olympiad_problem_2_antihydra.variants.set
 
 [BMO#3][https://wiki.bbchallenge.org/wiki/Beaver_Math_Olympiad#3._1RB0RB3LA4LA2RA_2LB3RA---3RA4RB_(bbch)_and_1RB1RB3LA4LA2RA_2LB3RA---3RA4RB_(bbch)] is equivalent to the non-termination of 2-state 5-symbol Turing machine [`1RB0RB3LA4LA2RA_2LB3RA---3RA4RB`](https://wiki.bbchallenge.org/wiki/1RB0RB3LA4LA2RA_2LB3RA---3RA4RB) (from all-0 tape).
 
-The machine was informally proven not to halt on [bbchallenge.org](bbchallenge.org) Discord's server, see [this discussion](https://discord.com/channels/960643023006490684/1084047886494470185/1252634913220591728).
+The machine was found and informally proven not to halt by [bbchallenge.org](bbchallenge.org) contributor Daniel Yuan on June 18th 2024, see [Discord discussion](https://discord.com/channels/960643023006490684/1084047886494470185/1252634913220591728).
 -/
 @[category research solved, AMS 5, AMS 11, AMS 68]
 theorem beaver_math_olympiad_problem_3
--- a/FormalConjectures/Other/BeaverMathOlympiad.lean
+++ b/FormalConjectures/Other/BeaverMathOlympiad.lean
@@ -50,7 +50,7 @@ using `↔ answer(sorry)`.
 The machine was discovered by [bbchallenge.org](bbchallenge.org) contributor Jason Yuen on
 June 25th 2024.
 -/
-@[category research open, AMS 5, AMS 11, AMS 68]
+@[category research open, AMS 5 11 68]
 theorem busy_beaver_math_olympiad_problem_1 (a : ℕ → ℕ) (b : ℕ → ℕ)
     (a_ini : a 0 = 1) (a_rec : ∀ n, a (n+1) = if (a n) ≥ (b n) then (a n) - (b n) else 2*(a n) + 1)
     (b_ini : b 0 = 2) (b_rec : ∀ n, b (n+1) = if (a n) ≥ (b n) then 4*(b n) + 2 else (b n) - (a n)):
@@ -74,7 +74,7 @@ that the machine does not halt is based on [a probabilistic argument](https://wi
 This machine and its mathematical reformulations were found by [bbchallenge.org](bbchallenge.org)
 contributors mxdys and Rachel Hunter on June 28th 2024.
 -/
-@[category research open, AMS 5, AMS 11, AMS 68]
+@[category research open, AMS 5 11 68]
 theorem beaver_math_olympiad_problem_2_antihydra
     (a : ℕ → ℕ) (b : ℕ → ℤ)
     (a_ini : a 0 = 8) (a_rec : ∀ n, a (n+1) = (3*(a n)/2 : ℕ))
@@ -88,7 +88,7 @@ theorem beaver_math_olympiad_problem_2_antihydra
 Alternative statement of beaver_math_olympiad_problem_2_antihydra
 using set size comparison instead of a recurrent sequence b.
 -/
-@[category research open, AMS 5, AMS 11, AMS 68]
+@[category research open, AMS 5 11 68]
 theorem beaver_math_olympiad_problem_2_antihydra.variants.set
     (a : ℕ → ℕ)
     (a_ini : a 0 = 8) (a_rec : ∀ n, a (n+1) = (3*(a n)/2 : ℕ)) :
@@ -104,7 +104,7 @@ theorem beaver_math_olympiad_problem_2_antihydra.variants.set
 The machine was found and informally proven not to halt by [bbchallenge.org](bbchallenge.org)
 contributor Daniel Yuan on June 18th 2024; see [Discord discussion](https://discord.com/channels/960643023006490684/1084047886494470185/1252634913220591728).
 -/
-@[category research solved, AMS 5, AMS 11, AMS 68]
+@[category research solved, AMS 5 11 68]
 theorem beaver_math_olympiad_problem_3
     (a : ℕ → ℕ)
     (a_ini : a 0 = 2)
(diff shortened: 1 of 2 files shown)


tcosmo:
Also, let me know if you require a specific maximum number of characters per line ~~(if you do, maybe it should be added in the README)~~ and I will update the long doc comments.

Reklle:
As far as I know, the limit on the length of a string is 100. This is the standard if you use Lean in VS Code.

tcosmo:
Ahh yep, I just saw it [here](https://github.com/google-deepmind/formal-conjectures/blob/8894ec5fdeff6ec2e56506f04670e3d5a732835d/.vscode/settings.json#L4)

But vscode did not enforce the rule on my code, probably due to some setting I have at the user level... I will add a commit correcting this. (Or it seems the ruler is just indicative)

Thank you!

tcosmo:
Done in b05e3242b368e1491aa50ef92b626b10ddb053d2: I broke lines that were > 100 characters apart from when this happened in the middle of a long link or as a consequence of a long link earlier in the sentence, e.g. `[BMO#2](https://wiki.bbchallenge.org/wiki/Beaver_Math_Olympiad#2._Hydra_and_Antihydra) formulation variant`.

FormalConjectures/Other/BeaverMathOlympiad.lean
```diff
@@ -0,0 +1,158 @@
...
+[BMO#1](https://wiki.bbchallenge.org/wiki/Beaver_Math_Olympiad#1._1RB1RE_1LC0RA_0RD1LB_---1RC_1LF1RE_0LB0LE_(bbch)) is equivalent to asking whether the 6-state Turing machine
+[`1RB1RE_1LC0RA_0RD1LB_---1RC_1LF1RE_0LB0LE`](https://wiki.bbchallenge.org/wiki/1RB1RE_1LC0RA_0RD1LB_---1RC_1LF1RE_0LB0LE) halts or not.
+
+There is presently no consensus on whether the machine halts or not, hence the problem is formulated
+using `↔ answer(sorry)`.
+
+The machine was discovered by [bbchallenge.org](bbchallenge.org) contributor Jason Yuen on
+June 25th 2024.
+-/
+@[category research open, AMS 5, AMS 11, AMS 68]
```
Paul-Lez: We usually write:
```suggestion
@[category research open, AMS 5 11 68]
```
(and similarly below)

FormalConjectures/Other/BeaverMathOlympiad.lean
```diff
@@ -0,0 +1,158 @@
...
+theorem beaver_math_olympiad_problem_2_antihydra.variants.set
+    (a : ℕ → ℕ)
+    (a_ini : a 0 = 8) (a_rec : ∀ n, a (n+1) = (3*(a n)/2 : ℕ)) :
+    ∀ n, ((Finset.Ico 0 n).filter fun x ↦ Odd (a x)).card ≤
+        2*((Finset.Ico 0 n).filter fun x ↦ Even (a x)).card := by
+  sorry
+
+/--
+[BMO#3][https://wiki.bbchallenge.org/wiki/Beaver_Math_Olympiad#3._1RB0RB3LA4LA2RA_2LB3RA---3RA4RB_(bbch)_and_1RB1RB3LA4LA2RA_2LB3RA---3RA4RB_(bbch)]
```
Paul-Lez, line 116: I think it would be helpful to add the informal statement of this problem in the docstring too!

FormalConjectures/Other/BeaverMathOlympiad.lean
```diff
@@ -0,0 +1,158 @@
...
+
+The machine was informally proven not to halt [bbchallenge.org](bbchallenge.org)
+contributor Daniel Yuan on July 19th 2024; see [sketched proof](https://wiki.bbchallenge.org/wiki/1RB3RB---1LB0LA_2LA4RA3LA4RB1LB) and [Discord discussion](https://discord.com/channels/960643023006490684/960643023530762343/1263666591900631210).
+-/
+@[category research solved, AMS 5, AMS 11, AMS 68]
+theorem beaver_math_olympiad_problem_4
+    (a : ℕ → ℕ)
+    (a_ini : a 0 = 2)
+    (a_rec : ∀ n, a (n+1) = if (a n)%3 = 0 then (a n)/3 + 2^n + 1
+                                           else ((a n) - 2)/3 + 2^n - 1) :
```
Paul-Lez:
```suggestion
    (a_rec : ∀ n, a (n+1) 
      = if (a n) % 3 = 0 then (a n) / 3 + 2^n + 1 else ((a n) - 2) / 3 + 2^n - 1) :
```

FormalConjectures/Other/BeaverMathOlympiad.lean
```diff
@@ -0,0 +1,158 @@
...
+in [Rocq](https://github.com/ccz181078/busycoq/blob/BB6/verify/1RB0LD_1LC0RA_1RA1LB_1LA1LE_1RF0LC_---0RE.v).
+-/
+def f (x: ℕ) :=
+  10*2^x - 1
+
+@[category research open, AMS 5, AMS 11, AMS 68]
+theorem beaver_math_olympiad_problem_5 (a : ℕ → ℕ) (b : ℕ → ℕ)
+    (a_ini : a 0 = 0) (a_rec : ∀ n, a (n+1) = if b n ≥ f (a n) then a n + 1 else a n)
+    (b_ini : b 0 = 5) (b_rec : ∀ n, b (n+1) = if b n ≥ f (a n) then b n - f (a n)
+                                                               else 3*b n + a n + 5) :
```
Paul-Lez:
```suggestion
    (b_ini : b 0 = 5) 
    (b_rec : ∀ n, b (n+1) = if b n ≥ f (a n) then b n - f (a n) else 3 * b n + a n + 5) :
```

FormalConjectures/Other/BeaverMathOlympiad.lean
```diff
@@ -0,0 +1,158 @@
...
+The machine was discovered by [bbchallenge.org](bbchallenge.org) contributor Jason Yuen on
+June 25th 2024.
+-/
+@[category research open, AMS 5, AMS 11, AMS 68]
+theorem busy_beaver_math_olympiad_problem_1 (a : ℕ → ℕ) (b : ℕ → ℕ)
+    (a_ini : a 0 = 1) (a_rec : ∀ n, a (n+1) = if (a n) ≥ (b n) then (a n) - (b n) else 2*(a n) + 1)
+    (b_ini : b 0 = 2) (b_rec : ∀ n, b (n+1) = if (a n) ≥ (b n) then 4*(b n) + 2 else (b n) - (a n)):
+    (∃ i, a i = b i) ↔ answer(sorry) := by
+  sorry
+
```
Paul-Lez, line 73: If you feel like it, it could be nice to add a Turing machine version of each of these statements and state the equivalence as a theorem with `category test`. For example, we could have something along the lines of:
```suggestion
inductive Γ
  | A
  | B
  | C
  | D
  | E
  | F

open Turing BusyBeaver

def BusyBeaver1 : Machine Γ (Fin 2) :=
  -- define the function using `match`...

theorem busy_beaver_math_olympiad_problem_1_halting_version : BusyBeaver1.IsHalting := by 
  sorry 
    
theorem busy_beaver_math_olympiad_problem_1_equivalence : 
    type_of% busy_beaver_math_olympiad_problem_1 ↔
       type_of% busy_beaver_math_olympiad_problem_1_halting_version := by
  sorry
```

That being said, this should be a lot faster once I've added the Turing Machine elaborator (since you wouldn't need to mess around with defining the function explicitly and so on), so this can be done in another PR:)

Paul-Lez (commented):
Thanks for the PR! A few initial comments - I'll take a another look later this week:)

FormalConjectures/Other/BeaverMathOlympiad.lean
tcosmo: OK, will update, note that the original file was using style `AMS 5, AMS 11`

FormalConjectures/Other/BeaverMathOlympiad.lean
tcosmo: Yes, I think it is a great idea to add those tests but I think it would make more sense to wait for your elaborator, so I will open a PR in the future for this :)

FormalConjectures/Other/BeaverMathOlympiad.lean
tcosmo: Capture d’écran 2025-06-30 à 13 22 09

This is the informal statement of the problem, are you sure it would be useful to the docstring?

FormalConjectures/Other/BeaverMathOlympiad.lean
tcosmo: Done in [f92c68d](https://github.com/google-deepmind/formal-conjectures/pull/251/commits/f92c68dd0735ed5f957c70db945a3d69e0376722)

FormalConjectures/Other/BeaverMathOlympiad.lean
Paul-Lez: I think it would (minus the discord link perhaps); the main use case is that it gives people an easy way to make sure that the contents of the statement in the docstring matches the formalisation, without having to leave a code editor and so on.

tcosmo: Do we want the other statements as well or just this one because of its cumbersomeness?

Paul-Lez: I think all statements would be great!

tcosmo: All added in 1e4681084b796421b826ca267f96c28a3917b685 :)

Paul-Lez:
I think you have merge conflicts that need to be fixed!

tcosmo:
> I think you have merge conflicts that need to be fixed!

I don't understand what's going on, I don't see any conflicts when I pull locally and github is unable to tell either:

Capture d’écran 2025-07-03 à 20 03 19

Paul-Lez:
Ah it seems the issue was that someone edited the file that you deleted - I've just pushed a fix!

PR #259, opened 2025-06-26
Add cardinality assumption to `erdos_402.variants.equality`

smmercuri:
AlphaProof showed that equality also holds for `{12}`, as it does for any singleton.

Digging into the references [[Sz86]](https://link.springer.com/article/10.1007/BF02579410) sounds like it expects `4 ≤ A.card`, while [[BaSo96]](https://eudml.org/doc/206861) explicitly asserts `5 ≤ A.card` (but this latter would mean `{2, 3, 4, 6}` should also be removed). Going with `4` for now.

PR #263, opened 2025-06-26
chore(ErdosProblems/402): fix misformalisation

b-mehta:
Reformulate the equality case.
- The equality case is intended to assert that the mutual gcd is 1: this is on the website and in the docstring already
- As a consequence, no assumption on the cardinality is needed (eg for singletons, we can only have the gcd being 1 if it's already {1}, and then we're in the first of the three cases, so the theorem holds).
- The equality case is stated here by saying that (a,b) >= a/|A| for all a,b, as written in the fourth paragraph of the introduction in [BaSo96].

### DIFF
--- a/FormalConjectures/ErdosProblems/402.lean
+++ b/FormalConjectures/ErdosProblems/402.lean
@@ -46,11 +46,11 @@ an iff statement.
 -/
 @[category research solved, AMS 11]
 theorem erdos_402.variants.equality (A : Finset ℕ) (h₁ : 0 ∉ A) (h₂ : A.Nonempty)
-    (h₃ : 4 ≤ A.card)
-    (hA : ∀ n, A ≠ Finset.Icc 1 n ∧
-      A ≠ ((Finset.Icc 1 n).image fun i => ((Finset.Icc 1 n).lcm id) / i) ∧
-      A ≠ {2,3,4,6}) :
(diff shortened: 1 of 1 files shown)


b-mehta:
As mentioned [here](https://github.com/google-deepmind/formal-conjectures/pull/259#issuecomment-3008929572), this fixes a misformalisation in the strong sense that the current version proves false.

FormalConjectures/ErdosProblems/402.lean
```diff
@@ -46,11 +46,11 @@ an iff statement.
 -/
 @[category research solved, AMS 11]
 theorem erdos_402.variants.equality (A : Finset ℕ) (h₁ : 0 ∉ A) (h₂ : A.Nonempty)
-    (h₃ : 4 ≤ A.card)
-    (hA : ∀ n, A ≠ Finset.Icc 1 n ∧
-      A ≠ ((Finset.Icc 1 n).image fun i => ((Finset.Icc 1 n).lcm id) / i) ∧
-      A ≠ {2,3,4,6}) :
-    ∃ᵉ (a ∈ A) (b ∈ A), a.gcd b < (a / A.card : ℚ) := by
+    (h₃ : A.gcd id = 1)
```
smmercuri, line 49: Ah this hypothesis was present in an older version but got lost later on -- this one is indeed very important, otherwise we need to include `{2k, 3k, 4k, 6k}`

PR #264, opened 2025-06-27
feat: Add conjecture about Fibonacci primes

tlringer:
This conjecture states that there are infintely many Fibonacci primes. Discussed this with @mo271 on Mathstodon; notes about the conjecture are in comments in the code. Feel free to edit to add examples or to make it more stylistically appropriate for Lean, or to fix any mistakes if you find them (Lean is not my native theorem proving language).

### DIFF
--- a/FormalConjectures/Wikipedia/FibonacciPrimes.lean
+++ b/FormalConjectures/Wikipedia/FibonacciPrimes.lean
@@ -16,13 +16,13 @@ limitations under the License.
 
 import FormalConjectures.Util.ProblemImports
 
-/-
- - This conjecture states that there are infinite Fibonacci primes
- - (i.e., Fibonacci numbers that are prime). It is unsolved. It is
- - stated on Wikipedia here: https://en.wikipedia.org/wiki/Fibonacci_prime.
- - It is also a barrier to defining a benchmark from this paper:
- - https://arxiv.org/html/2505.13938v1 (see Figure 8).
+/--
+ This conjecture states that there are infinite Fibonacci primes
+ (i.e., Fibonacci numbers that are prime). It is unsolved. It is
+ stated on Wikipedia here: https://en.wikipedia.org/wiki/Fibonacci_prime.
+ It is also a barrier to defining a benchmark from this paper:
+ https://arxiv.org/html/2505.13938v1 (see Figure 8).
  -/
-@[category research open]
-theorem fib_primes_infinite : { n : ℕ | n.fib.Prime}.Infinite :=
-  by sorry
+@[category research open, AMS 11]
+theorem fib_primes_infinite : {n : ℕ | n.fib.Prime}.Infinite := by
+  sorry
--- a/FormalConjectures/Wikipedia/FibonacciPrimes.lean
+++ b/FormalConjectures/Wikipedia/FibonacciPrimes.lean
@@ -17,12 +17,12 @@ limitations under the License.
 import FormalConjectures.Util.ProblemImports
 
 /--
- This conjecture states that there are infinite Fibonacci primes
- (i.e., Fibonacci numbers that are prime). It is unsolved. It is
- stated on Wikipedia here: https://en.wikipedia.org/wiki/Fibonacci_prime.
- It is also a barrier to defining a benchmark from this paper:
- https://arxiv.org/html/2505.13938v1 (see Figure 8).
- -/
+This conjecture states that there are infinite Fibonacci primes
+(i.e., Fibonacci numbers that are prime). It is unsolved. It is
+stated on Wikipedia here: https://en.wikipedia.org/wiki/Fibonacci_prime.
+It is also a barrier to defining a benchmark from this paper:
+https://arxiv.org/html/2505.13938v1 (see Figure 8).
+-/
 @[category research open, AMS 11]
 theorem fib_primes_infinite : {n : ℕ | n.fib.Prime}.Infinite := by
   sorry
--- a/FormalConjectures/Wikipedia/FibonacciPrimes.lean
+++ b/FormalConjectures/Wikipedia/FibonacciPrimes.lean
@@ -1,0 +1,35 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-
+ - This conjecture states that there are infinite Fibonacci primes
+ - (i.e., Fibonacci numbers that are prime). It is unsolved. It is
+ - stated on Wikipedia here: https://en.wikipedia.org/wiki/Fibonacci_prime.
+ - It is also a barrier to defining a benchmark from this paper:
+ - https://arxiv.org/html/2505.13938v1 (see Figure 8).
+ -/
+
+-- predicate that is true whenever a natural number n is a Fibonacci number
+def Fibonacci (n : ℕ) := ∃ (i : ℕ), Nat.fib i = n
+
+-- define the Fibonacci primes
+def fib_primes := { n : ℕ | Fibonacci n ∧ Prime n }
+
+-- state the conjecture
+theorem fib_primes_infinite : Infinite fib_primes :=
+  by sorry
--- a/FormalConjectures/Wikipedia/FibonacciPrimes.lean
+++ b/FormalConjectures/Wikipedia/FibonacciPrimes.lean
@@ -31,5 +31,6 @@ def Fibonacci (n : ℕ) := ∃ (i : ℕ), Nat.fib i = n
 def fib_primes := { n : ℕ | Fibonacci n ∧ Prime n }
 
 -- state the conjecture
+@[category research open]
 theorem fib_primes_infinite : Infinite fib_primes :=
   by sorry
--- a/FormalConjectures/Wikipedia/FibonacciPrimes.lean
+++ b/FormalConjectures/Wikipedia/FibonacciPrimes.lean
@@ -17,7 +17,7 @@ limitations under the License.
 import FormalConjectures.Util.ProblemImports
 
 /--
-This conjecture states that there are infinite Fibonacci primes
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/FibonacciPrimes.lean
```diff
@@ -0,0 +1,36 @@
...
+
+-- predicate that is true whenever a natural number n is a Fibonacci number
+def Fibonacci (n : ℕ) := ∃ (i : ℕ), Nat.fib i = n
+
+-- define the Fibonacci primes
+def fib_primes := { n : ℕ | Fibonacci n ∧ Prime n }
+
+-- state the conjecture
+@[category research open]
+theorem fib_primes_infinite : Infinite fib_primes :=
```
alreadydone: Welcome! I'd suggest the following formalization without any auxiliary defs:
```suggestion
theorem setOf_fib_prime_infinite : {n : ℕ | n.fib.Prime}.Infinite :=
```

tlringer: OK, I've updated the conjecture appropriately

FormalConjectures/Wikipedia/FibonacciPrimes.lean
```diff
@@ -0,0 +1,28 @@
...
+
+import FormalConjectures.Util.ProblemImports
+
+/-
+ - This conjecture states that there are infinite Fibonacci primes
+ - (i.e., Fibonacci numbers that are prime). It is unsolved. It is
+ - stated on Wikipedia here: https://en.wikipedia.org/wiki/Fibonacci_prime.
+ - It is also a barrier to defining a benchmark from this paper:
+ - https://arxiv.org/html/2505.13938v1 (see Figure 8).
+ -/
```
mo271:
```suggestion
/--
 This conjecture states that there are infinite Fibonacci primes
 (i.e., Fibonacci numbers that are prime). It is unsolved. It is
 stated on Wikipedia here: https://en.wikipedia.org/wiki/Fibonacci_prime.
 It is also a barrier to defining a benchmark from this paper:
 https://arxiv.org/html/2505.13938v1 (see Figure 8).
 -/
```

This way the comments get rendered in the documentation correctly

FormalConjectures/Wikipedia/FibonacciPrimes.lean
```diff
@@ -0,0 +1,28 @@
...
+@[category research open]
```
mo271:
```suggestion
@[category research open, AMS 11]
```
I suppose "Number Theory" is the suitable AMS category here

FormalConjectures/Wikipedia/FibonacciPrimes.lean
```diff
@@ -0,0 +1,28 @@
...
+theorem fib_primes_infinite : { n : ℕ | n.fib.Prime}.Infinite :=
+  by sorry
```
mo271:
```suggestion
theorem fib_primes_infinite : {n : ℕ | n.fib.Prime}.Infinite := by
  sorry
```

Styling nits

mo271 (approved):
Thanks for the contribution! LGTM

FormalConjectures/Wikipedia/FibonacciPrimes.lean
alreadydone: Indeed it's markdown and those `-` will probably be rendered as bullet points :)
but does doc-gen4 actually run on this repo? I can't find a link to the documentation.

nit: infinite -> infinitely many

b-mehta: It does: https://google-deepmind.github.io/formal-conjectures/

alreadydone: Thanks! Someone should make a PR to add it to the README. And the search functionality seems to be broken:
image

tlringer:
I'm OK with the remaining suggested changes, but will be traveling tomorrow
without an ability to commit them. Feel free to edit directly with those
changes, or I'll get to them next time I have Internet and no baby on my
lap :)

FormalConjectures/Wikipedia/FibonacciPrimes.lean
alreadydone: BTW I had déjà vu when I saw the paper you linked, but it turns out it's different from the work I saw a few days ago ([Verina](https://arxiv.org/pdf/2505.23135)) ...

FormalConjectures/Wikipedia/FibonacciPrimes.lean
```diff
@@ -0,0 +1,28 @@
...
+
+import FormalConjectures.Util.ProblemImports
+
+/--
+ This conjecture states that there are infinite Fibonacci primes
+ (i.e., Fibonacci numbers that are prime). It is unsolved. It is
+ stated on Wikipedia here: https://en.wikipedia.org/wiki/Fibonacci_prime.
+ It is also a barrier to defining a benchmark from this paper:
+ https://arxiv.org/html/2505.13938v1 (see Figure 8).
+ -/
```
mo271:
```suggestion
This conjecture states that there are infinite Fibonacci primes
(i.e., Fibonacci numbers that are prime). It is unsolved. It is
stated on Wikipedia here: https://en.wikipedia.org/wiki/Fibonacci_prime.
It is also a barrier to defining a benchmark from this paper:
https://arxiv.org/html/2505.13938v1 (see Figure 8).
-/
```

FormalConjectures/Wikipedia/FibonacciPrimes.lean
mo271: @alreadydone I think we hadn't linked it in the README, because it is linked in the about section in the github ui, but of course for people looking at the README outside of github it would be good to add it!

Actually doing
https://github.com/google-deepmind/formal-conjectures/pull/266
I see I even had made a TODO for it in the README already,
Thanks for bringing it up!

mo271: made an issue on the search functionality: https://github.com/google-deepmind/formal-conjectures/issues/267

FormalConjectures/Wikipedia/FibonacciPrimes.lean
```diff
@@ -0,0 +1,28 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/--
+This conjecture states that there are infinite Fibonacci primes
```
mo271:
```suggestion
This conjecture states that there are infinitely many Fibonacci primes
```

Paul-Lez (approved):

FormalConjectures/Wikipedia/FibonacciPrimes.lean
Shreyas4991: @alreadydone this correction is wrong. @tlringer 's version of the conjecture states that `n` is a fibonacci number and prime. In this change `n.fib` computes the `nth` fibonacci number. EDIT : This comment is after discussing with Talia. On second thoughts, this change might be equivalent to the original conjecture but it still isn't the intended meaning of the author.

tlringer: I trusted this correction because I don't know Lean syntax very well, but since they are indeed different, can we please revert to the original intention? I think each set will only be infinite if the other is too, but they are still fundamentally different sets, and the conjecture is about the set of Fibonacci primes, not the set of indices of Fibonacci primes.

mo271: @tlringer I'm sorry for the misunderstanding here: 
I assume the thought process of @alreadydone for the simplification went something like this: 
```lean
({n : ℕ | (∃ m : ℕ, m.fib = n) ∧ n.Prime}).Infinite 
```
... wait this can be written with the `''` notation for the image as 
```lean
(Nat.fib '' {n : ℕ | n.fib.Prime}).Infinite 
```
... wait `Nat.fib '' {n : ℕ | n.fib.Prime}` is clearly infinite iff `{n : ℕ | n.fib.Prime}` is infinite, so let's drop the `Nat.fib ''` part. 

Indeed here's a quickly written up proof of this (can certainly be golfed into oblivion)
```lean
theorem indices_infinite_iff_fib_primes_infinite : {n : ℕ | n.fib.Prime}.Infinite ↔
    ({n : ℕ | (∃ m : ℕ, m.fib = n) ∧ n.Prime}).Infinite := by
  simp [Set.infinite_iff_exists_gt]
  constructor
  · intros h a
    rcases h (a + 1) with ⟨b', ⟨hb₁, hb₂⟩⟩
    use Nat.fib b'
    exact ⟨⟨exists_apply_eq_apply Nat.fib b', hb₁⟩,
      by linear_combination b'.le_fib_add_one + hb₂⟩
  · intro h a
    rcases h (Nat.fib a) with ⟨b', ⟨⟨m, hm⟩, hb₁⟩, hb₂⟩
    use m
    constructor
    · simp_all
    · have := @m.fib_mono a
      omega
```

I have no strong opinion on if  `{n : ℕ | n.fib.Prime}.Infinite` or `({n : ℕ | (∃ m : ℕ, m.fib = n) ∧ n.Prime}).Infinite` is nicer, but if there is a consensus that the second version is more pretty/useful/readable I'm happy to prepare a pull request.

tlringer:
I think maybe the easiest thing to do is add a comment explaining that this
is phrased to be about the set of indices rather than the set of Fibonacci
primes, though one is infinite iff the other is infinite.

In terms of benchmarks, I do suspect in terms of automated proofs, it will
be easier if the set of Fibonacci primes is explicitly defined, since then
the AI system won't need to define it itself. But who knows, maybe the
proof about the indices is easier.

Wikis tend to treat these as two different but isomorphic questions, and so
do numeric sequences databases (two distinct sequences, one for the
Fibonacci primes and one for the indices of Fibonacci primes). So maybe
it's worth having both theorems as open and the proof that they are
equivalent statements included. But this isn't my benchmark suite so I
don't know for sure what you are optimizing for.

FormalConjectures/Wikipedia/FibonacciPrimes.lean
mo271: added the original variant in #355, comments welcome

mo271:
> In terms of benchmarks, I do suspect in terms of automated proofs, it will be easier if the set of Fibonacci primes is explicitly defined, since then the AI system won't need to define it itself. But who knows, maybe the proof about the indices is easier. 

I'd be surprised if there was an AI system that can make any progress on this if it cannot see through this equivalence with ease. I'm confident that current AI systems can prove the equivalence `indices_infinite_iff_fib_primes_infinite`. Hence I would expect that it doesn't matter, since it would be unlikely to have a system that can only solve one of the variants.

PR #265, opened 2025-06-28
Sum of three cubes

kckennylau:
An integer `n` is a sum of three cubes of integers iff `n` is not 4 or 5 mod 9.

### DIFF
--- a/FormalConjectures/Mathoverflow/100324.lean
+++ b/FormalConjectures/Mathoverflow/100324.lean
@@ -1,0 +1,60 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, looks great!

FormalConjectures/Mathoverflow/100324.lean
```diff
@@ -0,0 +1,60 @@
...
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Mathoverflow 100324
+
+An integer `n : ℤ` can be written as a sum of three cubes (of integers) if and only if
+`n` is not `4` or `5` mod `9`.
+
+*Reference:* [mathoverflow/100324](https://mathoverflow.net/a/100324)
+asked by user [*David Feldman*](https://mathoverflow.net/users/10909/david-feldman)
```
alreadydone, line 26: Also [on Wikipedia](https://en.wikipedia.org/wiki/Sums_of_three_cubes) :)

mo271: yeah, for this one Wikipedia is perhaps a better place, I will make a pr to move it later, if nobody beats me to it...

mo271: done in #281

PR #268, opened 2025-06-28
Covering a unit square with (1/(n+1))-by-(1/(n+2)) rectangles

kckennylau:


FormalConjectures/Mathoverflow/34145.lean
```diff
@@ -0,0 +1,189 @@
...
+/-- A rectangle is specified by its width, height, starting point, and rotation.
+The rectangle is assumed to start in the lower left corner. For example, the unit square
+`{ (x, y) | 0 ≤ x ≤ 1, 0 ≤ y ≤ 1 }` is specified as `⟨1, 1, (0, 0), 0⟩`  -/
+structure Rectangle : Type where
+  width : ℝ
+  height : ℝ
+  start : ℝ × ℝ
+  rotation : Angle
+
+/-- A combination of a rotation and a translation to map the a standard rectangle to the desired
```
mo271:
```suggestion
/-- A combination of a rotation and a translation to map the standard rectangle to the desired
```

FormalConjectures/Mathoverflow/34145.lean
```diff
@@ -0,0 +1,189 @@
...
+    convert tendsto_one_div_add_atTop_nhds_zero_nat.const_sub (1 : ℝ) (c := 0) using 1
+    norm_num
+
+/-- A configuration of rectangles of sides `1 / (n + 1)` and `1 / (n + 2)`. -/
+structure Configuration : Type where
+  rect (n : ℕ) : Rectangle
+  rect_width (n : ℕ) : (rect n).width = 1 / (n + 1)
+  rect_height (n : ℕ) : (rect n).height = 1 / (n + 2)
+
+/-- A "packing" means that the interiors of the rectangles are disjoint. -/
```
mo271:
```suggestion
/-- A "packing" means that the interiors of any two rectangles are disjoint. -/
```

FormalConjectures/Mathoverflow/34145.lean
```diff
@@ -0,0 +1,189 @@
...
+      (∀ n, (c.rect n).toSet ⊆ Rectangle.toSet ⟨133/132, 133/132, (0, 0), 0⟩) ∧
+      c.IsPacking) := by
+  sorry
+
+/-- It is known that packing the rectangles into a square of side length `501/500` is possible.
+
+Reference: https://www.sciencedirect.com/science/article/pii/S0167506008706009
+-/
+@[category research solved, AMS 51]
+theorem rectangles_pack_square_501_div_500 :
```
mo271, line 185: Is this the current recorcd? Then it might be a good idea to add as an open problem to find a smaller epsilon here? (Or have it as open probem if there is a `500 < k` with ... ⊆ Rectangle.toSet ⟨(k + 1/k, (k + 1/k, (0, 0), 0⟩)
Finding the precise ($1 + \varepsilon$) with an `answer(_)` as an open problem might not be so interesting because if the true answer is not $\varepsilon = 0$, then it might be hard to even write it down in an easy way.

mo271: Not the record, https://mathoverflow.net/questions/34145/can-we-cover-the-unit-square-by-these-rectangles/38319#38319
also see here: https://mathoverflow.net/questions/34145/can-we-cover-the-unit-square-by-these-rectangles/38319#comment80898_34379

PR #284, opened 2025-06-30
Add "Is Every Convex Polyhedron Rupert"

### LINKED ISSUE #283: Every Convex Polyhedron is Rupert
### What is the conjecture

Every convex polyhedron "has the Rupert property", that is, somewhat less formally put, a hole can be cut in the polyhedron through which another copy of the polyhedron can be passed. More formally, for any polyhedron (considered as a closed subset $X \subseteq \mathbb{R}^3$), there exists two affine
transformations $T_1, T_2 \in SE(3)$ such that the 2-dimensional projection
of $T_1 X$ lies entirely in the interior of the 2-dimensional projection of $T_2 X$.

### Literature

A citation I know of for this conjecture is
[R. P. Jerrard, J. E. Wetzel, and L. Yuan. Platonic passages. Math.
Mag., 90(2):87–98, 2017.](https://www.researchgate.net/publication/314715434_Platonic_Passages/link/5930413245851553b67f05a0/download?_tp=eyJjb250ZXh0Ijp7ImZpcnN0UGFnZSI6InB1YmxpY2F0aW9uIiwicGFnZSI6InB1YmxpY2F0aW9uIn19)
I welcome earlier citations if any can be found.

More information can be found in [wikipedia](https://en.wikipedia.org/wiki/Prince_Rupert%27s_cube) and the arXiv paper [An algorithmic approach to Rupert's problem](https://arxiv.org/pdf/2112.13754), which has a pretty good bibilography.

Progress has already been made formalizing this in Lean in [this github repository](https://github.com/dwrensha/Rupert.lean). I expect to slightly modify the definition from the one given there somewhat to anticipate some future changes to mathlib's treatment of Lp-normed spaces.

### Prerequisites needed

Definition of Rupert property. 

### Choose either option

jcreedcmu:
Closes #283.

### DIFF
--- a/FormalConjectures/ErdosProblems/672.lean
+++ b/FormalConjectures/ErdosProblems/672.lean
@@ -23,7 +23,7 @@ import FormalConjectures.Util.ProblemImports
 -/
 /-- Erdős problem 672 conjectures that the below holds for any $k ≥ 4$ and $l > 1$. -/
 def Erdos672With (k l : ℕ) : Prop :=
-  ∀ (s : Finset ℕ), s.Nonempty → s.card = k → ∃ᵉ (n > 0) (d > 0), n.gcd d = 1 →
+  ∀ (s : Finset ℕ), s.card = k → ∃ᵉ (n > 0) (d > 0), n.gcd d = 1 →
     Set.IsAPOfLengthWith s k n d → ∀ q, ∏ i ∈ s, i ≠ q ^ l
 
 /--
--- a/FormalConjectures/ForMathlib/Combinatorics/AP/Basic.lean
+++ b/FormalConjectures/ForMathlib/Combinatorics/AP/Basic.lean
@@ -134,13 +134,17 @@ theorem Set.not_isAPOfLength_empty {l : ℕ∞} (hl : 0 < l) :
 /-- We say that a set `s` is free of arithmetic progressions of length `l` if `s` contains no
 non-trivial arithmetic progressions of length `l`. Written as `Set.IsAPOfLengthFree s l`. --/
 def Set.IsAPOfLengthFree (s : Set α) (l : ℕ∞) : Prop :=
-  ∀ t ⊆ s, t.IsAPOfLength l → t.ncard = 1
+  ∀ t ⊆ s, t.IsAPOfLength l → l ≤ 1
 
 /-- Any set is free of arithmetic progressions of length `1`, because such APs are all trivial. -/
-theorem Set.IsAPOfLengthFree_one (s : Set α) : s.IsAPOfLengthFree 1 := by
+theorem Set.isAPOfLengthFree_one (s : Set α) : s.IsAPOfLengthFree 1 := by
   simp [Set.IsAPOfLengthFree]
 
-/-- Every set contains an arithmetic progression of length zero. -/
-@[simp]
-theorem Set.not_isAPOfLengthFree_zero (s : Set α) : ¬s.IsAPOfLengthFree 0 := by
-  simpa [Set.IsAPOfLengthFree] using fun x ↦ by simp [Ne.symm]
+/-- Any set is free of arithmetic progressions of length `0`, because such APs are all trivial. -/
+theorem Set.isAPOfLengthFree_zero (s : Set α) : s.IsAPOfLengthFree 0 := by
+  simp [Set.IsAPOfLengthFree]
+
+/-- Any non-trivial arithmetic progression cannot be free of arithmetic progressions. -/
+theorem Set.IsAPOfLength.not_isAPOfLengthFree {s : Set α} {l : ℕ∞}
+    (hs : s.IsAPOfLength l) (hl : 1 < l) : ¬s.IsAPOfLengthFree l := by
+  simpa [Set.IsAPOfLengthFree] using ⟨s, le_rfl, ⟨hs, hl⟩⟩
--- a/FormalConjectures/Wikipedia/GaussCircleProblem.lean
+++ b/FormalConjectures/Wikipedia/GaussCircleProblem.lean
@@ -63,7 +63,7 @@ $$
 [Ha59]  Hardy, G. H. (1959). _Ramanujan: Twelve Lectures on Subjects Suggested by His Life and Work_(3rd ed.). New York: Chelsea Publishing Company. p. 67
 -/
 @[category research solved, AMS 11]
-theorem error_le (r : ℝ) (hr : 0 ≤ r) : |E r| ≤ 2 * √2 * π * r := by
+theorem error_le (r : ℝ) (hr : 0 < r) : |E r| ≤ 2 * √2 * π * r := by
   sorry
 
 /--
@@ -73,8 +73,7 @@ $$
 $$
 -/
 @[category research solved, AMS 11]
-theorem error_not_isLittleO (r : ℝ) (hr : 0 ≤ r) :
-    ¬E =o[atTop] (fun r => √r * √√r.log) := by
+theorem error_not_isLittleO : ¬E =o[atTop] (fun r => √r * √√r.log) := by
   sorry
 
 /--
@@ -101,7 +100,7 @@ $$
 -/
 @[category research solved, AMS 11]
 theorem exact_form_floor (r : ℝ) (hr : 0 ≤ r) :
-    N r = 1 + 4 * ∑' i, (⌊r ^ 2 / (4 * i + 1)⌋ - ⌊r ^ 2 / (4 * i + 3)⌋) := by
+    N r = 1 + 4 * ∑' (i : ℕ), (⌊r ^ 2 / (4 * i + 1)⌋ - ⌊r ^ 2 / (4 * i + 3)⌋) := by
   sorry
 
 end GaussCircleProblem
--- a/FormalConjectures/Wikipedia/LehmerMahlerMeasureProblem.lean
+++ b/FormalConjectures/Wikipedia/LehmerMahlerMeasureProblem.lean
@@ -65,14 +65,15 @@ theorem lehmer_mahler_measure_problem.variants.not_reciprocal (f : ℤ[X])
     mahlerMeasureZ f ≥ mahlerMeasureZ (X^3 - X - 1) := by
   sorry
 
-def AllCoeffsOdd (f : Polynomial ℤ) : Prop :=
-  ∀ i : ℕ, Odd (f.coeff i)
+/-- `Polynomial.HasOddCoeffs f` means that all coefficients of `f : Polynomial ℤ` are odd. -/
+def Polynomial.HasOddCoeffs (f : Polynomial ℤ) : Prop :=
+  ∀ i ∈ f.support, Odd (f.coeff i)
 
 /--
 If all the coefficients of `f` are odd and `M(f)>1`, `M(f) ≥ M(X^2 - X - 1)`.
 -/
 @[category research solved, AMS 11]
 theorem lehmer_mahler_measure_problem.variants.odd (f : ℤ[X])
-    (hf : mahlerMeasureZ f > 1) (hf' : AllCoeffsOdd f) :
+    (hf : mahlerMeasureZ f > 1) (hf' : f.HasOddCoeffs) :
     mahlerMeasureZ f ≥ mahlerMeasureZ (X^2 - X - 1) := by
   sorry


jcreedcmu:
This is my first contribution to formal-conjectures. Thanks for the
clear contribution guidelines! Let me know if I missed anything.

I hope the definition of `IsRupert` is appropriate to simply have in the conjecture file.
I couldn't think of any trivial API theorems off the top of my head. Showing that
particular convex bodies are Rupert is possible, but a little involved.
(cf. the proofs for the [Triakis Tetrahedron](https://github.com/dwrensha/Rupert.lean/blob/main/Rupert/TriakisTetrahedron.lean), [Cube](https://github.com/dwrensha/Rupert.lean/blob/main/Rupert/Cube.lean))

I'd love to get @dwrensha's feedback on this PR as well.

FormalConjectures/Paper/Rupert.lean
```diff
@@ -0,0 +1,98 @@
...
+The projection of a vector from 3-space to 2-space by dropping the third coordinate.
+-/
+def proj_xy (v : Fin 3 → ℝ) : Fin 2 → ℝ :=
+  ![v 0, v 1]
+
+/--
+A convex polyhedron (given as a finite collection of vertices) is Rupert if
+there are two rotations in ℝ³ (called "inner" and "outer") and a translation in ℝ²
+such that the "inner shadow" (the projection to ℝ² of the inner rotation applied
+to the polyhedron, then translated) fits in the interior of the "outer shadow"
```
dwrensha, line 84: It would be good to note which topology this "interior" is with respect to.

dwrensha: (My understanding is that it's a topology that's different from but equivalent to the one we are currently using in Rupert.lean: https://github.com/dwrensha/Rupert.lean/blob/60dfda2137da215843d91c2af982bb3e68368102/Rupert/Basic.lean#L16-L22)

jcreedcmu: My understanding is that it's mathematically exactly the same topology (i.e. the standard topology on R^2) but maybe the way it arises from the lean implementation details might be a little different. (I think what's going on here is we're getting the product topology, and I think in the `Rupert.lean` development you link we were getting the metric topology on the L2 metric) If you think it's worth documenting the implementation details here, I could be persuaded.

dwrensha: Thanks for the explanation. I am persuaded that this is good, and I think we probably should update Rupert.lean to use the same definition.

dwrensha: (by Rupert.lean I mean https://github.com/dwrensha/Rupert.lean, not this file. :facepalm: )

jcreedcmu: Yeah, seems like a good idea! (to update the other repo)

jcreedcmu:
Oops, @jsm28 just observed on [#general > Rupert.lean](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Rupert.2Elean/with/526580884) a small but fatal error that makes the formal version of the conjecture uninteresting; I forgot "interior" when checking nonempty interior of convex closure.

He also suggested several generalizations of the definition of IsRupert that I think I might like to incorporate.

jcreedcmu:
I fixed the above-mentioned error.

I also investigated generalizing this beyond the conjecture as stated in the literature; I think it might be a little bit of work. My current preference is doing that work independently of this PR, as a mathlib contribution. Whenever that lands, and whenever formal-conjectures catches up to upstream, then I'd be interested in revisiting this to refer to the more general definition.

In other words, I'm curious what reviewers think about landing this PR as it is.

dwrensha (approved):
Looks good to me.

FormalConjectures/Paper/Rupert.lean
```diff
@@ -0,0 +1,98 @@
...
+(the projection to ℝ² of the outer rotation applied to the polyhedron)
+-/
+def IsRupert {ι : Type} [Fintype ι] (vertices : ι → ℝ³) : Prop :=
+   ∃ inner_rotation ∈ SO3, ∃ inner_offset : ℝ², ∃ outer_rotation ∈ SO3,
+   let hull := convexHull ℝ { vertices i | i }
+   let inner_shadow := { inner_offset + proj_xy (inner_rotation *ᵥ p) | p ∈ hull }
+   let outer_shadow := { proj_xy (outer_rotation *ᵥ p) | p ∈ hull }
```
callesonne: I think it would be nice to have these as separate definitions as well, so that theorems can be proven about these things specifically.

FormalConjectures/Paper/Rupert.lean
```diff
@@ -0,0 +1,98 @@
...
+   inner_shadow ⊆ interior outer_shadow
+
+/--
+Does the Rupert property hold of every convex polyhedron with nonempty interior?
```
callesonne:
```suggestion
Does the Rupert property hold for every convex polyhedron with nonempty interior?
```
Should this be for?

callesonne (commented):
Thanks a lot for this contribution! It would be great to get this merged :)

FormalConjectures/Paper/Rupert.lean
jcreedcmu: Ok! Did you have in mind hoisting the transformations as a single function like in the commit I just added?

FormalConjectures/Paper/Rupert.lean
jcreedcmu: Sure, if you want! I changed it to 'for'. A property holding of a subject and a property holding for a subject both sound acceptable to my mathematical-English ear.

jcreedcmu: (e.g. I find https://groups.seas.harvard.edu/courses/cs152/2023sp/lectures/lec03-inductive-proof.pdf which uses both words seemingly interchangeably in various sentences)

callesonne: Ah okay, I should also say I'm not an english native speaker, so do what you prefer then!

FormalConjectures/Paper/Rupert.lean
callesonne: Yes I think adding this function is great, but I was thinking of having separate definitions for `inner_shadow` and `outer_shadow` as well. These would take `vertices` as an argument.

jcreedcmu: Ok, how about this?

FormalConjectures/Paper/Rupert.lean
```diff
@@ -0,0 +1,120 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+import Mathlib.Analysis.InnerProductSpace.PiL2
+/-!
```
callesonne:
```suggestion
import Mathlib.Analysis.InnerProductSpace.PiL2

/-!
```

FormalConjectures/Paper/Rupert.lean
```diff
@@ -0,0 +1,120 @@
...
+-/
+def proj_xy (v : Fin 3 → ℝ) : Fin 2 → ℝ :=
+  ![v 0, v 1]
+
+/--
+The result of transforming a subset of ℝ³ by a chosen rotation and offset,
+and then projected to ℝ².
+--/
+def transformed_shadow (X : Set ℝ³) (offset : ℝ²) (rotation : SO3) : Set ℝ² :=
+  (λ p ↦ offset + proj_xy (rotation.1 *ᵥ p)) '' X
```
callesonne:
```suggestion
/--
The result of transforming a subset of ℝ³ by a chosen rotation and offset,
and then projected to ℝ².
--/
def transformed_shadow (X : Set ℝ³) (offset : ℝ²) (inner_rotation : SO3) : Set ℝ² :=
  (fun p ↦ offset + (inner_rotation *ᵥ p) ∘ Fin.castSucc) '' X
```
This changes 3 things:
1. I think using `Fin.castSucc` over an explicit def for `proj_xy` is better.
2.  Using `fun` over lambda is more standard.
3. The `.1` on inner_rotation is unnecessary since it gets coerced automatically.

FormalConjectures/Paper/Rupert.lean
```diff
@@ -0,0 +1,120 @@
...
+
+/--
+Get the "inner shadow" of a set of vertices, given the rotation and offset for it.
+--/
+def inner_shadow_of_vertices {ι : Type} [Fintype ι] (vertices : ι → ℝ³)
```
callesonne: I think its better if you make `vertices` a `Finset`, this makes it easier to use `convexHull`.

callesonne (commented):
Some more comments, I think it's almost ready now!

FormalConjectures/Paper/Rupert.lean
jcreedcmu: Good point, didn't know that was there! Thanks. Doing that seemed to require an extra type ascription in `is_every_convex_polyhedron_rupert` to help typeclass instantiation find the right topology for `interior`, so I added that.

I made most of the auxiliary definitions actually just take `Set ℝ³` instead of `Finset` because they make as much sense for arbitrary subsets as finite subsets. However, I deliberately left `IsRupert` requiring `Finset ℝ³`, and expanded the doc comment to explain why.

Basically, I hope my judgment is correct in supposing it's appropriate for this repository to stick closely to what's already in the published literature. Keeping `IsRupert` and the conjecture itself requiring `Finset ℝ³` is the closest to the definitions I could find, which are restricted to only compact polyhedra.

FormalConjectures/Paper/Rupert.lean
jcreedcmu: Good fixes, done! I kept the name `rotation` instead of `inner_rotation` because this helper is called by both `inner` and `outer`.

FormalConjectures/Paper/Rupert.lean
jcreedcmu: Done.

jcreedcmu:
Appreciate the review, let me know what you think now!

FormalConjectures/Paper/Rupert.lean
callesonne: Perfect, yes my change to `inner_rotation` was a mistake, I was experimenting with something else also and must have forgotten what you originally called it :)

FormalConjectures/Paper/Rupert.lean
callesonne: Perfect!

FormalConjectures/Paper/Rupert.lean
```diff
@@ -0,0 +1,124 @@
...
+  transformed_shadow (convexHull ℝ vertices) inner_offset inner_rotation
+
+/--
+Get the "outer shadow" of a set of vertices, given the rotation and offset for it.
+The outer shadow asymmetrically doesn't require a translation, since it would be
+redundant to specify two, since translation commutes with projection.
+--/
+def outer_shadow_of_vertices (vertices : Set ℝ³)
+  (outer_rotation : SO3) : Set ℝ² :=
+  transformed_shadow (convexHull ℝ vertices) 0 outer_rotation
```
callesonne: Now that I'm looking at this again, and everything has been cleaned up, I think actually only `transformed_shadow` should be a separate definition. As the others are just simple applications of it.

FormalConjectures/Paper/Rupert.lean
```diff
@@ -0,0 +1,124 @@
...
+  known results for Platonic, Archimedean, and Catalan solids.
+
+* This problem's name comes from the fact that it is a generalization
+  of [Prince Rupert's Cube](https://en.wikipedia.org/wiki/Prince_Rupert%27s_cube).
+
+-/
+open scoped Matrix
+
+notation "ℝ³" => Fin 3 → ℝ
+notation "ℝ²" => Fin 2 → ℝ
```
callesonne: I am a bit worried about these being outside a namespace, since it is likely that the notation `ℝ³` can be defined for other things in Mathlib (or here). Could you maybe put most of the definitions / theorems below in a `Rupert` namespace (except maybe `SO3`) and make these notations scoped to that namespace?

callesonne (approved):
LGTM modulo these final comments, thanks a lot again!

FormalConjectures/Paper/Rupert.lean
eric-wieser: Note we already have

https://github.com/google-deepmind/formal-conjectures/blob/8a69bc65453fc895d78fd9735f5fe10f3f5ec02a/FormalConjectures/ForMathlib/Geometry/2d.lean#L23

which will conflict.

jcreedcmu: I like the idea of namespacing to `Rupert` generally. About `ℝ²`/`ℝ³` I'm not sure what's best... the two issues I'm aware of are:

(1) `EuclideanSpace ℝ (Fin n)` vs. `Fin n → ℝ`. These are definitionally equal but mixing the two leads to needing to coerce to avoid missing, e.g. HAdd instances, which I added to `transformed_shadow`. If you already have `ℝ² = EuclideanSpace ℝ (Fin 2)` I'm inclined to match that convention.
(2) Where to put ℝ³? I didn't think it was worth adding a whole `3d.lean` file so I just added the definition scoped to `EuclideanGeometry`. Is that ok for now?

FormalConjectures/Paper/Rupert.lean
jcreedcmu: Ok, done! that was my original feeling, and I'm certainly happy to go with that :)

FormalConjectures/Paper/Rupert.lean
```diff
@@ -0,0 +1,112 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+import FormalConjectures.ForMathlib.Geometry.«2d»
```
callesonne:
```suggestion
```
This import is included in `ProblemImports`

FormalConjectures/Paper/Rupert.lean
```diff
@@ -0,0 +1,112 @@
...
+scoped[EuclideanGeometry] notation "ℝ³" => EuclideanSpace ℝ (Fin 3)
+
+namespace Rupert
+
+/--
+The result of transforming a subset of ℝ³ by a chosen rotation and offset,
+and then projected to ℝ².
+--/
+def transformed_shadow (X : Set ℝ³) (offset : ℝ²) (rotation : SO3) : Set ℝ² :=
+  let offset : Fin 2 → ℝ := offset
```
callesonne:
```suggestion
  letI offset : Fin 2 → ℝ := offset
```
so that this automatically gets replaced by `offset` (the previous one) whenever this definition is used.

callesonne (commented):
Some more small comments :) I actually prefer it without the coercions, and wouldnt mind having another notation for `ℝ²` for `Fin 2 → ℝ` as you had it before but in the Rupert namespace.

FormalConjectures/Paper/Rupert.lean
jcreedcmu: Conveniently, if ℝ² and ℝ³ are Fin 2 → ℝ and Fin 3 → ℝ, then this line isn't even needed.

jcreedcmu:
Terrific, then I'm perfectly happy leaving ℝ² and ℝ³ as namespace-local definitions here if that's ok with you all. The inner product structure of `EuclideanSpace` isn't relevant except for the meaning of the rotation group SO(3), and that's nicely compartmentalized in that definition.

FormalConjectures/Paper/Rupert.lean
```diff
@@ -0,0 +1,110 @@
...
+
+-/
+open scoped Matrix
+
+abbrev SO3 := Matrix.specialOrthogonalGroup (Fin 3) ℝ
+
+namespace Rupert
+
+notation "ℝ²" => Fin 2 → ℝ
+notation "ℝ³" => Fin 3 → ℝ
```
callesonne:
```suggestion
scoped notation "ℝ²" => Fin 2 → ℝ
scoped notation "ℝ³" => Fin 3 → ℝ
```

callesonne (approved):
Looks good to me now! (apart from this final comment)

jcreedcmu:
sgtm, thanks! I'm still learning about lean's module/namespacing features, thanks for your patience :)

PR #285, opened 2025-06-30
feat: Add 10 Graph Theory Conjectures

henrykmichalewski:
This commit introduces 10 formal conjectures (5 open, 5 resolved) from the [WoWII](http://cms.dt.uh.edu/faculty/delavinae/research/wowII/) database.

To support definitions not yet in the project's version of Mathlib, newer versions of `Bipartite.lean` and `Diam.lean` have been added to the `ForMathlib` directory. A few proofs in these files were replaced with `sorry` to avoid build issues.


henrykmichalewski:
@b-mehta - the first ten wowii graph conjectures in the chronological order

lakefile.lean
mo271, line 1: Is the only reason to have this to compile faster?
I think for now we haven't added any auxiliary libraries, because the build time for the entire repo is still small enough, especially when caching. 
I guess it doesn't hurt, but then perhaps we might also consider doing the same for other parts of the repo, like ErdosProblems.

FormalConjectures/GraphConjectures/README.md
mo271, line 1: I'm not sure how useful this file is. It basically contains only information also in the individual files and it will probably a pain to keep it in sync, when 
 - we change the doc string
 - change the status (from open to resolved)
 - add new conjectures
 Also it is probably better to get an overview directly on the WOWII site. I propose to remove the entire file.

FormalConjectures/GraphConjectures/Imports.lean
mo271, line 1: Instead of having this special import file, I propose to remove it an instead `import FormalConjectures.Util.ProblemImports` in the individual files.

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/Bipartite.lean
```diff
@@ -0,0 +1,285 @@
...
+  of `G`, instead `s ∪ t` is only required to cover the *support* of `G`, that is, the vertices
+  that form edges in `G`. This definition is equivalent to the expected definition. If `s` and `t`
+  do not cover all the vertices, one recovers a covering of all the vertices by unioning the
+  missing vertices `(s ∪ t)ᶜ` to either `s` or `t`.
+
+* `SimpleGraph.IsBipartite`: Predicate for a simple graph to be bipartite.
+  `G.IsBipartite` is defined as an abbreviation for `G.Colorable 2`.
+
+* `SimpleGraph.isBipartite_iff_exists_isBipartiteWith` is the proof that `G.IsBipartite` iff
+  `G.IsBipartiteWith s t`.
```
Paul-Lez, line 45:
```suggestion
  `G.IsBipartiteWith s t` for some `s` and `t`.
```

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
```diff
@@ -0,0 +1,90 @@
...
+
+open Classical
+
+variable {α : Type*} [Fintype α] [DecidableEq α]
+
+/-- `Ls G` is the maximum number of leaves over all spanning trees of `G`.
+It is defined to be `0` when `G` is not connected. -/
+noncomputable def Ls (G : SimpleGraph α) [DecidableRel G.Adj] : ℝ :=
+  if h : G.Connected then
+    let spanningTrees := { T : Subgraph G | T.IsSpanning ∧ IsTree T.coe }
```
Paul-Lez: I think using `let` to make the definition more readable is certainly useful. I personally prefer using `letI`, which automatically inlines the quantity, which makes using such definitions easier.

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
```diff
@@ -0,0 +1,90 @@
...
+namespace SimpleGraph
+
+open Classical
+
+variable {α : Type*} [Fintype α] [DecidableEq α]
+
+/-- `Ls G` is the maximum number of leaves over all spanning trees of `G`.
+It is defined to be `0` when `G` is not connected. -/
+noncomputable def Ls (G : SimpleGraph α) [DecidableRel G.Adj] : ℝ :=
+  if h : G.Connected then
```
Paul-Lez: I think you don't need to do a case disjunction here: when `G` is disconnected, the set `spanningTrees` is empty so the resulting supremum is `0`

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
```diff
@@ -0,0 +1,90 @@
...
+
+/-- `m G` is the size of a maximum matching of `G`. -/
+noncomputable def m (G : SimpleGraph α) [DecidableRel G.Adj] : ℝ :=
+  let matchings := { M : Subgraph G | M.IsMatching }
+  sSup (Set.image (fun M => (M.edgeSet.toFinset.card : ℝ)) matchings)
+
+/-- `a G` is a quantity defined via independent sets used in Conjecture 6. -/
+noncomputable def a (G : SimpleGraph α) [DecidableRel G.Adj] : ℝ :=
+  let independent_sets := { S : Finset α | IsIndepSet G S }
+  if h_indep : independent_sets.Nonempty then
```
Paul-Lez: Here too you can remove some of the case splitting, which will make the definition more usable. You would then be able to state theorems that record what occurs for "junk" values.

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
```diff
@@ -0,0 +1,90 @@
...
+  (largestInducedForestSize G : ℝ)
+
+/-- `largestInducedBipartiteSubgraphSize G` is the size of a largest induced
+bipartite subgraph of `G`. -/
+noncomputable def largestInducedBipartiteSubgraphSize (G : SimpleGraph α) : ℕ :=
+  sSup { n | ∃ s : Finset α, (G.induce s).IsBipartite ∧ s.card = n }
+
+/-- `b G` is the number of vertices of a largest induced bipartite subgraph of `G`.
+Returned as a real number. -/
+noncomputable def b (G : SimpleGraph α) : ℝ :=
```
Paul-Lez, line 75: For such definitions that are meant to be used as notation for stating the conjectures later, I would be tempted to used `scoped notation`. For example, this one would become 
```lean-4
/-- `b G` is the number of vertices of a largest induced bipartite subgraph of `G`.
Returned as a real number. -/
scoped notation "b" G:arg => (largestInducedBipartiteSubgraphSize G : ℝ)
```

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Domination.lean
```diff
@@ -0,0 +1,145 @@
...
+@[mk_iff]
+structure IsNDominatingSet (n : ℕ) (D : Finset α) : Prop where
+  isDominating : G.IsDominating D
+  card_eq : D.card = n
+
+/-! ### Domination number -/
+
+/-- The domination number of a graph `G` is the minimum size of a dominating
+set. It is `0` if there are no vertices. -/
+noncomputable def dominationNumber (G : SimpleGraph α) : ℕ :=
```
Paul-Lez, line 58: To avoid having too many structures, I would suggest simply inlining the properties of `IsNDominatingSet` in `dominationNumber`, e.g. 
`sInf {n | ∃ D : Finset α, G.IsDominating D ∧ D.card = n}`
and so on with other definitions.

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Invariants.lean
```diff
@@ -0,0 +1,144 @@
...
+noncomputable abbrev l_avg (G : SimpleGraph α) : ℝ := averageIndepNeighbors G
+
+/-- Independent domination number of `G`. -/
+noncomputable def gi (G : SimpleGraph α) : ℕ := G.indepDominationNumber
+
+/-- `temp_v G v = deg(v)/(n(G) - deg(v))`. -/
+noncomputable def temp_v (G : SimpleGraph α) [DecidableRel G.Adj] (v : α) : ℝ :=
+  let n := Fintype.card α
+  let deg := G.degree v
+  if n = deg then 0 else (deg : ℝ) / ((n : ℝ) - (deg : ℝ))
```
Paul-Lez, line 39: The `if else` here is redundant since division by `0` is `0` in the (Lean) Real numbers:)

FormalConjectures/GraphConjectures/GraphConjecture19.lean
```diff
@@ -0,0 +1,39 @@
...
+variable {α : Type*} [Fintype α] [DecidableEq α]
+
+/--
+WOWII Conjecture 19 (open):
+If `G` is connected then the size `b(G)` of a largest induced bipartite subgraph
+satisfies
+`b(G) ≥ FLOOR((∑ ecc(v))/(|V|) + sSup (range (l G)))`, where `ecc(v)` denotes
+eccentricity and `l(G)` is the independence number of neighbourhoods.
+-/
+-- WOWII conjecture: If G is a simple     connected graph, then b(G) ≥ FLOOR(average of ecc(v)+maximum of l (v))
```
Paul-Lez: Since the docstring above already contains the informal statement
```suggestion
```

FormalConjectures/GraphConjectures/GraphConjecture34.lean
```diff
@@ -0,0 +1,38 @@
...
+variable {α : Type*} [Fintype α] [DecidableEq α]
+
+/--
+WOWII Conjecture 34 (open):
+Let `path(G)` be the floor of the average distance of a connected graph `G`.
+Then
+`path(G) ≥ ceil( distavg(G, center) + distavg(G, maxEccentricityVertices G) )`.
+-/
+theorem conjecture34 [Nonempty α] (G : SimpleGraph α) (h_conn : G.Connected) :
+  (path G : ℤ) ≥ Int.ceil (distavg G (graphCenter G) + distavg G (maxEccentricityVertices G)) := by
```
Paul-Lez: The convention for inequalities is to write the smaller side on the left when sensible
```suggestion
    Int.ceil (distavg G (graphCenter G) + distavg G (maxEccentricityVertices G)) ≤ path G := by
```

FormalConjectures/GraphConjectures/GraphConjecture2.lean
```diff
@@ -0,0 +1,39 @@
...
+/--
+WOWII Conjecture 2 (open):
+For a simple connected graph `G`,
+`Ls(G) ≥ 2 · (l(G) - 1)` where `l(G)` is the average independence number of
+the neighbourhoods of the vertices of `G`.
+
+WOWII conjecture: If G is a simple   connected graph, then L s (G) ≥  2(average of l (v)   - 1)
+-/
+theorem conjecture2 (G : SimpleGraph α) (h : G.Connected) :
+  Ls G ≥ 2 * (l G - 1) := by sorry
```
Paul-Lez:
```suggestion
    2 * (l G - 1) ≤ Ls G := by
  sorry
```

FormalConjectures/GraphConjectures/GraphConjecture19.lean
```diff
@@ -0,0 +1,39 @@
...
+theorem conjecture19 (G : SimpleGraph α) [Nonempty α] (h_conn : G.Connected) :
+    b G ≥ FLOOR ((∑ v ∈ Finset.univ, ecc G v) / (Fintype.card α : ℝ) + (sSup (Set.range (indepNeighbors G)))) := by sorry
```
Paul-Lez:
```suggestion
    b G ≥ FLOOR ((∑ v, ecc G v) / (Fintype.card α : ℝ) + (sSup (Set.range (indepNeighbors G)))) := by sorry
```

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Invariants.lean
```diff
@@ -0,0 +1,144 @@
...
+
+/-- Maximum of `temp_v` over all vertices. -/
+noncomputable def max_temp (G : SimpleGraph α) [DecidableRel G.Adj] [Nonempty α] : ℝ :=
```
Paul-Lez: I think a name like `MaxTemp` would be a little more idiomatic - [here](https://leanprover-community.github.io/contribute/naming.html) are the Mathlib guidelines for this sort of stuff

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/Bipartite.lean
henrykmichalewski: This is imported from a newer version of Mathlib - should we change or keep as is?

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
henrykmichalewski: Done.

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
henrykmichalewski: On a closer look I refactored a and aprime - a is now very simple and aprime is very similar to the definition above. Let me know if the new version of aprime satisfied the requirements.

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Invariants.lean
henrykmichalewski: Done.

FormalConjectures/GraphConjectures/GraphConjecture19.lean
henrykmichalewski: Done.

FormalConjectures/GraphConjectures/GraphConjecture2.lean
henrykmichalewski: Done.

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Invariants.lean
henrykmichalewski: Done.

FormalConjectures/GraphConjectures/GraphConjecture19.lean
henrykmichalewski: Done.

FormalConjectures/GraphConjectures/GraphConjecture34.lean
henrykmichalewski: Done.

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
henrykmichalewski: Let me know if the current refactoring of Ls is satisfactory.

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
henrykmichalewski: This specific change explodes on GraphConjecture58.lean which looks like this

```
theorem conjecture58 (hG : G.Connected) :
  (G.f : ℝ) ≥ Nat.ceil ((G.b : ℝ) / G.l_avg) := by sorry
```
Should we fix it or leave for later?

henrykmichalewski:
#> * For the docstrings it would be nice to provide links to the WOWII site (I see there might be problem, since they don't have site for each individual links, but still a link would be good, even if one then has to scroll)

Done.

> * I think a better name name for the new directory for the problem collection  instead of "GraphConjectures" would be "WrittenOnTheWallII".

Done.

> * Instead of recording if the problem is open or resolved in the docstring, consider using `@[category research [solved|open], AMS 5]` (AMS 5 is "combinatorics, which I suppose is a good fit for most problems, since graph theory doesn't have its own subject number..)

Done.

lakefile.lean
henrykmichalewski: Let me know whether we should keep it or remove (I am keeping it for now).

FormalConjectures/GraphConjectures/README.md
henrykmichalewski: Good point. I need it only for coordination during the development process (I target 40 problems, 5-10 resolved, 30-35 open). What about keeping this README until I reach the target and then I will delete it?

FormalConjectures/GraphConjectures/Imports.lean
henrykmichalewski: The refactoring of imports looks as follows:

* The dedicated import file is deleted
* I need some single specialized imports not covered by `ProblemImports` - I am importing them in each file. 

Let me know if this meets the requirements.

lakefile.lean
mo271: Let's remove it

FormalConjectures/GraphConjectures/README.md
mo271: Sounds good, perhaps put a line `TODO: delete this file when...` or something similar so it is not forgotten?

FormalConjectures/GraphConjectures/Imports.lean
mo271: Sounds good, you could also consider adding the single specialized to https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/Util/ForMathlib.lean if they are in `ForMathlib`, this way they will be imported automatically when `ProblemImports` are imported.

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/Bipartite.lean
mo271: In this case I would say keep as is, because this will be deleted anyhow when we bump the version. If @Paul-Lez deems the changes important enough, he can upstream them...

lakefile.lean
henrykmichalewski: Done

FormalConjectures/GraphConjectures/README.md
henrykmichalewski: Done

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
Paul-Lez: LGTM!

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
Paul-Lez: Hm in that case let's leave that to later.

Paul-Lez (approved):

Issue #27, opened 2025-05-27
Sidon sets


b-mehta:
The formulation of Sidon sets currently in the repo is not right for mathlib:

- It works with sequences rather than sets. Historically in combinatorial number theory (as it was called then), infinite sets of naturals were usually expressed as strictly monotone sequences, but the modern formulation is to use infinite sets directly instead, and so the mathlib version should do the latter too.
- As a result, it requires a Sidon set to be infinite, and so is not a useful formulation for cases like ZMod n or other finite abelian groups, which are cases studied in the literature.

These are not *correctness* problems however: the problem statements using Sidon sets are taking in a valid assumption.

b-mehta:
If the GDM team agrees this should be changed here then I'm happy to work on it. Alternatively, since the current Erdos problems set works fine with this definition, I'm happy just to put the improved definition in mathlib, then this repo can be adjusted after the fact. Your call!

mo271:
I agree this should be changed, handling the increased sequences is a bit awkward, infinite sets sounds better. 

Both options sound good: fixing the improved definition in `ForMathlib` of this repo first or in parallel of upstreaming to mathlib is perhaps a good idea, since we don't get the benefits upstream changes to mathlib immediately, because we follow stable versions. 
Thanks for working on it, @b-mehta!

eric-wieser:
> since we don't get the benefits upstream changes to mathlib immediately, because we follow stable versions.

Interpreting this another way; you have three days to upstream this before the next stable mathlib release is cut!

PR #286, opened 2025-06-30
Erdos Problem 44

### LINKED ISSUE #222: Erdős problem 44: extending Sidon sets
### What is the conjecture
    
https://www.erdosproblems.com/44

  Let $N\geq 1$ and $A\subset \{1,\ldots,N\}$ be a Sidon set. Is it true that, for any $\epsilon>0$, there exist $M=M(\epsilon)$ and $B\subset \{N+1,\ldots,M\}$ such that $A\cup B\subset \{1,\ldots,M\}$ is a Sidon set of size at least $(1-\epsilon)M^{1/2}$?

Status: open

### Prerequisites needed
None, Sidon sets are already in `ForMathlib`

### Choose either option

saurabhraghuvanshii:
closes #222

### DIFF
--- a/FormalConjectures/ErdosProblems/44.lean
+++ b/FormalConjectures/ErdosProblems/44.lean
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    https://www.apache.org/licenses/LICENSE/2.0
+    https://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
@@ -29,7 +29,7 @@ noncomputable def maxSidonSetSize (N : ℕ) : ℕ :=
   sSup {(A.card) | (A : Finset ℕ) (_ : A ⊆ Finset.Icc 1 N) (_ : IsSidon A.toSet)}
 
 /-- The maximum size of a Sidon set in `{1, ..., N}` is less than or equal to `√N + 1`. -/
-@[category API, AMS 5 11]
+@[category undergraduate, AMS 5 11]
 theorem maxSidonSetSize_bound (N : ℕ) (hN : 1 ≤ N) :
     maxSidonSetSize N ≤ N.sqrt + 1 := by
   sorry
@@ -42,7 +42,7 @@ of size at least `(1−ε)M^{1/2}`?
 This problem asks whether any Sidon set can be extended to achieve a density
 arbitrarily close to the optimal density for Sidon sets.
 -/
-@[category research open, AMS 5 11]
+@[category undergraduate, AMS 5 11]
 theorem erdos_44 : (∀ᵉ (N ≥ (1 : ℕ)) (A ⊆ Finset.Icc 1 N), IsSidon A.toSet →
     ∀ᵉ (ε > (0 : ℝ)), ∃ᵉ (M > N) (B ⊆ Finset.Icc (N + 1) M),
       IsSidon (A ∪ B).toSet ∧ (1 - ε) * Real.sqrt M ≤ (A ∪ B).card) ↔ answer(sorry) := by
@@ -51,7 +51,7 @@ theorem erdos_44 : (∀ᵉ (N ≥ (1 : ℕ)) (A ⊆ Finset.Icc 1 N), IsSidon A.t
 /--
 A variant considering the extension to any larger range.
 -/
-@[category research open, AMS 5 11]
+@[category undergraduate, AMS 5 11]
 theorem erdos_44.variant : (∀ᵉ (N ≥ (1 : ℕ)) (A ⊆ Finset.Icc 1 N), IsSidon A.toSet →
     ∀ᵉ (ε > (0 : ℝ)) (M ≥ N), ∃ᵉ (B ⊆ Finset.Icc (N + 1) M),
       IsSidon (A ∪ B).toSet ∧ (1 - ε) * Real.sqrt M ≤ (A ∪ B).card) ↔ answer(sorry) := by


mo271:
I think it would be easier to review when this pull request is split into three, one for each problem. Also, as a first step, it might be a good idea to get the issue #27 done, because then we don't need to add new ad-hoc definition for the Sidon sets.

mo271:
After #287 is merged, you rebase or merge and then use `IsSidon A` for `A : Set ℕ` and `IsSidon A.toSet` for `A: Finset ℕ`, no need to redefine it. (Same for #288 and #289).

FormalConjectures/ErdosProblems/44.lean
```diff
@@ -0,0 +1,114 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+import FormalConjectures.ForMathlib.Combinatorics.Basic
```
callesonne:
```suggestion
```
Is this import really needed?

FormalConjectures/ErdosProblems/44.lean
```diff
@@ -0,0 +1,114 @@
...
+
+This problem asks whether any Sidon set can be extended to achieve a density
+arbitrarily close to the optimal density for Sidon sets.
+-/
+
+open Function Set
+
+/-- The maximum size of a Sidon set in `{1, ..., N}`. -/
+noncomputable def maxSidonSetSize (N : ℕ) : ℕ :=
+  sSup {A.card | (A : Finset ℕ) (_ : A ⊆ Finset.Icc 1 N) (_ : IsSidon A.toSet)}
```
callesonne:
```suggestion
  sSup {(A.card) | (A : Finset ℕ) (_ : A ⊆ Finset.Icc 1 N) (_ : IsSidon A.toSet)}
```
I think maybe you need this to not get errors?

FormalConjectures/ErdosProblems/44.lean
```diff
@@ -0,0 +1,114 @@
...
+
+/-- The optimal density for Sidon sets is approximately `√N`. -/
+@[category API, AMS 5 11]
+theorem maxSidonSetSize_bound (N : ℕ) (hN : 1 ≤ N) :
+    maxSidonSetSize N ≤ N.sqrt + 1 := by
+  sorry
+
+/--
+A Sidon set can be extended to achieve density close to optimal.
```
callesonne: I would reword this closer to how it is written on the website. In particular it should be phrased as a question rather than an assertion (especially since you are using the `answer(sorry)` formulation.

FormalConjectures/ErdosProblems/44.lean
```diff
@@ -0,0 +1,114 @@
...
+-/
+@[category research open, AMS 5 11]
+theorem erdos_44 : (∀ (N : ℕ) (A : Finset ℕ) (_ : 1 ≤ N) (_ : A ⊆ Finset.Icc 1 N) (_ : IsSidon A.toSet),
+    ∀ ε > 0, ∃ (M : ℕ) (B : Finset ℕ),
+      N < M ∧
+      B ⊆ Finset.Icc (N + 1) M ∧
+      IsSidon (A ∪ B).toSet ∧
+      (1 - ε) * M.sqrt ≤ (A ∪ B).card) ↔ answer(sorry) := by
```
callesonne:
```suggestion
theorem erdos_44 :
    (∀ (N : ℕ) (A : Finset ℕ) (_ : 1 ≤ N) (_ : A ⊆ Finset.Icc 1 N) (_ : IsSidon A.toSet),
      ∀ ε > (0 : ℝ), ∃ᵉ (M > N) (B ⊆ Finset.Icc (N + 1) M),
        IsSidon (A ∪ B).toSet ∧
        (1 - ε) * (M : ℝ).sqrt ≤ (A ∪ B).card) ↔ answer(sorry) := by
  sorry
```
A few suggestions here:
1. reformatting to keep below 100 character limit (and please do this in the other statements as well)
2. making sure that `ε` gets `ℝ` as its inferred type (currently it has type `ℕ`). Similarly making sure `M.sqrt` is using the real square root, currently it was using the natural number square root which involves taking a floor.
3. Finally using `∃ᵉ` the statement can be made a bit more compact.

callesonne (commented):
Thanks a lot for this contribution! Here are some initial comments :)

FormalConjectures/ErdosProblems/44.lean
saurabhraghuvanshii: I think yes, because many places we used `IsSidon A.toSet`, Without it, Lean would not be able to find the IsSidon definition.

callesonne: Have you tried deleting it to see if it compiles? Because `FormalConjectures.Util.ProblemImports` should include this import already, and if not I should fix this.

saurabhraghuvanshii: sorry for late, I checked  builds it is working without  `import FormalConjectures.ForMathlib.Combinatorics.Basic` so I removed it, thanks and only error I am seeing that copyright header..

FormalConjectures/ErdosProblems/44.lean
```diff
@@ -0,0 +1,107 @@
...
+# Erdős Problem 44: Extending Sidon Sets
+
+*Reference:* [erdosproblems.com/44](https://www.erdosproblems.com/44)
+
+Let N ≥ 1 and `A ⊆ {1,…,N}` be a Sidon set. Is it true that, for any ε > 0,
+there exist M = M(ε) and `B ⊆ {N+1,…,M}` such that `A ∪ B ⊆ {1,…,M}` is a Sidon set
+of size at least `(1−ε)M^{1/2}`?
+
+This problem asks whether any Sidon set can be extended to achieve a density
+arbitrarily close to the optimal density for Sidon sets.
```
callesonne: We usually put this description as the docstring to the erdos problem, and not as the module docstring.

FormalConjectures/ErdosProblems/44.lean
```diff
@@ -0,0 +1,107 @@
...
+-/
+
+open Function Set
+
+/-- The maximum size of a Sidon set in `{1, ..., N}`. -/
+noncomputable def maxSidonSetSize (N : ℕ) : ℕ :=
+  sSup {(A.card) | (A : Finset ℕ) (_ : A ⊆ Finset.Icc 1 N) (_ : IsSidon A.toSet)}
+
+/-- The optimal density for Sidon sets is approximately `√N`. -/
```
callesonne:
```suggestion
/-- The maximum size of a Sidon set in `{1, ..., N}` is less than or equal to `√N + 1`. -/
```
I would reword it like this, since currently this statement says nothing about optimality.

FormalConjectures/ErdosProblems/44.lean
```diff
@@ -0,0 +1,107 @@
...
+  sorry
+
+/--
+A Sidon set can be extended to achieve density close to optimal.
+-/
+@[category research open, AMS 5 11]
+theorem erdos_44 : (∀ (N : ℕ) (A : Finset ℕ) (_ : 1 ≤ N) (_ : A ⊆ Finset.Icc 1 N)
+    (_ : IsSidon A.toSet), ∀ (ε : ℝ) (_ : ε > 0), ∃ᵉ (M : ℕ) (B : Finset ℕ),
+      N < M ∧ B ⊆ Finset.Icc (N + 1) M ∧ IsSidon (A ∪ B).toSet ∧
+      (1 - ε) * Real.sqrt M ≤ (A ∪ B).card) ↔ answer(sorry) := by
```
callesonne:
```suggestion
theorem erdos_44 : (∀ᵉ (N ≥ (1 : ℕ)) (A ⊆ Finset.Icc 1 N), IsSidon A.toSet →
    ∀ᵉ (ε > (0 : ℝ)), ∃ᵉ (M > N) (B ⊆ Finset.Icc (N + 1) M),
      IsSidon (A ∪ B).toSet ∧ (1 - ε) * Real.sqrt M ≤ (A ∪ B).card) ↔ answer(sorry) := by
```
Here is a more compact way of writing this, utilizing `∃ᵉ` and `∀ᵉ` more. I also find it cleaner to use implications instead of unnamed hypothesis. Could you try to mimic these for the theorems below as well?

callesonne (commented):
Some more comments.

FormalConjectures/ErdosProblems/44.lean
```diff
@@ -0,0 +1,100 @@
...
+
+/-- The maximum size of a Sidon set in `{1, ..., N}`. -/
+noncomputable def maxSidonSetSize (N : ℕ) : ℕ :=
+  sSup {(A.card) | (A : Finset ℕ) (_ : A ⊆ Finset.Icc 1 N) (_ : IsSidon A.toSet)}
+
+/-- The maximum size of a Sidon set in `{1, ..., N}` is less than or equal to `√N + 1`. -/
+@[category API, AMS 5 11]
+theorem maxSidonSetSize_bound (N : ℕ) (hN : 1 ≤ N) :
+    maxSidonSetSize N ≤ N.sqrt + 1 := by
+  sorry
```
callesonne: Do you know how hard / easy this statement is? I would not use the tag `category API` but maybe `category undergraduate`, `category graduate` or `category research solved` depending on its difficulty.

FormalConjectures/ErdosProblems/44.lean
```diff
@@ -0,0 +1,100 @@
...
+
+/--
+A variant considering the extension to any larger range.
+-/
+@[category research open, AMS 5 11]
+theorem erdos_44.variant : (∀ᵉ (N ≥ (1 : ℕ)) (A ⊆ Finset.Icc 1 N), IsSidon A.toSet →
+    ∀ᵉ (ε > (0 : ℝ)) (M ≥ N), ∃ᵉ (B ⊆ Finset.Icc (N + 1) M),
+      IsSidon (A ∪ B).toSet ∧ (1 - ε) * Real.sqrt M ≤ (A ∪ B).card) ↔ answer(sorry) := by
+  sorry
```
callesonne: Will this variant not be false? If you pick `A` a "small" Sidon set, `ε` very small, and `M = N + 1`, then I'm guessing that there is no way to extend `A` to have optimal density in `[1, ..., M]` right? If you agree with this reasoning then I think this should be `category undergraduate` instead.

FormalConjectures/ErdosProblems/44.lean
```diff
@@ -0,0 +1,100 @@
...
+theorem sidon_set_lower_bound (N : ℕ) (hN : 1 ≤ N) :
+    ∃ᵉ (A ⊆ Finset.Icc 1 N), IsSidon A.toSet ∧ N.sqrt / 2 ≤ A.card := by
+  sorry
+
+/--
+The greedy construction gives a Sidon set of size approximately `√N`.
+-/
+@[category undergraduate, AMS 5 11]
+theorem greedy_sidon_construction (N : ℕ) (hN : 1 ≤ N) :
+    ∃ᵉ (A ⊆ Finset.Icc 1 N), IsSidon A.toSet ∧ A.card ≥ N.sqrt := by
```
callesonne, line 99: So given the theorem above this means that the set must have size `√N` or `√N + 1`?

callesonne (approved):
LGTM after these comments :) Thanks!

FormalConjectures/ErdosProblems/44.lean
saurabhraghuvanshii: hmhm, sorry I didn't thought about it

FormalConjectures/ErdosProblems/44.lean
saurabhraghuvanshii: Every Sidon set has size at most √N + 1.

FormalConjectures/ErdosProblems/44.lean
```diff
@@ -0,0 +1,100 @@
...
+  sorry
+
+/-! ## Related results and examples -/
+
+/--
+The set `{1, 2, 4, 8, 13}` is a Sidon set in `{1, ..., 13}`.
+-/
+@[category undergraduate, AMS 5 11]
+theorem example_sidon_set : IsSidon {1, 2, 4, 8, 13} := by
+  sorry
```
mo271, line 84:
```
intros a ha b hb c hc d hd h
  rcases ha with ha | ha | ha | ha | ha <;>
  rcases hb with hb | hb | hb | hb | hb <;>
  rcases hc with hc | hc | hc | hc | hc <;>
  rcases hd with hd | hd | hd | hd | hd <;>
  simp_all [ha, hc, hb, hd]
  ```
  would be a proof for this, but it is a little slow, so perhaps better not to include it.

mo271 (approved):

mo271:
Thanks!

FormalConjectures/ErdosProblems/44.lean
mo271: @saurabhraghuvanshii do you have a source for this?

saurabhraghuvanshii: no sir.

PR #287, opened 2025-07-01
refactor Sidon set definition

### LINKED ISSUE #27: Sidon sets

mo271:
closes #27

### DIFF
--- a/FormalConjectures/ForMathlib/Combinatorics/Basic.lean
+++ b/FormalConjectures/ForMathlib/Combinatorics/Basic.lean
@@ -20,19 +20,13 @@ import Mathlib.Order.Defs.PartialOrder
 
 open Function Set
 
-/-- A Sidon sequence (also called a Sidon set) is a sequence of natural numbers $a_0, a_1, ...$
-such that all pairwise sums $a_i + a_j$ are distinct for $i ≤ j$.
+variable {α γ: Type*} [AddCommMonoid α] [Membership α γ]
 
-In other words, all pairwise sums of elements are distinct apart from coincidences forced by the
-commutativity of addition. -/
-def IsSidon {ι α : Type*} [Preorder ι] [Preorder α] [Add α] (a : ι →o α) : Prop :=
-  ∀ i₁ j₁ i₂ j₂, i₁ ≤ j₁ → i₂ ≤ j₂ → a i₁ + a j₁ = a i₂ + a j₂ → i₁ = i₂ ∧ j₁ = j₂
-
-lemma IsSidon.injective {ι α : Type*} [Preorder ι] [Preorder α] [Add α] (a : ι →o α) (ha : IsSidon a) :
-    Injective a :=
-  fun i₁ i₂ hi ↦ (ha i₁ i₁ i₂ i₂ (le_refl _) (le_refl _) (by rw [hi])).1
-
-variable {α : Type*} [AddCommMonoid α]
+/-- A Sidon set is a set, such that such that all pairwise sums of elements are distinct apart from
+coincidences forced by the commutativity of addition. -/
+def IsSidon (A : γ) : Prop :=
+  Pairwise fun i₁ j₁ =>
+  i₁ ∈ A → j₁ ∈ A → ∀ i₂ ∈ A, ∀ j₂ ∈ A, i₁ + i₂ = j₁ + j₂ → (i₁ = j₂ ∧ i₂ = j₁)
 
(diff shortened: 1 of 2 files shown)


FormalConjectures/ForMathlib/Combinatorics/Basic.lean
```diff
@@ -20,19 +20,13 @@ import Mathlib.Order.Defs.PartialOrder
 
 open Function Set
 
-/-- A Sidon sequence (also called a Sidon set) is a sequence of natural numbers $a_0, a_1, ...$
-such that all pairwise sums $a_i + a_j$ are distinct for $i ≤ j$.
+variable {α γ: Type*} [AddCommMonoid α] [Membership α γ]
```
b-mehta: I would suggest specialising to gamma = Set alpha, the extra generality here is essentially spurious

FormalConjectures/ForMathlib/Combinatorics/Basic.lean
```diff
@@ -20,19 +20,13 @@ import Mathlib.Order.Defs.PartialOrder
...
-lemma IsSidon.injective {ι α : Type*} [Preorder ι] [Preorder α] [Add α] (a : ι →o α) (ha : IsSidon a) :
-    Injective a :=
-  fun i₁ i₂ hi ↦ (ha i₁ i₁ i₂ i₂ (le_refl _) (le_refl _) (by rw [hi])).1
-
-variable {α : Type*} [AddCommMonoid α]
+/-- A Sidon set is a set, such that such that all pairwise sums of elements are distinct apart from
+coincidences forced by the commutativity of addition. -/
+def IsSidon (A : γ) : Prop :=
+  Pairwise fun i₁ j₁ =>
+  i₁ ∈ A → j₁ ∈ A → ∀ i₂ ∈ A, ∀ j₂ ∈ A, i₁ + i₂ = j₁ + j₂ → (i₁ = j₂ ∧ i₂ = j₁)
```
b-mehta: This formulation is a bit surprising to me, I expected the more symmetric version (for all i1 j1 i2 j2 in A, i1 + i2 = j1 + j2 implies (i1=j1 and i2=j2) or (i1=j2 and i2=j1). I don't mind this one, but we should then have the lemma saying they're equivalent. Note also you could use Set.Pairwise with A.

FormalConjectures/ForMathlib/Combinatorics/Basic.lean
mo271: I had `A : Finset ℕ` in mind, because there seem to be a few conjectures around that. Not having the extra generality would force us to write
`IsSidon A.toSet` instead of `IsSidon A`, (and `IsSidon ↑A` also doesn't work) which seemed to justify the extra generality. What do you think?

FormalConjectures/ForMathlib/Combinatorics/Basic.lean
mo271: done, changed it to the symmetric version....

FormalConjectures/ForMathlib/Combinatorics/Basic.lean
b-mehta: I appreciate this point, but this is the general pattern in mathlib: we have `Set.Pairwise` and when we want to talk about a relation holding pairwise on a finset, we use `.toSet` for the conversion.

b-mehta:
I should add that I'd also been working on a general Sidon set definition for mathlib directly, but perhaps the easiest way is for me to PR that to mathlib and this one can be adjusted after the fact, so that neither one of us is blocked on the other?

mo271:
> I should add that I'd also been working on a general Sidon set definition for mathlib directly, but perhaps the easiest way is for me to PR that to mathlib and this one can be adjusted after the fact, so that neither one of us is blocked on the other?

Sounds like a plan. As soon as we have the mathlib version with your Sidon set definition we will remove this from our ForMathlib...

mo271:
needed for #286, #288 and #289

b-mehta (approved):
Thanks!

PR #291, opened 2025-07-01
update Erdős 730

mo271:
Reflecting the change on the website, adding also the conjecture that all pairs are of the form $(n, n+1)$.

### DIFF
--- a/FormalConjectures/ErdosProblems/730.lean
+++ b/FormalConjectures/ErdosProblems/730.lean
@@ -51,17 +51,25 @@ theorem erdos_730.variants.two_div_forall (n : ℕ) (h : 0 < n) : 2 ∣ (2*n).ch
 
 /--
 Show that $(n, n+1) ∈ S$ if and only if for all odd primes $p ≤ n$ both the base $p$
-representations of $n$ and $n+1$ either both have all digits less or equal to $(p-1)/2$
+representations of $n$ and $n+1$ either both have all digits less or equal to $\frac{p-1}{2}$
 or both don't.
-
-Note: currently there is stronger, but potentially false formulation of this criterion
-on erdosproblems.com.
 -/
 @[category undergraduate]
 theorem erdos_730.variants.succ_pair_criterion (n : ℕ) (h : 2 < n) :
     (n, n+1) ∈ S ↔ ∀ p ∈ Set.Ioc 2 n, ∀ [hp : Fact p.Prime],
     let kummer_condition: ℕ → Prop := fun n => (p.digits n).Forall (fun m => m ≤ (p - 1) / 2)
     (kummer_condition n ↔ kummer_condition (n + 1)) := by
+  sorry
+
+
+/--
+An observation by Thomas Bloom: we have $(n, n+1) ∈ S$ if for all primes $p≤n$ and $p=2n+1$
+(if such a prime exists) we have $n ∉ \{\frac{p - 1}{2}, p - 1\} \mod p$.
+-/
+@[category undergraduate]
+theorem erdos_730.variants.sufficient_criterion (n : ℕ) (h : 2 < n) :
+    (∀ p ∈ (Set.Ioc 2 n) ∪ {2 * p - 1}, ∀ [hp : Fact p.Prime],
+    (n : ZMod p) ≠ (p - 1) / 2 ∧ (n : ZMod p) ≠ p - 1) → (n, n + 1) ∈ S := by
   sorry
 
 open scoped Topology in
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/730.lean
```diff
@@ -64,6 +61,17 @@ theorem erdos_730.variants.succ_pair_criterion (n : ℕ) (h : 2 < n) :
...
   sorry
 
+
+/--
+An observation by Thomas Bloom: we have $(n, n+1) ∈ S$ if for all primes $p≤n$ and $p=2n+1$
+(if such a prime exists) we have $n ∉ \{\frac{p - 1}{2}, p - 1\} \mod p$.
+-/
+@[category undergraduate]
+theorem erdos_730.variants.sufficient_criterion (n : ℕ) (h : 2 < n) :
+    (∀ p ∈ (Set.Ioc 2 n) ∪ {2 * p - 1}, ∀ [hp : Fact p.Prime],
```
callesonne:
```suggestion
    (∀ p ∈ (Set.Ioc 2 n) ∪ {2 * n + 1}, ∀ [hp : Fact p.Prime],
```
Should it not be like this?

callesonne: Also this seems to be no longer mentioned on the website at all?

mo271: I was writing to Thomas Bloom and the site was changed yesterday again, because the observation didn't really make sense. Updated this pull request to reflect the changes. Now all statements should make sense.

FormalConjectures/ErdosProblems/730.lean
```diff
@@ -50,27 +51,8 @@ theorem erdos_730.variants.two_div_forall (n : ℕ) (h : 0 < n) : 2 ∣ (2*n).ch
...
-/--
-Standard heuristics then predict there should be $≫ \frac x {(\log x)^2}$
-many $n ≤ x$ such that $(n, n+1) ∈ S$.
+In every known example $(n, m) ∈ S$, we have $m = n + 1$.
 -/
 @[category research open, AMS 11]
-theorem erdos_730.variants.succ_pair_growth :
-    let C (x : ℝ) : ℝ := (Finset.Icc 0 ⌊x⌋₊ |>.filter fun n => (n, n+1) ∈ S).card
-    Filter.Tendsto (fun (x : ℝ) => x / (x.log^2) / C x) Filter.atTop (𝓝 0) := by
+theorem erdos_730.variants.delta_one (n m : ℕ) : (n, m) ∈ S → m = n + 1 := by
```
smmercuri: left-of-colon style is preferred
```suggestion
theorem erdos_730.variants.delta_one (n m : ℕ) (h : (n, m) ∈ S) : m = n + 1 := by
```

FormalConjectures/ErdosProblems/730.lean
```diff
@@ -50,27 +51,8 @@ theorem erdos_730.variants.two_div_forall (n : ℕ) (h : 0 < n) : 2 ∣ (2*n).ch
...
-    (n, n+1) ∈ S ↔ ∀ p ∈ Set.Ioc 2 n, ∀ [hp : Fact p.Prime],
-    let kummer_condition: ℕ → Prop := fun n => (p.digits n).Forall (fun m => m ≤ (p - 1) / 2)
-    (kummer_condition n ↔ kummer_condition (n + 1)) := by
-  sorry
-
-open scoped Topology in
-/--
-Standard heuristics then predict there should be $≫ \frac x {(\log x)^2}$
-many $n ≤ x$ such that $(n, n+1) ∈ S$.
+In every known example $(n, m) ∈ S$, we have $m = n + 1$.
```
smmercuri, line 54: From the website, does this variant not also require the condition that $n < m$? Ditto in the formal version

mo271: I found it preferable to change the definition of `S`, even though then the statement is a little less closely aligned with the informal version on the website, I think this is simpler.

FormalConjectures/ErdosProblems/730.lean
```diff
@@ -19,21 +19,22 @@ import FormalConjectures.Util.ProblemImports
...
 # Erdős Problem 730
 
-*Reference:* [erdosproblems.com/730](https://www.erdosproblems.com/730)
+*References:*
+  - [erdosproblems.com/730](https://www.erdosproblems.com/730)
+  - [A129515](https://oeis.org/A129515)
 -/
 private abbrev S :=
-  {(n, m) : ℕ × ℕ | n ≠ m ∧ ((2*n).choose n).primeFactors = ((2*m).choose m).primeFactors}
+  {(n, m) : ℕ × ℕ | n < m ∧ ((2*n).choose n).primeFactors = ((2*m).choose m).primeFactors}
```
callesonne, line 27: How come this change? On the website it still says `≠`

callesonne: Ah, I guess it doesn't matter if you are asking for the infinitude of this set, so I don't mind if it makes things simpler.

callesonne (approved):
LGTM!

Issue #276, opened 2025-06-30
Erdős Problem 64: Cycles of length power-of-two in graphs with min degree >= 3.


mo271:
### What is the conjecture
    
https://www.erdosproblems.com/64

  Does every finite graph with minimum degree at least 3 contain a cycle of length $2^k$ for some $k\geq 2$?

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

m-ow:
I'd like to work on this!

franzhusch:
Was added in #292

PR #292, opened 2025-07-01
Erdős 64

m-ow:
https://github.com/google-deepmind/formal-conjectures/issues/276

### DIFF
--- a/FormalConjectures/ErdosProblems/64.lean
+++ b/FormalConjectures/ErdosProblems/64.lean
@@ -1,0 +1,34 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/64.lean
```diff
@@ -0,0 +1,34 @@
...
+
+variable {V : Type*} (G : SimpleGraph V) [Fintype V] [DecidableRel G.Adj]
+/--
+Does every finite graph with minimum degree at least $3$
+contain a cycle of length $2^k$ for some $k \geq 2$?
+-/
+@[category research open, AMS 5]
+theorem erdos_64 (h : G.minDegree ≥ 3) :
+    ∃ (k : ℕ) (v : V) (c : G.Walk v v),
+      k ≥ 2 ∧ c.IsCycle ∧ c.length = 2^k ↔ answer(sorry) := by
```
callesonne: Here you need to be careful when using `↔ answer(sorry)`. It is important that this happens *outside* the quantification over `G` and `V` (because presumable a counterexample would involve providing a specific graph). So I would quantify like `(∀ {V : Type*} (G : SimpleGraph V) ...., ∃ (k : ℕ) ..., k ≥ 2 ∧ c.IsCycle ∧ c.length = 2^k) ↔ answer(sorry)`

callesonne (commented):
Thanks a lot for the contribution! If you want, we always encourage people to also formalize some of the related statements on the given Erdős problem page. For example, I think the following statement could also be interesting to formalize the variant that this holds for the minimum degree being at least the size of some given constant (so related to what Liu and Montgomery proved.

m-ow:
@callesonne Thank you for the feedback! I've implemented your suggestion by moving the `answer(sorry)` outside the quantification.
For the Liu-Montgomery variant, I'll focus on areas within my current expertise as their proof uses advanced methods. Hope someone with stronger graph theory skills tackles this later. 

Appreciate the review!

mo271 (approved):

PR #296, opened 2025-07-03
hadwiger

AlexKontorovich:


Paul-Lez:
Thanks for the PR! I see the wiki page also mentions a couple of generalizations and solved special cases - would you be up for adding a few as extra sanity checks?
I'll take a look at the maths tomorrow:)

AlexKontorovich:
Darn, it's not stated correctly. I should've stated it in terms of graph minors (which don't seem to be in Mathlib) instead of embeddings. I'll fix it in a few days (I'm off for the holiday...)

FormalConjectures/Wikipedia/Hadwiger.lean
```diff
@@ -0,0 +1,605 @@
...
+
+/--
+A `SimpleGraph'` has vertices `verts` which is a subset of some type `V`, and
+the condition that if two vertices are adjacent, `Adj u v`, then both are in
+the vertex set.
+-/
+structure SimpleGraph' (V : Type*) where
+  verts : Set V
+  Adj : V → V → Prop
+  Adj_of_verts : ∀ (u v : V), Adj u v → u ∈ verts ∧ v ∈ verts
```
YaelDillies:
```suggestion
  left_mem_verts_of_adj : ∀ (u v : V), Adj u v → u ∈ verts
```
The other implication comes from symmetry

FormalConjectures/Wikipedia/Hadwiger.lean
```diff
@@ -0,0 +1,605 @@
...
+  symm _ _ h := ⟨G.symm h.1, h.2.2, h.2.1⟩
+  loopless v h := G.loopless v h.1
+
+/--
+New graph obtained by deleting a given set of edges.
+-/
+def SimpleGraph'.deleteEdges (G : SimpleGraph' V) (s : Set (Sym2 V)) :
+    SimpleGraph' V where
+  verts := G.verts
+  Adj u v := if s(u, v) ∈ s then False else G.Adj u v
```
YaelDillies: That is a funny way to write
```suggestion
  Adj u v := s(u, v) ∉ s ∧ G.Adj u v
```
Probably
```suggestion
  Adj u v := G.Adj u v ∧ s(u, v) ∉ s
```
is better, to match `sdiff`

FormalConjectures/Wikipedia/Hadwiger.lean
YaelDillies:
```suggestion
  Adj_of_verts : ∀ (u v : V), Adj u v → u ∈ verts ∧ v ∈ verts := by aesop_graph
```
too, maybe?

FormalConjectures/Wikipedia/Hadwiger.lean
```diff
@@ -0,0 +1,605 @@
...
+      simp only [h_case', not_false_eq_true, true_and]
+      exact G.symm h
+  loopless v h := by
+    by_cases h_case : s(v, v) ∈ s
+    · simp [h_case] at h
+    · simp only [h_case, ↓reduceIte] at h
+      exact G.loopless v h
+
+def SimpleGraph'.edgeSet (G : SimpleGraph' V) : Set (Sym2 V) :=
+  {e | ∃ u v, s(u, v) = e ∧ G.Adj u v}
```
YaelDillies:
```suggestion
  {e | Sym2.lift ⟨(G.Adj · ·), sorry⟩ e}
```
This way, `s(u, v) ∈ G.edgeSet` is defeq to `G.Adj u v`

FormalConjectures/Wikipedia/Hadwiger.lean
```diff
@@ -0,0 +1,605 @@
...
+inductively delete edges, vertices, or contract an edge.
+-/
+inductive SimpleGraph'.Minor : SimpleGraph' V → SimpleGraph' V → Prop
+| refl (G : SimpleGraph' V) : G.Minor G
+| delete_edge {G H : SimpleGraph' V} (e : Sym2 V) :
+    H.Minor G → H.Minor (G.deleteEdges {e})
+| delete_vertex {G H : SimpleGraph' V} (v : V) :
+    H.Minor (G.deleteVerts {v}) → H.Minor G
+| contract_edge {G H : SimpleGraph' V} (e : Sym2 V) (he : e ∈ G.edgeSet) :
+    H.Minor (G.contractEdge he) → H.Minor G
```
YaelDillies, line 164: Is this really correct? You can't delete infinitely many vertices this way, so eg infinitely many disjoint copies of K5 don't contain a K5 minor according to your definition . I believe you should instead allow removing several vertices and several edges at once

AlexKontorovich:
> I am reviewing this as a mathlib PR, because I now believe it is the correct design for `SimpleGraph` in mathlib. Would be good to get this refactor going!

Thanks!!

FormalConjectures/Wikipedia/Hadwiger.lean
AlexKontorovich: Sure, I guess I had in mind applications to finite graphs, but you're right that this should be done generally!

YaelDillies:
It looks like we'll the refactor in mathlib: https://github.com/leanprover-community/mathlib4/pull/33466

AlexKontorovich:
> It looks like we'll the refactor in mathlib: [leanprover-community/mathlib4#33466](https://github.com/leanprover-community/mathlib4/pull/33466)

Please let me know when that goes through, so I can update this. Thanks!

PR #299, opened 2025-07-04
Replace lowerDensity with new one for Erdos 10

seewoo5:
It seems that the definition of `lowerDensity` in Erdos 10 is added before more general definition is added to `ForMathlib/Data/Set/Density.lean`. Remove it and replace with new one. Let me know if this is wrong.


mo271 (approved):

callesonne:
Great catch, thanks a lot!

PR #300, opened 2025-07-04
Erdos 9

### LINKED ISSUE #190: Erdős Problem 9: integers not of the form $p+2^k+2^l$
### What is the conjecture

https://www.erdosproblems.com/9

  Let $A$ be the set of all integers not of the form $p+2^k+2^l$ (where $k,l\geq 0$ and $p$ is prime). Is the upper density of $A$ positive?

Status: open

### Prerequisites needed

Upper density here: https://github.com/google-deepmind/formal-conjectures/blob/ed8a8090f9b10a188daeec228150351831790395/FormalConjectures/ForMathlib/Data/Set/Density.lean#L44

seewoo5:
Closes #190.

### DIFF
--- a/FormalConjectures/ErdosProblems/9.lean
+++ b/FormalConjectures/ErdosProblems/9.lean
@@ -25,7 +25,7 @@ import FormalConjectures.Util.ProblemImports
 /--
 The set of natural numbers that can be expressed as a prime plus two powers of 2.
 -/
-def Erdos9A : Set ℕ := { n | ∃ (p k l : ℕ), (Nat.Prime p) ∧ n = p + 2 ^ k + 2 ^ l }
+def Erdos9A : Set ℕ := { n | ¬ ∃ (p k l : ℕ), (Nat.Prime p) ∧ n = p + 2 ^ k + 2 ^ l }
 
 /--
 Is the upper density of the set of natural numbers that can be expressed as a prime plus
--- a/FormalConjectures/ErdosProblems/9.lean
+++ b/FormalConjectures/ErdosProblems/9.lean
@@ -27,6 +27,33 @@ The set of odd numbers that cannot be expressed as a prime plus two powers of 2.
 -/
 def Erdos9A : Set ℕ := { n | Odd n ∧ ¬ ∃ (p k l : ℕ), (Nat.Prime p) ∧ n = p + 2 ^ k + 2 ^ l }
 
+
+@[category test]
+example : 1 ∈ Erdos9A := by
+  constructor
+  · decide
+  · push_neg
+    intro p k l hp
+    linarith [Nat.Prime.two_le hp, @Nat.one_le_two_pow k, @Nat.one_le_two_pow l]
+
+@[category test]
+example : 3 ∈ Erdos9A := by
+  constructor
+  · decide
+  · push_neg
+    intro p k l hp
+    linarith [Nat.Prime.two_le hp, @Nat.one_le_two_pow k, @Nat.one_le_two_pow l]
+
+@[category test]
+example : 5 ∉ Erdos9A := by
+  unfold Erdos9A
+  simp only [exists_and_left, not_exists, not_and, Set.mem_setOf_eq, not_forall, Classical.not_imp,
+    Decidable.not_not]
+  intro
+  use 3, Nat.prime_three, 0, 0
+  simp only [pow_zero, Nat.reduceAdd]
+
+
 /--
 The set is known to be infinite. In [Er77c] Erdős credits Schinzel with proving that there are
 infinitely many odd integers not of this form, but gives no reference.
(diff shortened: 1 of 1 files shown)


callesonne (commented):
Perfect! Thanks a lot :) Could you also add the (solved) statement that this set is infinite? Other than that LGTM

seewoo5:
> Perfect! Thanks a lot :) Could you also add the (solved) statement that this set is infinite? Other than that LGTM

Added!

FormalConjectures/ErdosProblems/9.lean
```diff
@@ -0,0 +1,51 @@
...
+/--
+The set of odd numbers that cannot be expressed as a prime plus two powers of 2.
+-/
+def Erdos9A : Set ℕ := { n | Odd n ∧ ¬ ∃ (p k l : ℕ), (Nat.Prime p) ∧ n = p + 2 ^ k + 2 ^ l }
+
+/--
+The set is known to be infinite. Erdős [Er77c] credits Schinzel with proving that there are
+infinitely many odd integers not of this form, but gives no reference. Crocker [Cr71] has proved
+there are are $\gg \log \log N$ such integers in $\{1, \dots, N\}$.
+Pan [Pa11] improved this to $\gg_\epsilon N^{1 - \epsilon}$ for any $\epsilon > 0$.
```
callesonne: I would not add these last 2 sentences here, since they are not related to this particular statement. However if you wanted to add those statements as well then this would be very welcome! :)

seewoo5: You're right, I choose easier way 😅

FormalConjectures/ErdosProblems/9.lean
mo271, line 1: I suppose one could add test statements like 
```lean
example : 1 ∈ Erdos9A := by
  constructor
  · decide
  · push_neg
    intro p k l hp
    linarith [Nat.Prime.two_le hp, @Nat.one_le_two_pow k, @Nat.one_le_two_pow l]

example : 3 ∈ Erdos9A := by
  constructor
  · decide
  · push_neg
    intro p k l hp
    linarith [Nat.Prime.two_le hp, @Nat.one_le_two_pow k, @Nat.one_le_two_pow l]

example : 5 ∉ Erdos9A := by
  unfold Erdos9A
  simp only [exists_and_left, not_exists, not_and, Set.mem_setOf_eq, not_forall, Classical.not_imp,
    Decidable.not_not]
  intro
  use 3
  simp only [exists_prop]
  constructor
  · decide
  · use 0, 0
    norm_num
  ```
  
  but I'm on the fence about it if this is useful.
  
  What would be useful, is to find the first odd number bigger than 3 in in https://oeis.org/A006286, and write down the statement that it is in `Erdos9A`, which must exist, since the set is infitie, but looking at the sequence in the oeis, it looks like it might not be small.

**Review by mo271 — APPROVED** [review]
LGTM, thanks!
Further suggestions are optional, from my point of view we can merge as is.

File: FormalConjectures/ErdosProblems/9.lean [hunk 2]
```diff
@@ -0,0 +1,46 @@
...
+The set of odd numbers that cannot be expressed as a prime plus two powers of 2.
+-/
+def Erdos9A : Set ℕ := { n | Odd n ∧ ¬ ∃ (p k l : ℕ), (Nat.Prime p) ∧ n = p + 2 ^ k + 2 ^ l }
+
+/--
+The set is known to be infinite. Erdős credits Schinzel with proving that there are
+infinitely many odd integers not of this form, but gives no reference.
+
+Ref: Erdős, P., _Problems and results on combinatorial number theory. III._.
+-/
```
callesonne:
```suggestion
/--
The set is known to be infinite. In [Er77c] Erdős credits Schinzel with proving that there are
infinitely many odd integers not of this form, but gives no reference.

[Er77c] Erdős, P., _Problems and results on combinatorial number theory. III._.
-/
```

FormalConjectures/ErdosProblems/9.lean
seewoo5: I think it is good to have, I'll add your suggestions. Also, I checked myself up to 10^7, and cannot find other than 1, 3. This seems very rare. Based on the description of [Erdos 10](https://www.erdosproblems.com/10), Granville and Soundrarajan conjectured that every odd number is a sum of prime and *three* powers of 2, but probably *two* would be enough for odds > 3?

FYI, here's a sage code:
```python
from itertools import product


def A(N):
    """
    Set of odd integers 1 <= n <= N that cannot be expressed as p + 2 ^ a + 2 ^ b,
    where p is a prime and a, b are non-negative integers.
    """
    Ps = primes(N)
    in_A = dict.fromkeys(range(1, N + 1, 2), True)
    for p in Ps:
        for a, b in product(range(0, ceil(log(N, 2))), repeat=2):
            n = p + 2 ** a + 2 ** b
            if n % 2 == 1 and n <= N:
                in_A[n] = False
    return [n for n in in_A if in_A[n]]
```

PR #301, opened 2025-07-04
Add Erdos 307

seewoo5:


FormalConjectures/ErdosProblems/307.lean
```diff
@@ -0,0 +1,48 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 307
+
+*Reference:* [erdosproblems.com/307](https://www.erdosproblems.com/304)
```
callesonne:
```suggestion
*Reference:* [erdosproblems.com/307](https://www.erdosproblems.com/307)
```

FormalConjectures/ErdosProblems/307.lean
```diff
@@ -0,0 +1,48 @@
...
+
+/--
+Are there two finite set of primes $P$ and $Q$ such that
+
+$$
+1 = \left( \sum_{p \in P} \frac{1}{p} \right) \left( \sum_{q \in Q} \frac{1}{q} \right)
+$$
+? Asked by Barbeau.
+
+Ref: Barbeau, E. J., _Computer challenge corner: Problem 477: A brute force program._
```
callesonne:
```suggestion
?

Asked by Barbeau [Ba76].

[Ba76] Barbeau, E. J., _Computer challenge corner: Problem 477: A brute force program._
```

This is the format for references we have been using mostly.

FormalConjectures/ErdosProblems/307.lean
```diff
@@ -0,0 +1,48 @@
...
+-/
+@[category research open, AMS 5 11]
```
callesonne:
```suggestion
@[category research open, AMS 11]
```
To me this problem seems like it should only be tagged with `11` (happy to be convinced though!)

FormalConjectures/ErdosProblems/307.lean
```diff
@@ -0,0 +1,48 @@
...
+theorem erdos_307 : (∃ P Q : Finset ℕ, (∀ p ∈ P, p.Prime) ∧ (∀ q ∈ Q, q.Prime) ∧
+    (1 = (∑ p ∈ P, (p : ℚ)⁻¹) * (∑ q ∈ Q, (q : ℚ)⁻¹))) ↔ answer(sorry) := by
+  sorry
+
+/--
+Instead of set of primes, ask for all $p \in P$ to be relatively coprime,
+and similarly for $q \in Q$.
+-/
+@[category research open, AMS 5 11]
+theorem erdos_307_coprime : (∃ P Q : Finset ℕ, (∀ p₁ p₂ : P, p₁ ≠ p₂ → Nat.Coprime p₁ p₂) ∧
```
callesonne: Here I think it would be nice to use `Pairwise` to express that all primes are pairwise coprime

callesonne (commented):
Thanks! :tada:

FormalConjectures/ErdosProblems/307.lean
mo271: yeah, someting like
```
     Pairwise (fun (p₁ p₂ : P) => Nat.Coprime p₁ p₂) ∧
     Pairwise (fun (q₁ q₂ : Q) => Nat.Coprime q₁ q₂)
 ```

Comment (callesonne) [reply FormalConjectures/ErdosProblems/307.lean]: Yes, but maybe even something like `P.Pairwise Nat.Coprime` could work (using `Set.Pairwise`)

Comment (mo271) [reply FormalConjectures/ErdosProblems/307.lean]: does that work, or would we need `P.toSet.Pairwise Nat.Coprime`?

Comment (callesonne) [reply FormalConjectures/ErdosProblems/307.lean]: Ahh yes, you're probably right

File: FormalConjectures/ErdosProblems/307.lean (hunk 3 above)
Comment (seewoo5) [reply FormalConjectures/ErdosProblems/307.lean]: You're right, I forgot to remove 5 after copying from other files.

File: FormalConjectures/ErdosProblems/307.lean (hunk 4 above)
Comment (seewoo5) [reply FormalConjectures/ErdosProblems/307.lean]: Thanks! I didn't know `Pairwise` exists in mathlib.

File: FormalConjectures/ErdosProblems/307.lean [hunk 5]
```diff
@@ -0,0 +1,50 @@
...
+[Ba76] Barbeau, E. J., _Computer challenge corner: Problem 477: A brute force program._
+-/
+@[category research open, AMS 11]
+theorem erdos_307 : (∃ P Q : Finset ℕ, (∀ p ∈ P, p.Prime) ∧ (∀ q ∈ Q, q.Prime) ∧
+    (1 = (∑ p ∈ P, (p : ℚ)⁻¹) * (∑ q ∈ Q, (q : ℚ)⁻¹))) ↔ answer(sorry) := by
+  sorry
+
+/--
+Instead of set of primes, ask for all $p \in P$ to be relatively coprime,
+and similarly for $q \in Q$.
```
callesonne:
```suggestion
Instead of asking for sets of primes, ask only that all primes in the sets be relatively coprime.
```

mo271 (commented):
Thanks for the contribution!

Issue #314, opened 2025-07-04
The Hadwiger-Nelson Problem


BoltonBailey:
### What is the conjecture

The [Hadwiger-Nelson problem](https://en.wikipedia.org/wiki/Hadwiger%E2%80%93Nelson_problem) asks, what is

> "the minimum number of colors required to color the [plane](https://en.wikipedia.org/wiki/Plane_(mathematics)) such that no two [points](https://en.wikipedia.org/wiki/Point_(geometry)) at distance 1 from each other have the same color."

### Prerequisites needed

I noticed this problem is mentioned in the README, but a formalization doesn't actually seem to exist in the repository that I can find.

Luckily, I think it is fairly straightforward to define the graph using the SimpleGraph API from `mathlib`, and indeed, I did so a few weeks ago while looking at open problem formalizations for Project Numina:

```
def UnitDistancePlaneGraph : SimpleGraph (EuclideanSpace ℝ (Fin 2)) :=
  SimpleGraph.mk
    (fun x y => dist x y = 1)
    (by
      intros x y
      simp [dist_comm])
    (by
      intros x
      simp [dist_self])
```

I am happy to make a PR for this myself.

### AMS categories

* ams-52

### Choose either option
- [x] I plan on working on this conjecture
- [ ] This issue is up for grabs: I would like to see this conjecture added by somebody else

BoltonBailey:
Closed by #315

PR #315, opened 2025-07-04
feat: add Hadwiger-Nelson problem

BoltonBailey:
This PR adds the Hadwiger-Nelson problem as an open research problem. It also adds the lower bound of 5 as a solved research problem.

I have also taken the liberty of updating the README so that the Hadwiger-Nelson example there matches the formalization I provided.

Addresses issue #314

### DIFF
--- a/FormalConjectures/Wikipedia/HadwigerNelson.lean
+++ b/FormalConjectures/Wikipedia/HadwigerNelson.lean
@@ -40,8 +40,6 @@ def UnitDistancePlaneGraph : SimpleGraph (EuclideanSpace ℝ (Fin 2)) :=
 /--
 The Hadwiger–Nelson problem asks: How many colors are required to color the plane
 such that no two points at distance 1 from each other have the same color?
-
-It is known that 7 colors are sufficient.
 -/
 @[category research open, AMS 52]
 theorem HadwigerNelsonProblem :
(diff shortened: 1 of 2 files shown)


BoltonBailey:
Note: I used AMS 52 "Convex and discrete geometry" since that seemed most correct, but I'm not sure how to apply label tag to the PR or issue.

FormalConjectures/Wikipedia/HadwigerNelson.lean
```diff
@@ -0,0 +1,59 @@
...
+      intros x y
+      simp [dist_comm])
+    (by
+      intros x
+      simp [dist_self])
+/--
+The Hadwiger–Nelson problem asks: How many colors are required to color the plane
+such that no two points at distance 1 from each other have the same color?
+
+It is known that 7 colors are sufficient.
```
mo271: Would it make sense to also add `UnitDistancePlaneGraph.chromaticNumber ≤ 7` as a sepearate solved theorem?

mo271 (approved):
LGTM, thanks!

FormalConjectures/Wikipedia/HadwigerNelson.lean
BoltonBailey: I thought about this, but I was unsure if that's what we were going for. The "proof" is really just this picture from wikipedia, I am not sure if there is a formal source or whether it qualifies as a "research" problem.

callesonne: I think it would still be interesting to add, but maybe with the tag `high school` or `undergraduate`.

BoltonBailey: Ok then, I have now added it!

PR #343, opened 2025-07-09
feat: statements for degree sequences in triangle-free graphs

henrykmichalewski:
This commit introduces the formal statements (definitions, lemmas, and theorems) from the paper "Degree sequences in triangle-free graphs" by P. Erdős, S. Fajtlowicz, and W. Staton.

The work was published in Discrete Mathematics, Volume 92, Issues 1–3, in November 1991, pages 85-88. 

The paper answers one of the Graffiti conjectures. Graffiti conjectures appeared in this repository in the directory FormalConjectures/WrittenOnTheWallII.

### DIFF
--- a/FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
+++ b/FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
@@ -53,12 +53,9 @@ lemma lemma1_a
     (h_pos : ∀ k, 0 < d k)
     (h_no_three : ∀ k, d (k + 2) ≠ d k) :
     1 ≤ d (k + 2) - d k := by
-  rw [Nat.one_le_iff_ne_zero]
-  intro h
-  rw [Nat.sub_eq_zero_iff_le] at h
-  have h' : d k ≤ d (k + 2) := h_mono (le_add_of_nonneg_right (Nat.zero_le 2))
-  have h_eq : d (k + 2) = d k := le_antisymm h h'
-  exact h_no_three k h_eq
+  have : d k ≤ d (k + 2) := h_mono (by omega)
+  have := h_no_three k
+  omega
 
 /-- **Lemma 1 (b)**
 If a sequence `d` is nondecreasing and no three terms are equal, then terms at distance `2 * r` differ by at least `r`. -/


FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,173 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+import FormalConjectures.ForMathlib.Combinatorics.SimpleGraph.DiamExtra
+import FormalConjectures.ForMathlib.Combinatorics.SimpleGraph.GraphConjectures.Domination
+import FormalConjectures.ForMathlib.Combinatorics.SimpleGraph.Bipartite
+import FormalConjectures.ForMathlib.Combinatorics.SimpleGraph.GraphConjectures.Definitions
+import FormalConjectures.ForMathlib.Combinatorics.SimpleGraph.GraphConjectures.Invariants
```
mo271:
```suggestion
```

will be possible after https://github.com/google-deepmind/formal-conjectures/pull/344

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,173 @@
...
+theorem theorem1 (G : SimpleGraph α) [DecidableRel G.Adj] (h₁ : G.CliqueFree 3) (h₂ : f G = 2) :
+    G.IsBipartite ∧ δ G = 1 ∧ degreeSequenceCompact G := by
+  sorry
+
+/-- **Lemma 3.** For every `n` there exists a bipartite graph with
+`8 n` vertices, minimum degree `n + 1`, and `f = 3`. -/
+@[category research solved, AMS 5]
+lemma lemma3 (n : ℕ) :
+    ∃ (G : SimpleGraph (Fin (8 * n))) (_ : DecidableRel G.Adj),
+      G.IsBipartite ∧ δ G = n + 1 ∧ f G = 3 := by
```
Paul-Lez: I think you'll want `n` to be positive here!

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,173 @@
...
+noncomputable def δ (G : SimpleGraph α) [DecidableRel G.Adj] : ℕ :=
+  sInf (Set.range fun v : α => G.degree v)
+
+end SimpleGraph
+
+section lemmas
+
+variable (d : ℕ → ℕ) (n k r : ℕ)
+
+/-- **Lemma 1 (a)** -/
```
Paul-Lez: I think it will be helpful to also have an informal version of this lemma in the docstring (and similarly for the ones below)

Paul-Lez (commented):
Nice work! Two very minor comments - I'll take another look at the maths tomorrow.

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
mo271: #344 now merged, so this suggestion can be applied

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,173 @@
...
+
+/-- `δ G` is the minimum degree of `G`. -/
+noncomputable def δ (G : SimpleGraph α) [DecidableRel G.Adj] : ℕ :=
+  sInf (Set.range fun v : α => G.degree v)
+
+end SimpleGraph
+
+section lemmas
+
+variable (d : ℕ → ℕ) (n k r : ℕ)
```
Paul-Lez, line 46: Here the paper also assumes that `d` is nondecreasing, that all terms are positive and that no three terms are equal.

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,173 @@
...
+variable {α : Type*} [Fintype α] [DecidableEq α]
+
+/-- The degree sequence of a graph, sorted in nondecreasing order. -/
+noncomputable def degreeSequence (G : SimpleGraph α) [DecidableRel G.Adj] : List ℕ :=
+  (Finset.univ.val.map fun v : α => G.degree v).sort (· ≤ ·)
+
+/-- The degree sequence of `G` is **compact** if it satisfies
+`IsCompactSequence`. -/
+def degreeSequenceCompact (G : SimpleGraph α) [DecidableRel G.Adj] : Prop :=
+  IsCompactSequence fun k => (degreeSequence G).getD k 0
```
Paul-Lez: Here you'll run into trouble because the compactness is only meant to hold for small `k`  (for large `k`, you will have `d_{k+2} - d_k = 0`). One way out of this would be to define an `IsCompactSequenceOn` predicate, which takes as extra parameter a set, and states that the sequence is compact _for inputs from that set_.

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: Done, thank you!

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: Done, thank you!

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,168 @@
...
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+Title: Degree sequences in triangle-free graphs
+Authors: P. Erdős, S. Fajtlowicz and W. Staton,
+Published in Discrete Mathematics 92 (1991) 85–88.
+-/
+
+open BigOperators
```
YaelDillies, line 25: This has been a noop for a while now
```suggestion
```

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,168 @@
...
+open Classical
```
YaelDillies, line 26: If you do
```suggestion
open Classical
open scoped Finset
```

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,168 @@
...
+def IsCompactSequence (d : ℕ → ℕ) : Prop :=
+  ∀ k, d (k + 2) = d k + 1
+
+namespace SimpleGraph
+
+variable {α : Type*} [Fintype α] [DecidableEq α]
+
+/-- The number of vertices of `G` having degree `d`. -/
+noncomputable def degreeFreq (G : SimpleGraph α) (d : ℕ) : ℕ :=
+  (Finset.univ.filter fun v : α => G.degree v = d).card
```
YaelDillies: Then here you can do
```suggestion
  #{v | G.degree v = d}
```

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,168 @@
...
+
+/-- `δ G` is the minimum degree of `G`. -/
+noncomputable def δ (G : SimpleGraph α) [DecidableRel G.Adj] : ℕ :=
+  sInf (Set.range fun v : α => G.degree v)
```
YaelDillies: This is already `G.minDegree`

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,168 @@
...
+section lemmas
+
+variable (d : ℕ → ℕ) (n k r : ℕ)
+
+/-- **Lemma 1 (a)** -/
+@[category research solved, AMS 5]
+lemma lemma1_a
+    (h_no_three : ∀ k, d (k + 2) ≠ d k) :
+    1 ≤ d (k + 2) - d k := by
+  sorry
```
YaelDillies: Idle question: Once you fix the statement to assume that `d` is monotone, this is a triviality. Do we want trivialities in formal-conjectures? Erdos and al stated it as a lemma, but I assume it is by convenience rather than because it is truly important

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,168 @@
...
+    r ≤ d (k + 2 * r) - d k := by
+  sorry
+
+/-- **Lemma 2 (a)** -/
+@[category research solved, AMS 5]
+lemma lemma2_a
+    (h_no_three : ∀ i, d (i + 2) ≠ d i) :
+    2 * n * n ≤
+      (∑ i ∈ Finset.Icc (2 * n + 1) (4 * n), d i) -
+        (∑ i ∈ Finset.Icc 1 (2 * n), d i) := by
```
YaelDillies: FYI precedence does what you would expect here:
```suggestion
      ∑ i ∈ .Icc (2 * n + 1) (4 * n), d i -
        ∑ i ∈ .Icc 1 (2 * n), d i := by
```

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,168 @@
...
+      let J := H.induce (Set.compl (Set.range i))
+      J.CliqueFree 3 ∧ f J = 3 ∧ δ J ≥ 2 * Fintype.card α := by
+  sorry
+
+/-- **Theorem 2.** Every triangle-free graph is an induced subgraph of one
+with `f = 3`. -/
+@[category research solved, AMS 5]
+theorem theorem2 (G : SimpleGraph α) [DecidableRel G.Adj] (h : G.CliqueFree 3) :
+    ∃ (β : Type*) (_ : Fintype β) (H : SimpleGraph β) (_ : DecidableRel H.Adj) (i : α ↪ β),
+      (∀ u w, G.Adj u w ↔ H.Adj (i u) (i w)) ∧ H.CliqueFree 3 ∧ f H = 3 := by
```
YaelDillies:
```suggestion
    ∃ (β : Type*) (_ : Fintype β) (H : SimpleGraph β) (_ : DecidableRel H.Adj) (i : α ↪g β),
      H.CliqueFree 3 ∧ f H = 3 := by
```

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,168 @@
...
+      G.IsBipartite ∧ δ G = n + 1 ∧ f G = 3 := by
+  sorry
+
+/-- **Lemma 4.** A triangle-free graph can be extended by one to a
+triangle-free graph whose new part has minimum degree at least
+`2 n` and `f = 3`. -/
+@[category research solved, AMS 5]
+lemma lemma4 (G : SimpleGraph α) [DecidableRel G.Adj] (h₁ : G.CliqueFree 3) (v : α) :
+    ∃ (β : Type*) (_ : Fintype β) (H : SimpleGraph β) (_ : DecidableRel H.Adj) (i : α ↪ β),
+      (∀ u w, G.Adj u w ↔ H.Adj (i u) (i w)) ∧
```
YaelDillies:
```suggestion
    ∃ (β : Type*) (_ : Fintype β) (H : SimpleGraph β) (_ : DecidableRel H.Adj) (i : α ↪g β),
```

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: Done

.gitignore
YaelDillies, line 1: What happened here?

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: Done

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: Thank you, I think that I managed to apply it.

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: This assumption is not inlined in all statements.

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: To address the comment regarding degreeSequenceCompact and the "small k" issue, we refactored the definition of compactness to be relative to a set of indices.

Specifically, we made the following changes in  `FormalConjectures/Paper/DegreeSequencesTriangleFree.lean`:

1. Introduced ` IsCompactSequenceOn`:

Instead of a global IsCompactSequence that applies to all k, we defined a new predicate that takes a set S of valid indices:
```lean
def IsCompactSequenceOn (d : ℕ → ℕ) (S : Set ℕ) : Prop :=
  ∀ k ∈ S, d (k + 2) = d k + 1
```

2. Updated `degreeSequenceCompact`: 

We then updated the `degreeSequenceCompact `definition to use this new predicate, restricting the check to indices` k` where `k + 2` is strictly less than the number of vertices (`Fintype.card α`):

```lean
def degreeSequenceCompact (G : SimpleGraph α) [DecidableRel G.Adj] : Prop :=
  IsCompactSequenceOn (fun k => (degreeSequence G).getD k 0) {k | k + 2 < Fintype.card α}
```

This is supposed to ensure  that the condition `d(k+2) = d(k) + 1` is only enforced for relevant indices, preventing the issue where the condition would fail (or be trivial/wrong) for large k where the degree sequence values would default to `0`.

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: Thank you,  I think that this change got merged as suggsted.

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: Good point. To make it moot I am including a proof, hence it should be automatically skipped by systems looking for "sorry". Let me know whether this is sufficient.

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: I think that it got merged, thank you!

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: I think that it got merged, thank you!

.gitignore
henrykmichalewski: For a short moment I mistakenly assumed that all files in `scripts` are temporary files added by me. This is corrected now.

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: I think that it got merged, thank you!

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
henrykmichalewski: Thank you, it is now replaced with mindegree.

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,189 @@
...
+
+/-- **Lemma 2 (b)**
+Inequality involving sums of terms of a nondecreasing sequence with no three terms equal. -/
+@[category research solved, AMS 5]
+lemma lemma2_b
+    (h_mono : Monotone d)
+    (h_pos : ∀ k, 0 < d k)
+    (h_no_three : ∀ i, d (i + 2) ≠ d i) :
+    2 * n * n + 2 * n + 1 ≤
+      (∑ i ∈ .Icc (2 * n + 1) (4 * n + 1), d i) -
```
YaelDillies:
```suggestion
      ∑ i ∈ .Icc (2 * n + 1) (4 * n + 1), d i -
```
etc

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,189 @@
...
+
+variable {α : Type*} [Fintype α] [DecidableEq α]
+
+/-- The number of vertices of `G` having degree `d`. -/
+noncomputable def degreeFreq (G : SimpleGraph α) (d : ℕ) : ℕ :=
+  #{v | G.degree v = d}
+
+
+
+end SimpleGraph
```
YaelDillies:
```suggestion
end SimpleGraph
```

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,189 @@
...
+
+section lemmas
+
+variable (d : ℕ → ℕ) (n k r : ℕ)
+
+/-- **Lemma 1 (a)**
+If a sequence `d` is nondecreasing and no three terms are equal, then terms at distance 2 differ by at least 1. -/
+@[category research solved, AMS 5]
```
YaelDillies: I don't think we should mark lemmas as `research solved` because that means "solved conjecture". The fact that `a \le b` and `a \ne b` implies `a < b` was never a conjecture

What about instead
```suggestion
/-- **Lemma 1 (a)**
If a sequence `d` is nondecreasing and no three terms are equal, then terms at distance 2 differ by at least 1. -/
@[category API, AMS 5]
```

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,189 @@
...
+
+/-- **Lemma 2 (a)**
+Inequality involving sums of terms of a nondecreasing sequence with no three terms equal. -/
+@[category research solved, AMS 5]
+lemma lemma2_a
+    (h_mono : Monotone d)
+    (h_pos : ∀ k, 0 < d k)
+    (h_no_three : ∀ i, d (i + 2) ≠ d i) :
+    2 * n * n ≤
+      (∑ i ∈ .Icc (2 * n + 1) (4 * n), d i) -
```
YaelDillies:
```suggestion
      ∑ i ∈ .Icc (2 * n + 1) (4 * n), d i -
```

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,189 @@
...
+
+variable {α : Type*} [Fintype α] [DecidableEq α]
+
+/-- The degree sequence of a graph, sorted in nondecreasing order. -/
+noncomputable def degreeSequence (G : SimpleGraph α) [DecidableRel G.Adj] : List ℕ :=
+  (Finset.univ.val.map fun v : α => G.degree v).sort (· ≤ ·)
+
+/-- The degree sequence of `G` is **compact** if it satisfies
+`IsCompactSequenceOn` for all valid indices `k` such that `k + 2 < Fintype.card α`. -/
+def degreeSequenceCompact (G : SimpleGraph α) [DecidableRel G.Adj] : Prop :=
```
YaelDillies:
```suggestion
def HasCompactdegreeSequence (G : SimpleGraph α) [DecidableRel G.Adj] : Prop :=
```
instead maybe?

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,187 @@
...
+    (h_mono : Monotone d)
+    (h_pos : ∀ k, 0 < d k)
+    (h_no_three : ∀ k, d (k + 2) ≠ d k) :
+    1 ≤ d (k + 2) - d k := by
+  rw [Nat.one_le_iff_ne_zero]
+  intro h
+  rw [Nat.sub_eq_zero_iff_le] at h
+  have h' : d k ≤ d (k + 2) := h_mono (le_add_of_nonneg_right (Nat.zero_le 2))
+  have h_eq : d (k + 2) = d k := le_antisymm h h'
+  exact h_no_three k h_eq
```
YaelDillies:
```suggestion
  have : d k ≤ d (k + 2) := h_mono (by omega)
  have := h_no_three k
  omega
```

FormalConjectures/Paper/DegreeSequencesTriangleFree.lean
```diff
@@ -0,0 +1,187 @@
...
+
+/-- **Lemma 3.** For every `n` there exists a bipartite graph with
+`8 n` vertices, minimum degree `n + 1`, and `f = 3`. -/
+@[category API, AMS 5]
+lemma lemma3 (n : ℕ) (hn : 0 < n) :
+    ∃ (G : SimpleGraph (Fin (8 * n))) (_ : DecidableRel G.Adj),
+      G.IsBipartite ∧ G.minDegree = n + 1 ∧ f G = 3 := by
+  sorry
+
+/-- **Lemma 4.** A triangle-free graph can be extended by one to a
```
YaelDillies: Extended by one what? By one vertex or by one part?

henrykmichalewski: The original statement of Lemma 4 from the paper (this is OCR)):

Lemma 4. Let $G$ be a triangle-free graph with $n$ vertices and let $v$ be a vertex of $G$. There exists a triangle-free graph $H$ containing $G$ as an induced subgraph such that:
(i) the degree of $v$ in $H$ is one more than its degree in $G$;
(ii) for every vertex $w$ of $G$ other than $v$ the degree of $w$ in $H$ is the same as its degree in $G$;
(iii) if $J$ is the subgraph of $H$ induced by the vertices not in $G$, then $f(J)=3$ and $\delta(J) \geqslant 2 n$.

YaelDillies: Yeah, that statement makes more sense to me

PR #350, opened 2025-07-11
Erdos 208

### LINKED ISSUE #51: Formalise Erdos Problem 208

callesonne:
closes #51.

### DIFF
--- a/FormalConjectures/ErdosProblems/208.lean
+++ b/FormalConjectures/ErdosProblems/208.lean
@@ -23,25 +23,32 @@ import FormalConjectures.Util.ProblemImports
 
 open Filter Real
 
--- TODO: Need to fix wording on website...
+/-- The sequence of squarefree numbers, denoted by `s` as in Erdős problem 208. -/
+noncomputable def erdos208.s : ℕ → ℕ := Nat.nth Squarefree
+
+open erdos208
 
 /--
-Let `s1 < s2 < ⋯` be a sequence of squarefree numbers. Is it true that, for any `ϵ > 0` and large `n`,
-`s (n+1) − s n ≪_ϵ (s n)^ε`?
+Let `s1 < s2 < ⋯` be the sequence of squarefree numbers. Is it true that,
+for any `ϵ > 0` and large `n`, `s (n+1) − s n ≪_ϵ (s n)^ε`?
 -/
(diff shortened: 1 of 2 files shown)


FormalConjectures/ErdosProblems/208.lean
```diff
@@ -0,0 +1,56 @@
...
+
+open erdos208
+
+/--
+Let $s_1 < s_2 < ⋯$ be the sequence of squarefree numbers. Is it true that,
+for any $ϵ > 0$ and large $n$, $s_{n+1} − s_n ≪_ϵ s_n^ε$?
+-/
+@[category research open, AMS 11]
+theorem erdos_208.i :
+    (∀ ε > 0, ∃ C > 0, ∀ᶠ n in atTop, s (n + 1) - s n < C * (s n)^ε) ↔ answer(sorry) := by sorry
```
Paul-Lez: Could this be written using the big O notation?

mo271: I think it is fine as is

FormalConjectures/ErdosProblems/208.lean
```diff
@@ -0,0 +1,56 @@
...
+
+open Filter Real
+
+/-- The sequence of squarefree numbers, denoted by `s` as in Erdős problem 208. -/
+noncomputable def erdos208.s : ℕ → ℕ := Nat.nth Squarefree
+
+open erdos208
+
+/--
+Let $s_1 < s_2 < ⋯$ be the sequence of squarefree numbers. Is it true that,
```
mo271:
```suggestion
Let $s_1 < s_2 < ⋯$ be the sequence of squarefree numbers. Is it true that
```
for consistency with the similar sentence below

FormalConjectures/ErdosProblems/208.lean
callesonne: It can yes, but the issue is that you want `C` to depend on `ε`, so you need to use `IsBigOWith`, which basically has no API anyway.

Paul-Lez: Wouldn't the dependence be automatic if you write 
```lean
∀ ε > 0, (fun n => s (n + 1) - s n) =O[atTop] (fun n => (s n)^ε)
``` 
?

callesonne: True! Then I will change this later :)

PR #355, opened 2025-07-15
Variant for infinitude of Fibonacci primes.

mo271:
Also restructure documentation a bit.

### DIFF
--- a/FormalConjectures/Wikipedia/FibonacciPrimes.lean
+++ b/FormalConjectures/Wikipedia/FibonacciPrimes.lean
@@ -16,13 +16,46 @@ limitations under the License.
 
 import FormalConjectures.Util.ProblemImports
 
+/-!
+# Fibonacci Primes
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Fibonacci_prime)
+-/
+
 /--
-This conjecture states that there are infinitely many Fibonacci primes
-(i.e., Fibonacci numbers that are prime). It is unsolved. It is
-stated on Wikipedia here: https://en.wikipedia.org/wiki/Fibonacci_prime.
+There are infinitely many Fibonacci primes, i.e., Fibonacci numbers that are prime
 It is also a barrier to defining a benchmark from this paper:
 https://arxiv.org/html/2505.13938v1 (see Figure 8).
+
 -/
 @[category research open, AMS 11]
-theorem fib_primes_infinite : {n : ℕ | n.fib.Prime}.Infinite := by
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/FibonacciPrimes.lean
```diff
@@ -16,13 +16,46 @@ limitations under the License.
...
-theorem fib_primes_infinite : {n : ℕ | n.fib.Prime}.Infinite := by
+theorem fib_primes_infinite.variant : {n : ℕ | n.fib.Prime}.Infinite := by
   sorry
+
+/--
+The two ways of phrasing the conjecture are equivalent.
+-/
+@[category test, AMS 11]
+theorem indices_infinite_iff_fib_primes_infinite : type_of% fib_primes_infinite.variant ↔
+    type_of%fib_primes_infinite := by
```
callesonne:
```suggestion
    type_of% fib_primes_infinite := by
```

FormalConjectures/Wikipedia/FibonacciPrimes.lean
```diff
@@ -16,13 +16,46 @@ limitations under the License.
...
+@[category research open, AMS 11]
+theorem fib_primes_infinite : {n : ℕ | (∃ m : ℕ, m.fib = n) ∧ n.Prime}.Infinite := by
+  sorry
+
+/--
+There are infinitely many indices $i$, such that the $i$-th Fibonacci is prime.
 -/
 @[category research open, AMS 11]
-theorem fib_primes_infinite : {n : ℕ | n.fib.Prime}.Infinite := by
+theorem fib_primes_infinite.variant : {n : ℕ | n.fib.Prime}.Infinite := by
```
callesonne, line 39: So how come you want this to be the variant?

mo271: I don't have a preference, this was more accidental. 
What names would you suggest?

callesonne: No, this doesn't matter so much to me either, but I guess what is your motivation for introducing the new variant?

mo271: see discussion at #264

mo271: it would be even more obvious that the two statements are equivalent if Nat.fib would be injective, but it isn't because `Nat.fib 1 = 1 = Nat.fib 2`

PR #432, opened 2025-07-18
Erdős Problem 128

### LINKED ISSUE #367: Erdős Problem 128: Graphs such that every large subgraph has enough edges contain a triangle
### What is the conjecture
    
https://www.erdosproblems.com/128

  Let $G$ be a graph with $n$ vertices such that every subgraph on $\geq n/2$ vertices has more than $n^2/50$ edges. Must $G$ contain a triangle?

Status: open

### Choose either option

rdivyanshu:
Fixes #367

### DIFF
--- a/FormalConjectures/ErdosProblems/128.lean
+++ b/FormalConjectures/ErdosProblems/128.lean
@@ -1,0 +1,37 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/128.lean
```diff
@@ -0,0 +1,37 @@
...
+
+/--
+Let G be a graph with n vertices such that every subgraph on ≥ $n/2$
+vertices has more than $n^2/50$ edges. Must G contain a triangle?
+-/
+@[category research open, AMS 5]
+theorem erdos_128 :
+    ((∀ (G' : G.Subgraph) [Fintype G'.verts] [Fintype G'.edgeSet],
+        2 * G'.verts.toFinset.card ≥ Fintype.card V →
+        50 * G'.edgeSet.toFinset.card > (Fintype.card V)^2) → ¬ (G.CliqueFree 3))
```
mo271:
```suggestion
        letI n := Fintype.card V;
        2 * G'.verts.toFinset.card ≥ n →
        50 * G'.edgeSet.toFinset.card > n^2) → ¬ (G.CliqueFree 3))
```
I suppose we could define `n` here, but this is more a stylistic choice, feel free to ignore it.

rdivyanshu: Thanks for suggestion. Fixed.

PR #442, opened 2025-07-18
Erdős Problem 251

mo271:


FormalConjectures/ErdosProblems/251.lean
```diff
@@ -0,0 +1,32 @@
...
+
+/-!
+# Erdős Problem 251
+
+*Reference:* [erdosproblems.com/251](https://www.erdosproblems.com/251)
+-/
+
+/--
+Is
+$$\sum_{n} \frac{φ_n}{2^n}$$
```
Paul-Lez:
```suggestion
$$\sum_{n} \frac{p_n}{2^n}$$
```

Paul-Lez (approved):
LGTM. Maybe worth adding a comment that the sum we're claiming is irrational is half the sum in the original formulation of the problem but that these are of course trivially equivalent?

mo271:
@Paul-Lez added a remark on the index shift

Paul-Lez:
> @Paul-Lez added a remark on the index shift

LGMT!

Issue #441, opened 2025-07-18
Complex structure on 6-sphere


peabrainiac:
### What is the conjecture
It is a famous open question whether the 6-sphere $S^6$ admits a complex structure, i.e. whether it can be turned into a complex manifold. Over the years several papers have appeared claiming to resolve the conjecture either way, but as far as I can tell there is still no solution that is widely accepted as correct. See e.g. [this MO question](https://mathoverflow.net/q/1973/503660).

### Prerequisites needed
Nothing; the 6-sphere is in Mathlib in the form of `Metric.sphere (0 : EuclideanSpace ℝ (Fin 7)) 1`, and complex structures on it should be easy to talk about with Mathlib's manifold API.

### [AMS categories](https://github.com/google-deepmind/formal-conjectures/labels?q=ams-)

* ams-32
### Choose either option
- [x] I plan on working on this conjecture
- [ ] This issue is up for grabs: I would like to see this conjecture added by somebody else

PR #447, opened 2025-07-18
Complex structure on 6-sphere

peabrainiac:
Adds the open question of whether $S^6$ admits a complex structure, closing #441.

### DIFF
--- a/FormalConjectures/Mathoverflow/1973.lean
+++ b/FormalConjectures/Mathoverflow/1973.lean
@@ -1,0 +1,41 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+open scoped Manifold
+
+/-!
+# Mathoverflow 1973
+
+Does the 6-sphere $S^6$ admit the structure of a complex manifold?
+
+*Reference:* [mathoverflow/1973](https://mathoverflow.net/questions/1973/)
+asked by user [*Fetchinson0234*](https://mathoverflow.net/users/41312/victor-ramos)
+-/
+
+/-- The unit `n`-sphere, defined as `Metric.sphere 0 1` in `EuclideanSpace ℝ (Fin (n + 1))`. -/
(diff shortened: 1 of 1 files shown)


peabrainiac:
[This answer](https://mathoverflow.net/a/1984/503660) to the referenced MathOverflow question points out that if such a complex structure existed, it would in particular imply the existence of a non-standard complex structure on $\mathbb CP^3$, by blowup of $S^6$ at a point. While this implication would be hard to formalise (I don't even know mathematically why the resulting complex structure on $\mathbb CP^3$ would have to be nonstandard), the statement that there exists a nonstandard complex structure on $\mathbb CP^3$ should be easy to formalise - we don't have the standard smooth structure on it yet, but could simply ask if there are two smooth structures on it that are not complex-diffeomorphic. So maybe it would make sense to add this as a variant of the question, though I'm not sure because I haven't really seen exotic $\mathbb CP^3$'s mentioned as an open problem before anywhere else.

FormalConjectures/Mathoverflow/1973.lean
```diff
@@ -0,0 +1,41 @@
...
+# Mathoverflow 1973
+
+Does the 6-sphere $S^6$ admit the structure of a complex manifold?
+
+*Reference:* [mathoverflow/1973](https://mathoverflow.net/questions/1973/)
+asked by user [*Fetchinson0234*](https://mathoverflow.net/users/41312/victor-ramos)
+-/
+
+/-- The unit `n`-sphere, defined as `Metric.sphere 0 1` in `EuclideanSpace ℝ (Fin (n + 1))`. -/
+abbrev unitSphere (n : ℕ) := Metric.sphere (0 : EuclideanSpace ℝ (Fin (n + 1))) 1
```
smmercuri:
```suggestion
abbrev unitSphere (n : ℕ) : Set (EuclideanSpace ℝ (Fin (n + 1)) := Metric.sphere 0 1
```

FormalConjectures/Mathoverflow/1973.lean
```diff
@@ -0,0 +1,41 @@
...
+
+/--
+Does the 6-sphere admit a complex structure, i.e. an atlas of holomorphically compatible charts
+relating it to `EuclideanSpace ℂ (Fin 3)`?
+-/
+@[category research open, AMS 32]
+theorem mathoverflow_1973 :
+    (∃ atlas : ChartedSpace (EuclideanSpace ℂ (Fin 3)) (unitSphere 6),
+      @IsManifold _ _ _ _ _ _ _ 𝓘(ℂ, _) 1 _ _ atlas) ↔ answer(sorry) := by
```
smmercuri: You can avoid this `@` if you remove the hole in `𝓘(ℂ, _)`
```suggestion
      IsManifold 𝓘(ℂ, EuclideanSpace ℂ (Fin 3)) 1 (unitSphere 6)) ↔ answer(sorry) := by
```

peabrainiac: I keep forgetting that instances are picked up on by typeclass synthesis even if they're just around in the local context and not actually registered as instances :)
Well-spotted, thanks!

smmercuri (approved):
LGTM, thank you!

PR #450, opened 2025-07-22
Add Erdős Problem 236

HKalbasi:


FormalConjectures/ErdosProblems/236.lean
```diff
@@ -0,0 +1,48 @@
...
+
+*Reference:* [erdosproblems.com/236](https://www.erdosproblems.com/236)
+-/
+
+open Filter Asymptotics
+
+/-- 
+$f(n)$ counts the number of solutions to $n=p+2^k$ for prime $p$ and $k\geq 0$.
+-/
+def f (n : ℕ) : ℕ :=
```
Paul-Lez:
```suggestion
private def f (n : ℕ) : ℕ :=
```

FormalConjectures/ErdosProblems/236.lean
```diff
@@ -0,0 +1,48 @@
...
+  ((List.range (Nat.log2 n + 1)).filter (fun k => Nat.Prime (n - 2^k))).length
+
+/--
+Let $f(n)$ count the number of solutions to $n=p+2^k$ for prime $p$ and $k\geq 0$. Show that $f(n)=o(\log n)$.
+-/
+@[category research open, AMS 5 11]
+theorem erdos_236: (fun n : ℝ => (f (Nat.ceil n) : ℝ)) =o[ atTop ] Real.log := by
```
Paul-Lez: I would suggest this in order to stay a little closer to the informal formulation of the problem.
```suggestion
theorem erdos_236: (fun n => (f n : ℝ)) =o[atTop] Real.log (n : ℝ) := by 
```

FormalConjectures/ErdosProblems/236.lean
```diff
@@ -0,0 +1,48 @@
...
+  sorry
+
+/--
+Numbers $n > 2$ such that $n - 2^k$ is a prime for all $k > 0$ with $2^k < n$. 
+Erdős conjectures that these are the only values of n with this property.
+
+https://oeis.org/A039669
+-/
+@[category research open, AMS 5 11]
+theorem erdos_236.related: ∀ (n : ℕ), n > 2 → f n = Nat.log2 n → n ∈ [4, 7, 15, 21, 45, 75, 105] := by
```
Paul-Lez: This doesn't seem to match the informal statement -  would you mind adding a sanity check statement (using `category test`) that the two are equivalent?

hkalbasi: I can even remove this since it is probably a trivial corollary of the above + some computer analysis to rule out small numbers. What do you prefer? Removing it, rewriting it to be closer to the informal statement, or writing a test (I need to read more about this one)?

Paul-Lez: I think keeping it is fine! If this is easily solvable, I would also consider another `category` value (e.g `research solved`, or even `high_school`/`undergraduate`/`graduate`, depending on the difficulty)

hkalbasi: It's not solved, but it is checked with computer to be true until 1e20, so if a proof of the erdos_236 shows that `f n` is strictly less than `log n` for a number sufficiently large then this will follow.

hkalbasi: I removed it to unblock this PR.

Issue #239, opened 2025-06-16
Inscribed square problem


peabrainiac:
### What is the conjecture
The [inscribed square problem](https://en.wikipedia.org/wiki/Inscribed_square_problem) or Toeplitz conjecture asks if / states that every Jordan curve (i.e. simple closed curve in $\mathbb R^2$) admits an inscribed square, i.e. a square whose vertices all lie on the curve. The general case of this is still open, but there are also many solved variants for special classes of maps that could be interesting to state here too: for example, it was shown in 2020 [here](https://arxiv.org/abs/2005.09193) that *smooth* Jordan curves admit inscribed rectangles of any given aspect ratio (in particular squares for the aspect ratio 1:1), and [in 2023](https://arxiv.org/abs/2011.05216) this was even generalised to inscribed cyclic quadrilaterals. There is also a classic proof from 1977 that every Jordan curve admits some inscribed rectangle, just with no way to control the aspect ratio; you might have seen a video about it [here](https://www.youtube.com/watch?v=AmgkSdhK4K8).

This conjecture was already listed in #104, I'm just opening this issue to add some further information. This should really be a sub-issue, but I couldn't see how to open one.

### Prerequisites needed
[...]

PR #451, opened 2025-07-23
Inscribed square problem

peabrainiac:
Adds the inscribed square problem and several variants, closing #239.

### DIFF
--- a/FormalConjectures/Wikipedia/InscribedSquare.lean
+++ b/FormalConjectures/Wikipedia/InscribedSquare.lean
@@ -1,0 +1,81 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Inscribed square problem
+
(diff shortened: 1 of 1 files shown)


peabrainiac:
The survey paper I listed as a reference also contains many more variants of the problem, for example on inscribed triangles, inscribed circular quadrilaterals and the number of inscribed squares. I've kept them out of this PR for now, but they might be interesting future additions.

FormalConjectures/Wikipedia/InscribedSquare.lean
```diff
@@ -0,0 +1,81 @@
...
+theorem exists_inscribed_rectangle (γ : Circle → EuclideanSpace ℝ (Fin 2)) (hγ : IsEmbedding γ) :
+    ∃ t₁ t₂ t₃ t₄ r, IsRectangle (γ t₁) (γ t₂) (γ t₃) (γ t₄) r :=
+  sorry
+
+/--
+It is known that every *smooth* Jordan curve admits inscribed rectangles of all aspect ratios.
+-/
+@[category research solved, AMS 51]
+theorem exists_inscribed_rectangle_of_smooth (γ : Circle → EuclideanSpace ℝ (Fin 2))
+    (hγ : IsEmbedding γ) (hγ' : ContMDiff (𝓡 1) (𝓡 2) ∞ γ) (r : ℝ) (hr : r > 0) :
```
Paul-Lez, line 79: Out of curiosity, is anything similar known for squares when assuming sufficient regularity of the curve? Could possibly be a nice extra problem to add;)

Paul-Lez (commented):
Very cool stuff!

Paul-Lez (approved):

FormalConjectures/Wikipedia/InscribedSquare.lean
peabrainiac: Existence of inscribed squares in smooth Jordan curves is a special case of this simply by setting `r := 1`. I think inscribed squares are also known to exist for all C^2 curves rather than just smooth ones; this might be an interesting addition. I'm currently travelling though and so can't add it until I'm home next week.

Paul-Lez: Cool! Let's get this merged, we can always open a new one with more stuff later!

PR #452, opened 2025-07-26
Erdős Problem 212

### LINKED ISSUE #424: Erdős Problem 212: dense subset of R² with only rational distances
### What is the conjecture
    
https://www.erdosproblems.com/212

  Is there a dense subset of $\mathbb{R}^2$ such that all pairwise distances are rational?

Status: open

### Choose either option

rdivyanshu:
Fixes #424

### DIFF
--- a/FormalConjectures/ErdosProblems/212.lean
+++ b/FormalConjectures/ErdosProblems/212.lean
@@ -28,4 +28,4 @@ are rational?
 -/
 @[category research open, AMS 52]
 theorem erdos_212 : (∃ u : Set ℂ,
-  Dense u ∧ (∀ c₁ ∈ u, ∀ c₂ ∈ u, ¬ Irrational ‖c₁ - c₂‖)) ↔ answer(sorry) := by sorry
+  Dense u ∧ u.Pairwise fun c₁ c₂ => dist c₁ c₂ ∈ Set.range Rat.cast) ↔ answer(sorry) := by sorry


FormalConjectures/ErdosProblems/212.lean
```diff
@@ -0,0 +1,31 @@
...
+*Reference:* [erdosproblems.com/212](https://www.erdosproblems.com/212)
+-/
+
+/--
+Is there a dense subset of ℝ^2 such that all pairwise distances
+are rational?
+-/
+@[category research open, AMS 52]
+theorem erdos_212 : (∃ u : Set ℂ,
+  Dense u ∧ (∀ c₁ ∈ u, ∀ c₂ ∈ u, ¬ Irrational ‖c₁ - c₂‖)) ↔ answer(sorry) := by sorry
```
mo271:
```suggestion
@[category research open, AMS 52]
theorem erdos_212 : (∃ u : Set ℂ, Dense u ∧
    u.Pairwise fun c₁ c₂ => dist c₁ c₂ ∈ Set.range Rat.cast) ↔ answer(sorry) := by sorry
```

Using "pairwise" to be closer to the informal statement, although it doesn't make a difference of course, since the distance between two equal points is rational anyhow. 
Also avoiding the double negation of "not irrational" to describe a rational number, but this is a question of taste I suppose.

mo271 (approved):
LGTM, thanks!

FormalConjectures/ErdosProblems/212.lean
rdivyanshu: Thanks for review. Happy to incorporate both suggestions

PR #455, opened 2025-07-28
feat: formalize matrix group conjecture

### LINKED ISSUE #152: A conjecture about matrix groups
### What is the conjecture
Conjecture: for any g in G/Γ such that Dg is [relatively compact](https://en.wikipedia.org/wiki/Relatively_compact) (in G/Γ), Dg is closed. Here G = SLn(R), Γ = SLn(Z), and the subgroup D of diagonal matrices in G.

Reference: [Littlewood conjecture - Connection to further conjecture](https://en.wikipedia.org/wiki/Littlewood_conjecture#Connection_to_further_conjectures)

### Prerequisites needed
I think this should be doable in the current state of Mathlib.

* ams-22
* ams-15
* ams-11

### Choose either option

Paul-Lez:
Fixes #152.


FormalConjectures/Wikipedia/LittlewoodConjecture.lean
```diff
@@ -43,3 +43,49 @@ to the nearest integer.
...
+
+section MatrixGroupConjecture
+
+open Matrix
+open scoped MatrixGroups
+
+section
+variable (n : Type*) [DecidableEq n] [Fintype n] (R : Type*) [CommRing R]
+
+private def Matrix.diagonalHom : (n → Rˣ) →* GeneralLinearGroup n R where
```
smmercuri: There's some `GL` notation for `GeneralLinearGroup`

smmercuri: I feel this `def` could be made public and put in a `ForMathlib` file

smmercuri: Oh ignore me, that's for `Fin n` only

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
```diff
@@ -43,3 +43,49 @@ to the nearest integer.
...
+/-- The group of invertible diagonal matrices. -/
+private def DiagonalMatricesMultiplicative : Subgroup (GeneralLinearGroup n R) :=
+  Subgroup.map (Matrix.diagonalHom n R) ⊤
+
+/-- The group of invertible diagonal matrices with determinant 1. -/
+private def SpecialDiagonalMatrices : Subgroup (SpecialLinearGroup n R) :=
+  (DiagonalMatricesMultiplicative n R).comap Matrix.SpecialLinearGroup.toGL
+
+-- TODO(Paul-Lez): do we want to upstream this to Mathlib?
+instance : TopologicalSpace (SpecialLinearGroup n ℝ) :=
```
smmercuri: Maybe `Matrix.SpecialLinearGroup` should be an `abbrev` to avoid this, like `Matrix.GeneralLinearGroup` is

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
```diff
@@ -43,3 +43,49 @@ to the nearest integer.
...
+section
+variable (n : Type*) [DecidableEq n] [Fintype n] (R : Type*) [CommRing R]
+
+private def Matrix.diagonalHom : (n → Rˣ) →* GeneralLinearGroup n R where
+  toFun := fun d ↦ ⟨Matrix.diagonal (fun i ↦ d i), Matrix.diagonal (fun i ↦ (d i).inv), by simp, by simp⟩
+  map_mul' x y := by ext; simp [-mul_diagonal, -diagonal_mul]
+  map_one' := by ext; simp
+
+/-- The group of invertible diagonal matrices. -/
+private def DiagonalMatricesMultiplicative : Subgroup (GeneralLinearGroup n R) :=
```
smmercuri: Ditto above, I think this can be public and named something like `Matrix.GeneralLinearGroup.diagonalSubgroup`

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
```diff
@@ -43,3 +43,49 @@ to the nearest integer.
...
+  map_mul' := by simp
+  map_one' := by simp
+
+end
+
+/-- Let `n ≥ 3`, `G = SL_n(ℝ)` and `Γ = SL_(ℤ)`, and the subgroup `D` of diagonal matrices in `G`.
+
+Conjecture: for any `g` in `G/Γ` such that `Dg` is relatively compact (in `G/Γ`), then `Dg` is closed.-/
+@[category research open]
+theorem matrix_group_conjecture {n : ℕ} (hm : 3 ≤ n) (g : SL(n, ℝ))
```
smmercuri:
```suggestion
theorem matrix_group_conjecture {n : ℕ} (hn : 3 ≤ n) (g : SL(n, ℝ))
```

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
```diff
@@ -43,3 +43,49 @@ to the nearest integer.
...
+    (Dg : Set (SL(n, ℝ) ⧸ (⊤ : Subgroup _).map Matrix.SpecialLinearGroup.coeIntGroupHom))
+    (Dg_def : Dg = { QuotientGroup.mk (d * g) | (d : SL(n, ℝ)) (_ : d ∈ SpecialDiagonalMatrices _ _) })
```
smmercuri: Could we define an appropriate `MulAction` then we can just use `MulAction.orbit` to avoid a separate `Dg` def here?

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
```diff
@@ -43,3 +43,49 @@ to the nearest integer.
...
+  toFun x := ↑x
+  map_mul' := by simp
+  map_one' := by simp
+
+end
+
+/-- Let `n ≥ 3`, `G = SL_n(ℝ)` and `Γ = SL_(ℤ)`, and the subgroup `D` of diagonal matrices in `G`.
+
+Conjecture: for any `g` in `G/Γ` such that `Dg` is relatively compact (in `G/Γ`), then `Dg` is closed.-/
+@[category research open]
```
smmercuri: nit: AMS tag?

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
```diff
@@ -43,3 +43,49 @@ to the nearest integer.
...
+
+-- TODO(Paul-Lez): do we want to upstream this to Mathlib?
+instance : TopologicalSpace (SpecialLinearGroup n ℝ) :=
+  inferInstanceAs (TopologicalSpace { A : Matrix n n ℝ // A.det = 1 })
+
+variable {n R} in
+def Matrix.SpecialLinearGroup.coeIntGroupHom : SpecialLinearGroup n ℤ →* SpecialLinearGroup n R where
+  toFun x := ↑x
+  map_mul' := by simp
+  map_one' := by simp
```
smmercuri: this is `SpecialLinearGroup.map (Int.castRingHom R)`

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
Paul-Lez: Oh I should write a linter for this

Paul-Lez: Done in #516

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
Paul-Lez: Opened a Mathlib PR about this: https://github.com/leanprover-community/mathlib4/pull/27596.

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
Paul-Lez: Done.

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
```diff
@@ -43,3 +43,42 @@ to the nearest integer.
...
+-- This instance can be made more general (this is being done in https://github.com/leanprover-community/mathlib4/pull/27596)
+instance : TopologicalSpace (SpecialLinearGroup n ℝ) :=
+  inferInstanceAs (TopologicalSpace { A : Matrix n n ℝ // A.det = 1 })
+
+end
+
+open Pointwise
+
+/-- `SL_n(ℤ)`, viewed as a subgroup of `SL_n(ℝ)`-/
+private abbrev SpecialLinearGroupIntSubgroup (n : ℕ) : Subgroup SL(n, ℝ) :=
```
smmercuri: This `def` isn't actually used, I see it's inlined in the below theorem. Let's remove it?

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
```diff
@@ -43,3 +43,42 @@ to the nearest integer.
...
+
+/-
+Formalisation note: the original wiki page seems uses left quotients, but it seems (based on
+the existing litterature that one should be taking right quotients. This is what has been done
+here).
+-/
+
+/-- Let `n ≥ 3`, `G = SL_n(ℝ)` and `Γ = SL_n(ℤ)`, and the subgroup `D` of diagonal matrices in `G`.
+
+Conjecture: for any `g` in `Γ\G` such that `Dg` is relatively compact (in `Γ\G`), then `Dg` is closed.-/
```
smmercuri: Is there a reference for this?

Paul-Lez: The only reference I know of is the wiki page for the Littlewood conjecture

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
```diff
@@ -43,3 +43,42 @@ to the nearest integer.
...
+
+/-- `SL_n(ℤ)`, viewed as a subgroup of `SL_n(ℝ)`-/
+private abbrev SpecialLinearGroupIntSubgroup (n : ℕ) : Subgroup SL(n, ℝ) :=
+  (⊤ : Subgroup _).map (SpecialLinearGroup.map (Int.castRingHom ℝ))
+
+/-
+Formalisation note: the original wiki page seems uses left quotients, but it seems (based on
+the existing litterature that one should be taking right quotients. This is what has been done
+here).
+-/
```
smmercuri: I think what's happening here is the double coset `Γ\G/D` so there's both a right and left action. This makes me wonder if it would be simpler to state this result using [`DoubleCoset`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/GroupTheory/DoubleCoset.html#DoubleCoset.mk) somehow

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
smmercuri: Found one here https://arxiv.org/pdf/2504.17644v3

smmercuri: I think that should be `G / Γ` in the docstring

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
smmercuri: Looking at the above reference, we are looking at `D \ G / Γ` and right orbits `Dg`. But Probably we don't need to actually use `DoubleCoset`

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
```diff
@@ -43,3 +43,42 @@ to the nearest integer.
...
+
+/-- Let `n ≥ 3`, `G = SL_n(ℝ)` and `Γ = SL_n(ℤ)`, and the subgroup `D` of diagonal matrices in `G`.
+
+Conjecture: for any `g` in `Γ\G` such that `Dg` is relatively compact (in `Γ\G`), then `Dg` is closed.-/
+@[category research open, AMS 11 15 22]
+theorem matrix_group_conjecture {n : ℕ} (hn : 3 ≤ n) (g : MulOpposite SL(n, ℝ))
+    (D : Set (MulOpposite SL(n, ℝ) ⧸
+      ((⊤ : Subgroup _).map (SpecialLinearGroup.map (Int.castRingHom ℝ))).op))
+    (D_def : D = QuotientGroup.mk '' (Matrix.SpecialLinearGroup.diagonalSubgroup _ _).op.carrier)
+    (hg : IsCompact <| closure (g • D)) : IsClosed (g • D) :=
```
smmercuri: I think this works (?), avoids the need for `MulOpposite` and the `D` definition
```suggestion
theorem matrix_group_conjecture {n : ℕ} (hn : 3 ≤ n)
    (g : SL(n, ℝ) ⧸ Subgroup.map (map (Int.castRingHom ℝ)) ⊤)
    (hg : IsCompact <| closure (MulAction.orbit (diagonalSubgroup (Fin n) ℝ) g)) :
    IsClosed <| MulAction.orbit (diagonalSubgroup (Fin n) ℝ) g := 
```

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
Paul-Lez: Nice find for the ref!  In that case, I think the conjecture should really be in the arXiv subdirectory.

FormalConjectures/Wikipedia/LittlewoodConjecture.lean
Paul-Lez: Yeah I guess if we don't actually have to deal with right actions in the end things do get easier!

FormalConjectures/Arxiv/2504.17644/Margulis.lean
```diff
@@ -0,0 +1,54 @@
...
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# A conjecture by Margulis on matrix groups
+
+
+-/
```
smmercuri:
```suggestion
/-!
# A conjecture by Margulis on matrix groups

*Reference:* [arxiv/2504.17644](https://arxiv.org/pdf/2504.17644v3)
**Bounded diagonal orbits in homogeneous spaces over function fields**
by *Qianlin Huang, Ronggang Shi*
-/
```

smmercuri (approved):

PR #462, opened 2025-07-28
Erdős Problem 269

mo271:


FormalConjectures/ErdosProblems/269.lean
```diff
@@ -0,0 +1,73 @@
...
+/--
+Let $P$ be a finite set of primes with $|P| \ge 2$ and let
+$\{a_1 < a_2 < \dots\}$ be the set of positive integers whose prime factors
+are all in $P$. Is the sum
+$$ \sum_{n=1}^\infty \frac{1}{[a_1,\ldots,a_n]} $$
+rational or irrational?
+-/
+@[category research open, AMS 11]
+theorem erdos_269 : (∀ᵉ (P : Finset ℕ) (h : ∀ p ∈ P, p.Prime) (h_card : P.card ≥ 2),
+    (Irrational (series (P : Set ℕ)) ↔ answer(sorry))) ↔ answer(sorry) := by
```
mo271: the two `answer`s here is not a typo, but I thought it would reflect the two questions asked here:
 - is it always rational?
 - is it always irrational?
 This way one can (dis)prove both by choosing the sorrys correctly...

mo271: feeback welcome!

FormalConjectures/ErdosProblems/269.lean
```diff
@@ -0,0 +1,73 @@
...
+-/
+noncomputable def a (P : Set ℕ) : ℕ → ℕ := Nat.nth <| HasPrimeFactorsIn P
+
+/--
+The `n`-th partial least common multiple, $[a_0, \dots, a_n]$, which is
+the LCM of the first `n` integers in the sequence.
+-/
+noncomputable def partialLcm (P : Set ℕ) (n : ℕ) : ℕ :=
+  -- We take the LCM of `{a P 0, ..., a P n}`.
+  (Finset.range n).image (a P) |>.lcm id
```
Paul-Lez: I think this also works?
```suggestion
  (Finset.range n).lcm (a P)
```

FormalConjectures/ErdosProblems/269.lean
Paul-Lez: That sounds reasonable to me!

mo271: splitting it up into two questions now

PR #449, opened 2025-07-22
Erdos 364

seewoo5:
Closes #373.

PR #513, opened 2025-07-29
Erdős Problems 364 and 366

mo271:


mo271:
I had overlooked #449, let's get that reviewed and merged, than I will rebase accordingly, generalizing `Powerful` to `$k$-full`

FormalConjectures/ErdosProblems/366.lean
```diff
@@ -0,0 +1,61 @@
...
+  sorry
+
+/--
+Note that $8$ is $3$-full and $9$ is 2-full.
+-/
+@[category test, AMS 11]
+example : (∃ (n : ℕ), (3).Full n ∧ (2).Full (n + 1)) := by
+  use 8
+  unfold Nat.Full
+  simp [show (8).primeFactors = {2} by
```
Paul-Lez: There should be a simproc for this! (I'll add one soonish!)

Paul-Lez (commented):
LGTM for 366 - I'll do another review once #449 is merged.

FormalConjectures/ErdosProblems/366.lean
mo271: Ack

FormalConjectures/ErdosProblems/366.lean
```diff
@@ -0,0 +1,61 @@
...
+    rw [show 8 = 2 ^ 3 by decide, Nat.primeFactors_pow]
+    · norm_num
+    · decide,
+    show (8 + 1).primeFactors = {3} by
+    rw [show (8 + 1) = 3 ^ 2 by decide, Nat.primeFactors_pow]
+    · norm_num
+    · decide]
```
Paul-Lez:
```suggestion
open Lean Meta Qq in
/-- Simproc to compute the set `Nat.primeFactors`. -/
dsimproc Nat.primeFactorsEq (Nat.primeFactors _) := fun e ↦ do
  unless e.isAppOfArity `Nat.primeFactors 1 do return .continue
  let some n ← fromExpr? e.appArg! | return .continue
  let outAsList : List Q(ℕ) := (unsafe n.primeFactors.val.unquot).map mkNatLit
  let outAsFinset : Q(Finset ℕ) := outAsList.foldl (fun s n ↦ q(insert $n $s)) q({})
  return .done outAsFinset


/--
Note that $8$ is $3$-full and $9$ is 2-full.
-/
@[category test, AMS 11]
example : (∃ (n : ℕ), (3).Full n ∧ (2).Full (n + 1)) := by
  use 8
  simp [Nat.Full, Nat.primeFactorsEq]
```
(maybe the simproc should go in `ForMathlib`! I'm hoping to upstream some version of it eventually, though the current implementation is certainly not the most efficient)

mo271: Nice, it is definitely good enough for our `ForMathlib`...

PR #514, opened 2025-07-30
Erdős Problem 352

### LINKED ISSUE #506: Erdős Problem 352
### What is the conjecture
    
https://www.erdosproblems.com/352

  Is there some $c>0$ such that every measurable $A\subseteq \mathbb{R}^2$ of measure $\geq c$ contains the vertices of a triangle of area 1?

Status: open

### Prerequisites needed
There is `Affine.Triangle`, and of course the Measure on $\mathbb{R}^2$, so all that is needed should be there.

rdivyanshu:
Fixes #506

### DIFF
--- a/FormalConjectures/ErdosProblems/352.lean
+++ b/FormalConjectures/ErdosProblems/352.lean
@@ -15,17 +15,17 @@ limitations under the License.
 -/
 
 import FormalConjectures.Util.ProblemImports
-open scoped EuclideanGeometry
 
 /-!
 # Erdős Problem 352
-
 *Reference:* [erdosproblems.com/352](https://www.erdosproblems.com/352)
 -/
 
+open scoped EuclideanGeometry
+
 /--
 Is there some c > 0 such that every measurable A ⊆ ℝ² of measure ≥ c
- contains the vertices of a triangle of area 1?
+ contains the vertices of a triangle of area 1 ?
 -/
 
 @[category research open, AMS 51]
--- a/FormalConjectures/ErdosProblems/352.lean
+++ b/FormalConjectures/ErdosProblems/352.lean
@@ -1,0 +1,39 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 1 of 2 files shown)


rdivyanshu:
Closed by mistake. Reopening.

FormalConjectures/ErdosProblems/352.lean
```diff
@@ -0,0 +1,39 @@
...
+*Reference:* [erdosproblems.com/352](https://www.erdosproblems.com/352)
+-/
+
+open scoped EuclideanGeometry
+
+/--
+Is there some c > 0 such that every measurable A ⊆ ℝ² of measure ≥ c
+ contains the vertices of a triangle of area 1 ?
+-/
+
```
callesonne:
```suggestion
-/
```

FormalConjectures/ErdosProblems/352.lean
```diff
@@ -0,0 +1,39 @@
...
+@[category research open, AMS 51]
+theorem erdos_352 :
+    (∃ c : ℝ, ∀ A : Set ℝ², Measurable A →
```
callesonne:
```suggestion
    (∃ c > (0 : ℝ), ∀ A : Set ℝ², MeasurableSet A →
```
Right now this says that `A` is measurable as a function from `ℝ²` to `Prop`, which is most likely equivalent (since I guess that `Prop` has the discrete measure) but more awkward to work with.

I also added `c > (0 : ℝ)`, which whilst not needed, is more faithful to the natural language statement.

callesonne (approved):
Thanks! Looks good to me modulo these small comments :)

FormalConjectures/ErdosProblems/352.lean
```diff
@@ -0,0 +1,39 @@
...
+       MeasureTheory.MeasureSpace.volume A ≥ c.toEReal
```
smmercuri: If you `open scoped ProbabilityTheory` you can use `ℙ A` here instead

FormalConjectures/ErdosProblems/352.lean
rdivyanshu: Thank you for review. 

Incorporated both suggestion.

FormalConjectures/ErdosProblems/352.lean
rdivyanshu: That is good suggestion. Thanks.

FormalConjectures/ErdosProblems/352.lean
```diff
@@ -0,0 +1,39 @@
...
+
+*Reference:* [erdosproblems.com/352](https://www.erdosproblems.com/352)
+-/
+
+open scoped EuclideanGeometry
+open scoped ProbabilityTheory
+
+/--
+Is there some c > 0 such that every measurable A ⊆ ℝ² of measure ≥ c
+ contains the vertices of a triangle of area 1 ?
```
mo271:
```suggestion
 contains the vertices of a triangle of area 1?
```

PR #521, opened 2025-07-31
Erdős problem 137

### LINKED ISSUE #372: Erdős Problem 137
### What is the conjecture
    
https://www.erdosproblems.com/137

  Let $k\geq 3$. Can the product of any $k$ consecutive integers $N$ ever be powerful? That is, must there always exist a prime $p\mid N$ such that $p^2\nmid N$?

Status: open

### Prerequisites needed
None (after manually defining "powerful")

See also #373

mo271:
fixes #372

### DIFF
--- a/FormalConjectures/ErdosProblems/137.lean
+++ b/FormalConjectures/ErdosProblems/137.lean
@@ -1,0 +1,57 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 137
+
+*References:*
+- [erdosproblems.com/137](https://www.erdosproblems.com/137)
+- [Er82c] Erdős, Paul, "Miscellaneous problems in number theory". Congr. Numer. (1982), 25-45.,
+- [ES75] P. Erdös, J. L. Selfridge, "The product of consecutive integers is never a power",
+  Illinois J. Math. 19(2): 292-301, 1975
+-/
+
+/--
+Let $k\geq 3$. Can the product of any $k$ consecutive integers $N$ ever be powerful? That is,
+must there always exist a prime $p\mid N$ such that $p^2\nmid N$?
+-/
+@[category research open, AMS 11]
+  theorem erdos_137 : (∀ k ≥ 3, ∀ n, ¬ ((Finset.Ioc n (n + k)).prod id).Powerful) ↔ answer(sorry) := by
+  sorry
+
+/--
+Let $k\geq 2$. Erdős and Selfridge [ES75] proved that the product of any $k$ consecutive
+integers $N$ cannot be a perfect power.
+-/
+@[category research solved, AMS 11]
+theorem erdos_137.variants.perfect_power (k : ℕ) (hk : k ≥ 2) (n : ℕ) (x l : ℕ)
+(hl : 2 ≤ l):
+    ((Finset.Ioc n (n + k)).prod id) ≠ x ^ l := by
+  sorry
+
+/--
+Erdős [Er82c] conjectures that, if $m$, $k$ are fixed and $n$ sufficiently large, then there must
+be at least $k$ distict primes $p$ such that $p\mid m(m+1)\cdots (m+n)$ and yet $p^2$ does not
+divide the right hand side.
+-/
+@[category research open, AMS 11]
+theorem erdos_137.multiple_powerful_factors (m k : ℕ) (hm : 0 < m): ∃ (n₀ : ℕ), ∀ n > n₀,
+    letI N := (Finset.Icc m (m + n)).prod id
+    ∃ P : Finset ℕ, P.card = k ∧ ∀ p ∈ P, p.Prime ∧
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/137.lean
```diff
@@ -0,0 +1,57 @@
...
+  theorem erdos_137 : (∀ k ≥ 3, ∀ n, ¬ ((Finset.Ioc n (n + k)).prod id).Powerful) ↔ answer(sorry) := by
+  sorry
+
+/--
+Let $k\geq 2$. Erdős and Selfridge [ES75] proved that the product of any $k$ consecutive
+integers $N$ cannot be a perfect power.
+-/
+@[category research solved, AMS 11]
+theorem erdos_137.variants.perfect_power (k : ℕ) (hk : k ≥ 2) (n : ℕ) (x l : ℕ)
+(hl : 2 ≤ l):
```
callesonne:
```suggestion
theorem erdos_137.variants.perfect_power (k : ℕ) (hk : k ≥ 2) (n : ℕ) (x l : ℕ) (hl : 2 ≤ l) :
```
I think this line-break might be by mistake?

FormalConjectures/ErdosProblems/137.lean
```diff
@@ -0,0 +1,57 @@
...
+  Illinois J. Math. 19(2): 292-301, 1975
+-/
+
+/--
+Let $k\geq 3$. Can the product of any $k$ consecutive integers $N$ ever be powerful? That is,
+must there always exist a prime $p\mid N$ such that $p^2\nmid N$?
+-/
+@[category research open, AMS 11]
+  theorem erdos_137 : (∀ k ≥ 3, ∀ n, ¬ ((Finset.Ioc n (n + k)).prod id).Powerful) ↔ answer(sorry) := by
+  sorry
```
callesonne:
```suggestion
theorem erdos_137 :
    (∀ k ≥ 3, ∀ n, ¬ (∏ x ∈ Finset.Ioc n (n + k), x).Powerful) ↔ answer(sorry) := by
  sorry
```
theorem here is indented by mistake, and I prefer the product notation for readability (as this affects how it's printed in the goal)

FormalConjectures/ErdosProblems/137.lean
```diff
@@ -0,0 +1,57 @@
...
+
+/--
+Let $k\geq 2$. Erdős and Selfridge [ES75] proved that the product of any $k$ consecutive
+integers $N$ cannot be a perfect power.
+-/
+@[category research solved, AMS 11]
+theorem erdos_137.variants.perfect_power (k : ℕ) (hk : k ≥ 2) (n : ℕ) (x l : ℕ)
+(hl : 2 ≤ l):
+    ((Finset.Ioc n (n + k)).prod id) ≠ x ^ l := by
```
callesonne:
```suggestion
    (∏ x ∈ Finset.Ioc n (n + k), x) ≠ x ^ l := by
```
As above

FormalConjectures/ErdosProblems/137.lean
```diff
@@ -0,0 +1,57 @@
...
+  sorry
+
+/--
+Erdős [Er82c] conjectures that, if $m$, $k$ are fixed and $n$ sufficiently large, then there must
+be at least $k$ distict primes $p$ such that $p\mid m(m+1)\cdots (m+n)$ and yet $p^2$ does not
+divide the right hand side.
+-/
+@[category research open, AMS 11]
+theorem erdos_137.multiple_powerful_factors (m k : ℕ) (hm : 0 < m): ∃ (n₀ : ℕ), ∀ n > n₀,
+    letI N := (Finset.Icc m (m + n)).prod id
```
callesonne:
```suggestion
    letI N := ∏ x ∈ Finset.Ioc m (m + n), x
```
as above

FormalConjectures/ErdosProblems/137.lean
```diff
@@ -0,0 +1,57 @@
...
+theorem erdos_137.variants.perfect_power (k : ℕ) (hk : k ≥ 2) (n : ℕ) (x l : ℕ)
+(hl : 2 ≤ l):
+    ((Finset.Ioc n (n + k)).prod id) ≠ x ^ l := by
+  sorry
+
+/--
+Erdős [Er82c] conjectures that, if $m$, $k$ are fixed and $n$ sufficiently large, then there must
+be at least $k$ distict primes $p$ such that $p\mid m(m+1)\cdots (m+n)$ and yet $p^2$ does not
+divide the right hand side.
+-/
```
callesonne:
```suggestion
divide the right hand side.

[Er82c] Erdős, P., Miscellaneous problems in number theory. Congr. Numer. (1982), 25-45.
-/
```

FormalConjectures/ErdosProblems/137.lean
```diff
@@ -0,0 +1,57 @@
...
+/--
+Let $k\geq 3$. Can the product of any $k$ consecutive integers $N$ ever be powerful? That is,
+must there always exist a prime $p\mid N$ such that $p^2\nmid N$?
+-/
+@[category research open, AMS 11]
+  theorem erdos_137 : (∀ k ≥ 3, ∀ n, ¬ ((Finset.Ioc n (n + k)).prod id).Powerful) ↔ answer(sorry) := by
+  sorry
+
+/--
+Let $k\geq 2$. Erdős and Selfridge [ES75] proved that the product of any $k$ consecutive
```
callesonne, line 36: I can't find this reference [ES75] on the corresponding webpage to this erdos problem. Could you add the reference here?

mo271: I had put the reference [ES75], not from erdosproblems, but added by myself, at the top of the file under "references", but from your remark, I see it is better to put it closer. We don't yet have a very disciplined way of dealing with references, especially when they appear in more than one conjecture. We probably should at some point, I will open an issue.

callesonne: Ah I missed that! Okay I agree with this, and probably adding the references like you did is better long-term but maybe a bit confusing since we don't follow it everywhere yet.

PR #524, opened 2025-07-31
feat: Improve linter warnings 

Paul-Lez:
This PR improves the warning messages from the category and AMS tag linters and prevents the header linter from firing on the file `FormalConjectures.All`.


callesonne (approved):
LGTM!

Potential suggestion would be that maybe since `@atts` needs to be extracted in `toCategory` and `toAMS` anyways, maybe its possible to extract `atts` first, which `toCategory` and `toAMS` takes as a parameter, instead of `(stx : TSyntax ``Command.declModifiers)`. But this is all very minor so feel free to merge as is.

Paul-Lez:
That's a good idea! Let me do that in a subsequent PR!

Issue #229, opened 2025-06-13
Generic Turing machine-based conjectures


tcosmo:
Hello all,

It is really cool that you integrated [bbchallenge's Antihydra](https://wiki.bbchallenge.org/wiki/Antihydra) (Turing machine [1RB1RA_0LC1LE_1LD1LC_1LA0LB_1LF1RE_---0RA](https://bbchallenge.org/1RB1RA_0LC1LE_1LD1LC_1LA0LB_1LF1RE_---0RA)) in your list:

https://github.com/google-deepmind/formal-conjectures/blob/4aba75a98f9721524210f788dc71f173115308f1/FormalConjectures/Other/BusyBeaverAntihydra.lean#L25-L36

Wouldn't it be interesting if we went more general and created a generic way to add the nonhalting problem from blank tape of any TM in the repo?

For instance, you could then inherit for free of the current 4,408 [BB(6) holdouts](https://wiki.bbchallenge.org/wiki/Holdouts_lists) and more generally, use bbchallenge's work as a quasi-infinite conjecture generator (e.g. there are ~100 millions of [7-state holdouts](https://wiki.bbchallenge.org/wiki/BB(7))).

Thankfully, [Thomas Vigouroux](https://github.com/vigoux) has already implemented the representation and parsing of TMs in Lean:

https://git.sr.ht/~vigoux/busybeaver/tree/master/item/Busybeaver/

Then, we could add a conjecture in the form of `nonhalt <machine>`, for instance using [bbchallenge's standard format](https://discuss.bbchallenge.org/t/standard-tm-text-format/60) for TMs, e.g. for Antihydra, `nonhalt "1RB1RA_0LC1LE_1LD1LC_1LA0LB_1LF1RE_---0RA"`. (only caveat is that in its current formulation, this format is limited to... 26 states)

For reference, any Pi_1 statement can be formulated that way (that includes for instance [Erdős's problem 406 (15 states)](https://arxiv.org/abs/2107.12475), [Goldbach's conjecture (25 states)](https://github.com/lengyijun/goldbach_tm), [Riemann Hypothesis (748 states)](https://github.com/sorear/metamath-turing-machines/blob/master/riemann-matiyasevich-aaronson.nql), Consistency of PA / ZF, etc...).

Would there be interest for this?

If yes, I am interested working on this.

Paul-Lez:
[...]

PR #526, opened 2025-08-01
feat: Implement Turing Machine Elaborator

Paul-Lez:
See discussion in #229. 

For now this is just a prototype - the code needs to be cleaned up a little and properly tested before we merge this!

### DIFF
--- a/FormalConjectures/ForMathlib/Computability/TuringMachineNotation.lean
+++ b/FormalConjectures/ForMathlib/Computability/TuringMachineNotation.lean
@@ -1,0 +1,101 @@
+import FormalConjectures.ForMathlib.Computability.TuringMachine
+
+open Turing BusyBeaver
+
+section
+
+open Lean Elab Meta Parser Command Term Qq
+
+
+def Char.toDir (c : Char) : Dir := if c.val == 82 then Dir.right else Dir.left
+
+--Outputs the alphabet index if `c` is a capital letter, e.g. `A -> 0`, `Z -> 25`
+def Char.toAlphabetNat (c : Char) : Nat :=
+  let AN := (c.val.toNat - 65 : Nat)
+  if AN < 15 then AN else panic! "Bad state encodding"
+
+def Alphabet := ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P',
+  'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z']
+
+/--
+Take as input a list of strings and return an array of `matchAltExpr` syntaxes
+mapping `(state, symbol)` pairs to actions of a Turing Machine,
+given by a term of type `Option (State n)`.
+-/
+def getStateDict (L : List String) (stateType : Name) : TermElabM (Array (TSyntax ``matchAltExpr)) := do
+  /-
+  The encoding of a TM move by a (sub)string `"ABC"` works as follows:
+    - The move the TM is currently at (the dictionary's key) is determined
+      by the position of the substring
+    - `A` gives the new symbol (`0` or `1`), `B` the direction the head moves (`L` or `R`) and `C` the new state (`0, ..., n-1` or `none`)
+  -/
+  let dirToStx : Dir → TermElabM Term
+    | .left => do return ← `(Dir.left)
+    | .right => do return ← `(Dir.right)
+
+  let natToStateStx (n : Nat) : TermElabM Term := do
+    let .some stateChar := Alphabet.get? n
+      | throwError "All state characters should be between A and Z"
+    return ← `($(Lean.mkIdent <| .str stateType stateChar.toString))
+
+  let toOptionState (s : String) : TermElabM Term := do
+    if s.length != 3 then return ← `(none) else
+      if '-' ∈ s.toList then return ← `(none) else
+        let state ← natToStateStx (s.get! ⟨2⟩).toAlphabetNat
+        let stmt ← `(Stmt.write $(quote (s.get! ⟨0⟩).toString.toNat!) $(← dirToStx (s.get! ⟨1⟩).toDir))
+        return ← `(some ($state, $stmt))
+  --The moves when the TM head read `0`
+  let moves0 ← L.toArray.zipIdx.mapM fun (s, i) => do
+    let s_first : Term ← toOptionState (s.extract ⟨0⟩ ⟨2⟩)
+    return ← `(matchAltExpr| | $(← natToStateStx i), (0 : Fin 2) => $s_first)
+  --The moves when the TM head read `1`
+  let moves1 ← L.toArray.zipIdx.mapM fun (s, i) => do
+    let s_first : Term ← toOptionState (s.extract ⟨3⟩ ⟨6⟩)
+    return ← `(matchAltExpr| | $(← natToStateStx i), (1 : Fin 2) => $s_first)
+  return moves0 ++ moves1
+
+def mkStateType (n : ℕ) : TermElabM Unit := do
+  let stateName := .mkSimple s!"State{n}"
+  if ← hasConst stateName then do return
+  let indName : Ident := Lean.mkIdent <| stateName
+  -- First construct the inductive type
+  let typeConstructors : List (TSyntax ``ctor) ←
+    (List.range n).mapM fun i ↦ `(Command.ctor| | $(Lean.mkIdent <|
+      .mkSimple s!"{Alphabet[i]!}"):ident : $(indName))
+  let typeConstructors := typeConstructors.toArray
+  let stx ← `(command| inductive $(indName) $typeConstructors:ctor*)
+  Lean.liftCommandElabM <| elabCommand stx
+  -- Create the inhabited instance as follows in order to have access to it e.g. in proofs
+  let inhabitedStx ←
+     `(command | instance : Inhabited $(Lean.mkIdent (.mkSimple s!"State{n}"))
+      := ⟨$(Lean.mkIdent <| (.mkStr2 s!"State{n}" "A"))⟩)
+  Lean.liftCommandElabM <| elabCommand inhabitedStx
+
+def parseTuring (descr : String) : TermElabM Expr := do
+  let moveListStr := descr.splitOn "_"
+  let nStates := moveListStr.length
+  mkStateType nStates
+  match ← checkTypeQ (.const (.mkSimple s!"State{nStates}") []) q(Type) with
+  | some stateType =>
+    let funConstructors ← getStateDict moveListStr (.mkSimple s!"State{nStates}")
+    let tm ← `(term | fun a b => match a, b with $funConstructors:matchAlt*)
+    -- let defaultEltName := Name.mkStr2 s!"State{nStates}" "A"
+    -- let defaultElt : Q($stateType) := .const defaultEltName []
+    -- let _ : Q(Inhabited $stateType) := q(⟨$defaultElt⟩)
+    let _ ← synthInstanceQ q(Inhabited $stateType)
+    let type := q(Machine (Fin 2) $stateType)
+    try
+      let expr ← elabTermEnsuringTypeQ tm type
+      return (expr : Expr)
+    catch _ =>
+      IO.println "ohno"
+    return default
+  | none => throwError "Invalid state type"
+
+end
+
+elab "turing_machine%" str:Lean.Parser.strLit : term => do
+  Lean.Elab.Term.withDeclName `FindABetterName do
+    parseTuring str.getString
+
+#check (turing_machine% "1RB1LC_1RC1RB_1RD0LE_1LA1LD_1RZ0LA")


FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,193 @@
...
+
+The machine's transitions are encoded as a single string, with each state's transitions
+separated by an underscore (`_`).
+For each state, a 3m-character substring defines the behavior:
+- The first 3 characters `"ABC"` describe the action when the head reads `0`:
+  - `A`: The symbol to write (`0` or `1`).
+  - `B`: The direction to move the head (`L` or `R`).
+  - `C`: The new state (`A` through `Z`).
+- The next 3 characters `"DEF"` describe the action when the head reads `1` using the same format.
+- So on...
```
tcosmo, line 36: This section seems outdated (multisymbol TMs)

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,193 @@
...
+
+The character `Z` is reserved for the halting state. The string `"---"` can be used to represent
```
tcosmo, line 38: There is a slight subtlety here:

- We generally use "1RZ" (or actually any `1R<X>` with `<X>` bigger than the number of states) to signify that *we know* that the machine halts
- `---` is used to either mean: "we know it is never reached" OR "we don't know if it is ever reached

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,193 @@
...
+`0, ..., 9` and throws an error in other cases. This is used when parsing the "symbol"
+component of turing machine string representations. -/
+private def Char.toNumeralSyntax (c : Char) : TermElabM Term := do
+  unless 48 ≤ c.val && c.val ≤ 57 do
+    throwError m!"Invalid write instruction: {c} is not a numeral."
+  let n := c.val - 48
+  `($(Lean.Quote.quote n.toNat))
+
+/-- `Char.toStateSyntax c stateName` outputs the syntax of the constructor corresponding to `c`
+if `c` is a capital letter (i.e. something betwen `A` and `Z`.). For example, `A` would output
```
tcosmo: betwen -> between

tcosmo (commented):
Minor comments only :)

If you want to add tests for the simulation of Turing machines here are some ideas:

- Check that [1RB1LC_1RC1RB_1RD0LE_1LA1LD_1RZ0LA](https://wiki.bbchallenge.org/wiki/1RB1LC_1RC1RB_1RD0LE_1LA1LD_1RZ0LA) halts after   47,176,870 steps
- Check that [1RB2LA1RA1RA_1LB1LA3RB1RZ](https://wiki.bbchallenge.org/wiki/BB(2,4)) halts after 3,932,964 steps

In case you want to add tests about nonhalting machines then you probably need to have at least one decider implemented (i.e. algorithm that proves that a machine does not halt) and if you do, the simplest target are to syntactically recognise machines with no halting transition (i.e. no undefined transition and no mention of a state bigger than state count), like for instance [1RB0LA_1LB0RB](https://bbchallenge.org/1RB0LA_1LB0RB). The nonhalting proof is trivial: there is no instruction that makes it halt.

Slightly more ambitious but certainly off-topic for your project are loops (machines that eventually repeat the same configuration, potentially translated in space) and here are some references for implementation:

- Cyclers only (repeating exactly the same configuration, no translation, e.g. [1RB0RB_1LC1RD_0LA1RA_---1RE_1LD0LD](https://bbchallenge.org/1RB0RB_1LC1RD_0LA1RA_---1RE_1LD0LD)): [Section 2](https://www.arxiv.org/pdf/2504.20563)
- Translated Cyclers legacy algorithm (repeating the same configuration, translated in space, e.g. [1RA---_------](https://bbchallenge.org/1RA---_------); [1RB0LE_1LC0RD_---1LD_1RE0LA_1LA0RE](https://bbchallenge.org/1RB0LE_1LC0RD_---1LD_1RE0LA_1LA0RE)): [Section 3](https://www.arxiv.org/pdf/2504.20563)
- Coq-BB5 algorithms for loops, which detects both of the above with a new, elegant, algorithm: [Section 4.3](https://github.com/bbchallenge/bbchallenge-paper/tree/build-paper-pdf)

Note that these deciders have already been implemented in Lean: https://git.sr.ht/~vigoux/busybeaver/tree/master/item/Busybeaver/Deciders

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
Paul-Lez: Fixed!

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
Paul-Lez: Now updated!

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
Paul-Lez: That's good to know! 

One question I have about this: my mental model of how an `n`-state busy beaver TM works is that we have our `n` states `A, B, ...` *plus* one extra halting state `Z`, and we can halt in two different ways: 
- By transitioning to the halting state `Z`/ some-sufficiently-large-other-letter
- By transitioning to a state (not necessarily `Z`) that has no transitions to other states. This case subsumes the previous one.

Is this actually correct? (If not, we may also want to tweak the contents of `ForMathlib.Computability.TuringMachine`. Incidentally, the definition of `HaltsAt` isn't quite right; I have a branch with some fixes that I'll PR shortly).

tcosmo: Yeah, I think we are saying the same thing.

The way we've done it at bbchallenge (for practical reasons, see [TNF Enumeration](https://wiki.bbchallenge.org/wiki/Tree_Normal_Form)) is as follows, transitions are either:

- Undefined ('---'), and the machine halts if it ever reaches one
- Defined, always of format `<b><D><s>` with `b` a symbol, `D` is L/R and `s` a state. If `s` corresponds to a state that is bigger than the number of defined states (typically, `Z`), then the machine stops

tcosmo:
Note: I'm going offline tomorrow until about Aug 25th, I apologise in advance for this delay for future reviews

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
Paul-Lez: Ok perfect!

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,194 @@
...
+
+private def Alphabet := ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N',
+  'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z']
+
+/-- `Char.toDirSyntax c` outputs the syntax corresponding to the character `c` if `c` is
+`R` or `L` and throws an error in other cases. This is used when parsing the "direction"
+component of turing machine string representations. -/
+private def Char.toDirSyntax : Char → TermElabM Term
+  | ⟨82, _⟩ => return ← `(Dir.right)
+  | ⟨76, _⟩ => return ← `(Dir.left)
```
eric-wieser: Does
```suggestion
  | 'R' => `(Dir.right)
  | 'L' => `(Dir.left)
```
work here?

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,194 @@
...
+  | char => throwError "Invalid direction {char}."
+
+/-- `Char.toBinarySyntax c` outputs the syntax corresponding to the character `c` if `c` is in
+`0, ..., 9` and throws an error in other cases. This is used when parsing the "symbol"
+component of turing machine string representations. -/
+private def Char.toNumeralSyntax (c : Char) : TermElabM Term := do
+  unless 48 ≤ c.val && c.val ≤ 57 do
```
eric-wieser:
```suggestion
  unless c.isDigit do
```

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,194 @@
...
+    throwError m!"Invalid write instruction: {c} is not a numeral."
+  let n := c.val - 48
```
eric-wieser:
```suggestion
  let n := c.val - '0'.val
```

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,194 @@
...
+  `($(Lean.Quote.quote n.toNat))
+
+/-- `Char.toStateSyntax c stateName` outputs the syntax of the constructor corresponding to `c`
+if `c` is a capital letter (i.e. something between `A` and `Z`.). For example, `A` would output
+the syntax `stateName.A` where `stateName` is the name of the type used to index sets.
+This is used when parsing the "state" component of turing machine string representations. -/
+private def Char.toStateSyntax (c : Char) (stateName : Name) : TermElabM Term := do
+  if c.val < 65 || c.val > 90 then
```
eric-wieser:
```suggestion
  if c.isUpper then
```

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,194 @@
...
+    throwError m!"Invalid state character: {c} should be between A and Z"
+  -- The convention is to use the character `Z` to denote the extra halting state.
+  if c == 'Z' then
+    `(none)
+  else
+    `($(Lean.mkIdent <| .str stateName c.toString))
```
eric-wieser:
```suggestion
    return Lean.mkIdent <| .str stateName c.toString
```

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,194 @@
...
+  unless s.length == 3 do
+    throwError m!"Invalid transition encoding: {s} should be 3 characters long."
+  if s == "---" then
+    `(none)
+  else
+    let state ← (s.get! ⟨2⟩).toStateSyntax stateName
+    let symbol ← (s.get! ⟨0⟩).toNumeralSyntax
+    let direction ← (s.get! ⟨1⟩).toDirSyntax
+    let stmt ← `(Stmt.write $symbol $direction)
+    `(some ($state, $stmt))
```
eric-wieser: I think better as 
```suggestion
private def String.toStmtSyntax (s : String) (stateName : Name) : TermElabM Term := do
  let [c_symbol, c_dir, c_state] := s.data | 
    throwError m!"Invalid transition encoding: {s} should be 3 characters long."
  if c_symbol = '-' ∧ c_dir = '-' ∧ c_state = '-' then
    `(none)
  else
    let state ← c_state.toStateSyntax stateName
    let symbol ← c_symbol.toNumeralSyntax
    let direction ← c_dir.toDirSyntax
    let stmt ← `(Stmt.write $symbol $direction)
    `(some ($state, $stmt))
```
`s.get! ⟨2⟩` is badly behaved, as `2` is a byte offset, but `.length` was measuring a character offset.

FormalConjectures/ForMathlib/Test/Computability/TuringMachine.lean
```diff
@@ -45,12 +45,41 @@ match l with
 instance : alwaysHaltingMachine.IsHalting := by
   rw [isHalting_iff_exists_haltsAt]
   -- halts after zero steps
-  exact ⟨0, by aesop⟩ 
+  exact ⟨0, by aesop⟩
+
+example : turing_machine% "------" = fun _ _ => none := by
```
eric-wieser, line 52: Please add a test where the parser fails

Paul-Lez: Added a few!

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,195 @@
...
+    throwError "All portions of the string separated by `_` should have the same length"
+  let numSymbols := entryLengths[0]! / 3
+  let stateName := numStates.toStateName
+  mkStateType numStates stateName
+  let .some stateType ← checkTypeQ (.const stateName []) q(Type)
+    | throwError m!"The constant {stateName} is not a type."
+  let funConstructors ← mkMachineMatchAltExpr moveListStr (numStates.toStateName) numSymbols numStates
+  let tm ← `(term | fun a b => match a, b with $funConstructors:matchAlt*)
+  let _ ← synthInstanceQ q(Inhabited $stateType)
+  let numSymbolsQ : Q(Nat) := ToExpr.toExpr numSymbols
```
eric-wieser:
```suggestion
  let numSymbolsQ : Q(Nat) := Lean.toExpr numSymbols
```

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,195 @@
...
+  let moveListStr := descr.splitOn "_"
+  let numStates := moveListStr.length
+  let entryLengths := moveListStr.map String.length |>.dedup
+  unless entryLengths.length == 1 do
+    throwError "All portions of the string separated by `_` should have the same length"
+  let numSymbols := entryLengths[0]! / 3
+  let stateName := numStates.toStateName
+  mkStateType numStates stateName
+  let .some stateType ← checkTypeQ (.const stateName []) q(Type)
+    | throwError m!"The constant {stateName} is not a type."
```
eric-wieser:
```suggestion
    | throwError m!"The constant {.ofConstName stateName} is not a type."
```

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,195 @@
...
+exists in the environment then this will be reused without checking that this is the right
+inductive type - such checks are left to the user.
+-/
+def parseTuring (descr : String) : TermElabM Expr := do
+  let moveListStr := descr.splitOn "_"
+  let numStates := moveListStr.length
+  let entryLengths := moveListStr.map String.length |>.dedup
+  unless entryLengths.length == 1 do
+    throwError "All portions of the string separated by `_` should have the same length"
+  let numSymbols := entryLengths[0]! / 3
```
eric-wieser, line 185: I assume you also want to check here that it is divisible by `3`?

eric-wieser: I take this back, I think the previous version actually makes more sense, since then you get an error on the last piece that isn't long enough.

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,197 @@
...
+  The encoding of a TM move by a (sub)string `"ABC"` works as follows:
+    - The move the TM is currently at (the dictionary's key) is determined
+      by the position of the substring
+    - `A` gives the new symbol (`0` or `1`), `B` the direction the head moves
+      (`L` or `R`) and `C` the new state (`0, ..., n-1` or `none`)
+  -/
+
+  let moves ← L.toArray.zipIdx.mapM fun (s, i) =>
+    Array.range numSymbols |>.mapM fun symbolIdx ↦ do
+      let s_first : Term ← (s.extract ⟨3 * symbolIdx⟩ ⟨3 * (symbolIdx + 1)⟩).toStmtSyntax stateName numStates
```
eric-wieser: Using `⟨` to construct a string position is pretty much always wrong, as it it ignores the difference between characters and utf-8-encoded bytes. You should use `String.next` to count instead.

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
Paul-Lez: Reverted back to the previous version

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
Paul-Lez: Done.

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
Paul-Lez: Ah actually I think it might be worth keeping this (otherwise one can run into unspecified behaviour of `String.next`)

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,203 @@
...
+- The first 3 characters `"ABC"` describe the action when the head reads `0`:
+  - `A`: The symbol to write (`0, 1, ..., m-1`).
+  - `B`: The direction to move the head (`L` or `R`).
+  - `C`: The new state (`A` through `Z`).
+- The next 3 characters `"DEF"` describe the action when the head reads `1` using the same format.
+- The next 3 characters `"GHI"` describe the action when the head reads `2` using the same format.
+- So on...
+
+The character `Z` is reserved for the halting state. The string `"---"` can be used to represent
+a transition to the halting state without writing or moving.
```
smmercuri, line 39: nit: a concrete example here might be nice to fully illustrate usage

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,203 @@
...
+  | 'L' => `(Dir.left)
+  | char => throwError "Invalid direction {char}."
+
+/-- `Char.toBinarySyntax c` outputs the syntax corresponding to the character `c` if `c` is in
+`0, ..., 9` and throws an error in other cases. This is used when parsing the "symbol"
+component of turing machine string representations. -/
+private def Char.toNumeralSyntax (c : Char) : TermElabM Term := do
+  unless c.isDigit do
+    throwError m!"Invalid write instruction: {c} is not a numeral."
+  let n := c.val - '0'.val
```
smmercuri, line 70: maybe a comment somewhere explaining this line

FormalConjectures/ForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -0,0 +1,203 @@
...
+/-- `Nat.toStateSyntax n stateName` outputs the syntax of the `n`-th constructor of the type used to
+index states. -/
+private def Nat.toStateSyntax (n : Nat) (stateName : Name) : TermElabM Term := do
+  if n > 25 then
+    throwError m!"Invalid state index {n}"
+  `($(Lean.mkIdent <| .str stateName Alphabet[n]!.toString))
+
+/-- `String.toStmtSyntax s stateName` parses a component of a Turing machine string representation
+and outputs the syntax of the corresponding (state, statement) pair
+(i.e. term of type `State × Stmt`). -/
```
smmercuri:
```suggestion
and outputs the syntax of the corresponding `(state, statement) : State × Stmt` pair. -/
```

smmercuri (approved):
LGTM!

PR #528, opened 2025-08-03
Erdős Problems 939

mo271:


FormalConjectures/ForMathlib/Data/Nat/Prime/Finset.lean
```diff
@@ -0,0 +1,26 @@
...
+
+import Mathlib
+
+/-- A Finset of numbers is coprime, or relatively prime, if its `gcd` is 1. -/
+@[reducible] def Finset.Coprime (S : Finset ℕ) : Prop := S.gcd id = 1
+
+theorem Finset.Coprime_pair_iff (S : Finset ℕ) (a b : ℕ) (h : S = {a, b}) :
+    S.Coprime ↔ Nat.Coprime a b := by
+  subst h
+  simp only [Finset.Coprime, Nat.Coprime, gcd_insert, gcd_singleton, normalize_eq]
```
Paul-Lez: minigolf
```suggestion
  simp only [h, Finset.Coprime, Nat.Coprime, gcd_insert, gcd_singleton, normalize_eq]
```

FormalConjectures/ErdosProblems/939.lean
```diff
@@ -0,0 +1,115 @@
...
+  The following proof works in princicple, but throws a "(kernel) deep recursion detected" error,
+  caused by the simproc `primeFactorsEq`.
+
+  constructor
+  · unfold Finset.Coprime
+    aesop
+  · simp only [Full, primeFactorsEq, Finset.mem_insert, insert_emptyc_eq, Finset.mem_singleton,
+    forall_eq_or_imp, reducePow, forall_eq]
+    norm_num
+  -/
```
Paul-Lez, line 85: For this, I did a quick investigation and I think what's happening here is that the kernel is reducing very large numbers. I'm not sure this can be avoided with the current implementation of the simproc, but then I plan on reimplementing it anyway when upstreaming it to Mathlib.

FormalConjectures/ErdosProblems/939.lean
```diff
@@ -0,0 +1,115 @@
...
+  sorry
+
+/--
+Are there infinitely many triples of coprime $3$-powerful numbers $a, b, c$ such that $a + b = c$?
+-/
+@[category research open, AMS 11]
+theorem erdos_939.variants.triples :
+    {(a,b,c) | ({a, b, c} : Finset ℕ).Coprime ∧
+    (3).Full a ∧ (3).Full b ∧ (3).Full c ∧
+    a + b = c}.Infinite ↔ answer(sorry) := by sorry
```
Paul-Lez:
```suggestion
      (3).Full a ∧ (3).Full b ∧ (3).Full c ∧
      a + b = c}.Infinite ↔ answer(sorry) := by
  sorry
```

FormalConjectures/ErdosProblems/939.lean
mo271: Ack

PR #532, opened 2025-08-04
golf proof for `lbMeasure_rigidMotion`

smmercuri:
golfed proof for `lbMeasure_rigidMotion` adapted from AlphaProof's solution


callesonne (approved):
Nice!

PR #542, opened 2025-08-07
Erdős Problem 945

mo271:


FormalConjectures/ErdosProblems/945.lean
```diff
@@ -0,0 +1,90 @@
...
+ - [erdosproblems.com/945](https://www.erdosproblems.com/945)
+ - [ErMi52] Erdős, P. and Mirsky, L., The distribution of values of the divisor function {$d(n)$}. Proc. London Math. Soc. (3) (1952), 257--271.
+-/
+
+open Filter Real
+
+
+private abbrev τ  := fun (n : ℕ) => n.divisors.card
+
+private noncomputable def F (x : ℝ) : ℕ :=
```
Paul-Lez, line 36: I think it would be nice to have a docstring here.

FormalConjectures/ErdosProblems/945.lean
```diff
@@ -0,0 +1,90 @@
...
+  sSup {k | ∃ (n : ℕ), n + k ≤ x ∧ (Set.Ioc n (n + k)).InjOn τ}
+
+/--
+Is it true that $F(x) \leq (\log x)^{O(1)}$?
+-/
+@[category research open, AMS 11]
+theorem erdos_945 : (∃ (O : ℝ → ℝ), O =O[Filter.atTop] (1 : ℝ → ℝ) ∧ ∀ (x : ℝ),
+    F x ≤ x.log ^ (O x)) ↔ answer(sorry) := by
```
Paul-Lez: Do we want to only require this to be true asymptotically? (I haven't thought about it too carefully, but this might prevent issues with junk values)

FormalConjectures/ErdosProblems/945.lean
```diff
@@ -0,0 +1,90 @@
...
+  sorry
+
+/--
+Is there a constant $C > 0$ such that, for all large $x$, every interval $[x, x+(\log x)C]$
+contains two integers with the same number of divisors?
+-/
+@[category research open, AMS 11]
+theorem erdos_945.variants.constant : (∃ (C : ℝ), C > 0 ∧ ∃ x₀, ∀ (x : ℝ), x₀ < x →
```
Paul-Lez: I'd be tempted to write this using filters.

FormalConjectures/ErdosProblems/945.lean
```diff
@@ -0,0 +1,90 @@
...
+    ↑b ∈ Set.Icc x (x + (x.log)^C) ∧
+    τ a = τ b) ↔ answer(sorry) := by
+  sorry
+
+-- TODO(firsching): show equivalence
+/--
+The two ways of phrasing the conjecture are equivalent.
+-/
+@[category undergraduate, AMS 11]
+theorem erdos_945.equivalence :
```
Paul-Lez: Here wouldn't it be better to use `type_of%` ?

FormalConjectures/Wikipedia/RamanujanTau.lean
```diff
@@ -32,9 +32,9 @@ There are two conjectures related to the Ramanujan τ-function:
 
 open PowerSeries PowerSeries.WithPiTopology
 
-noncomputable def Δ : PowerSeries ℤ := X * ∏' (n : ℕ+), (1 - X ^ (n : ℕ)) ^ 24
+private noncomputable def Δ : PowerSeries ℤ := X * ∏' (n : ℕ+), (1 - X ^ (n : ℕ)) ^ 24
```
Paul-Lez, line 35: I wonder if we shouldn't just lint for non-private definitions...

Paul-Lez (commented):
Nice!

FormalConjectures/Wikipedia/RamanujanTau.lean
mo271: Would be a good idea, I think we rarely re-use definitions from other files and prefer to include them in ForMathlib

FormalConjectures/ErdosProblems/945.lean
mo271: good idea, added a docstring

FormalConjectures/ErdosProblems/945.lean
mo271: Using `∀ᶠ` now, please check if it is correct like this.

FormalConjectures/ErdosProblems/945.lean
mo271: Agreed, this should be nicer..

FormalConjectures/ErdosProblems/945.lean
mo271: I first had that, but then I notived that when we have a `answer(sorry)` in the comments, it makes the equivalence contain two sorries that need to be filled in. Then a a human would need to check that they are the same, right?
I suppose this could be avoided by first defining two props and then using them in the two statements (adding the `answer(sorry)`s and then in the erquivalence theorem. 
Or is there another way to handle this? 
What are your thoughts here?

Paul-Lez: Ah good point! I'd be tempted to go via the "defining props" approach, in order to make sure everything stays consistent if we ever go back and change some of the statements in the file.

mo271: done

Paul-Lez (approved):

Issue #504, opened 2025-07-29
Erdős Problem 349


mo271:
### What is the conjecture
    
https://www.erdosproblems.com/349

  For what values of $t,\alpha \in (0,\infty)$ is the sequence $\lfloor t\alpha^n\rfloor$ complete (that is, all sufficiently large integers are the sum of distinct integers of the form $\lfloor t\alpha^n\rfloor$)?

Status: open

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

Paul-Lez:
#549

PR #549, opened 2025-08-08
feat: add formalisation of erdos_349

Paul-Lez:
#504


FormalConjectures/ErdosProblems/349.lean
```diff
@@ -0,0 +1,81 @@
...
+often and even infinitely often?
+-/
+@[category research open, AMS 11]
+theorem erdos_349.variants.floor_3_halves_odd :
+    {n | Odd ⌊(3/2 : ℝ) ^ n⌋}.Infinite ↔ answer(sorry) := by
+  sorry
+
+/--
+Is it true that the terms of the sequence $\lfloor (3/2)^n\rfloor$ are odd infinitely
+often and even infinitely often?
```
mo271:
```suggestion
Is it true that the terms of the sequence $\lfloor (3/2)^n\rfloor$ are even infinitely often?
```
To more closely align with the formal version.

PR #550, opened 2025-08-08
feat: Add formalisation of Mahler's 3/2 problem

Paul-Lez:


FormalConjectures/Wikipedia/Mahler32.lean
```diff
@@ -0,0 +1,55 @@
...
+@[category undergraduate, AMS 11]
+theorem mahler_conjecture.variants.consequence (H : type_of% mahler_conjecture) :
+    1 / 2 < Ω (3 / 2) := by
+  sorry
+
+/-- It is known that for all rational `p/q > 1` in lowest terms, we have `Ω(p/q) > 1/p`. -/
+@[category research solved, AMS 11]
+theorem mahler_conjecture.variants.flatto_lagarias_pollington (p q : ℕ) (hp : 1 ≤ p) (hq : 1 ≤ q)
+    (hpq : p.Coprime q) (hpq' : q ≤ p) : 1 / p < Ω (p / q) := by
+  sorry
```
mo271: I think we really need `p/q > 1` or `q <  p`, otherwise a counterexample would be:
```
example : ¬ type_of% mahler_conjecture.variants.flatto_lagarias_pollington := by
  push_neg
  use 1, 1
  simp [Ω]
```

FormalConjectures/Wikipedia/Mahler32.lean
```diff
@@ -0,0 +1,55 @@
...
+`1/2` for all positive integers `n`. -/
+private def IsZNumber (x : ℝ) : Prop :=
+  ∀ n > 0, Int.fract (x * (3 / 2 : ℝ) ^ n) < 1 / 2
+
+/-- The **Mahler Conjecture** states that there are no Z-numbers. -/
+@[category research open, AMS 11]
+theorem mahler_conjecture (x : ℝ) (hx : IsZNumber x) : False := by
+  sorry
+
+/-- If Mahler's conjecture is true, i.e. there are no `Z` numbers, then `Ω(3/2)` exceeds `1/2`.  -/
```
mo271:
```suggestion
/-- If Mahler's conjecture is true, i.e. there are no Z-numbers, then `Ω(3/2)` exceeds `1/2`.  -/
```
For consistency with the writing above

FormalConjectures/Wikipedia/Mahler32.lean
```diff
@@ -0,0 +1,55 @@
...
+@[category undergraduate, AMS 11]
+theorem mahler_conjecture.variants.consequence (H : type_of% mahler_conjecture) :
+    1 / 2 < Ω (3 / 2) := by
+  sorry
+
+/-- It is known that for all rational `p/q > 1` in lowest terms, we have `Ω(p/q) > 1/p`. -/
+@[category research solved, AMS 11]
+theorem mahler_conjecture.variants.flatto_lagarias_pollington (p q : ℕ) (hp : 1 ≤ p) (hq : 1 ≤ q)
+    (hpq : p.Coprime q) (hpq' : q ≤ p) : 1 / p < Ω (p / q) := by
```
mo271:
```suggestion
    (hpq : p.Coprime q) (hpq' : q < p) : 1 / p < Ω (p / q) := by
```

FormalConjectures/Wikipedia/Mahler32.lean
Paul-Lez: Oups, that was a typo! Now fixed.

PR #553, opened 2025-08-11
feat: state results and conjectures on VCₘ dim of convex sets in ℝⁿ

YaelDillies:
VC dimension of a set family is an old notion in learning theory. VC dimension of a *set* in a group (defined as the VC dimension of its family of translates) is a more recent notion motivated by additive combinatorics. VCₙ dimension of a set family is a very recent notion that appeared in the context of model theory. Therefore very few questions have been asked and answered at the intersection of both, i.e. about the VCₙ dimension of a family of translates

This PR offers some conjectures in this direction. The conjectures are all mine and do not appear in the literature. They are very easily stated due to the elementary nature of VCₙ dimension.

### DIFF
--- a/FormalConjectures/Other/VCDimConvex.lean
+++ b/FormalConjectures/Other/VCDimConvex.lean
@@ -14,8 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -/
 import FormalConjectures.Util.ProblemImports
-import Mathlib.Analysis.Convex.Basic
-import Mathlib.Data.Real.Basic
 
 /-!
 # VCₙ dimension of convex sets in ℝⁿ, ℝⁿ⁺¹, ℝⁿ⁺²
@@ -62,7 +60,7 @@ lemma vc2_lt_two_of_convex_r3 {C : Set (Fin 3 → ℝ)} (hC : Convex ℝ C)
     {x y : Fin 2 → Fin 3 → ℝ} {z : Set (Fin 2 × Fin 2) → Fin 3 → ℝ}
     (hxy : ∀ i j s, x i + y j + z s ∈ C ↔ (i, j) ∈ s) : False := sorry
 
-/-- Every convex set in ℝⁿ⁺¹ has VC₂ dimension at most 1. -/
+/-- Every convex set in ℝⁿ⁺¹ has VCₙ dimension at most 1. -/
 @[category research open, AMS 5 52]
 lemma exists_vcn_le_of_convex_rn_add_one (n : ℕ) :
     ∃ d : ℕ, ∀ (C : Set (Fin (n + 1) → ℝ)) (hC : Convex ℝ C) (x : Fin n → ℕ → Fin (n + 1) → ℝ)
--- a/FormalConjectures/ForMathlib/Combinatorics/Additive/VCDim.lean
+++ b/FormalConjectures/ForMathlib/Combinatorics/Additive/VCDim.lean
@@ -1,0 +1,34 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


FormalConjectures/Other/VCDimConvex.lean
```diff
@@ -0,0 +1,70 @@
...
+and conjectures that every convex set in ℝⁿ⁺¹ has finite VCₙ dimension.
+-/
+
+/-! ### What's known in the literature -/
+
+/-- Every convex set in ℝ² has VC dimension at most 3. -/
+@[category research solved, AMS 5 52]
+lemma vc_lt_four_of_convex_r2 {C : Set (Fin 2 → ℝ)} (hC : Convex ℝ C)
+    {x : Fin 4 → Fin 2 → ℝ} {y : Set (Fin 4) → Fin 2 → ℝ}
+    (hxy : ∀ i s, x i + y s ∈ C ↔ i ∈ s) : False := sorry
```
Paul-Lez, line 36: nit: the convention we follow is `by sorry` rather than `sorry`.

FormalConjectures/Other/VCDimConvex.lean
```diff
@@ -0,0 +1,70 @@
...
+import Mathlib.Data.Real.Basic
+
+/-!
+# VCₙ dimension of convex sets in ℝⁿ, ℝⁿ⁺¹, ℝⁿ⁺²
+
+In the literature it is known that every convex set in ℝ² has VC dimension at most 3,
+and there exists a convex set in ℝ³ with infinite VC dimension (even more strongly,
+which shatters an infinite set).
+
+This file states that every convex set in ℝⁿ has finite VCₙ dimension, constructs a convex set in
```
Paul-Lez, line 25: Which of the formal statements below corresponds to "every convex set in ℝⁿ has finite VCₙ dimension"?

FormalConjectures/Other/VCDimConvex.lean
Paul-Lez, line 36: How annoying would it be to add a local definition of the VC dimension and use it in these formalisations? This wouldn't have to be at Mathlib level generality (but could definitely be added to `ForMathlib` and eventually upstreamed;))

Paul-Lez (commented):
Thanks for the contribution:)

FormalConjectures/Other/VCDimConvex.lean
YaelDillies: It's definitely possible, but it would make the statement less concrete for no theoretical benefit (the first thing you do when dealing with VC dimension is unfolding the definition, quite often). What do you think?

FormalConjectures/Other/VCDimConvex.lean
YaelDillies: Whoops, that got lost in translation

FormalConjectures/Other/VCDimConvex.lean
Paul-Lez: I see! In my eyes, the main benefit of using the general definition is maintainability e.g. if we later decide we need to make some changes related to that definition, we would only need to modify one `def` rather than all the theorems (which might end up introducing error/inconsistencies). A good middle ground could involve:
1) Stating a general definition
2) Adding a tiny bit of API (which you can leave sorried out if it's not in the `ForMathlib` directory) to help unfolding the definition.

What do you think?

FormalConjectures/Other/VCDimConvex.lean
```diff
@@ -0,0 +1,70 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+import FormalConjectures.Util.ProblemImports
+import Mathlib.Analysis.Convex.Basic
+import Mathlib.Data.Real.Basic
```
smmercuri: this should suffice 
```suggestion
import FormalConjectures.Util.ProblemImports
```

YaelDillies:
Sorry, looks like rebasing this branch caused trouble because it was the main branch of my fork. I'll reopen a new PR.

PR #557, opened 2025-08-12
Erdős Problem 20

mo271:


FormalConjectures/ErdosProblems/20.lean
```diff
@@ -0,0 +1,58 @@
...
+noncomputable def f (n k : ℕ) : ℕ :=
+    sInf {m | ∀ {α : Type}, ∀ (F : Set (Set α)),
+    ((∀ f ∈ F, f.ncard = n) ∧ m ≤ F.ncard) → ∃ S ⊆ F, S.ncard = k ∧ IsSunflower S}
+
+
+/--
+Is it true that $f(n,k) < c^n_k$ for some constant $c_k>0?
+-/
+@[category research open, AMS 5]
+theorem erdos_20 : ∃ (c : ℕ → ℕ), ∀ n k, f n k < (c k)^n := by
```
mo271, line 54: not sure if this is nicer as the equivalent (when assuming `Clasical`):
```
theorem erdos_20' : ∀ k, ∃ (c : ℕ), ∀ n, f α n k < c^n := by 
```

Paul-Lez: I'm happy with either!

PR #558, opened 2025-08-12
Erdős Problem 30

### LINKED ISSUE #205: Erdős Problem 30: asymptotics of max size of Sidon set.
### What is the conjecture
    
https://www.erdosproblems.com/30

  Let $h(N)$ be the maximum size of a Sidon set in $\{1,\ldots,N\}$. Is it true that, for every $\epsilon>0$,
$$h(N) = N^{1/2}+O_\epsilon(N^\epsilon)?$$

Status: open

### Prerequisites needed
None, Sidon sets are in `ForMathlib`.

### Choose either option

mo271:
closes #205 

deleting `maxSidonSetSize_bound`, since this seems to be 
 - not used anywhere
 -  wrong

### DIFF
--- a/FormalConjectures/ErdosProblems/30.lean
+++ b/FormalConjectures/ErdosProblems/30.lean
@@ -31,7 +31,7 @@ noncomputable abbrev h := maxSidonSetSize
 open Filter
 
 /--
-Is it true that, for every $\varepsilon > 0$, $h(N) = \sqrt N + O_{\varespilon}(N_{\varespilon})
+Is it true that, for every $\varepsilon > 0$, $h(N) = \sqrt N + O_{\varespilon}(N^\varespilon)
 -/
 @[category research open, AMS 11]
 theorem erdos_30 :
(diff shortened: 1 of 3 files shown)


FormalConjectures/ErdosProblems/30.lean
```diff
@@ -0,0 +1,42 @@
...
+/--
+Let $h(N)$ be the maximum size of a Sidon set in $\{1, \dots, N\}$.
+-/
+noncomputable abbrev h := maxSidonSetSize
+
+
+open Filter
+
+/--
+Is it true that, for every $\varepsilon > 0$, $h(N) = \sqrt N + O_{\varespilon}(N_{\varespilon})
```
Paul-Lez: Should this be:
```suggestion
Is it true that, for every $\varepsilon > 0$, $h(N) = \sqrt N + O_{\varespilon}(N^\varespilon)
```

Paul-Lez (approved):

PR #559, opened 2025-08-12
Erdős Problems 590, 591, 592

mo271:


FormalConjectures/ForMathlib/SetTheory/Cardinal/SimpleGraph.lean
```diff
@@ -0,0 +1,44 @@
...
+
+/--
+This proposition asserts the Ramsey property `κ → (κ, c)₂`, where `κ` is the
+cardinality of the ordinal `ω ^ β`.
+
+It states that for any 2-coloring of the edges of a complete graph on `κ`
+vertices, there must be a monochromatic red clique of size `κ` or a
+monochromatic blue clique of size `c`.
+-/
+def OmegaPowerRamsey  (β : Ordinal.{u}) (c : Cardinal.{u}): Prop :=
```
Paul-Lez:
```suggestion
def OmegaPowerRamsey  (β : Ordinal.{u}) (c : Cardinal.{u}) : Prop :=
```

FormalConjectures/ForMathlib/SetTheory/Cardinal/SimpleGraph.lean
```diff
@@ -0,0 +1,44 @@
...
+  -- Let α be the ordinal ω^β
+  let κ := ω ^ β
+  -- Let V be a type with the cardinality of α.
+  let V := κ.toType
```
Paul-Lez: Here rather than choosing a type of cardinality `κ`, I think it might be better to quantify over all `V` that have that cardinality, since this will make `OmegaPowerRamsey` easier to apply.

FormalConjectures/ForMathlib/SetTheory/Cardinal/SimpleGraph.lean
```diff
@@ -0,0 +1,44 @@
...
+import Mathlib.Order.CompletePartialOrder
+import Mathlib.SetTheory.Ordinal.Exponential
+
+open Cardinal Ordinal
+
+universe u
+
+/--
+This proposition asserts the Ramsey property `κ → (κ, c)₂`, where `κ` is the
+cardinality of the ordinal `ω ^ β`.
```
Paul-Lez: (just adding this for other people like myself who aren't super familiar with this area;))
```suggestion
cardinality of the ordinal `ω ^ β` and `c` is some cardinal.
```

FormalConjectures/ErdosProblems/591.lean
```diff
@@ -0,0 +1,35 @@
...
+open Cardinal Ordinal
+
+universe u
+
+/--
+Let $α$ be the infinite ordinal $\omega^{\omega^2}$. Is it true that any red/blue colouring of the
+edges of $K_α$ there is either a red $K_α$ or a blue $K_3$.
+-/
+@[category research open, AMS 3]
+theorem erdos_591 : OmegaPowerRamsey (ω^2) 3 ↔ answer(sorry) := by
```
Paul-Lez:
```suggestion
theorem erdos_591 : OmegaPowerRamsey (ω ^ 2) 3 ↔ answer(sorry) := by
```

FormalConjectures/ForMathlib/SetTheory/Cardinal/SimpleGraph.lean
mo271: I added an argument $V$ for the type. Is this what you had in mind?

Paul-Lez: Ah what I had in mind was quantifying inside the definition, e.g. something along the lines of
```lean
def OmegaPowerRamsey (β : Ordinal.{u}) (c : Cardinal.{u}) : Prop :=
  let κ = (ω ^ β).card
  -- For any red/blue edge coloring of the complete graph on V...
  -- (represented by two graphs G_red and G_blue that are complements)
  ∀ (V : Type u),  κ = Cardinal.mk V → 
    ∀ (G_red G_blue : SimpleGraph V), IsCompl G_red G_blue →
      -- ...there is either a red K_α
      (∃ (s : Set V), G_red.IsClique s ∧ #s = κ) ∨
      -- ...or there is a blue K_3.
      (∃ (s : Set V), G_blue.IsClique s ∧ #s = c)
```
(maybe up to some permutation of the quantifiers and so on)

mo271: ah, that makes more sense, done!

PR #561, opened 2025-08-13
fix: add condition that type is non-empty

Paul-Lez:


FormalConjectures/Wikipedia/UnionClosed.lean
smmercuri, line 1: Can you change the year to `2025` in the copyright header?

Paul-Lez: I think this is intentional (iirc this formalisation was started in 2024)

smmercuri: Oh! then, LGTM

PR #565, opened 2025-08-13
fix: `error_le` is not true for `r` close to `0`

smmercuri:
As formalised this fails, e.g., for `r = 1/16`.


FormalConjectures/Wikipedia/GaussCircleProblem.lean
```diff
@@ -63,7 +63,7 @@ $$
 [Ha59]  Hardy, G. H. (1959). _Ramanujan: Twelve Lectures on Subjects Suggested by His Life and Work_(3rd ed.). New York: Chelsea Publishing Company. p. 67
 -/
```
Paul-Lez, line 65: The docstring should change too.

PR #566, opened 2025-08-13
fix: `maxSidonSetSize_bound` should use `2 * Real.sqrt N`

smmercuri:
The upper bound of `Nat.sqrt N + 1` is false (consider `N = 7` and the Sidon set of `{1, 2, 5, 7}` of size `4`). Replacing this with  the easy upper bound of `2 * Real.sqrt N` since it has `category undergraduate`.

### DIFF
--- a/FormalConjectures/ErdosProblems/30.lean
+++ b/FormalConjectures/ErdosProblems/30.lean
@@ -1,0 +1,42 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


mo271:
This is already "fixed" as drive-by in https://github.com/google-deepmind/formal-conjectures/pull/558. There I just removed it. See also
https://github.com/google-deepmind/formal-conjectures/pull/286#discussion_r2271612715 .
I think your fix is better, by actually replacing it with a bound that makes sense.

mo271:
I took the liberty of resolving the merge conflict that I caused...

smmercuri:
Ah right I didn't notice that! In the future we might want to encourage more references to help with maintenance

PR #571, opened 2025-08-14
fix: add missing condition to erdos_11

Paul-Lez:
The condition was also missing from the informal statements.


mo271:
Not sure if we want to also tell Thomas Bloom about this condition. (It is of course trivial)
Seems like erdosproblems.com now has a comment function, perhaps that could be used?

mo271 (approved):

PR #576, opened 2025-08-14
fix: `conjecture6_3` distribution function -> density function

smmercuri:
Conjecture 6.3 in https://arxiv.org/pdf/0911.2077 is stated using the distribution function rather than the density function


FormalConjectures/Arxiv/0911.2077/Conjecture6_3.lean
```diff
@@ -28,7 +28,7 @@ open NNReal ENNReal ProbabilityTheory
 
 
 /-- As usual, let $\Phi$ and $\phi$ be the distribution function and density of the standard normal. -/
```
Paul-Lez: Let's maybe change this docstring since we're only using one of the two functions that are mentioned.

PR #577, opened 2025-08-14
Erdős Problem 28

### LINKED ISSUE #203: Erdős Problem 28
### What is the conjecture
    
https://www.erdosproblems.com/28

  If $A\subseteq \mathbb{N}$ is such that $A+A$ contains all but finitely many integers then $\limsup 1_A\ast 1_A(n)=\infty$.

Status: open

see also #204 

### Prerequisites needed
Not sure if  we have already the convolution $\ast$ in a convenient way.

mo271:
fixes #203 
drive-by: fix wrong definition of convolution for Erdős 943

### DIFF
--- a/FormalConjectures/ErdosProblems/28.lean
+++ b/FormalConjectures/ErdosProblems/28.lean
@@ -22,7 +22,7 @@ import FormalConjectures.Util.ProblemImports
 *Reference:* [erdosproblems.com/28](https://www.erdosproblems.com/28)
 -/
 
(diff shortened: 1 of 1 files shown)


mo271:
I think using `PowerSeries` seems like the best approach for the convolution $f\ast g(n) = \sum_{a+b=n}f(a)g(b)$, at least I didn't find anything more convenient...

Paul-Lez (approved):

Issue #88, opened 2025-05-28
Brocard's Conjecture


mh-tanver:

BoltonBailey:
> For any integer n≥2, there exists at least one prime p such that n^2<p<(n+1)^2.

This is the Legendre conjecture, which is already [here](https://github.com/google-deepmind/formal-conjectures/blob/662fea92e726f4c770d5df77ab4246447c1f280e/FormalConjectures/LandauProblems/LegendreConjecture.lean#L29). Perhaps you meant [this](https://en.wikipedia.org/wiki/Brocard%27s_conjecture) or [this](https://en.wikipedia.org/wiki/Brocard%27s_problem)?

mh-tanver:
Thanks for pointing it out. Yes I meant the first one.

mh-tanver:
Why can't I pick the assignee?

mo271:
do you want to assign this to yourself, @mh-tanver?

mh-tanver:
Yes I want to

Paul-Lez:
@mh-tanver It seems Brocard's conjecture has already been done - sorry about that! Would you like to take up [Brocard's Problem](https://en.wikipedia.org/wiki/Brocard%27s_problem) instead?

mh-tanver:
Alright then.

BoltonBailey:
@mh-tanver have you made any progress on this? If it helps, I already have a preexisting formalization of Brocard's problem, I'd really like to see it added and I'd be happy to make the PR myself.

mh-tanver:
Hi @BoltonBailey , sure go ahead. Sorry I was busy lately.

PR #578, opened 2025-08-14
feat: add Brocard's Problem / Erdős 398

BoltonBailey:
Related to #88.

Note this adds Brocard's Problem, rather than Brocard's Conjecture, which already is present. This PR moves the latter file to a name that clarifies this difference.


mo271:
Cool, thanks!
This is also https://www.erdosproblems.com/398 
I propose to do the same here as for https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/36.lean :
Formalize the version in the `ErdosProblems` directory and then import it in the `Wikipedia` directory like so:
https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/Wikipedia/MinimalOverlapProblem.lean

So this would mean to 
`mv FormalConjectures/Wikipedia/BrocardProblem.lean FormalConjectures/ErdosProblems/398.lean`
change the docstring a little and then write `import FormalConjectures.ErdosProblems.«398»` and the copyright header to `FormalConjectures/Wikipedia/BrocardProblem.lean`

mo271 (changes requested):
LGTM other then making a new file for Erdős problems as described in the comment.

FormalConjectures/Wikipedia/BrocardProblem.lean
```diff
@@ -0,0 +1,33 @@
...
+-/
+
+open Nat
+
+/--
+**Brocard's Problem**
+Does $n! + 1 = m^2$ have integer solutions other than $n = 4, 5, 7$?
+-/
+@[category research open, AMS 11]
+theorem brocard_problem : (∃ n ≥ 8, ∃ m, n ! + 1 = m ^ 2) ↔ answer(sorry) := by
```
mo271:
```suggestion
theorem brocard_problem : {n | ∃ m, n ! + 1 = m ^ 2} = {4, 5, 7} ↔ answer(sorry) := by
```
This way the formal statement is closer to the docstring.

mo271:
Thanks, @BoltonBailey, keep the good stuff coming!

PR #583, opened 2025-08-18
Correct assumption of Invariant subspace problem

yoh-tanimoto:
For a positive solution, the invariant subspace problem requires that the dimension of the vector space `H` is at least 2, because otherwise any subspace would be either `H` or `{0}`. Do also some other minor corrections in docstrings.


Paul-Lez (approved):
Good catch!

PR #586, opened 2025-08-19
fix build warnings

mo271:
cc-ing @henrykmichalewski for the change in the definition of `IsEdgeCover`

### DIFF
--- a/FormalConjectures/Wikipedia/UnionClosed.lean
+++ b/FormalConjectures/Wikipedia/UnionClosed.lean
@@ -1,5 +1,5 @@
 /-
-Copyright 2024 The Formal Conjectures Authors.
+Copyright 2025 The Formal Conjectures Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


FormalConjectures/Wikipedia/UnionClosed.lean
```diff
@@ -1,5 +1,5 @@
 /-
-Copyright 2024 The Formal Conjectures Authors.
+Copyright 2025 The Formal Conjectures Authors.
```
Paul-Lez, line 2: For this, shouldn't we allow different years in the header? (I think this is a sign I should fix the copyright header linter!)

mo271: If we want this to be in sync with the javascript check we chould match  `(!/^2[0-9]{3}$/.test(year))`, i.e. the copyright should be a 2 followed by three digits.
I think the linter needs to be fixed some time before 2026-01-01...

However, for this particular file `2024` should just be `2025`, since the file is not that old.

Paul-Lez: I think the reason that this file has 2024 is that the formalisation was written by Bhavik a while back (in 2024) but only added to formal conjectures recently. But arguably this doesn't matter too much anyway!

Paul-Lez (approved):

PR #587, opened 2025-08-19
Erdős Problem 40

### LINKED ISSUE #204: Erdős Problem 40
### What is the conjecture
    
https://www.erdosproblems.com/40

  For what functions $g(N)\to \infty$ is it true that 
$$\lvert A\cap \{1,\ldots,N\}\rvert \gg N^{1/2}g(N)$$
implies $\limsup 1_A\ast 1_A(n)=\infty$?

Status: open

see also #203

mo271:
fixes #204

### DIFF
--- a/FormalConjectures/ErdosProblems/40.lean
+++ b/FormalConjectures/ErdosProblems/40.lean
@@ -34,10 +34,10 @@ implies $\limsup 1_A\ast 1_A(n)=\infty$.
 -/
 def Erdos40For (g : ℕ → ℝ) : Prop :=
   ∀ (A : Set ℕ),
-  ((fun (N : ℕ) => (N : ℝ).sqrt/(g N)) =O[atTop] (fun (N : ℕ) => ((A ∩ Set.Icc 1 N).ncard : ℝ))) →
-  (limsup (fun (N : ℕ) =>
-  letI a := PowerSeries.mk (indicator A 1)
-  (a * a).coeff ℕ N) atTop = (⊤ : ℕ∞))
+    ((fun (N : ℕ) => (N : ℝ).sqrt/(g N)) =O[atTop] (fun (N : ℕ) => ((A ∩ Set.Icc 1 N).ncard : ℝ))) →
+    (limsup (fun (N : ℕ) =>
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/40.lean
```diff
@@ -0,0 +1,94 @@
...
+The predicate for a function $g\colon\mathbb{N} → \mathbb{R})$ that
+$$\lvert A\cap \{1,\ldots,N\}\rvert \gg \frac{N^{1/2}}{g(N)}$$
+implies $\limsup 1_A\ast 1_A(n)=\infty$.
+-/
+def Erdos40For (g : ℕ → ℝ) : Prop :=
+  ∀ (A : Set ℕ),
+  ((fun (N : ℕ) => (N : ℝ).sqrt/(g N)) =O[atTop] (fun (N : ℕ) => ((A ∩ Set.Icc 1 N).ncard : ℝ))) →
+  (limsup (fun (N : ℕ) =>
+  letI a := PowerSeries.mk (indicator A 1)
+  (a * a).coeff ℕ N) atTop = (⊤ : ℕ∞))
```
Paul-Lez:
```suggestion
def Erdos40For (g : ℕ → ℝ) : Prop :=
  ∀ (A : Set ℕ),
    ((fun (N : ℕ) => (N : ℝ).sqrt/(g N)) =O[atTop] (fun (N : ℕ) => ((A ∩ Set.Icc 1 N).ncard : ℝ))) →
    (limsup (fun (N : ℕ) =>
    letI a := PowerSeries.mk (indicator A 1)
    (a * a).coeff ℕ N) atTop = (⊤ : ℕ∞))
```

Paul-Lez: We seem to be using this power series trick quite a bit - maybe it would be worth adding something in ForMathlib about that at some point?

Paul-Lez (approved):

FormalConjectures/ErdosProblems/40.lean
mo271: Yes, good plan, will do next time I use it

PR #588, opened 2025-08-19
bump toolchain and Mathlib to v4.20.1

mo271:


mo271:
Here doc-gen4 doesn't work properly, and the "Extract documentation" step doesn't write output which leads the step "Inject stats into index.html" to fail. We do not attempt to fix it, since it works again in the next version and also `v4.22.0`: https://github.com/google-deepmind/formal-conjectures/actions/runs/18711723422/job/53361540372
https://github.com/google-deepmind/formal-conjectures/actions/runs/18712095972/job/53362725181

FormalConjectures/ErdosProblems/366.lean
```diff
@@ -50,7 +50,7 @@ theorem erdos_366.variant.three_two :
 /--
 Are there any consecutive pairs of $3$-full integers?
 -/
-@[category undergraduate, AMS 11]
+@[category research open, AMS 11]
```
mo271:
```suggestion
@[category undergraduate, AMS 11]
```

FormalConjectures/Wikipedia/Conway99Graph.lean
```diff
@@ -25,7 +25,12 @@ import FormalConjectures.Util.ProblemImports
 namespace Conway99Graph
 
 --TODO(firsching): Consider using SimpleGraph.IsSRGWith to formulate the conjecture.
+
+open SimpleGraph
+
 variable {V : Type} {G : SimpleGraph V}
+
+
```
eric-wieser, line 33: nit: just one newline?

FormalConjectures/Wikipedia/Conway99Graph.lean
```diff
@@ -25,7 +25,12 @@ import FormalConjectures.Util.ProblemImports
 namespace Conway99Graph
 
 --TODO(firsching): Consider using SimpleGraph.IsSRGWith to formulate the conjecture.
+
+open SimpleGraph
```
eric-wieser, line 29: perhaps `open scoped` if you only need this for notation?

mo271: will fix that later on head in order to avoid running the ci again, if that's ok.

FormalConjectures/Wikipedia/Conway99Graph.lean
mo271: same, will do later

FormalConjectures/Wikipedia/Conway99Graph.lean
mo271: this was actually needed in a lot of places, and now that it is open we can remove a buch more `SimpleGraph`. done in #1141

FormalConjectures/Wikipedia/Conway99Graph.lean
mo271: done in #1141

PR #589, opened 2025-08-19
Erdős Problem 213

### LINKED ISSUE #425: Erdős Problem 213
### What is the conjecture
    
https://www.erdosproblems.com/213

  Let $n\geq 4$. Are there $n$ points in $\mathbb{R}^2$, no three on a line and no four on a circle, such that all pairwise distances are integers?

Status: open

### Prerequisites needed
see also #369 and #328 with similar hypothesis.

rdivyanshu:
Fixes #425


FormalConjectures/ErdosProblems/213.lean
```diff
@@ -0,0 +1,38 @@
...
+-/
+
+open scoped EuclideanGeometry
+
+/--
+Let n ≥ 4. Are there n points in ℝ², no three on a line and no four on a circle,
+such that all pairwise distances are integers?
+-/
+@[category research open, AMS 52]
+theorem erdos_213 : ∀ n : ℕ, n ≥ 4 →
```
rdivyanshu: I feel this is very strong reading of conjecture. Maybe just increasing upper bound or may be for all n there exists a set of size greater than n satisfying those criteria is better ?

Paul-Lez: Wouldn't those versions be equivalent to the original formulation (e.g. if there is such a set `S` of size greater than `n`, then we can just take any subset of size `n`)?

FormalConjectures/ErdosProblems/213.lean
```diff
@@ -0,0 +1,38 @@
...
+    (∃ S : Set ℝ², S.Finite ∧ S.ncard = n ∧
+                   (∀ T : Set ℝ², T ⊆ S ∧ T.ncard = 3 → ¬ Collinear ℝ T) ∧
+                   (∀ Q : Set ℝ², Q ⊆ S ∧ Q.ncard = 4 → ¬ EuclideanGeometry.Cospherical Q) ∧
+                   (S.Pairwise fun p₁ p₂ => dist p₁ p₂ ∈ Set.range Int.cast))
```
Paul-Lez:
```suggestion
      (∀ T : Set ℝ², T ⊆ S ∧ T.ncard = 3 → ¬ Collinear ℝ T) ∧
      (∀ Q : Set ℝ², Q ⊆ S ∧ Q.ncard = 4 → ¬ EuclideanGeometry.Cospherical Q) ∧
      (S.Pairwise fun p₁ p₂ => dist p₁ p₂ ∈ Set.range Int.cast))
```

FormalConjectures/ErdosProblems/213.lean
rdivyanshu: Yes, correct.

Paul-Lez: Here I would be tempted to do the following: define a predicate `IsGood` (perhaps with a better name:)) stating that for a given `n`, such a set exists. You would then be able to state the main result as `(∀ n : ℕ, n ≥ 4 →IsGood n) ↔ answer(sorry)`. 

One reason this would be useful is that it would allow you to state some of the proved special cases, which is often a good sanity check that the main conjecture statement is correct.

rdivyanshu: @Paul-Lez Sorry for delay. I have factored out predicate and added proved special case. PTAL.

FormalConjectures/ErdosProblems/213.lean
```diff
@@ -0,0 +1,49 @@
...
+namespace Erdos213
+
+def Erdos213For (n : ℕ) : Prop := ∃ S : Set ℝ², S.Finite ∧ S.ncard = n ∧
+    (∀ T : Set ℝ², T ⊆ S ∧ T.ncard = 3 → ¬ Collinear ℝ T) ∧
+    (∀ Q : Set ℝ², Q ⊆ S ∧ Q.ncard = 4 → ¬ EuclideanGeometry.Cospherical Q) ∧
+    (S.Pairwise fun p₁ p₂ => dist p₁ p₂ ∈ Set.range Int.cast)
+
+/--
+Let n ≥ 4. Are there n points in ℝ², no three on a line and no four on a circle,
+such that all pairwise distances are integers?
```
Paul-Lez:
```suggestion
Let $n ≥ 4$. Are there $n$ points in $\mathbb{R}^2$, no three on a line and no four on a circle,
such that all pairwise distances are integers?
```

FormalConjectures/ErdosProblems/213.lean
```diff
@@ -0,0 +1,49 @@
...
+-/
+@[category research open, AMS 52]
+theorem erdos_213 : (∀ n : ℕ, n ≥ 4 → Erdos213For n)
+    ↔ answer(sorry) := by
+  sorry
+
+/--
+Best construction to date, due to Kreisel and Kurz, has n = 7.
```
Paul-Lez:
```suggestion
The best construction to date, due to Kreisel and Kurz, has $n = 7$.
```

FormalConjectures/ErdosProblems/213.lean
```diff
@@ -0,0 +1,49 @@
...
+# Erdős Problem 213
+
+*Reference:* [erdosproblems.com/213](https://www.erdosproblems.com/213)
+-/
+
+open scoped EuclideanGeometry
+
+namespace Erdos213
+
+def Erdos213For (n : ℕ) : Prop := ∃ S : Set ℝ², S.Finite ∧ S.ncard = n ∧
```
Paul-Lez, line 33:
```suggestion
/--
The predicate (on $n$) that there exist $n$ points in $\mathbb{R}^2$, no three on a line and no four on a circle,
such that all pairwise distances are integers?
-/
def Erdos213For (n : ℕ) : Prop := ∃ S : Set ℝ², S.Finite ∧ S.ncard = n ∧
```

PR #598, opened 2025-08-20
Feat: formalize Erdős Problem 851

Paul-Lez:


FormalConjectures/ErdosProblems/851.lean
```diff
@@ -0,0 +1,54 @@
...
+
+/-!
+# Erdős Problem 851
+
+*Reference:* [erdosproblems.com/851](https://www.erdosproblems.com/851)
+-/
+
+namespace Erdos851
+
+#check Set.upperDensity
```
smmercuri: left in by mistake?

Paul-Lez: Oups indeed.

PR #599, opened 2025-08-20
Feat: add formalisation of Erdős Problem 830

Paul-Lez:


FormalConjectures/ErdosProblems/830.lean
```diff
@@ -0,0 +1,91 @@
...
+-/
+@[mk_iff]
+structure IsAmicable (a b : ℕ) where
+  left : σ 1 a = a + b
+  right : σ 1 b = a + b
+
+/--
+Let $A(x)$ counts the number of amicable $1\leq a\leq b\leq x$.
+-/
+noncomputable abbrev A (x : ℝ) : ℝ :=
```
smmercuri, line 41: What do you think about making `x` to have type `Nat`? Would make things simpler and IIUC `Real` typing is probably not needed?

Paul-Lez: I generally follow the convention of using the real typing when `x` is the variable name (e.g. in analogy with the prime counting function and so on). But I agree here this doesn't matter too much.

callesonne (approved):
LGTM!

PR #600, opened 2025-08-20
Feat: add formalisation of Conjecture 1.10 in 2501.03234 

### LINKED ISSUE #596: 2501.03234 Conjecture 1.10
### What is the conjecture
This conjecture located in [2501.0324](https://arxiv.org/pdf/2501.03234). 
It concerns the sums:
$$P(h, k):=\sum_{j=1}^{k-1}(-1)^{j+1+[h j / k]}$$ and $$S(k):=\sum_{h=1}^{k-1} P(h, k)$$
for $$h$$ and $$k$$ positive integers and [ ] is the standard floor function. The conjecture states that for odd prime $$k$$ that $$S(k)>0$$.

### Prerequisites needed
All required content is in mathlib already.

* ams-11

### Choose either option

YanYablonovskiy:
Adding Conjecture 1.10 from [2501.03234](https://arxiv.org/pdf/2501.03234).

closes #596

### DIFF
--- a/FormalConjectures/Arxiv/2501.03234/Conjecture1_10.lean
+++ b/FormalConjectures/Arxiv/2501.03234/Conjecture1_10.lean
@@ -1,0 +1,29 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Conjecture 1.10
+
+*Reference:* [arxiv.org/pdf/2501.03234](https://arxiv.org/pdf/2501.03234)
+**THEOREMS AND CONJECTURES ON AN ARITHMETIC SUM ASSOCIATED WITH THE CLASSICAL THETA FUNCTION θ₃**
+by *Yan Yablonovskiy*
+-/
+
+
+open BigOperators
+open Finset
+open Nat
--- a/FormalConjectures/Arxiv/2501.03234/Conjecture1_10.lean
+++ b/FormalConjectures/Arxiv/2501.03234/Conjecture1_10.lean
@@ -24,6 +24,8 @@ by *Yan Yablonovskiy*
 -/
 
 
-open BigOperators
-open Finset
-open Nat
+
+open Nat Finset BigOperators in
+theorem Conjecture1_10 (k: ℕ) (hprim: Nat.Prime k) (hodd: Odd k) : (∑ h ∈ range (k-2), (∑ j ∈ range (h-2),
+    (-1:ℤ)^( (j+1) + 1 + (floor ((h+1)*(j+1) / k)) ) )) > 0 := by
+  sorry
--- a/FormalConjectures/Arxiv/2501.03234/Conjecture1_10.lean
+++ b/FormalConjectures/Arxiv/2501.03234/Conjecture1_10.lean
@@ -26,6 +26,11 @@ by *Yan Yablonovskiy*
 
 
 open Nat Finset BigOperators in
-theorem Conjecture1_10 (k: ℕ) (hprim: Nat.Prime k) (hodd: Odd k) : (∑ h ∈ range (k-2), (∑ j ∈ range (h-2),
-    (-1:ℤ)^( (j+1) + 1 + (floor ((h+1)*(j+1) / k)) ) )) > 0 := by
+/--
+**Conjecture 1.10**: For any odd prime k, the sum associated with the classical theta function θ₃
+is positive.
+-/
+@[category research open, AMS 11]
+theorem Conjecture1_10 (k: ℕ) (hprim: Nat.Prime k) (hodd: k > 2) : (∑ h ∈ range (k-2),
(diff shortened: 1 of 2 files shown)


FormalConjectures/Arxiv/2501.03234/Conjecture1_10.lean
```diff
@@ -0,0 +1,36 @@
...
+
+open Nat Finset BigOperators in
+/--
+**Conjecture 1.10**: For any odd prime k, the sum associated with the classical theta function θ₃
+is positive.
+-/
+@[category research open, AMS 11]
+theorem Conjecture1_10 (k: ℕ) (hprim: Nat.Prime k) (hodd: k > 2) : (∑ h ∈ range (k-2),
+    (∑ j ∈ range (h-2),(-1:ℤ)^( (j+1) + 1 + ⌊(h + 1) * (j + 1) / k⌋₊ ) ) ) > 0 := by
+  sorry
```
mo271:
```suggestion

/--
Define the sum
$$S'(h, k) := \sum_{j=1}^{k-1}(-1)^{j + 1 + \lfloor \frac{hj}{k}\rfloor}.$$
-/
-- Using "S'" instead of "S", like it is in the paper, to avoid overloading function name.
private def S' (h k : ℕ) : ℤ := ∑ j ∈ Finset.Ico 1 k, (-1 : ℤ) ^ (j + 1 + ⌊(h * j : ℚ) / k⌋₊)

/--
Define the sum
$$S(k) := \sum_{h=1}^{k-1}S'(h, k)$$
-/
private def S (k : ℕ) : ℤ := ∑ h ∈ Finset.Ico 1 k, S' h k

-- note that in Table 1 in  https://arxiv.org/abs/2501.03234v1, there seems to be an error: 11 appears twice
example : List.map S (List.range 10) = [0, 0, 1, 2, 5, 4, 7, 10, 11, 8] := by
  unfold S
  decide +kernel

/--
**Conjecture 1.10**: For any odd prime k, the sum associated with the classical theta function θ₃, $S(k)$ is positive.
-/
@[category research open, AMS 11]
theorem Conjecture1_10 (k : ℕ) (hk : 0 < k) (hprim: Nat.Prime k) (hodd: Odd k) :
    0 < S k := by
  sorry
```

As stated the theorem is false: take k=3, then the outer sum only runs over `{0}`, and hence is 0. Also there was unintentional Nat division (we need to cast to Q before) and the index shift was too confusing, so I changed it to be more aligned with the paper. 

I propose to add definitions. 
Funny that in the paper a value is calculated incorrectly.

FormalConjectures/Arxiv/2501.03234/Conjecture1_10.lean
```diff
@@ -0,0 +1,36 @@
...
+-/
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Conjecture 1.10
+
+*Reference:* [arxiv.org/pdf/2501.03234](https://arxiv.org/pdf/2501.03234)
+**THEOREMS AND CONJECTURES ON AN ARITHMETIC SUM ASSOCIATED WITH THE CLASSICAL THETA FUNCTION θ₃**
+by *Yan Yablonovskiy*
+-/
```
mo271:
```suggestion
*Reference:* [arxiv/2501.03234](https://arxiv.org/abs/2501.03234)
**Theorems and Conjectures on an Arithmetic Sum Associated with the Classical Theta Function θ3**
by *Bruce C. Berndt, Raghavendra N. Bhat, Jeffrey L. Meyer, Likun Xie, Alexandru Zaharescu*
-/
```

YanYablonovskiy:
> Thanks, that's great, those are exactly the conjecture we are looking for!
> 
> Consider also adding the other conjectures from the paper. (Not necessarily in this pull request...)

No worries! And yea, I was waiting to get this PR first in order to avoid mistakes. But once this one is verified, the others will follow suit.

YanYablonovskiy:
Proposing adding the namespace as well since there are definitions now, for the future conjectures.

FormalConjectures/Arxiv/2501.03234/Conjecture1_10.lean
```diff
@@ -0,0 +1,56 @@
...
+-- note that in Table 1 in  https://arxiv.org/abs/2501.03234v1, there seems to be an error: 11 appears twice
+example : List.map S (List.range 10) = [0, 0, 1, 2, 5, 4, 7, 10, 11, 8] := by
+  unfold S
+  decide +kernel
+
+/--
+**Conjecture 1.10**: For any odd prime k, the sum associated with the classical theta function θ₃, $S(k)$ is positive.
+-/
+@[category research open, AMS 11]
+theorem Conjecture1_10 (k : ℕ) (hk : 0 < k) (hprim: Nat.Prime k) (hodd: Odd k) :
```
Paul-Lez: Positivity is implied by the other assumptions;)
```suggestion
theorem Conjecture1_10 (k : ℕ) (hprim: Nat.Prime k) (hodd: Odd k) :
```

FormalConjectures/Arxiv/2501.03234/Conjecture1_10.lean
```diff
@@ -0,0 +1,56 @@
...
+-/
+def S (k : ℕ) : ℤ := ∑ h ∈ Finset.Ico 1 k, S' h k
+
+-- note that in Table 1 in  https://arxiv.org/abs/2501.03234v1, there seems to be an error: 11 appears twice
+example : List.map S (List.range 10) = [0, 0, 1, 2, 5, 4, 7, 10, 11, 8] := by
+  unfold S
+  decide +kernel
+
+/--
+**Conjecture 1.10**: For any odd prime k, the sum associated with the classical theta function θ₃, $S(k)$ is positive.
```
Paul-Lez: In the paper, isn't this Conjecture 1.1 rather than 1.10?

mo271: (if that is changed, then also rename the file and the Title of the docstring...)

mo271:
Thanks again, @YanYablonovskiy!

PR #601, opened 2025-08-21
Inscribed square problem addendum

peabrainiac:
One more solved special case of the inscribed square problem that was added in #451. See the review comments there.


peabrainiac:
The result is relatively old (1929) and listed in the survey article that is already listed as a reference, so no extra reference is needed.

Paul-Lez (approved):
Nice!

PR #604, opened 2025-08-21
Proof of `IsSidon.avoids_isAPOfLength_three`

smmercuri:


FormalConjectures/ForMathlib/Combinatorics/Basic.lean
```diff
@@ -28,9 +28,26 @@ coincidences forced by the commutativity of addition. -/
...
+    rw [Set.eq_of_subset_of_ncard_le (Set.inter_subset_right) (by linarith)]
+  have ha : a ∈ A := Set.mem_of_mem_inter_left <| hss (hY ▸ ⟨0, by norm_num, by simp⟩)
+  have ha₁ : a + d ∈ A := Set.mem_of_mem_inter_left <| hss (hY ▸ ⟨1, by norm_num, by simp⟩)
+  have ha₂ : a + 2 • d ∈ A := Set.mem_of_mem_inter_left <| hss (hY ▸ ⟨2, by norm_num, by simp⟩)
+  have := hA _ ha _ ha₁ _ ha₂ _ ha₁ (by simp; omega)
+  simp at this
+  simp [hY, this.1, Set.setOf_and] at hY_card
+  linarith [Set.ncard_singleton _ ▸ Set.ncard_inter_le_ncard_right {a | ∃ x, x < 3} {a}]
+
+instance (A : Finset ℕ) : Decidable (IsSidon A.toSet) :=
```
Paul-Lez, line 49: One small thought (maybe for a later PR): would it be worth refactoring the definition of `IsSidon` to use `SetLike` rather than `Set`? That way you wouldn't need to use `.toSet` here.

Paul-Lez (approved):

FormalConjectures/ForMathlib/Combinatorics/Basic.lean
smmercuri: Yeah that would be nice, I'll follow up with another PR

PR #605, opened 2025-08-21
feat: Sum of three cubes: 42 and rationals

### LINKED ISSUE #581: Sum of three cubes
### What is the conjecture

Wikipedia: https://en.wikipedia.org/wiki/Sums_of_three_cubes

For any $n \not\equiv 4, 5 \pmod{9}$, there exist $x, y, z \in \mathbb{Z}$ such that $x^3 + y^3 + z^3 = n$.

There are famous and notorious examples such as 33 and 42 where the solutions are recently found by Booker and Sutherland. 
Known for all $n < 100$, and up to $n < 1000$, it is open for 114, 390, 627, 633, 732, 921, and 975. Also, there are infinite families of solutions for $n = 1, 2$. Also, any integer (in fact, any rational number) is a sum of three *rational* cubes: see [here](https://math.stackexchange.com/questions/494424/proving-that-any-rational-number-can-be-represented-as-the-sum-of-the-cubes-of-t) [Ryley 1825, Richmond 1930].

### Prerequisites needed
None

* ams-11

### Choose either option

seewoo5:
Closes #581.

The solution for 42 is added. Also, Ryley's result on sum of three rational cubes. Fix AMS tag.

### DIFF
--- a/FormalConjectures/Wikipedia/SumOfThreeCubes.lean
+++ b/FormalConjectures/Wikipedia/SumOfThreeCubes.lean
@@ -28,24 +28,22 @@ An integer `n : ℤ` can be written as a sum of three cubes (of integers) if and
 asked by user [*David Feldman*](https://mathoverflow.net/users/10909/david-feldman)
 -/
 
-/-- The predicate that `n : ℤ` is a sum of three (integer) cubes. -/
-def IsSumOfThreeCubes (n : ℤ) : Prop :=
-  ∃ x y z : ℤ, n = x^3 + y^3 + z^3
+variable {R : Type*} [Ring R]
 
-/-- The predicate that `n : ℚ` is a sum of three *rational* cubes. -/
-def IsSumOfThreeCubesRat (n : ℚ) : Prop :=
-  ∃ x y z : ℚ, n = x^3 + y^3 + z^3
+/-- The predicate that `n : R` is a sum of three (integer) cubes. -/
+def IsSumOfThreeCubes (n : R) : Prop :=
+  ∃ x y z : R, n = x^3 + y^3 + z^3
 
 @[category test, AMS 11]
-theorem isSumOfThreeCubes_2 : IsSumOfThreeCubes 2 :=
+theorem isSumOfThreeCubes_2 : IsSumOfThreeCubes (2 : ℤ) :=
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/SumOfThreeCubes.lean
```diff
@@ -32,6 +32,10 @@ asked by user [*David Feldman*](https://mathoverflow.net/users/10909/david-feldm
 def IsSumOfThreeCubes (n : ℤ) : Prop :=
   ∃ x y z : ℤ, n = x^3 + y^3 + z^3
 
+/-- The predicate that `n : ℚ` is a sum of three *rational* cubes. -/
+def IsSumOfThreeCubesRat (n : ℚ) : Prop :=
+  ∃ x y z : ℚ, n = x^3 + y^3 + z^3
+
```
mo271: I would suggest to generalize here instead of copying the definition:
```
variable {R : Type*} [Ring R]

/-- The predicate that `n : R` is a sum of three cubes. -/
def IsSumOfThreeCubes (n : R) : Prop :=
  ∃ x y z : R, n = x^3 + y^3 + z^3
```
Then one needs to explicitely give the type below: replace `2` by `(2 : ℤ) ` etc.
```
theorem isSumOfThreeCubes_2 : IsSumOfThreeCubes (2 : ℤ) :=

``

File: FormalConjectures/Wikipedia/SumOfThreeCubes.lean [hunk 2]
```diff
@@ -30,18 +30,24 @@ asked by user [*David Feldman*](https://mathoverflow.net/users/10909/david-feldm
 
 namespace SumOfThreeCubes
 
-/-- The predicate that `n : ℤ` is a sum of three (integer) cubes. -/
-def IsSumOfThreeCubes (n : ℤ) : Prop :=
-  ∃ x y z : ℤ, n = x^3 + y^3 + z^3
+variable {R : Type*} [Ring R]
+
+/-- The predicate that `n : R` is a sum of three (integer) cubes. -/
```
mo271:
```suggestion
/-- The predicate that `n : R` is a sum of three cubes. -/
```

mo271 (approved):
Thanks, nice contribution!

PR #606, opened 2025-08-21
feat: Add formalization of Moser's Worm problem and related theorems

### LINKED ISSUE #593: Moser's Worm Problem
### What is the conjecture

[Wikipedia](https://en.wikipedia.org/wiki/Moser%27s_worm_problem)

Moser's worm problem asks: "What is the minimal area of a shape in the plane that can be used to cover any length-1 curve by rotation and translation?"

There are also variants of the problem that specify that the shape should be convex, or that the curves should start and end at the same point.

### Prerequisites needed

It would be necessary to have some way of describing the set of "curves of length one". This might require calculus, but it could also be stated more simply in terms of "curves of length at most one" by describing such curves as 1-Lipschitz functions from the unit interval.

* ams-52 Convex and discrete geometry

BoltonBailey:
This PR fixes #593 

* Adds a formalization of Moser's worm problem in terms of 1-Lipschitz maps from the unit interval.
* Adds a variant for convex shapes
* Adds a number of explicit bounds from the literature.
* Also adds a high-school level theorem, which I partially proved as a sanity check (although I noticed Mathlib's metric space API could stand to be enhanced, so I didn't include the proof itself).

### DIFF
--- a/FormalConjectures/Wikipedia/MoserWorm.lean
+++ b/FormalConjectures/Wikipedia/MoserWorm.lean
@@ -28,14 +28,14 @@ open EuclideanGeometry MeasureTheory
 The set of worms is the set of curves of length (at most) 1.
 We formalize this as the set of ranges of 1-Lipschitz functions from `[0,1]` to `ℝ²`.
 -/
-def Worms : Set (Set (ℝ²)) :=
+def Worms : Set (Set ℝ²) :=
     {s | ∃ f : (Set.Icc (0 : ℝ) (1 : ℝ)) → ℝ², LipschitzWith 1 f ∧ Set.range f = s}
 
 /--
 The set of covers is the set of (measurable) sets
 that cover every worm by translation and rotation (i.e. through an isometry).
 -/
-def WormCovers : Set (Set (ℝ²)) :=
+def WormCovers : Set (Set ℝ²) :=
     {X | MeasurableSet X ∧ ∀ w ∈ Worms, ∃ iso, Isometry iso ∧ w ⊆ iso '' X}
 
 /--
@@ -97,7 +97,7 @@ Acta Mathematica Sinica, 49 (4): 835–846, MR 2264090.
 -/
 @[category research solved, AMS 52]
 theorem convex_mosers_worm_problem_upper_bound :
-    ∃ X : (Set (ℝ²)), MeasurableSet X ∧ Convex ℝ X ∧ volume X = 0.270911861 ∧
+    ∃ X : Set ℝ², MeasurableSet X ∧ Convex ℝ X ∧ volume X = 0.270911861 ∧
       ∀ w ∈ Worms, ∃ iso, Isometry iso ∧ w ⊆ iso '' X := by
   sorry
 
--- a/FormalConjectures/Wikipedia/MoserWorm.lean
+++ b/FormalConjectures/Wikipedia/MoserWorm.lean
@@ -54,9 +54,7 @@ of a shape that can cover every unit-length curve?
 -/
 @[category research open, AMS 52]
 theorem mosers_worm_problem :
-    IsGLB
-      {v | ∃ X ∈ WormCovers, volume X = v}
-      answer(sorry) := by
+    IsGLB {v | ∃ X ∈ WormCovers, volume X = v} answer(sorry) := by
   sorry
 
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/MoserWorm.lean
```diff
@@ -0,0 +1,125 @@
...
+Khandhawit, Tirasan; Pagonakis, Dimitrios; Sriswasdi, Sira (2013),
+"Lower Bound for Convex Hull Area and Universal Cover Problems",
+International Journal of Computational Geometry & Applications,
+23 (3): 197–212, arXiv:1101.5638, doi:10.1142/S0218195913500076, MR 3158583, S2CID 207132316.
+-/
+@[category research solved, AMS 52]
+theorem convex_mosers_worm_problem_lower_bound :
+    0.232239 ∈
+      lowerBounds
+      {v | ∃ X ∈ WormCovers, Convex ℝ X ∧ volume X = v } := by
```
Paul-Lez: I'd be tempted to have this all on the same line
```suggestion
    0.232239 ∈ lowerBounds {v | ∃ X ∈ WormCovers, Convex ℝ X ∧ volume X = v} := by
```

FormalConjectures/Wikipedia/MoserWorm.lean
```diff
@@ -0,0 +1,125 @@
...
+
+/--
+**Moser's Worm Problem**
+What is the minimal area (or greatest lower bound on the area)
+of a shape that can cover every unit-length curve?
+-/
+@[category research open, AMS 52]
+theorem mosers_worm_problem :
+    IsGLB
+      {v | ∃ X ∈ WormCovers, volume X = v }
```
Paul-Lez: small nitpick (and same below)
```suggestion
      {v | ∃ X ∈ WormCovers, volume X = v}
```

Paul-Lez (approved):
Nice! Thanks for the contribution!

FormalConjectures/Wikipedia/MoserWorm.lean
```diff
@@ -0,0 +1,116 @@
...
+-/
+def Worms : Set (Set (ℝ²)) :=
+    {s | ∃ f : (Set.Icc (0 : ℝ) (1 : ℝ)) → ℝ², LipschitzWith 1 f ∧ Set.range f = s}
+
+/--
+The set of covers is the set of (measurable) sets
+that cover every worm by translation and rotation (i.e. through an isometry).
+-/
+def WormCovers : Set (Set (ℝ²)) :=
+    {X | MeasurableSet X ∧ ∀ w ∈ Worms, ∃ iso, Isometry iso ∧ w ⊆ iso '' X}
```
Paul-Lez: Very small nit: these inner brackets shouldn't be needed
```suggestion
def Worms : Set (Set ℝ²) :=
    {s | ∃ f : (Set.Icc (0 : ℝ) (1 : ℝ)) → ℝ², LipschitzWith 1 f ∧ Set.range f = s}

/--
The set of covers is the set of (measurable) sets
that cover every worm by translation and rotation (i.e. through an isometry).
-/
def WormCovers : Set (Set ℝ²) :=
    {X | MeasurableSet X ∧ ∀ w ∈ Worms, ∃ iso, Isometry iso ∧ w ⊆ iso '' X}
```

FormalConjectures/Wikipedia/MoserWorm.lean
```diff
@@ -0,0 +1,116 @@
...
+/--
+There is a convex set of area 0.270911861 that covers all worms.
+
+*Reference:*
+Wang, Wei (2006), "An improved upper bound for the worm problem",
+Acta Mathematica Sinica, 49 (4): 835–846, MR 2264090.
+-/
+@[category research solved, AMS 52]
+theorem convex_mosers_worm_problem_upper_bound :
+    ∃ X : (Set (ℝ²)), MeasurableSet X ∧ Convex ℝ X ∧ volume X = 0.270911861 ∧
```
Paul-Lez:
```suggestion
    ∃ X : Set ℝ², MeasurableSet X ∧ Convex ℝ X ∧ volume X = 0.270911861 ∧
```

Paul-Lez (approved):

PR #610, opened 2025-08-23
Erdős Problem 155

### LINKED ISSUE #391: Erdős Problem 155
### What is the conjecture
    
https://www.erdosproblems.com/155

  Let $F(N)$ be the size of the largest Sidon subset of $\\{1,\ldots,N\\}$. Is it true that for every $k\geq 1$ we have 
$$F(N+k)\leq F(N)+1$$
for all sufficiently large $N$?

Status: open

### Prerequisites needed
None, we already have Sidon sets in `ForMathlib`.

rao107:
Fixes #391

### DIFF
--- a/FormalConjectures/ErdosProblems/155.lean
+++ b/FormalConjectures/ErdosProblems/155.lean
@@ -40,7 +40,9 @@ for all sufficiently large $N$?
 -/
 @[category research open, AMS 5]
 theorem erdos_155 :
-    ∀ᶠ N in atTop, ∀ k ≥ 1, F (N + k) ≤ F N + 1 := by
+    (∀ k ≥ 1, ∀ᶠ N in atTop, F (N + k) ≤ F N + 1) ↔ answer(sorry) := by
   sorry
+
+-- TODO: This may even hold with $k \approx ε * N ^ (1 / 2)$.
 
 end Erdos155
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/155.lean
```diff
@@ -0,0 +1,46 @@
...
+/--
+Is it true that for every $k \geq 1$ we have
+$$
+F(N + k) \leq F(N) + 1
+$$
+for all sufficiently large $N$?
+-/
+@[category research open, AMS 5]
+theorem erdos_155 :
+    ∀ᶠ N in atTop, ∀ k ≥ 1, F (N + k) ≤ F N + 1 := by
```
mo271:
```suggestion
    ∀ k ≥ 1, ∀ᶠ N in atTop,F (N + k) ≤ F N + 1 := by
```

Check out problem 4 here: https://users.renyi.hu/~sos/1994_On_Sum_Sets_of_Sidon_Sets_I.pdf
I think the `N` is allowed to depend on `k`, hence I propose to change the order of the quantifiers.

mo271: also since this is a question, can you put the entire statement in parenthesis and then write `\iff answer(sorry)` (see some other problems for examples of this)

rao107: Thanks for catching that! I've added your suggestions

PR #612, opened 2025-08-25
Erdős Problem 723

rao107:
Adds [Erdős Problem 723](https://www.erdosproblems.com/723).

I'm not sure if this is actually due to Erdős. This seems to be the Prime Power Conjecture. I looked through the linked paper and I couldn't find it. I think it might be better to move the contents to `FormalConjectures/Other/PrimePower.lean` and import it into `FormalConjectures/ErdosProblems/723.lean`, similar to the Brocard Problem.

### DIFF
--- a/FormalConjectures/ErdosProblems/723.lean
+++ b/FormalConjectures/ErdosProblems/723.lean
@@ -1,0 +1,55 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 723
+
+Also known as the prime power conjecture.
+
+*Reference:* [erdosproblems.com/723](https://www.erdosproblems.com/723)
+-/
+
+open Configuration
+
+namespace Erdos723
+
+variable {P L : Type} [Membership P L]
+
+/--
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/723.lean
```diff
@@ -0,0 +1,55 @@
...
+-/
+@[category research open, AMS 5]
+theorem erdos_732 : (∀ pp : ProjectivePlane P L, IsPrimePow pp.order) ↔ answer(sorry) := by
+  sorry
+
+/--
+These always exist if $n$ is a prime power.
+-/
+@[category research solved, AMS 5]
+theorem erdos_732.prime_pow_is_projplane_order :
```
rao107, line 44: I wasn't too sure about the name here. If anyone has a better name, I'd gladly change it.

FormalConjectures/ErdosProblems/723.lean
```diff
@@ -0,0 +1,55 @@
...
+    ∀ n, IsPrimePow n → ∃ pp : ProjectivePlane P L, pp.order = n := by
+  sorry
+
+/--
+This conjecture has been proved for $n \leq 11$.
+-/
+@[category research solved, AMS 5]
+theorem erdos_732.leq_11 : ∀ pp : ProjectivePlane P L, pp.order ≤ 11 → IsPrimePow pp.order := by
+  sorry
+
```
mo271, line 54:
```suggestion

/--
It is open whether there exists a projective plane of order 12.
-/
@[category research open, AMS 5]
theorem erdos_732.eq_12 : (∃ pp : ProjectivePlane P L, pp.order = 12) ↔ answer(sorry) :=  by
  sorry

```

Let's add the first open case?!

FormalConjectures/ErdosProblems/723.lean
mo271: looks good to me, we often also use ".variants" in the name.

mo271 (approved):
LGTM, thanks!

Perhaps you could also add the Bruck and Ryser result mentioned (or leave a TODO for that)

rao107:
> Perhaps you could also add the Bruck and Ryser result mentioned (or leave a TODO for that)

Added Bruck and Ryser theorem and the first open case :slightly_smiling_face:

mo271:
Thanks again!

PR #614, opened 2025-08-26
Abstract `IsSidon` 

smmercuri:
Redefine `IsSidon` for any type `S` with the membership property. This allows `IsSidon A` to work when `A` is a `Finset` without having to write `A.toSet`.

### DIFF
--- a/FormalConjectures/ForMathlib/Combinatorics/Basic.lean
+++ b/FormalConjectures/ForMathlib/Combinatorics/Basic.lean
@@ -94,12 +94,12 @@ namespace Finset
 theorem isSidon_toSet {A : Finset α} : IsSidon A.toSet ↔ IsSidon A := by
   simp [IsSidon]
 
-instance (A : Finset ℕ) : Decidable (IsSidon A) :=
+instance (A : Finset α) [DecidableEq α] : Decidable (IsSidon A) :=
   decidable_of_iff (∀ᵉ (i₁ ∈ A) (j₁ ∈ A) (i₂ ∈ A) (j₂ ∈ A), _) <| by rfl
 
-/-- The maximum size of a Sidon set in `{1, ..., N}`. -/
-noncomputable def maxSidonSetSize (N : ℕ) : ℕ :=
-  sSup {(A.card) | (A : Finset ℕ) (_ : A ⊆ Finset.Icc 1 N) (_ : IsSidon A)}
+/-- The maximum size of a Sidon set in the supplied `Finset`. -/
+def IsSidon.maxSubsetCard (A : Finset α) [DecidableEq α] : ℕ :=
+  (A.powerset.filter IsSidon).sup Finset.card
 
 theorem IsSidon.insert {A : Finset α} {m : α} [DecidableEq α] [IsRightCancelAdd α]
     [IsLeftCancelAdd α] (hA : IsSidon A) :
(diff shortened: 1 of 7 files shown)


FormalConjectures/ForMathlib/Combinatorics/Basic.lean
```diff
@@ -89,32 +81,54 @@ theorem IsSidon.insert {A : Set α} {m : α} [IsRightCancelAdd α] [IsLeftCancel
...
+
+@[simp, push_cast]
+theorem isSidon_toSet {A : Finset α} : IsSidon A.toSet ↔ IsSidon A := by
+  simp [IsSidon]
+
+instance (A : Finset ℕ) : Decidable (IsSidon A) :=
+  decidable_of_iff (∀ᵉ (i₁ ∈ A) (j₁ ∈ A) (i₂ ∈ A) (j₂ ∈ A), _) <| by rfl
+
+/-- The maximum size of a Sidon set in `{1, ..., N}`. -/
+noncomputable def maxSidonSetSize (N : ℕ) : ℕ :=
```
Paul-Lez: I think the name here should change, or the definition (i.e. the name should refer to Icc, or the definition should be for arbitrary Finsets, perhaps with an extra definition for the `Icc 1 N` case)

FormalConjectures/ForMathlib/Combinatorics/Basic.lean
```diff
@@ -89,32 +81,54 @@ theorem IsSidon.insert {A : Set α} {m : α} [IsRightCancelAdd α] [IsLeftCancel
     · simp_all
       exact fun _ _ _ _ _ ↦ by simp_all [add_comm]
 
+end Set
+
+namespace Finset
+
+@[simp, push_cast]
+theorem isSidon_toSet {A : Finset α} : IsSidon A.toSet ↔ IsSidon A := by
```
Paul-Lez, line 94: I think this theorem should be also be proven for general objects that have the `SetLike` instance (and we can then remove the Finset special case once leanprover-community/mathlib4#28241 is merged)

FormalConjectures/ForMathlib/Combinatorics/Basic.lean
smmercuri: Agreed, done!

FormalConjectures/ForMathlib/Combinatorics/Basic.lean
```diff
@@ -89,32 +85,55 @@ theorem IsSidon.insert {A : Set α} {m : α} [IsRightCancelAdd α] [IsLeftCancel
...
+-- TODO: remove once https://github.com/leanprover-community/mathlib4/pull/28241 is merged
+@[simp, push_cast]
+theorem isSidon_toSet {A : Finset α} : IsSidon A.toSet ↔ IsSidon A := by
+  simp [IsSidon]
+
+instance (A : Finset α) [DecidableEq α] : Decidable (IsSidon A) :=
+  decidable_of_iff (∀ᵉ (i₁ ∈ A) (j₁ ∈ A) (i₂ ∈ A) (j₂ ∈ A), _) <| by rfl
+
+/-- The maximum size of a Sidon set in the supplied `Finset`. -/
+def IsSidon.maxSubsetCard (A : Finset α) [DecidableEq α] : ℕ :=
```
Paul-Lez: Small nitpick: I'm not sure the `IsSidon` namespace is the best here since `IsSidon` doesn't appear in the hypotheses. Perhaps the following is better?
```suggestion
def maxSidonSubsetCard (A : Finset α) [DecidableEq α] : ℕ :=
```

b-mehta:
I just noticed this, I think this change is a bad one, see previous discussion here https://github.com/google-deepmind/formal-conjectures/pull/287#discussion_r2177810321

PR #618, opened 2025-08-27
Erdős Problem 828

rao107:
Adds [Erdős Problem 828](https://www.erdosproblems.com/828).

Though the site specifies _a_ to be an integer, the original problem does not specify this ([link](https://users.renyi.hu/~p_erdos/1983-27.pdf)). I elected to make it a natural number to avoid coercions with the totient function.

### DIFF
--- a/FormalConjectures/ErdosProblems/828.lean
+++ b/FormalConjectures/ErdosProblems/828.lean
@@ -35,17 +35,17 @@ theorem erdos_828 : (∀ a : ℤ, Set.Infinite {n : ℕ | ↑(φ n) ∣ n + a}) 
   sorry
 
 /--
-Lehmer conjectured that $\phi(n) | n - 1$ if and only if $n$ is prime.
+When $n > 1$, Lehmer conjectured that $\phi(n) | n - 1$ if and only if $n$ is prime.
 -/
 @[category research open, AMS 11]
 theorem erdos_828.variants.lehmer_conjecture : (∀ n > 1, φ n ∣ n - 1 ↔ Prime n) ↔ answer(sorry) := by
   sorry
 
 /--
-It is an easy exercise to show that $\phi(n) | n$ if and only if $n = 2^a 3^b$.
+It is an easy exercise to show that, for $n > 1$, $\phi(n) | n$ if and only if $n = 2^a 3^b$.
 -/
 @[category undergraduate, AMS 11]
-theorem erdos_828.variants.phi_dvd_self_iff_pow2_pow3 : ∀ n, φ n ∣ n ↔ ∃ a b, n = 2^a * 3^b := by
+theorem erdos_828.variants.phi_dvd_self_iff_pow2_pow3 : ∀ n > 1, φ n ∣ n ↔ ∃ a b, n = 2^a * 3^b := by
   sorry
 
 end Erdos828
--- a/FormalConjectures/ErdosProblems/828.lean
+++ b/FormalConjectures/ErdosProblems/828.lean
@@ -31,7 +31,7 @@ Is it true that, for any $a \in \mathbb{Z}$, there are infinitely many $n$ such 
 $$\phi(n) | n + a$$?
 -/
 @[category research open, AMS 11]
-theorem erdos_828 : (∀ a : ℤ, Infinite {n : ℕ | ↑(φ n) ∣ n + a}) ↔ answer(sorry) := by
+theorem erdos_828 : (∀ a : ℤ, Set.Infinite {n : ℕ | ↑(φ n) ∣ n + a}) ↔ answer(sorry) := by
   sorry
 
 /--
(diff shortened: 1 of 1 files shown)


smmercuri:
I think we might want `a` be an integer in this case in order to include subtractions as well (e.g., `ϕ(n)∣n − 1`)?

rao107:
> I think we might want `a` be an integer in this case in order to include subtractions as well (e.g., `ϕ(n)∣n − 1`)?

Oh! Good point, I just realized Lehmer's conjecture could be "proven" false if we use 0 (since 0 - 1 = 0 for natural numbers). I've changed it so the first two conjectures use integers instead.

smmercuri:
> > I think we might want `a` be an integer in this case in order to include subtractions as well (e.g., `ϕ(n)∣n − 1`)?
> 
> Oh! Good point, I just realized Lehmer's conjecture could be "proven" false if we use 0 (since 0 - 1 = 0 for natural numbers). I've changed it so the first two conjectures use integers instead.

Actually looking at https://en.wikipedia.org/wiki/Lehmer%27s_totient_problem Lehmer is really asking whether there are any composite numbers such that  `ϕ(n)∣n − 1`, so we should just exclude `0` and `1` anyway

FormalConjectures/ErdosProblems/828.lean
```diff
@@ -0,0 +1,51 @@
...
+-/
+@[category research open, AMS 11]
+theorem erdos_828 : (∀ a : ℤ, Infinite {n : ℕ | ↑(φ n) ∣ n + a}) ↔ answer(sorry) := by
+  sorry
+
+/--
+Lehmer conjectured that $\phi(n) | n - 1$ if and only if $n$ is prime.
+-/
+@[category research open, AMS 11]
+theorem erdos_828.variants.lehmer_conjecture : (∀ n, (φ n : ℤ) ∣ ↑n - 1 ↔ Prime n) ↔ answer(sorry) := by
```
smmercuri: wrt to my comment on the main page, Lehmer's Conjecture only concerns composites so we can stay in `Nat` for this one
```suggestion
theorem erdos_828.variants.lehmer_conjecture : (∀ n > 1, φ n ∣ n - 1 ↔ Prime n) ↔ answer(sorry) := by
```

FormalConjectures/ErdosProblems/828.lean
```diff
@@ -0,0 +1,51 @@
...
+namespace Erdos828
+
+open scoped Nat
+
+/--
+Is it true that, for any $a \in \mathbb{Z}$, there are infinitely many $n$ such that
+$$\phi(n) | n + a$$?
+-/
+@[category research open, AMS 11]
+theorem erdos_828 : (∀ a : ℤ, Infinite {n : ℕ | ↑(φ n) ∣ n + a}) ↔ answer(sorry) := by
```
smmercuri: While this is fine, `Infinite` is a predicate that applies to types, while `{n : ...}` is a set. I think it's preferred to use `Set.Infinite` here instead, which avoids coercing the set to a type.
```suggestion
theorem erdos_828 : (∀ a : ℤ, Set.Infinite {n : ℕ | ↑(φ n) ∣ n + a}) ↔ answer(sorry) := by
```

rao107:
Updated with your suggestions

FormalConjectures/ErdosProblems/828.lean
```diff
@@ -0,0 +1,51 @@
...
+-/
+@[category research open, AMS 11]
+theorem erdos_828.variants.lehmer_conjecture : (∀ n > 1, φ n ∣ n - 1 ↔ Prime n) ↔ answer(sorry) := by
+  sorry
+
+/--
+It is an easy exercise to show that $\phi(n) | n$ if and only if $n = 2^a 3^b$.
+-/
+@[category undergraduate, AMS 11]
+theorem erdos_828.variants.phi_dvd_self_iff_pow2_pow3 : ∀ n, φ n ∣ n ↔ ∃ a b, n = 2^a * 3^b := by
```
smmercuri: This I think is false for `n = 0` because `φ 0 ∣ 0` but we cannot write `0` as `2 ^ a * 3 ^ b`

FormalConjectures/ErdosProblems/828.lean
```diff
@@ -0,0 +1,51 @@
...
+/--
+Is it true that, for any $a \in \mathbb{Z}$, there are infinitely many $n$ such that
+$$\phi(n) | n + a$$?
+-/
+@[category research open, AMS 11]
+theorem erdos_828 : (∀ a : ℤ, Set.Infinite {n : ℕ | ↑(φ n) ∣ n + a}) ↔ answer(sorry) := by
+  sorry
+
+/--
+Lehmer conjectured that $\phi(n) | n - 1$ if and only if $n$ is prime.
```
smmercuri: may as well update the docstring to reflect the change here
```suggestion
When $n > 1$, Lehmer conjectured that $\phi(n) | n - 1$ if and only if $n$ is prime.
```

FormalConjectures/ErdosProblems/828.lean
rao107: Yeah, I just added another hypothesis to make sure n > 1 like Lehmer's conjecture

Issue #511, opened 2025-07-29
Erdős Problem 359


mo271:
### What is the conjecture
    
https://www.erdosproblems.com/359

  Let $a_1<a_2<\cdots$ be an infinite sequence of integers such that $a_1=n$ and $a_{i+1}$ is the least integer which is not a sum of consecutive earlier $a_j$ s. What can be said about the density of this sequence?In particular, in the case $n=1$, can one prove that $a_k/k\to \infty$ and $a_k/k^{1+c}\to 0$ for any $c>0$?

Status: open

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

Paul-Lez:
#620

franzhusch:
Was added in #620

PR #620, opened 2025-08-27
Feat: formalise Erdős Problem 359

Paul-Lez:
#511

### DIFF
--- a/FormalConjectures/ErdosProblems/359.lean
+++ b/FormalConjectures/ErdosProblems/359.lean
@@ -24,32 +24,11 @@ import FormalConjectures.Util.ProblemImports
 
 namespace Erdos359
 
--- open MeasureTheory
--- open Polynomial
--- open scoped BigOperators
--- open scoped Classical
--- open scoped ENNReal
--- open scoped EuclideanGeometry
--- open scoped InnerProductSpace
--- open scoped intervalIntegral
--- open scoped List
--- open scoped Matrix
--- open scoped Nat
--- open scoped NNReal
--- open scoped Pointwise
--- open scoped ProbabilityTheory
--- open scoped Real
--- open scoped symmDiff
--- open scoped Topology
+open Filter Asymptotics
 
-
-
-
--- open Filter Asymptotics
-
-/-- The predicate that `A 0 = n` and for all `j`, `A (j + 1)` is the smallest natural number that
+/-- The predicate that `A` is monotone, `A 0 = n` and for all `j`, `A (j + 1)` is the smallest natural number that
 cannot be written as a sum of consecutive terms of `A 0, ..., A j` -/
-def IsGoodFor (A : ℕ → ℕ) (n : ℕ) : Prop := A 0 = n ∧
+def IsGoodFor (A : ℕ → ℕ) (n : ℕ) : Prop := A 0 = n ∧ StrictMono A ∧
   ∀ j, IsLeast
     {m : ℕ | A j < m ∧ ∀ a b, Finset.Icc a b ⊆ Finset.Iic j → m ≠ ∑ i ∈ Finset.Icc a b, A i}
     (A <| j + 1)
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/359.lean
```diff
@@ -0,0 +1,75 @@
...
+def IsGoodFor (A : ℕ → ℕ) (n : ℕ) : Prop := A 0 = n ∧
+  ∀ j, IsLeast
+    {m : ℕ | ∀ a b, Finset.Icc a b ⊆ Finset.Iic j → m ≠ ∑ i ∈ Finset.Icc a b, A i}
+    (A <| j + 1)
+
+/-- Let $a_1< a_2 < ⋯ $ be an infinite sequence of integers such that $a_1=1$ and $a_{i+1}$ is the
+least integer which is not a sum of consecutive earlier $a_j$s. Show that $a_k / k \to \infty$. -/
+@[category research open, AMS 11]
+theorem erdos_359.parts.i (A : ℕ → ℕ) (hA : IsGoodFor A 1) :
+    atTop.Tendsto (fun k ↦ A k / k) atTop := by
```
smmercuri: I guess it doesn't matter since it goes to infinity anyway, but maybe cast to `Rat` or `Real` to avoid nat division?

FormalConjectures/ErdosProblems/359.lean
```diff
@@ -0,0 +1,75 @@
...
+theorem erdos_359.variants.isGoodFor_1_low_values (A : ℕ → ℕ) (hA : IsGoodFor A 1) :
+    A '' (Set.Iic 7) = {1, 2, 4, 5, 8, 10, 14, 15} := by
+  sorry
+
+/-- Suppose sequence $A$ satisfies the following: `A 0 = n` and for all `j`, `A (j + 1)` is the
+smallest natural number that cannot be written as a sum of consecutive terms of `A 0, ..., A j`.
+Then $A$ is strictly increasing. -/
+@[category test, AMS 11]
+theorem erdos_359.variants.strictMono_of_isGoodFor (A : ℕ → ℕ) (n : ℕ) (hA : IsGoodFor A n) :
+    StrictMono A := by
```
smmercuri: I don't think this is true given the definition of `IsGoodFor` for `n > 1`. For example, if `n = 2`, then `A 0 = 2` and `A 1 = 1`. This assumption might need to go in the definition of `IsGoodFor` itself

FormalConjectures/ErdosProblems/359.lean
Paul-Lez: Good point, done!

FormalConjectures/ErdosProblems/359.lean
Paul-Lez: Good point - fixed!

FormalConjectures/ErdosProblems/359.lean
```diff
@@ -0,0 +1,75 @@
...
+
+*Reference:* [erdosproblems.com/359](https://www.erdosproblems.com/359)
+-/
+
+namespace Erdos359
+
+open Filter Asymptotics
+
+/-- The predicate that `A 0 = n` and for all `j`, `A (j + 1)` is the smallest natural number that
+cannot be written as a sum of consecutive terms of `A 0, ..., A j` -/
```
smmercuri: This no longer matches the formal statement, we should include that `A (j + 1)` is greater than `A j` (or equivalently that `A` is strict mono

FormalConjectures/ErdosProblems/359.lean
smmercuri: This is basically by definition now, so I'm not sure it warrants inclusion as a test. Proof:
```lean
  intro a b hab
  induction hab with
  | refl => exact (hA.2 a).1.1
  | step _ a_ih => exact a_ih.trans (hA.2 _).1.1
```

Paul-Lez: Good point, removed.

smmercuri (approved):

PR #621, opened 2025-08-27
Erdős Problem 74

### LINKED ISSUE #304: Erdős Problem 74
### What is the conjecture
    
https://www.erdosproblems.com/74

  Let $f(n)\to \infty$ (possibly very slowly). Is there a graph of infinite chromatic number such that every finite subgraph on $n$ vertices can be made bipartite by deleting at most $f(n)$ edges?

### Prerequisites needed
Mathematical context that needs to be formalized so that this conjecture can be added.
Consider linking any open Mathlib issues / PRs here.

mo271:
fixes #304

### DIFF
--- a/FormalConjectures/ErdosProblems/74.lean
+++ b/FormalConjectures/ErdosProblems/74.lean
@@ -33,16 +33,68 @@ open Erdos74
 universe u
 variable {V : Type u}
 
-
-
+/--
+For a given subgraph `A`, this is the set of all numbers `k` such that `A` can be made
+bipartite by deleting `k` edges.
+-/
 def SimpleGraph.edgeDistancesToBipartite {G : SimpleGraph V} (A : G.Subgraph) : Set ℕ :=
   { (E.ncard) | (E : Set (Sym2 V)) (_ : IsBipartite (A.deleteEdges E).coe)}
 
+/--
+The set of edge distances to a bipartite graph is always non-empty because deleting all edges
+from a graph makes it bipartite.
+-/
+@[category test, AMS 5]
+theorem SimpleGraph.edgeDistancesToBipartite_nonempty {G : SimpleGraph V} (A : G.Subgraph) :
+    SimpleGraph.edgeDistancesToBipartite A |>.Nonempty := by
+  dsimp only [edgeDistancesToBipartite,Set.nonempty_def]
+  refine ⟨_, Set.univ, ?_, rfl⟩
+  use fun v => 0
+  simp
+
+/--
+The minimum number of edges that must be deleted from a subgraph `A` to make it bipartite.
+-/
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/74.lean
```diff
@@ -0,0 +1,80 @@
...
+noncomputable def SimpleGraph.max_subgraph_bipartite_edge_deletion_num
+    (G : SimpleGraph V) (n : ℕ) : ℕ :=
+  sSup { (sInf { (E.ncard) | (E : Set (Sym2 V)) (_ : IsBipartite (A.deleteEdges E).coe)}) |
+    (A : Subgraph G) (_ : (A.verts.ncard = n))}
+
+
+/--
+Deleting all edges will always be sufficient to make the graph bipartite.
+-/
+@[category undergraduate, AMS 5]
```
Paul-Lez: Another sanity check that would be useful to add here would be that the set we're taking a sup of is always bounded above (so that taking the sup doesn't result in a junk value)

FormalConjectures/ErdosProblems/74.lean
```diff
@@ -0,0 +1,80 @@
...
+
+/--
+For a given Graph $G$, define a number such that any subgraph of G
+on $n$ vertices can be made bipartite after deleting at most this many edges.
+This is Definition 3.1 in [EHS82].
+
+[EHS82] Erdős, P. and Hajnal, A. and Szemerédi, E.,
+  *On almost bipartite large chromatic graphs* Theory and practice of combinatorics (1982), 117-123.
+-/
+noncomputable def SimpleGraph.max_subgraph_bipartite_edge_deletion_num
```
Paul-Lez: nit: since this is a def I guess we should use the `camelCase` convention?
```suggestion
noncomputable def SimpleGraph.maxSubgraphBipartiteEdgeDeletionNum
```

FormalConjectures/ErdosProblems/74.lean
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Erdos74
+
+open Erdos74
+
+universe u
+variable {V : Type u}
+
+/--
+For a given Graph $G$, define a number such that any subgraph of G
```
Paul-Lez: Shouldn't the docstring mention that this number is optimal (in a suitable sense?)

Paul-Lez (approved):
Nice!

FormalConjectures/ErdosProblems/74.lean
mo271: adding two sanity checks now: inf is `Nonempty` and sup is `BddAbove`

FormalConjectures/ErdosProblems/74.lean
mo271: good point, done

PR #623, opened 2025-08-27
chore: fix formalisation of Artin's conjecture

Paul-Lez:


FormalConjectures/Wikipedia/ArtinPrimitiveRootsConjecture.lean
```diff
@@ -76,33 +76,35 @@ theorem artin_primitive_roots.variants.part_ii_square_or_minus_one
 
 /--
 **Artin's Conjecture on Primitive Roots**, second half, power version
-If `a` is a perfect pth power for prime `p`, then the density of the set `S(a)` of
-primes `p` such that `a` is a primitive root modulo `p`
-is given by `p(p-2) / (p^2 - p - 1) * C` where `C` is Artin's constant. If there are
+If `a` is a perfect pth power for an odd prime `p` of a number that isn't `≡ 1 [MOD 4]`,
```
smmercuri:
```suggestion
If `a` is a perfect pth power for an odd prime `p` of a number whose squarefree part isn't `≡ 1 [MOD 4]`,
```

FormalConjectures/Wikipedia/ArtinPrimitiveRootsConjecture.lean
```diff
@@ -76,33 +76,35 @@ theorem artin_primitive_roots.variants.part_ii_square_or_minus_one
...
 more than one such prime `p`, then the number needs to be multiplied by
 ``p(p-2) / (p^2 - p - 1)` for all such primes `p`.
 -/
 @[category research open, AMS 11]
-theorem artin_primitive_roots.variants.part_ii_prime_power
-    (a m b : ℕ) (ha : a = b^m) (hb : ∀ u v, 1 < u → b ≠ v^u) (hm₁ : 1 < m)
-    (hm₂ : m.primeFactorsList.Nodup)  :
+theorem artin_primitive_roots.variants.part_ii_prime_power_squarefree_not_modeq_one
+    (a m b : ℕ) (ha : a = b ^ m) (hb : ∀ u v, 1 < u → b ≠ v^u) (hm₁ : 1 < m)
+    (hm₂ : m.primeFactorsList.Nodup) (hm₃ : Odd m) (hb' : ¬ b ≡ 1 [MOD 4]):
```
smmercuri: `hb'` needs to only check the squarefree part of `b` isn't `1 (mod 4)`, rather than `b` itself

smmercuri: This might be useful #627

FormalConjectures/Wikipedia/ArtinPrimitiveRootsConjecture.lean
```diff
@@ -76,33 +76,35 @@ theorem artin_primitive_roots.variants.part_ii_square_or_minus_one
...
-``p(p-2) / (p^2 - p - 1)` for all such primes `p`.
+`p(p-2) / (p^2 - p - 1)` for all such primes `p`.
 -/
 @[category research open, AMS 11]
-theorem artin_primitive_roots.variants.part_ii_prime_power
-    (a m b : ℕ) (ha : a = b^m) (hb : ∀ u v, 1 < u → b ≠ v^u) (hm₁ : 1 < m)
-    (hm₂ : m.primeFactorsList.Nodup)  :
+theorem artin_primitive_roots.variants.part_ii_prime_power_squarefree_not_modeq_one
+    (a m b : ℕ) (ha : a = b ^ m) (hb : ∀ u v, 1 < u → b ≠ v^u) (hm₁ : 1 < m)
+    (hm₃ : Odd m) (hb' : ¬ b.squarefreePart ≡ 1 [MOD 4]) :
```
smmercuri: nit: I guess `hm₃` can go back to `hm₂`?

smmercuri: Or do we want to keep `hm₂` from before?

smmercuri: I don't think it's necessary looking at https://www.numericana.com/answer/constants.htm#artin

FormalConjectures/Wikipedia/ArtinPrimitiveRootsConjecture.lean
```diff
@@ -76,33 +76,35 @@ theorem artin_primitive_roots.variants.part_ii_square_or_minus_one
...
 more than one such prime `p`, then the number needs to be multiplied by
-``p(p-2) / (p^2 - p - 1)` for all such primes `p`.
+`p(p-2) / (p^2 - p - 1)` for all such primes `p`.
 -/
 @[category research open, AMS 11]
-theorem artin_primitive_roots.variants.part_ii_prime_power
-    (a m b : ℕ) (ha : a = b^m) (hb : ∀ u v, 1 < u → b ≠ v^u) (hm₁ : 1 < m)
-    (hm₂ : m.primeFactorsList.Nodup)  :
+theorem artin_primitive_roots.variants.part_ii_prime_power_squarefree_not_modeq_one
+    (a m b : ℕ) (ha : a = b ^ m) (hb : ∀ u v, 1 < u → b ≠ v^u) (hm₁ : 1 < m)
```
smmercuri, line 87: I think it's probably enough to assert `(hb : ¬ IsSquare b)` here so that the first row is avoided, but I can't remember whether there was some edge case requiring `hb` of this form

FormalConjectures/Wikipedia/ArtinPrimitiveRootsConjecture.lean
```diff
@@ -76,33 +76,35 @@ theorem artin_primitive_roots.variants.part_ii_square_or_minus_one
...
 /--
-**Artin's Conjecture on Primitive Roots**, second half, perfect prime power version
-Write `a = a_0 b^2` where `a_0` is squarefree.
-If `a_0` is congruent to `1 mod 4`, then the density of the set `S(a)` of primes `p` such that `a` is a primitive root modulo `p`
-is given by `C * ∏_p, p(p-1) / (p^2 - p - 1)` where `C` is Artin's constant and the product is taken over the prime factors `p` of `a_0`
+**Artin's Conjecture on Primitive Roots**, second half, power version
+If `a` is a perfect mth power (for an odd `m`) of a number whose squarefree part is `≡ 1 [MOD 4]`,
+then the density of the set `S(a)` of primes `p` such that `a` is a primitive root modulo `p`
+is given by $$C \left(\prod_{p \mid m} \frac{p(p-2)}{(p ^ 2 - p - 1)}\right)
+\left(\prod_{p \mid a} \frac{p(p-1)}{(p ^ 2 - p - 1)}\right)$ where $C$ is Artin's constant.
```
smmercuri: I think this is not quite right. In this case one first hits the second row of the table in https://www.numericana.com/answer/constants.htm#artin, giving $C(b)\left(\prod_{p \mid m} \frac{p(p-2)}{(p ^ 2 - p - 1)}\right)$. Now we're assuming that $sf(b) = 1 \pmod{4}$, so that $C(b)$ hits the third row of the table. This then gives $C\left(\prod_{p \mid m} \frac{p(p-2)}{(p ^ 2 - p - 1)}\right)\left(1 - \prod_{q \mid sf(b)}\frac{1}{(1 + q - q^2)}\right)$

smmercuri: The factors included in the Wikipedia page seems to be coming from the special case when `a` is a prime $\equiv 1\pmod{4}$ (see below the table)

FormalConjectures/Wikipedia/ArtinPrimitiveRootsConjecture.lean
Paul-Lez: I think we also want `m` to be maximal?

FormalConjectures/Wikipedia/ArtinPrimitiveRootsConjecture.lean
Paul-Lez: Done

FormalConjectures/Wikipedia/ArtinPrimitiveRootsConjecture.lean
Paul-Lez: Indeed, good point. I've updated the formalisation and the docstring.

FormalConjectures/Wikipedia/ArtinPrimitiveRootsConjecture.lean
```diff
@@ -76,33 +76,36 @@ theorem artin_primitive_roots.variants.part_ii_square_or_minus_one
...
-theorem artin_primitive_roots.variants.part_ii
-    (a a_0 b : ℕ) (ha : a = a_0 * b^2) (ha' : ∀ n m, m ≠ 1 → a ≠ n^m)
-    (ha' : Squarefree a) (ha_0' : a_0 ≡ 1 [MOD 4]):
+theorem artin_primitive_roots.variants.part_ii_prime_power_squarefree_modeq_one
+    (a m b : ℕ) (ha : a = b ^ m) (hb : ∀ u v, 1 < u → b ≠ v^u) (hm₁ : 1 < m)
+    (hm₂ : m.primeFactorsList.Nodup) (hm₃ : Odd m) (hb' : b.squarefreePart ≡ 1 [MOD 4]):
     (S a).HasDensity
-      (ArtinConstant * ∏ p ∈ a_0.primeFactors, p * (p-1 : ℝ) / (p^2 - p - 1))
+      (ArtinConstant * (∏ p ∈ m.primeFactors, p * (p - 2 : ℝ) / (p ^ 2 - p - 1))
+        * (1 - ∏ q ∈ b.primeFactors, 1 / (1 + q - q ^ 2 : ℝ)))
```
smmercuri: I think that should be `b.squarefreePart.primeFactors`?
```suggestion
        * (1 - ∏ q ∈ b.squarefreePart.primeFactors, 1 / (1 + q - q ^ 2 : ℝ)))
```

FormalConjectures/Wikipedia/ArtinPrimitiveRootsConjecture.lean
smmercuri: right yes, so this doesn't really matter in the table of https://www.numericana.com/answer/constants.htm#artin because it is applied iteratively, but in our formulation we do need it since we're giving the final answer

smmercuri (changes requested):

FormalConjectures/Wikipedia/ArtinPrimitiveRootsConjecture.lean
Paul-Lez: Yeah exactly

FormalConjectures/Wikipedia/ArtinPrimitiveRootsConjecture.lean
```diff
@@ -76,33 +76,36 @@ theorem artin_primitive_roots.variants.part_ii_square_or_minus_one
...
-If `a_0` is congruent to `1 mod 4`, then the density of the set `S(a)` of primes `p` such that `a` is a primitive root modulo `p`
-is given by `C * ∏_p, p(p-1) / (p^2 - p - 1)` where `C` is Artin's constant and the product is taken over the prime factors `p` of `a_0`
+**Artin's Conjecture on Primitive Roots**, second half, power version
+If `a` is a perfect mth power (for an odd `m`) of a number `b` whose squarefree part is `≡ 1 [MOD 4]`,
+then the density of the set `S(a)` of primes `p` such that `a` is a primitive root modulo `p`
+is given by $$C \left(\prod_{p \mid m} \frac{p(p-2)}{(p ^ 2 - p - 1)}\right)
+\left(\prod_{q \mid \mathrm{sf}(b)} \frac{1}{(1 + q - q ^ 2)}\right)$ where $C$ is Artin's constant,
+and $\mathrm{sf}(n)$ denotes the squarefree part of a natural number $n$.
 -/
 @[category research open, AMS 11]
```
smmercuri:
```suggestion
-/
-- TODO: There might be a missing factor of `μ b` here, see https://mast.queensu.ca/~murty/intelligencer.pdf
@[category research open, AMS 11]
```

smmercuri:
Let's get this merged as it now matches the table. However, I think there might be a factor of $\mu(b)$ missing from the website we were using as reference (see https://mast.queensu.ca/~murty/intelligencer.pdf). This would explain the switch in negative sign from the Wikipedia page as well. Let's add the suggested TODO for now and revisit later

smmercuri (approved):

PR #624, opened 2025-08-27
chore: change tests from `example` to `theorem`

Paul-Lez:
The goal here is to ensure that all test are accounted for when we count entries that have the `category test` attribute.


mo271 (approved):
LGTM, do we want to lint against using `example` at some point?

Paul-Lez:
I think linting might be a good idea, but I'm not sure whether we want to fully ban examples - perhaps we should lint against examples that have a category attribute?

FormalConjectures/Arxiv/2501.03234/Conjecture_1_1.lean
```diff
@@ -41,7 +41,7 @@ $$S(k) := \sum_{h=1}^{k-1}S'(h, k)$$
 def S (k : ℕ) : ℤ := ∑ h ∈ Finset.Ico 1 k, S' h k
 
 -- note that in Table 1 in  https://arxiv.org/abs/2501.03234v1, there seems to be an error: 11 appears twice
-example : List.map S (List.range 10) = [0, 0, 1, 2, 5, 4, 7, 10, 11, 8] := by
+theorem S_values_first_ten : List.map S (List.range 10) = [0, 0, 1, 2, 5, 4, 7, 10, 11, 8] := by
```
mo271: this is missing a category tag (also was missing it before...)

Will be fixed after rebase/merge. This file was deleted/moved

Paul-Lez: Fixed (by merging!)

Issue #546, opened 2025-08-07
Feit-Thompson conjecture for primes


rbsmith-google:
### What is the conjecture
The [Feit-Thompon conjecture](https://en.wikipedia.org/wiki/Feit%E2%80%93Thompson_conjecture) states that for all primes p and q, (q^p - 1) / (q - 1) \nmid (p^q - 1) / (p - 1)

### Prerequisites needed
No work required.

### [AMS categories](https://github.com/google-deepmind/formal-conjectures/labels?q=ams-11)

* ams-11

### Choose either option
- [X] I plan on adding this conjecture to the repository
- [ ] This issue is up for grabs: I would like to see this conjecture added by somebody else

franzhusch:
Was solved in #625

PR #625, opened 2025-08-28
Add the Feit-Thompson conjecture for primes to the wikipedia problems

rbsmith-google:
The Feit-Thompson conjecture for primes is an open problem: https://en.wikipedia.org/wiki/Feit%E2%80%93Thompson_conjecture

This is issue https://github.com/google-deepmind/formal-conjectures/issues/546

### DIFF
--- a/FormalConjectures/Wikipedia/FeitThompsonPrimeConjecture.lean
+++ b/FormalConjectures/Wikipedia/FeitThompsonPrimeConjecture.lean
@@ -26,6 +26,6 @@ import FormalConjectures.Util.ProblemImports
 There are no distinct primes $p$ and $q$ such that $\frac{q^p - 1}{q - 1}$ divides $\frac{p^q - 1}{p - 1}$
 -/
 @[category research open, AMS 11]
-theorem feit_thompson_primes (p q : ℕ) (hp : Prime p) (hq : Prime q) (h : p < q) :
+theorem feit_thompson_primes (p q : ℕ) (hp : p.Prime) (hq : q.Prime) (h : p < q) :
     ¬ (q ^ p - 1) / (q - 1) ∣ (p ^ q - 1) / (p - 1) := by
   sorry
--- a/FormalConjectures/Wikipedia/FeitThompsonPrimeConjecture.lean
+++ b/FormalConjectures/Wikipedia/FeitThompsonPrimeConjecture.lean
@@ -1,0 +1,31 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Feit-Thompson conjecture on primes
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Feit%E2%80%93Thompson_conjecture)
+-/
+
+/--
+Are there any primes p and q such that (q^p - 1) / (q - 1) divides (p^q - 1) / (p - 1)?
+-/
+@[category research open, AMS 11]
+theorem feit_thompson_primes (p q : ℕ) (hp : Prime p ∧ Prime q) (h : p < q) :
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/FeitThompsonPrimeConjecture.lean
```diff
@@ -0,0 +1,31 @@
...
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Feit-Thompson conjecture on primes
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Feit%E2%80%93Thompson_conjecture)
+-/
+
+/--
+Are there any primes p and q such that (q^p - 1) / (q - 1) divides (p^q - 1) / (p - 1)?
```
mo271:
```suggestion
Are there any primes p and q such that $\frac{q^p - 1}{q - 1}$ divides $\frac{p^q - 1}{p - 1}$?
```

FormalConjectures/Wikipedia/FeitThompsonPrimeConjecture.lean
```diff
@@ -0,0 +1,31 @@
...
+-/
+@[category research open, AMS 11]
+theorem feit_thompson_primes (p q : ℕ) (hp : Prime p ∧ Prime q) (h : p < q) :
```
mo271:
```suggestion
theorem feit_thompson_primes (p q : ℕ) (hp : Prime p) (hq : Prime q) (h : p < q) :
```
let's split those

mo271 (approved):
Nice, thanks!
Two small comments, but other than that LGTM.

I suppose we could also add the solved variants for q =2 and q =3 at some point.

FormalConjectures/Wikipedia/FeitThompsonPrimeConjecture.lean
```diff
@@ -0,0 +1,31 @@
...
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Feit%E2%80%93Thompson_conjecture)
+-/
+
+/--
+Are there any primes p and q such that $\frac{q^p - 1}{q - 1}$ divides $\frac{p^q - 1}{p - 1}$?
+-/
+@[category research open, AMS 11]
+theorem feit_thompson_primes (p q : ℕ) (hp : Prime p) (hq : Prime q) (h : p < q) :
+  ¬ (q ^ p - 1) / (q - 1) ∣ (p ^ q - 1) / (p - 1) := by
```
smmercuri: nit: statements are indented four spaces
```suggestion
    ¬ (q ^ p - 1) / (q - 1) ∣ (p ^ q - 1) / (p - 1) := by
```

FormalConjectures/Wikipedia/FeitThompsonPrimeConjecture.lean
```diff
@@ -0,0 +1,31 @@
...
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Feit-Thompson conjecture on primes
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Feit%E2%80%93Thompson_conjecture)
+-/
+
+/--
+Are there any primes p and q such that $\frac{q^p - 1}{q - 1}$ divides $\frac{p^q - 1}{p - 1}$?
```
smmercuri: The Conjecture on the wiki page and the formal statement are on the lines of "There are no primes p and q such that ..." so I would have the informal statement reflect that. 
```suggestion
There are no distinct primes $p$ and $q$ such that $\frac{q^p - 1}{q - 1}$ divides $\frac{p^q - 1}{p - 1}$
```

FormalConjectures/Wikipedia/FeitThompsonPrimeConjecture.lean
```diff
@@ -0,0 +1,31 @@
...
+# Feit-Thompson conjecture on primes
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Feit%E2%80%93Thompson_conjecture)
+-/
+
+/--
+There are no distinct primes $p$ and $q$ such that $\frac{q^p - 1}{q - 1}$ divides $\frac{p^q - 1}{p - 1}$
+-/
+@[category research open, AMS 11]
+theorem feit_thompson_primes (p q : ℕ) (hp : Prime p) (hq : Prime q) (h : p < q) :
```
smmercuri: Note that `hp` and `hq` are using `_root_.Prime` rather than `Nat.Prime`. They are equivalent via [`Nat.prime_iff`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Nat/Prime/Defs.html#Nat.prime_iff) but I think we might prefer use `Nat.Prime` here instead so that there's direct access to all the `Nat.Prime` API
```suggestion
theorem feit_thompson_primes (p q : ℕ) (hp : p.Prime) (hq : q.Prime) (h : p < q) :
```

smmercuri:
LGTM apart from my final comment above -- thank you!

Issue #207, opened 2025-06-11
Erdős Problem 33


mo271:
### What is the conjecture
    
https://www.erdosproblems.com/33

  Let $A\subset\mathbb{N}$ be such that every large integer can be written as $n^2+a$ for some $a\in A$ and $n\geq 0$. What is the smallest possible value of 
$$\limsup \frac{\lvert A\cap\{1,\ldots,N\}\rvert}{N^{1/2}}?$$

Status: open

### Prerequisites needed
None

### Choose either option
- [x] @dturjo77  plans to work on this conjecture
- [ ] This issue is up for grabs: I would like to see this conjecture added by somebody else

dturjo77:
I’m interested in exploring this conjecture. Any suggestions on where to begin?

mo271:
Nice, perhaps best to formalize "every large integer can be written as $n^2+a$ for some $a\in A$ and $n\geq 0$" and then look into how to best to the $\limsup$ and in conjunction with the `answer\(\)` formalism.
Feel free to make a draft pull request even with partial progress!

mo271:
@dturjo77 are you still working on this?

dturjo77:
I have worked on this, but i couldn't make any good improvements. I sincerely apologize for that.

zhenningdavidliu:
Happy to give it a go if no-one is working on this right now.

Paul-Lez:
> I have worked on this, but i couldn't make any good improvements. I sincerely apologize for that.

No problem at all @dturjo77! Feel free to have a go @zhenningdavidliu:)

franzhusch:
Was added in #631

PR #630, opened 2025-08-28
Issue 207

zhenningdavidliu:
First attempt at formalising Erdos 33 (Issue #207 ).


zhenningdavidliu:
Currently the file only has the base theorem, and I plan on adding the related theorems after a bit of feedback on the first draft.

PR #631, opened 2025-08-28
Erdős Problem 33

zhenningdavidliu:
PR for #207 - Erdos 33


zhenningdavidliu:
First draft PR for Erdos 33. I plan on adding the current upper bound on the minimal value once I get some feedback on the first draft. (Also, apologies for the messy commit history, somehow messed it up)

FormalConjectures/ErdosProblems/33.lean
```diff
@@ -0,0 +1,51 @@
...
+-/
+def AdditiveBasisCondition (A : Set ℕ) : Prop :=
+  ∀ (k : ℕ), ∃ (n : ℕ) (a : ℕ), a ∈ A ∧ k = a + n^2
+
+/-- Let `A ⊆ ℕ` be a set such that every integer can be written as `n^2 + a` for some `a` in `A` and `n ≥ 0`. What is the smallest possible value of `lim sup n → ∞ |A ∩ {1, …, N}| / N^(1/2) = 0`?
+-/
+@[category research open, AMS 11]
+theorem erdos_33 : IsLeast
+    { c : ℝ | ∃ (A : Set ℕ), AdditiveBasisCondition A ∧
+      Filter.atTop.limsup (fun N => (Set.bdd A N).card / (N : ℝ)^(1/2)) = c}
```
smmercuri: `(N : ℝ)^(1/2)` is using `Nat` division in the exponent, so this is actually `N^0`. You can use `√N` instead

FormalConjectures/ErdosProblems/33.lean
```diff
@@ -0,0 +1,51 @@
...
+
+open Classical
+
+open scoped Finset
+
+namespace Erdos33
+
+/-- Given a set of natural numbers `A`, `Set.bdd A N` is the set `{1,...,N} ∩ A`-/
+private noncomputable def Set.bdd (A : Set ℕ) (N : ℕ) : Finset ℕ :=
+    Finset.Icc 1 N |>.filter (· ∈ A)
```
smmercuri: two space indents for `def`s 
```suggestion
  Finset.Icc 1 N |>.filter (· ∈ A)
```

FormalConjectures/ErdosProblems/33.lean
```diff
@@ -0,0 +1,51 @@
...
+
+/-!
+# Erdős Problem 41
+
+*Reference:* [erdosproblems.com/33](https://www.erdosproblems.com/33)
+-/
+
+open Classical
+
+open scoped Finset
```
smmercuri: Currently this `open scoped Finset` is not used or needed so it can be removed. It does enable `#` notation for `Finset.card` however, so you can replace `(Set.bdd A N).card` with `#(Set.bdd A N)` if you prefer

FormalConjectures/ErdosProblems/33.lean
zhenningdavidliu: Oh interesting, does this mean that my formalisation in [Erdos 41](https://github.com/google-deepmind/formal-conjectures/blob/3369444aa88051f0211ecc14cab5290706b334e8/FormalConjectures/ErdosProblems/41.lean#L43-L46) has the same error?

Paul-Lez: Indeed! Fixed in #637 :)

FormalConjectures/ErdosProblems/33.lean
```diff
@@ -0,0 +1,70 @@
...
+  sorry
+
+/--
+Erdos observed that this value is finite and > 1.
+-/
+@[category research solved, AMS 11]
+theorem erdos_33_i :
+    1 ∈ lowerBounds { c : ℝ | ∃ (A : Set ℕ), AdditiveBasisCondition A ∧
+      Filter.atTop.limsup (fun N => (Set.bdd A N).card / (√N : ℝ)) = c}
+    := by
```
smmercuri: styling
```suggestion
      Filter.atTop.limsup (fun N => (Set.bdd A N).card / (√N : ℝ)) = c} := by
```

FormalConjectures/ErdosProblems/33.lean
```diff
@@ -0,0 +1,70 @@
...
+-/
+def AdditiveBasisCondition (A : Set ℕ) : Prop :=
+  ∀ (k : ℕ), ∃ (n : ℕ) (a : ℕ), a ∈ A ∧ k = a + n^2
+
+/-- Let `A ⊆ ℕ` be a set such that every integer can be written as `n^2 + a` for some `a` in `A` and `n ≥ 0`. What is the smallest possible value of `lim sup n → ∞ |A ∩ {1, …, N}| / N^(1/2) = 0`?
+-/
+@[category research open, AMS 11]
+theorem erdos_33 : IsLeast
+    { c : ℝ | ∃ (A : Set ℕ), AdditiveBasisCondition A ∧
+      Filter.atTop.limsup (fun N => (Set.bdd A N).card / (√N : ℝ)) = c}
```
smmercuri: nit: I think you should be able to do without the type ascription and just write `√N` rather than `(√N : ℝ)`. Ditto below

FormalConjectures/ErdosProblems/33.lean
```diff
@@ -0,0 +1,70 @@
...
+    := by
+  sorry
+
+/--
+The smallest possible value of `lim sup n → ∞ |A ∩ {1, …, N}| / N^(1/2) = 0` is at most `2φ^(5/2) ≈ 6.66`, with `φ` equal to the golden ratio. Proven by Wouter van Doorn.
+-/
+@[category research solved, AMS 11]
+theorem erdos_33_Wouter :
+    2 * (Real.rpow ((1 + √5) / 2 ) (5/2)) ∈ lowerBounds { c : ℝ | ∃ (A : Set ℕ), AdditiveBasisCondition A ∧
```
smmercuri: If you `open scoped goldenRatio` you can use `φ` here if you like. Or you can just `Real.goldenRatio`.

smmercuri: Also `Real.rpow` isn't needed, you can use `^ ((5 : ℝ) / 2)`

smmercuri:
Feel free to add yourself to `AUTHORS` if you haven't done so yet

FormalConjectures/ErdosProblems/33.lean
zhenningdavidliu: Would you be able to explain the difference between this case and the previous one where we had `N^(1/2)`? Is it true that in that case `N^((1 : ℝ) / 2)` still would not have worked?

smmercuri: It is true that `N^((1 : ℝ) / 2)` would not have worked because `N : ℕ` and this does not have an rpow (I.e., there is no `Pow ℕ ℝ` instance). But if you coerce `N` as well then this would have worked: `(N : ℝ)^((1 : ℝ) / 2)`. Note that this is _the same_ as calling `Real.rpow N (1 / 2)`, see next para. Similarly `√N` is the same as writing `Real.sqrt N`, and handles the coercions automatically.

More generally, say `x : ℝ` and `n : ℕ`, then `x ^ n` is notation for the power function coming from the `Pow ℝ ℕ` instance. If `y : ℝ`, then `x ^ y` is notation for the power function coming from the [`Pow ℝ ℝ`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/SpecialFunctions/Pow/Real.html#Real.instPow) instance. You can see on that link that the underlying function is `Real.rpow`. Of course when you write `(1/2)` this gets automatically interpreted as `ℕ` so we're back to the first case. So you have two choices: (1) type ascript `x ^ ((1 : ℝ) / 2)`; (2) use the underlying function `Real.rpow` which will coerce for you because then it definitely knows which power function you mean. Both work, but (1) is much more idiomatic (in the same way as why we use `x | y` instead of `Dvd.dvd x y`).

zhenningdavidliu: Thanks a lot for the detailed answer!

FormalConjectures/ErdosProblems/33.lean
```diff
@@ -0,0 +1,70 @@
...
+open Classical
+open scoped goldenRatio
+
+namespace Erdos33
+
+/-- Given a set of natural numbers `A`, `Set.bdd A N` is the set `{1,...,N} ∩ A`-/
+private noncomputable def Set.bdd (A : Set ℕ) (N : ℕ) : Finset ℕ :=
+  Finset.Icc 1 N |>.filter (· ∈ A)
+
+/-- Let `A ⊆ ℕ` be a set such that every integer can be written as `n^2 + a` for some `a` in `A` and `n ≥ 0`. (Changed 'every large integer' to 'every integer' as for the statement these conditions are equivalent. Also, this was the formulation in the original paper `by Erdos.)
```
smmercuri: Typically we separate out comments/justifications for Formalisation choices away from docstrings as follows
```suggestion
/-- Let `A ⊆ ℕ` be a set such that every integer can be written as `n^2 + a` for some `a` in `A` and `n ≥ 0`. -/
-- Formalisation note: Changed 'every large integer' to 'every integer' as for the statement these 
-- conditions are equivalent. Also, this was the formulation in the original paper `by Erdos.
```

FormalConjectures/ErdosProblems/33.lean
```diff
@@ -0,0 +1,70 @@
...
+-/
+def AdditiveBasisCondition (A : Set ℕ) : Prop :=
+  ∀ (k : ℕ), ∃ (n : ℕ) (a : ℕ), a ∈ A ∧ k = a + n^2
+
+/-- Let `A ⊆ ℕ` be a set such that every integer can be written as `n^2 + a` for some `a` in `A` and `n ≥ 0`. What is the smallest possible value of `lim sup n → ∞ |A ∩ {1, …, N}| / N^(1/2) = 0`?
```
smmercuri: to keep within 100 char line limit
```suggestion
/-- Let `A ⊆ ℕ` be a set such that every integer can be written as `n^2 + a` 
for some `a` in `A` and `n ≥ 0`. What is the smallest possible value of 
`lim sup n → ∞ |A ∩ {1, …, N}| / N^(1/2) = 0`?
```

FormalConjectures/ErdosProblems/33.lean
```diff
@@ -0,0 +1,70 @@
...
+Erdos observed that this value is finite and > 1.
+-/
+@[category research solved, AMS 11]
+theorem erdos_33_i :
+    1 ∈ lowerBounds { c : ℝ | ∃ (A : Set ℕ), AdditiveBasisCondition A ∧
+      Filter.atTop.limsup (fun N => (Set.bdd A N).card / √N) = c} := by
+  sorry
+
+/--
+The smallest possible value of `lim sup n → ∞ |A ∩ {1, …, N}| / N^(1/2) = 0` is at most `2φ^(5/2) ≈ 6.66`, with `φ` equal to the golden ratio. Proven by Wouter van Doorn.
```
smmercuri:
```suggestion
The smallest possible value of `lim sup n → ∞ |A ∩ {1, …, N}| / N^(1/2) = 0` 
is at most `2φ^(5/2) ≈ 6.66`, with `φ` equal to the golden ratio. Proven by 
Wouter van Doorn.
```

FormalConjectures/ErdosProblems/33.lean
```diff
@@ -0,0 +1,76 @@
...
+namespace Erdos33
+
+/-- Given a set of natural numbers `A`, `Set.bdd A N` is the set `{1,...,N} ∩ A`-/
+private noncomputable def Set.bdd (A : Set ℕ) (N : ℕ) : Finset ℕ :=
+  Finset.Icc 1 N |>.filter (· ∈ A)
+
+/-- Let `A ⊆ ℕ` be a set such that every integer can be written as `n^2 + a` for some `a` in `A` and `n ≥ 0`. -/
+-- Formalisation note: Changed 'every large integer' to 'every integer' as for the statement these 
+-- conditions are equivalent. Also, this was the formulation in the original paper `by Erdos.
+-/
```
smmercuri: final line char fix
```suggestion
/-- Let `A ⊆ ℕ` be a set such that every integer can be written as `n^2 + a` for some `a` in `A` 
and `n ≥ 0`. -/
-- Formalisation note: Changed 'every large integer' to 'every integer' as for the statement these 
-- conditions are equivalent. Also, this was the formulation in the original paper `by Erdos.
```

smmercuri:
By the way, you can add 
```
"editor.rulers": [
        100
    ],
```
to `settings.json` in vscode to help you keep within 100 chars in the future if you like! Unlike mathlib, it's not a strict convention we follow here, but is always preferred for readability

mo271:
> By the way, you can add
> 
> ```
> "editor.rulers": [
>         100
>     ],
> ```
> 
> to `settings.json` in vscode to help you keep within 100 chars in the future if you like! Unlike mathlib, it's not a strict convention we follow here, but is always preferred for readability

I think we already have that set for this repo, so the line should already be there, not need to add it globally?!
https://github.com/google-deepmind/formal-conjectures/blob/71257280583ea7ce7ee1dbf61d03b233fe7f8c76/.vscode/settings.json#L4

FormalConjectures/ErdosProblems/33.lean
```diff
@@ -0,0 +1,76 @@
...
+    { c : ℝ | ∃ (A : Set ℕ), AdditiveBasisCondition A ∧
+      Filter.atTop.limsup (fun N => (Set.bdd A N).card / √N) = c}
+    answer(sorry) := by
+  sorry
+
+/--
+Erdos observed that this value is finite and > 1.
+-/
+@[category research solved, AMS 11]
+theorem erdos_33_i :
```
smmercuri: The naming convention for additional results is `erdos_*.variants.*`
```suggestion
theorem erdos_33.variants.one_mem_lowerBounds :
```

FormalConjectures/ErdosProblems/33.lean
```diff
@@ -0,0 +1,76 @@
...
+      Filter.atTop.limsup (fun N => (Set.bdd A N).card / √N) = c} := by
+  sorry
+
+/--
+The smallest possible value of `lim sup n → ∞ |A ∩ {1, …, N}| / N^(1/2) = 0` 
+is at most `2φ^(5/2) ≈ 6.66`, with `φ` equal to the golden ratio. Proven by 
+Wouter van Doorn.
+-/
+@[category research solved, AMS 11]
+theorem erdos_33_Wouter :
```
smmercuri:
```suggestion
theorem erdos_33.variants.vanDoorn :
```

smmercuri:
LGTM now after the final suggestions! Thank you!

smmercuri (approved):

PR #638, opened 2025-09-01
Add Erdős 107

### LINKED ISSUE #338: Erdős Problem 107: minimal number of points containing an $n$-gon
### What is the conjecture
    
https://www.erdosproblems.com/107

  Let $f(n)$ be minimal such that any $f(n)$ points in $\mathbb{R}^2$, no three on a line, contain $n$ points which form the vertices of a convex $n$-gon. Prove that $f(n)=2^{n-2}+1$.

### Prerequisites needed
 - "convex $n$-gon"

Vtec234:
Closes #338. I have used definitions from [EmptyHexagonLean](https://github.com/bsubercaseaux/EmptyHexagonLean/blob/3399900b0d2dea2e3ec790249dacab9f5331da98/Lean/Geo.lean).


FormalConjectures/ErdosProblems/107.lean
```diff
@@ -0,0 +1,121 @@
...
+
+/-- The set `S` is in convex position. -/
+def ConvexIndep (S : Set Point) : Prop :=
+  ∀ a ∈ S, a ∉ convexHull ℝ (S \ {a})
+
+/-- The set `P` contains a convex `n`-gon. -/
+def HasConvexNGon (n : ℕ) (P : Set Point) : Prop :=
+  ∃ S : Finset Point, S.card = n ∧ ↑S ⊆ P ∧ ConvexIndep S
+
+noncomputable def f (n : ℕ) : ℕ :=
```
Vtec234, line 38: I have roughly followed the formalization of Erdős 126, except that [there](https://github.com/google-deepmind/formal-conjectures/blob/ed41ca10bdce5343dfe037abc84befc685c86016/FormalConjectures/ErdosProblems/126.lean#L54) the claims are stated as 'for any $f(n)$ matching the spec, ..', whereas here I have written down $f(n)$ and claimed that it has some properties. This is slightly stronger because it implies $f(n)$ exists and is not trivial (with mathlib's `sInf`, `f n = 0` if the set `{ N ∣ .. }` is empty).

FormalConjectures/ErdosProblems/107.lean
```diff
@@ -0,0 +1,121 @@
...
+
+namespace Point
+
+/-- The three points `a,b,c` are not collinear. -/
+def InGenPos₃ (a b c : Point) : Prop :=
+  Matrix.det !![a 0, b 0, c 0 ; a 1, b 1, c 1 ; 1, 1, 1] ≠ 0
+
+/-- No three points in `S` are collinear. -/
+def SetInGenPos (S : Set Point) : Prop :=
+  ∀ ⦃p q r : Point⦄, {p,q,r} ⊆ S → List.Nodup [p,q,r] → InGenPos₃ p q r
```
mo271:
```suggestion
  ∀ ⦃p q r : Point⦄, {p,q,r} ⊆ S → {p,q,r}.ncard = 3 → InGenPos₃ p q r
```
Would that be closer to the informal "three"?
But in general I would use 
```
def NonTrilinear (A : Set ℝ²) : Prop := ∀ᵉ (x ∈ A) (y ∈ A) (z ∈ A),
  A.Triplewise (fun x y z ↦ ¬ Collinear ℝ {x, y, z})
```
and move that from https://github.com/google-deepmind/formal-conjectures/blob/ed41ca10bdce5343dfe037abc84befc685c86016/FormalConjectures/ErdosProblems/846.lean#L37 to ForMathlib. Then we can re-use it (also in FormalConjectures/ErdosProblems/213.lean)

FormalConjectures/ErdosProblems/107.lean
```diff
@@ -0,0 +1,121 @@
...
+# Erdős Problem 107
+
+*Reference:* [erdosproblems.com/107](https://www.erdosproblems.com/107)
+-/
+
+open Filter
+
+namespace Erdos107
+
+abbrev Point := EuclideanSpace ℝ (Fin 2)
```
mo271: here it would be preferred to just use 
https://github.com/google-deepmind/formal-conjectures/blob/ed41ca10bdce5343dfe037abc84befc685c86016/FormalConjectures/ForMathlib/Geometry/2d.lean#L23 the "ℝ²" notation

FormalConjectures/ErdosProblems/107.lean
```diff
@@ -0,0 +1,121 @@
...
+  ∀ ⦃p q r : Point⦄, {p,q,r} ⊆ S → List.Nodup [p,q,r] → InGenPos₃ p q r
+
+end Point
+
+/-- The set `S` is in convex position. -/
+def ConvexIndep (S : Set Point) : Prop :=
+  ∀ a ∈ S, a ∉ convexHull ℝ (S \ {a})
+
+/-- The set `P` contains a convex `n`-gon. -/
+def HasConvexNGon (n : ℕ) (P : Set Point) : Prop :=
```
mo271: perhaps **https://github.com/google-deepmind/formal-conjectures/blob/603575bec6867fa91ab460c7065e83aa00906e7f/FormalConjectures/ForMathlib/Geometry/2d.lean#L51** could be used here?

FormalConjectures/ErdosProblems/107.lean
mo271: one could make this implicit, by having another theorem showing that you set is nonempty (ne need to fill in the proof..)

FormalConjectures/ErdosProblems/107.lean
```diff
@@ -0,0 +1,121 @@
...
+theorem erdos_107 : (∀ n ≥ 3, f n = 2^(n-2) + 1) ↔ answer(sorry) := by
+  sorry
+
+/-- Depending on details of definitions,
+the statement is false or trivial for $n < 3$. -/
+@[category test, AMS 52]
+theorem erdos_107.f_zero_eq : f 0 = 0 := by
+  have : ∀ P, HasConvexNGon 0 P := by
+    intro; use ∅; simp [ConvexIndep]
+  simp [f, this]
```
mo271: Another good test statement here might be the case $n=3$.

FormalConjectures/ErdosProblems/107.lean
```diff
@@ -0,0 +1,121 @@
...
+  sInf { N | ∀ (pts : Finset Point), pts.card = N → Point.SetInGenPos pts →
+    HasConvexNGon n pts }
+
+/--
+Let $f(n)$ be minimal such that any $f(n)$ points in $ℝ^2$, no three on a line,
+contain $n$ points which form the vertices of a convex $n$-gon.
+Prove that $f(n) = 2^{n-2} + 1$.
+-/
+@[category research open, AMS 52]
+theorem erdos_107 : (∀ n ≥ 3, f n = 2^(n-2) + 1) ↔ answer(sorry) := by
```
mo271:
```suggestion
theorem erdos_107 : (∀ n ≥ 3, f n = 2^(n - 2) + 1) ↔ answer(sorry) := by
```

FormalConjectures/ErdosProblems/107.lean
```diff
@@ -0,0 +1,121 @@
...
+
+[ErSz60] Erdős, P. and Szekeres, G., _On some extremum problems in elementary geometry_.
+  Ann. Univ. Sci. Budapest. Eötvös Sect. Math. (1960/61), 53-62.
+
+[ErSz35] Erdős, P. and Szekeres, G., _A combinatorial problem in geometry_.
+  Compos. Math. (1935), 463-470.
+-/
+@[category research solved, AMS 52]
+theorem erdos_107.variants.ersz_bounds :
+    ∀ n ≥ 3, 2^(n-2) + 1 ≤ f n ∧ f n ≤ Nat.choose (2*n - 4) (n - 2) + 1 := by
```
mo271:
```suggestion
    ∀ n ≥ 3, 2^(n - 2) + 1 ≤ f n ∧ f n ≤ Nat.choose (2 * n - 4) (n - 2) + 1 := by
```

FormalConjectures/ErdosProblems/107.lean
mo271, line 1: Thanks for the great contribution! 
I would sometime in the docstring mention the name "happy ending problem". 
In fact even better: perhaps you could also add a file in the `FormalConjectures/Wikipedia/HappyEndingProblem.lean` and only include 
```
import FormalConjectures.ErdosProblems.«107»
```
under the copyright header? This way people trying to formalize the happy ending problem will have an easier time seeing that you already did it...

FormalConjectures/ErdosProblems/107.lean
Vtec234: That predicate had a silly typo, actually:
```
def NonTrilinear (A : Set ℝ²) : Prop := ∀ᵉ (x ∈ A) (y ∈ A) (z ∈ A),
  Triplewise A (fun x y z ↦ ¬ Collinear ℝ {x, y, z})
```
The three outer binders `x,y,z` are never used.

FormalConjectures/ErdosProblems/107.lean
Vtec234: It could (or maybe the `IsConvexPolygon` variant), but I slightly prefer the current formulation because (1) `Fin n -> P` has the extra structure of an ordering on `p`, yet abstractly we just want to talk about sets of points and (2) going between "all angles are CCW" and "the set is convex" requires some proof.

Would you prefer `Is(Ccw)ConvexPolygon` despite this?

FormalConjectures/ErdosProblems/107.lean
Vtec234: ~~One could! Are you suggesting to implement that, or just pointing out another formulation?~~ Added.

FormalConjectures/ErdosProblems/107.lean
Vtec234: I added the statement. I also started filling in a proof, but it's getting a bit long; I understand you don't want those in this repo. (We have a proof modulo assuming that some CNF is unsatisfiable [here](https://github.com/bsubercaseaux/EmptyHexagonLean/blob/main/Lean/Geo.lean#L46).)

FormalConjectures/ErdosProblems/107.lean
```diff
@@ -0,0 +1,129 @@
...
+def ConvexIndep (S : Set ℝ²) : Prop :=
+  ∀ a ∈ S, a ∉ convexHull ℝ (S \ {a})
+
+/-- The set `P` contains a convex `n`-gon.
+See also `IsConvexPolygon`. -/
+def HasConvexNGon (n : ℕ) (P : Set ℝ²) : Prop :=
+  ∃ S : Finset ℝ², S.card = n ∧ ↑S ⊆ P ∧ ConvexIndep S
+
+def cardSet (n : ℕ) := { N | ∀ (pts : Finset ℝ²), pts.card = N → NonTrilinear pts.toSet →
+  HasConvexNGon n pts }
```
Paul-Lez, line 35: Small nit: let's also add a docstring here and for the definition below

FormalConjectures/ErdosProblems/107.lean
```diff
@@ -0,0 +1,129 @@
...
+/--
+Let $f(n)$ be minimal such that any $f(n)$ points in $ℝ^2$, no three on a line,
+contain $n$ points which form the vertices of a convex $n$-gon.
+Prove that $f(n) = 2^{n-2} + 1$.
+-/
+@[category research open, AMS 52]
+theorem erdos_107 : (∀ n ≥ 3, f n = 2^(n - 2) + 1) ↔ answer(sorry) := by
+  sorry
+
+namespace erdos_107
```
Paul-Lez: I'd be tempted to remove this since we're already in `Erdos107` namespace
```suggestion
```

Paul-Lez (approved):
Nice work, thanks!

FormalConjectures/ErdosProblems/107.lean
Paul-Lez: I think the current formulation is reasonable! (in fact I think it would be reasonable to add it to `ForMathlib` since it's useful when we don't want to assume that extra structure)

Vtec234: Moved to `ForMathlib`.

FormalConjectures/ErdosProblems/107.lean
Vtec234: I think there's some awkwardness to adding docstrings everywhere because you end up with a lot of duplicated phrases. I documented `f` as 'the function in `erdos_107`' to avoid that to some extent.

PR #639, opened 2025-09-01
workflow to push update to teorth/erdosproblems

mo271:


mo271:
will do this slightly differently...

PR #641, opened 2025-09-01
workflow to trigger update on teorth/erdosproblems

mo271:


mo271:
needs https://github.com/teorth/erdosproblems/pull/7 in order to work.

eric-wieser (approved):

PR #645, opened 2025-09-02
 Feat: Erdős 196

### LINKED ISSUE #417: Erdős Problem 196
### What is the conjecture
    
https://www.erdosproblems.com/196

  Must every permutation of $\mathbb{N}$ contain a monotone 4-term arithmetic progression $x_1<x_2<x_3<x_4$?

Status: open

### Prerequisites needed
None, we already have arithmetic progressions in `ForMathlib`.

YanYablonovskiy:
Adding [Erdős 196](https://www.erdosproblems.com/196) . 

Closes #417.

### DIFF
--- a/FormalConjectures/ErdosProblems/196.lean
+++ b/FormalConjectures/ErdosProblems/196.lean
@@ -23,7 +23,7 @@ import FormalConjectures.Util.ProblemImports
 -/
 namespace Erdos196
 
-/-- Must every permutation of $\mathbb{N}$, contain a monotone 4-term arithmetic progression. -/
+/-- Must every permutation of $\mathbb{N}$, contain a monotone 4-term arithmetic progression?-/
 @[category research open, AMS 5 11]
 theorem erdos_196 : (∀ (f : ℕ ≃ ℕ), ∃ (a : List ℕ),
     ((a.Sorted (· < · )) ∨ (a.Sorted (· > · ))) ∧ Set.IsAPOfLength (f '' a.toFinset) 4)
--- a/FormalConjectures/ErdosProblems/196.lean
+++ b/FormalConjectures/ErdosProblems/196.lean
@@ -1,0 +1,33 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


callesonne (commented):
Thanks for your contribution!

To me the statement of the Erdos problem seems slightly ambiguous, am I right in understanding that the permutation "containing" a monotone 4-term arithmetic sequence means that there are 4 consecutive natural numbers such that the permutation sends them to a 4-term arithmetic sequence? If so, I would recommend using `Set.Icc n (n + 3)` for some given `n` as opposed to the list `a`.

YanYablonovskiy:
> Thanks for your contribution!
> 
> To me the statement of the Erdos problem seems slightly ambiguous, am I right in understanding that the permutation "containing" a monotone 4-term arithmetic sequence means that there are 4 consecutive natural numbers such that the permutation sends them to a 4-term arithmetic sequence? If so, I would recommend using `Set.Icc n (n + 3)` for some given `n` as opposed to the list `a`.

You are very right, I had the same thought [here](https://leanprover.zulipchat.com/#narrow/channel/524981-Formal-conjectures/topic/Erd.C5.91s.20196), and it does change its strength . @mo271 What do you think?

mo271:
@YanYablonovskiy @callesonne 


> Thanks for your contribution!
> 
> To me the statement of the Erdos problem seems slightly ambiguous, am I right in understanding that the permutation "containing" a monotone 4-term arithmetic sequence means that there are 4 consecutive natural numbers such that the permutation sends them to a 4-term arithmetic sequence?

I don't think that is what is meant here: What about the permutation 
$$0, 2, 4, 1, 6, 8, 10, 3, 12, 14, 16, 5, 18, 20, 22, 7, 24, ...$$
given by counting interleaving even and odd numbers increasingly, but  always going three even, then one odd?
This clearly has the property that 4 consecutive numbers contain three even and one odd, and therefore cannot be a 4-term arithmetic progression. 
Also see the intro here:
https://mathweb.ucsd.edu/~ronspubs/77_06_no_long_aps.pdf

Perhaps it would be worth clarifying this with Thomas Bloom, how about using the comment section on [Erdős 196](https://www.erdosproblems.com/196)?

mo271:
I added a comment/question to [Erdős 196](https://www.erdosproblems.com/196) , but since this is my first it is pending for review...

FormalConjectures/ErdosProblems/196.lean
```diff
@@ -0,0 +1,33 @@
...
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 196
+
+*Reference:* [erdosproblems.com/196](https://www.erdosproblems.com/196)
+-/
+namespace Erdos196
+
+/-- Must every permutation of $\mathbb{N}$, contain a monotone 4-term arithmetic progression. -/
```
mo271:
```suggestion
/-- Must every permutation of $\mathbb{N}$, contain a monotone 4-term arithmetic progression?-/
```

mo271 (approved):

PR #654, opened 2025-09-05
Add `erdos_145`

### LINKED ISSUE #381: Erdős Problem 145
### What is the conjecture
    
https://www.erdosproblems.com/145

  Let $s_1<s_2<\cdots$ be the sequence of squarefree numbers. Is it true that, for any $\alpha \geq 0$,
$$\lim_{x\to \infty}\frac{1}{x}\sum_{s_n\leq x}(s_{n+1}-s_n)^\alpha$$
exists?

Status: open

### Choose either option

smmercuri:
Fixes #381

### DIFF
--- a/FormalConjectures/ErdosProblems/145.lean
+++ b/FormalConjectures/ErdosProblems/145.lean
@@ -1,0 +1,81 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    https://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 145
+*Reference:* [erdosproblems.com/145](https://www.erdosproblems.com/145)
+-/
+
+namespace Erdos145
+
+open Filter
+open scoped Topology
+
+/-- Let $s_1 < s_2 < \cdots$ be the sequence of squarefree numbers. -/
+noncomputable abbrev s (n : ℕ) : ℕ := Nat.nth Squarefree n
+
+/-- Let $A(x)$ denote the set of indices $n$ for which $s_n \leq x$. -/
+noncomputable abbrev A (x : ℝ) : Finset ℕ :=
+  (Finset.Icc 0 ⌊x⌋₊).preimage s (Nat.nth_injective Nat.squarefree_infinite).injOn
+
+/--
+Let $s_1 < s_2 < \cdots$ be the sequence of squarefree numbers. Is it true that, for any
+$\alpha\geq 0$,
+$$
+\lim_{x\to\infty} \frac{1}{x}\sum_{s_n\leq x}(s_{n+1}-s_n)^\alpha
+$$
+exists?
+-/
+@[category research open, AMS 11]
+theorem erdos_145 :
+    (∀ α ≥ (0 : ℝ), ∃ β : ℝ,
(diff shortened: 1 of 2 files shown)


FormalConjectures/ErdosProblems/145.lean
```diff
@@ -0,0 +1,88 @@
...
+
+open Filter
+open scoped Topology
+
+/-- Let $s_1 < s_2 < \cdots$ be the sequence of squarefree numbers. -/
+noncomputable abbrev s (n : ℕ) : ℕ := Nat.nth Squarefree n
+
+/-- Let $A(x)$ denote the set of indices $n$ for which $s_n \leq x$. -/
+noncomputable abbrev A (x : ℝ) : Finset ℕ :=
+  (Finset.Icc 0 ⌊x⌋₊).preimage s (Nat.nth_injective Nat.squarefree_infinite).injOn
```
mo271, line 35: I'm a bit confused about the indexing. In the cited papers [Er51] and [Ho73], sqaurefree numbers are 1-indexed and the sum is 
$$\lim_{x\to \infty}\frac{1}{x}\sum_{s_{n + 1}\leq x}(s_{n+1}-s_n)^\alpha$$
but on erdosprolems (while also 1-indexed) the sum is

$$\lim_{x\to \infty}\frac{1}{x}\sum_{s_n\leq x}(s_{n+1}-s_n)^\alpha$$

So I think the way it is done here is consistent with what is done in the papers (but not on erdosproblems.com, perhaps worth making a comment on that in the comment section there?)

Of course the existence of the limit should not really be affected in any case...

FormalConjectures/ErdosProblems/145.lean
```diff
@@ -0,0 +1,88 @@
...
+-/
+@[category research open, AMS 11]
+theorem erdos_145 :
+    (∀ α ≥ (0 : ℝ), ∃ β : ℝ,
+      atTop.Tendsto (fun x : ℝ ↦ 1 / x * ∑ n ∈ A x, (s (n + 1) - s n : ℝ) ^ α) (𝓝 β)) ↔
+    answer(sorry) := by
+  sorry
+
+/--
+Erd\"os [Er51] proved this for all $0\leq \alpha\leq 2$.
```
mo271:
```suggestion
Erdős [Er51] proved this for all $0\leq \alpha\leq 2$.
```

FormalConjectures/ErdosProblems/145.lean
```diff
@@ -0,0 +1,88 @@
...
+theorem erdos_145.variants.le_two {α : ℝ} (hα : α ∈ Set.Icc 0 2) :
+    ∃ β : ℝ,
+      atTop.Tendsto (fun x : ℝ ↦ 1 / x * ∑ n ∈ A x, (s (n + 1) - s n : ℝ) ^ α) (𝓝 β) := by
+  sorry
+
+/--
+Hooley [Ho73] extended this to all $\alpha\leq 3$.
+
+[Ho73] Hooley, Christopher, On the intervals between consecutive terms of sequences.
+  (1973), 129-140.
```
mo271:
```suggestion
[Ho73] Hooley, Christopher, On the intervals between consecutive terms of sequences. Proc. Symp. Pure Math, vol. 24, pp. 129-140. 1973.
```

Not sure why the name of the journal is missing on erdosproblems.com, probably a bug.

mo271 (approved):
Great!

FormalConjectures/ErdosProblems/145.lean
smmercuri: Indeed I guess this does not have an impact on the final answer since it is just a shift in terms in the partial sum. 

On the other hand, I think we are aligning with the website in this case, but not with the papers. Let $s_n'$ be our zero-indexing ($s_0' = 2, s_1' = 3, s_2' = 5, ...$) and let $s_n$ be one-indexing ($s_1 = 2, s_2 = 3, ...$). If we fix $x$, then the condition $s_n' \leq x$ and $s_n \leq x$ is independent of the indexing, we just return all squarefree numbers $\leq x$. So for $x = 3$ we have two terms in the sum $(3 - 2)^{\alpha}$ and $(5 - 3)^{\alpha}$. On the other hand, in the paper's convention $s_{n + 1}\leq 3$ is only satisfied for one $n = 1$, giving only one term $(3 - 2)^{\alpha}$ in the sum.

In any case, this does not have impact on the mathematical content so I'd vote for just leaving it as stated.

mo271: sounds good to me!

PR #664, opened 2025-09-10
AlphaProof's counterexample  erdos_730.variants.delta_one

mo271:
AlphaProof found a counterexample to what we had in `erdos_730.variants.delta_one`: 10003 and 10005.

We adapt the disproof and change the negated statement's name to `erdos_730.variants.delta_ne_one.

A numerical search for more counterexamples is implemented here: https://github.com/mo271/kummer

### DIFF
--- a/FormalConjectures/ErdosProblems/730.lean
+++ b/FormalConjectures/ErdosProblems/730.lean
@@ -53,7 +53,7 @@ theorem erdos_730.variants.two_div_forall (n : ℕ) (h : 0 < n) : 2 ∣ (2*n).ch
   sorry
 
 /--
-There are examples where $(n, m) ∈ S$, we have $m ≠ n + 1$.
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/730.lean
```diff
@@ -53,10 +53,22 @@ theorem erdos_730.variants.two_div_forall (n : ℕ) (h : 0 < n) : 2 ∣ (2*n).ch
   sorry
 
 /--
-In every known example $(n, m) ∈ S$, we have $m = n + 1$.
+There are examples where $(n, m) ∈ S$, we have $m ≠ n + 1$.
```
callesonne:
```suggestion
There are examples where $(n, m) ∈ S$ with $m ≠ n + 1$.
```

PR #665, opened 2025-09-11
Erdos 946

seewoo5:


FormalConjectures/ErdosProblems/946.lean
```diff
@@ -0,0 +1,79 @@
...
+@[category research solved, AMS 11]
+theorem erdos_946.hildebrand_lower_bound : (fun x => x / (log (log x))^3) =O[atTop] Erdos946Count
+    := by
+  sorry
+
+/--
+Upper bound in [EPS87]: $O(x / \sqrt{\log \log x})$.
+-/
+@[category research solved, AMS 11]
+theorem erdos_946.upper_bound : Erdos946Count =O[atTop] (fun x => x / sqrt (log (log x))) := by
```
mo271:
```suggestion
theorem erdos_946.upper_bound : Erdos946Count =O[atTop] (fun x => x / x.log.log.sqrt) := by
```
should also work, same in some of the other theorems, but perhaps a question of taste

mo271:
LGTM, thanks!

FormalConjectures/ErdosProblems/946.lean
```diff
@@ -0,0 +1,79 @@
...
+ - [He84] Heath-Brown, D. R., The divisor function at consecutive integers. Mathematika 31 (1984), no. 2, 141--149.
+ - [Hi85] Hildebrand, A., The divisor function at consecutive integers. Pacific J. Math. (1987), 307--319
+ - [EPS87] Erdős, P., Pomerance, C., and Sarkőzy, A., On locally repeated values of arithmetic functions. III. Proc. Amer. Math. Soc. (1987), 1--7.
+-/
+
+open Filter Real
+
+namespace Erdos946
+
+private abbrev τ := fun (n : ℕ) => n.divisors.card
```
smmercuri: There is notation `σ 0` already in [mathlib](https://leanprover-community.github.io/mathlib4_docs/Mathlib/NumberTheory/ArithmeticFunction.html#ArithmeticFunction.termσ) for this if you `open scoped ArithmeticFunction`

FormalConjectures/ErdosProblems/946.lean
```diff
@@ -0,0 +1,79 @@
...
+-/
+@[category research solved, AMS 11]
+theorem erdos_946 : {n : ℕ | τ n = τ (n + 1)}.Infinite := by
+  sorry
+
+/--
+There are infinitely many $n$ such that $τ(n) = τ(n + 5040)$. Proved in [Sp81].
+-/
+@[category research solved, AMS 11]
+theorem erdos_946.spiro_5040 : {n : ℕ | τ n = τ (n + 5040)}.Infinite := by
```
smmercuri:
```suggestion
theorem erdos_946.variants.spiro_5040 : {n : ℕ | τ n = τ (n + 5040)}.Infinite := by
```
and similar for other variants in this file

FormalConjectures/ErdosProblems/946.lean
```diff
@@ -0,0 +1,79 @@
...
+  sorry
+
+/-- Number of $n \le x$ with $τ(n) = τ(n+1)$. -/
+noncomputable def Erdos946Count (x : ℝ) : ℝ
```
smmercuri: Following Mathlib naming conventions this should be as follows
```suggestion
noncomputable def erdos946Count (x : ℝ) : ℝ
```
Functions inherit the naming convention of the return value, and terms of types are in `lowerCamelCase`

FormalConjectures/ErdosProblems/946.lean
smmercuri: Could also consider `√x.log.log` as an option

mo271: even nicer!

FormalConjectures/ErdosProblems/946.lean
seewoo5: I actually considred this first, but I found that [945](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/945.lean) did in this way (and I copied from there). What do you think?

mo271: perhaps having 
```
open scoped ArithmeticFunction

private abbrev τ := σ 0
```
here and in 945 would be best? @smmercuri

smmercuri: > I actually considred this first, but I found that [945](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/945.lean) did in this way (and I copied from there). What do you think?

In general, I think that if something is already defined in mathlib we should use that. It gives API access and minimises divergence from mathlib as much as possible. In this case `τ` is easy enough to define already, and `n.divisors.card` is perhaps the more natural definition, so it's not unreasonable to have the redefinition. ~~I'd still use `σ 0` however, at least it saves a line of code!~~ (Actually you'd just replace that line with `open scoped ArithmeticFunction`). I guess either works for me, so I'll leave it up to your personal preference!

smmercuri: > perhaps having
> 
> ```
> open scoped ArithmeticFunction
> 
> private abbrev τ := σ 0
> ```
> 
> here and in 945 would be best? @smmercuri

If we are going to use `σ 0`, then I would just remove the `abbrev` and use `σ 0` inline. Otherwise I think keeping the `τ` definition is actually fine on second thoughts!

seewoo5: I replaced with `σ 0`. Probably I'll make new PR for 945.

mo271: LGTM, although now the docstring doesn't quite match the formal statement, but that is ok, I suppose. 
Perhaps when you are making PR for 945, we could even define `τ` in ForMathlib, similarly to how Mathlib defines `σ`.

smmercuri: The Erdős website does clarify the meaning of `τ`, so perhaps we could include that in the docstring also?

FormalConjectures/ErdosProblems/946.lean
```diff
@@ -0,0 +1,78 @@
...
+/-!
+# Erdős Problem 946
+
+*References:*
+ - [erdosproblems.com/946](https://www.erdosproblems.com/946)
+ - [ErMi52] Erdős, P. and Mirsky, L., The distribution of values of the divisor function {$d(n)$}. Proc. London Math. Soc. (3) (1952), 257--271.
+ - [Sp81] Spiro, C. A., The frequency with which an integral-valued, prime-independent, multiplicative or additive function of n divides a polynomial function of n.
+ - [He84] Heath-Brown, D. R., The divisor function at consecutive integers. Mathematika 31 (1984), no. 2, 141--149.
+ - [Hi85] Hildebrand, A., The divisor function at consecutive integers. Pacific J. Math. (1987), 307--319
+ - [EPS87] Erdős, P., Pomerance, C., and Sarkőzy, A., On locally repeated values of arithmetic functions. III. Proc. Amer. Math. Soc. (1987), 1--7.
```
smmercuri: some styling suggestions to keep in 100 lines 
```suggestion
*References:*
 - [erdosproblems.com/946](https://www.erdosproblems.com/946)
 - [ErMi52] Erdős, P. and Mirsky, L., The distribution of values of the divisor function {$d(n)$}. 
   Proc. London Math. Soc. (3) (1952), 257--271.
 - [Sp81] Spiro, C. A., The frequency with which an integral-valued, prime-independent, 
   multiplicative or additive function of n divides a polynomial function of n.
 - [He84] Heath-Brown, D. R., The divisor function at consecutive integers. 
   Mathematika 31 (1984), no. 2, 141--149.
 - [Hi85] Hildebrand, A., The divisor function at consecutive integers. Pacific J. Math. 
   (1987), 307--319
 - [EPS87] Erdős, P., Pomerance, C., and Sarkőzy, A., On locally repeated values of 
   arithmetic functions. III. Proc. Amer. Math. Soc. (1987), 1--7.
```

FormalConjectures/ErdosProblems/946.lean
```diff
@@ -0,0 +1,78 @@
...
+/--
+There are infinitely many $n$ such that $τ(n) = τ(n + 5040)$. Proved in [Sp81].
+-/
+@[category research solved, AMS 11]
+theorem erdos_946.variants.spiro_5040 : {n : ℕ | σ 0 n = σ 0 (n + 5040)}.Infinite := by
+  sorry
+
+/-- Number of $n \le x$ with $τ(n) = τ(n+1)$. -/
+noncomputable def erdos946Count (x : ℝ) : ℝ
+  := ((Finset.range (⌊x⌋₊ + 1)).filter (fun n => σ 0 n = σ 0 (n + 1))).card
```
smmercuri: `:=` should always go at the end of a line 
```suggestion
noncomputable def erdos946Count (x : ℝ) : ℝ :=
  ((Finset.range (⌊x⌋₊ + 1)).filter (fun n => σ 0 n = σ 0 (n + 1))).card
```

FormalConjectures/ErdosProblems/946.lean
```diff
@@ -0,0 +1,78 @@
...
+
+/--
+The number of $n \le x$ with $τ(n) = τ(n+1)$ is at least $x / (\log x)^7$ for all sufficiently large $x$.
+Proved in [He84].
```
smmercuri:
```suggestion
/--
The number of $n \le x$ with $τ(n) = τ(n+1)$ is at least $x / (\log x)^7$ for all sufficiently 
large $x$. Proved in [He84].
```

FormalConjectures/ErdosProblems/946.lean
```diff
@@ -0,0 +1,78 @@
...
+-/
+@[category research solved, AMS 11]
+theorem erdos_946.variants.heathbrown_lower_bound : (fun x => x / (x.log)^7) =O[atTop] erdos946Count := by
+  sorry
```
smmercuri:
```suggestion
theorem erdos_946.variants.heathbrown_lower_bound : 
    (fun x => x / (x.log)^7) =O[atTop] erdos946Count := by
  sorry
```

FormalConjectures/ErdosProblems/946.lean
```diff
@@ -0,0 +1,78 @@
...
+
+/--
+Improved lower bound in [Hi85]: $Ω(x / (\log \log x)^3)$.
+-/
+@[category research solved, AMS 11]
+theorem erdos_946.variants.hildebrand_lower_bound : (fun x => x / (x.log.log)^3) =O[atTop] erdos946Count
+    := by
+  sorry
```
smmercuri:
```suggestion
theorem erdos_946.variants.hildebrand_lower_bound : 
    (fun x => x / (x.log.log)^3) =O[atTop] erdos946Count := by
  sorry
```

smmercuri:
LGTM after styling fixes!

FormalConjectures/ErdosProblems/946.lean
seewoo5: I added a line for that in docstring!

smmercuri: Great, thanks!

smmercuri (approved):

PR #666, opened 2025-09-14
Erdős Problem 470

rao107:
Adds [Erdős Problem 470](https://www.erdosproblems.com/470)

This PR also moves the definition of `primeGap` to its own separate file under `FormalConjectures/ForMathlib/NumberTheory/PrimeGap.lean` since this problem also uses the definition.

### DIFF
--- a/FormalConjectures/ErdosProblems/470.lean
+++ b/FormalConjectures/ErdosProblems/470.lean
@@ -45,7 +45,7 @@ theorem erdos_470.part1 : (∃ n : ℕ, n.Weird ∧ Odd n) ↔ answer(sorry) := 
 Are there infinitely many primitive weird numbers?
 -/
 @[category research open, AMS 11]
-theorem erdos_470.part2 : (Set.Infinite {n : ℕ | PrimitiveWeird n}) ↔ answer(sorry) := by
+theorem erdos_470.part2 : (Set.Infinite PrimitiveWeird) ↔ answer(sorry) := by
   sorry
 
 /--
--- a/FormalConjectures/ErdosProblems/470.lean
+++ b/FormalConjectures/ErdosProblems/470.lean
@@ -88,7 +88,7 @@ that an odd weird number must have at least 6 prime divisors.
 -/
 @[category research solved, AMS 11]
 theorem erdos_470.variants.odd_weird_prime_div :
-    ∀ n : ℕ, n.Weird → {m | m ∈ n.divisors ∧ m.Prime}.ncard > 6 := by
+    ∀ n : ℕ, Odd n → n.Weird → {m | m ∈ n.divisors ∧ m.Prime}.ncard > 6 := by
   sorry
 
 /--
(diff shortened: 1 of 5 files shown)


FormalConjectures/ErdosProblems/470.lean
```diff
@@ -0,0 +1,102 @@
...
+-/
+@[category research open, AMS 11]
+theorem erdos_470.part1 : (∃ n : ℕ, n.Weird ∧ Odd n) ↔ answer(sorry) := by
+  sorry
+
+/--
+Are there infinitely many primitive weird numbers?
+-/
+@[category research open, AMS 11]
+theorem erdos_470.part2 : (Set.Infinite {n : ℕ | PrimitiveWeird n}) ↔ answer(sorry) := by
```
mo271:
```suggestion
theorem erdos_470.part2 : (Set.Infinite PrimitiveWeird) ↔ answer(sorry) := by
```
simpler?

FormalConjectures/ErdosProblems/470.lean
```diff
@@ -0,0 +1,102 @@
...
+/--
+Melfi [Me15](https://mathscinet.ams.org/mathscinet/relay-station?mr=3276337) has proved that there
+are infinitely many primitive weird numbers, conditional on the fact that
+$p_{n+1} - p_n < \frac{1}{10} p_n^{1/2}$ for all large $n$, which in turn would follow from
+well-known conjectures concerning prime gaps.
+-/
+@[category research solved, AMS 11]
+theorem erdos_470.variants.prime_gap_imp_inf_prim_weird :
+    ∀ᶠ n in Filter.atTop, primeGap n < √ (n.nth Nat.Prime) / 10 →
+      Set.Infinite {n : ℕ | PrimitiveWeird n} := by
```
mo271:
```suggestion
      Set.Infinite PrimitiveWeird  := by
```

FormalConjectures/ErdosProblems/470.lean
```diff
@@ -0,0 +1,102 @@
...
+theorem erdos_470.variants.odd_weird_10_pow_21 : ∀ n < 10 ^ 21, Odd n → ¬n.Weird := by
+  sorry
+
+/--
+Liddy and Riedl [LiRi18](https://ideaexchange.uakron.edu/honors_research_projects/728/) have shown
+that an odd weird number must have at least 6 prime divisors.
+-/
+@[category research solved, AMS 11]
+theorem erdos_470.variants.odd_weird_prime_div :
+    ∀ n : ℕ, Odd n → n.Weird → {m | m ∈ n.divisors ∧ m.Prime}.ncard > 6 := by
```
mo271:
```suggestion
    ∀ n : ℕ, Odd n → n.Weird → 6 ≤ {m | m ∈ n.divisors ∧ m.Prime}.ncard := by
```

The order is cosmetic, but the important difference it that $6$ is not ruled out by the paper, only up to 5 it seems..

FormalConjectures/ErdosProblems/470.lean
rao107: simpler!

PR #670, opened 2025-09-18
Erdős Problem 312

ayush31010:
Summary

This PR introduces a scaffold for Erdős Problem 312 in the FormalConjectures/ErdosProblems directory.
The theorem is stated in terms of Fin n → ℕ families and subsets Finset (Fin n) rather than multisets, which makes multiplicities explicit and avoids ambiguity with ⊆.

Changes

New file: FormalConjectures/ErdosProblems/312.lean

Added helper definitions:

hsumIdx: harmonic sum of a subfamily restricted to indices in a Finset.

hsumAll: total harmonic sum of the full indexed family.

Proved simple supporting lemmas (hsumIdx_empty, hsumAll_eq_univ_sum).

Stated the main theorem erdos_312_fintype:

Existence of a universal constant c > 0.

For all K > 1, sufficiently large families with harmonic sum > K contain a subfamily with harmonic sum in (1 - exp(-c*K), 1].

Added docstrings and inline comments for clarity.

Included a proof sketch as guidance for future formalization (currently marked sorry).

Motivation

The fintype-indexed formulation is simpler to handle in Lean:

It avoids issues with sub-multisets and multiplicity tracking.Summary

This PR introduces a scaffold for Erdős Problem 312 in the FormalConjectures/ErdosProblems directory.
The theorem is stated in terms of Fin n → ℕ families and subsets Finset (Fin n) rather than multisets, which makes multiplicities explicit and avoids ambiguity with ⊆.

Changes

New file: FormalConjectures/ErdosProblems/312.lean

Added helper definitions:

hsumIdx: harmonic sum of a subfamily restricted to indices in a Finset.

hsumAll: total harmonic sum of the full indexed family.

Proved simple supporting lemmas (hsumIdx_empty, hsumAll_eq_univ_sum).

Stated the main theorem erdos_312_fintype:

Existence of a universal constant c > 0.

For all K > 1, sufficiently large families with harmonic sum > K contain a subfamily with harmonic sum in (1 - exp(-c*K), 1].

Added docstrings and inline comments for clarity.

Included a proof sketch as guidance for future formalization (currently marked sorry).

Motivation

The fintype-indexed formulation is simpler to handle in Lean:

It avoids issues with sub-multisets and multiplicity tracking.

It keeps the problem statement closer to the original mathematical intent.

Provides a clear scaffold for collaborators to build on and formalize the probabilistic method proof.

Next Steps

Replace sorry with a formal proof, likely via a probabilistic method (Chebyshev/Hoeffding).

[...]

PR #671, opened 2025-09-18
Update AUTHORS

Solventerritory:


mo271:
let's do these changes together with https://github.com/google-deepmind/formal-conjectures/pull/670

PR #672, opened 2025-09-18
Updated copyright

Solventerritory:


mo271:
let's do these changes together with #670

PR #673, opened 2025-09-21
Add formalization of Erdős Problem 188

### LINKED ISSUE #412: Erdős Problem 188
### What is the conjecture
    
https://www.erdosproblems.com/188

  What is the smallest $k$ such that $\mathbb{R}^2$ can be red/blue coloured with no pair of red points unit distance apart, and no $k$-term arithmetic progression of blue points with distance $1$?

Status: open

### Prerequisites needed

What we have for arithmetic progressions should work in `Set ℂ` out of the box, and likely also for some version of subsets of $\mathbb{R}^2$. 

### Choose either option

rdivyanshu:
Fix #412


FormalConjectures/ErdosProblems/188.lean
```diff
@@ -0,0 +1,43 @@
...
+/-!
+# Erdős Problem 188
+
+*Reference:* [erdosproblems.com/188](https://www.erdosproblems.com/188)
+-/
+
+namespace Erdos188
+
+/--
+The set of numbers ($k$) such that $\mathbb{R}^2$ can be red/blue coloured with no pair of red
```
Paul-Lez:
```suggestion
The set of numbers $k$ such that $\mathbb{R}^2$ can be red/blue coloured with no pair of red
```

FormalConjectures/ErdosProblems/188.lean
```diff
@@ -0,0 +1,43 @@
...
+points unit distance apart, and no $k$-term arithmetic progression of blue points with distance 1.
+-/
+def erdos_188.s := { k : ℕ | ∃ blue : Set ℂ,
```
Paul-Lez: Since you already have the `Erdos188` namespace in this file, we might as well remove this:)
```suggestion
def s := { k : ℕ | ∃ blue : Set ℂ,
```

FormalConjectures/ErdosProblems/188.lean
```diff
@@ -0,0 +1,43 @@
...
+    (Set.univ \ blue).Pairwise (fun c₁ c₂ => dist c₁ c₂ ≠ 1) ∧
+    ¬ (∃ bs ⊆ blue, (∃ s, bs.IsAPOfLengthWith k s 1)) }
+
+/--
+What is the smallest $k$ such that $\mathbb{R}^2$ can be red/blue coloured with no pair of red
+points unit distance apart, and no $k$-term arithmetic progression of blue points with distance 1?
+-/
+@[category research open, AMS 5]
+theorem erdos_188 : IsEmpty erdos_188.s ∨ IsLeast erdos_188.s answer(sorry)
+    := by sorry
```
Paul-Lez:
```suggestion
/--
What is the smallest $k$ such that $\mathbb{R}^2$ can be red/blue coloured with no pair of red
points unit distance apart, and no $k$-term arithmetic progression of blue points with distance 1?
-/
@[category research open, AMS 5]
theorem erdos_188 : IsEmpty erdos_188.s ∨ IsLeast erdos_188.s answer(sorry) := by
  sorry
```

FormalConjectures/ErdosProblems/188.lean
```diff
@@ -0,0 +1,43 @@
...
+def erdos_188.s := { k : ℕ | ∃ blue : Set ℂ,
+    (Set.univ \ blue).Pairwise (fun c₁ c₂ => dist c₁ c₂ ≠ 1) ∧
+    ¬ (∃ bs ⊆ blue, (∃ s, bs.IsAPOfLengthWith k s 1)) }
+
+/--
+What is the smallest $k$ such that $\mathbb{R}^2$ can be red/blue coloured with no pair of red
+points unit distance apart, and no $k$-term arithmetic progression of blue points with distance 1?
+-/
+@[category research open, AMS 5]
+theorem erdos_188 : IsEmpty erdos_188.s ∨ IsLeast erdos_188.s answer(sorry)
```
Paul-Lez: I'd be tempted to remove the `IsEmpty erdos_188.s` clause from this statement and instead add a separate statement `erdos_188.s.Nonempty` (since this seems to be known, based on [the webpage](https://www.erdosproblems.com/188)). 

It could also be nice to add the known stuff on this problem (e.g. that `k` is at least 5 and less than 10000000)

rdivyanshu: I have hard time parsing what website says vs what paper proves. I believe this is coming from following.

> Erdos et al. asked in [2, p. 535] whether the following proposition is valid: if the points of E^2 are 2-colored (red and blue) so that no two blue points are at distance 1 apart then there will be four red points forming a unit
square (for E^3, the analogous statement is obvious). In this paper we show, that the answer is affirmative even if we say “arbitrary four-point configuration” instead of “unit square” (Theorem 1). A supplementary result
(Theorem 2) shows that Theorem 1 becomes certainly false if we replace “four” by “twelve” in it. The interval 4 < n < 12 is no man’s land at present. 

This proves that coloring exists such that no two blue point 1 distance apart and four red points form AP sequence of distance 1 (Red and blue is exchanged in paper). But it doesn't prove or refute that there is coloring exists such that no two blue point 1 distance apart and no four red points form AP sequence of distance 1. Hence I kept IsEmpty initially.

Paul-Lez: I found this in the book by Erdős & Graham (_Old and new problems and results in combinatorial number theory: van der Waerden's theorem and related topics_):
> It has been shown that there is a large M so that it is possible to partition
> E2 into two sets A and B so that A contains no pair of points with distance 1
> and B contains no A.P. of length M.

so hopefully we're good (though they don't provide a proof nor a reference...)

rdivyanshu: Added, PTAL. I wanted to use [IsCoinitialFor](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/Bounds/Defs.html#IsCoinitialFor), but I think it is added recently.

PR #675, opened 2025-09-22
feat: add basic API for halting of TMs

Paul-Lez:


FormalConjectures/ForMathlib/Computability/TuringMachine/BusyBeavers.lean
```diff
@@ -174,6 +181,40 @@ lemma haltsAfter_zero_iff (s : Cfg Γ Λ) :
 lemma isHalting_iff_exists_haltsAt : IsHalting M ↔ ∃ n, M.HaltsAfter (init []) n :=
   ⟨fun _ ↦ eval_dom_iff.mpr IsHalting.halts, fun H ↦ ⟨eval_dom_iff.mp H⟩⟩
 
+lemma notHaltsAfter_iff (s : Cfg Γ Λ) (n : ℕ) (H : ¬M.HaltsAfter s n) :
```
callesonne: is this name accurate? I don't see an `iff`

Paul-Lez: Good point, fixed!

FormalConjectures/ForMathlib/Test/Computability/TuringMachine.lean
```diff
@@ -36,24 +37,36 @@ def alwaysHaltingMachine : Machine Γ Λ := fun _ _ =>
...
   · use { q := some Λ.T, tape := ⟨Γ.A, Quotient.mk'' [Γ.A], default⟩}
     rfl
   · rfl
+
+def neverHalts : Machine Γ Λ := fun l s =>
+  match l, s with
+  | .S, .A => some (Λ.T, Stmt.write default Dir.right)
+  | .T, .A => some (Λ.S, Stmt.write default Dir.right)
+  | .S, .B => some (Λ.T, Stmt.write default Dir.left)
+  | .T, .B => some (Λ.T, Stmt.write default Dir.left)
```
callesonne: I'm slightly confused why we are putting definitions in a test file? Is it because this is not going into `Mathlib` eventually?

FormalConjectures/ForMathlib/Test/Computability/TuringMachine.lean
```diff
@@ -36,24 +37,36 @@ def alwaysHaltingMachine : Machine Γ Λ := fun _ _ =>
...
+  exact ⟨1, by aesop⟩
 
 theorem haltsAfterOne_haltingNumber : haltsAfterOne.haltingNumber = 1 := by
   apply haltingNumber_def
   · use { q := some Λ.T, tape := ⟨Γ.A, Quotient.mk'' [Γ.A], default⟩}
     rfl
   · rfl
+
+def neverHalts : Machine Γ Λ := fun l s =>
+  match l, s with
```
YaelDillies:
```suggestion
def neverHalts : Machine Γ Λ :=
```

FormalConjectures/ForMathlib/Computability/TuringMachine/BusyBeavers.lean
```diff
@@ -174,6 +181,40 @@ lemma haltsAfter_zero_iff (s : Cfg Γ Λ) :
 lemma isHalting_iff_exists_haltsAt : IsHalting M ↔ ∃ n, M.HaltsAfter (init []) n :=
   ⟨fun _ ↦ eval_dom_iff.mpr IsHalting.halts, fun H ↦ ⟨eval_dom_iff.mp H⟩⟩
 
+lemma exists_of_notHaltsAfter (s : Cfg Γ Λ) (n : ℕ) (H : ¬M.HaltsAfter s n) :
```
YaelDillies:
```suggestion
lemma exists_of_not_haltsAfter (s : Cfg Γ Λ) (n : ℕ) (H : ¬M.HaltsAfter s n) :
```

FormalConjectures/ForMathlib/Computability/TuringMachine/BusyBeavers.lean
```diff
@@ -174,6 +181,40 @@ lemma haltsAfter_zero_iff (s : Cfg Γ Λ) :
...
+      suffices ∀ (a : Λ), a ∉ (Option.get _ HH).q by
+        rw [← Option.eq_none_iff_forall_not_mem.mpr this, ← Option.eq_some_of_isSome]
+      intro a h
+      apply (H a (Option.get _ HH).tape) <| h.symm ▸ (Option.eq_some_of_isSome HH)
+    rw [HaltsAfter, multiStep_succ, this]
+    rfl
+  · apply multiStep_succ_eq_none_of_multiStep_eq_none
+    rwa [Bool.not_eq_true, Option.not_isSome, Option.isNone_iff_eq_none] at HH
+
+lemma not_isHalting_iff_forall_multiStep_ne_none : ¬IsHalting M
```
YaelDillies:
```suggestion
lemma not_isHalting_iff_forall_isSome_multiStep : ¬IsHalting M
```

FormalConjectures/ForMathlib/Test/Computability/TuringMachine.lean
Paul-Lez: Yeah the idea here is just having a few examples to check the definitions make sense.

Paul-Lez:
> This is mainly proofs so looks good to me, but I didn't review the original PR adding these files so can I ask: is the option in `def step (M : Machine Γ Λ) : Cfg Γ Λ → Option (Cfg Γ Λ)` really necessary? Couldn't you send `⟨none, t⟩` to `⟨none, t⟩` again? Instead of sending it to `none`. This is so that it could take values in `Cfg Γ Λ` which I feel would make some case work later easier. But I am not super familiar with Turing machines.

For this, the `none`s are encoding two different things: a `none` in the output of `step (M : Machine Γ Λ) : Cfg Γ Λ → Option (Cfg Γ Λ)` encodes the fact that the `state, symbol` pair in the input configuration had no transition defined, while a `none` in the state corresponds to the machine reaching a special state called the halting state (which has no transitions defined to other states, so the latter implies the former, but these are not the same since a machine could have several states that may lead to halting; and these lead to different total step counts of the machine).

FormalConjectures/ForMathlib/Computability/TuringMachine/BusyBeavers.lean
```diff
@@ -174,6 +181,40 @@ lemma haltsAfter_zero_iff (s : Cfg Γ Λ) :
...
+        rw [← Option.eq_none_iff_forall_not_mem.mpr this, ← Option.eq_some_of_isSome]
+      intro a h
+      apply (H a (Option.get _ HH).tape) <| h.symm ▸ (Option.eq_some_of_isSome HH)
+    rw [HaltsAfter, multiStep_succ, this]
+    rfl
+  · apply multiStep_succ_eq_none_of_multiStep_eq_none
+    rwa [Bool.not_eq_true, Option.isSome_eq_false_iff, Option.isNone_iff_eq_none] at HH
+
+lemma not_isHalting_iff_forall_isSome_multiStep : ¬IsHalting M
+    ↔ ∀ n, M.multiStep (init []) (n + 1) |>.isSome := by
```
YaelDillies: For readability, I recommend not starting a multiline statement in the middle of a line
```suggestion
lemma not_isHalting_iff_forall_isSome_multiStep :
    ¬ IsHalting M ↔ ∀ n, M.multiStep (init []) (n + 1) |>.isSome := by
```

FormalConjectures/ForMathlib/Computability/TuringMachine/BusyBeavers.lean
```diff
@@ -132,6 +132,13 @@ lemma multiStep_succ (M : Machine Γ Λ) (config : Cfg Γ Λ) (n : ℕ) :
     M.multiStep config (n + 1) = Option.bind (M.multiStep config n) M.step := by
   rw [multiStep, Function.iterate_succ', Function.comp_apply, multiStep]
 
+@[simp]
+lemma multiStep_succ_eq_none_of_multiStep_eq_none {M : Machine Γ Λ} {config : Cfg Γ Λ} {n : ℕ}
+    (H : M.multiStep config n = none) :
+    M.multiStep config (n + 1) = none := by
```
YaelDillies: Maybe generalise to
```suggestion
lemma multiStep_eq_none_of_le {M : Machine Γ Λ} {config : Cfg Γ Λ} {m n : ℕ}
    (H : M.multiStep config n = none) (hnm : n ≤ m) :
    M.multiStep config m = none := by
```

FormalConjectures/ForMathlib/Computability/TuringMachine/BusyBeavers.lean
```diff
@@ -174,6 +181,40 @@ lemma haltsAfter_zero_iff (s : Cfg Γ Λ) :
 lemma isHalting_iff_exists_haltsAt : IsHalting M ↔ ∃ n, M.HaltsAfter (init []) n :=
   ⟨fun _ ↦ eval_dom_iff.mpr IsHalting.halts, fun H ↦ ⟨eval_dom_iff.mp H⟩⟩
 
+lemma exists_of_not_haltsAfter (s : Cfg Γ Λ) (n : ℕ) (H : ¬M.HaltsAfter s n) :
+    ∃ (a : Λ) (b : Tape Γ), M.multiStep s n = some ⟨a, b⟩ := by
+  contrapose! H
+  by_cases HH : M.multiStep s n |>.isSome
```
YaelDillies: Is it easier to do a match here possibly?
```suggestion
  match HH : M.multiStep s n with
```

YaelDillies (approved):
Looks good to me otherwise

FormalConjectures/ForMathlib/Computability/TuringMachine/BusyBeavers.lean
```diff
@@ -132,6 +132,13 @@ lemma multiStep_succ (M : Machine Γ Λ) (config : Cfg Γ Λ) (n : ℕ) :
     M.multiStep config (n + 1) = Option.bind (M.multiStep config n) M.step := by
   rw [multiStep, Function.iterate_succ', Function.comp_apply, multiStep]
 
+@[simp]
+lemma multiStep_eq_none_of_le {M : Machine Γ Λ} {config : Cfg Γ Λ} {m n : ℕ}
```
YaelDillies:
```suggestion
lemma multiStep_eq_none_mono {M : Machine Γ Λ} {config : Cfg Γ Λ} {m n : ℕ}
```
is more inspiring maybe?

FormalConjectures/ForMathlib/Test/Computability/TuringMachine.lean
```diff
@@ -38,11 +39,11 @@ def alwaysHaltingMachine : Machine Γ Λ := fun _ _ =>
   none
 
 def haltsAfterOne : Machine Γ Λ := fun l _ =>
```
YaelDillies: No need to `fun` then `match`:
```suggestion
def haltsAfterOne : Machine Γ Λ := fun
```

FormalConjectures/ForMathlib/Test/Computability/TuringMachine.lean
```diff
@@ -38,11 +39,11 @@ def alwaysHaltingMachine : Machine Γ Λ := fun _ _ =>
...
-match l with
-| --If the state is `S`, change state to `T` and move head to the right
-  .S => some (Λ.T, Stmt.write default Dir.right)
-| --If the state is already `T` then halt
-  .T => none
+  match l with
+  | --If the state is `S`, change state to `T` and move head to the right
+    .S => some (Λ.T, Stmt.write default Dir.right)
+  | --If the state is already `T` then halt
+    .T => none
```
YaelDillies:
```suggestion
  | -- If the state is `S`, change state to `T` and move head to the right
    .S, _ => some (Λ.T, Stmt.write default Dir.right)
  | -- If the state is already `T` then halt
    .T, _ => none
```

FormalConjectures/ForMathlib/Test/Computability/TuringMachine.lean
Paul-Lez: But I like having `fun`!

Issue #646, opened 2025-09-02
Seventeen square packing problem


BoltonBailey:
### What is the conjecture

The square packing problem [covered by wikipedia here](https://en.wikipedia.org/wiki/Square_packing). The case of n=17 is on particular interest due to the complicated form of the smallest solution found so far. I propose adding an open problem asking what the minimal size of the 17-packing is. I could also go further and add some of the other related questions linked on the wikipedia page about optimal packings involving circles.

### Prerequisites needed

Should not require many prerequisites, squares can be represented as images of the unit square under isometries.

### [AMS categories](https://github.com/google-deepmind/formal-conjectures/labels?q=ams-)

* ams-51 Geometry

ayushdebnath012:
Can you plz assign me this isuue

Paul-Lez:
This issue has already been claimed by @BoltonBailey

BoltonBailey:
Sorry for sleeping on this, #681 is a PR.

BoltonBailey:
Closed by #681

PR #681, opened 2025-09-24
feat(SquarePacking): add formalization of square packing theorems

BoltonBailey:
This PR adds some theorems on square packing described in #646. In particular it has a problem about the minimum size of a square in which 17 unit squares can be packed.

### DIFF
--- a/FormalConjectures/Wikipedia/SquarePacking.lean
+++ b/FormalConjectures/Wikipedia/SquarePacking.lean
@@ -1,0 +1,94 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Square Packing
+
+*References:*
+- [Wikipedia on packing of squares in squares](https://en.wikipedia.org/wiki/Square_packing#In_a_square)
+- Friedman, Erich (2009), "Packing unit squares in squares: a survey and new results",
+  Electronic Journal of Combinatorics, 1000, Dynamic Survey 7
+- A website with visualizations of packings:
+  [link](https://erich-friedman.github.io/packing/)
+-/
+
+open EuclideanGeometry
+
+universe u
+
+namespace SquarePacking
+
+/--
+A square of a particular side length as a subset of the Euclidean plane.
+-/
+def Square (side : ℝ) : Set ℝ² :=
+  {p : ℝ² | 0 < p 0 ∧ p 0 < side ∧ 0 < p 1 ∧ p 1 < side}
+
+/--
+The unit square as a subset of the Euclidean plane
+(not including border, so that squares that touch at the border are disjoint).
+-/
+def UnitSquare : Set ℝ² := Square 1
+
+/--
+A structure representing a packing of `n` isometric embeddings
+of a set `s` inside a (presumably larger) set `S`.
+-/
+structure Packing (n : ℕ) (s : Set ℝ²) (S : Set ℝ²) where
+  /-- The isometric equivalences
+  that represent the transformations of the base shape to their locations in the packing. -/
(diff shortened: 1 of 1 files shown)


BoltonBailey:
This fixes #646

BoltonBailey:
Apparently it is claimed that the best known solution for squares in circles [is not proved optimal for n=3](https://erich-friedman.github.io/packing/squincir/). This problem seems like it can't possibly be that hard to solve, so I am a bit hesitant to include it. Nevertheless, I added it, I'll leave it up to the reviewers if this or other variants of the problem should stay.

FormalConjectures/Wikipedia/SquarePacking.lean
```diff
@@ -0,0 +1,185 @@
...
+
+/--
+The unit square as a subset of the Euclidean plane.
+-/
+def UnitSquare : Set ℝ² := Square 1
+
+/--
+A circle of a particular radius as a subset of the Euclidean plane.
+Not including border, so that squares that touch at the border are disjoint,
+but a square internal to another shape is a subset of that shape.
```
mo271:
```suggestion
Not including border, so that circles that touch at the border are disjoint,
but a circle internal to another shape is a subset of that shape.
```

FormalConjectures/Wikipedia/SquarePacking.lean
```diff
@@ -0,0 +1,185 @@
...
+/--
+A structure representing a packing of `n` isometric embeddings
+of a set `s` inside a (presumably larger) set `S`.
+-/
+structure Packing (n : ℕ) (s : Set ℝ²) (S : Set ℝ²) where
+  /-- The isometric equivalences
+  that represent the transformations of the base shape to their locations in the packing. -/
+  embeddings : Fin n → (ℝ² ≃ᵢ ℝ²)
+  /-- The images of the embeddings are pairwise disjoint -/
+  disjoint : ∀ i j : Fin n, i ≠ j → Disjoint (embeddings i '' s) (embeddings j '' s)
```
mo271: Might it make sense to use `Pairwise` here?

FormalConjectures/Wikipedia/SquarePacking.lean
```diff
@@ -0,0 +1,185 @@
...
+  sorry
+
+/--
+What is the smallest square that can contain 11 unit squares?
+
+Reference: https://en.wikipedia.org/wiki/Square_packing#In_a_square
+-/
+@[category research open, AMS 51]
+theorem least_eleven_square_packing_in_square :
+    IsLeast {x : ℝ | Nonempty (Packing 11 UnitSquare (Square x))} answer(sorry) := by
```
mo271, line 98: while it is probably true that the answer here will be an algebraic number, I think it might be a good idea to add an easier version here, finding a better than the known bound, but not perfect?
That would mean looking at 
https://link.springer.com/article/10.1007/s00454-004-1129-z
finding the most known digits of this constant and then setting up a theorem finding a number that is better. The idea here being, even if the constant described in that paper is optimal a "new" result would be more digits of that constant, or a strictly smaller one when it is not optimal. 
(But this is just and idea and there is no need to add it in this pull request.)

mo271: same remark for some of the other results here.

mo271 (approved):

FormalConjectures/Wikipedia/SquarePacking.lean
BoltonBailey: I think all the numbers which are solutions to any of the problems here are definitely algebraic ([here](https://math.stackexchange.com/questions/2153357/transcendental-solutions-to-constrained-polynomial-optimization-problems) is a stackexchange answer). I have been assuming that the way these numbers were derived was something along the lines of:

1. Obtaining some algebraic constraints on the locations of the shapes
2. Using a CAS to derive a particular algebraic representation for the number in question 
3. Computing it to a few digits of precision 

(indeed, some of the numbers given for some of these problems are expressed exactly, I assume the only reason not all of them are is that the polynomials become too big or unsolvable by radicals). 

This being the case, I feel that computing a few more digits would be hard to justify as being "new" or an "open problem", because in principle, the same author who came up with the numbers in the first place could just run their program again for a bit longer to get additional precision. The real open problem is to prove that the configuration in question is optimal (of course, this is decidable too in principle, but is a much harder technical feat). And the closed problems I am providing just as a stand in to the demonstrations of the particular constructions.

I could also think about changing these to conjectures, by saying something like 
`IsLeast {x : ℝ | Nonempty (Packing 11 UnitSquare (Square x))} ... <-> answer(sorry)`, but I am reluctant to do this because, again, I think it would take quite a lot of text to fill in the ... as a precisely-stated algebraic number, and even if some of these constructions turn out not to be optimal, the problem remains interesting.

mo271:
Thanks!

Issue #271, opened 2025-06-30
Erdős Problem 51: Extremal property of Euler's totient


mo271:
### What is the conjecture
    
https://www.erdosproblems.com/51

  Is there an infinite set $A\subset \mathbb{N}$ such that for every $a\in A$ there is an integer $n$ such that $\phi(n)=a$, and yet if $n_a$ is the smallest such integer then $n_a/a\to \infty$ as $a\to\infty$?

Status: open

### Prerequisites needed
Euler's totient function is there and also filters to express the limit needed here. 

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

cosmicgenius:
Hi! I'd like to take on this one.

PR #682, opened 2025-09-25
Erdős Problem 51

cosmicgenius:
Fix https://github.com/google-deepmind/formal-conjectures/issues/271

### DIFF
--- a/FormalConjectures/ErdosProblems/51.lean
+++ b/FormalConjectures/ErdosProblems/51.lean
@@ -1,0 +1,46 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 51
+
+*Reference:* [erdosproblems.com/51](https://www.erdosproblems.com/51)
+-/
+
+namespace Erdos51
+
+/--
+Is there an infinite set A ⊆ ℕ such that for every a ∈ A,
+there is an integer n such that ϕ(n)=a, and
+yet if nₐ is the smallest such integer, then nₐ/a → ∞ as a → ∞?
+-/
+@[category research open, AMS 11]
+theorem erdos_51 :
+    ∃ A : Set ℕ, ∃ n : A → ℕ,
+      A.Infinite ∧
+      ∀ a : A, IsLeast (Nat.totient ⁻¹' {(a : ℕ)}) (n a) ∧
+      Filter.Tendsto (fun a : A => (n a : ℝ) / (a : ℝ)) Filter.atTop Filter.atTop
+    ↔ answer(sorry) := by
+  sorry
+
+/-
+The remarks from the erdosproblems site are the same as those in
+[erdosproblems.com/694](https://www.erdosproblems.com/694).
+-/
+
+end Erdos51


FormalConjectures/ErdosProblems/51.lean
```diff
@@ -0,0 +1,46 @@
...
+    ∃ A : Set ℕ, ∃ n : A → ℕ,
+      A.Infinite ∧
+      ∀ a : A, IsLeast (Nat.totient ⁻¹' {(a : ℕ)}) (n a) ∧
+      Filter.Tendsto (fun a : A => (n a : ℝ) / (a : ℝ)) Filter.atTop Filter.atTop
+    ↔ answer(sorry) := by
+  sorry
+
+/-
+The remarks from the erdosproblems site are the same as those in
+[erdosproblems.com/694](https://www.erdosproblems.com/694).
```
cosmicgenius, line 46: For this reason, I didn't add any of the extra statements there. Should we still add them anyways? Is this comment necessary?

mo271: Yeah, I think it is fine not to repeat them, and good to mention it in a comment!

FormalConjectures/ErdosProblems/51.lean
```diff
@@ -0,0 +1,46 @@
...
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 51
+
+*Reference:* [erdosproblems.com/51](https://www.erdosproblems.com/51)
+-/
+
```
mo271, line 24:
```suggestion

open Filter
open scoped Nat

```
This way we don't need to write `Filter` below. 
This then allows to use `φ` instead of `Nat.totient` below.

FormalConjectures/ErdosProblems/51.lean
```diff
@@ -0,0 +1,46 @@
...
+Is there an infinite set A ⊆ ℕ such that for every a ∈ A,
+there is an integer n such that ϕ(n)=a, and
+yet if nₐ is the smallest such integer, then nₐ/a → ∞ as a → ∞?
+-/
+@[category research open, AMS 11]
+theorem erdos_51 :
+    ∃ A : Set ℕ, ∃ n : A → ℕ,
+      A.Infinite ∧
+      ∀ a : A, IsLeast (Nat.totient ⁻¹' {(a : ℕ)}) (n a) ∧
+      Filter.Tendsto (fun a : A => (n a : ℝ) / (a : ℝ)) Filter.atTop Filter.atTop
```
mo271:
```suggestion
      (∀ a : A, IsLeast (φ ⁻¹' {(a : ℕ)}) (n a)) ∧
      Tendsto (fun a : A => (n a : ℝ) / (a : ℝ)) atTop atTop
```
note also the parentheses around the `∀ a`, which makes for a more natural statement.

FormalConjectures/ErdosProblems/51.lean
```diff
@@ -0,0 +1,46 @@
...
+
+*Reference:* [erdosproblems.com/51](https://www.erdosproblems.com/51)
+-/
+
+namespace Erdos51
+
+/--
+Is there an infinite set A ⊆ ℕ such that for every a ∈ A,
+there is an integer n such that ϕ(n)=a, and
+yet if nₐ is the smallest such integer, then nₐ/a → ∞ as a → ∞?
```
mo271:
```suggestion
Is there an infinite set $A ⊆ ℕ$ such that for every $a ∈ A$,
there is an integer n such that $\phi(n)=a$, and
yet if $n_a$ is the smallest such integer, then $\frac{n_a}{a} → ∞$ as $a → ∞$?
```

Let's use more LaTex markdown

FormalConjectures/ErdosProblems/51.lean
mo271: I think we need parenthesis around everything before the `↔ answer(sorry)`.

mo271 (commented):
Thanks, looks good!

FormalConjectures/ErdosProblems/51.lean
```diff
@@ -0,0 +1,49 @@
...
+there is an integer n such that $\phi(n)=a$, and
+yet if $n_a$ is the smallest such integer, then $\frac{n_a}{a} → \infty$ as $a → ∞$?
+-/
+@[category research open, AMS 11]
+theorem erdos_51 :
+    ∃ A : Set ℕ, ∃ n : A → ℕ,
+      (A.Infinite ∧
+      (∀ a : A, IsLeast (φ ⁻¹' {(a : ℕ)}) (n a)) ∧
+      Tendsto (fun a : A => (n a : ℝ) / (a : ℝ)) atTop atTop)
+    ↔ answer(sorry) := by
```
mo271, line 41:
```suggestion
theorem erdos_51 :
    (∃ A : Set ℕ, ∃ n : A → ℕ,
      A.Infinite ∧
      (∀ a : A, IsLeast (φ ⁻¹' {(a : ℕ)}) (n a)) ∧
      Tendsto (fun a : A => (n a : ℝ) / (a : ℝ)) atTop atTop)
    ↔ answer(sorry) := by
```

mo271 (approved):

PR #685, opened 2025-09-25
fix: Fix definition of completeness in erdos problem 348

Paul-Lez:


mo271 (approved):
I suppose one could make multiplicative versions of these and use `to_additive`, but perhaps let's now worry about that until there is a real world use case...

PR #691, opened 2025-09-26
Erdos 975

seewoo5:
There might be a better way to deal with nonnegative irreducible polynomials, and feel free to give a comment on it.

### DIFF
--- a/FormalConjectures/ErdosProblems/975.lean
+++ b/FormalConjectures/ErdosProblems/975.lean
@@ -1,0 +1,62 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+
+/-!
+# Erdős Problem 975
+
+*References:*
+ - [erdosproblems.com/975](https://www.erdosproblems.com/975)
+ - [Va39] van der Corput, J. G., Une in\'egalit\'e{} relative au nombre des diviseurs. Nederl. Akad. Wetensch., Proc. (1939), 547--553.
+ - [Er52b] Erd\"os, P., On the sum {$\sum^x_{k=1} d(f(k))$}. J. London Math. Soc. (1952), 7--15.
+-/
+
+open Filter Real Polynomial
+
+namespace Erdos975
+
+/- Sum of $\tau(f(n))$ for a function $f : \mathbb{N} \to \mathbb{N}$. -/
+noncomputable def erdos975Sum (f : ℕ → ℕ) (x : ℝ) : ℝ :=
+  ∑ n ∈ Finset.range (⌊x⌋₊ + 1), (f n).divisors.card
+
+/--
+For an irreducible polynomial $f \in \mathbb{Z}[x]$ with $f(n) \ge 0$ for all $n \ge 0$,
+does there exists a constant $c = c(f) > 0$ such that
+$\sum_{n \le x} \tau(f(n)) \approx c \cdot x \log x$?
+-/
+@[category research open, AMS 11]
+theorem erdos_975 (f : ℕ → ℕ) (hf : ∃ g : ℤ[X], Irreducible g ∧ ∀ n : ℕ, f n = g.eval ↑n) :
+    ∃ (c : ℝ), c > 0 ∧ Tendsto (fun x ↦ erdos975Sum f x / (x * x.log)) (atTop) (nhds c) := by
+  sorry
+
+/--
+The correctness of growth rate is shown in [Va39] (lower bound) and [Er52b] (upper bound).
+-/
+@[category research solved, AMS 11]
+theorem erdos_975.variant.upper_bound (f : ℕ → ℕ)
+    (hf : ∃ g : ℤ[X], Irreducible g ∧ ∀ n : ℕ, f n = g.eval ↑n) :
+    erdos975Sum f =O[atTop] (fun x ↦ x * x.log) := by
+  sorry
+
+@[category research solved, AMS 11]
+theorem erdos_975.variant.lower_bound (f : ℕ → ℕ)
+    (hf : ∃ g : ℤ[X], Irreducible g ∧ ∀ n : ℕ, f n = g.eval ↑n) :
+    (fun x ↦ x * x.log) =O[atTop] (erdos975Sum f) := by
+  sorry
+
+end Erdos975
--- a/FormalConjectures/ErdosProblems/975.lean
+++ b/FormalConjectures/ErdosProblems/975.lean
@@ -24,15 +24,31 @@ import FormalConjectures.Util.ProblemImports
  - [erdosproblems.com/975](https://www.erdosproblems.com/975)
  - [Va39] van der Corput, J. G., Une in\'egalit\'e{} relative au nombre des diviseurs. Nederl. Akad. Wetensch., Proc. (1939), 547--553.
  - [Er52b] Erd\"os, P., On the sum {$\sum^x_{k=1} d(f(k))$}. J. London Math. Soc. (1952), 7--15.
+ - [Ho63] Hooley, Christopher, On the number of divisors of a quadratic polynomial. Acta Math. (1963), 97--114.
+ - [Mc95] McKee, James, On the average number of divisors of quadratic polynomials. Math. Proc. Cambridge Philos. Soc. (1995), 389--392.
+ - [Mc97] McKee, James, A note on the number of divisors of quadratic polynomials. (1997), 275--281.
+ - [Mc99] McKee, James, The average number of divisors of an irreducible quadratic polynomial. Math. Proc. Cambridge Philos. Soc. (1999), 17--22.
+ - [T] T. Tao, Erdos' divisor bound, https://terrytao.wordpress.com/2011/07/23/erdos-divisor-bound/
 -/
 
 open Filter Real Polynomial
+open scoped ArithmeticFunction
 
 namespace Erdos975
 
-/- Sum of $\tau(f(n))$ for a function $f : \mathbb{N} \to \mathbb{N}$. -/
-noncomputable def erdos975Sum (f : ℕ → ℕ) (x : ℝ) : ℝ :=
-  ∑ n ∈ Finset.range (⌊x⌋₊ + 1), (f n).divisors.card
+/-
+Sum of $\tau(f(n))$ for a function $f : \mathbb{N} \to \mathbb{N}$.
+Here $\tau$ is the divisor counting function, which is `σ 0` in mathlib.
+-/
+noncomputable def Erdos975Sum (f : ℕ → ℕ) (x : ℝ) : ℝ :=
+  ∑ n ∈ Finset.range (⌊x⌋₊ + 1), σ 0 (f n)
+
+/- Auxiliary definition for nonnegative irreducible polynomials over $\mathbb{Z}$ on $\mathbb{N}$. -/
+def Erdos975NonnegIrredPoly (f : ℕ → ℕ) (g : ℤ[X]) : Prop :=
+  Irreducible g ∧ ∃ N : ℕ, ∀ n ≥ N, f n = g.eval ↑n ∧ 0 ≤ f n
+
+def Erdos975Asymptotic (f : ℕ → ℕ) (c : ℝ) : Prop :=
+  Tendsto (fun x ↦ Erdos975Sum f x / (x * x.log)) (atTop) (nhds c)
 
 /--
 For an irreducible polynomial $f \in \mathbb{Z}[x]$ with $f(n) \ge 0$ for all $n \ge 0$,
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,103 @@
...
+open scoped ArithmeticFunction
+
+namespace Erdos975
+
+/-
+Sum of $\tau(f(n))$ for a function $f : \mathbb{N} \to \mathbb{N}$.
+Here $\tau$ is the divisor counting function, which is `σ 0` in mathlib.
+-/
+noncomputable def Erdos975Sum (f : ℕ → ℕ) (x : ℝ) : ℝ :=
+  ∑ n ∈ Finset.range (⌊x⌋₊ + 1), σ 0 (f n)
```
Paul-Lez: I think this should work:
```suggestion
  ∑ n ≤ ⌊x⌋₊, σ 0 (f n)
```

FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,103 @@
...
+
+/- Auxiliary definition for nonnegative irreducible polynomials over $\mathbb{Z}$ on $\mathbb{N}$. -/
+def Erdos975NonnegIrredPoly (f : ℕ → ℕ) (g : ℤ[X]) : Prop :=
+  Irreducible g ∧ ∃ N : ℕ, ∀ n ≥ N, f n = g.eval ↑n ∧ 0 ≤ f n
+
+def Erdos975Asymptotic (f : ℕ → ℕ) (c : ℝ) : Prop :=
+  Tendsto (fun x ↦ Erdos975Sum f x / (x * x.log)) (atTop) (nhds c)
```
Paul-Lez:
```suggestion
  Tendsto (fun x ↦ Erdos975Sum f x / (x * x.log)) atTop (nhds c)
```

FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,103 @@
...
+
+/--
+For an irreducible polynomial $f \in \mathbb{Z}[x]$ with $f(n) \ge 0$ for all $n \ge 0$,
```
Paul-Lez: The Erdős Problem website seems to work with different assumptions on `f` (namely `f` being at least `1` for sufficiently large `n`). Would you mind adding a comment saying from which source this formulation is coming from?

FormalConjectures/ErdosProblems/975.lean
seewoo5: Thanks! I think there are other Erdos files using `Finset.range` summation, such as 946 (which I did write following other Erdos file that I forget which one). I can make another PR for such minor updates later.

FormalConjectures/ErdosProblems/975.lean
seewoo5: The comment actually follows order statement of the website. I think the new one is better, so I'll update docstring accordingly.

FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,103 @@
...
+ - [T] T. Tao, Erdos' divisor bound, https://terrytao.wordpress.com/2011/07/23/erdos-divisor-bound/
+-/
+
+open Filter Real Polynomial
+open scoped ArithmeticFunction
+
+namespace Erdos975
+
+/-
+Sum of $\tau(f(n))$ for a function $f : \mathbb{N} \to \mathbb{N}$.
```
Paul-Lez:
```suggestion
Sum of $\tau(f(n))$ from `0` to `⌊x⌋` to for a function $f : \mathbb{N} \to \mathbb{N}$.
```

FormalConjectures/ErdosProblems/975.lean
Paul-Lez: That sounds great, thanks!

FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,103 @@
...
+def Erdos975Asymptotic (f : ℕ → ℕ) (c : ℝ) : Prop :=
+  Tendsto (fun x ↦ Erdos975Sum f x / (x * x.log)) atTop (nhds c)
+
+/--
+For an irreducible polynomial $f \in \mathbb{Z}[x]$ with $f(n) \ge 1$ for sufficiently large $n$,
+does there exists a constant $c = c(f) > 0$ such that
+$\sum_{n \le x} \tau(f(n)) \approx c \cdot x \log x$?
+-/
+@[category research open, AMS 11]
+theorem erdos_975 (f : ℕ → ℕ) (g : ℤ[X]) (hf : Erdos975NonnegIrredPoly f g) :
```
Paul-Lez: Here since the problem is stated as a question, I think it would be better to use the `answer(sorry)` convention.

seewoo5: Updated!

FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,103 @@
...
+/-
+Sum of $\tau(f(n))$ from `0` to `⌊x⌋` to for a function $f : \mathbb{N} \to \mathbb{N}$.
+Here $\tau$ is the divisor counting function, which is `σ 0` in mathlib.
+-/
+noncomputable def Erdos975Sum (f : ℕ → ℕ) (x : ℝ) : ℝ :=
+  ∑ n ≤ ⌊x⌋₊, σ 0 (f n)
+
+/- Auxiliary definition for nonnegative irreducible polynomials over $\mathbb{Z}$ on $\mathbb{N}$. -/
+def Erdos975NonnegIrredPoly (f : ℕ → ℕ) (g : ℤ[X]) : Prop :=
+  Irreducible g ∧ ∃ N : ℕ, ∀ n ≥ N, f n = g.eval ↑n ∧ 0 ≤ f n
```
Paul-Lez: I would be tempted to remove this condition, and just do everything in terms of the polynomial `g` (and `g.eval` when you want to use the corresponding function!)

FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,103 @@
...
+
+/--
+For an irreducible polynomial $f \in \mathbb{Z}[x]$ with $f(n) \ge 1$ for sufficiently large $n$,
+does there exists a constant $c = c(f) > 0$ such that
+$\sum_{n \le x} \tau(f(n)) \approx c \cdot x \log x$?
+-/
+@[category research open, AMS 11]
+theorem erdos_975 (f : ℕ → ℕ) (g : ℤ[X]) (hf : Erdos975NonnegIrredPoly f g) :
+    (∃ (c : ℝ), 0 < c ∧ Erdos975Asymptotic f c) ↔ answer(sorry) := by
+  sorry
```
Paul-Lez, line 57: Here, I would recommend moving the quantification "inside". That way, replacing `sorry` by false will correspond to the negation of the statement.

FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,103 @@
...
+When $f$ is an irreducible quadratic polynomial, the question is answered first by Hooley [Ho63].
+More compact expression of the constant in terms of Hurwitz class numbers (when $a = 1$)
+is given by McKey in [Mc95], [Mc97], [Mc99].
+
+TODO: formalize Hurwitz class numbers and the expression of the constant in terms of them.
+-/
+@[category research solved, AMS 11]
+theorem erdos_975.variant.quadratic (f : ℕ → ℕ) (g : ℤ[X]) (hf : Erdos975NonnegIrredPoly f g)
+    (hg_degree : g.degree = 2) :
+    ∃ (c : ℝ), 0 < c ∧ Erdos975Asymptotic f c := by
```
Paul-Lez: Here you could possibly use `answer(sorry)` as a placeholder for `c` (and then later replace `sorry` by the actual closed form!)

FormalConjectures/ErdosProblems/975.lean
seewoo5: Do you have any suggestions? One issue is that `σ 0` is only defined on `ℕ`, so I cannot just use polynomials over integers (and I also want to consider something like $x^2 - 2$ since the result is known for all quadratic integral polynomials. One possible way is to extend the divisor sum function to `ℤ` by 0, which is equivalent to the original statement. Or maybe I misunderstood your comment!

YaelDillies: I would extend the divisor sum function to the integers by using `Int.toNat` to turn an integer `n` into the natural number `max n 0`

FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,103 @@
...
+-/
+noncomputable def Erdos975Sum (f : ℕ → ℕ) (x : ℝ) : ℝ :=
+  ∑ n ≤ ⌊x⌋₊, σ 0 (f n)
+
+/- Auxiliary definition for nonnegative irreducible polynomials over $\mathbb{Z}$ on $\mathbb{N}$. -/
+def Erdos975NonnegIrredPoly (f : ℕ → ℕ) (g : ℤ[X]) : Prop :=
+  Irreducible g ∧ ∃ N : ℕ, ∀ n ≥ N, f n = g.eval ↑n ∧ 0 ≤ f n
+
+def Erdos975Asymptotic (f : ℕ → ℕ) (c : ℝ) : Prop :=
+  Tendsto (fun x ↦ Erdos975Sum f x / (x * x.log)) atTop (nhds c)
```
YaelDillies: Could you drop all these definitions? I feel like they don't help understanding the statements, but instead obscure them

FormalConjectures/ErdosProblems/975.lean
seewoo5: I'm not sure if I understand your suggestion - could you elaborate?

FormalConjectures/ErdosProblems/975.lean
seewoo5: I updated, although I'm unsure if this is what you meant.

FormalConjectures/ErdosProblems/975.lean
seewoo5: Removed! I agree that it made things more complicated.

seewoo5:
@Paul-Lez @YaelDillies Sorry for dragging this PR, I addressed some of your comments.

FormalConjectures/ErdosProblems/975.lean
YaelDillies:
```suggestion
/--
For an irreducible polynomial $f \in \mathbb{Z}[x]$ with $f(n) \ge 1$ for sufficiently large $n$,
does there exists a constant $c = c(f) > 0$ such that
$\sum_{n \le x} \tau(f(n)) \approx c \cdot x \log x$?
-/
@[category research open, AMS 11]
theorem erdos_975 : answer(sorry) ↔
    \forall (f : ℕ → ℕ) (g : ℤ[X]), Erdos975NonnegIrredPoly f g → ∃ c > (0 : ℝ), Erdos975Asymptotic f c := by
  sorry
```
is roughly what Paul means

FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,101 @@
...
+When $f$ is an irreducible quadratic polynomial, the question is answered first by Hooley [Ho63].
+More compact expression of the constant in terms of Hurwitz class numbers (when $a = 1$)
+is given by McKey in [Mc95], [Mc97], [Mc99].
+
+TODO: formalize Hurwitz class numbers and the expression of the constant in terms of them.
+-/
+@[category research solved, AMS 11]
+theorem erdos_975.variant.quadratic (f : ℤ[X]) (hf : Irreducible f)
+    (hf_pos : ∀ᶠ n : ℕ in atTop, 1 ≤ f.eval ↑n) (hf_degree : f.degree = 2) (c : ℝ) :
+    c = answer(sorry) ∧ 0 < c ∧ Tendsto (fun x ↦ Erdos975Sum f x / (x * x.log)) atTop (nhds c) := by
```
YaelDillies: This is unprovable. I think you meant
```suggestion
    (hf_pos : ∀ᶠ n : ℕ in atTop, 1 ≤ f.eval ↑n) (hf_degree : f.degree = 2) (c : ℝ) :
    c = answer(sorry) → 0 < c ∧ Tendsto (fun x ↦ Erdos975Sum f x / (x * x.log)) atTop (nhds c) := by
```
instead?

FormalConjectures/ErdosProblems/975.lean
seewoo5: Thanks! I updated.

FormalConjectures/ErdosProblems/975.lean
seewoo5: Yes you're right, I updated.

FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,100 @@
...
+  ∑ n ≤ ⌊x⌋₊, σ 0 (f.eval ↑n).toNat
+
+/--
+For an irreducible polynomial $f \in \mathbb{Z}[x]$ with $f(n) \ge 1$ for sufficiently large $n$,
+does there exists a constant $c = c(f) > 0$ such that
+$\sum_{n \le x} \tau(f(n)) \approx c \cdot x \log x$?
+-/
+@[category research open, AMS 11]
+theorem erdos_975 : (∀ (f : ℤ[X]), Irreducible f → (∀ᶠ n : ℕ in atTop, 1 ≤ f.eval ↑n) →
+    (∃ c > (0 : ℝ), Tendsto (fun x ↦ Erdos975Sum f x / (x * x.log)) atTop (nhds c))) ↔
```
YaelDillies:
```suggestion
    ∃ c > (0 : ℝ), Tendsto (fun x ↦ Erdos975Sum f x / (x * log x)) atTop (nhds c)) ↔
```

FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,100 @@
...
+    answer(sorry) := by
+  sorry
+
+/--
+The correctness of growth rate is shown in [Va39] (lower bound) and [Er52b] (upper bound).
+-/
+@[category research solved, AMS 11]
+theorem erdos_975.variant.upper_bound (f : ℤ[X]) (hf : Irreducible f)
+    (hf_pos : ∀ᶠ n : ℕ in atTop, 1 ≤ f.eval ↑n) : Erdos975Sum f =O[atTop] (fun x ↦ x * x.log) := by
```
YaelDillies:
```suggestion
    (hf_pos : ∀ᶠ n : ℕ in atTop, 1 ≤ f.eval ↑n) : Erdos975Sum f =O[atTop] (fun x ↦ x * log x) := by
```
Same below

FormalConjectures/ErdosProblems/975.lean
```diff
@@ -0,0 +1,100 @@
...
+$\sum_{n \le x} \tau(f(n)) \approx c \cdot x \log x$?
+-/
+@[category research open, AMS 11]
+theorem erdos_975 : (∀ (f : ℤ[X]), Irreducible f → (∀ᶠ n : ℕ in atTop, 1 ≤ f.eval ↑n) →
+    (∃ c > (0 : ℝ), Tendsto (fun x ↦ Erdos975Sum f x / (x * x.log)) atTop (nhds c))) ↔
+    answer(sorry) := by
+  sorry
+
+/--
+The correctness of growth rate is shown in [Va39] (lower bound) and [Er52b] (upper bound).
```
YaelDillies:
```suggestion
The correctness of the growth rate is shown in [Va39] (lower bound) and [Er52b] (upper bound).
```
maybe?

PR #692, opened 2025-09-28
Erdős Problem 422

rao107:
Adds [Erdős Problem 422](https://www.erdosproblems.com/422).

### DIFF
--- a/FormalConjectures/ErdosProblems/422.lean
+++ b/FormalConjectures/ErdosProblems/422.lean
@@ -1,0 +1,46 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 422
+
+*Reference:* [erdosproblems.com/422](https://www.erdosproblems.com/422)
+-/
+
+namespace Erdos422
+
+/--
+Let $f(0) = f(1) = 1$ and for $n > 1$
+$$
+f(n) = f(n - f(n - 1)) + f(n - f(n - 2)).
+$$
+Note: It is not known whether $f(n)$ is well-defined for all $n$.
+-/
+partial def f : ℕ → ℕ
+| 0 => 1
+| 1 => 1
+| n => f (n - f (n - 1)) + f (n - f (n - 2))
+
+/--
+Does $f(n)$ miss infinitely many integers?
+-/
+@[category research open, AMS 11]
+theorem erdos_422 : Set.Infinite {n | ∀ x, f x ≠ n} := by
+  sorry
+
+end Erdos422
--- a/FormalConjectures/ErdosProblems/422.lean
+++ b/FormalConjectures/ErdosProblems/422.lean
@@ -25,7 +25,7 @@ import FormalConjectures.Util.ProblemImports
 namespace Erdos422
 
 /--
-Let $f(0) = f(1) = 1$ and for $n > 1$
+Let $f(1) = f(2) = 1$ and for $n > 2$
 $$
 f(n) = f(n - f(n - 1)) + f(n - f(n - 2)).
 $$
--- a/FormalConjectures/ErdosProblems/422.lean
+++ b/FormalConjectures/ErdosProblems/422.lean
@@ -31,9 +31,9 @@ f(n) = f(n - f(n - 1)) + f(n - f(n - 2)).
 $$
 Note: It is not known whether $f(n)$ is well-defined for all $n$.
 -/
-partial def f : ℕ → ℕ
-| 0 => 1
+partial def f : ℕ+ → ℕ+
 | 1 => 1
+| 2 => 1
 | n => f (n - f (n - 1)) + f (n - f (n - 2))
 
 /--


FormalConjectures/ErdosProblems/422.lean
```diff
@@ -0,0 +1,46 @@
...
+Let $f(1) = f(2) = 1$ and for $n > 2$
+$$
+f(n) = f(n - f(n - 1)) + f(n - f(n - 2)).
+$$
+Note: It is not known whether $f(n)$ is well-defined for all $n$.
+-/
+partial def f : ℕ+ → ℕ+
+| 1 => 1
+| 2 => 1
+| n => f (n - f (n - 1)) + f (n - f (n - 2))
```
Paul-Lez:
```suggestion
  | 1 => 1
  | 2 => 1
  | n => f (n - f (n - 1)) + f (n - f (n - 2))
```

FormalConjectures/ErdosProblems/422.lean
```diff
@@ -0,0 +1,46 @@
...
+
```
Paul-Lez, line 42: Here I think it would be interesting to add two statements:
1) The question of the surjectivity of `f` (seems to be open)
2) The question of whether the sequence `f` must eventually terminate (presumably not? admittedly I haven't thought too much about whether or not this is hard to prove)

rao107: Those sound like good ideas. I'll also add in a question of the function's growth rate since that's open as well.

rao107: Hey, I added your suggestions but I had a bit of trouble with the second.

For it, I removed the `partial` definition for `f` and defined the function with `sorry`. I don't think this is what you had in mind when you suggested it so I'd welcome any other ways to define that `f` must eventually terminate.

smmercuri: Perhaps by termination @Paul-Lez meant whether `f` becomes stationary at some point? I.e., does there exist `m` and `N` for which `f n = m` for all `n` larger than `N`. Termination can be quite surprising in this sense, see for example the [Goodstein sequence](https://oeis.org/A266201) which eventually decreases and terminates at 0.

The other interpretation of termination here is whether one reaches a point at which `f n` cannot be defined (e.g, if `n - f (n - 1) < 0`), but I think this sense is already captured in the definition.

rao107: Apologies for the delay! I've added another variant for `f` eventually becoming stationary. :slightly_smiling_face:

FormalConjectures/ErdosProblems/422.lean
```diff
@@ -0,0 +1,73 @@
...
+-/
+@[category research open, AMS 11]
+theorem erdos_422.variants.growth_rate : f =O[atTop] (answer(sorry) : ℕ+ → ℕ+) := by
+  sorry
+
+/--
+Does $f$ become stationary at some point?
+-/
+@[category research open, AMS 11]
+theorem erdos_422.variants.eventually_const : (∃ n, EventuallyConst f n) ↔ answer(sorry) := by
```
callesonne:
```suggestion
theorem erdos_422.variants.eventually_const : EventuallyConst f atTop ↔ answer(sorry) := by
```
This does not look quite right to me, as you are asking for the existence of a filter `n`. I'm guessing this is what you want?

FormalConjectures/ErdosProblems/422.lean
```diff
@@ -0,0 +1,73 @@
...
+Let $f(1) = f(2) = 1$ and for $n > 2$
+$$
+f(n) = f(n - f(n - 1)) + f(n - f(n - 2)).
+$$
+-/
+partial def f : ℕ+ → ℕ+
+  | 1 => 1
+  | 2 => 1
+  | n => f (n - f (n - 1)) + f (n - f (n - 2))
+-- Note: It is not known whether $f(n)$ is well-defined for all $n$.
```
callesonne:
```suggestion
$$

Note: It is not known whether $f(n)$ is well-defined for all $n$.
-/
partial def f : ℕ+ → ℕ+
  | 1 => 1
  | 2 => 1
  | n => f (n - f (n - 1)) + f (n - f (n - 2))
```
Moved the comment into the docstring.

FormalConjectures/ErdosProblems/422.lean
```diff
@@ -0,0 +1,73 @@
...
+*Reference:* [erdosproblems.com/422](https://www.erdosproblems.com/422)
+-/
+
+namespace Erdos422
+
+open Filter
+open scoped Topology
+
+instance : Norm ℕ+ where
+  norm n := n
```
callesonne: I'm not sure if you want to add this as a global instance. For example there seems to be no `Norm` instance on `ℕ` in mathlib (that I can find). I believe the reason for this is that you don't want it as an instance as a priori there are multiple possible norms (such as the p-adic norms as well). I'm assuming you are adding this in order to talk about the growth rate below, but what you can do is just to view `f` as a function valued in R for that statement. See my suggestion below.

FormalConjectures/ErdosProblems/422.lean
```diff
@@ -0,0 +1,73 @@
...
+-/
+@[category research open, AMS 11]
+theorem erdos_422.variants.surjective : f.Surjective ↔ answer(sorry) := by
+  sorry
+
+/--
+How does $f$ grow?
+-/
+@[category research open, AMS 11]
+theorem erdos_422.variants.growth_rate : f =O[atTop] (answer(sorry) : ℕ+ → ℕ+) := by
```
callesonne:
```suggestion
theorem erdos_422.variants.growth_rate :
    (fun n ↦ (f n : ℝ)) =O[atTop] (answer(sorry) : ℕ+ → ℝ) := by
```

FormalConjectures/ErdosProblems/422.lean
callesonne:
```suggestion
```
I'm not sure if you want to add this as a global instance. For example there seems to be no `Norm` instance on `ℕ` in mathlib (that I can find). I believe the reason for this is that you don't want it as an instance as a priori there are multiple possible norms (such as the p-adic norms as well). I'm assuming you are adding this in order to talk about the growth rate below, but what you can do is just to view `f` as a function valued in R for that statement. See my suggestion below.

rao107: Oh, I didn't realize I could do that! Thank you so much!

PR #900, opened 2025-10-08
Erdős Problem 686

### LINKED ISSUE #879: Erdős Problem 686
### What is the conjecture
    
https://www.erdosproblems.com/686

  Can every integer $N\geq 2$ be written as
$$N=\frac{\prod_{1\leq i\leq k}(m+i)}{\prod_{1\leq i\leq k}(n+i)}$$
for some $k\geq 2$ and $m\geq n+k$?
  

Status:  open

### Choose either option

sboukortt:
Fixes #879


FormalConjectures/ErdosProblems/686.lean
```diff
@@ -0,0 +1,43 @@
...
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 686
+*Reference:* [erdosproblems.com/686](https://www.erdosproblems.com/686)
+-/
+
+namespace Erdos686
+
+open BigOperators
```
Paul-Lez: I think this isn't needed?
```suggestion
```

sboukortt: Oh, right, I think we added it for Π but indeed, it seems that it’s unnecessary. Thanks!

PR #903, opened 2025-10-08
Remove occurences of `private` keyword

Paul-Lez:


Paul-Lez:
> I assume the reasoning is now that we have everything in namespaces, we can drop the `private`?

Yeah exactly! I had started to do that in previous PRs but it seems that I missed a few.

PR #1120, opened 2025-10-15
feat: state P vs NP

### LINKED ISSUE #676: P != NP
### What is the conjecture

Generally agreed to be the most important open problem in computer science. The conjecture is that the class of problems that can be solved in polynomial time is not the same as the class of problems with answers that can be checked in polynomial time.

[Wikipedia Reference](https://en.wikipedia.org/wiki/P_versus_NP_problem)

### Prerequisites needed

Mathlib already has the concept of a function that can be computed in polynomial time: [Turing.TM2ComputableInPolyTime](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/TMComputable.html#Turing.TM2ComputableInPolyTime)

As well as the concept of a language over an alphabet: [Language](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/Language.html#Language)

What remains for prerequisites are to formalize the concept of the classes P and NP. Perhaps the simplest way of doing this (and closest to how introductory courses present the material) would be as `Set`s of `Language`s. Note that care must be taken to avoid making statements that are vacuously true or false due to trivial or infinite alphabets.

* ams-68 Computer Science

### Choose either option
- [X] I plan on adding this conjecture to the repository

BoltonBailey:
Closes #676: The P vs NP problem.

I have tried to keep the definitions as simple as possible to avoid misstatements. This PR defines:

- Decision problems as functions from `List`s of `Bool`s to `Bool`
- Complexity classes as sets of decision problems
- `P` as the class of decision problems computable with `TM2ComputableInPolyTime` using a generic encoding for `List Bool`
- `NP` using an adaptation of the definition in Arora-Barak.

As a bonus, this includes a definition of co-NP and the NP vs co-NP conjecture.

### DIFF
--- a/FormalConjectures/Wikipedia/ComplexityTheory.lean
+++ b/FormalConjectures/Wikipedia/ComplexityTheory.lean
@@ -101,6 +101,9 @@ theorem NP_ne_coNP : NP ≠ coNP := by sorry
 
 /--
 The theorem that the set of complements of languages in P is itself P.
+
+This can be proven by observing that the boolean negation function is computable in polynomial time,
+and that compositions of poly-time computable functions are also poly-time computable.
 -/
 @[category undergraduate, AMS 68]
 theorem coP_eq_P :
@@ -109,6 +112,10 @@ theorem coP_eq_P :
 
 /--
 The theorem that P is a subset of NP.
+
+This can be proven by observing that for any language in P,
+we can construct a verifier that ignores the witness and simply runs the poly-time decider for the
+language.
 -/
 @[category undergraduate, AMS 68]
 theorem P_subset_NP :


BoltonBailey:
Also, be advised, I took some of the code around the encodings of lists from an old Zulip thread discussing this problem, I am not sure if the person who wrote that has gone through the CLA process - should I try to get permission from them or rewrite that part of the code?

mo271:
> Also, be advised, I took some of the code around the encodings of lists from an old Zulip thread discussing this problem, I am not sure if the person who wrote that has gone through the CLA process - should I try to get permission from them or rewrite that part of the code?

If you think they contributed substantially, perhaps it would be good to add them as co-author by amending one of the commits with something like `Co-authored-by: Foo Bar <foobar@baz.org>`. This will make the CLA bot pick it up, I suppose. But of course this is a question of your judgment on how substantial their contribution was. 

Definitely nice that you link the zulip discussion in the code!

BoltonBailey:
I was able to replace Daniel Weber's encodings with a more compact implementation, perhaps we can revisit it if we want a more general implementation in the future, but for now it should be fine.

FormalConjectures/ForMathlib/Computability/Encoding.lean
```diff
@@ -0,0 +1,75 @@
...
+/-
+These encodings are used in the formalization of complexity classes such as P and NP.
+
+Note that in a Zulip discussion thread, Daniel Weber contributed some more general encodings.
+We might eventually want to replace these with his more general versions.
+
+see: https://leanprover.zulipchat.com/#narrow/channel/116395-maths/topic/Formalise.20the.20proposition.20P.20.E2.89.A0NP/with/453031558
+-/
+
+def finEncodingListBool : Computability.FinEncoding (List Bool) where
```
madvorak, line 32: I just started reading and I cannot help but think that the same definition could be written nicer (sorry for not getting any further in my reading). Something like:
```
  Γ := Bool
  encode := id
  decode := Option.some
  decode_encode _ := rfl
  ΓFin := Bool.fintype
```

BoltonBailey: Yes, in fact I think it was this way when it was simply Daniel's version, I'll change it back.

FormalConjectures/Wikipedia/ComplexityTheory.lean
```diff
@@ -0,0 +1,120 @@
...
+
+/--
+The type of complexity classes. We define these as sets of decision problems.
+-/
+abbrev ComplexityClass := Set DecisionProblem
+
+/--
+A simple definition to abstract the notion of a poly-time Turing machine into a predicate.
+-/
+def isComputableInPolyTime {α β : Type} (ea : FinEncoding α) (eb : FinEncoding β) (f : α → β) :=
```
YaelDillies:
```suggestion
def IsComputableInPolyTime {α β : Type} (ea : FinEncoding α) (eb : FinEncoding β) (f : α → β) :=
```
maybe?

FormalConjectures/Wikipedia/ComplexityTheory.lean
```diff
@@ -0,0 +1,120 @@
...
+  toFun : List Bool → Bool
+
+instance : CoeFun DecisionProblem (fun _ ↦ List Bool → Bool) :=
+  ⟨DecisionProblem.toFun⟩
+
+instance : Membership (List Bool) DecisionProblem :=
+  ⟨fun X x ↦ X x⟩
+
+instance : HasCompl DecisionProblem :=
+  ⟨fun X ↦ ⟨fun x ↦ not (X x)⟩⟩
```
YaelDillies: Why not
```suggestion
abbrev DecisionProblem := List Bool → Bool
```
? (I don't understand the point of this membership instance

FormalConjectures/Wikipedia/ComplexityTheory.lean
```diff
@@ -0,0 +1,120 @@
...
+@[category research open, AMS 68]
+theorem NP_ne_coNP : NP ≠ coNP := by sorry
+
+/--
+The theorem that P is a subset of NP.
+-/
+@[category undergraduate, AMS 68]
+theorem P_subset_NP :
+    P ⊆ NP := by
+  sorry
```
YaelDillies, line 123: Maybe also state that P is self-dual and `P ⊆ coNP`?

BoltonBailey: Good idea!

BoltonBailey:
Thanks for the suggestions @YaelDillies, I have implemented them. Let me know if there are any more comments.

PR #1123, opened 2025-10-15
Formalization of the open problems about Euclid numbers.

### LINKED ISSUE #1122: Euclid Numbers
### What is the conjecture
Conjectures about Euclid numbers:
https://en.wikipedia.org/wiki/Euclid_number

-  It is not known whether there is an infinite number of prime Euclid numbers ([primorial primes](https://en.wikipedia.org/wiki/Primorial_prime)).[[3]](https://en.wikipedia.org/wiki/Euclid_number#cite_note-3) 

- It is also unknown whether every Euclid number is a [squarefree number](https://en.wikipedia.org/wiki/Squarefree_number).

### Choose either option

szabadka:
fixes #1122

### DIFF
--- a/FormalConjectures/ErdosProblems/1065.lean
+++ b/FormalConjectures/ErdosProblems/1065.lean
@@ -1,0 +1,50 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/Euclid.lean
```diff
@@ -0,0 +1,46 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# *Euclic Numbers conjecture
```
mo271:
```suggestion
# Euclid Numbers conjecture
```

Issue #1074, opened 2025-10-14
Erdős Problem 1017


mo271:
### What is the conjecture

https://www.erdosproblems.com/1017

  Let $f(n,k)$ be such that every graph on $n$ vertices and $k$ edges can be partitioned into at most $f(k)$ edge-disjoint complete graphs. Estimate $f(n,k)$ for $k>n^2/4$.

Status:  open

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

cyb3r17:
This issue can be closed? [PR has been merged](https://github.com/google-deepmind/formal-conjectures/pull/1125)

smmercuri:
> This issue can be closed? [PR has been merged](https://github.com/google-deepmind/formal-conjectures/pull/1125)

That PR contains a different problem number (no. 1074). I'm not sure why GitHub is saying above that that PR references this issue since the PR description only references a different issue [#1114](https://github.com/google-deepmind/formal-conjectures/issues/1114)

cyb3r17:
I think there's been a mistake on my side, I have confused Issue 1074 with Erdos 1074, this is embarrassing, sorry for the inconvenience 😓

PR #1125, opened 2025-10-15
Erdős problem #1074

### LINKED ISSUE #1114: Erdős Problem 1074
### What is the conjecture

https://www.erdosproblems.com/1074

  Let $S$ be the set of all $m\geq 1$ such that there exists a prime $p\not\equiv 1\pmod{m}$ such that $m!+1\equiv 0\pmod{p}$. Does
$$\lim \frac{\lvert S\cap [1,x]\rvert}{x}$$
exist? What is it?

Similarly, if $P$ is the set of all primes $p$ such that there exists an $m$ with $p\not\equiv 1\pmod{m}$ such that $m!+1\equiv 0\pmod{p}$, then does
$$\lim \frac{\lvert P\cap [1,x]\rvert}{\pi(x)}$$
exist? What is it?

Status:  open

### Choose either option

smmercuri:
Fixes #1114

### DIFF
--- a/FormalConjectures/ErdosProblems/1074.lean
+++ b/FormalConjectures/ErdosProblems/1074.lean
@@ -26,7 +26,7 @@ open scoped Nat
 
 /-- The EHS numbers (after Erdős, Hardy, and Subbarao) are those $m\geq 1$ such that there
 exists a prime $p\not\equiv 1\pmod{m}$ such that $m! + 1 \equiv 0\pmod{p}$.-/
-abbrev Nat.EHSNumbers : Set ℕ := {m | 1 ≤ m ∧ ∃ p, ¬p ≡ 1 [MOD m] ∧ p ∣ m ! + 1}
+abbrev Nat.EHSNumbers : Set ℕ := {m | 1 ≤ m ∧ ∃ p, p.Prime ∧ ¬p ≡ 1 [MOD m] ∧ p ∣ m ! + 1}
 
 /-- The Pillai primes are those primes $p$ such that there exists an $m$ with
 $p\not\equiv 1\pmod{m}$ such that $m! + 1 \equiv 0\pmod{p}$-/
@@ -41,7 +41,7 @@ that $m! + 1 \equiv 0\pmod{p}$. Does
 $$
   \lim\frac{|S\cap[1, x]|}{x}
 $$
-exist? What is it? -/
+exist? -/
 @[category research open, AMS 11]
 theorem erdos_1074.parts.i.i : (∃ c, EHSNumbers.HasDensity c) ↔ answer(sorry) := by
   sorry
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/1074.lean
```diff
@@ -0,0 +1,94 @@
...
+# Erdős Problem 1074
+
+*Reference:* [erdosproblems.com/1074](https://www.erdosproblems.com/1074)
+-/
+
+open scoped Nat
+
+/-- The EHS numbers (after Erdős, Hardy, and Subbarao) are those $m\geq 1$ such that there
+exists a prime $p\not\equiv 1\pmod{m}$ such that $m! + 1 \equiv 0\pmod{p}$.-/
+abbrev Nat.EHSNumbers : Set ℕ := { m | 1 ≤ m ∧ ∃ p, ¬ p ≡ 1 [MOD m] ∧ p ∣ m ! + 1}
```
Paul-Lez:
```suggestion
abbrev Nat.EHSNumbers : Set ℕ := {m | 1 ≤ m ∧ ∃ p, ¬p ≡ 1 [MOD m] ∧ p ∣ m ! + 1}
```

Paul-Lez: Should the first two definitions go in `ForMathlib`?

FormalConjectures/ErdosProblems/1074.lean
```diff
@@ -0,0 +1,94 @@
...
+
+/-- Let $S$ be the set of all $m\geq 1$ such that there exists a prime $p\not\equiv 1\pmod{m}$ such
+that $m! + 1 \equiv 0\pmod{p}$. Does
+$$
+  \lim\frac{|S\cap[1, x]|}{x}
+$$
+exist? What is it? -/
+@[category research open, AMS 11]
+theorem erdos_1074.parts.i : (∃ c, EHSNumbers.HasDensity c) ↔ answer(sorry) := by
+  sorry
```
Paul-Lez, line 56: Arguably there should also be a version of this statement that allows the solver to provide the value of the limit using `answer(sorry)`

smmercuri: Something like this you mean?
```
∃ c, c = answer(sorry) ∧ EHSNumbers.HasDensity c 
```

FormalConjectures/ErdosProblems/1074.lean
smmercuri: Pillai primes seem to be a proper thing but I'm not sure about EHS numbers. It doesn't look like Pillai primes appear in any other Erdos problem so I'm inclined to leave it here for now, but I don't have strong feelings either way

FormalConjectures/ErdosProblems/1074.lean
callesonne: Does `EHSNumbers.HasDensity answer(sorry)` not work?

FormalConjectures/ErdosProblems/1074.lean
callesonne: Should there not be a condition that `p` is prime here?

smmercuri: yes there definitely should!

PR #1128, opened 2025-10-17
feat: Add erdos 276

### LINKED ISSUE #466: Erdős Problem 276
### What is the conjecture
    
https://www.erdosproblems.com/276

  Is there an infinite Lucas sequence $a_0,a_1,\ldots,$ where $(a_0,a_1)=1$ and $a_{n+2}=a_{n+1}+a_n$ for $n\geq 0$ such that all $a_k$ are composite, and yet no integer has a common factor with every term of the sequence?

Status: open

### Choose either option

callesonne:
Adds Erdos 276 as well as `Nat.Composite` to ForMathlib. Fixes #466.

### DIFF
--- a/FormalConjectures/ErdosProblems/276.lean
+++ b/FormalConjectures/ErdosProblems/276.lean
@@ -32,6 +32,6 @@ term of the sequence?
 -/
 @[category research open, AMS 11]
 theorem erdos_274 : (∃ (a : ℕ → ℕ),
-    IsLucasSequence a ∧ (∀ k, (a k).Composite) ∧ (∀ n > 1, ∃ k, ¬ n ∣ (a k))) ↔
+    IsLucasSequence a ∧ (∀ k, (a k).Composite) ∧ (∀ n > 1, ∃ k, Nat.gcd n (a k) = 1)) ↔
     answer(sorry) := by
   sorry
--- a/FormalConjectures/ErdosProblems/276.lean
+++ b/FormalConjectures/ErdosProblems/276.lean
@@ -23,7 +23,17 @@ import FormalConjectures.Util.ProblemImports
 [erdosproblems.com/276](https://www.erdosproblems.com/276)
 -/
 
-def IsLucasSequence (a : ℕ → ℕ) : Prop := ∀ n, a (n + 2) = a (n + 1) + a n
+/--
+We define a Lucas sequence to be a Fibonacci sequence with arbitrary starting points
+`L 0` and `L 1`.
+
+TODO: There seems to be multiple definitions in the literature, some of which also
+allow coefficients in the reccurence relation. For now this simple definition has been
+chosen as it agrees best with the Erdős problem in this same file.
+However before moving this into `ForMathlib` one should make a concious decision about
+which definition to choose.
+-/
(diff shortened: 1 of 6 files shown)


FormalConjectures/ErdosProblems/541.lean
smmercuri, line 1: What's happening in this file? It seems to be replacing 541 with 1077?

FormalConjectures/ErdosProblems/276.lean
```diff
@@ -0,0 +1,37 @@
...
+
+def IsLucasSequence (a : ℕ → ℕ) : Prop := ∀ n, a (n + 2) = a (n + 1) + a n
+
+/--
+Is there an infinite Lucas sequence $a_0, a_1, \ldots$ where $a_{n+2} = a_{n+1} + a_n$ for
+$n \ge 0$ such that all $a_k$ are composite, and yet no integer has a common factor with every
+term of the sequence?
+-/
+@[category research open, AMS 11]
+theorem erdos_274 : (∃ (a : ℕ → ℕ),
```
smmercuri:
```suggestion
theorem erdos_276 : (∃ (a : ℕ → ℕ),
```

FormalConjectures/ErdosProblems/541.lean
callesonne: Oh jesus christ yeah that isnt supposed to go there!

FormalConjectures/ErdosProblems/276.lean
```diff
@@ -0,0 +1,51 @@
...
+We define a Lucas sequence to be a Fibonacci sequence with arbitrary starting points
+`L 0` and `L 1`.
+
+TODO: There seems to be multiple definitions in the literature, some of which also
+allow coefficients in the reccurence relation. For now this simple definition has been
+chosen as it agrees best with the Erdős problem in this same file.
+However before moving this into `ForMathlib` one should make a concious decision about
+which definition to choose.
+-/
+def IsLucasSequence (L : ℕ → ℕ) : Prop := ∀ n, L (n + 2) = L (n + 1) + L n
```
Paul-Lez, line 36: I think this has already been defined in `ForMathlib` in some form [here](https://github.com/google-deepmind/formal-conjectures/blob/b212d70d9e681cd3545a1e6fb18c83e0eaccaef1/FormalConjectures/ForMathlib/NumberTheory/WallSunSunPrimes.lean). Would this work in your context?

callesonne: Ah nice catch, but unfortunately neither work. I think my definition should be expressable as some sort of linear combination of the two ones defined in that file, but to me this seems to overcomplicate things more than is needed for this problem. What do you think?

Paul-Lez: Yeah I think that would be a little too complicated. Let's just leave this as is for now!

Paul-Lez (approved):

FormalConjectures/ErdosProblems/276.lean
smmercuri: I think the one you need is the third one from that file, the sequence `lucasNumber`

PR #1131, opened 2025-10-19
Add Erdős Problem 371

### LINKED ISSUE #49: Formalise Erdos Problem 371

edwag:
fixes #49

### DIFF
--- a/FormalConjectures/ErdosProblems/503.lean
+++ b/FormalConjectures/ErdosProblems/503.lean
@@ -1,0 +1,84 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 503
+
+*Reference:* [erdosproblems.com/503](https://www.erdosproblems.com/503)
+-/
+
+namespace Erdos503
+
+open scoped EuclideanGeometry
+
+/--
+What is the size of the largest $A \subseteq \mathbb{R}^n$ such that every three points from $A$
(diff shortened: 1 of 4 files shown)


FormalConjectures/ErdosProblems/371.lean
```diff
@@ -0,0 +1,36 @@
...
+
+/--
+Let $P(n)$ denote the largest prime factor of $n$. Show that the set of $n$
+with $P(n+1) > P(n)$ has density 1/2.
+-/
+@[category research open, AMS 11]
+theorem erdos_371 :
+    { n | Nat.maxPrimeFac (n + 1) > Nat.maxPrimeFac n }.HasDensity (1/2) := by
+  sorry
+
```
mo271, line 35:
```suggestion

-- TODO: add the statements from the additional material
```

mo271:
Thanks for the contribution! The formalization looks good to me!

FormalConjectures/ErdosProblems/371.lean
edwag: I believe it is open. As I understand the comments on erdosproblems.com, two related things have been proved:

1. the analogous statement for logarithmic density, as opposed to natural density, holds;
2. this statement would follow from a conjecture which is itself still open;

but the statement itself remains open.

Which additional statements do you think we should add here? I could add (1) as part of this PR, but adding (2) would require formalising the "Elliott-Halberstam conjecture for friable integers", which I'm afraid is beyond me and feels out of scope for #49 . We could open an issue to track that? (It looks like "The Elliott-Halberstam Conjecture" is already referenced in a couple of open issues, but I think this is not the same thing?).

I could also have a go at adding the question about the density of $n$ such that $P(n+1) > P(n) n^\alpha$, as part of this PR.

mo271: Ok, great regarding the openness of the statement!

I think only adding a TODO would be enough for this pr. For later, the two statements I see are:
- This set and its complement both have positive upper density. (two theorems, solved)
- the statment with $\alpha$ (as pointed out by you)

Not sure about adding the other stuff you mentioned, at least not in this file...

mo271 (approved):
Thanks!

FormalConjectures/ErdosProblems/371.lean
```diff
@@ -0,0 +1,37 @@
...
+# Erdős Problem 371
+
+*Reference:* [erdosproblems.com/371](https://www.erdosproblems.com/371)
+-/
+
+namespace Erdos371
+
+/--
+Let $P(n)$ denote the largest prime factor of $n$. Show that the set of $n$
+with $P(n+1) > P(n)$ has density 1/2.
```
mo271:
```suggestion
with $P(n+1) > P(n)$ has density $\frac{1}{2}$.
```
one more nit

PR #1135, opened 2025-10-21
feat(ErdosProblems): 1072

### LINKED ISSUE #1112: Erdős Problem 1072
### What is the conjecture

https://www.erdosproblems.com/1072

  For any prime $p$, let $f(p)$ be the least integer such that $f(p)!+1\equiv 0\pmod{p}$.

Is it true that there are infinitely many $p$ for which $f(p)=p-1$?
Is it true that $f(p)/p\to 0$ for almost all $p$?

Status:  open

### Choose either option

callesonne:
Fixes #1112

### DIFF
--- a/FormalConjectures/ErdosProblems/1072.lean
+++ b/FormalConjectures/ErdosProblems/1072.lean
@@ -23,6 +23,7 @@ import FormalConjectures.Util.ProblemImports
 -/
 
 open Nat Filter Finset
+open scoped Topology
 
 namespace Erdos1072
 


FormalConjectures/ErdosProblems/1072.lean
```diff
@@ -0,0 +1,58 @@
...
+
+*Reference:* [erdosproblems.com/1065](https://www.erdosproblems.com/1072)
+-/
+
+open Nat Filter Finset
+
+namespace Erdos1072
+
+/-- For any prime $p$, let $f(p)$ be the least integer such that $f(p)! + 1 \equiv 0 \mod p$.-/
+noncomputable def f (p : ℕ) : ℕ := sInf {n | (n)! + 1 ≡ 0 [MOD p]}
```
Paul-Lez, line 31: Are the brackets around the `n` really needed here?

FormalConjectures/ErdosProblems/1072.lean
```diff
@@ -0,0 +1,58 @@
...
+@[category research open, AMS 11]
+theorem erdos_1072a : Set.Infinite {p | p.Prime ∧ f p = p - 1} ↔ answer(sorry) := by
+  sorry
+
+/-- Is it true that $f(p)/p \to 0$ for almost all $p$? -/
+@[category research open, AMS 11]
+theorem erdos_1072b :
+    (∃ (P : Finset ℕ), (∀ p ∈ P, p.Prime) ∧
+      ∀ᵉ ε > (0 : ℝ), ∃ N₀, ∀ p ≥ N₀, p.Prime ∧ p ∉ P → f p / p < ε)
+    ↔ answer(sorry) := by
```
Paul-Lez: (not an expert in this sort of stuff). I feel like one could also read this statement in terms of the density of the set of primes that verify the property, which would then be a little weaker than the current version. Do you think this is worth including as a separate statement?

FormalConjectures/ErdosProblems/1072.lean
```diff
@@ -0,0 +1,58 @@
...
+/--
+Erdős, Hardy, and Subbarao [HaSu02], believed that the number of $p \le x$ for which $f(p)=p−1$
+is $o(x/logx)$.
+
+[HaSu02] Hardy, G. E. and Subbarao, M. V., _A modified problem of Pillai and some related questions._
+Amer. Math. Monthly (2002), 554--559.
+-/
+@[category research open, AMS 11]
+theorem erdos_1072a.variants.littleo :
+    (fun x ↦ (#({p | p.Prime ∧ f p = p - 1} ∩ Finset.range (x + 1)).toFinset : ℝ)) =o[atTop]
```
Paul-Lez: I'd be tempted to do this for consistency with some of the other formalisations.
```suggestion
    (fun x ↦ (({p | p.Prime ∧ f p = p - 1} ∩ Set.interIcc 0 x).ncard : ℝ)) =o[atTop]
```

FormalConjectures/ErdosProblems/1072.lean
callesonne: Either brackets or a space between `n` and `!`. I tend to find brackets easier to read than the space.

FormalConjectures/ErdosProblems/1072.lean
smmercuri: Should that be `{p | p.Prime ∧ f p = p - 1}.interIcc 0 x).ncard` rather than `∩ Set.interIcc 0 x`?

Paul-Lez: Indeed - I seem to have confused myself there!

FormalConjectures/ErdosProblems/1072.lean
callesonne: What do you mean by this? That there is a density 1 subset of all primes for which the statement holds?

Edit: added a potential variant. Let me know if this was what you had in mind or not.

FormalConjectures/ErdosProblems/1072.lean
callesonne:
```suggestion
    (fun x ↦ ({p | p.Prime ∧ f p = p - 1}.interIcc 0 x).ncard : ℝ)) =o[atTop]
```

FormalConjectures/ErdosProblems/1072.lean
```diff
@@ -0,0 +1,66 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 1072
+
+*Reference:* [erdosproblems.com/1065](https://www.erdosproblems.com/1072)
```
YaelDillies:
```suggestion
*Reference:* [erdosproblems.com/1072](https://www.erdosproblems.com/1072)
```

FormalConjectures/ErdosProblems/1072.lean
```diff
@@ -0,0 +1,66 @@
...
+
+/-- For any prime $p$, let $f(p)$ be the least integer such that $f(p)! + 1 \equiv 0 \mod p$.-/
+noncomputable def f (p : ℕ) : ℕ := sInf {n | (n)! + 1 ≡ 0 [MOD p]}
+
+/-- Is it true that there are infinitely many $p$ for which $f(p) = p − 1$? -/
+@[category research open, AMS 11]
+theorem erdos_1072a : Set.Infinite {p | p.Prime ∧ f p = p - 1} ↔ answer(sorry) := by
+  sorry
+
+/-- Is it true that $f(p)/p \to 0$ for almost all $p$? -/
```
YaelDillies: That's... vague! I see this is how it is stated on the website, but my interpretation of it is rather that the density of such `p` tends to zero as we take more and more primes. In other words,
```suggestion
/-- Is it true that $f(p)/p \to 0$ as $p \to \infty$, except along a density zero set of primes? -/
```

FormalConjectures/ErdosProblems/1072.lean
```diff
@@ -0,0 +1,66 @@
...
+@[category research open, AMS 11]
+theorem erdos_1072b :
+    (∃ (P : Finset ℕ), (∀ p ∈ P, p.Prime) ∧
+      ∀ᵉ ε > (0 : ℝ), ∃ N₀, ∀ p ≥ N₀, p.Prime ∧ p ∉ P → f p / p < ε)
+    ↔ answer(sorry) := by
```
YaelDillies: Indeed, your statement is equivalent to
```suggestion
    (∀ᵉ ε > (0 : ℝ), ∃ N₀, ∀ p ≥ N₀, p.Prime → f p / p < ε)
    ↔ answer(sorry) := by
```
No prime is in `P` after some stage. Here is my suggested replacement
```suggestion
    (∃ (P : Set ℕ), P.HasDensity 0 {p | p.Prime} ∧
      Tendsto (fun p => f p / p) (atTop \inf principal ({p | p.Prime} \ P)) (\nhds 0)
    ↔ answer(sorry) := by
```

callesonne: Oh right, very good point. How does this compare to the statement `erdos_1072b.variants.density` below? I guess I should just make this the only version then?

FormalConjectures/ErdosProblems/1072.lean
```diff
@@ -0,0 +1,66 @@
...
+/-- Is it true that there are infinitely many $p$ for which $f(p) = p − 1$? -/
+@[category research open, AMS 11]
+theorem erdos_1072a : Set.Infinite {p | p.Prime ∧ f p = p - 1} ↔ answer(sorry) := by
+  sorry
+
+/-- Is it true that $f(p)/p \to 0$ as $p \to \infty$, except along a density zero set of primes? -/
+@[category research open, AMS 11]
+theorem erdos_1072b :
+    (∃ (P : Set ℕ), P.HasDensity 0 {p | p.Prime} ∧
+      Tendsto (fun p => f p / p) (atTop \inf principal ({p | p.Prime} \ P)) (\nhds 0)
```
YaelDillies: Won't work :)

FormalConjectures/ErdosProblems/1072.lean
```diff
@@ -0,0 +1,66 @@
...
+    ↔ answer(sorry) := by
+  sorry
+
+/-- Is it true that $f(p)/p \to 0$ for $p$ in a density 1 subset of the primes? -/
+@[category research open, AMS 11]
+theorem erdos_1072b.variants.density :
+    (∃ (P : Set ℕ), P ⊆ {p | p.Prime} ∧ P.HasDensity 1 {p | p.Prime} ∧
+      ∀ᵉ ε > (0 : ℝ), ∃ N₀, ∀ p ≥ N₀, p ∈ P → f p / p < ε)
```
YaelDillies: Can you write this using `Tendsto`?
```suggestion
      Tendsto (fun p => f p / p) (atTop ⊓ principal P) (𝓝 0))
```

FormalConjectures/ErdosProblems/1072.lean
```diff
@@ -0,0 +1,66 @@
...
+    ↔ answer(sorry) := by
+  sorry
+
+/--
+Erdős, Hardy, and Subbarao [HaSu02], believed that the number of $p \le x$ for which $f(p)=p−1$
+is $o(x/logx)$.
```
YaelDillies:
```suggestion
is $o(x/\log x)$.
```

FormalConjectures/ErdosProblems/1072.lean
callesonne:
```suggestion
      Tendsto (fun p => f p / p) (atTop ⊓ principal ({p | p.Prime} \ P)) (𝓝 0)
```

callesonne:
@YaelDillies should I not just remove the density 1 version now, given that the two should be equivalent? Or maybe make it the main version, since it is a bit cleaner to state?

YaelDillies:
Yeah, let's keep the density 1 version only

YaelDillies:
You might need to rerun CI once the github incident is over

PR #1136, opened 2025-10-21
Add infinitude of Mersenne primes

edwag:
I'm not sure where this should actually live. For now I've put it in `Wikipedia` with a reference to the wikipedia page for Mersenne primes.


mo271:
Perhaps this could be part of https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/Wikipedia/NewMersenne.lean 
(and then possibly move `NewMersenne.lean` to just `Mersenne.lean`?
At least it is definitely good to re-use `GivesMersennePrime` instead of redefining it as `IsMersennePrime`

FormalConjectures/Wikipedia/Mersenne.lean
```diff
@@ -85,4 +85,12 @@ theorem new_mersenne_conjecture.variants.prime (p : ℕ) (hp : p.Prime) :
     NewMersenneConjectureStatement p := by
   sorry
 
-end NewMersenne
+/--
+Are there infinitely many Mersenne primes?
+-/
+@[category research open, AMS 11]
+theorem infinitely_many_mersenne_primes :
+  Set.Infinite { p : ℕ | p.GivesMersennePrime } ↔ answer(sorry) := by
```
mo271, line 93: I personally also like `Set.Infinite (GivesMersennePrime : Set ℕ)`, but I think there is not consensus for this here.

PR #1138, opened 2025-10-22
Erdős 707: update solution status and add `0<n` assumption

### LINKED ISSUE #1137: Erdős Problem 707 minor issue
Erdős problem 707 asks whether a finite Sidon set can always be extended to a (finite!) perfect difference set.  https://www.erdosproblems.com/forum/thread/707

The file https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/707.lean has several versions of this conjecture, one of which is called `erdos_707.variants.weaker`:
https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/707.lean#L65-L71

There is a minor oversight here, in that the natural number `n` might be zero.  This is not the intended conjecture.

### Trivial proof/disproof
It is not hard to see that every Sidon set can be extended to a perfect difference set in Z/nZ with n=0.  (The proof is "just do it".)

For what it's worth, I have tried proving this in Lean and found it difficult...

### Suggested fix
Add the condition that n is nonzero.

### Additional context
The conjecture has been resolved, as reported in the forums [here](https://www.erdosproblems.com/forum/thread/707#post-1350).  It is false.  This includes `erdos_707` and `erdos_707.variants.weaker`.

A counterexample is {1,2,4,8,13}.  A Lean proof IS available.

As a result, the inclusion/exclusion of `n=0` flips the truth value of `erdos_707.variants.weaker`.

### Choose either option

mo271:
fixes #1137


mo271:
I propose to remove `erdos_707.variants.smallest_prime` and `erdos_707.variants.constructive` since they make less sense in light of the (re)discovered counterexamples.

FormalConjectures/ErdosProblems/707.lean
```diff
@@ -33,43 +36,66 @@ open Function Set
...
+A weaker version asking for any modulus, not necessarily of the form `p^2 + p + 1`.
 -/
-@[category research open, AMS 5 11]
-theorem erdos_707.variants.smallest_prime :
-    sInf {p : ℕ | IsPrimePow p ∧ ∃ (A : Set ℕ) (B : Set ℕ), A.Finite ∧ IsSidon A ∧
-      A ⊆ B ∧ IsPerfectDifferenceSet B (p^2 + p + 1)} =
-    answer(sorry) := by
+@[category research solved, AMS 5 11]
+theorem erdos_707.variants.weaker : (∀ (A : Set ℕ) (h : A.Finite), IsSidon A →
+    ∃ (B : Set ℕ), ∃ᵉ (n > 0), A ⊆ B ∧ IsPerfectDifferenceSet B n) ↔ false := by
```
smmercuri: was this the intention? Otherwise the subscript `e` can be removed. 
```suggestion
    ∃ᵉ (B : Set ℕ) (n > 0), A ⊆ B ∧ IsPerfectDifferenceSet B n) ↔ false := by
```

mo271: yes

FormalConjectures/ErdosProblems/707.lean
```diff
@@ -33,43 +36,66 @@ open Function Set
...
+difference set modulo `p^2 + p + 1` for some prime power `p`.
+
+As described in [A25], a counterexample is provided in [Ha47], see below.
 -/
-@[category research open, AMS 5 11]
+@[category research solved, AMS 5 11]
 theorem erdos_707 : (∀ (A : Set ℕ) (h : A.Finite), IsSidon A →
     ∃ (B : Set ℕ) (p : ℕ), IsPrimePow p ∧ A ⊆ B ∧
-    IsPerfectDifferenceSet B (p^2 + p + 1)) ↔ answer(sorry) := by
+    IsPerfectDifferenceSet B (p^2 + p + 1)) ↔ false := by
```
Paul-Lez:
```suggestion
    IsPerfectDifferenceSet B (p^2 + p + 1)) ↔ False := by
```

FormalConjectures/ErdosProblems/707.lean
```diff
@@ -48,15 +48,20 @@ theorem erdos_707 : (∀ (A : Set ℕ) (h : A.Finite), IsSidon A →
   simp
   sorry
 
+/--
+A  version asking for prime modulus.
```
plby: I think the word "order" should be used instead of "modulus".

Details: The modulus $p^2+p+1$ isn't necessarily prime.  However, every perfect difference set modulo $v=q^2+q+1$ corresponds to a projective plane with some extra structure on it.  This projective plane has order $q$.

FormalConjectures/ErdosProblems/707.lean
```diff
@@ -48,15 +48,20 @@ theorem erdos_707 : (∀ (A : Set ℕ) (h : A.Finite), IsSidon A →
...
+@[category research solved, AMS 5 11]
+theorem erdos_707.variants.weaker : (∀ (A : Set ℕ) (h : A.Finite), IsSidon A →
+    ∃ᵉ (B : Set ℕ) (p : ℕ), p.Prime ∧ A ⊆ B ∧ IsPerfectDifferenceSet B (p^2 + p + 1)) ↔ false := by
+  sorry
 
 /--
 A weaker version asking for any modulus, not necessarily of the form `p^2 + p + 1`.
 -/
 @[category research solved, AMS 5 11]
 theorem erdos_707.variants.weaker : (∀ (A : Set ℕ) (h : A.Finite), IsSidon A →
```
plby: Is there any issue with having two things named `erdos_707.variants.weaker`?

(I understand separately that you would like to change the name `weaker`.)

plby:
Our paper is now available on the arxiv at https://arxiv.org/abs/2510.19804 with only human authors, so the reference can be updated.

plby:
Regarding the names, I personally would have the version of general modulus $v>0$ be the main conjecture and then have two variants, one for prime order and one for prime power order.

One reasoning for this is that while the general modulus is the weakest conjecture when stated positively, its *disproof* is the strongest statement.  Another reasoning in favor is that it has the cleanest statement imo.

One argument against is that the one time I am aware of that Erdős offered 1000 dollars for the problem, he specifically stated the prime case.

plby:
I don't know how this is done in this project, but I think it's worth mentioning that the proofs of the counterexamples have been formalized in Lean?

FormalConjectures/ErdosProblems/707.lean
mo271: done, that makes sense...

FormalConjectures/ErdosProblems/707.lean
mo271: yeah, we can't use the same declaration twice, (already fixed)

mo271:
> Our paper is now available on the arxiv at https://arxiv.org/abs/2510.19804 with only human authors, so the reference can be updated.

done

mo271:
> Regarding the names, I personally would have the version of general modulus v > 0 be the main conjecture and then have two variants, one for prime order and one for prime power order.
> 
> One reasoning for this is that while the general modulus is the weakest conjecture when stated positively, its _disproof_ is the strongest statement. Another reasoning in favor is that it has the cleanest statement imo.
> 
> One argument against is that the one time I am aware of that Erdős offered 1000 dollars for the problem, he specifically stated the prime case.

Sounds reasonable. Normally we try just to at least roughly match what is written erdosproblems.com, but your suggestions to also mention the prime case. Completely agree regarding the "weaker".

mo271:
> I don't know how this is done in this project, but I think it's worth mentioning that the proofs of the counterexamples have been formalized in Lean?

Currently we don't track that information, but we should, and there might be an issue about that, if not I will create one. 
We explicitly don't want to add long proofs to this repository, since it will be much harder to maintain.

FormalConjectures/ErdosProblems/707.lean
```diff
@@ -32,44 +35,81 @@ open Function Set
 
 namespace Erdos707
 
+
 /--
-**Erdős Problem 707**: Any finite Sidon set can be embedded in a perfect difference set modulo
-`p^2 + p + 1` for some prime power `p`.
+**Erdős Problem 707**: Is is false that any finite Sidon set can be embedded in a perfect
```
b-mehta:
```suggestion
**Erdős Problem 707**: It is false that any finite Sidon set can be embedded in a perfect
```

FormalConjectures/ErdosProblems/707.lean
```diff
@@ -32,44 +35,81 @@ open Function Set
...
+
+/--
+This conjecture was actually first disproved by Hall in 1947 [Ha47], long before Erdős asked
+this question.
+A counterexample for any modulus from from [Ha47] in the paragraph following Theorem 4.3, where it
+was given as $\{-8, -6, 0, 1, 4\}$, but this can be shifted to natural numbers
+as pointed out in [arxiv/2510.19804].
+-/
+@[category research solved, AMS 5 11]
+theorem erdos_707.counterexample_hall(A : Set ℕ) (hA : A = {1, 3, 9, 10, 13}) :
```
b-mehta:
```suggestion
theorem erdos_707.counterexample_hall (A : Set ℕ) (hA : A = {1, 3, 9, 10, 13}) :
```

b-mehta (approved):

b-mehta:
> > I don't know how this is done in this project, but I think it's worth mentioning that the proofs of the counterexamples have been formalized in Lean?
> 
> Currently we don't track that information, but we should, and there might be an issue about that, if not I will create one.

The approach taken in the other files in which a solution has been formalised elsewhere is just to mention it in the theorem docstring, which might be enough?

plby:
> > Regarding the names, I personally would have the version of general modulus v > 0 be the main conjecture and then have two variants, one for prime order and one for prime power order.
> > One reasoning for this is that while the general modulus is the weakest conjecture when stated positively, its _disproof_ is the strongest statement. Another reasoning in favor is that it has the cleanest statement imo.
> > One argument against is that the one time I am aware of that Erdős offered 1000 dollars for the problem, he specifically stated the prime case.
> 
> Sounds reasonable. Normally we try just to at least roughly match what is written erdosproblems.com, but your suggestions to also mention the prime case. Completely agree regarding the "weaker".

erdosproblems.com currently asks about "prime", perhaps because that's the version that accompanied the $1000 offer.

Of course I'm okay with whatever you decide -- just sharing some thoughts.

FormalConjectures/ErdosProblems/707.lean
```diff
@@ -32,44 +35,84 @@ open Function Set
...
+**Erdős Problem 707**: It is false that any finite Sidon set can be embedded in a perfect
+different set modulo some $n$.
+
+As described in [arxiv/2510.19804], a counterexample is provided in [Ha47], see below.
+The proof of this has been formalized.
 -/
-@[category research open, AMS 5 11]
+@[category research solved, AMS 5 11]
 theorem erdos_707 : (∀ (A : Set ℕ) (h : A.Finite), IsSidon A →
+    ∃ᵉ (B : Set ℕ) (n > 0), A ⊆ B ∧ IsPerfectDifferenceSet B n) ↔ false := by
```
Paul-Lez:
```suggestion
    ∃ᵉ (B : Set ℕ) (n > 0), A ⊆ B ∧ IsPerfectDifferenceSet B n) ↔ False := by
```

Paul-Lez (approved):

FormalConjectures/ErdosProblems/707.lean
```diff
@@ -32,44 +35,84 @@ open Function Set
...
+/--
+It is false that any finite Sidon set can be embedded in a perfect
+difference set modulo `p^2 + p + 1` for some prime `p`.
+
+As described in [arxiv/2510.19804], a counterexample is provided in [Ha47], see below.
+The proof of this has been formalized.
+-/
+@[category research solved, AMS 5 11]
+theorem erdos_707.variants.prime : (∀ (A : Set ℕ) (h : A.Finite), IsSidon A →
+    ∃ᵉ (B : Set ℕ) (p : ℕ), p.Prime ∧ A ⊆ B ∧ IsPerfectDifferenceSet B (p^2 + p + 1)) ↔ false := by
```
mo271:
```suggestion
    ∃ᵉ (B : Set ℕ) (p : ℕ), p.Prime ∧ A ⊆ B ∧ IsPerfectDifferenceSet B (p^2 + p + 1)) ↔ False := by
```

PR #1139, opened 2025-10-22
silence `dupNamespace` linter in `AMSLinter`

mo271:
fallout from #1127


mo271:
Not sure if this is the best fix, but when renaming one of the `AMSLinter`, we should do so consistently for the other linters?

Paul-Lez:
That looks reasonable for now - I'll probably rename the AMS linter afterwards!

PR #1141, opened 2025-10-22
fixes after bump to `v4.22.0`

mo271:


Paul-Lez (approved):

FormalConjectures/ErdosProblems/730.lean
```diff
@@ -58,7 +58,7 @@ There are examples where $(n, m) ∈ S$ with $m ≠ n + 1$.
 (Found by AlphaProof, although it was implicit already in [A129515])
 -/
 @[category research solved, AMS 11]
-theorem erdos_730.variants.delta_ne_one : ∃ (n m : ℕ) (h : (n, m) ∈ S), m ≠ n + 1 := by
+theorem erdos_730.variants.delta_ne_one : ∃ (n m : ℕ) (_ : (n, m) ∈ S), m ≠ n + 1 := by
```
eric-wieser: I think using `\and` would be cleaner

eric-wieser (approved):

FormalConjectures/ErdosProblems/730.lean
mo271: agreed, done.

PR #1142, opened 2025-10-23
fix overwrite_index: importModules changes

mo271:
The `overwrite_index.lean` script failed to count attributes after a Lean version bump. This was due to changes in how module imports handle environment extensions, as introduced in
https://github.com/leanprover/lean4/pull/6325 .

`Lean.withImportModules` no longer loads environment extensions. This commit refactors the `runWithImports` function to use the lower-level `importModulesCore` and `finalizeImport` functions, explicitly setting `loadExts := true` to ensure attributes are loaded and counted correctly.

### DIFF
--- a/docbuild/scripts/overwrite_index.lean
+++ b/docbuild/scripts/overwrite_index.lean
@@ -96,13 +96,17 @@ unsafe def runWithImports {α : Type} (actionToRun : CoreM α) : IO α := do
   -- which we probably don't want to count here.
   let moduleImportNames := #[`FormalConjectures.All]
   initSearchPath (← findSysroot)
-  let modulesToImport : Array Import := moduleImportNames.map ({ module := · })
+
+  let imports : Array Import := moduleImportNames.map ({ module := · })
   let currentCtx := { fileName := "", fileMap := default }
   Lean.enableInitializersExecution
 
-  Lean.withImportModules modulesToImport {} fun env => do
-    let (result, _newState) ← Core.CoreM.toIO actionToRun currentCtx { env := env }
-    return result
+  let env ← withImporting do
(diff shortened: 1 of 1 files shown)


docbuild/scripts/overwrite_index.lean
```diff
@@ -96,13 +96,17 @@ unsafe def runWithImports {α : Type} (actionToRun : CoreM α) : IO α := do
...
+  let imports : Array Import := moduleImportNames.map ({ module := · })
   let currentCtx := { fileName := "", fileMap := default }
   Lean.enableInitializersExecution
 
-  Lean.withImportModules modulesToImport {} fun env => do
-    let (result, _newState) ← Core.CoreM.toIO actionToRun currentCtx { env := env }
-    return result
+  let env ← withImporting do
+    let (_, s) ← importModulesCore imports |>.run
+    finalizeImport s imports {} (trustLevel := 1024) (leakEnv := false) (loadExts := true)
```
Paul-Lez: OOC, what's the rationale for doing this rather than calling the slightly higher level `Lean.importModules` (looking at the implementation, that function seems to do almost the same thing as you!)

mo271: That is precisely the point: "almost the same" means a hardcoded `(loadExts := false)` and apparently that makes a differences...

mo271: see also https://github.com/Kha/lean4/blob/654c2241391e1e115ee4bd8e0ce01c628af9af8c/releases_drafts/environment.md

mo271: no, I understand, using that now

Paul-Lez (approved):
Nice, LGTM!

PR #1144, opened 2025-10-23
feat: Formalize Erdős Problem 97

### LINKED ISSUE #327: Erdős Problem 97
### What is the conjecture
    
https://www.erdosproblems.com/97

  Does every convex polygon have a vertex with no other $4$ vertices equidistant from it?

### Prerequisites needed
Definition of `convex polygon` not yet in mathlib.

junseokl:
Fixes #327.

I've used a slightly different definition for `HasNEquidistantProperty` than the `hasMinEquidistantProperty` found in [92.lean](https://github.com/google-deepmind/formal-conjectures/blob/4b01fd3ebc008540e4fec7c19d85063ecda8ad54/FormalConjectures/ErdosProblems/92.lean#L30-L52) using a modular, layered approach. 

As a bonus, I've also added the specific 9-point set for `erdos_97.variants.three_equidistant` which I found from the Danzer's construction (without proof). Just for fun! 😉

### DIFF
--- a/FormalConjectures/ErdosProblems/97.lean
+++ b/FormalConjectures/ErdosProblems/97.lean
@@ -39,7 +39,7 @@ A set of points $A$ has n equidistant points on a set of points $B$
 if for every point in $B$, there exist at least $n$ other points in $A$ that are equidistant from it.
 -/
 def HasNEquidistantPointsOn (n : ℕ) (A : Finset ℝ²) (B : Finset ℝ²) : Prop :=
-  B.Nonempty ∧ ∀ p ∈ B, HasNEquidistantPointsAt n A p
+  ∀ p ∈ B, HasNEquidistantPointsAt n A p
 
 /--
 A set of points $A$ has n equidistant property
@@ -60,7 +60,7 @@ A set of points $A$ has n unit distance points on a set of points $B$
 if for every point in $B$, there exist at least $n$ other points in $A$ that are at unit distance from it.
 -/
 def HasNUnitDistancePointsOn (n : ℕ) (A : Finset ℝ²) (B : Finset ℝ²) : Prop :=
-  B.Nonempty ∧ ∀ p ∈ B, HasNUnitDistancePointsAt n A p
+  ∀ p ∈ B, HasNUnitDistancePointsAt n A p
 
 /--
 A set of points $A$ has n unit distance property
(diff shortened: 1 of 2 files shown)


FormalConjectures/ErdosProblems/97.lean
```diff
@@ -0,0 +1,143 @@
...
+-/
+def HasNEquidistantPointsAt (n : ℕ) (A : Finset ℝ²) (p : ℝ²) : Prop :=
+  ∃ r : ℝ, r > 0 ∧ (A.filter fun q ↦ dist p q = r).card ≥ n
+
+/--
+A set of points $A$ has n equidistant points on a set of points $B$
+if for every point in $B$, there exist at least $n$ other points in $A$ that are equidistant from it.
+-/
+def HasNEquidistantPointsOn (n : ℕ) (A : Finset ℝ²) (B : Finset ℝ²) : Prop :=
+  B.Nonempty ∧ ∀ p ∈ B, HasNEquidistantPointsAt n A p
```
callesonne: To me it seems unnatural to put the `Nonempty` condition here, I would rather put it in the statement of `erdos_97` below (and wherever else its needed). I see that `erdos_92` also assumes `Nonempty` at a similar place, but I intend to try to refactor that file a bit anyways and then I will move that assumption elsewhere.

callesonne (approved):
Very clean, thanks a lot!

FormalConjectures/ErdosProblems/97.lean
junseokl: Pushed. I used `∧` to combine it with `ConvexIndep`. Happy to change it to `A.Nonempty → ...` or `(h : A.Nonempty) :` if you prefer that style.

FormalConjectures/ErdosProblems/97.lean
```diff
@@ -74,7 +74,7 @@ Does every convex polygon have a vertex with no other 4 vertices equidistant fro
 -/
 @[category research open, AMS 52]
 theorem erdos_97 :
-    (∀ A : Finset ℝ², ConvexIndep A → ¬HasNEquidistantProperty 4 A) ↔ answer(sorry) := by
+    (∀ A : Finset ℝ², A.Nonempty ∧ ConvexIndep A → ¬HasNEquidistantProperty 4 A) ↔ answer(sorry) := by
```
callesonne: I think here its better to use implication as this seems to be the standard convention elsewhere in this repo.

FormalConjectures/ErdosProblems/97.lean
callesonne: Added a comment that `A.Nonempty → ...` agrees better with the conventions in this repo. After that I'm happy to merge :)

FormalConjectures/ErdosProblems/97.lean
junseokl: Done!

callesonne: Thanks!

PR #1145, opened 2025-10-24
OEIS A228828

mo271:


FormalConjectures/Oeis/228828.lean
```diff
@@ -0,0 +1,67 @@
...
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+Numbers n such that n^2 + pi(n) is prime.
+
+*Reference:* [A228828](https://oeis.org/A228828)
+-/
+
+namespace Oeis228828
```
callesonne:
```suggestion
namespace OeisA228828
```
Do you need an `A` here also?

mo271: I guess it wouldn't hurt

FormalConjectures/Oeis/228828.lean
```diff
@@ -0,0 +1,67 @@
...
+  · decide
+
+/--
+Conjecture: the sequence A228828 is infinite.
+-/
+@[category research open, AMS 11]
+theorem a.infinite : {a n | n}.Infinite := by
+  sorry
+
+end Oeis228828
```
mo271:
```suggestion
end OeisA228828
```

PR #1151, opened 2025-10-28
fix: `erdos_128` should use induced subgraphs

smmercuri:
Original statement ([source](https://www.tandfonline.com/doi/epdf/10.1080/16073606.1993.9631741?needAccess=true)) uses induced subgraphs rather than general subgraphs


mo271:
not sure if we want to add a comment to https://www.erdosproblems.com/128 or if "every subgraph" in that statement is sufficiently clear already

mo271 (approved):

smmercuri:
> not sure if we want to add a comment to https://www.erdosproblems.com/128 or if "every subgraph" in that statement is sufficiently clear already

Good shout, I'll do that

PR #1179, opened 2025-11-07
Erdős Problem 332

### LINKED ISSUE #492: Erdős Problem 332
### What is the conjecture
    
https://www.erdosproblems.com/332

  Let $A\subseteq \mathbb{N}$ and $D(A)$ be the set of those numbers which occur infinitely often as $a_1-a_2$ with $a_1,a_2\in A$. What conditions on $A$ are sufficient to ensure $D(A)$ has bounded gaps?

Status: open

mo271:
fixes #492

### DIFF
--- a/FormalConjectures/ErdosProblems/332.lean
+++ b/FormalConjectures/ErdosProblems/332.lean
@@ -48,7 +48,7 @@ has to provide what the sufficient conditions are, and proof that they imply the
 If the condition is a solution to the problem is up to human judgement.
 -/
 @[category research open, AMS 11]
-theorem erdos_332 : ∀ (A : Set ℕ), (answer(sorry) : Set ℕ → Prop) A → HasBoundedGaps (D_A A) := by
+theorem erdos_332 (A : Set ℕ) : (answer(sorry) : Set ℕ → Prop) A → HasBoundedGaps (D_A A) := by
   sorry
 
 --TODO(firsching): formalize additional statements


FormalConjectures/ErdosProblems/332.lean
```diff
@@ -0,0 +1,56 @@
...
+-/
+noncomputable def D_A (A : Set ℕ) : Set ℤ :=
+  { d : ℤ | Set.Infinite { p : ℕ × ℕ | p.1 ∈ A ∧ p.2 ∈ A ∧ (p.1 : ℤ) - (p.2 : ℤ) = d } }
+
+/--
+A set $S \subseteq \mathbb{Z}$ has bounded gaps if it is syndetic, meaning there is a uniform
+bound $M$ such that every interval of length $M$ contains an element of $S$.
+-/
+def HasBoundedGaps (S : Set ℤ) : Prop :=
+  ∃ M : ℕ, M > 0 ∧ ∀ z : ℤ, ∃ s ∈ S, z ≤ s ∧ s < z + (M : ℤ)
```
mo271, line 39: some other problems use this, the we could potentially move it to ForMathlib, but maybe let's do it when we first formalise on of those problems that use it.

callesonne, line 39: What is the motivation for this interpretation of having bounded gaps in this case? It seems that this definition implies that the set is infinite.

FormalConjectures/ErdosProblems/332.lean
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Erdos332
+
+open Set
+
+/--
+The set of numbers $D(A)$ which occur infinitely often as $a_1 - a_2$ with $a_1, a_2 \in A$.
+-/
+noncomputable def D_A (A : Set ℕ) : Set ℤ :=
+  { d : ℤ | Set.Infinite { p : ℕ × ℕ | p.1 ∈ A ∧ p.2 ∈ A ∧ (p.1 : ℤ) - (p.2 : ℤ) = d } }
```
callesonne:
```suggestion
  { d : ℤ | Set.Infinite { (a, b) : ℕ × ℕ | a ∈ A ∧ b ∈ A ∧ (a : ℤ) - (b : ℤ) = d } }
```

callesonne (approved):
LGTM apart from the comments above

FormalConjectures/ErdosProblems/332.lean
mo271: I don't quite understand the comment. Do you suggest to add a test statement showing that the definition implies that the set is infinte?

mo271: @callesonne?

Paul-Lez: I think the point @callesonne was trying to make is that this definition is stronger than the usual definition of bounded gaps: as any finite set has bounded gaps if one uses the same definition as Erdos when he states the problem (which simply says that for all `n`, `(n+1).nth (· ∈ S)  - n.nth ( · ∈ S)` is at most `M` for some constant `M`). 

That being said, I think it's clear from the context that all sets under consideration are infinite so I think this isn't too bad.

FormalConjectures/ErdosProblems/332.lean
```diff
@@ -0,0 +1,56 @@
...
+Let $A\subseteq \mathbb{N}$ and $D(A)$ be the set of those numbers which occur infinitely often as
+$a_1 - a_2$ with $a_1, a_2\in A$. What conditions on $A$ are sufficient to ensure $D(A)$ has bounded
+gaps?
+
+This is formalised here using the `answer(sorry)` mechanism. In order to solve this problem one
+has to provide what the sufficient conditions are, and proof that they imply the desired condition.
+If the condition is a solution to the problem is up to human judgement.
+-/
+@[category research open, AMS 11]
+theorem erdos_332 : ∀ (A : Set ℕ), (answer(sorry) : Set ℕ → Prop) A → HasBoundedGaps (D_A A) := by
```
Paul-Lez:
```suggestion
theorem erdos_332 (A : Set ℕ) : (answer(sorry) : Set ℕ → Prop) A → HasBoundedGaps (D_A A) := by
```

PR #1189, opened 2025-11-11
fix: positivity in `erdos_253`

smmercuri:
Statement concerns only positive integers, ref. [Cassels' solution paper](https://www.brand.site.co.il/riddles/201507a_files/21111124.pdf)

### DIFF
--- a/FormalConjectures/ErdosProblems/253.lean
+++ b/FormalConjectures/ErdosProblems/253.lean
@@ -39,7 +39,7 @@ integers which are the sum of distinct $a_i$ then every sufficiently large integ
 the sum of distinct $a_i$.
 -/
 @[category research solved, AMS 11]
-theorem erdos_253 : ¬ ∀ a : ℕ → ℕ, a 0 > 0 →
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/253.lean
```diff
@@ -26,20 +26,20 @@ namespace Erdos253
 
 open scoped Topology
 
-/-- The predicate that `a : ℕ → ℤ` is a strictly monotone sequence such that every infinite
+/-- The predicate that `a : ℕ → ℕ` is a strictly monotone sequence such that every infinite
 arithmetic progression contains infinitely many integers that are the sum of distinct $a_i$s.-/
 @[inline]
-def RepresentsAPs (a : ℕ → ℤ) : Prop :=
+def RepresentsAPs (a : ℕ → ℕ) : Prop :=
```
Paul-Lez, line 32: I guess this edit isn't really needed?

smmercuri: It's required if we keep the codomain change in `erdos_253`. I guess neither are really required since we assert `0 < a 0` but seems more natural for it to be `ℕ` anyway

PR #1193, opened 2025-11-11
fix: add cardinality condition to `erdos_274` 

smmercuri:
1. All claims are false if `ENat.card G ≤ 1` because no partition of `G` of size `> 1` can exist.
2. The website has since been updated to remove the condition that `G` is abelian since that case was solved.

### DIFF
--- a/FormalConjectures/ErdosProblems/274.lean
+++ b/FormalConjectures/ErdosProblems/274.lean
@@ -35,9 +35,8 @@ If `G` is a group then can there exist an exact covering of `G` by more than one
 different sizes? (i.e. each element is contained in exactly one of the cosets.)
 -/
 @[category research open, AMS 20]
-theorem erdos_274 :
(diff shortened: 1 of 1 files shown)


mo271 (approved):
We might want to keep the abelian variant as a solved one?

smmercuri:
> We might want to keep the abelian variant as a solved one?

Added back in. Seems to me like we might want to take the quantification over `G` out of the `↔ answer(sorry)` in the main result `erdos_274` because `answer(False)` does not simplify to the same statement as `variants.abelian` now. What do you think?

mo271:
> > We might want to keep the abelian variant as a solved one?
> 
> Added back in. Seems to me like we might want to take the quantification over `G` out of the `↔ answer(sorry)` in the main result `erdos_274` because `answer(False)` does not simplify to the same statement as `variants.abelian` now. What do you think?

yes, let's do that!

PR #1199, opened 2025-11-12
Correct Littlewood Conjecture category to 'open'

YanYablonovskiy:
Correcting the category of the Littlewood Conjecture to match source.


mo271 (approved):
Thanks!

PR #1207, opened 2025-11-14
Erdős Problem 317

### LINKED ISSUE #483: Erdős Problem 317
### What is the conjecture
    
https://www.erdosproblems.com/317

  Is there some constant $c>0$ such that for every $n\geq 1$ there exists some $\delta_k\in \\{-1,0,1\\}$ for $1\leq k\leq n$ with $0< \left\lvert \sum_{1\leq k\leq n}\frac{\delta_k}{k}\right\rvert < \frac{c}{2^n}$?
Is it true that for sufficiently large $n$, for any $\delta_k\in \\{-1,0,1\\}$,
$\left\lvert \sum_{1\leq k\leq n}\frac{\delta_k}{k}\right\rvert > \frac{1}{[1,\ldots,n]}$
whenever the left-hand side is not zero?

Status: open

### Choose either option

Molybdaen42:
Implemented Erdős Problem 317

fixes #483

### DIFF
--- a/FormalConjectures/ErdosProblems/317.lean
+++ b/FormalConjectures/ErdosProblems/317.lean
@@ -69,7 +69,7 @@ theorem erdos_317.variants.counterexample : ¬ (∀  δ : (Fin 4) → ℚ, δ ''
     lhs ≠ 0 → lhs > (1 : ℚ) / ((Icc 1 4).lcm id : ℕ)) := by
   push_neg
   use ![0, 1, -1, -1]
-  norm_num[Finset.sum]
+  norm_num [Finset.sum]
   refine ⟨by grind, le_of_eq ?_⟩
   exact (abs_of_nonneg (by norm_num)).trans (one_div _)
 


FormalConjectures/ErdosProblems/317.lean
```diff
@@ -0,0 +1,68 @@
...
+-/
+
+namespace Erdos317
+
+/--
+The least common multiple of the numbers 1,...,n.
+-/
+abbrev lcm_one_to_n : ℕ → ℕ
+    | 0 => 1
+    | n + 1 => (lcm_one_to_n n).lcm (n + 1)
```
mo271: I would tend to not have that definition, but rather put 
`open Finset` in the beginning and then just use `(Icc 1 n).lcm id` or whatever. This is almost as short and no new definition needed. Even without the `open Finset` one could just do `(Finset.Icc 1 n).lcm id`

FormalConjectures/ErdosProblems/317.lean
```diff
@@ -0,0 +1,68 @@
...
+
+/--
+Inequality is obvious for `erdos_317.variants.claim2`, the problem is strict inequality.
+This fails for small $n$, for example\[\frac{1}{2}-\frac{1}{3}-\frac{1}{4}=-\frac{1}{12}.\]
+-/
+@[category API, AMS 11]
+lemma claim2_inequality : ∃ N : ℕ, ∀ n ≥ N,
```
mo271: here and below one can use `∀ᶠ n in Filter.atTop,` (or just `∀ᶠ n in atTop,` after adding `open Filter`) for "for all sufficiently large `n`"

FormalConjectures/ErdosProblems/317.lean
```diff
@@ -0,0 +1,68 @@
...
+    ∀ δ : ℕ → ℝ, δ '' (Finset.Icc 1 n) ⊆ {-1,0,1} →
+    abs (∑ k ∈ Finset.Icc 1 n, ((δ k : ℝ) / k)) ≠ 0 →
+        abs (∑ k ∈ Finset.Icc 1 n, ((δ k : ℝ) / k)) ≥ 1 / (lcm_one_to_n n : ℝ) := by
```
mo271:
```suggestion
    letI lhs := |∑ k ∈ Finset.Icc 1 n, ((δ k : ℝ) / k)|
    lhs  ≠ 0 → lhs ≥ 1 / (lcm_one_to_n n : ℝ) := by
```

Consider this inline definition, so that you don't have to repeat defining the sum. Also notation `|...|` might be nicer than `abs`

Here and everywhere below

FormalConjectures/ErdosProblems/317.lean
```diff
@@ -0,0 +1,68 @@
...
+-/
+abbrev lcm_one_to_n : ℕ → ℕ
+    | 0 => 1
+    | n + 1 => (lcm_one_to_n n).lcm (n + 1)
+
+/--
+Inequality is obvious for `erdos_317.variants.claim2`, the problem is strict inequality.
+This fails for small $n$, for example\[\frac{1}{2}-\frac{1}{3}-\frac{1}{4}=-\frac{1}{12}.\]
+-/
+@[category API, AMS 11]
```
mo271:
```suggestion
@[category undergraduate, AMS 11]
```

FormalConjectures/ErdosProblems/317.lean
```diff
@@ -0,0 +1,68 @@
...
+\[\left\lvert \sum_{1\leq k\leq n}\frac{\delta_k}{k}\right\rvert > \frac{1}{[1,\ldots,n]}\]
+whenever the left-hand side is not zero?
+-/
+@[category research open, AMS 11]
+theorem erdos_317.variants.claim2 : (∃ N : ℕ, ∀ n ≥ N,
+    ∀ δ : ℕ → ℝ, δ '' (Finset.Icc 1 n) ⊆ {-1,0,1} →
+    abs (∑ k ∈ Finset.Icc 1 n, ((δ k : ℝ) / k)) ≠ 0 →
+        abs (∑ k ∈ Finset.Icc 1 n, ((δ k : ℝ) / k)) > 1 / (lcm_one_to_n n : ℝ)) ↔ answer(sorry) := by
+  sorry
+
```
mo271, line 51:
```suggestion
@[category graduate, AMS 11]
theorem erdos_317.variants.counterexample : ¬ (∀  δ : (Fin 4) → ℚ, δ '' Set.univ ⊆ {-1,0,1} →
    letI lhs := |∑ k, ((δ k : ℚ) / (k + 1))|
    lhs ≠ 0 → lhs > (1 : ℚ) / ((Icc 1 4).lcm id : ℕ)) := by
  push_neg
  use ![0, 1, -1, -1]
  norm_num[Finset.sum]
  refine ⟨by grind, le_of_eq ?_⟩
  exact (abs_of_nonneg (by norm_num)).trans (one_div _)
```

Here's the counterexample, which perhaps could also be added. Now using the additional idea to use `Fin n` instead of `Finset.Icc 1 n`, because that makes some things easier, (but one has to index shift by `k+1`, which might be not so nice, I suppose a question of taste, but it is not nice to define `δ` on more numbers than necessary.

mo271 (commented):
Thanks for the contributions!

A first few remarks, but looks good in general!

Molybdaen42:
Thanks for your comments! I think I took all of them into account.

FormalConjectures/ErdosProblems/317.lean
```diff
@@ -0,0 +1,76 @@
...
+open Finset
+open Filter
+
+/--
+Is there some constant $c>0$ such that for every $n\geq 1$ there exists some $\delta_k\in \{-1,0,1\}$ for $1\leq k\leq n$ with
+\[0< \left\lvert \sum_{1\leq k\leq n}\frac{\delta_k}{k}\right\rvert < \frac{c}{2^n}?\]
+-/
+@[category research open, AMS 11]
+theorem erdos_317 : (∃ c : ℝ, c > 0 → ∀ n ≥ 1,
+    ∃ δ : (Fin n) → ℚ, δ '' Set.univ ⊆ {-1,0,1}∧
```
mo271:
```suggestion
    ∃ δ : (Fin n) → ℚ, δ '' Set.univ ⊆ {-1,0,1} ∧
```

FormalConjectures/ErdosProblems/317.lean
```diff
@@ -0,0 +1,76 @@
...
+`erdos_317.variants.claim2` fails for small $n$, for example
+\[\frac{1}{2}-\frac{1}{3}-\frac{1}{4}=-\frac{1}{12}.\]
+-/
+@[category graduate, AMS 11]
+theorem erdos_317.variants.counterexample : ¬ (∀  δ : (Fin 4) → ℚ, δ '' Set.univ ⊆ {-1,0,1} →
+    letI lhs := |∑ k, ((δ k : ℚ) / (k + 1))|
+    lhs ≠ 0 → lhs > (1 : ℚ) / ((Icc 1 4).lcm id : ℕ)) := by
+  push_neg
+  use ![0, 1, -1, -1]
+  norm_num[Finset.sum]
```
mo271:
```suggestion
  norm_num [Finset.sum]
```

(I know you copied  what I suggested earlier, but I apparently forgot that space..)

mo271:
> Thanks for your comments! I think I took all of them into account.

In that case it would be good to click "Resolve Conversation", so it is easier to keep track of what comments have been addressed and which ones are still open. (Also you can just accept edit suggestions in the github ui, if that is easier..)

mo271:
Thanks again!

PR #1210, opened 2025-11-17
Erdős Problem 749

### LINKED ISSUE #926: Erdős Problem 749
### What is the conjecture

https://www.erdosproblems.com/749

  Let $\epsilon>0$. Does there exist $A\subseteq \mathbb{N}$ such that the lower density of $A+A$ is at least $1-\epsilon$ and yet $1_A\ast 1_A(n) \ll_\epsilon 1$ for all $n$?

Status:  open

mo271:
fixes #926

### DIFF
--- a/FormalConjectures/ErdosProblems/749.lean
+++ b/FormalConjectures/ErdosProblems/749.lean
@@ -33,7 +33,7 @@ and yet $1_A\ast 1_A(n) \ll_\epsilon 1$ for all $n$?
 @[category research open, AMS 11]
 theorem erdos_749 : (∀ᵉ (ε > (0 : ℝ)),
     ∃ A : Set ℕ, 1 - ε ≤ lowerDensity (A + A) ∧
-    ((Nat.cast (R := ℝ) ∘ sumRep A) ≪ Nat.cast (R := ℝ) ∘ 1)) ↔ answer(sorry) := by
+    ((Nat.cast (R := ℝ) ∘ sumRep A) ≪ (fun n => (1: ℝ)))) ↔ answer(sorry) := by
   sorry
 
 


FormalConjectures/ErdosProblems/749.lean
```diff
@@ -13,28 +13,29 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 -/
-noncomputable def A (x : ℕ) : ℝ := {u | u.Composite ∧ ∃ n, n ! + 1 ≡ 0 [MOD u] ∧ u < x}.ncard
-
-/-- Is it true that $A(x) ≤ x^{o(1)}$? -/
 @[category research open, AMS 11]
-theorem erdos_1073 :
-    (∃ (o : ℕ → ℝ), o =o[atTop] (1 : ℕ → ℝ) ∧ ∀ x, A x ≤ x ^ (o x)) ↔ answer(sorry) := by
+theorem erdos_749 : (∀ᵉ (ε > (0 : ℝ)),
+    ∃ A : Set ℕ, 1 - ε ≤ lowerDensity (A + A) ∧
+    ((Nat.cast (R := ℝ) ∘ sumRep A) ≪ Nat.cast (R := ℝ) ∘ 1)) ↔ answer(sorry) := by
```
Paul-Lez: Not hyper important, but I think this is a little more readable. 
```suggestion
    ((Nat.cast (R := ℝ) ∘ sumRep A) ≪ (fun n => 1))) ↔ answer(sorry) := by
```

mo271: I think one needs to provide the type of `1` in that case:
```lean4
    ((Nat.cast (R := ℝ) ∘ sumRep A) ≪ (fun n => (1 : ℝ)))) ↔ answer(sorry) := by
```
Then it is almost as ugly, no?

PR #1219, opened 2025-11-18
Erdős Problem 705

### LINKED ISSUE #906: Erdős Problem 705
### What is the conjecture

https://www.erdosproblems.com/705

  Let $G$ be a finite unit distance graph in $\mathbb{R}^2$ (i.e. the vertices are a finite collection of points in $\mathbb{R}^2$ and there is an edge between two points if and only if the distance between them is $1$).

Is there some $k$ such that if $G$ has girth $\geq k$ (i.e. $G$ contains no cycles of length $<k$) then $\chi(G)\leq 3$?

Status:  open

80mb3rm4n:
fixes #906

### DIFF
--- a/FormalConjectures/ErdosProblems/1054.lean
+++ b/FormalConjectures/ErdosProblems/1054.lean
@@ -1,0 +1,68 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 1054
+
+*Reference:* [erdosproblems.com/1054](https://www.erdosproblems.com/1054)
+-/
+
+namespace Erdos1054
+
+open Classical Filter Asymptotics
+
+/-- Let $f(n)$ be the minimal integer $m$ such that $n$ is the sum of the $k$ smallest
+divisors of $m$ for some $k\geq 1$.-/
+noncomputable def f (n : ℕ) : ℕ :=
+  if h : ∃ᵉ (m) (k ≥ 1), n = ∑ i < k, Nat.nth (· ∈ m.divisors) i then
+    Nat.find h
+  else 0
+
+/-- Let $f(n)$ be the minimal integer $m$ such that $n$ is the sum of the $k$ smallest divisors
+of $m$ for some $k\geq 1$. Is it true that $f(n)=o(n)$?-/
+@[category research open, AMS 11]
+theorem erdos_1054.parts.i : (fun n ↦ (f n : ℝ)) =o[atTop] (fun n ↦ (n : ℝ)) ↔ answer(sorry) := by
+  sorry
+
+/-- Let $f(n)$ be the minimal integer $m$ such that $n$ is the sum of the $k$ smallest divisors
+of $m$ for some $k\geq 1$. Is it true that $f(n)=o(n)$ for almost all $n$? -/
+@[category research open, AMS 11]
+theorem erdos_1054.parts.ii : (∃ (A : Set ℕ), A.HasDensity 1 ∧
+    (fun (n : A) ↦ (f ↑n : ℝ)) =o[atTop] (fun n ↦ (n : ℝ))) ↔ answer(sorry) := by
+  sorry
+
+/-- Let $f(n)$ be the minimal integer $m$ such that $n$ is the sum of the $k$ smallest divisors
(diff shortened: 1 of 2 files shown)


FormalConjectures/ErdosProblems/705.lean
```diff
@@ -0,0 +1,50 @@
...
+
+/-!
+# Erdős Problem 705
+
+*Reference:* [erdosproblems.com/705](https://www.erdosproblems.com/705)
+-/
+
+namespace Erdos705
+
+open Erdos705
```
mo271:
```suggestion
open scoped EuclideanGeometry
```

FormalConjectures/ErdosProblems/705.lean
```diff
@@ -0,0 +1,50 @@
...
+
+variable {V : Finset (EuclideanSpace ℝ (Fin 2))}
```
mo271:
```suggestion
variable {V : Finset ℝ²}
```
this should work with `open scoped EuclideanGeometry`

FormalConjectures/ErdosProblems/705.lean
```diff
@@ -0,0 +1,50 @@
...
+
+/--
+A finite unit distance graph in ℝ^2:
+A graph where the vertices are a finite collection of points in ℝ² and there is an edge between
+two points if and only if the distance between them is 1.
+-/
+
+def unitDistancePlaneGraph (G : SimpleGraph V) : Prop :=
+  ∀ x y : V, G.Adj x y ↔ dist x y = 1
```
mo271: potentially merge definitions with https://github.com/google-deepmind/formal-conjectures/blob/a91ab6df34970fdd0fd8c919e693b2ba81fa4fbb/FormalConjectures/Wikipedia/HadwigerNelson.lean#L33 
I know this is slightly different, but perhaps one could make a more general definition and then special case in both files. I suppose there will be more conjectures talking about the unit distance graph.
Or just leave a TODO for it.

FormalConjectures/ErdosProblems/705.lean
```diff
@@ -0,0 +1,50 @@
...
+/--
+Let G be a finite unit distance graph in R².
+Is there some k such that if G has girth ≥ k, then χ(G) ≤ 3?
+-/
+
+@[category research open, AMS 5]
+theorem erdos_705:
+  (∀ (G : SimpleGraph V) (h : unitDistancePlaneGraph G),
+    ∃ k, SimpleGraph.girth G ≥ k → SimpleGraph.chromaticNumber G ≤ 3) ↔ answer(sorry) := by sorry
+
```
mo271, line 42: Please add a TODO for the statements in the additional material (the concrete constructions etc)

FormalConjectures/ErdosProblems/705.lean
80mb3rm4n: We generalized the definition and updated it in both files. The definition appears in the Definition file of the directory GraphConjectures

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
```diff
@@ -87,4 +88,16 @@ noncomputable def indepNeighbors (G : SimpleGraph α) (v : α) : ℝ :=
 noncomputable def averageIndepNeighbors (G : SimpleGraph α) : ℝ :=
   (∑ v ∈ Finset.univ, indepNeighbors G v) / (Fintype.card α : ℝ)
 
+/-- A finite unit distance graph in ℝ^2:
+A graph where the vertices are a finite collection of points in ℝ² and there is an edge between
+two points if and only if the distance between them is 1. -/
```
mo271: Please update the docstring: now there is not finiteness condition here anymore, right?

mo271 (approved):
Looks good to me, modulo the comment about the docstring.

Nice to unify the files in this way!

FormalConjectures/ForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
80mb3rm4n: True, now it simply passes the set of vertices of \R^2, which could be finite or infinite

FormalConjectures/ErdosProblems/705.lean
```diff
@@ -0,0 +1,43 @@
...
+/--
+Let G be a finite unit distance graph in R².
+Is there some k such that if G has girth ≥ k, then χ(G) ≤ 3?
+-/
+@[category research open, AMS 5]
+theorem erdos_705:
+  (∀ (V : Set ℝ²) (hf: V.Finite),
+    ∃ k, (UnitDistancePlaneGraph V).girth ≥ k → (UnitDistancePlaneGraph V).chromaticNumber ≤ 3) ↔
+    answer(sorry) := by sorry
```
smmercuri, line 38: You need `∧` rather than `→` otherwise there is a contradiction by choosing any `k > (UnitDistancePlaneGraph V).girth`
```suggestion
  (∀ (V : Set ℝ²) (hf: V.Finite),
    ∃ k, (UnitDistancePlaneGraph V).girth ≥ k ∧ (UnitDistancePlaneGraph V).chromaticNumber ≤ 3) ↔
    answer(sorry) := by sorry
```

80mb3rm4n: You are right, I think the original statement in the Erd\H{o}s problems webpage has also this problem.

PR #1222, opened 2025-11-18
fix: zero indexing in `erdos_695.variants.upperBound` 

smmercuri:
Sequence is one-indexed on [the website](https://www.erdosproblems.com/forum/thread/695) while we zero-index here. While this has no impact on the main problem, `k = 0` leads to a counterexample in the upper bound variant.

### DIFF
--- a/FormalConjectures/ErdosProblems/695.lean
+++ b/FormalConjectures/ErdosProblems/695.lean
@@ -55,6 +55,7 @@ theorem erdos_695.variant.upperBound :
       (∀ i, q (i + 1) % q i = 1) ∧
       ∃ o : ℕ → ℝ,
         (o =o[atTop] (1 : ℕ → ℝ)) ∧
+        -- We use `(k + 1)` here as the informal statement is 1-indexed.
         ∀ k, q k ≤ exp ((k + 1) * (log (k + 1)) ^ (1 + o k))) ↔
     answer(sorry) := by
   sorry


FormalConjectures/ErdosProblems/695.lean
```diff
@@ -55,7 +55,7 @@ theorem erdos_695.variant.upperBound :
       (∀ i, q (i + 1) % q i = 1) ∧
       ∃ o : ℕ → ℝ,
         (o =o[atTop] (1 : ℕ → ℝ)) ∧
-        ∀ k, q k ≤ exp (k * (log k) ^ (1 + o k))) ↔
+        ∀ k, q k ≤ exp ((k + 1) * (log (k + 1)) ^ (1 + o k))) ↔
```
Paul-Lez, line 59: I'd be tempted to just require `k` to be positive, rather than re-indexing. What do you think?

smmercuri: I think we might then have to also require `i` to be positive in the properties about `q` above it as well, so the shift seemed slightly cleaner. Or perhaps we should just index with `PNat`. I have no strong feelings here about which is better

callesonne, line 59:
```suggestion
        -- We use `(k + 1)` here as the informal statement is 1-indexed.
        ∀ k, q k ≤ exp ((k + 1) * (log (k + 1)) ^ (1 + o k))) ↔
```

PR #1225, opened 2025-11-18
Erdos 119

### LINKED ISSUE #360: Erdős Problem 119
### What is the conjecture
    
https://www.erdosproblems.com/119

  Let $z_i$ be an infinite sequence of complex numbers such that $\lvert z_i\rvert=1$ for all $i\geq 1$, and for $n\geq 1$ let
$$p_n(z)=\prod_{i\leq n} (z-z_i).$$
Let $M_n=\max_{\lvert z\rvert=1}\lvert p_n(z)\rvert$. Is it true that $\limsup M_n=\infty$? Is it true that there exists $c>0$ such that for infinitely many $n$ we have $M_n > n^c$? Is it true that there exists $c>0$ such that, for all large $n$,
$$\sum_{k\leq n}M_k > n^{1+c}?$$

Status: open

### Prerequisites needed
All that is needed should be in mathlib

### Choose either option

porianes:
fixes #360

### DIFF
--- a/FormalConjectures/ErdosProblems/119.lean
+++ b/FormalConjectures/ErdosProblems/119.lean
@@ -26,41 +26,47 @@ open Filter Finset Set
 
 namespace Erdos119
 
-/-- Let $z_i$ be an infinite sequence of complex numbers such that $|z_i| = 1$ for all $i \geq 1$. -/
-variable (z : PNat → ℂ) (hz : ∀ i : PNat, ‖z i‖ = 1)
+/-
+Here we use 0-indexing for generality and convenience, while in the original problem
+formulation 1-indexing was used. This change does not affect the meaning of the problem.
+In the description of the problem below we remain faithful to the original one.
+-/
 
-/-- For $n \geq 1$ let $p_n(z) = \prod_{i \leq n} (z - z_i)$. -/
-noncomputable def p (n : PNat) : ℂ → ℂ :=
-    fun w => ∏ i ∈ range n, (w - z ⟨i + 1, by linarith⟩)
+/-- Let $z_i$ be an infinite sequence of complex numbers such that $|z_i| = 1$ for all $i \geq 1$.
+For $n \geq 1$ let $p_n(z) = \prod_{i \leq n} (z - z_i)$. -/
+noncomputable def p (z : ℕ → ℂ) (n : ℕ) : ℂ → ℂ :=
+    fun w => ∏ i ∈ range n, (w - z i)
 
 /-- Let $M_n = \max_{|z| = 1} |p_n(z)|$. -/
-noncomputable def M (n : PNat) : ℝ :=
+noncomputable def M (z : ℕ → ℂ) (n : ℕ) : ℝ :=
     sSup { (‖p z n w‖) | (w : ℂ) (_ : ‖w‖ = 1) }
 
 /-- Question 1:
 
 Is it true that $\limsup M_n = \infty$?
 
-Wagner [Wa80] proved that there is some c > 0 with Mₙ > (log n)ᶜ infintely often.
+Wagner [Wa80] proved that there is some $c > 0$ with $M_n > (\log n)^c$ infintely often.
 
 [Wa80] Wagner, Gerold, On a problem of {E}rdős in {D}iophantine approximation. Bull. London Math. Soc. (1980), 81--88.
 -/
 @[category research solved, AMS 30]
 theorem erdos_119_1 :
-    atTop.limsup (fun n => (M z n : EReal)) = ⊤ ↔ answer(True) := by
+    (∀ (z : ℕ → ℂ) (hz : ∀ i : ℕ, ‖z i‖ = 1),
+      atTop.limsup (fun n => (M z n : EReal)) = ⊤) ↔ answer(True) := by
   sorry
 
 /-- Question 2:
 
 Is it true that there exists $c > 0$ such that for infinitely many $n$ we have $M_n > n^c$?
 
-Beck [Be91] proved that there exists some c > 0 such that max_{n ≤ N} Mₙ > Nᶜ.
+Beck [Be91] proved that there exists some $c > 0$ such that $\max_{n \leq N} M_n > N^c$.
 
 [Be91] Beck, J., The modulus of polynomials with zeros on the unit circle: A problem of Erdős. Annals of Math. (1991), 609-651.
 -/
 @[category research solved, AMS 30]
 theorem erdos_119_2 :
-    ∃ (c : ℝ) (hc : c > 0), Set.Infinite {n : PNat | M z n > n ^ c} ↔ answer(True) := by
+    (∀ (z : ℕ → ℂ) (hz : ∀ i : ℕ, ‖z i‖ = 1),
+      ∃ (c : ℝ) (hc : c > 0), Infinite {n : ℕ | M z n > n ^ c}) ↔ answer(True) := by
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/119.lean
```diff
@@ -0,0 +1,85 @@
...
+n ≥ 1 let pₙ(z) = ∏_{i ≤ n} (z - zᵢ).
+
+Let Mₙ = max_{|z| = 1} |pₙ(z)|.
+
+1. Is it true that lim sup Mₙ = ∞?
+
+2. Is it true that there exists c > 0 such that for infinitely many n we have Mₙ > nᶜ?
+
+3. Is it true that there exists c > 0 such that, for all large n, ∑_{k ≤ n} Mₖ > n^{1 + c}?
+-/
```
mo271: Please put these docstrings to the definitions/theorem they belong. Please use latex markdown.

FormalConjectures/ErdosProblems/119.lean
```diff
@@ -0,0 +1,85 @@
...
+
+/-- Definition of the set ℕ_{≥ 1}: -/
+def N1 : Type := {n : ℕ // n ≥ 1}
```
mo271:
```suggestion
```
I'm not entirely sure if indexing in this problem should start at 0 or one, but in any case `ℕ+` exists and doesn't need redefinition, if that is what you want to use (I'd propose rather using `ℕ` if possible).

FormalConjectures/ErdosProblems/119.lean
```diff
@@ -0,0 +1,85 @@
...
+/-- Definition of the unit sequence: -/
+variable (z : N1 → ℂ) (hz : ∀ i : N1, ‖z i‖ = 1)
+
+/-- Definition of the polynomial pₙ(z): -/
+noncomputable def p (n : N1) : ℂ → ℂ :=
+    fun w => ∏ i : range (n.val), (w - z ⟨i + 1, by linarith⟩)
+
+/-- Definition of Mₙ: -/
+noncomputable def M (n : N1) : ℝ :=
+    sSup {x | ∃ (w : ℂ), ‖w‖ = 1 ∧ x = ‖p z n w‖}
```
mo271:
```suggestion
    sSup { (‖p z n w‖) | (w : ℂ) (_ : ‖w‖ = 1)}
```
Slightly shorter and avoids giving a name to the absolute value.

FormalConjectures/ErdosProblems/119.lean
```diff
@@ -0,0 +1,85 @@
...
+
+/-- Question 1:
+
+Wagner [Wa80] proved that there is some c > 0 with Mₙ > (log n)ᶜ infintely often.
+
+[Wa80] Wagner, Gerold, On a problem of {E}rdős in {D}iophantine approximation. Bull. London Math. Soc. (1980), 81--88.
+-/
+@[category research closed, AMS 30]
```
mo271:
```suggestion
@[category research solved, AMS 30]
```

FormalConjectures/ErdosProblems/119.lean
```diff
@@ -0,0 +1,85 @@
...
+theorem erdos_119_1 :
+    ∀ (R : ℝ), ∃ n, M z n ≥ R ↔ answer(True) := by
```
mo271: seems like `atTop.limsup (fun n => (M z n : EReal)) = ⊤` would be closer to the original formulation? Might be equivalent..

FormalConjectures/ErdosProblems/119.lean
```diff
@@ -0,0 +1,85 @@
...
+
+/-- Question 2:
+
+Beck [Be91] proved that there exists some c > 0 such that max_{n ≤ N} Mₙ > Nᶜ.
+
+[Be91] Beck, J., The modulus of polynomials with zeros on the unit circle: A problem of Erdős. Annals of Math. (1991), 609-651.
+-/
+@[category research closed, AMS 30]
+theorem erdos_119_2 :
+    ∃ (c : ℝ) (hc : c > 0), ∀ (N : ℕ), ∃ n, n ≥ N ∧ M z ⟨n + 1, by linarith⟩ > n ^ c ↔ answer(True) := by
```
mo271:
```suggestion
    (∃ (c : ℝ) (hc : c > 0), ∀ (N : ℕ), ∃ n, n ≥ N ∧ M z ⟨n + 1, by linarith⟩ > n ^ c) ↔ answer(True) := by
```
One could rephrase equivalently that the set where the inequality is satisfied is infinite to be closer to the original formulation, but I suppose that is a question of taste.

FormalConjectures/ErdosProblems/119.lean
```diff
@@ -0,0 +1,85 @@
...
+  sorry
+
+/-- Question 3: -/
+@[category research open, AMS 30]
+theorem erdos_119_3 :
+    ∃ (c : ℝ) (hc : c > 0), ∃ (N : ℕ), ∀ n, n ≥ N →
+      ∑ k : range n, M z ⟨k + 1, by linarith⟩ > n ^ (1 + c) ↔ answer(sorry) := by
```
mo271:
```suggestion
    (∃ (c : ℝ) (hc : c > 0), ∃ (N : ℕ), ∀ n, n ≥ N →
      ∑ k : range n, M z ⟨k + 1, by linarith⟩ > n ^ (1 + c)) ↔ answer(sorry) := by
```
use `∀ᶠ n in atTop` to translate the informal "for all large $n$"

porianes:
Thanks a lot for the comments! I didn't realize there is a `ℕ+` and used `PNat` instead in the new version. I used this but not `ℕ` because in the statement of this problem it asks for all `i ≥ 1` and for `n ≥ 1`. I also fixed other suggestions mentioned.

mo271:
Could you mark those comments that are resolved by clicking  "Resolve conversation"?

Why not using for `ℕ+` notation for `PNat`? 

I don't think the meaning of the problem changes when one uses `ℕ` instead and drop the conditions $i\ge 1$ and $n\ge 1$, but it does become more simple. Perhaps you could do that while adding a comment saying "we use 0-indexing here, while in the original problem formulation 1-indexing was used". A reason do to so is that many more theorems are available for `ℕ` compared to `ℕ+` so likely it is already easier to (dis)prove the conjecture when using that (although it should also be quite simple to translate between them in a proof...)

FormalConjectures/ErdosProblems/119.lean
```diff
@@ -0,0 +1,76 @@
...
+*Reference:* [erdosproblems.com/119](https://www.erdosproblems.com/119)
+-/
+
+open Filter Finset Set
+
+namespace Erdos119
+
+/-- Let $z_i$ be an infinite sequence of complex numbers such that $|z_i| = 1$ for all $i \geq 1$. -/
+variable (z : PNat → ℂ) (hz : ∀ i : PNat, ‖z i‖ = 1)
+
```
mo271, line 34: this doesn't work because `/-- ...-/` expects a lemma or theorem. One could change it to `/- ... -/`. 
Consider also to putting the definition for those directly in the theorems/definitions?

porianes:
You're totally right with the indexing, I fixed it and added a comment at the top. I also put the "definition" of the sequence z_i into each definition/theorem below.

I thought `ℕ+` and `PNat` are equivalent. What is the difference between these two?

mo271:
Thanks again!

PR #1226, opened 2025-11-19
add Erdősproblem 170

### LINKED ISSUE #401: Erdős Problem 170: The Sparse Ruler problem
### What is the conjecture
    
https://www.erdosproblems.com/170

  Let $F(N)$ be the smallest possible size of $A\subset \\{0,1,\ldots,N\\}$ such that $\\{0,1,\ldots,N\\}\subset A-A$. Find the value of 
$$\lim_{N\to \infty}\frac{F(N)}{N^{1/2}}.$$

Status: open

see also: https://en.wikipedia.org/wiki/Sparse_ruler

### Choose either option

eladyn:
Fixes #401.

This adds the sparse ruler problem to the list. In theory, this is also on Wikipedia, but I only formalized the Erdős variant of the problem, so I thought that it would be best suited in `erdosproblems/`. I'm happy to change things about that, however!

### DIFF
--- a/FormalConjectures/ErdosProblems/170.lean
+++ b/FormalConjectures/ErdosProblems/170.lean
@@ -26,21 +26,23 @@ open scoped Topology
 
 namespace Erdos170
 
-/-- An `N`-perfect ruler is a subset `A ⊆ ℕ` (the marks), such that each positive integer
-`k ≤ N` can be measured, that is, expressed as a difference `k = a₁ - a₀` with `a₁, a₂ ∈ A`. -/
+/-- An $N$-perfect ruler is a finite subset $A \subseteq \mathbb{N}$ (the marks), such that each
+positive integer $k \leq N$ can be measured, that is, expressed as a difference $k = a_1 - a_0$
+with $a_0, a_1 \in A$. The set $A$ is then also called a difference basis w.r.t. $N$. -/
 @[reducible]
 def PerfectRuler (N : ℕ) (A : Finset ℕ) : Prop :=
   ∀ k ∈ Finset.range (N + 1), ∃ᵉ (a₀ ∈ A) (a₁ ∈ A), k = a₁ - a₀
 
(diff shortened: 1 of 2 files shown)


FormalConjectures/ErdosProblems/170.lean
```diff
@@ -0,0 +1,57 @@
...
+    exact ⟨by rfl, fun k hk => ⟨0, by simp, k, hk, rfl⟩⟩
+
+/-- We define a function `F N` as the minimum cardinality of an `N`-perfect ruler of length `N`. -/
+def F (N : ℕ) : ℕ :=
+    Finset.min' (Finset.image Finset.card (PerfectRulersLengthN N))
+                (Finset.image_nonempty.mpr ⟨TrivialRuler N, trivial_ruler_is_perfect N⟩)
+
+/-- The problem is to determine the limit of the sequence `(F N) / √N` as `N → ∞`. -/
+@[category research open, AMS 05]
+lemma erdos170 : Filter.Tendsto (fun N => F N / √N) Filter.atTop (𝓝 answer(sorry)) := sorry
```
mo271:
```suggestion
lemma erdos170 : Filter.Tendsto (fun N => F N / √N) Filter.atTop (𝓝 answer(sorry)) := by sorry

end Erdos170
```

mo271 (commented):
Thanks for the contribution very nice. mathematically sound, only a few remarks:
 
 - Please use latex markdown for all comments
 - Please add a TODO for the formalization of the statements below the main box (or formalize them as well)
 - consider adding a file FormalConjectures/Wikipedia/SparseRuler, which has a line
`import FormalConjectures.ErdosProblems.«170»` and a docstring describing the problem. Then later other statements from the wikipedia page https://en.wikipedia.org/wiki/Sparse_ruler can be added.

mo271 (approved):
Thanks again!

PR #1231, opened 2025-11-19
feat: add Erdős Problem 508

### LINKED ISSUE #761: Erdős Problem 508: Hadwiger-Nelson problem
### What is the conjecture
    
https://www.erdosproblems.com/508

  What is the chromatic number of the plane? That is, what is the smallest number of colours required to colour $\mathbb{R}^2$ such that no two points of the same colour are distance $1$ apart?
  
The [Hadwiger-Nelson problem](https://en.wikipedia.org/wiki/Hadwiger%E2%80%93Nelson_problem), see also #314 

Status: open

juli5n:
fixes #761

I'm not very experienced with lean, so maybe you should consider reviewing this contribution a little more carefully. Also note that I haven't managed to build the documentation yet, so the format of the doc comments could be a little messed up, sorry about that!!

I am very busy over the next few weeks, but maybe I could also prove the lower bound 4 for the chromatic number of the plane using the [Moser-Spindel graph](https://de.wikipedia.org/wiki/Moser-Spindel)  if there is any interest, but I can't promise anything!! 

Also note that I am not sure about whether I've placed the labels undergraduate vs research correctly.

Thank you for reviewing this!

Best regards,
Julian

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -7,6 +7,7 @@
 Google LLC
 
 # Please keep the list sorted alphabetically
+Abel Doñate
 Anirudh Rao
 Anthony Wang
 Ayush Debnath
@@ -14,6 +15,7 @@ Bhavik Mehta
 Calle Sönne
 Eric Wieser
 James Jordan
+Jofre Costa
 Junseok Lee
 Junyan Xu
 Moritz Firsching
--- a/FormalConjectures/ErdosProblems/1003.lean
+++ b/FormalConjectures/ErdosProblems/1003.lean
@@ -27,7 +27,7 @@ namespace Erdos1003
 open scoped Nat
 
 /--
-Are there infinitely many solutions to $\phi(n) = \phi(n+1)$, where \phi is the Euler totient
+Are there infinitely many solutions to $\phi(n) = \phi(n+1)$, where $\phi$ is the Euler totient
 function?
 -/
 @[category research open, AMS 11]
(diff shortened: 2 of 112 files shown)


mo271:

This problem is already in formal-conjectures, see #314 and #315. 
When and Erdős problem is also a wikipedia problem, we usually put it in one place and then just import the file in the other place to make sure to have it there. 

In your formalization you also provide the proof `3 ≤ χ(ℝ²)` which is nice, if we can shorten it a few lines. 
We don't want to keep long proofs in this repo, because of the implied maintenance burden. 
But perhaps if the Moser spindle proof can be shortened significantly. 

To summarize, I propose to perhaps move the stuff from https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/Wikipedia/HadwigerNelson.lean ot here and then 
`import FormalConjectures.ErdosProblems.«508»` in HadwigerNelson.lean or vice versa, potentially adding  `3 ≤ χ(ℝ²)` in the form 
```
@[category research solved, AMS 52]
theorem HadwigerNelsonAtLeastThree :
    3 ≤ UnitDistancePlaneGraph.chromaticNumber := by
   ...
   
```

juli5n:
Hello @mo271 ,

I've incorporated the changes you suggested (i.e. moving the code from the Wikipedia/HadwigerNelson.lean file to ErdosProblems/508.lean).

Best regards,
Julian

Issue #874, opened 2025-10-07
Erdős Probelm 680


mo271:
### What is the conjecture
    
https://www.erdosproblems.com/680

  Is it true that, for all sufficiently large $n$, there exists some $k$ such that 
$$p(n+k)>k^2+1,$$
where $p(m)$ denotes the least prime factor of $m$?

Can one prove this is false if we replace $k^2+1$ by $e^{(1+\epsilon)\sqrt{k}}+C_\epsilon$, for all $\epsilon>0$, where $C_\epsilon>0$ is some constant?
  

Status:  open

PR #1233, opened 2025-11-19
feat: add Erdős Problem 680

triangular2357:
Adds formalization of Erdős Problem 680

Erdős Problem 680: 
 Is it true that, for all sufficiently large $n$, there exists some $k$ such that
$$p(n+k)>k^2+1,$$
where $p(m)$ denotes the least prime factor of $m$?
Can one prove this is false if we replace $k^2+1$ by $e^{(1+\epsilon)\sqrt{k}}+C_\epsilon$, for all $\epsilon>0$, where $C_\epsilon>0$ is some constant?
see [https://www.erdosproblems.com/680](https://www.erdosproblems.com/680)

Closes https://github.com/google-deepmind/formal-conjectures/issues/874

### DIFF
--- a/FormalConjectures/ErdosProblems/680.lean
+++ b/FormalConjectures/ErdosProblems/680.lean
@@ -36,7 +36,7 @@ where $p(m)$ denotes the least prime factor of $m$?
 -/
 @[category research open, AMS 11]
 theorem erdos_680 :
-  (∀ᶠ (n : ℕ) in Filter.atTop, ∃ k, Nat.minFac (n + k) > k^2 + 1)
+  (∀ᶠ (n : ℕ) in Filter.atTop, ∃ k ≠ 0, Nat.minFac (n + k) > k^2 + 1)
   ↔ answer(sorry) := sorry
 
 /--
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/680.lean
```diff
@@ -0,0 +1,49 @@
...
+
+/--
+Is it true that, for all sufficiently large $n$, there exists some $k$ such that
+\[
+p(n+k)>k^2+1,
+\]
+where $p(m)$ denotes the least prime factor of $m$?
+-/
+@[category research open, AMS 11]
+theorem erdos_680 : ∃ N, ∀ n ≥ N, ∃ k, Nat.minFac (n + k) > k^2 + 1 := sorry
```
Paul-Lez: For problems stated as questions, our convention is to use `answer(sorry)` to convey that people don't seem to have an expected answer (true or false) in mind yet.
```suggestion
theorem erdos_680 : (∃ N, ∀ n ≥ N, ∃ k, Nat.minFac (n + k) > k^2 + 1) ↔ answer(sorry) := sorry
```

Paul-Lez: You'll want to add the condition that `k` is non-zero here otherwise the theorem holds trivially!

Paul-Lez: I'd also be tempted to formalise this using `Filter.Eventually` since that convention is used in many other files. Feel free to ask if you're not sure how to do this!

florianflorianofficial: I now used `Filter.Eventually` to restate the conjectures. I hope I did it correctly, feel free to correct me, if I did it wrong.

florianflorianofficial: It should be fixed now.

PR #1235, opened 2025-11-19
add erdős problem 677

### LINKED ISSUE #872: Erdős Problem 677
### What is the conjecture
    
https://www.erdosproblems.com/677

  Let $M(n,k)=[n+1,\ldots,n+k]$ be the least common multiple of $\\{n+1,\ldots,n+k\\}$.

Is it true that for all $m\geq n+k$
$$M(n,k) \neq M(m,k)?$$
  

Status:  open

HerrLaal:
This adds a formalization of Erdős problem 677. If you'd like me to change something or to add for example the Thue-Siegel theorem, I'm happy to do so :)

fixes #872

### DIFF
--- a/FormalConjectures/ErdosProblems/678.lean
+++ b/FormalConjectures/ErdosProblems/678.lean
@@ -56,7 +56,7 @@ Cambie [Ca24] found the example $M(36, 8) > M(48, 9)$.
 lemma lcmInterval_lt_example4 : lcmInterval 47 9 < lcmInterval 36 8 := by decide
 
 /--
-Write $M(n, k)$ be the least common multiple of ${n+1, \dotsc, n+k}$.
+Write $M(n, k)$ be the least common multiple of $\{n+1, \dotsc, n+k\}$.
 Let $k$ be sufficiently large. Are there infinitely many $m, n$ with $m \geq n + k$ such that
 $$
 M(n, k) > M(m, k + 1)
--- a/AUTHORS
+++ b/AUTHORS
@@ -7,6 +7,7 @@
 Google LLC
 
 # Please keep the list sorted alphabetically
+Abel Doñate
 Anirudh Rao
 Anthony Wang
 Ayush Debnath
@@ -14,6 +15,7 @@ Bhavik Mehta
 Calle Sönne
 Eric Wieser
 James Jordan
+Jofre Costa
 Junseok Lee
 Junyan Xu
 Moritz Firsching
--- a/FormalConjectures/ErdosProblems/1003.lean
+++ b/FormalConjectures/ErdosProblems/1003.lean
@@ -27,7 +27,7 @@ namespace Erdos1003
 open scoped Nat
 
 /--
-Are there infinitely many solutions to $\phi(n) = \phi(n+1)$, where \phi is the Euler totient
+Are there infinitely many solutions to $\phi(n) = \phi(n+1)$, where $\phi$ is the Euler totient
 function?
 -/
 @[category research open, AMS 11]
--- a/FormalConjectures/ErdosProblems/1041.lean
+++ b/FormalConjectures/ErdosProblems/1041.lean
@@ -1,0 +1,69 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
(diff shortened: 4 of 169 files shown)


FormalConjectures/ErdosProblems/677.lean
```diff
@@ -0,0 +1,53 @@
...
+
+/--
+Write $M(n, k)$ be the least common multiple of ${n+1, \dotsc, n+k}$, denoted here as
+`lcmInterval n k`.
+-/
+def lcmInterval (n k : ℕ) : ℕ := (Finset.Ioc n (n + k)).lcm id
+
+/-
+Erdős expected very few solutions for M(n,k)=M(m,l), where m ≥ n+k and l > 1. The only solutions he
+ knew were the followings.
```
callesonne:
```suggestion
 knew were the following.
```

FormalConjectures/ErdosProblems/677.lean
```diff
@@ -0,0 +1,53 @@
...
+
+/-
+We use the same definition as is Erdős Problem 678.
+-/
+
+/--
+Write $M(n, k)$ be the least common multiple of ${n+1, \dotsc, n+k}$, denoted here as
+`lcmInterval n k`.
+-/
+def lcmInterval (n k : ℕ) : ℕ := (Finset.Ioc n (n + k)).lcm id
```
callesonne: Since you are reusing a definition from `678.lean` I suggest you create a new file in `ForMathlib` containing this definition. Possibly it could be put in a new file `Data/Nat/Interval.lean`. A bonus (but optional) would be if you wanted to try to find the minimal assumptions so that this definition makes sense (it should make sense outside just the naturals). For example `Finset.lcm` that you are using is defined in any `NormalizedGCDMonoid`.

FormalConjectures/ErdosProblems/677.lean
```diff
@@ -0,0 +1,53 @@
...
+lemma lcmInterval_eq_example2 : lcmInterval 3 4 = lcmInterval 19 2 := by decide
+
+/--
+Denote by M(n,k) the least common multiple of the finite set {n + 1, ...,  n + k}.
+Is it true that for all m ≥ n + k, we get M(m, k) ≠ M(n, k)?
+-/
+@[category research open, AMS 11]
+theorem erdos_677 :
+    ∀ m n k, m ≥ n + k → lcmInterval m k ≠ lcmInterval n k := by
+  sorry
```
callesonne, line 43: Here you need to be careful with your assumptions. I suspect that at least you need to assume `k > 0` (even though `Finset.Ico n n` is empty, the lcm will have the default value `1`).

callesonne:
And thanks for the contribution! :)

FormalConjectures/ErdosProblems/677.lean
```diff
@@ -0,0 +1,48 @@
...
+/-!
+# Erdős Problem 677
+*Reference:* [erdosproblems.com/677](https://www.erdosproblems.com/677)
+-/
+
+namespace Erdos677
+
+open Finset
+
+/- Denote by $M(n, k)$ the least common multiple of ${n+1, \dotsc, n+k}$. -/
```
callesonne: This comment is a bit out of place, Could you just copy paste it as the starting line to each of the two comments below?

FormalConjectures/ErdosProblems/677.lean
```diff
@@ -0,0 +1,48 @@
...
+
+/-
+Erdős expected very few solutions for M(n,k)=M(m,l), where m ≥ n+k and l > 1. The only solutions he
+ knew were the following.
+-/
```
callesonne: I would either remove this, or make it a docstring to each of the examples below (we generally use very few non-docstring comments in formal conjectures).

FormalConjectures/ErdosProblems/678.lean
```diff
@@ -25,11 +25,7 @@ open Asymptotics Filter Finset
 
 namespace Erdos678
 
-/--
-Write $M(n, k)$ be the least common multiple of ${n+1, \dotsc, n+k}$, denoted here as
-`lcmInterval n k`.
--/
-def lcmInterval (n k : ℕ) : ℕ := (Finset.Ioc n (n + k)).lcm id
+/- Denote by $M(n, k)$ the least common multiple of ${n+1, \dotsc, n+k}$. -/
```
callesonne: Same as above regarding this comment.

FormalConjectures/ErdosProblems/677.lean
```diff
@@ -0,0 +1,48 @@
...
+lemma lcmInterval_eq_example2 : lcmInterval 3 4 = lcmInterval 19 2 := by decide
+
+/--
+Denote by M(n,k) the least common multiple of the finite set {n + 1, ...,  n + k}.
+Is it true that for all m ≥ n + k, we get M(m, k) ≠ M(n, k)?
+-/
+@[category research open, AMS 11]
+theorem erdos_677 :
+    ∀ (m n k : ℕ), k > 0 → m ≥ n + k → lcmInterval m k ≠ lcmInterval n k := by
+  sorry
```
callesonne, line 43: Could you add a TODO below here asking to formalize the other statements on the webpage (and of course you are very welcome to formalize any of them as well if you want)?

PR #1236, opened 2025-11-19
feat(ErdosProblems): 123

### LINKED ISSUE #363: Erdős Problem 123
### What is the conjecture
    
https://www.erdosproblems.com/123

  Let $a,b,c$ be three integers which are pairwise coprime. Is every large integer the sum of distinct integers of the form $a^kb^lc^m$ ($k,l,m\geq 0$), none of which divide any other?

Status: open

### Prerequisites needed
None

See also #364.

### Choose either option

jganten:
I am not sure if the proven variants are to be included, but here we go :)

fixes #363

### DIFF
--- a/FormalConjectures/ErdosProblems/123.lean
+++ b/FormalConjectures/ErdosProblems/123.lean
@@ -13,7 +13,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 -/
-
 import FormalConjectures.Util.ProblemImports
 
 /-!
@@ -48,8 +47,8 @@ Characterizes a "snug" finite set of natural numbers:
 all elements are within a multiplicative factor $(1 + ε)$ of the minimum.
 Specifically, for a finite set $A$ and $ε > 0$, all $a ∈ A$ satisfy $a < (1 + ε) · min(A)$.
 -/
-def IsSnug (ε : ℝ) (A : Finset ℕ) (hA : A.Nonempty) : Prop :=
-  ∀ a ∈ A, a < (1 + ε) * A.min' hA
+def IsSnug (ε : ℝ) (A : Finset ℕ) : Prop :=
+  ∃ hA : A.Nonempty, ∀ a ∈ A, a < (1 + ε) * A.min' hA
 
 /--
 Predicate for pairwise coprimality of three integers.
@@ -72,8 +71,7 @@ to be greater than one and distinct.
 @[category research open, AMS 11]
 theorem erdos_123 (a b c : ℕ) (ha : a > 1) (hb : b > 1) (hc : c > 1)
     (h_coprime : PairwiseCoprime a b c) :
-    IsDComplete ((powers a : Set ℕ) * (powers b : Set ℕ) *
-      (powers c : Set ℕ)) ↔ answer(sorry) := by sorry
+    IsDComplete (↑(powers a) * ↑(powers b) * ↑(powers c)) ↔ answer(sorry) := by sorry
 
 /--
 Erdős and Lewin proved this conjecture when $a = 3$, $b = 5$, and $c = 7$.
@@ -83,7 +81,7 @@ _$d$-complete sequences of integers_. Math. Comp. (1996), 837-840.
 -/
 @[category research solved, AMS 11]
 theorem erdos_123.variants.erdos_lewin_3_5_7 :
-    IsDComplete ((powers 3 : Set ℕ) * (powers 5 : Set ℕ) * (powers 7 : Set ℕ)) := by sorry
+    IsDComplete (↑(powers 3) * ↑(powers 5) * ↑(powers 7)) := by sorry
 
 /--
 A simpler case: the set of numbers of the form $2^k 3^l$ ($k, l ≥ 0$) is d-complete.
@@ -98,8 +96,7 @@ Reference: [Er92b] Erdős, Paul, _Some of my favourite problems in various branc
 of combinatorics_. Matematiche (Catania) (1992), 231-240.
 -/
 @[category research solved, AMS 11]
-theorem erdos_123.variants.powers_2_3 :
-    IsDComplete ((powers 2 : Set ℕ) * (powers 3 : Set ℕ)) := by sorry
+theorem erdos_123.variants.powers_2_3 : IsDComplete (↑(powers 2) * ↑(powers 3)) := by sorry
 
 /--
 A stronger conjecture for numbers of the form $2^k 3^l 5^j$.
@@ -109,10 +106,8 @@ $b_1 < ... < b_t$ of the form $2^k 3^l 5^j$ where $b_t < (1 + ϵ) b_1$.
 -/
 @[category research open, AMS 11]
 theorem erdos_123.variants.powers_2_3_5_snug :
-  (∀ ε > 0, ∀ᶠ (n : ℕ) in atTop,
-    ∃ (A : Finset ℕ) (hA : (A : Set ℕ) ⊆ (powers 2 : Set ℕ) *
-      (powers 3 : Set ℕ) * (powers 5 : Set ℕ))
-      (hAnonempty : A.Nonempty) (hAsnug : IsSnug ε A hAnonempty),
-    A.sum id = n) ↔ answer(sorry) := by sorry
+    (∀ ε > 0, ∀ᶠ n in atTop,
+      ∃ A : Finset ℕ, (A : Set ℕ) ⊆ ↑(powers 2) * ↑(powers 3) * ↑(powers 5) ∧ IsSnug ε A ∧
+        ∑ x ∈ A, x = n) ↔ answer(sorry) := by sorry
 
 end Erdos123


FormalConjectures/ErdosProblems/123.lean
```diff
@@ -0,0 +1,113 @@
...
+This represents numbers whose prime factors are restricted to those of $p$, $q$, and $r$.
+-/
+def PowersOfThree (p q r : ℕ) : Set ℕ :=
+  {n : ℕ | ∃ a b c : ℕ, n = p^a * q^b * r^c}
+
+/--
+Given three natural numbers `a`, `b`, `c`, this is the set of all natural numbers
+of the form $a^k b^l c^m$ where $k, l, m ≥ 0$.
+-/
+def powersOfThree (a b c : ℕ) : Set ℕ :=
```
Paul-Lez: Isn't this the same as the previous definition?

FormalConjectures/ErdosProblems/123.lean
```diff
@@ -0,0 +1,113 @@
...
+/--
+The set of all natural numbers of the form $p^a q^b$ where $a, b ≥ 0$.
+This represents numbers whose prime factors are restricted to those of $p$ and $q$.
+-/
+def PowersOfTwo (p q : ℕ) : Set ℕ :=
+  {n : ℕ | ∃ a b : ℕ, n = p^a * q^b}
+
+/--
+The set of all natural numbers of the form $p^a q^b r^c$ where $a, b, c ≥ 0$.
+This represents numbers whose prime factors are restricted to those of $p$, $q$, and $r$.
```
Paul-Lez: Note that the set you've defined isn't always the same as this set, since you need to account for multiplicities. For example, taking `(p, q, r) = (2^2, 3^2, 5^2)` should give you a counterexample.

FormalConjectures/ErdosProblems/123.lean
```diff
@@ -0,0 +1,113 @@
...
+  inductive hypothesis to $n - 3^k$ (which is even).
+
+Reference: [Er92b] Erdős, Paul, _Some of my favourite problems in various branches
+of combinatorics_. Matematiche (Catania) (1992), 231-240.
+-/
+@[category research solved, AMS 11]
+theorem erdos_123.variants.powers_2_3 :
+    IsDComplete (PowersOfTwo 2 3) := by
+  sorry
+
```
Paul-Lez, line 100: (you're more than welcome to add the extra problem if you want to, but no pressure!)
```suggestion

-- TODO(Paul-Lez): add the last problem from the webpage. 

```

FormalConjectures/ErdosProblems/123.lean
```diff
@@ -0,0 +1,113 @@
...
+of the form $a^k b^l c^m$ where $k, l, m ≥ 0$.
+-/
+def powersOfThree (a b c : ℕ) : Set ℕ :=
+  {n | ∃ k l m : ℕ, n = a^k * b^l * c^m}
+
+/--
+Helper predicate for pairwise coprimality of three integers.
+-/
+def PairwiseCoprime (a b c : ℕ) : Prop :=
+  Nat.Coprime a b ∧ Nat.Coprime b c ∧ Nat.Coprime a c
```
Paul-Lez: I'd be tempted to write this as (untested)
```suggestion
  Set.Pairwise {a, b, c} Nat.Coprime
```

Paul-Lez:
> I am not sure if the proven variants are to be included, but here we go :)
> 
> fixes #363

Proved variants are very welcome!

FormalConjectures/ErdosProblems/123.lean
jganten: yes!

FormalConjectures/ErdosProblems/123.lean
jganten: Okay, I corrected this to our powers being a subset of the numbers whose prime factors are restricted to those of $p$, $q$, and $r$.

Should I maybe just omit this statement since its a little unnecessary?

FormalConjectures/ErdosProblems/123.lean
jganten: will look into it!

FormalConjectures/ErdosProblems/123.lean
jganten: So trying to prove the equivalence, I realized using `Set.Pairwise` introduces subtle changes to the property. 
To sum it up, using `Set.Pairwise` also allows for `a=b`, which is technically allowed by our definition. 
This sums it up:
```lean4
def PairwiseCoprime (a b c : ℕ) : Prop :=
  Nat.Coprime a b ∧ Nat.Coprime b c ∧ Nat.Coprime a c

def testPairwiseCoprime (a b c : ℕ) : Prop := Set.Pairwise {a, b, c} Nat.Coprime

example (a b : ℕ) (hab: (a=b)): Set.Pairwise {a, b} Nat.Coprime:= by
  simp [Set.Pairwise]
  tauto

example (a b : ℕ) (hab: (a=b)): a.Coprime b ↔ a=1 := by
  rw [hab]
  exact Nat.coprime_self b

example (a b : ℕ) (hab: ¬(a=b)): (¬(a=1 ∧ b=1) ∧ (a.Coprime b ∧ b.Coprime a) ) ↔ Set.Pairwise {a, b} Nat.Coprime:= by
  simp [Set.pairwise_insert, Nat.Coprime]
  grind
```

So I think its a more elegant way to put it, but it might make a standard form a little harder to get.

FormalConjectures/ErdosProblems/123.lean
jganten: just committed the last problem

FormalConjectures/ErdosProblems/123.lean
```diff
@@ -0,0 +1,119 @@
...
+**Erdős Problem #123**
+
+Let $a, b, c$ be three integers which are pairwise coprime. Is every large integer
+the sum of distinct integers of the form $a^k b^l c^m$ ($k, l, m ≥ 0$), none of which
+divide any other?
+
+Equivalently: is the set $\{a^k b^l c^m : k, l, m \geq 0\}$ d-complete?
+-/
+@[category research open, AMS 11]
+theorem erdos_123 (a b c : ℕ) (h_coprime : PairwiseCoprime a b c) :
```
smmercuri: I think we need to assert that `a`, `b`, and `c` are pairwise distinct here, otherwise this reduces to the corresponding question with two pairwise coprime integers. This was the original setting which motivated this Conjecture in the [source paper](https://scispace.com/pdf/on-some-sequences-of-integers-15llhntro7.pdf) of Erdős and Lewin.

Similarly, if any of `a`, `b`, `c` are `1` then this also reduces to the two-integer case, so I think the implicit assumption they they are all `>1` is present in the paper.

jganten: You are right, in these cases the conjecture would not make a lot of sense, because in the two-integer case the conjecture only holds for `2`,`3`.

FormalConjectures/ErdosProblems/123.lean
```diff
@@ -0,0 +1,124 @@
...
+    (s : Set ℕ) ⊆ A ∧                  -- The summands come from A
+    IsAntichain (· ∣ ·) (s : Set ℕ) ∧   -- No summand divides another
+    s.sum id = n                        -- They sum to n
+
+/--
+The set of all natural numbers of the form $p^a q^b$ where $a, b ≥ 0$.
+This is a subset of numbers whose prime factors are restricted to those of $p$ and $q$.
+-/
+def PowersOfTwo (p q : ℕ) : Set ℕ :=
+  {n : ℕ | ∃ a b : ℕ, n = p^a * q^b}
```
YaelDillies: May I complain that "powers of two" doesn't mean "p^a * q^b"? 😄 Maybe
```suggestion
def TwoPrimePowers (p q : ℕ) : Set ℕ :=
  {n : ℕ | ∃ a b : ℕ, n = p^a * q^b}
```

FormalConjectures/ErdosProblems/123.lean
```diff
@@ -0,0 +1,124 @@
...
+
+/--
+The set of all natural numbers of the form $p^a q^b r^c$ where $a, b, c ≥ 0$.
+This is a subset of numbers whose prime factors are restricted to those of $p$, $q$, and $r$.
+-/
+def PowersOfThree (p q r : ℕ) : Set ℕ :=
+  {n : ℕ | ∃ a b c : ℕ, n = p^a * q^b * r^c}
+
+/--
+The subsets of powers of three which satisfy the criterion of being close to its smallest element. Namely for an arbitrary $ε > 0$ a finite subset $A$ of PowersOfThree is snug if for all $a ∈ A$, we have $a < (1 + ε) * min(A)$.
```
YaelDillies: Can you break the long line?

FormalConjectures/ErdosProblems/123.lean
```diff
@@ -0,0 +1,124 @@
...
+-/
+def IsSnug (ε : ℝ) (A : Finset ℕ) (hA : A.Nonempty) : Prop :=
+  ∀ a ∈ A, a < (1 + ε) * A.min' hA
+
+/--
+Helper predicate for pairwise coprimality of three integers.
+-/
+def PairwiseCoprime (a b c : ℕ) : Prop := Set.Pairwise {a, b, c} Nat.Coprime
```
YaelDillies: A more faithful alternative to Paul's suggestion would be
```suggestion
def PairwiseCoprime (a b c : ℕ) : Prop := Pairwise (Nat.Coprime on ![a, b, c])
```

FormalConjectures/ErdosProblems/123.lean
```diff
@@ -0,0 +1,124 @@
...
+the sum of distinct integers of the form $a^k b^l c^m$ ($k, l, m ≥ 0$), none of which
+divide any other?
+
+Equivalently: is the set $\{a^k b^l c^m : k, l, m \geq 0\}$ d-complete?
+
+Note: For this not to reduce to the two-integer case, we need the integers
+to be greater than one and distinct.
+-/
+@[category research open, AMS 11]
+theorem erdos_123 (a b c : ℕ) (h_distinct : a ≠ b ∧ b ≠ c ∧ a ≠ c)
```
YaelDillies: ... then this condition will become redundant
```suggestion
theorem erdos_123 (a b c : ℕ)
```

FormalConjectures/ErdosProblems/123.lean
```diff
@@ -0,0 +1,124 @@
...
+    (h_gt_one : 1 < a ∧ 1 < b ∧ 1 < c) (h_coprime : PairwiseCoprime a b c) :
```
YaelDillies: Please avoid assuming an `And`. Assume each constituent part separately

FormalConjectures/ErdosProblems/123.lean
```diff
@@ -0,0 +1,124 @@
...
+  sorry
+
+/--
+A stronger conjecture for numbers of the form $2^k 3^l 5^j$.
+
+For any $ε > 0$, all large integers $n$ can be written as the sum of distinct integers $b_1 < ... < b_t$ of the form $2^k 3^l 5^j$ where $b_t < (1 + ϵ) b_1$.
+-/
+@[category research open, AMS 11]
+theorem erdos_123.variants.powers_2_3_5_snug :
+  (∀ ε > 0, ∀ᶠ (n : ℕ) in atTop, ∃ (A : Finset ℕ) (hA : (A : Set ℕ) ⊆ PowersOfThree 2 3 5) (hAnonempty : A.Nonempty) (hAsnug: IsSnug ε A hAnonempty), A.sum id = n) ↔ answer(sorry) := by sorry
```
YaelDillies: Long line again

FormalConjectures/ErdosProblems/123.lean
YaelDillies: The definition doesn't actually mention `PowersOfThree` anywhere. Could you clarify the docstring?

FormalConjectures/ErdosProblems/123.lean
jganten: Two prime powers seems also misleading to me, since p and q do not have to be prime, just coprime (which is not part of the definition but how it is used).
I read this as 'PowersOfTwo natural numbers'. I struggle to find a more descriptive name that is less wrong, but of course I am open to other suggestions.

FormalConjectures/ErdosProblems/123.lean
jganten: Thank you! This makes it a lot nicer - though the `on` notation did not work, so I used `Function.onFun`.

FormalConjectures/ErdosProblems/123.lean
YaelDillies: Yeah, I see (although then I will complain that maybe you should be using `a` and `b` instead of `p` and `q`!). I have another suggestion: Don't write a new definition at all, but instead use `Submonoid.powers` (aka `powers` if you `open Submonoid`) and pointwise multiplicative of sets (available with `open scoped Pointwise`), ie your `PowersOfTwo a b` could become `powers a * powers b` and `PowersOfThree a b c` could become `powers a * powers b * powers c` (possibly with some `Set ℕ` annotations). At this point, I would suggest dropping the intermediate definitions.

FormalConjectures/ErdosProblems/123.lean
YaelDillies: Sorry, the `on` notation is now hidden in the `Function` scope. You can do `open scoped Function` to access it

PR #1241, opened 2025-11-20
fix: positivity in `358.lean`

smmercuri:
`u` and `v` should be positive integers, since adding zero will inflate the number of additive representations

### DIFF
--- a/FormalConjectures/ErdosProblems/358.lean
+++ b/FormalConjectures/ErdosProblems/358.lean
@@ -31,7 +31,7 @@ Let $a$ be an infinite sequence of integers. `intervalRepresentations A n` is th
 to \[n=\sum_{u\leq i\leq v}a_i.\] where `u` and `v` are positive integers.
 -/
 def intervalRepresentations (A : ℕ → ℕ) (n : ℕ) : Set (ℕ × ℕ) :=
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/358.lean
```diff
@@ -24,14 +24,14 @@ import FormalConjectures.Util.ProblemImports
...
+open Filter Finset
 
 /-
 Let $a$ be an infinite sequence of integers. `intervalRepresentations A n` is the set of solutions
-to \[n=\sum_{u\leq i\leq v}a_i.\]
+to \[n=\sum_{u\leq i\leq v}a_i.\] where `u` and `v` are positive integers.
 -/
 def intervalRepresentations (A : ℕ → ℕ) (n : ℕ) : Set (ℕ × ℕ) :=
-  {(u, v) | n = ∑ i ∈ Finset.Icc u v, A i}
+  {(u, v) | (u : ℕ) (v : ℕ) (_ : 0 < u) (_ : 0 < v) (_ : n = ∑ i ∈ Icc u v, A i)}
```
callesonne: Why do you need to specify `(u : ℕ) (v : ℕ)` after this change? Can you not say:
`0 < u ∧ 0 < v ∧ ..`

PR #1249, opened 2025-11-21
Erdős Problem 930

### LINKED ISSUE #1024: Erdős Probelm 930
### What is the conjecture

https://www.erdosproblems.com/930

  Is it true that, for every $r$, there is a $k$ such that if $I_1,\ldots,I_r$ are disjoint intervals of consecutive integers, all of length at least $k$, then
$$\prod_{1\leq i\leq r}\prod_{m\in I_i}m$$
is not a perfect power?

Status:  open

### Choose either option

Mal-Pat:
This adds a formalization of Erdős Problem 930, along with two variants for the specific `r = 1` case. Let me know if any improvements or corrections need to be made.

Closes #1024

### DIFF
--- a/FormalConjectures/ErdosProblems/930.lean
+++ b/FormalConjectures/ErdosProblems/930.lean
@@ -59,8 +59,8 @@ def nextPrime (k : ℕ) : ℕ :=
 /--
 Let $k$, $l$, $n$ be integers such that $k \ge 3$, $l \ge 2$ and $n + k \ge p^{(k)}$,
 where $p^{(k)}$ is the least prime satisfying $p^{(k)} \ge k$.
-Then there is a prime $p \ge k$ for which $\alpha_p \not\equiv 0 \pmod{l}$,
-where $\alpha_p$ is the power of $p$ dividing $(n + 1) \ldots (n + k)$.
+Then there is a prime $p \ge k$ for which $l$ does not divide
+the multiplicity of the prime factor $p$ in $(n + 1) \ldots (n + k)$.
 
 Theorem 2 from [ErSe75].
 


Mal-Pat:
I could formalize the proof that Theorem 2 (`erdos_930.variant.consecutive_strong`) implies Theorem 1 (`erdos_930.variant.consecutive_integers`) in a future PR.

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,99 @@
...
+
+open Finset
+
+namespace Erdos930
+
+/--
+$n$ is a perfect power if there exist natural numbers $m$ and $l$
+such that $1 < m$, $1 < l$, and $m^l = n$.
+-/
+def PerfectPower (n : ℕ) : Prop :=
```
callesonne:
```suggestion
def IsPower (n : ℕ) : Prop :=
```
I prefer this name :-)

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,99 @@
...
+  ∃ m l, 1 < m ∧ 1 < l ∧ m^l = n
```
callesonne:
```suggestion
  ∃ m l, 1 < l ∧ m^l = n
```
I think `1` should be allowed to be a perfect power

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,99 @@
...
+Is it true that, for every $r$, there is a $k$ such that
+if $I_1,\ldots,I_r$ are disjoint intervals of consecutive integers,
+all of length at least $k$, then
+$$
+  \prod_{1\leq i\leq r}\prod_{m\in I_i}m
+$$
+is not a perfect power?
+-/
+@[category research open, AMS 11]
+theorem erdos_930 :
```
callesonne, line 46: I think this statement would become a bit cleaner if `I : Fin r → ℕ \times ℕ` picking out the start and end point of the interval.

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,99 @@
...
+  ∀ r, ∃ k, ∀ (I : Fin r → Finset ℕ),
+    (∀ i j : Fin r, i ≠ j → Disjoint (I i) (I j)) →
+      (∀ i : Fin r, k ≤ (I i).card ∧ ∃ a, I i = Ico a (a + (I i).card)) →
+        ¬ PerfectPower (∏ i : Fin r, ∏ m ∈ I i, m) := by
```
callesonne:
```suggestion
  (∀ r, ∃ k, ∀ (I : Fin r → Finset ℕ),
    (∀ i j : Fin r, i ≠ j → Disjoint (I i) (I j)) →
      (∀ i : Fin r, k ≤ (I i).card ∧ ∃ a, I i = Ico a (a + (I i).card)) →
        ¬ PerfectPower (∏ i : Fin r, ∏ m ∈ I i, m)) ↔ answer(sorry) := by
```
Remember to add `↔ answer(sorry)` since the problem is asking if it is true or not!

Mal-Pat:
Changed `I : Fin r → Finset ℕ` to `I : Fin r → ℕ × ℕ` in the formalized statement of `erdos_930`, along with the other suggested changes.

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,99 @@
...
+$$
+  \prod_{1\leq i\leq r}\prod_{m\in I_i}m
+$$
+is not a perfect power?
+-/
+@[category research open, AMS 11]
+theorem erdos_930 :
+    (∀ r, ∃ k, ∀ (I : Fin r → ℕ × ℕ),
+      (∀ i : Fin r, (I i).1 ≤ (I i).2 ∧ k ≤ (I i).2 - (I i).1 + 1) →
+        (∀ i j : Fin r, i ≠ j → (I i).2 < (I j).1 ∨ (I j).2 < (I i).1) →
```
callesonne:
```suggestion
        (∀ i j : Fin r, i < j → (I i).2 < (I j).1) →
```
The standard order on `Fin r` is the non-cyclic one, i.e. `0 < 1 < ... < r - 1`, so we can use this here.

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,99 @@
...
+if $I_1,\ldots,I_r$ are disjoint intervals of consecutive integers,
+all of length at least $k$, then
+$$
+  \prod_{1\leq i\leq r}\prod_{m\in I_i}m
+$$
+is not a perfect power?
+-/
+@[category research open, AMS 11]
+theorem erdos_930 :
+    (∀ r, ∃ k, ∀ (I : Fin r → ℕ × ℕ),
```
callesonne:
```suggestion
    (∀ r > 0, ∃ k, ∀ (I : Fin r → ℕ × ℕ),
```

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,99 @@
...
+      (∀ i : Fin r, (I i).1 ≤ (I i).2 ∧ k ≤ (I i).2 - (I i).1 + 1) →
```
callesonne:
```suggestion
      (∀ i : Fin r, 0 < (I i).1∧ (I i).1 ≤ (I i).2 ∧ k ≤ (I i).2 - (I i).1 + 1) →
```
Making sure we don't get `0` in our interval :)

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,99 @@
...
+        (∀ i j : Fin r, i ≠ j → (I i).2 < (I j).1 ∨ (I j).2 < (I i).1) →
+          ¬ IsPower (∏ i : Fin r, ∏ m ∈ Icc (I i).1 (I i).2, m)) ↔ answer(sorry) := by
+  sorry
+
+/--
+$p$ is the least prime satisfying $k \le p$
+if for any other prime $q$, $k \le q$ implies $p \le q$.
+-/
+def leastPrime (k p : ℕ) : Prop :=
```
callesonne: I would suggest implementing `Nat.nextPrime` of type `ℕ` using `Nat.find`. This would not be a `Prop` but a natural number, and given only the parameter `k` (so instead of giving it `p, k` you give it only `k` and it returns `p`).

callesonne (commented):
Thanks! Here are some more comments :)

Mal-Pat:
Thanks for the suggestion - I've implemented `nextPrime` using `Nat.find`.

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,92 @@
...
+  sorry
+
+/--
+Returns the least prime satisfying $k \le p$
+-/
+def nextPrime (k : ℕ) : ℕ :=
+  Nat.find (Nat.exists_infinite_primes k)
+
+/--
+Let $k$, $l$, $n$ be integers such that $k \ge 3$, $l \ge 2$ and $n + k \ge p^{(k)}$,
```
callesonne, line 60: The comment here says $n + k \ge p^{(k)}$, but the statement says `nextPrime k < n + k`. Which on

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,92 @@
...
+where $p^{(k)}$ is the least prime satisfying $p^{(k)} \ge k$.
+Then there is a prime $p \ge k$ for which $\alpha_p \not\equiv 0 \pmod{1}$,
+where $\alpha_p$ is the power of $p$ dividing $(n + 1) \ldots (n + k)$.
```
callesonne: I don't see \alpha_p in the lean formalization, am I missing something?

Mal-Pat: `α_p` equals `Nat.factorization (∏ m ∈ Icc (n + 1) (n + k), m) p`.

Would it be better to rewrite the docstring without `α_p`? 
I just copied the exact statement of Theorem 2 from [ErSe75].

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,92 @@
...
+if $I_1,\ldots,I_r$ are disjoint intervals of consecutive integers,
+all of length at least $k$, then
+$$
+  \prod_{1\leq i\leq r}\prod_{m\in I_i}m
+$$
+is not a perfect power?
+-/
+@[category research open, AMS 11]
+theorem erdos_930 :
+    (∀ r > 0, ∃ k, ∀ (I : Fin r → ℕ × ℕ),
```
YaelDillies:
```suggestion
    (∀ r > 0, ∃ k, ∀ I₁ I₂ : Fin r → ℕ,
```
will make your statement more readable

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,92 @@
...
+      (∀ i : Fin r, 0 < (I i).1 ∧ (I i).1 ≤ (I i).2 ∧ k ≤ (I i).2 - (I i).1 + 1) →
```
YaelDillies: The following is more clearly correct
```suggestion
      (∀ i : Fin r, 0 < (I i).1 ∧ (I i).1 + k ≤ (I i).2 + 1) →
```

FormalConjectures/ErdosProblems/930.lean
```diff
@@ -0,0 +1,92 @@
...
+Theorem 2 from [ErSe75].
+
+[ErSe75] Erdős, P. and Selfridge, J. L., The product of consecutive integers is never a power. Illinois J. Math. (1975), 292-301.
+-/
+@[category research solved, AMS 11]
+theorem erdos_930.variant.consecutive_strong :
+    ∀ k l n, 3 ≤ k → 2 ≤ l → nextPrime k ≤ n + k →
+      ∃ p, k ≤ p ∧ p.Prime ∧
+        ¬ (l ∣ Nat.factorization (∏ m ∈ Icc (n + 1) (n + k), m) p) := by
+  sorry
```
YaelDillies, line 74: This looks too simple: Take `p` to be a prime greater than `∏ m ∈ Icc (n + 1) (n + k), m` and `k`. Then `Nat.factorization (∏ m ∈ Icc (n + 1) (n + k), m) p = 1`, so `l` clearly doesn't divide it. Are you sure you are correctly quoting the statement?

FormalConjectures/ErdosProblems/930.lean
YaelDillies: Yes, probably. It's an unnecessary level of indirection IMO

FormalConjectures/ErdosProblems/930.lean
Mal-Pat: If `p` is greater than `∏ m ∈ Icc (n + 1) (n + k), m`, then `Nat.factorization (∏ m ∈ Icc (n + 1) (n + k), m) p` would be `0`, not `1`. 

I'm sure I have correctly quoted the exact statement of Theorem 2 from *Erdős, P. and Selfridge, J. L., The product of consecutive integers is never a power. Illinois J. Math. (1975), 292-301.*

YaelDillies: Ah, it's the docstring that is confusing! For me, the "power of 5 that divides 25" is 25, not 2. You probably mean "exponent" instead of "power" here

Mal-Pat: I was initially confused as well - "power" is the term used in the paper.

I have changed it to "multiplicity of `p`" for clarity.

YaelDillies: Yes, please!

FormalConjectures/ErdosProblems/930.lean
Mal-Pat: Thank you for that simplification - I missed it.

FormalConjectures/ErdosProblems/930.lean
Mal-Pat: Apologies for the confusion. I have changed the docstring.

PR #1254, opened 2025-11-21
more `open scoped EuclideanGeometry`

mo271:


FormalConjectures/Wikipedia/Kakeya.lean
```diff
@@ -32,14 +34,14 @@ For simplicity, we omit the compactness assumption here.
 For a discussion on the equivalence of definitions with and without compactness, see
 [this paper](https://arxiv.org/pdf/2203.15731).
 -/
-def IsKakeya {n : ℕ} (S : Set (EuclideanSpace ℝ (Fin n))) : Prop :=
+def IsKakeya {n : ℕ} (S : Set (ℝ^n)) : Prop :=
```
smmercuri, line 37: Can the parents around `ℝ^n` go or are they needed for this case?

mo271: they are needed unfortunately

PR #1258, opened 2025-11-22
add github action + checker to updates on erdos problem pages 

ayushman1210:
- Added (workflow):  .github/workflows/erdos-watcher.yml — GitHub Actions workflow that runs daily (and manually) to check Erdős problem pages, produce updates, create issues labeled erdos-update, and commit the last-seen dates file.
- Added (script):  .github/scripts/check_erdos_updates.py — Python checker that:
- finds https://www.erdosproblems.com/<id> URLs in *.lean files,
- fetches pages and extracts the "last edited / updated" date,
- compares with the persisted dates and emits an updates.json list.
- Supports --dry-run, --sample-html (test a local HTML file), and normal mode.
- Added (state):  .github/erdos-updates/erdos_dates.json — initial empty JSON used to persist last-seen dates.
- Added (docs):  .github/erdos-updates/README.md — usage notes and local testing instructions.
- Added (deps):  .github/erdos-updates/requirements.txt — Python packages needed by the checker.
- Added (examples):  .github/erdos-updates/examples/erdos_1.html — sample HTML for quick local testing.
- Added (local automation):  scripts/run_erdos_check.ps1 — PowerShell helper that creates a venv, (optionally) installs deps, and runs the checker in dry-run (default) or full mode (-FullRun).
- Added (editor task):  .vscode/tasks.json — VS Code tasks to run the dry-run or full checker from the editor.
- Edited (script):  .github/scripts/check_erdos_updates.py — improved date-extraction to prefer the site's "last edited" phrasing and added --dry-run / --sample-html modes (if you run the file now it will show results instead of immediately writing files).

### DIFF
--- a/.vscode/copyright.code-snippets
+++ b/.vscode/copyright.code-snippets
@@ -1,0 +1,23 @@
+{
+	"Copyright header": {
+		"scope": "lean4",
+		"prefix": "copyright",
+		"body": [
+      "/-",
+      "Copyright ${CURRENT_YEAR} The Formal Conjectures Authors.",
+      "",
+      "Licensed under the Apache License, Version 2.0 (the \"License\");",
+      "you may not use this file except in compliance with the License.",
+      "You may obtain a copy of the License at",
+      "",
+      "    https://www.apache.org/licenses/LICENSE-2.0",
+      "",
+      "Unless required by applicable law or agreed to in writing, software",
+      "distributed under the License is distributed on an \"AS IS\" BASIS,",
+      "WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.",
+      "See the License for the specific language governing permissions and",
+      "limitations under the License.",
+      "-/",
+		]
+	},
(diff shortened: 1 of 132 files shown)


seewoo5:
I'm not sure if I'm the best person to review this, at least I don't know if I can test it since I'm not a maintainer of this repo. At least it would be nice if you could tell how to test it ourselves.

mo271:
Just a few first thoughts on this: 
 - I think it would be easier to add something on the side of the (non-public?) repo that generates erdosproblems.com. 
 - Instead of keeping state, we could write a simple script that 
   (1) gets for each problem the state of each problem (open/solved) and
   (2) compares with our attribute for the main problem, which is typically names "erdos_$num". 
 For (1), we can rely on https://github.com/teorth/erdosproblems/blob/main/data/problems.yaml, which is kept up-to-date, I believe. 
 Wdyt?

mo271:
closing due to inactivity

PR #1259, opened 2025-11-22
fix(GromovPolynomialGrowth): `HasPolynomialGrowth` should only quantify over positive `n`

smmercuri:
- fix `HasPolynomialGrowth` to quantify only over `n > 0`
- add some more API and tests to demonstrate the issue with `n = 0`: if `G` is an infinite group, then no matter what `d` is, we can always find `n` such that `C * n ^ d < GrowthFunction S n`. If `d > 0` then `n = 0` trivially works. On the other hand, if `d = 0` then this reduces to showing that `GrowthFunction S n` is unbounded which holds in the infinite case.
- this generalises the counterexample of `Multiplicative ℤ` discovered by AlphaProof to `GromovPolynomialGrowthTheorem`, since this is an infinite finitely-generated nilpotent group.

### DIFF
--- a/FormalConjectures/Wikipedia/GromovPolynomialGrowth.lean
+++ b/FormalConjectures/Wikipedia/GromovPolynomialGrowth.lean
@@ -147,8 +147,8 @@ theorem tendsto_atTop_growthFunction_of_infinite [Infinite G] {S : Set G} (hS : 
   intro A
   obtain rfl | hA := A.eq_empty_or_nonempty
   · aesop
-  · choose n hn using fun (a : A) ↦ exists_cayleyBall_mem_of_closure_eq h a
-    let N : ℕ := (Set.range n).toFinset.max' (by simpa [Finset.nonempty_iff_ne_empty])
+  · choose n hn using fun (a : A) ↦ exists_cayleyBall_mem_of_closure_eq_top h a
+    let N : ℕ := (Set.range n).toFinset.max' (by simp [hA])
     refine ⟨N, fun a ha ↦ ?_⟩
     simpa using cayleyBall_monotone S (Finset.le_max' _ _ (by aesop)) (hn ⟨a, ha⟩)
 


FormalConjectures/Wikipedia/GromovPolynomialGrowth.lean
```diff
@@ -91,17 +114,73 @@ lemma CayleyBall_inv {G : Type*} [Group G] (S : Set G) {g : G} {n : ℕ}
   simp only [inv_inv] at this
   exact this.symm
 
+-- end of Claude generated statements
```
callesonne, line 117: what does this mean? It seems to be at the beginning of this block, and above no statements are added?

FormalConjectures/Wikipedia/GromovPolynomialGrowth.lean
```diff
@@ -30,37 +30,60 @@ Claude 4.0 Sonnet: https://claude.ai/share/918bb269-bd28-4c09-b84e-cab579c836e8
...
+variable {G : Type*} [Group G]
+
 /-- The `CayleyBall` is the ball of radius `n` in the Cayley graph of a group `G` with generating
     set `S`. -/
-def CayleyBall {G : Type*} [Group G] (S : Set G) (n : ℕ) : Set G :=
+def CayleyBall (S : Set G) (n : ℕ) : Set G :=
   {g : G | ∃ (l : List G), l.length ≤ n ∧ (∀ s ∈ l, s ∈ S ∨ s⁻¹ ∈ S) ∧ l.prod = g}
 
+@[category API, AMS 20]
+theorem cayleyBall_zero {G : Type*} [Group G] (S : Set G) :
```
callesonne: Given the variable above?
```suggestion
theorem cayleyBall_zero (S : Set G) :
```

FormalConjectures/Wikipedia/GromovPolynomialGrowth.lean
```diff
@@ -30,37 +30,60 @@ Claude 4.0 Sonnet: https://claude.ai/share/918bb269-bd28-4c09-b84e-cab579c836e8
...
+  exact (this.image List.prod).subset fun _ _ ↦ by aesop (add simp [CayleyBall])
+
 /-- The `GrowthFunction` of a group `G` with respect to a set `S` counts the number
     of group elements that can be reached by words of length at most `n` in `S`. -/
-noncomputable def GrowthFunction {G : Type*} [Group G] (S : Set G) (n : ℕ) : ℕ :=
+noncomputable def GrowthFunction (S : Set G) (n : ℕ) : ℕ :=
   (CayleyBall S n).ncard
 
+@[category API, AMS 20]
+theorem growthFunction_zero {G : Type*} [Group G] (S : Set G) :
```
callesonne:
```suggestion
theorem growthFunction_zero (S : Set G) :
```

FormalConjectures/Wikipedia/GromovPolynomialGrowth.lean
```diff
@@ -91,17 +114,73 @@ lemma CayleyBall_inv {G : Type*} [Group G] (S : Set G) {g : G} {n : ℕ}
...
   exact this.symm
 
+-- end of Claude generated statements
+
+@[category API, AMS 20]
+lemma mem_cayleyBall_one_of_mem {S : Set G} {g : G} (hg : g ∈ S) : g ∈ CayleyBall S 1 :=
+  ⟨[g], by simp_all⟩
+
+@[category API, AMS 20]
+lemma exists_mem_cayleyBall_of_closure_eq {S : Set G} (h : Subgroup.closure S = ⊤) (g : G) :
```
callesonne:
```suggestion
lemma mem_some_cayleyBall_of_closure_eq {S : Set G} (h : Subgroup.closure S = ⊤) (g : G) :
```
This name makes more sense to me, but I don't know if it disagrees with any naming conventions. I found the last name slightly ambiguous, but up to you.

FormalConjectures/Wikipedia/GromovPolynomialGrowth.lean
smmercuri: l. 43 has the following comment `-- Basic properties of CayleyBall and GrowthFunction (Claude generated statements, human proofs)` so I thought it'd be best to distinguish the end of that for clarity

FormalConjectures/Wikipedia/GromovPolynomialGrowth.lean
smmercuri: I'll modify it slightly to `exists_cayleyBall_mem_of_closure_eq` as I think `some` is not really used for exists statements

FormalConjectures/Wikipedia/GromovPolynomialGrowth.lean
callesonne: Okay that sounds perfect.

FormalConjectures/Wikipedia/GromovPolynomialGrowth.lean
```diff
@@ -91,17 +114,73 @@ lemma CayleyBall_inv {G : Type*} [Group G] (S : Set G) {g : G} {n : ℕ}
...
   exact this.symm
 
+-- end of Claude generated statements
+
+@[category API, AMS 20]
+lemma mem_cayleyBall_one_of_mem {S : Set G} {g : G} (hg : g ∈ S) : g ∈ CayleyBall S 1 :=
+  ⟨[g], by simp_all⟩
+
+@[category API, AMS 20]
+lemma exists_cayleyBall_mem_of_closure_eq {S : Set G} (h : Subgroup.closure S = ⊤) (g : G) :
```
YaelDillies:
```suggestion
lemma exists_cayleyBall_mem_of_closure_eq_top {S : Set G} (h : Subgroup.closure S = ⊤) (g : G) :
```
for clarity

FormalConjectures/Wikipedia/GromovPolynomialGrowth.lean
```diff
@@ -91,17 +114,73 @@ lemma CayleyBall_inv {G : Type*} [Group G] (S : Set G) {g : G} {n : ℕ}
...
+@[category API, AMS 20]
+theorem tendsto_atTop_growthFunction_of_infinite [Infinite G] {S : Set G} (hS : S.Finite)
+    (h : Subgroup.closure S = ⊤) : atTop.Tendsto (GrowthFunction S) atTop := by
+  delta GrowthFunction
+  have (n : ℕ) : Fintype (CayleyBall S n) := (cayleyBall_finite hS n).fintype
+  apply ((Finset.tendsto_card_atTop).comp (f := fun n ↦ (CayleyBall S n).toFinset) ?_).congr
+    (by simp [Set.ncard_eq_toFinset_card'])
+  apply tendsto_atTop_atTop_of_monotone fun _ _ ↦ by simpa using cayleyBall_monotone S
+  intro A
+  by_cases hA : A = ∅
```
YaelDillies:
```suggestion
  obtain rfl | hA := A.eq_empty_or_nonempty
```
will save you a few characters below

Paul-Lez (approved):

PR #1262, opened 2025-11-22
fix: remove use of `Partition` in `274.lean`

smmercuri:
- `P : Partition (⊤ : Subgroup G)` was too strong a hypothesis for an exact covering of a group. `Partition` asserts that the supremum over parts covers `G`, but subgroup supremum is subgroup generation. We require union of sets.
- Using `P : Partition (⊤ : Set G)` works but is a bit messy because we have to separately assert that parts represent cosets and lose access to `Subgroup.index`
- Define `Group.ExactCovering` to reconcile this

### DIFF
--- a/FormalConjectures/ErdosProblems/274.lean
+++ b/FormalConjectures/ErdosProblems/274.lean
@@ -35,7 +35,7 @@ representative `{g_1, ..., g_k}` such that the cosets `g_iH_i` are pairwise disj
 union covers `G`.
 
 Note that this differs from `Partition (α := Subgroup G)` because the covering condition there
-invokes `Subgroup.sub` which is multiplication and thus stronger than union. This definition
+invokes `Subgroup.sup` which is subgroup generation and thus stronger than union. This definition
 is easier to use in this contect than the alternative `Partition (α := Set G)`, which lacks
 subgroup definitions such as `Subgroup.index`. -/
 structure Group.ExactCovering (G : Type*) [Group G] (ι : Type*) [Fintype ι] where


FormalConjectures/ErdosProblems/274.lean
```diff
@@ -30,16 +30,29 @@ namespace Erdos274
...
 
+/-- An exact covering of a group `G` is a finite collection of subgroups `{H_1, ..., H_k}` and
+representative `{g_1, ..., g_k}` such that the cosets `g_iH_i` are pairwise disjoint and their
+union covers `G`.
+
+Note that this differs from `Partition (α := Subgroup G)` because the covering condition there
+invokes `Subgroup.sup` which is subgroup generation and thus stronger than union. This definition
+is easier to use in this contect than the alternative `Partition (α := Set G)`, which lacks
+subgroup definitions such as `Subgroup.index`. -/
+structure Group.ExactCovering (G : Type*) [Group G] (ι : Type*) [Fintype ι] where
```
callesonne, line 41: I think `ι` should be part of the data here, what do you think?

callesonne (approved):
LGTM apart from minor comment

FormalConjectures/ErdosProblems/274.lean
smmercuri: I couldn't decide between the two options. What would be the main benefit of including it in the data?

callesonne: That you don't have to quantify over it separately. As this `ι` will usually be used only for this it is nicer to not have to do that.

smmercuri: After thinking about it a bit, I think it's better to leave it as a parameter. Otherwise, it's a bit opaque and difficult to work with:
- Including `ι` as data requires extra API to extract the cardinality, which we need to assert is `> 1` in some hypotheses later.
- Similarly, we might also want to assert other hypotheses and type-classes on `ι`  in the future, so it's slightly easier to leave it unbundled

Issue #159, opened 2025-06-04
Reed's Conjecture for triangle-free graphs with maximum degree 6


andrew-d-king:
### What is the conjecture

Reed's Conjecture in its general form is that the chromatic number of a graph is at most halfway between its trivial lower bound (clique number omega) and its trivial upper bound (maximum degree Delta, plus one), rounded up.

I suggest formalizing the simplest open case: Delta=6, omega=2, as described here:
https://mathoverflow.net/questions/37923/does-every-triangle-free-graph-with-maximum-degree-at-most-6-have-a-5-colouring

### Prerequisites needed

Let me take a crack at giving the conjecture in a slightly more formal language.

For any undirected graph G=(V,E) where V is a set of vertices and E is a set of undirected edges,

if for every v in V, the set of edges in E containing v has size at most 6, and

if for every v in V, there are no two other vertices u,w in V with the property that ( {v,u} in E and {v,w} in E and {u,w} in E )

then there is a function f:V->{1,2,3,4,5} such that for all {v,u} in E, f(v) != f(u).

### [AMS categories](https://github.com/google-deepmind/formal-conjectures/labels?q=ams-)

* ams-05C15

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

b-mehta:
I have a slight preference for stating the conjecture in full generality, rather than just the simplest open case. But it's better to have a special case than nothing at all!

Paul-Lez:
We could have both;)

edwag:
I've taken a crack at adding both the general and the special case at https://github.com/google-deepmind/formal-conjectures/pull/1264

PR #1264, opened 2025-11-22
feat: Reed's omega, delta, and chi conjecture

edwag:
Addresses https://github.com/google-deepmind/formal-conjectures/issues/159

### DIFF
--- a/FormalConjectures/Paper/ReedOmegaDeltaChi.lean
+++ b/FormalConjectures/Paper/ReedOmegaDeltaChi.lean
@@ -21,7 +21,7 @@ import FormalConjectures.Util.ProblemImports
 
 *References*:
 - [B. Reed,  ω Δ and χ, J. Graph Theory 27 (1998) 177-212.](https://onlinelibrary.wiley.com/doi/10.1002/(SICI)1097-0118(199804)27:4%3C177::AID-JGT1%3E3.0.CO;2-K)
-- [openproblemgarden(http://www.openproblemgarden.org/op/reeds_omega_delta_and_chi_conjecture)
+- [openproblemgarden](http://www.openproblemgarden.org/op/reeds_omega_delta_and_chi_conjecture)
 - [mathoverflow/37923](https://mathoverflow.net/questions/37923) asked by user [Andrew D. King](https://mathoverflow.net/users/4580/andrew-d-king)
 -/
 
@@ -45,8 +45,8 @@ theorem reed_omega_delta_chi_conjecture :
   sorry
 
 /--
-For a finite graph $G$, we define $\Delta(G)$ to be the maximum degree, $\omega(G)$ to be the 
-size of the largest clique subgraph, and $\chi(G)$ to be the chromatic number. Reed's omega, 
+For a finite graph $G$, we define $\Delta(G)$ to be the maximum degree, $\omega(G)$ to be the
+size of the largest clique subgraph, and $\chi(G)$ to be the chromatic number. Reed's omega,
 delta, and chi conjecture states that $$\chi(G) \leq \lceil \frac{1}{2}(\omega(G) + \Delta(G) + 1) \rceil.$$
 -/
 @[category research open, AMS 5]


FormalConjectures/Other/ReedOmegaDeltaChi.lean
```diff
@@ -0,0 +1,64 @@
...
+
+*Reference* (https://mathoverflow.net/questions/37923)
+-/
+
+namespace ReedOmegaDeltaChi
+
+open Classical
+
+noncomputable
+def maxDegree {V : Type} (G : SimpleGraph V) : ℕ∞ :=
```
edwag: I'm not sure if this is already defined somewhere in Mathlib, or if perhaps it would make sense to put this into ForMathlib? Mathlib does define `SimpleGraph.maxDegree`, but as far as I can see this only exists for finite graphs.

andrew-d-king: I think it would also be fine to state the conjecture for finite graphs, if it's easier.

edwag:
@mo271 could I please get a review on this? 🙏

FormalConjectures/Other/ReedOmegaDeltaChi.lean
```diff
@@ -0,0 +1,71 @@
...
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Reed's omega, delta, and chi conjecture
+
+*Reference:* (http://www.openproblemgarden.org/op/reeds_omega_delta_and_chi_conjecture)
+
+*Reference* (https://mathoverflow.net/questions/37923)
```
mo271:
```suggestion
# Reed's omega, delta, and chi conjecture

*References*:
- [B. Reed,  ω Δ and χ, J. Graph Theory 27 (1998) 177-212.](https://onlinelibrary.wiley.com/doi/10.1002/(SICI)1097-0118(199804)27:4%3C177::AID-JGT1%3E3.0.CO;2-K) 
-
[openproblemgarden(http://www.openproblemgarden.org/op/reeds_omega_delta_and_chi_conjecture)
- [mathoverflow/37923](https://mathoverflow.net/questions/37923) asked by user [Andrew D. King](https://mathoverflow.net/users/4580/andrew-d-king)
```

I think it would be best to move the file to the `Paper` directory, for instance `FormalConjectures/Paper/OmegaDeltaChi.lean`

FormalConjectures/Other/ReedOmegaDeltaChi.lean
```diff
@@ -0,0 +1,71 @@
...
+open Classical
+
+noncomputable
+def maxDegree {V : Type} (G : SimpleGraph V) : ℕ∞ :=
+  if h : ∀ v, Finite (G.neighborSet v) then
+    have : ∀ v, Fintype (G.neighborSet v) :=
+      fun v => @Fintype.ofFinite _ (h v)
+    ⨆ v, G.degree v
+  else
+    ⊤
```
mo271: Is this givin the junk value of "0" for a graph that has Finite degree for each node, but infinitely many nodes, and nodes of arbitrarily high degree? Should it be \top in that case as well. 
Perhaps more something like
```
noncomputable
def maxDegree {V : Type} (G : SimpleGraph V) : ℕ∞ :=
  ⨆ v, if h : (G.neighborSet v).Finite then
    have : Fintype (G.neighborSet v) := by exact h.fintype
    (G.degree v : ℕ∞)
  else
    ⊤
```
and then make sure `maxCliqueSize` is not depending on that junk value?

mo271 (commented):
Very nice! Just some minor comments!

FormalConjectures/Other/ReedOmegaDeltaChi.lean
```diff
@@ -0,0 +1,73 @@
...
+conjecture states that $$\chi(G) \leq \lceil \frac{1}{2}(\omega(G) + \Delta(G) + 1) \rceil.$$
+-/
+@[category research open, AMS 5]
+theorem reed_omega_delta_chi_conjecture :
+  ∀ {V : Type} (G : SimpleGraph V),
+    let χ := G.chromaticNumber
+    let ω := maxCliqueSize G
+    let Δ := maxDegree G
+  /- (d+1)/2 = ⌈d/2⌉ -/
+  χ ≤ (ω + Δ + 1).map (fun d => (d + 1) / 2) := by
```
YaelDillies: Here is my suggestion, for finite graphs:
```suggestion
    (∀ {V : Type} [Fintype V] [DecidableEq V] (G : SimpleGraph V) [DecidableRel G.Adj],
      let χ := G.chromaticNumber
      let ω := G.cliqueNum
      let Δ := G.maxDegree
      /- (d+1)/2 = ⌈d/2⌉ -/
      2 * χ ≤ ω + Δ + 2 ↔ answer(sorry) := by
```

FormalConjectures/Other/ReedOmegaDeltaChi.lean
YaelDillies: Yes, let's state it for finite graphs for now

FormalConjectures/Other/ReedOmegaDeltaChi.lean
edwag: How would I actually check if this gives `0`? I had the case of unbounded degree in mind when writing this, and I thought that the supremum would be \top in this case, since the return type is N_\infty. Maybe it would be good to include some kind of unit test here. Like define a graph with finite but unbounded degrees and prove that the `maxDegree` is `\top`.

FormalConjectures/Other/ReedOmegaDeltaChi.lean
edwag: 1. I thought that the convention in this repo is to use the `answer(sorry)` style if the conjecture was phrased as a question. I'm not actually familiar with this conjecture, I'm just going by the page at http://www.openproblemgarden.org/op/reeds_omega_delta_and_chi_conjecture, where the conjecture is phrased as a statement. Happy to switch to `answer(sorry)` if this conjecture is usually phrased as a question but would just like to confirm that we want this change!
2. I think we should add this finite version as a separate theorem statement, since we do also want to state the conjecture for infinite graphs.

YaelDillies: 1. Sorry, you might be right. I am an expert mathlib reviewer, but not an expert conjectures one 😅
2. But then shouldn't the infinite version be stated in terms of cardinals? Right now, you have an in-between statement.

FormalConjectures/Other/ReedOmegaDeltaChi.lean
YaelDillies: I believe the current definition is correct, but it's certainly a bit weird compared to what it could be. I would expect something like:
```suggestion
def edegree (G : SimpleGraph V) (v : V) : ℕ∞ := (G.neighborSet v).encard

def emaxDegree {V : Type*} (G : SimpleGraph V) : ℕ∞ := ⨆ v, G.edegree v
```

FormalConjectures/Other/ReedOmegaDeltaChi.lean
edwag: As I understand it, this conjecture is only for the case where the clique number and max degree are both finite, so I think we can state everything using just $N_\infty$? I'd like to defer to @andrew-d-king on this one, since I think he has a better idea than I do of the precise statement of the conjecture.

FormalConjectures/Other/ReedOmegaDeltaChi.lean
edwag: This looks much nicer, thanks!

FormalConjectures/Other/ReedOmegaDeltaChi.lean
```diff
@@ -0,0 +1,69 @@
...
+
+def emaxDegree {V : Type*} (G : SimpleGraph V) : ℕ∞ := ⨆ v, G.edegree v
+
+/--
+For a nonempty graph, `G.cliqueNum` is 0 iff the clique size is unbounded. For the purposes of
+stating the conjecture, we actually want the max clique size to be `∞` in this case.
+-/
+noncomputable
+def maxCliqueSize {V : Type} (G : SimpleGraph V) : ℕ∞ :=
+  if Nonempty V ∧ G.cliqueNum = 0 then ⊤ else G.cliqueNum
```
YaelDillies:
```suggestion
def ecliqueNum {V : Type} (G : SimpleGraph V) : ℕ∞ := ⨆ (s : Finset V) (_ : G.IsClique s), #s
```
if you add `open scoped Finset` earlier in the file

FormalConjectures/Other/ReedOmegaDeltaChi.lean
YaelDillies: At least, could you fix the indentation and avoid this weird `ENat.map`?
```suggestion
    (∀ {V : Type} (G : SimpleGraph V),
      let χ := G.chromaticNumber
      let ω := G.ecliqueNum
      let Δ := G.emaxDegree
      2 * χ ≤ ω + Δ + 2 ↔ answer(sorry) := by
```

mo271:
@edwag Seems like there is not much left to do here? 
Would be great to get this conjecture merged!

edwag:
> @edwag Seems like there is not much left to do here? Would be great to get this conjecture merged!

@mo271 I should be able to finish this off in the next week or so - have been very busy for the last few weeks but will have some more free time opening up shortly!

FormalConjectures/Other/ReedOmegaDeltaChi.lean
edwag: I've kept the statement for general graphs, but have also added a second statement for finite graphs, following @YaelDillies suggested implementation.

edwag:
I've pushed changes which I think resolve all the outstanding feedback.

FormalConjectures/Paper/ReedOmegaDeltaChi.lean
```diff
@@ -0,0 +1,83 @@
...
+noncomputable
+def edegree {V : Type*} (G : SimpleGraph V) (v : V) : ℕ∞ := (G.neighborSet v).encard
+
+noncomputable
+def emaxDegree {V : Type*} (G : SimpleGraph V) : ℕ∞ := ⨆ v, G.edegree v
+
+noncomputable
+def ecliqueNum {V : Type} (G : SimpleGraph V) : ℕ∞ := ⨆ (s : Finset V) (_ : G.IsClique s), #s
+
+end SimpleGraph
```
YaelDillies: Can you put these in `ForMathlib`?

FormalConjectures/Paper/ReedOmegaDeltaChi.lean
```diff
@@ -0,0 +1,65 @@
...
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Reed's omega, delta, and chi conjecture
+
+*References*:
+- [B. Reed,  ω Δ and χ, J. Graph Theory 27 (1998) 177-212.](https://onlinelibrary.wiley.com/doi/10.1002/(SICI)1097-0118(199804)27:4%3C177::AID-JGT1%3E3.0.CO;2-K)
+-
+[openproblemgarden(http://www.openproblemgarden.org/op/reeds_omega_delta_and_chi_conjecture)
```
mo271:
```suggestion
- [openproblemgarden(http://www.openproblemgarden.org/op/reeds_omega_delta_and_chi_conjecture)
```

FormalConjectures/Paper/ReedOmegaDeltaChi.lean
```diff
@@ -0,0 +1,65 @@
...
+@[category research open, AMS 5]
+theorem reed_omega_delta_chi_conjecture :
+  ∀ {V : Type} (G : SimpleGraph V),
+    let χ := G.chromaticNumber
+    let ω := G.ecliqueNum
+    let Δ := G.emaxDegree
+    2 * χ ≤ ω + Δ + 2 := by
+  sorry
+
+@[category research open, AMS 5]
```
mo271, line 52:
```suggestion

/--
For a finite graph $G$, we define $\Delta(G)$ to be the maximum degree, $\omega(G)$ to be the 
size of the largest clique subgraph, and $\chi(G)$ to be the chromatic number. Reed's omega, 
delta, and chi conjecture states that $$\chi(G) \leq \lceil \frac{1}{2}(\omega(G) + \Delta(G) + 1) \rceil.$$
-/
@[category research open, AMS 5]
```

FormalConjectures/Paper/ReedOmegaDeltaChi.lean
edwag: I'm not sure how to edit this suggested change in the UI, but I think we're also missing a closing `]` here. I'll make this change locally and push.

FormalConjectures/Paper/ReedOmegaDeltaChi.lean
edwag: I've moved these into `Combinatorics/SimpleGraph/GraphConjectures/Definitions`

PR #1267, opened 2025-11-23
fix: `erdos_196`

smmercuri:
Casting from `List` to `Finset` loses the ordering asserted in `erdos_196` making the problem trivial (because `{0, 1, 2, 3}` is always constructible). Create `List.IsAPOfLength` to resolve this.

### DIFF
--- a/FormalConjectures/ForMathlib/Combinatorics/AP/Basic.lean
+++ b/FormalConjectures/ForMathlib/Combinatorics/AP/Basic.lean
@@ -56,7 +56,7 @@ theorem card (h : s.IsAPOfLengthWith l a d) : ENat.card s = l := h.1
 theorem eq (h : s.IsAPOfLengthWith l a d) : s = {a + n • d | (n : ℕ) (_ : n < l)} := h.2
 
 /-- An arithmetic progression with first term `a` and difference `d` is of length zero if and only
-if the difference is non-zero and `s` is empty. -/
+if `s` is empty. -/
 @[simp]
 theorem zero : s.IsAPOfLengthWith 0 a d ↔ s = ∅ := by
   simpa [Set.IsAPOfLengthWith] using fun _ => by aesop
@@ -78,7 +78,7 @@ theorem length (h : s.IsAPOfLengthWith l a d) : s.length = l := by
   simp [h]
 
 /-- An arithmetic progression with first term `a` and difference `d` is of length zero if and only
-if the difference is non-zero and `s` is empty. -/
+if `s` is empty. -/
 @[simp]
 theorem zero : s.IsAPOfLengthWith 0 a d ↔ s = [] := by
   simp [IsAPOfLengthWith]


FormalConjectures/ForMathlib/Combinatorics/AP/Basic.lean
```diff
@@ -41,6 +41,13 @@ value `⊤`.
 def Set.IsAPOfLengthWith (s : Set α) (l : ℕ∞) (a d : α) : Prop :=
   ENat.card s = l ∧ s = {a + n • d | (n : ℕ) (_ : n < l)}
```
callesonne, line 42: It is now unclear to me why we enforce the length here (this is not related to your PR, but maybe you know)?

FormalConjectures/ForMathlib/Combinatorics/AP/Basic.lean
```diff
@@ -62,6 +69,28 @@ theorem one : s.IsAPOfLengthWith 1 a d ↔ s = {a} := by
...
+namespace List.IsAPOfLengthWith
+
+variable {s : List α} {l : ℕ} {a d : α}
+
+theorem length (h : s.IsAPOfLengthWith l a d) : s.length = l := by
+  rw [IsAPOfLengthWith] at h
+  simp [h]
+
+/-- An arithmetic progression with first term `a` and difference `d` is of length zero if and only
+if the difference is non-zero and `s` is empty. -/
```
callesonne: Is this comment accurate? Where is the statement about non-zero difference?

FormalConjectures/ForMathlib/Combinatorics/AP/Basic.lean
smmercuri: [from yourself actually!](https://github.com/google-deepmind/formal-conjectures/pull/255#discussion_r2184910452) Basically to avoid having to do a weird case split on `d = 0`

FormalConjectures/ForMathlib/Combinatorics/AP/Basic.lean
smmercuri: Oh that's what the result used to say with the case split on `d = 0` I think

FormalConjectures/ForMathlib/Combinatorics/AP/Basic.lean
callesonne: Haha yeah I definitely recall at least thinking about this before!

FormalConjectures/ForMathlib/Combinatorics/AP/Basic.lean
```diff
@@ -41,6 +41,13 @@ value `⊤`.
 def Set.IsAPOfLengthWith (s : Set α) (l : ℕ∞) (a d : α) : Prop :=
   ENat.card s = l ∧ s = {a + n • d | (n : ℕ) (_ : n < l)}
 
+/--
+A list version of `Set.IsAPOfLengthWith`. Useful when order preservation is required, for example
+when considering images under arbitrary functions.
+-/
+def List.IsAPOfLengthWith (s : List α) (l : ℕ) (a d : α) : Prop :=
+  s = (List.range l).map fun n ↦ a + n • d
```
callesonne, line 49: ~~But then is this definition fine if `d = 0`?~~ Edit: I guess yeah its fine to interpret `[a, a, a, a, a]` as an arithmetic sequence.

PR #1270, opened 2025-11-24
add `erdos_26`

smmercuri:
- `erdos_26` plus variants. While this has status disproved on the website, it appears to still be open if restricting to thick sets
- add to `Set.HasDensity` API and remove some unnecessary hypotheses there

### DIFF
--- a/FormalConjectures/ErdosProblems/26.lean
+++ b/FormalConjectures/ErdosProblems/26.lean
@@ -89,7 +89,7 @@ for some $a\in A$?
 -/
 @[category research open, AMS 11]
 theorem erdos_26 (A : ℕ → ℕ) (hA : StrictMono A) (h : IsThick A) :
-    (∃ k, IsBehrend (A · + k)) ↔ answer(True) := by
+    (∃ k, IsBehrend (A · + k)) ↔ answer(sorry) := by
   sorry
 
 /--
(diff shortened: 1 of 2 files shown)


FormalConjectures/ErdosProblems/26.lean
```diff
@@ -0,0 +1,113 @@
...
+  exact (add_lt_of_neg_right _ hε).trans_le (Set.lowerDensity_le_one _)
+
+/--
+Let $A\subset\mathbb{N}$ be infinite such that $\sum_{a \in A} \frac{1}{a} = \infty$. Must
+there exist some $k\geq 1$ such that almost all integers have a divisor of the form $a+k$
+for some $a\in A$?
+-/
+@[category research open, AMS 11]
+theorem erdos_26 (A : ℕ → ℕ) (hA : StrictMono A) (h : IsThick A) :
+    (∃ k, IsBehrend (A · + k)) ↔ answer(True) := by
```
Paul-Lez: Is this what you intended to write?
```suggestion
    (∃ k, IsBehrend (A · + k)) ↔ answer(sorry) := by
```

Paul-Lez (approved):
Nice!

FormalConjectures/ErdosProblems/26.lean
smmercuri: yes!

PR #1271, opened 2025-11-24
feat(ErdosProblems): 1035, hypercube conjecture

### LINKED ISSUE #1086: Erdős Problem 1035
### What is the conjecture

https://www.erdosproblems.com/1035

  Is there a constant $c>0$ such that every graph on $2^n$ vertices with minimum degree $>(1-c)2^n$ contains the $n$-dimensional hypercube $Q_n$?

Status:  open

### Choose either option

Karanjot786:
Adds formalization of Erdős Problem 1035 about hypercube subgraphs.

**Problem**: Does there exist a constant c > 0 such that every graph on 2^n vertices 
with min degree > (1-c)·2^n contains the n-dimensional hypercube Q_n?

- Defines hypercube graph Q_n
- Includes two related variant questions from Erdős
- AMS category: 05 (Combinatorics)

Closes #1086


FormalConjectures/ErdosProblems/1035.lean
```diff
@@ -0,0 +1,92 @@
...
+*Reference:* [erdosproblems.com/1035](https://www.erdosproblems.com/1035)
+-/
+
+namespace Erdos1035
+
+/--
+The $n$-dimensional hypercube graph $Q_n$ has $2^n$ vertices (corresponding to binary strings
+of length $n$) with edges between vertices that differ in exactly one coordinate.
+-/
+def Hypercube (n : ℕ) : SimpleGraph (Fin n → Bool) where
```
eric-wieser:
```suggestion
def hypercube (n : ℕ) : SimpleGraph (Fin n → Bool) where
```
though I'd recommend abstracting to
```suggestion
def hypercube (n : Type*) : SimpleGraph (n → Bool) where
```

FormalConjectures/ErdosProblems/1035.lean
```diff
@@ -0,0 +1,92 @@
...
+
+/--
+Is there a constant $c > 0$ such that every graph on $2^n$ vertices with minimum degree
+$> (1-c) \cdot 2^n$ contains the $n$-dimensional hypercube $Q_n$?
+
+See also [576] for the extremal number of edges that guarantee a $Q_n$.
+-/
+@[category research open, AMS 05]
+theorem erdos_1035 :
+    (∃ c > 0, ∀ n : ℕ, ∀ (G : SimpleGraph (Fin (2^n))) [Fintype (Fin (2^n))]
```
eric-wieser: `[Fintype (Fin (2^n))]` is a bad idea; why did you add it?

FormalConjectures/ErdosProblems/1035.lean
```diff
@@ -0,0 +1,92 @@
...
+  loopless v h := by
+    obtain ⟨i, hi, _⟩ := h
+    exact hi rfl
+
+/--
+A graph $G$ contains (or embeds) a graph $H$ if there exists an injective graph homomorphism
+from $H$ to $G$.
+-/
+def contains {α β : Type*} (G : SimpleGraph α) (H : SimpleGraph β) : Prop :=
+  ∃ f : β → α, Function.Injective f ∧ ∀ {v w : β}, H.Adj v w → G.Adj (f v) (f w)
```
eric-wieser: Are you sure mathlib doesn't already have this? Note the naming convention says this should be `Contains` since it's a `Prop`.

mo271:
@Karanjot786 are you still working on this?

Karanjot786:
> @Karanjot786 are you still working on this?

yes I just got a little bit busy with my personal life, and by tomorrow, I will update it

FormalConjectures/ErdosProblems/1035.lean
Karanjot786: You're right. `Fin` types are already `Fintype` instances. I'll remove these redundant instances from all theorem statements. Thanks for catching this!

FormalConjectures/ErdosProblems/1035.lean
Karanjot786: I will rename it to `hypercube`. For the `Type*` suggestion, the adjacency relation needs exactly one coordinate to differ. How would you express this for a general type? Should I use `BitVector n` or another approach?

FormalConjectures/ErdosProblems/1035.lean
Karanjot786: I will rename it to `Contains`. I checked Mathlib and found `SimpleGraph.Embedding`, but it requires bidirectional edge preservation. This problem needs only an injective homomorphism. Should I use `∃ f : H →g G, Function.Injective` f instead of defining a new predicate?

Karanjot786:
Hi @mo271 @eric-wieser I addressed the review comments:

- Renamed Hypercube to hypercube
- Renamed contains to Contains
- Removed redundant Fintype instances

Ready for another look when you have time.

FormalConjectures/ErdosProblems/1035.lean
eric-wieser: You would change exactly nothing in your definition of `Adj`, it works just fine after the change to `Type*` as you have it already!

FormalConjectures/ErdosProblems/1035.lean
eric-wieser: Yes, that sounds good to me.

eric-wieser: Or at least, define the predicate in terms of that.

Karanjot786: Applied! Contains is now defined as `∃ f : H →g G, Function.Injective f`

FormalConjectures/ErdosProblems/1035.lean
Karanjot786: Done! Changed to `def hypercube (n : Type*) : SimpleGraph (n → Bool)` and updated theorem calls to use `hypercube (Fin n)`.

FormalConjectures/ErdosProblems/1035.lean
```diff
@@ -0,0 +1,92 @@
...
+[Er93] Erdős, P., _Some of my favourite unsolved problems_. A tribute to Paul Erdős (1993),
+467-478.
+-/
+@[category research open, AMS 05]
+theorem erdos_1035.variant_u :
+    (∃ u : ℕ → ℕ, ∀ n : ℕ, ∀ (G : SimpleGraph (Fin (2^n)))
+      [DecidableRel G.Adj],
+      (∀ v, G.degree v > 2^n - u n) →
+      Contains G (hypercube (Fin n))) ↔ answer(sorry) := by
+  sorry
```
YaelDillies, line 90: These are trivially true, as in there is a trivial lower/upper bound to these questions. You need to make explicit conjectures of the form "This upper/lower bound is true"

FormalConjectures/ErdosProblems/1035.lean
```diff
@@ -0,0 +1,92 @@
...
+  loopless v h := by
+    obtain ⟨i, hi, _⟩ := h
+    exact hi rfl
+
+/--
+A graph $G$ contains (or embeds) a graph $H$ if there exists an injective graph homomorphism
+from $H$ to $G$.
+-/
+def Contains {α β : Type*} (G : SimpleGraph α) (H : SimpleGraph β) : Prop :=
+  ∃ f : H →g G, Function.Injective f
```
YaelDillies, line 46: This already exists in mathlib, can you use it?

Paul-Lez (approved):

Issue #1273, opened 2025-11-25
Erdős problem 56 bounds issue


plby:
### Describe the misformalization
This issue concerns Erdős problem 56, available at https://www.erdosproblems.com/56

Aristotle (from Harmonic) found a simple counterexample to the conjecture with N=k=2 and the set {1,2}.

I think that some condition is needed for k and N to make the conjecture nontrivial.  I have not yet figured out what it is, but I'll try to comment here when I do.

### Trivial proof/disproof

N=k=2, A = {1,2}.  Lean version:

/--
Suppose $A \subseteq \{1,\dots,N\}$ is such that there are no $k+1$ elements of $A$ which are
relatively prime. An example is the set of all multiples of the first $k$ primes.
Is this the largest such set?
-/
theorem erdos_56 : (∀ᵉ (N ≥ 2) (k > 0), (MaxWeaklyDivisible N k = (FirstPrimesMultiples N k).card)) ↔
    False := by
  bound;
  · contrapose! a;
    refine' ⟨ 2, by norm_num, 2, by norm_num, _ ⟩;
    unfold MaxWeaklyDivisible FirstPrimesMultiples;
    refine' ne_of_gt ( lt_of_lt_of_le _ <| le_csSup _ <| show 2 ∈ _ from _ );
    · rw [ Finset.Icc_eq_cons_Ioc ] <;> norm_num;
      rw [ Finset.filter_insert, Finset.filter_singleton ] ; norm_num;
      split_ifs <;> simp_all +decide [ Nat.Prime.ne_one ];
    · exact ⟨ _, fun n hn => hn.choose_spec.2.2 ▸ Finset.card_le_card hn.choose_spec.1 ⟩;
    · simp +decide [ WeaklyDivisible ];
  · norm_num +zetaDelta at *

### Suggested fix
I will try to figure out the right bound on k.

### Additional context
There are at least 2 papers available about this subject, both available online.  I'll be looking through them.
[AhKh94] Ahlswede, Rudolf and Khachatrian, Levon H., On extremal sets without coprimes. Acta Arith. (1994), 89-99. http://matwbn.icm.edu.pl/ksiazki/aa/aa66/aa6617.pdf
[AhKh95] Rudolf Ahlswede, and Levon H. Khachatrian. "Maximal sets of numbers not containing k+1 pairwise coprime integers." Acta Arithmetica 72.1 (1995): 77-100. https://eudml.org/doc/206786

### Choose either option
[...]

PR #1274, opened 2025-11-25
Update theorem erdos_56 with additional condition

plby:
Attempted to fix https://github.com/google-deepmind/formal-conjectures/issues/1273

See also https://www.erdosproblems.com/forum/thread/56#post-1845

### DIFF
--- a/FormalConjectures/ErdosProblems/56.lean
+++ b/FormalConjectures/ErdosProblems/56.lean
@@ -109,10 +109,11 @@ lemma weaklyDivisible_firstPrimesMultiples (N k : ℕ) (hN : 1 ≤ N) :
 /--
 Suppose $A \subseteq \{1,\dots,N\}$ is such that there are no $k+1$ elements of $A$ which are
 relatively prime. An example is the set of all multiples of the first $k$ primes.
-Is this the largest such set?
+Is this the largest such set?  To avoid trivial counterexamples, we must insist that $N$ be at
+least the $k$th prime.
 -/
 @[category research solved, AMS 11]
-theorem erdos_56 : (∀ᵉ (N ≥ 2) (k > 0), k.nth Nat.Prime ≤ N →
+theorem erdos_56 : (∀ᵉ (k > 0) (N ≥ (k-1).nth Nat.Prime),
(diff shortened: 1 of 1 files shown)


eric-wieser:
(since we tried to fix this before in #517)

FormalConjectures/ErdosProblems/56.lean
```diff
@@ -112,7 +112,8 @@ relatively prime. An example is the set of all multiples of the first $k$ primes
 Is this the largest such set?
 -/
 @[category research solved, AMS 11]
-theorem erdos_56 : (∀ᵉ (N ≥ 2) (k > 0), (MaxWeaklyDivisible N k = (FirstPrimesMultiples N k).card)) ↔
+theorem erdos_56 : (∀ᵉ (N ≥ 2) (k > 0), N ≥ k.nth Nat.Prime →
```
smmercuri: I suppose we can get rid of `N ≥ 2` now? 
```suggestion
theorem erdos_56 : (∀ᵉ (N) (k > 0), N ≥ k.nth Nat.Prime →
```

FormalConjectures/ErdosProblems/56.lean
```diff
@@ -112,7 +112,8 @@ relatively prime. An example is the set of all multiples of the first $k$ primes
 Is this the largest such set?
```
smmercuri: Could you also add the new condition to the docstring?

FormalConjectures/ErdosProblems/56.lean
smmercuri: Do we need to be careful about indexing here? if `k = 1` then should this correspond to $p_1 = 2$? If so, then we should use `(k - 1).nth Prime` since it's zero-indexed.

smmercuri: nit: `le` is preferred outside of quantification binders 
```suggestion
theorem erdos_56 : (∀ᵉ (N ≥ 2) (k > 0), k.nth Nat.Prime ≤ N →
```

plby: I think you are correct.

PR #1279, opened 2025-11-27
feat(ErdosProblems): 1043

### LINKED ISSUE #1278: Erdős Problem 1043
### What is the conjecture

https://www.erdosproblems.com/1043

Let $f\in \mathbb{C}[x]$ be a monic polynomial. Must there exist a straight line $\ell$ such that the projection of\[\{ z: \lvert f(z)\rvert\leq 1\}\]onto $\ell$ has measure at most $2$?

Status:  solved

### Choose either option

callesonne:
fixes #1278

### DIFF
--- a/FormalConjectures/ErdosProblems/1043.lean
+++ b/FormalConjectures/ErdosProblems/1043.lean
@@ -43,7 +43,7 @@ Pommerenke [Po61] proved that the answer is no.
 -/
 @[category research solved, AMS 28 30]
 theorem erdos_1043 :
-    (∀ (f : ℂ[X]), f.Monic →
+    (∀ (f : ℂ[X]), f.Monic → f.degree ≥ 1 →
       ∃ (u : ℂ), ‖u‖ = 1 ∧
       volume ((ℝ ∙ u).orthogonalProjection '' levelSet f) ≤ 2) ↔ answer(False) := by
   sorry


FormalConjectures/ErdosProblems/1043.lean
```diff
@@ -0,0 +1,62 @@
...
+Pommerenke [Po61] proved that the answer is no.
+
+[Po61] Pommerenke, Ch., _On metric properties of complex polynomials._ Michigan Math. J. (1961),
+97-115.
+-/
+@[category research solved, AMS 28 30]
+theorem erdos_1043 :
+    ∃ (f : ℂ[X]), f.Monic ∧
+      ∀ (u : ℂ), ‖u‖ = 1 ∧
+      volume ((ℝ ∙ u).orthogonalProjection '' levelSet f) > 2 := by
```
smmercuri: should be this instead otherwise there is a contradiction with any `u` without norm `1`:
```suggestion
      ∀ (u : ℂ), ‖u‖ = 1 →
      volume ((ℝ ∙ u).orthogonalProjection '' levelSet f) > 2 := by
```

FormalConjectures/ErdosProblems/1043.lean
```diff
@@ -0,0 +1,62 @@
...
+  sorry
+
+/--
+On the other hand, Pommerenke also proved there always exists a line such that the projection has
+measure at most 3.3.
+-/
+@[category research solved, AMS 28 30]
+theorem erdos_1043.variants.weak :
+    ∀ (f : ℂ[X]), f.Monic ∧
+      ∃ (u : ℂ), ‖u‖ = 1 ∧
```
smmercuri:
```suggestion
    ∀ (f : ℂ[X]), f.Monic →
      ∃ (u : ℂ), ‖u‖ = 1 ∧
```

smmercuri (changes requested):

FormalConjectures/ErdosProblems/1043.lean
```diff
@@ -0,0 +1,62 @@
...
+      ∀ (u : ℂ), ‖u‖ = 1 ∧
+      volume ((ℝ ∙ u).orthogonalProjection '' levelSet f) > 2 := by
+  sorry
+
+/--
+On the other hand, Pommerenke also proved there always exists a line such that the projection has
+measure at most 3.3.
+-/
+@[category research solved, AMS 28 30]
+theorem erdos_1043.variants.weak :
```
smmercuri, line 56: For this variant, we should exclude constant polynomials

FormalConjectures/ErdosProblems/1043.lean
```diff
@@ -0,0 +1,62 @@
...
+*Reference:* [erdosproblems.com/1043](https://www.erdosproblems.com/1043)
+-/
+
+namespace Erdos1043
+
+open MeasureTheory Polynomial
+
+/-- The set $\{ z \in \mathbb{C} : \lvert f(z)\rvert\leq 1\}$ -/
+def levelSet (f : Polynomial ℂ) : Set ℂ :=
+  {z : ℂ | ‖Polynomial.eval z f‖ ≤ 1}
```
eric-wieser: Could abbreviate to `f.eval z`, maybe?

FormalConjectures/ErdosProblems/1043.lean
callesonne: Thanks for spotting!

FormalConjectures/ErdosProblems/1043.lean
```diff
@@ -0,0 +1,62 @@
...
+Pommerenke [Po61] proved that the answer is no.
+
+[Po61] Pommerenke, Ch., _On metric properties of complex polynomials._ Michigan Math. J. (1961),
+97-115.
+-/
+@[category research solved, AMS 28 30]
+theorem erdos_1043 :
+    ∃ (f : ℂ[X]), f.Monic ∧
+      ∀ (u : ℂ), ‖u‖ = 1 →
+      volume ((ℝ ∙ u).orthogonalProjection '' levelSet f) > 2 := by
```
eric-wieser: Doesn't this say "greater than two" instead of "at most 2"?

YaelDillies: The answer is "no", so Calle negated the statement

eric-wieser: Don't we normally write this as `answer(False)`? I think it's better for the negation to be explicit, so the formalization matches the question, not just the answer

mo271: we have not been completely consistent about it, but I agree with @eric-wieser

callesonne:
@smmercuri could you resolve your changes requested? I will fix the final comment by Eric and then its ready to be merged I think.

Issue #1305, opened 2025-11-29
Erdős Problem 1108


mo271:
### What is the conjecture

https://www.erdosproblems.com/1108

  Let 
$$A = \left\\{ \sum_{n\in S}n! : S\subset \mathbb{N}\textrm{ finite}\right\\}.$$ 
If $k\geq 2$, then does $A$ contain only finitely many $k$-th powers? Does it contain only finitely many powerful numbers?

Status:open

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

PR #1318, opened 2025-12-02
Erdos 1108

ayush31010:
This PR adds the formalization of Erdős Problem 1108.#1305
The problem studies the set 

𝐴={∑𝑛∈𝑆𝑛!:𝑆⊂𝑁 finite}, A={n∈S∑n!:S⊂N `finite},

and asks two open questions:

1.For each integer 𝑘≥2, are there only finitely many perfect k-th powers in A?

2.Are there only finitely many powerful numbers in A (those in which every prime divisor appears with exponent ≥ 2)?

Included in this PR

- Definition of Nat.FactorialSums.

- Predicates for Nat.IsKthPower and Nat.IsPowerful.

- Subsets of factorial sums that are perfect powers or powerful.

- Conjecture stubs for both parts, marked as research open.

- A few basic test lemmas for sanity checks.

The file follows the style of other Erdős problem entries.
I intend to continue working on this problem, so this sets up the structure for future progress.


mo271:
This does not compile

PR #1331, opened 2025-12-03
fix: `OmegaPowerRamsey` should keep track of the order type

mo271:
This is now not just checking the cardinality but the order type

### DIFF
--- a/FormalConjectures/ForMathlib/SetTheory/Cardinal/SimpleGraph.lean
+++ b/FormalConjectures/ForMathlib/SetTheory/Cardinal/SimpleGraph.lean
@@ -15,7 +15,6 @@ limitations under the License.
 -/
 
 import Mathlib.Combinatorics.SimpleGraph.Clique
-import Mathlib.Order.CompletePartialOrder
 import Mathlib.SetTheory.Ordinal.Exponential
 
 open Cardinal Ordinal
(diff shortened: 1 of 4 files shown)


FormalConjectures/ForMathlib/SetTheory/Cardinal/SimpleGraph.lean
```diff
@@ -20,25 +20,26 @@ import Mathlib.SetTheory.Ordinal.Exponential
...
 open Cardinal Ordinal
 
-open scoped Ordinal
-
 universe u
 
 /--
-This proposition asserts the Ramsey property `κ → (κ, c)₂`, where `κ` is the
-cardinality of the ordinal `ω ^ β` and `c` is some cardinal.
+This proposition asserts the ordinal Ramsey property `ω^β → (ω^β, c)²`.
```
smmercuri: This `2` was a subscript previously, which is the right notation here?

FormalConjectures/ForMathlib/SetTheory/Cardinal/SimpleGraph.lean
```diff
@@ -20,25 +20,26 @@ import Mathlib.SetTheory.Ordinal.Exponential
...
+* There is a `G_blue` clique of cardinality `c`.
 -/
 def OmegaPowerRamsey (β : Ordinal.{u}) (c : Cardinal.{u}) : Prop :=
-  let κ := (ω ^ β).card
-  ∀ (V : Type u),  κ = Cardinal.mk V →
-  -- For any red/blue edge coloring of the complete graph on V...
-  -- (represented by two graphs G_red and G_blue that are complements)
+  let α_ord := ω ^ β
+  -- We require V to have an order (LinearOrder)
+  ∀ (V : Type u) [LinearOrder V] [LE (Quotient.out α_ord).α],
```
smmercuri: Why is `LinearOrder` required vs `[LE V]`? I suppose `V` will inherit a linear order from the order isomorphism, but apriori it needn't have it.

smmercuri: It seems like `[LE (Quotient.out α_ord).α]` shouldn't be required, given that it's the underlying type of a well order. Indeed, you should use `α_ord.toType` instead which has the required instances

smmercuri: Or looking at the [latest docs](https://leanprover-community.github.io/mathlib4_docs/Mathlib/SetTheory/Ordinal/Basic.html#Ordinal.ToType), we should use `Set.Iio α` instead perhaps?

FormalConjectures/ForMathlib/SetTheory/Cardinal/SimpleGraph.lean
mo271:
```suggestion
This proposition asserts the ordinal Ramsey property `ω^β → (ω^β, c)₂`.
```

mo271: done

YaelDillies:
Sorry for hijacking the PR. There were many things wrong with the current definition from the point of view of formalising further ordinal Ramsey properties, so I rewrote it all. In the future we could add a proper Rado arrow notation, that would be fun.

mo271 (commented):
LGTM

PR #1337, opened 2025-12-04
fix: `erdos_853`

smmercuri:
- Need to exclude `t = 0` from `r`, otherwise `r n = 0` always.
- Compactify statements using `Filter.Tendsto`

### DIFF
--- a/FormalConjectures/ErdosProblems/853.lean
+++ b/FormalConjectures/ErdosProblems/853.lean
@@ -40,8 +40,7 @@ integer $t$ such that $d_n = t$ has no solutions for $n \le x$.
 Is it true that $r(x) \to \infty$?
 -/
 @[category research open, AMS 11]
-theorem erdos_853 :
-    atTop.Tendsto r atTop := by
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/853.lean
```diff
@@ -39,7 +41,8 @@ Is it true that $r(x) \to \infty$?
 -/
 @[category research open, AMS 11]
 theorem erdos_853 :
-  ∀ (M : ℕ), ∃ (X : ℕ), ∀ x ≥ X, r x > M := by sorry
+    atTop.Tendsto r atTop := by
+  sorry
```
callesonne, line 44: I don't see how these are different? Also below? If its true for sufficiently large `M`, then should it not be true for any `M' < M` as well using the same `X`?

smmercuri: Oh yes, you're right these are equivalent. The only issue was the inclusion of `t = 0` in the definition of `r`. May as well keep the `Tendsto` spelling now it's done

Issue #1350, opened 2025-12-05
Erdős problem 137: flipped quantifiers


plby:
### Describe the misformalization

This concerns Erdős problem 137 which can be found at https://www.erdosproblems.com/137

The original source is https://renyi.hu/~p_erdos/1982-08.pdf

I think this is incorrect:
https://github.com/google-deepmind/formal-conjectures/blob/37a34a76ea03c9ab157981bf4495a8c3a1add68a/FormalConjectures/ErdosProblems/137.lean#L49-L61

I think both the comment and the Lean should have a different quantifier order.  It currently says "∀ m k, ∃ n₀, ∀ n" but I think it should be "∀  k, ∃ n₀, ∀ n m".

### Trivial proof/disproof
I think the current order is an okay undergraduate exercise, but not the intended conjecture.

### Suggested fix
Introduce `m` later in the statement, with the quantifier "for all".

### Additional context
This was found by Aristotle.

### Choose either option
- [ ] I plan on working on this issue
- [x] This issue is up for grabs: I would like to see this misformalization fixed by somebody else

shamEiNew:
I've added a fix for this in pr https://github.com/google-deepmind/formal-conjectures/pull/1353

shamEiNew:
I think this issue can be closed as #1353 is merged. @Paul-Lez

PR #1353, opened 2025-12-06
Erdos 137: flipped quantifiers fix

shamEiNew:
This is the fix for the issue https://github.com/google-deepmind/formal-conjectures/issues/1350. 
Changed the order of quantifiers from "∀ m k, ∃ n₀, ∀ n" to "∀ k, ∃ n₀, ∀ n m" as per the original [source](https://renyi.hu/~p_erdos/1982-08.pdf) in page 27. 

It's stated clearly that 
"... In fact I would conjecture that for every $k$ there is an $n_0$ so that for $n_0 > n$ at least $k$ of the them have exponents must be $1$...."

### DIFF
--- a/FormalConjectures/ErdosProblems/137.lean
+++ b/FormalConjectures/ErdosProblems/137.lean
@@ -47,14 +47,14 @@ theorem erdos_137.variants.perfect_power (k : ℕ) (hk : k ≥ 2) (n : ℕ) (x l
   sorry
 
 /--
-Erdős [Er82c] conjectures that, if $k$ is fixed, then for all $n$ sufficiently large and
-all positive integers $m$, then there must be at least $k$ distict primes $p$ such that
+Erdős [Er82c] conjectures that, if $k$ is fixed, then for all $n$ sufficiently large and all
+positive integers $m$, there must be at least $k$ distinct primes $p$ such that
 $p\mid m(m+1)\cdots (m+n)$ and yet $p^2$ does not divide the right hand side.
 
 [Er82c] Erdős, Paul, "Miscellaneous problems in number theory". Congr. Numer. (1982), 25-45.,
 -/
 @[category research open, AMS 11]
-theorem erdos_137.multiple_powerful_factors (k : ℕ) : ∃ (n₀ : ℕ), ∀ n > n₀,
+theorem erdos_137.multiple_powerful_factors (k : ℕ) : ∀ᶠ n in Filter.atTop,
     ∀ (m : ℕ) (hm : 0 < m),
     letI N := ∏ x ∈ Finset.Ioc m (m + n), x
     ∃ P : Finset ℕ, P.card = k ∧ ∀ p ∈ P, p.Prime ∧


plby:
The issue from the original source is the position of the quantifier on `m`, which is somewhat implicit.

FormalConjectures/ErdosProblems/137.lean
```diff
@@ -47,14 +47,16 @@ theorem erdos_137.variants.perfect_power (k : ℕ) (hk : k ≥ 2) (n : ℕ) (x l
   sorry
 
 /--
-Erdős [Er82c] conjectures that, if $m$, $k$ are fixed and $n$ sufficiently large, then there must
+Erdős [Er82c] conjectures that, if $k$ are fixed, $n$ sufficiently large and $m$ be any
+positive integer, then there must
 be at least $k$ distict primes $p$ such that $p\mid m(m+1)\cdots (m+n)$ and yet $p^2$ does not
```
YaelDillies:
```suggestion
Erdős [Er82c] conjectures that, if $k$ is fixed, then for all $n$ sufficiently large and
all positive integers $m$, then there must be at least $k$ distinct primes $p$ such that
$p\mid m(m+1)\cdots (m+n)$ and yet $p^2$ does not divide the right hand side.
```

shamEiNew: I've made changes as recommended in the latest [commit ](https://github.com/google-deepmind/formal-conjectures/pull/1353/commits/11fbfd13ce7b5a36a3819bf8795bf34b07ba286e)

FormalConjectures/ErdosProblems/137.lean
```diff
@@ -47,14 +47,15 @@ theorem erdos_137.variants.perfect_power (k : ℕ) (hk : k ≥ 2) (n : ℕ) (x l
...
-divide the right hand side.
+Erdős [Er82c] conjectures that, if $k$ is fixed, then for all $n$ sufficiently large and
+all positive integers $m$, then there must be at least $k$ distict primes $p$ such that
+$p\mid m(m+1)\cdots (m+n)$ and yet $p^2$ does not divide the right hand side.
 
 [Er82c] Erdős, Paul, "Miscellaneous problems in number theory". Congr. Numer. (1982), 25-45.,
 -/
 @[category research open, AMS 11]
-theorem erdos_137.multiple_powerful_factors (m k : ℕ) (hm : 0 < m): ∃ (n₀ : ℕ), ∀ n > n₀,
+theorem erdos_137.multiple_powerful_factors (k : ℕ) : ∃ (n₀ : ℕ), ∀ n > n₀,
```
YaelDillies: Maybe take the opportunity to use filter notation?
```suggestion
theorem erdos_137.multiple_powerful_factors (k : ℕ) : ∀ᶠ n in atTop,
```

FormalConjectures/ErdosProblems/137.lean
```diff
@@ -47,14 +47,15 @@ theorem erdos_137.variants.perfect_power (k : ℕ) (hk : k ≥ 2) (n : ℕ) (x l
   sorry
 
 /--
-Erdős [Er82c] conjectures that, if $m$, $k$ are fixed and $n$ sufficiently large, then there must
-be at least $k$ distict primes $p$ such that $p\mid m(m+1)\cdots (m+n)$ and yet $p^2$ does not
-divide the right hand side.
+Erdős [Er82c] conjectures that, if $k$ is fixed, then for all $n$ sufficiently large and
+all positive integers $m$, then there must be at least $k$ distict primes $p$ such that
```
YaelDillies:
```suggestion
all positive integers $m$, then there must be at least $k$ distinct primes $p$ such that
```

FormalConjectures/ErdosProblems/137.lean
shamEiNew: I've made the change in the latest commit.

PR #1358, opened 2025-12-08
Update theorem statement in 330.lean

Woett:
The original statement on the Erdos problem website was incorrect. This is now fixed, and the formalisation should be fixed as well.

### DIFF
--- a/FormalConjectures/ErdosProblems/330.lean
+++ b/FormalConjectures/ErdosProblems/330.lean
@@ -42,13 +42,13 @@ def MinAsymptoticAddBasis (A : Set ℕ) : Prop :=
   IsAsymptoticAddBasis A ∧ ∀ n ∈ A, ¬ IsAsymptoticAddBasis (A \ {n})
 
 /--
-Suppose $A \subset \mathbb{N}$ is a minimal basis with positive density.
-Is it true that, for any $n \in A$, the (upper) density of integers which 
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/330.lean
```diff
@@ -42,13 +42,13 @@ def MinAsymptoticAddBasis (A : Set ℕ) : Prop :=
...
-Suppose $A \subset \mathbb{N}$ is a minimal basis with positive density.
-Is it true that, for any $n \in A$, the (upper) density of integers which 
+Does there exist a minimal basis $A \subset \mathbb{N}$ with positive density
+such that, for any $n \in A$, the (upper) density of integers which 
 cannot be represented without using $n$ is positive?
 -/
 @[category research open, AMS 5 11]
 theorem erdos_330_statement :
-    (∀ (A : Set ℕ),  MinAsymptoticAddBasis A → A.HasPosDensity →
+    (∃ (A : Set ℕ),  MinAsymptoticAddBasis A → A.HasPosDensity →
```
Paul-Lez: (otherwise the statement can be proven by taking `A` to be a set that doesn't satisfy `MinAsymptoticAddBasis A`!)
```suggestion
    (∃ (A : Set ℕ),  MinAsymptoticAddBasis A ∧ A.HasPosDensity ∧
```

Paul-Lez (approved):
Thanks!

Issue #1359, opened 2025-12-08
Novikov Conjecture:invariance of higher signatures


SrijanRoy123-github:
### What is the conjecture
https://en.wikipedia.org/wiki/Novikov_conjecture

The Novikov Conjecture asserts that for any closed, oriented manifold $M$ with fundamental group $\pi_1(M)$, all of its higher signatures remain invariant under oriented homotopy equivalence.

Equivalently, via the Mishchenko–Fomenko and Kasparov formulations, tThe conjecture states that the topological K-theory assembly map

$$
K_\ast(B\pi_1(M)) \longrightarrow K_\ast(C_r^\ast(\pi_1(M)))
$$

is rationally injective.

This is one of the central open problems in topology and noncommutative geometry, connected to the Baum–Connes conjecture, positive scalar curvature, and rigidity theory.

---

### Prerequisites needed

Since the full analytic machinery is not yet available in mathlib, the conjecture can be included using abstract placeholders such as:

```lean
constant NovikovConjecture (G : Type _) [Group G] : Prop
```

### [AMS categories](https://github.com/google-deepmind/formal-conjectures/labels?q=ams-)

* ams-XX

### Choose either option
- [x] I plan on adding this conjecture to the repository  
- [ ] This issue is up for grabs: I would like to see this conjecture added by somebody else

SrijanRoy123-github:
@maintainers When you have the time, could you please take a look and let me know if any adjustments are needed?
I’d be very happy to revise the file according to your feedback.

PR #1361, opened 2025-12-08
Adding Novikov conjecture to the list

SrijanRoy123-github:
**Closes Issue #1359 **

## Changes

- Added a new file:  
  `FormalConjectures/Wikipedia/NovikovConjecture.lean`

Hope this works. I’ve added a formalisation of the **Novikov Conjecture** based on its global formulation on Wikipedia.

In this version, I model the conjecture by an abstract predicate on discrete groups and record the statement that it holds for every group. Since the conjecture is open, I mark it as `research open` and use `answer(sorry)`.

### DIFF
--- a/FormalConjectures/ErdosProblems/330.lean
+++ b/FormalConjectures/ErdosProblems/330.lean
@@ -42,13 +42,13 @@ def MinAsymptoticAddBasis (A : Set ℕ) : Prop :=
   IsAsymptoticAddBasis A ∧ ∀ n ∈ A, ¬ IsAsymptoticAddBasis (A \ {n})
 
 /--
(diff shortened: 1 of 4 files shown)


mo271:
I think it is not useful to add this with a placeholder definition of NovikovConjectureForGroup and essentially nothing else. We can add that file when the appropriate definition is in mathlib. Le'ts re-open it then. 

Also you unintentionally change 3 files in this pull request.

SrijanRoy123-github:
Thanks for pointing out. I thought of including the surrounding analytic/K-theoretic setup. Shall wait for it's mathlib inclusion.Thanks very much.

Issue #1302, opened 2025-11-29
Erdős Problem 1105


mo271:
### What is the conjecture

https://www.erdosproblems.com/1105

  The anti-Ramsey number $\mathrm{AR}(n,G)$ is the maximum possible number of colours in which the edges of $K_n$ can be coloured without creating a rainbow copy of $G$ (i.e. one in which all edges have different colours).Let $C_k$ be the cycle on $k$ vertices. Is it true that $$\mathrm{AR}(n,C_k)=\left(\frac{k-2}{2}+\frac{1}{k-1}\right)n+O(1)?$$ Let $P_k$ be the path on $k$ vertices and $\ell=\lfloor\frac{k-1}{2}\rfloor$. If $n\geq k\geq 5$ then is $\mathrm{AR}(n,P_k)$ equal to $$\max\left(\binom{k-2}{2}+1, \binom{\ell-1}{2}+(\ell-1)(n-\ell+1)+\epsilon\right)$$ where $\epsilon=1$ if $k$ is odd and $\epsilon=2$ otherwise?

Status:open

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

Arsh123344423:
@mo271 can i work upon this?

jdhruv555:
Hi, I have done something similar before. Im interested to work on this.

PR #1362, opened 2025-12-08
Erdos-1105

ayush31010:
Solved #1302


mo271:
This is not a correct formalization: it only contains placeholder definitions and no mathematical content. I consider this spam.

Issue #171, opened 2025-06-06
The Gap Conjecture


Paul-Lez:
### What is the conjecture
See [here](https://en.wikipedia.org/wiki/Gromov%27s_theorem_on_groups_of_polynomial_growth#The_gap_conjecture) and the references thereof.

### Prerequisites needed
Seems doable with the current state of Mathlib.

edwag:
I'm interested in taking a look at this but it looks like it might require a fair amount of setup so I wanted to check what we think is required here. I think we need, but don't have, the notion of the growth rate of a group, which will itself require us to have a notion of the minimal word length of a group element, which requires us to be able to talk about a finite symmetric set of generators. Mathlib does already have a definition for "finitely generated group", but I think everything else that I've just listed is absent and would need to be implemented in order to state this conjecture. Does that sound right? If so, would it then be sensible for all of those supporting definitions to live inside `ForMathlib.Algebra.Group`?

edwag:
I've started taking a look at this and my current rough plan is:

- Define the "word $N$-ball" with respect to a finite subset $S$ of a group $G$, i.e. the set of all elements which can be represented as products of $\leq N$ elements from $S$.
- Prove that, for two finite generating sets $S$, $S^\prime$, the growth rate of the $N$-ball wrt $S$ is big $O$ equivalent to the growth rate wrt $S^\prime$ (I think this is true, and I don't need to additionally assume that these generating sets are symmetric?).
- Define the big $O$ growth rate of the $N$-ball for a finitely generated group $G$, independent of a particular generating set.

Once that's done, I think we would be in a position to state the conjecture pretty straightforwardly. I am not 100% sure how to achieve all of this, and I'm also not sure how quickly I will be able to get around to actually implementing it. And I have no idea at all where it should all live.

@Paul-Lez I would appreciate any thoughts you have on this plan 😃

edwag:
I've put up a PR for this at https://github.com/google-deepmind/formal-conjectures/pull/1366. I may have got a bit carried away with implementing the machinery for talking about the growth rate of a group 😅

Paul-Lez:
Very cool! Thanks for adding this (and sorry for missing your initial messages in this issue...).

Smetalo:
Is this resolved in #1366?

PR #1366, opened 2025-12-08
feat: growth rate of groups and the Gap Conjecture

edwag:
This is an attempt to implement https://github.com/google-deepmind/formal-conjectures/issues/171. To support the statement of the conjecture, I've put together a formalisation of the concept of the growth rate of a finitely generated group - this machinery makes up almost all of the content of this PR.


YaelDillies (commented):
I'm sorry, but it looks like a lot of your code might be duplicating existing work!
1. Look at `GrowthFunction` in this repo to replace `growthRate_of_wordBalls`
2. Look at `Mathlib.Analysis.Asymptotics.Asymptotics` for a systematic treatment of growth of functions that should substitute to your `GrowthClass`.

edwag:
> I'm sorry, but it looks like a lot of your code might be duplicating existing work!
> 
> 1. Look at `GrowthFunction` in this repo to replace `growthRate_of_wordBalls`
> 2. Look at `Mathlib.Analysis.Asymptotics.Asymptotics` for a systematic treatment of growth of functions that should substitute to your `GrowthClass`.

Thanks for pointing this out!

1. This all looks relevant, including all of the `CayleyBall` definitions, and should probably be pulled out into `ForMathlib`? I'm not sure what's the right way to combine this with my `GrowthRate.lean`, which contains a lot more content. I think most of `GrowthRate.lean` is still worth adding to the repo in some form.
2. I did take a look at `Mathlib.Analysis.Asymptotics`, but I don't think any of this covers the notion of growth that's used here. As far as I can see this all covers notions like big O, where you have bounds that look something like $f(x) \le C \cdot g(x)$. Here we have bounds like $f(x) \le g(C \cdot x)$, which is a different notion. E.g. big O distinguishes $e^x$ from $e^{2x}$, but `GrowthClass` doesn't. Please let me know if I'm missing something in Mathlib which does cover the required notion of equivalent growth.

YaelDillies:
1. Yes, feel free to move `GrowthFunction` to ForMathlib. What is not covered by `GrowthFunction` seems to exist already in mathlib. A cursory look at your code makes me think you are reinventing the `FreeGroup` API.
2. If you take log on both sides, then mathlib's asymptotics library lets you define superexponential functions, right?

edwag:
1. To your `FreeGroup` point, it's certainly related, but I don't know exactly how one would use the existing `FreeGroup` API to express what's in here. The point of `GrowthRate.lean` is really to (a) establish the ability to talk about Cayley balls (which are currently called `wordBall` in my code), and (b) establish that the growth rate of the Cayley balls for a group is independent of the choice of generating set, so that the growth rate of a group in terms of the growth rate of its Cayley balls is well defined. This feels sufficiently different from free groups that I think it's likely worth having specific definitions/lemmas for it (though perhaps this could/should be built on `FreeGroup` under the hood?). I don't know the `FreeGroup` API at all, nor do I know the philosophy for what should live in Mathlib, so I'm interested to hear your thoughts!
2. I don't think I follow. I can't fruitfully take log on both sides of $f(x) \le g(C \cdot x)$, where $f$ and $g$ are arbitrary functions. We also don't need to talk about superexponential functions here, since the growth rate of a group is at most exponential.

YaelDillies:
1. My viewpoint is that a Cayley graph is no more than a surjective hom from a free group, and Cayley balls are images of balls in said free group.
2. Sorry, I meant superpolynomial. All I'm thinking is that your notion of function growth is non-standard, and you're applying it to a case that could just as well be stated by hand.

FormalConjectures/ForMathlib/Algebra/Group/GrowthRate.lean
```diff
@@ -0,0 +1,726 @@
...
+import Mathlib.Data.Finset.Lattice.Basic
+import Mathlib.Data.Finset.Union
+import Mathlib.Data.Fintype.BigOperators
+import Mathlib.Data.Real.Basic
+import Mathlib.GroupTheory.Finiteness
+import Mathlib.Order.BoundedOrder.Basic
+import Mathlib.Order.Filter.AtTopBot.Basic
+import Mathlib.Order.Filter.AtTopBot.Defs
+import Mathlib.Order.Filter.Basic
+import Mathlib.Order.MinMax
```
mo271:
```suggestion
import Mathlib.Data.Fin.VecNotation
import Mathlib.Data.Fintype.BigOperators
import Mathlib.Data.Real.Basic
import Mathlib.GroupTheory.Finiteness
import Mathlib.Order.Filter.AtTopBot.Basic
```

`#min_imports`

mo271 (commented):
This would be fantastic to have @edwag!
Having the growth rate would allow us to state many conjectures from the Kourovka notebook!

Thanks for the contributions. I see there are still some unaddressed comments from @YaelDillies, but this looks very good.

edwag:
> This would be fantastic to have @edwag! Having the growth rate would allow us to state many conjectures from the Kourovka notebook!
> 
> Thanks for the contributions. I see there are still some unaddressed comments from @YaelDillies, but this looks very good.

@mo271 I'm keen to spend some time getting this finished off, but I'm unclear on how to proceed. I believe the main contributions of this PR, and the things that we want to keep in some form, are:

1. The definition of `growthRate`
2. The proof that `growthRate` is well defined

From the discussion above, I think that @YaelDillies may disagree that these are needed (and possibly that they are currently absent from Mathlib?). One point, I think, is that this conjecture specifically could be stated without setting up the whole machinery for `growthRate`, but I think based on your comment above that `growthRate` would also be useful more broadly?

Another point that @YaelDillies made is that all of this can be thought of in terms of free groups. I'm certainly happy to take a crack at setting up the definitions here to be based on the Mathlib free group API, if we think that would be valuable (though I'd likely need some guidance on which parts of the free group API are relevant here).

(Also @YaelDillies sorry for disappearing before resolving the discussion above. Thanks for going back and forth with me - it's enormously valuable to get expert feedback and I really appreciate it :) )

YaelDillies:
The issue is that your notion of `GrowthClass` is non-standard: eg `log x` and `2 log x` are considered separate growth rates by your definition. Furthermore, from a cursory reading of https://en.wikipedia.org/wiki/Gromov%27s_theorem_on_groups_of_polynomial_growth#The_gap_conjecture you won't even need the full power of `GrowthClass` since all you'll want is something like `VirtuallyNilpotent G \iff growthFun G =O[atTop] f`.

Let me put it another way: One can express all possible asymptotic behaviors using `Exists`, `Forall` and inequalities. Mathlib decided to layer `Filter.Eventually` and asymptotic notation on top of the bare quantifier/inequality soup because these turn out to be a useful language that can express most things of interest, and they are standard in mathematical practice. Your framework in comparison is neither standard nor have you demonstrated that it is useful. Therefore we should not have it.

Note that this discussion also applies to #1481 to some extent, which is why progress there has been slow: We agree that mathlib's language isn't general enough to express everything we care about, but we aren't yet sure how to improve on it.

Unrelatedly, please merge master and subsequently fix the build. We have moved to v4.27.0 and adopted the module system 🎉

FormalConjectures/Wikipedia/GapConjecture.lean
```diff
@@ -0,0 +1,46 @@
...
+
+/-
+If the growth rate of a finitely generated group $G$ is superpolynomial, then the growth rate is at
+least $e^{\sqrt{n}}$.
+-/
+@[category research open, AMS 20]
+theorem gap_conjecture :
+  ∀ (G : Type) [Group G] [Group.FG G],
+    SuperPolynomial (growthRate G) → growthRate G ≥ expSqrt := by
+      sorry
```
Paul-Lez:
```suggestion
  sorry
```

Paul-Lez (commented):
@edwag 

I think it would be nice to get this merged soon! Here's what I suggest: 
- for now, let's work with Filter.Eventually and state the conjecture in terms of that
- Make a separate PR for the GrowthRate infrastructure. This seems like a nice addition, but I think the current definition is quite non-standard, so we should try to find a nice way to encapsulate all use cases (without delaying adding this conjecture further!). One such way could involve defining the growth rate equivalence classes associated to an arbitrary equivalence relation on functions, and then specialise to your use case. That being said, it's not clear to me that this is better in any way than just working with the bare equivalence relation (i.e. are there theorems or definitions where we need to work with the whole equivalence class of `g`, rather than just saying something like "let `f ~ g` ..."

edwag:
@YaelDillies Thanks for the explanation, this is all clear now! 🙂

@Paul-Lez I'm happy to proceed as you suggest, but I may not have time to get to this until June.

Paul-Lez (approved):

PR #1377, opened 2025-12-11
fix erdos_1003.variants.eps87

mo271:


FormalConjectures/ErdosProblems/1003.lean
```diff
@@ -45,16 +46,18 @@ theorem erdos_1003.variants.Icc :
     (∀ k ≥ 1, Set.Infinite {n | ∀ i ∈ Set.Icc 1 k, φ n = φ (n + i)}) ↔ answer(sorry) := by
   sorry
 
+
+
```
callesonne:
```suggestion
```
Should these be here?

PR #1378, opened 2025-12-11
feat(ErdosProblems): 1091

### LINKED ISSUE #1290: Erdős Problem 1091
### What is the conjecture

https://www.erdosproblems.com/1091

  Let $G$ be a $K_4$-free graph with chromatic number $4$. Must $G$ contain an odd cycle with at least two diagonals? More generally, is there some $f(r)\to \infty$ such that every graph with chromatic number $4$, in which every subgraph on $\leq r$ vertices has chromatic number $\leq 3$, contains an odd cycle with at least $f(r)$ diagonals?

Status:open

### Choose either option

ayush31010:
closes #1290

### DIFF
--- a/.github/workflows/build-and-docs.yml
+++ b/.github/workflows/build-and-docs.yml
@@ -33,7 +33,7 @@ jobs:
     steps:
 
       - name: Checkout project
-        uses: actions/checkout@v4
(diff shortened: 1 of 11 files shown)


ayushdebnath012:
@YaelDillies

PR #1382, opened 2025-12-12
fix(ErdosProblems/14): correctly bracket the statement

YaelDillies:


mo271 (approved):

YaelDillies:
Should we switch to `answer(sorry) \iff statement` to avoid further bracketing issues?

PR #1397, opened 2025-12-18
Erdős Prolem 248: update solution status

mo271:


FormalConjectures/ErdosProblems/248.lean
```diff
@@ -30,8 +32,8 @@ namespace Erdos248
 Are there infinitely many $n$ such that $\omega(n + k) \ll k$ for all $k \geq 1$?
 Here $\omega(n)$ is the number of distinct prime divisors of $n$.
 -/
-@[category research open, AMS 11]
-theorem erdos_248 : (∃ C > (0 : ℝ), { n | ∀ k ≥ 1, ω (n + k) ≤ C * k }.Infinite) ↔ answer(sorry) := by
+@[category research solved, AMS 11]
+theorem erdos_248 : (∃ C > (0 : ℝ), { n | ∀ k ≥ 1, ω (n + k) ≤ C * k }.Infinite) := by
```
Paul-Lez, line 36: nit: I think our convention is to use `answer(True)` since this is phrased as a question?
```suggestion
theorem erdos_248 : (∃ C > (0 : ℝ), { n | ∀ k ≥ 1, ω (n + k) ≤ C * k }.Infinite) ↔ answer(True) := by
```

Paul-Lez (approved):

PR #1398, opened 2025-12-18
feat: tensor networks

robinsong2:


mo271:
we should soon be at v4.27.0...

YaelDillies:
We have now bumped the repo past the mathlib commit required to make progress here.

PR #1402, opened 2025-12-20
feat(ErdosProblems): 38

### LINKED ISSUE #217: Erdős Problem 38
### What is the conjecture
    
https://www.erdosproblems.com/38

  Does there exist $B\subset\mathbb{N}$ which is not an additive basis, but is such that for every set $A\subseteq\mathbb{N}$ of Schnirelmann density $\alpha$ and every $N$ there exists $b\in B$ such that 
$$\lvert (A\cup (A+b))\cap \{1,\ldots,N\}\rvert\geq (\alpha+f(\alpha))N$$
where $f(\alpha)>0$ for $0<\alpha <1 $? The Schnirelmann density is defined by 
$$d_s(A) = \inf_{N\geq 1}\frac{\lvert A\cap\{1,\ldots,N\}\rvert}{N}.$$

Status: open

### Prerequisites needed
Schnirelmann density (check if it is also used in other problems)

### Choose either option

vltanh:
Closes #217

### DIFF
--- a/FormalConjectures/ErdosProblems/38.lean
+++ b/FormalConjectures/ErdosProblems/38.lean
@@ -24,7 +24,7 @@ import FormalConjectures.Util.ProblemImports
 
 open Classical Set Pointwise
 
-section Erdos38
+namespace Erdos38
 
 /--
 Does there exist $B \subset \mathbb{N}$ which is not an additive basis,


FormalConjectures/ErdosProblems/38.lean
```diff
@@ -0,0 +1,43 @@
...
+Does there exist a set B ⊆ ℕ which is not an additive basis but guarantees a density increase?
+-/
+@[category research open, AMS 11]
+theorem erdos_38 :
+    (
+      ∃ B : Set ℕ, ¬ B.IsAddBasis ∧
+      ∃ f : ℝ → ℝ, (∀ α : ℝ, 0 < α → α < 1 → f α > 0) ∧
+        ∀ A : Set ℕ, let α := schnirelmannDensity A
+        ∀ N : ℕ,
+          ∃ b ∈ B, ((Ioc 0 N) ∩ (A ∪ (A + {b}))).ncard ≥ (α + f α) * N
```
YaelDillies:
```suggestion
          ∃ b ∈ B, (Ioc 0 N ∩ (A ∪ (A + {b}))).ncard ≥ (α + f α) * N
```

FormalConjectures/ErdosProblems/38.lean
```diff
@@ -0,0 +1,43 @@
...
+    ) ↔ answer(True) := by
```
YaelDillies:
```suggestion
    ) ↔ answer(sorry) := by
```
since we don't know whether it is true or false

vltanh: For existence, if I understand correctly, I think it is true (e.g. Linnik's work as discuss in [Erdos 38](https://www.erdosproblems.com/38)). 
Also discussed there is the interest in if we can improve the lower bound, i.e., find a different B, but I don't know how to formalize that part.

YaelDillies: If I read the discussion correctly, Linnik's work proves the existence of something weaker:
> This is a stronger property than B being an essential component (see [[37]](https://www.erdosproblems.com/37)). Linnik [[Li42]](https://www.erdosproblems.com/38#cite-Li42-38) gave the first construction of an essential component which is not an additive basis.

At any rate, `answer(True)` is incompatible with `category research solved`

vltanh: I see. I have misread it.

FormalConjectures/ErdosProblems/38.lean
```diff
@@ -0,0 +1,43 @@
...
+
+*Reference:* [erdosproblems.com/38](https://www.erdosproblems.com/38)
+-/
+
+open Classical Set Pointwise
+
+section Erdos38
+
+/--
+Does there exist a set B ⊆ ℕ which is not an additive basis but guarantees a density increase?
```
YaelDillies: Could you be clearer about what "guarantees a density increase" mean here? Personally I am an additive combinatorialist, yet I couldn't tell what that meant.

FormalConjectures/ErdosProblems/38.lean
```diff
@@ -0,0 +1,43 @@
...
+-/
+@[category research open, AMS 11]
+theorem erdos_38 :
+    (
+      ∃ B : Set ℕ, ¬ B.IsAddBasis ∧
+      ∃ f : ℝ → ℝ, (∀ α : ℝ, 0 < α → α < 1 → f α > 0) ∧
+        ∀ A : Set ℕ, let α := schnirelmannDensity A
+        ∀ N : ℕ,
+          ∃ b ∈ B, (Ioc 0 N ∩ (A ∪ (A + {b}))).ncard ≥ (α + f α) * N
+    ) ↔ answer(sorry) := by
```
YaelDillies:
```suggestion
theorem erdos_38 : answer(sorry) ↔
    ∃ B : Set ℕ, ¬ B.IsAddBasis ∧ ∃ f : ℝ → ℝ, (∀ α, 0 < α → α < 1 → f α > 0) ∧
      ∀ (A : Set ℕ) (N : ℕ),
        let α := schnirelmannDensity A
        ∃ b ∈ B, (Ioc 0 N ∩ (A ∪ (A + {b}))).ncard ≥ (α + f α) * N := by
```
feels more readable

FormalConjectures/ErdosProblems/38.lean
YaelDillies: Also please use latex for those problem docstrings!

vltanh: I was trying to simplify it but maybe I shouldn't have. The idea was that which ever b in B will increase the density (by f(alpha) > 0). I have switched to the original statement from the website. Will that be more appropriate?

YaelDillies: Yes, certainly :)

FormalConjectures/ErdosProblems/38.lean
```diff
@@ -0,0 +1,46 @@
...
+
+/-!
+# Erdős Problem 38
+
+*Reference:* [erdosproblems.com/38](https://www.erdosproblems.com/38)
+-/
+
+open Classical Set Pointwise
+
+section Erdos38
```
YaelDillies:
```suggestion
namespace Erdos38
```

PR #1405, opened 2025-12-22
feat: add Lebesgue-Nagell equation conjecture (#431)

### LINKED ISSUE #431: Lebesgue-Nagell equation
### What is the conjecture

For odd prime $p$, the only integer solutions for the equation $x^2 - 2 = y^p$ are the trivial ones $(\pm 1, -1)$.

I found this from [this paper](https://arxiv.org/pdf/2507.12397) uploaded yesterday, which gives partial progress and also related references. It can be thought as a variation of Catalan equation, and probably putting under [Catalan.lean](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/Wikipedia/Catalan.lean) also makes sense. Actually, it may also possible to define all generalized Catalan equations and gather up these equations in a single file (I was not aware of these variants when I wrote Catalan.lean).

### Prerequisites needed

None

* ams-11

### Choose either option

ruskaruma:
## Summary

Adds the Lebesgue-Nagell equation conjecture to the Wikipedia section, extending `Catalan.lean` with related Diophantine equation content.

## The Conjecture

For any odd prime $p$, the only integer solutions $(x, y)$ to the equation:

$$x^2 - 2 = y^p$$

are $(x, y) = (\pm 1, -1)$.

## Reference

- Ethan Katz and Kyle Pratt, *"On the Lebesgue-Nagell equation $x^2 - 2 = y^p$"*, [arXiv:2507.12397](https://arxiv.org/abs/2507.12397)

## What's Added

- `lebesgue_nagell`: Main conjecture statement
- `lebesgue_nagell_solution_pos_one`: Verification that $(1, -1)$ is a solution ✓
- `lebesgue_nagell_solution_neg_one`: Verification that $(-1, -1)$ is a solution ✓

Closes #431


ruskaruma:
**Quick style question:**
I used a biconditional (`↔`) for `lebesgue_nagell`, which isn't consistent with `catalans_conjecture` on the same page that uses an implication form.

My reasoning for using `↔`: The conjecture states these are the *only* solutions, so a biconditional naturally captures "solutions exist if and only if they have this form." The reverse direction (`←`) is trivially true, as verified by the two test theorems, so we get a strictly stronger statement at no extra cost. Additionally, the symmetric solution pair `(±1, -1)` felt like a natural fit for a characterization-style biconditional.

That said, happy to refactor if implication form is preferred here. Just let me know.

ruskaruma:
Thanks for the review.

PR #1408, opened 2025-12-23
feat: state results and conjectures on VCₘ dim of convex sets in ℝⁿ

YaelDillies:
VC dimension of a set family is an old notion in learning theory. VC dimension of a *set* in a group (defined as the VC dimension of its family of translates) is a more recent notion motivated by additive combinatorics. VCₙ dimension of a set family is a very recent notion that appeared in the context of model theorym to extend the VC dimension of a set of translates.

This PR offers some conjectures in this direction. The conjectures are all mine and do not appear in the literature. They are very easily stated due to the elementary nature of VCₙ dimension.

### DIFF
--- a/FormalConjectures/Other/VCDimConvex.lean
+++ b/FormalConjectures/Other/VCDimConvex.lean
@@ -31,37 +31,38 @@ open scoped EuclideanGeometry Pointwise
 
 /-! ### What's known in the literature -/
 
-/-- Every convex set in ℝ² has VC dimension at most 3. -/
+/-- Every convex set in $\mathbb R^2$ has VC dimension at most 3. -/
 @[category research solved, AMS 5 52]
 lemma hasAddVCDimAtMost_three_of_convex_r2 {C : Set ℝ²} (hC : Convex ℝ C) : HasAddVCDimAtMost C 3 :=
   sorry
 
-/-- There exists a set in ℝ³ shattering an infinite set. -/
+/-- There exists a set in $\mathbb R^3$ shattering an infinite set. -/
 @[category research solved, AMS 5 52]
 lemma exists_infinite_convex_r3_shatters :
     ∃ A C : Set ℝ³, A.Infinite ∧ Convex ℝ C ∧ Shatters {t +ᵥ C | t : ℝ³} A := sorry
 
 /-! ### What's not in the literature -/
 
-/-- There exists a set of infinite VCₙ dimension in ℝⁿ⁺². -/
+/-- There exists a set of infinite $\mathrm{VC}_n$ dimension in $\mathbb R^{n + 2}$. -/
 @[category research solved, AMS 5 52]
 lemma exists_convex_rn_add_two_vc_n_forall_not_hasAddVCNDimAtMost (n : ℕ) :
     ∃ C : Set (Fin (n + 2) → ℝ), Convex ℝ C ∧ ∀ d, ¬ HasAddVCNDimAtMost C n d := sorry
 
 /-! ### Conjectures -/
 
-/-- Every convex set in ℝ³ has VC₂ dimension at most 1. -/
+/-- Every convex set in $\mathbb R^3$ has $\mathrm{VC}_2$ dimension at most 1. -/
 @[category research open, AMS 5 52]
 lemma hasAddVCNDimAtMost_two_one_of_convex_r3 {C : Set ℝ³} (hC : Convex ℝ C) :
     HasAddVCNDimAtMost C 2 1 := sorry
 
-/-- For every `n` there exists some `d` such that every convex set in ℝⁿ⁺¹ has VCₙ dimension at
-most `d`. -/
+/-- For every $n$ there exists some $d$ such that every convex set in $\mathbb R^{n + 1}$ has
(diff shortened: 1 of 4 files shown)


FormalConjecturesForMathlib/Combinatorics/Additive/VCDim.lean
```diff
@@ -0,0 +1,106 @@
...
+-/
+import FormalConjectures.ForMathlib.Combinatorics.SetFamily.VCDim
+import Mathlib
+
+open scoped Pointwise
+
+variable {G : Type*} [CommGroup G] {A B : Set G} {n d d' : ℕ}
+
+variable (A d n) in
+/-- A set `A` in an abelian group has VC dimension at most `d` iff one cannot find two sequences
```
mo271, line 33:
```suggestion
/-- A set `A` in an abelian group has Vapnik–Chervonenkis (VC) dimension
at most `d` iff one cannot find two sequences
```

or potentially add a module docstring?

FormalConjectures/ForMathlib/Combinatorics/Additive/VCDim.lean
```diff
@@ -0,0 +1,106 @@
...
+variable {G : Type*} [CommGroup G] {A B : Set G} {n d d' : ℕ}
+
+variable (A d n) in
+/-- A set `A` in an abelian group has VC dimension at most `d` iff one cannot find two sequences
+`x` and `y` of elements indexed by `[d + 1]` and `2 ^ [d + 1]ⁿ` respectively such that
+`y s * x i ∈ A ↔ i ∈ s` for all `i ∈ [d + 1]ⁿ`, `s ⊆ [d + 1]ⁿ`. -/
+@[to_additive
+/-- A set `A` in an abelian group has VCₙ dimension at most `d` iff one cannot find two sequences
+`x` and `y` of elements indexed by `[n] × [d + 1]` and `2 ^ [d + 1]ⁿ` respectively such that
+`y s + ∑ k, x (k, i k) ∈ A ↔ i ∈ s` for all `i ∈ [d + 1]ⁿ`, `s ⊆ [d + 1]ⁿ`. -/]
```
mo271:
```suggestion
`x` and `y` of elements indexed by `[d + 1]` and `2 ^ [d + 1]` respectively such that
`y s * x i ∈ A ↔ i ∈ s` for all `i ∈ [d + 1]`, `s ⊆ [d + 1]`. -/
@[to_additive
/-- A set `A` in an abelian group has VCₙ dimension at most `d` iff one cannot find two sequences
`x` and `y` of elements indexed by `[d + 1]` and `2 ^ [d + 1]` respectively such that
`y s + ∑ k, x i ∈ A ↔ i ∈ s` for all `i ∈ [d + 1]`, `s ⊆ [d + 1]`. -/]
```
copy paste from below? this is the $n=1$ case, right?

FormalConjectures/ForMathlib/Combinatorics/Additive/VCDim.lean
```diff
@@ -0,0 +1,106 @@
...
+limitations under the License.
+-/
+import FormalConjectures.ForMathlib.Combinatorics.SetFamily.VCDim
+import Mathlib
+
+open scoped Pointwise
+
+variable {G : Type*} [CommGroup G] {A B : Set G} {n d d' : ℕ}
+
+variable (A d n) in
```
mo271: $n$ could also only be declared only after the next definition, so it becomes clear that is it not used.
Also there is already another `n` in line 21, can that be removed?

FormalConjectures/ForMathlib/Combinatorics/SetFamily/VCDim.lean
```diff
@@ -0,0 +1,82 @@
...
+
+lemma univ_shatters : Shatters .univ A := .of_forall_subset <| by simp
+
+@[simp] lemma shatters_univ : Shatters 𝒜 .univ ↔ 𝒜 = .univ := by
+  simp [Shatters, Set.eq_univ_iff_forall]
+
+variable (𝒜 d) in
+/-- A set family `𝒜` has VC dimension at most `d` if there are no families `x` of elements indexed
+by `[d + 1]` and `A` of sets of `𝒜` indexed by `2^[d + 1]` such that `x i ∈ A s ↔ i ∈ s` for all
+`i ∈ [d], s ⊆ [d]`. -/
```
mo271:
```suggestion
`i ∈ [d + 1], s ⊆ [d + 1]`. -/
```
should it be like this or is the index `d` excluded for some reason? (doesn't look like it is excluded in the formal definition.

FormalConjectures/Other/VCDimConvex.lean
mo271, line 1: probably nicer to use latex markdown for the docstrings, perhaps look at the docgen output before merging?

mo271 (approved):
nice conjectures!
I'm just surprised that they aren't mentioned in the literature anywhere: they do seem natural.

Just a few docstring nits

FormalConjectures/ForMathlib/Combinatorics/Additive/VCDim.lean
YaelDillies: Good catch

PR #1415, opened 2025-12-26
feat(ErdosProblems): 517

### LINKED ISSUE #767: Erdős Problem 517
### What is the conjecture
    
https://www.erdosproblems.com/517

  Let $f(z)=\sum_{k=1}^\infty a_kz^{n_k}$ be an entire function. Is it true that if $n_k/k\to \infty$ then $f(z)$ assumes every value infinitely often?
  

Status: open

### Choose either option

CoolRmal:
This PR adds Erdos Problem 517.

Closes #767


FormalConjectures/ErdosProblems/517.lean
```diff
@@ -0,0 +1,78 @@
...
+ - [Bi28] Biernacki, Miécislas, Sur les équations algébriques contenant des paramétres arbitraires.
+    (1928), 145.
+ - [Wa01] Wang, Yuefei. "On the Fatou set of an entire function with gaps." Tohoku Mathematical
+    Journal, Second Series 53.1 (2001): 163-170.
+-/
+
+open Set Filter Topology
+
+/-- This is the terminology adopted in [Wa01] and some other sources. -/
+def hasFabryGaps (n : ℕ → ℕ) : Prop := StrictMono n ∧ Tendsto (fun k => n k / (k : ℝ)) atTop atTop
```
YaelDillies:
```suggestion
def HasFabryGaps (n : ℕ → ℕ) : Prop := StrictMono n ∧ Tendsto (fun k => n k / (k : ℝ)) atTop atTop
```
since it is a predicate. Same below

FormalConjectures/ErdosProblems/517.lean
```diff
@@ -0,0 +1,78 @@
...
+  _ ≤ 2 * 1 / (2 * b) := by grind
+  _ = 1 / b := by grind
+
+namespace Erdos517
+
+/-- If `f(z) = ∑ aₖzⁿₖ` is an entire function such that `nₖ / k → ∞`, is it true that `f` assumes
+every value infinitely often? -/
+@[category research open, AMS 30]
+theorem erdos_517.fabry {f : ℂ → ℂ} {n : ℕ → ℕ} (hn : hasFabryGaps n) {a : ℕ → ℂ}
+    (hf : ∀ z, HasSum (fun k => (a k) * z ^ (n k)) (f z)) (z : ℂ) :
```
YaelDillies:
```suggestion
    (hf : ∀ z, HasSum (fun k => a k * z ^ n k) (f z)) (z : ℂ) :
```
Same below

FormalConjectures/ErdosProblems/517.lean
```diff
@@ -0,0 +1,78 @@
...
+    {x : ℂ | f x = z}.Infinite := by
```
YaelDillies, line 37: The statement is a question. Could you therefore turn the statement into 
```suggestion
/-- If `f(z) = ∑ aₖzⁿₖ` is an entire function such that `nₖ / k → ∞`, is it true that `f` assumes
every value infinitely often? -/
@[category research open, AMS 30]
theorem erdos_517.fabry : answer(sorry) \iff
    \forall {f : ℂ → ℂ} {n : ℕ → ℕ} (hn : hasFabryGaps n) {a : ℕ → ℂ}
      (hf : ∀ z, HasSum (fun k => (a k) * z ^ (n k)) (f z)) (z : ℂ),
        {x : ℂ | f x = z}.Infinite := by
```

YaelDillies:
Apparently, #767 classifies this under AMS-32. Could you clarify?

CoolRmal:
AMS-32 is Several Complex Variables and Analytic Spaces, but here I think we only care about one-variable functions, and this is why I used AMS-30 instead in the PR.

FormalConjectures/ErdosProblems/517.lean
```diff
@@ -0,0 +1,78 @@
...
+  /- use the Cauchy criterion of series. -/
+  have : ∃ k > 0, ∀ m ≥ k, ∑ j : Icc ⌊m / 2⌋₊ m , 1 / (n j : ℝ)
+    ≤ 1 / (2 * b) := by
+    have : Icc (-1 / (2 * b)) (1 / (2 * b)) ∈ (𝓝 0) := by sorry
+    obtain ⟨k, hk⟩ := hn.2.nat_tsum_vanishing this
+    refine ⟨2 * k + 1, by linarith, fun m hm => ?_⟩
+    have : Icc ⌊m / 2⌋₊ m ⊆ {n | k ≤ n} := by sorry
+    have := (hk (Icc ⌊m / 2⌋₊ m) this).2
+    simpa [tsum_fintype] using this
+  obtain ⟨k, hk⟩ := this
```
YaelDillies:
```suggestion
   obtain ⟨k, hk⟩ : ∃ k > 0, ∀ m ≥ k, ∑ j : Icc ⌊m / 2⌋₊ m , 1 / (n j : ℝ)
    ≤ 1 / (2 * b) := by
    have : Icc (-1 / (2 * b)) (1 / (2 * b)) ∈ (𝓝 0) := by sorry
    obtain ⟨k, hk⟩ := hn.2.nat_tsum_vanishing this
    refine ⟨2 * k + 1, by linarith, fun m hm => ?_⟩
    have : Icc ⌊m / 2⌋₊ m ⊆ {n | k ≤ n} := by sorry
    have := (hk (Icc ⌊m / 2⌋₊ m) this).2
    simpa [tsum_fintype] using this
```
but also could you use `filter_upwards` to avoid undoing the `Filter` primitives? You shouldn't have to use `eventually_atTop`

FormalConjectures/ErdosProblems/517.lean
```diff
@@ -0,0 +1,114 @@
...
+ - [erdosproblems.com/517](https://www.erdosproblems.com/517)
+ - [Bi28] Biernacki, Miécislas, Sur les équations algébriques contenant des paramétres arbitraires.
+    (1928), 145.
+ - [Wa01] Wang, Yuefei. "On the Fatou set of an entire function with gaps." Tohoku Mathematical
+    Journal, Second Series 53.1 (2001): 163-170.
+-/
+
+open Set Filter Topology
+
+/-- This is the terminology adopted in [Wa01] and some other sources. -/
```
YaelDillies: Could you make this docstring descriptive? Right now I can't tell what it means without reading the body of the definition

FormalConjectures/ErdosProblems/517.lean
```diff
@@ -0,0 +1,114 @@
...
+def HasFabryGaps (n : ℕ → ℕ) : Prop := StrictMono n ∧ Tendsto (fun k => n k / (k : ℝ)) atTop atTop
+
+def HasFejerGaps (n : ℕ → ℕ) : Prop := StrictMono n ∧ Summable (fun k => 1 / (n k : ℝ))
```
YaelDillies: Could you add a docstring here please?

CoolRmal:
-awaiting-author

The proof of `HasFejerGaps.HasFabryGaps` contains a lot of calculations and is thus a bit messy. The essential idea is that if a sequence `nₖ` satisfies `Summable (fun k => 1 / (n k : ℝ))`, then for large `k`, `∑ j : Icc (k / 2) k , 1 / (n j : ℝ)` is small and this term is larger than `k /(2 * n k)`.

FormalConjectures/ErdosProblems/517.lean
```diff
@@ -0,0 +1,114 @@
...
+
+@[category API, AMS 40]
+theorem hasFejerGaps.hasFabryGaps {n : ℕ → ℕ} (hn : HasFejerGaps n) : HasFabryGaps n := by
+  refine ⟨hn.1, ?_⟩
+  simp only [tendsto_atTop_atTop]
```
YaelDillies: This still looks wrong. You shouldn't have to use this lemma. Do you how to use `filter_upwards`?

CoolRmal: Yes I do. But I think somehow `filter_upwards` is not very convenient to use because I need to consider the cases `b>0` and `b ≤ 0` separately. I have already simplified my proof a lot and the version right now is the best I can come up with. May you take a look again?

FormalConjectures/ErdosProblems/517.lean
```diff
@@ -0,0 +1,99 @@
...
+
+open Set Filter Topology
+
+/-- A sequence of natural numbers `n₀ < n₁ < ...` is said to have Fabry gaps if `nₖ / k → ∞`.
+This is the terminology adopted in [Wa01] and some other sources. -/
+def HasFabryGaps (n : ℕ → ℕ) : Prop := StrictMono n ∧ Tendsto (fun k => n k / (k : ℝ)) atTop atTop
+
+/-- A sequence of natural numbers `n₀ < n₁ < ...` is said to have Fejér gaps if `∑' 1 / nₖ < ∞`.
+This is the terminology adopted in [Wa01] and some other sources. -/
+def HasFejerGaps (n : ℕ → ℕ) : Prop := StrictMono n ∧ Summable (fun k => 1 / (n k : ℝ))
```
CoolRmal: I just realized that these two definitions are also needed in Erdös Problem 516. Should we move them into another file?

YaelDillies: Sure, that is a good idea! Please put it in a new file in ForMathlib and import it in Util.ForMathlib

FormalConjectures/ForMathlib/Analysis/HasGaps.lean
```diff
@@ -0,0 +1,79 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import Mathlib.Algebra.Order.Ring.Star
+import Mathlib.Analysis.RCLike.Basic
+import Mathlib.Data.Int.Star
+import Mathlib.Data.Nat.Cast.Order.Field
```
YaelDillies: Can you reduce the imports further?

FormalConjectures/ForMathlib/Analysis/HasGaps.lean
```diff
@@ -0,0 +1,79 @@
...
+
+/-! #
+
+*References:*
```
YaelDillies:
```suggestion
/-!
# Insert title here

## References:
```

FormalConjectures/ForMathlib/Analysis/HasGaps.lean
CoolRmal: This is generated by #min_imports

YaelDillies: There's no guarantee that `#min_imports` yields a minimal set of imports. It is easily fooled by typeclass search.

FormalConjectures/ErdosProblems/517.lean
YaelDillies: You shouldn't have to deal with the `b ≤ 0` case because eventually `b > 0` anyway. I have pushed a golf

PR #1421, opened 2025-12-27
fix(ErdosProblems/728): make interesting

YaelDillies:
Although the statement on the website wasn't updated, everyone agrees a and b should be less than n, and Tao has further suggested that a + b - n should be on the order of log n.

### DIFF
--- a/FormalConjectures/ErdosProblems/728.lean
+++ b/FormalConjectures/ErdosProblems/728.lean
@@ -35,15 +35,15 @@ and
 $$C \log n < a + b - n < C' \log n ?$$
 -/
 @[category research open, AMS 11]
-theorem erdos_728 : answer(sorry) ↔
-    ∀ C > (0 : ℝ), ∀ C' > C, ∀ᶠ ε : ℝ in 𝓝[>] 0,
+theorem erdos_728 :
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/728.lean
```diff
@@ -22,22 +22,30 @@ import FormalConjectures.Util.ProblemImports
...
       ε * n < a ∧
       ε * n < b ∧
-      Nat.factorial a * Nat.factorial b ∣ Nat.factorial n * Nat.factorial (a + b - n) ∧
-      a + b > n + C * Real.log n) ↔ answer(sorry) := by
+      a < (1 - ε) * n ∧
+      b < (1 - ε) * n ∧
+      a ! * b ! ∣ n ! * (a + b - n)! ∧
+      a + b > n + C * log n ∧
+      a + b < n + C' * log n := by
   sorry
```
mo271, line 49: When putting `answer(False)` above, then AlphaProof finds
```
  use default, (by match· _ _ _ one_pos one_pos one_lt_two with | ⟨e,y,k,s, a, right, M, _⟩ =>linarith only[a.trans M])
```
I think this corresponds to the counter example with `ε = C = 1` and `C' = 2`.

YaelDillies: Whoops, good point. Should be fixed now

mo271 (approved):

PR #1426, opened 2025-12-29
chore(ErdosProblems/1077): tag as `solved`

pitmonticone:
A reference was recently found for this problem and it is now marked as solved on the [site](https://www.erdosproblems.com/forum/thread/1077).

### DIFF
--- a/FormalConjectures/ErdosProblems/1077.lean
+++ b/FormalConjectures/ErdosProblems/1077.lean
@@ -43,7 +43,7 @@ theorem erdos_1077 :
           IsBalanced H.coe D ∧
           m > (n : ℝ) ^ (1 - α) ∧
           H.edgeSet.ncard > ε * m ^ (1 + α))
-      ↔ answer(sorry) := by
+      ↔ answer(False) := by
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/1077.lean
```diff
@@ -34,7 +34,7 @@ Let $ε, α > 0$ and $D$ and $n$ be sufficiently large. If $G$ is a graph on $n$
 least $n^{1+α}$ edges, then must $G$ contain a $D$-balanced subgraph on $m > n^{1-α}$ vertices with
 at least $εm^{1+α}$ edges?
 -/
-@[category research open, AMS 5]
+@[category research solved, AMS 5]
```
Paul-Lez, line 37: It seems that this was disproved - could you also replace the `answer(sorry)` by `answer(False)`?

pitmonticone: Oh sure, sorry.

YaelDillies (approved):
Thanks!

PR #1432, opened 2025-12-30
feat(ErdosProblems): 513

### LINKED ISSUE #763: Erdős Problem 513
### What is the conjecture
    
https://www.erdosproblems.com/513

  Let $f=\sum_{n=0}^\infty a_nz^n$ be a transcendental entire function. What is the greatest possible value of
$$\liminf_{r\to \infty} \frac{\max_n\lvert a_nr^n\rvert}{\max_{\lvert z\rvert=r}\lvert f(z)\rvert}?$$
  

Status: open

### Choose either option

CoolRmal:
Closes #763

### DIFF
--- a/FormalConjectures/ForMathlib/Algebra/Polynomial/Algebra.lean
+++ b/FormalConjectures/ForMathlib/Algebra/Polynomial/Algebra.lean
@@ -24,12 +24,6 @@ import Mathlib.RingTheory.Algebraic.Pi
 
 variable {R S : Type*} [CommSemiring R] [CommSemiring S] [Algebra R S]
 
-namespace Algebra
-
-@[simp] lemma toRingHom_ofId : ofId R S = algebraMap R S := rfl
-
-end Algebra
-
 namespace Polynomial
 
 @[simp] theorem eval₂_id {p : Polynomial R} {x : R} : eval₂ (RingHom.id _) x p = p.eval x := rfl


FormalConjectures/ErdosProblems/513.lean
```diff
@@ -0,0 +1,56 @@
...
+  (⨆ n, ‖iteratedDeriv n f 0 * ((n)! : ℝ)⁻¹ * r ^ n‖) / (⨆ z : {z : ℂ // ‖z‖ = r}, ‖f z‖)
+
+/-- Let `f` be a transcendental entire function. What is the greatest possible value of
+`liminf (fun r : ℝ => ratio r f) atTop`? -/
+@[category research open, AMS 30]
+theorem erdos_513.sup : answer(sorry) = ⨆ f : {f : ℂ → ℂ // IsTranscendental f},
+    (liminf (fun r : ℝ => ratio r f) atTop) := by
+  sorry
+
+/-- For all transcendetal entire function `f`, `liminf (fun r : ℝ => ratio r f) atTop ≤ 2 / π - c`
```
YaelDillies:
```suggestion
/-- For all transcendental entire function `f`, `liminf (fun r : ℝ => ratio r f) atTop ≤ 2 / π - c`
```
same below

FormalConjectures/ErdosProblems/513.lean
```diff
@@ -0,0 +1,56 @@
...
+  (1964), 143-186.
+-/
+
+open scoped Nat Real
+open Filter
+
+namespace Erdos513
+
+noncomputable def ratio (r : ℝ) (f : ℂ → ℂ) : ℝ :=
+  (⨆ n, ‖iteratedDeriv n f 0 * ((n)! : ℝ)⁻¹ * r ^ n‖) / (⨆ z : {z : ℂ // ‖z‖ = r}, ‖f z‖)
```
YaelDillies: Alternatively,
```suggestion
  (⨆ n, ‖iteratedDeriv n f 0 * (n ! : ℝ)⁻¹ * r ^ n‖) / (⨆ z : {z : ℂ // ‖z‖ = r}, ‖f z‖)
```

FormalConjectures/ForMathlib/Analysis/Transcendental.lean
```diff
@@ -0,0 +1,28 @@
...
+import Mathlib.Analysis.Calculus.FDeriv.Defs
+
+/-!
+# Transcendental Entire Functions
+
+-/
+
+open Complex
+
+def IsTranscendental (f : ℂ → ℂ) : Prop := Differentiable ℂ f ∧ ∀ p : Polynomial ℂ, p.eval ≠ f
```
YaelDillies: This definition is only accurate if the function is already entire. Maybe you could formalise the more general definition? https://en.wikipedia.org/wiki/Transcendental_function

I suspect https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/Algebraic/Defs.html#Transcendental is already the correct definition, in fact

CoolRmal: I see, so maybe sth like`Transcendental (Polynomial ℂ) f`?

This definition requires the instance `Algebra (Polynomial ℂ) (ℂ → ℂ)` though, which seems to be missing in Mathlib.

YaelDillies: No, not quite. Instead it should be
```
import Mathlib

example (f : ℂ → ℂ) : IsAlgebraic ℂ f := by
  simp [IsAlgebraic, funext_iff]
  -- ∃ p, ¬p = 0 ∧ ∀ (x : ℂ), Polynomial.eval (f x) p = 0
```

CoolRmal: But here `p` is a one variable polynomial. I think we need a two variable polynomial. The wikipedia page says that $f(x)=\frac{a x+b}{c x+d}$ is not transcendental because it is the root of the polynomial $(ax+ b)-(cx+d)y=0\in ℂ[x][y]$. 

Here's another source which supports my argument: https://math.stackexchange.com/questions/2961583/transcendental-entire-functions

YaelDillies: Indeed, you are right, here is the correct statement:
```
import Mathlib

namespace Polynomial

variable {R S : Type*} [CommSemiring R] [CommSemiring S] [Algebra R S]

instance instAlgebraPi : Algebra R[X] (S → S) :=
  (Pi.ringHom fun x ↦ (Polynomial.aeval x).toRingHom).toAlgebra

@[simp] lemma aeval_polynomial_pi (p : R[X][X]) (f : S → S) (x : S) :
    p.aeval f x = aevalAeval x (f x) p := by simp [instAlgebraPi, aeval, eval₂, sum]

example (f : ℂ → ℂ) : IsAlgebraic ℂ[X] f := by
  simp [IsAlgebraic, funext_iff]
  -- ∃ p ≠ 0, ∀ x : ℂ, eval x (aeval (C (f x)) p) = 0

end Polynomial
```

CoolRmal: I created a new file in the folder about polynomials and moved the instance `instAlgebraPi` into it. There's a warning when I move the lemma `aeval_polynomial_pi` into this new file because `aevalAeval` is not defined for `CommSemiring` in this repo.

YaelDillies: Ah indeed, it was generalised from `CommRing` in https://github.com/leanprover-community/mathlib4/pull/30867. Maybe just state everything for `CommRing` for now and leave a TODO?

CoolRmal: I can still use `CommSemiring` for the instance `instAlgebraPi`. I just can't prove the lemma `aeval_polynomial_pi`, which is also not needed in the problem statement of 513, so I think it is fine to not have it right now.

YaelDillies: But can you add the ungeneralised lemma in the same file? I am pretty sure that whatever AI will try to solve it is will thank you for having this lemma available

CoolRmal: I add the lemma.

PR #1436, opened 2025-12-30
feat: Ferreira's result for the Andrica, Brocard, Legendre, Oppermann conjectures

seewoo5:
The main result of Ferreira [Real exponential sums over primes and prime gaps](https://arxiv.org/abs/2307.08725) implies Andrica, Brocard, Legendre, and Oppermann's conjecture for sufficiently large $n$.

### DIFF
--- a/FormalConjectures/Wikipedia/BrocardConjecture.lean
+++ b/FormalConjectures/Wikipedia/BrocardConjecture.lean
@@ -16,8 +16,6 @@ limitations under the License.
 
 import FormalConjectures.Util.ProblemImports
 
-open Filter
-
 /-!
 # Brocard's Conjecture
 
@@ -26,7 +24,7 @@ open Filter
 - [Luan Alberto Ferreira, *Real exponential sums over primes and prime gaps*](https://arxiv.org/abs/2307.08725)
 -/
 
-open Finset
+open Finset Filter
 
 namespace Brocard
 


FormalConjectures/Wikipedia/Andrica.lean
```diff
@@ -16,10 +16,14 @@ limitations under the License.
...
 
+open Filter
+
 /-!
 # Andrica's conjecture
 
-*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Andrica%27s_conjecture)
+*References:*
+- [Wikipedia](https://en.wikipedia.org/wiki/Andrica%27s_conjecture)
+- [Fe23] Luan Alberto Ferreira, Real exponential sums over primes and prime gaps, arXiv:2307.08725
```
YaelDillies: This reference is broken: The syntax is `[text]{link)` or `[text][name of ref]` where `name of ref` is in the references file. Here I would suggest
```suggestion
- [Luan Alberto Ferreira, *Real exponential sums over primes and prime gaps*](https://arxiv.org/abs/2307.08725)
```

seewoo5:
> Thanks! Ferreira's result is really powerful, I remember reading about all those conjectures years ago and thinking they were out of reach.

Indeed! All the previous works I saw before this were existence of prime in short intervals $[x, x + x^\lambda]$, where many of the results assume RH and still didn't make $\lambda$ smaller than $1/2$. This is really amazing result.

PR #1442, opened 2026-01-01
fix(ErdosProblems/1038): exclude constant polynomials

CoolRmal:
The infimum should range over nonconstant monic polynomials. Otherwise the infimum will be trivially equal to zero.


YaelDillies (commented):
Can you also update the docstrings?

PR #1457, opened 2026-01-03
feat(Navier-Stokes): formalization of Navier–Stokes existence and smoothness

### LINKED ISSUE #2293: Navier Stokes Equations
### What is the conjecture
One of the 7 Millennium Problems set by the Clay Mathematics Institute.

Detailed information can be found here:
https://www.claymath.org/wp-content/uploads/2022/06/navierstokes.pdf

lecopivo:
fixes #2293

Formalization of Navier-Stokes existence and smoothness problem as described in https://www.claymath.org/wp-content/uploads/2022/06/navierstokes.pdf

Potentially issues:
 - not immediately obvious that `fderiv ℝ (v · t) x (v x t)` is the convective term and that `div` is divergence
 - the growth conditions uses `‖iteratedFDeriv ℝ n v₀ x‖` instead of `|∂^α u(x,t)|` but it should be equivalent
 - at various places I'm not completely sure if there should be `t ≥ 0` or `t > 0` as `fderiv` might not give what we think at `t=0`. Mainly in `NavierStokesExistenceAndSmoothness.navierStokes`

Note:
 - I didn't use the unicode character nu for viscosity as it looks too similar to the letter `v` which is used for velocity.
 - bikeshead: should it use `v(x,t)`(aligned with the official statement) or `v(t,x)` that is more mathlib aligned?
 - should it be formulated in an arbitrary dimension and then instantiated for `n:=3`?


FormalConjectures/Wikipedia/NavierStokes.lean
YaelDillies, line 1: Maybe create a new folder `FormalConjectures.Millenium`? This surely deserves not to be drowned with all the conjectures from Wikipedia!

FormalConjectures/Wikipedia/NavierStokes.lean
```diff
@@ -0,0 +1,130 @@
...
+# Existence And Smoothness Of The Navier–Stokes Equation
+
+*Reference:*
+  [Wikipedia](https://en.wikipedia.org/wiki/Navier%E2%80%93Stokes_existence_and_smoothness)
+  [Clay Mathematics Institute](https://www.claymath.org/wp-content/uploads/2022/06/navierstokes.pdf)
+-/
+
+open ContDiff Set InnerProductSpace MeasureTheory
+
+local macro "ℝ³"  : term => `(EuclideanSpace ℝ (Fin 3))
```
YaelDillies: There is both an `ℝ²` and an `ℝ^ d` notation in the `EuclideanGeometry` scope that are defined in this repo. Maybe you could add your notation next to them?

FormalConjectures/Wikipedia/NavierStokes.lean
```diff
@@ -0,0 +1,130 @@
...
+
+noncomputable
+def div (v : ℝ³ → ℝ³) (x : ℝ³) : ℝ := (fderiv ℝ v x).trace ℝ ℝ³
```
YaelDillies: I know this is standard terminology, but the confusion with `/` could maybe be avoided by calling this `divergence` and giving it the usual notation instead?

YaelDillies (commented):
Could you add *way* more documentation? I would make sure we're at least at mathlib standard (every def has a docstring), and hopefully go even further by writing down explicitly all the design decisions you've taken.

> I didn't use the unicode character nu for viscosity as it looks too similar to the letter v which is used for velocity.

This seems fine.

> bikeshead: should it use v(x,t) (aligned with the official statement) or v(t,x) that is more mathlib aligned?

I don't mind, so long as the discrepancy is clearly documented.

> should it be formulated in an arbitrary dimension and then instantiated for n:=3?

Sure! That seems very reasonable so long as it isn't too difficult to do so. I assume it's already solved for `n = 2`? If so, it seems reasonable to have one statement for `n = 2`, one for `n = 3` and one for all `n`, the first one being solved and the other two open.

lecopivo:
@YaelDillies Hopefully, I have addressed all the issues. 

The docs strings are mostly AI generated but they sound good to me and I don't really know what else I would write there.

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+
+  /-- The pressure is periodic in space with period 1 for all times t ≥ 0 (condition 10). -/
+  pressure_periodic : ∀ t ≥ 0, IsPeriodic (p · t)
+
+
+/-- (A) Existence and smoothness of Navier–Stokes solutions on ℝ³. -/
+@[category research open, AMS 35]
+theorem navier_stokes_existence_and_smoothness_R3 (nu : ℝ) (hnu : nu > 0)
+  (v₀ : ℝ³ → ℝ³) (hv₀ : InitialVelocityConditionRn v₀) :
+  ∃ v p, NavierStokesExistenceAndSmoothnessRn (n:=3) nu v₀ (f:=0) v p := sorry
```
YaelDillies:
```suggestion
    (v₀ : ℝ³ → ℝ³) (hv₀ : InitialVelocityConditionRn v₀) :
    ∃ v p, NavierStokesExistenceAndSmoothnessRn (n:=3) nu v₀ (f:=0) v p := sorry
```
Same below

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+  extends NavierStokesExistenceAndSmoothness nu v₀ f v p where
+
+  /-- The velocity is square-integrable at each time t ≥ 0. -/
+  integrable : ∀ t ≥ 0, Integrable (‖v · t‖^2)
+
+  /-- The kinetic energy ∫|v(x,t)|²dx remains uniformly bounded for all time (condition 7). -/
+  globally_bounded_energy : ∃ E, ∀ t ≥ 0, (∫ x : ℝ^n, ‖v x t‖^2) < E
+
+
+/--
```
YaelDillies:
```suggestion
/--
```

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+Initial velocity conditions for the Navier-Stokes problem on all of ℝⁿ.
+
+In addition to being smooth and divergence-free, the velocity must decay
+faster than any polynomial at spatial infinity (condition 4 in Fefferman's paper).
+
+This condition ensures the velocity field has finite energy and reasonable
+behavior as |x| → ∞.
+-/
+structure InitialVelocityConditionRn {n : ℕ} (v₀ : ℝ^n → ℝ^n) : Prop
```
YaelDillies: What about
```suggestion
structure InitialVelocityConditionDecay {n : ℕ} (v₀ : ℝ^n → ℝ^n) : Prop
```
instead? In both cases, the function has domain `R^n`, so `Rn` is a bit confusing as a way to distinguish

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+  velocity_periodic : ∀ t ≥ 0, IsPeriodic (v · t)
+
+  /-- The pressure is periodic in space with period 1 for all times t ≥ 0 (condition 10). -/
+  pressure_periodic : ∀ t ≥ 0, IsPeriodic (p · t)
+
+
+/-- (A) Existence and smoothness of Navier–Stokes solutions on ℝ³. -/
+@[category research open, AMS 35]
+theorem navier_stokes_existence_and_smoothness_R3 (nu : ℝ) (hnu : nu > 0)
+  (v₀ : ℝ³ → ℝ³) (hv₀ : InitialVelocityConditionRn v₀) :
```
YaelDillies: The source uses
```suggestion
  (u₀ : ℝ³ → ℝ³) (hv₀ : InitialVelocityConditionRn u₀) :
```
instead. Maybe you could do that so that `nu` can become `\nu`?

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+structure NavierStokesExistenceAndSmoothnessPeriodic {n : ℕ}
+    (nu : ℝ) (v₀ : ℝ^n → ℝ^n) (f : ℝ^n → ℝ → ℝ^n)
+    (v :  ℝ^n → ℝ → ℝ^n) (p : ℝ^n → ℝ → ℝ) : Prop
+  extends NavierStokesExistenceAndSmoothness nu v₀ f v p where
+
+  /-- The velocity is periodic in space with period 1 for all times t ≥ 0 (condition 10). -/
+  velocity_periodic : ∀ t ≥ 0, IsPeriodic (v · t)
+
+  /-- The pressure is periodic in space with period 1 for all times t ≥ 0 (condition 10). -/
+  pressure_periodic : ∀ t ≥ 0, IsPeriodic (p · t)
```
YaelDillies: Technically, the periodicity of pressure isn't stated anywhere. Could you leave a comment to that effect?

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+4. The solution is smooth (C^∞) for all time t ≥ 0 (equations 6, 11)
+-/
+structure NavierStokesExistenceAndSmoothness {n : ℕ}
+    (nu : ℝ) (v₀ : ℝ^n → ℝ^n) (f : ℝ^n → ℝ → ℝ^n)
+    (v :  ℝ^n → ℝ → ℝ^n) (p : ℝ^n → ℝ → ℝ) : Prop where
+
+  /-- The Navier-Stokes equation (equation 1):
+      ∂v/∂t + (v·∇)v = ν∆v - ∇p + f -/
+  navier_stokes : ∀ x, ∀ t > 0,
+    deriv (v x ·) t + fderiv ℝ (v · t) x (v x t) = nu • Δ (v · t) x - gradient (p · t) x + f x t
```
YaelDillies: Condition 1 is also stated for t = 0. Could you explain the discrepancy in a comment?

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+In addition to the basic solution properties, we require:
+- The velocity is square-integrable at each time (finite kinetic energy)
+- The total energy remains bounded for all time (condition 7)
+-/
+structure NavierStokesExistenceAndSmoothnessRn {n : ℕ}
+    (nu : ℝ) (v₀ : ℝ^n → ℝ^n) (f : ℝ^n → ℝ → ℝ^n)
+    (v :  ℝ^n → ℝ → ℝ^n) (p : ℝ^n → ℝ → ℝ) : Prop
+  extends NavierStokesExistenceAndSmoothness nu v₀ f v p where
+
+  /-- The velocity is square-integrable at each time t ≥ 0. -/
```
YaelDillies:
```suggestion
  /-- The velocity is square-integrable at each time t ≥ 0 (condition 7). -/
```

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+theorem navier_stokes_existence_and_smoothness_periodic (nu : ℝ) (hnu : nu > 0)
+  (v₀ : ℝ³ → ℝ³) (hv₀ : InitialVelocityConditionPeriodic v₀) :
+  ∃ v p, NavierStokesExistenceAndSmoothnessPeriodic (n:=3) nu v₀ (f:=0) v p := sorry
+
+/-- (C) Breakdown of Navier–Stokes solutions on ℝ³. -/
+@[category research open, AMS 35]
+theorem navier_stokes_breakdown_R3 (nu : ℝ) (hnu : nu > 0) :
+  ∃ (v₀ : ℝ³ → ℝ³) (f : ℝ³ → ℝ → ℝ³),
+    InitialVelocityConditionRn v₀ ∧ ForceConditionRn f ∧
+    ¬(∃ v p, NavierStokesExistenceAndSmoothnessRn (n:=3) nu v₀ f v p) := sorry
```
YaelDillies:
```suggestion
    ∃ (v₀ : ℝ³ → ℝ³) (f : ℝ³ → ℝ → ℝ³),
      InitialVelocityConditionRn v₀ ∧ ForceConditionRn f ∧
      \forall v p, ¬ NavierStokesExistenceAndSmoothnessRn (n:=3) nu v₀ f v p := sorry
```
maybe?

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+The force must be smooth, periodic in space, and decay in time
+(conditions 8-9 in Fefferman's paper).
+-/
+structure ForceConditionPeriodic {n : ℕ} (f : ℝ^n → ℝ → ℝ^n) : Prop
+  extends ForceCondition f where
+  /-- The force is periodic in space with period 1 for all times (condition 8) -/
+  periodic : ∀ t ≥ 0, IsPeriodic (f · t)
+  /-- All derivatives of f decay faster than any polynomial in time (condition 9). -/
+  decay : ∀ m : ℕ, ∀ K : ℝ, ∃ C : ℝ, ∀ x, ∀ t > 0,
+    ‖iteratedFDeriv ℝ m (↿f) (x,t)‖ ≤ C / (1 + t)^K
```
YaelDillies: Can you make it clear that this is a time decay?

FormalConjectures/Millenium/NavierStokes.lean
mirefek: Errata on page 6 says also $p$ should be periodic.

mirefek:
We looked at it with Tomáš Bárta (teacher of ODE at Charles University), and it looks good to us.

mirefek:
More from Tomáš Bárta:

the formulation of NS seems correct to me but I have some remarks:

1. Should there be 'for all `nu>0`' in all theorems?

It seems strange that there is 'for all `nu`' in parts A, B (existence of global smooth solutions) and also 'for all `nu`' in parts C, D (non-existence of globally defined smooth solutions). In fact, if non-existence is proved for one nu, by a rescaling procedure it is proved for all positive `nu` (with another `v_0` and `f`). So, it does not matter whether there is 'for all `nu`' or 'there exists `nu`'. But may be it would be better to write 'there exists `nu>0`' in parts C, D? I suggest to ask an NS expert concerning this.

2. Problems with $t>0$ and $t\ge 0$.

The present lean formulation writes $t>0$ in some conditions instead of $t\ge 0$ and the reason seems to be the language. Why don't you use `iteratedFDerivWithin` instead of `iteratedFDeriv`. Wouldn't it allow to work on $t\ge 0$? Maybe this would be a more elegant solution. (`iteratedFDerivWithin` seems to be used in definition of `ContDiffOn`, so we need it anyway)

FormalConjectures/Millenium/NavierStokes.lean
YaelDillies: Sorry, I realise I can't read

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+- `navier_stokes_breakdown_periodic`: (D) Existence of breakdown scenario on ℝ³/ℤ³
+-/
+
+open ContDiff Set InnerProductSpace MeasureTheory EuclideanGeometry
+
+/--
+The divergence of a vector field `v : ℝⁿ → ℝⁿ` at point `x`,
+computed as the trace of the Jacobian matrix.
+
+In coordinates: div v = Σᵢ ∂vᵢ/∂xᵢ
```
grunweg: This formula should be written as a LaTeX formula; likewise for the formula three lines above and in the other doc-strings below.

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+-/
+noncomputable
+def divergence {n : ℕ} (v : ℝ^n → ℝ^n) (x : ℝ^n) : ℝ := (fderiv ℝ v x).trace ℝ (ℝ^n)
```
grunweg: This definition has junk value zero when v is not differentiable at x (because then the `fderiv` is defined to be zero).

grunweg: Maybe also add that as a simp lemma. And some basic API lemmas, such as linearity?

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+
+notation "∇⬝" => divergence
```
grunweg, line 76:
```suggestion
@[inherit_doc]
notation "∇⬝" => divergence
```
Can you also mention the notation in the doc-string, please?

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+
+/--
+A function `f : ℝⁿ → α` is periodic if it is periodic in each coordinate
+with period 1, i.e., `f(x + eᵢ) = f(x)` for each unit vector `eᵢ`.
+This captures functions on the n-torus ℝⁿ/ℤⁿ.
+-/
+def IsPeriodic {α : Sort*} {n : ℕ} (f : ℝ^n → α) : Prop :=
```
grunweg: Perhaps rename this to `IsOnePeriodic`?

FormalConjectures/Millenium/NavierStokes.lean
grunweg: You're only using this notation once. Can you also use it on line 75 --- or remove it?

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+
+This condition ensures the velocity field has finite energy and reasonable
+behavior as |x| → ∞.
+-/
+structure InitialVelocityConditionRn {n : ℕ} (v₀ : ℝ^n → ℝ^n) : Prop
+  extends InitialVelocityCondition v₀ where
+  /-- All derivatives of v₀ decay faster than any polynomial (condition 4).
+
+      For any derivative order m and any decay rate K, there exists a constant C
+      such that |∂ᵐv₀(x)| ≤ C/(1+|x|)^K. -/
```
grunweg: Personal preference, to some extent:
```suggestion
  /-- All derivatives of v₀ decay faster than any polynomial (condition 4).
  For any derivative order m and any decay rate K, there exists a constant C
  such that |∂ᵐv₀(x)| ≤ C/(1+|x|)^K. -/
```
(and please apply the LaTeX formatting also)

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+  decay : ∀ m : ℕ, ∀ K : ℝ, ∃ C : ℝ, ∀ x, ‖iteratedFDeriv ℝ m v₀ x‖ ≤ C / (1 + ‖x‖)^K
```
grunweg:
```suggestion
  decay : ∀ m : ℕ, ∀ K : ℝ, ∃ C : ℝ, ∀ x, ‖iteratedFDeriv ℝ m v₀ x‖ ≤ C / (1 + ‖x‖) ^ K
```

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+
+/--
+Initial velocity conditions for the periodic Navier-Stokes problem on ℝⁿ/ℤⁿ.
+
+The velocity must be smooth, divergence-free, and periodic with period 1
```
grunweg: I also like the phrasing "1-periodic"; this may read more nicely. Optional/subjective.

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+in each coordinate (condition 8 in Fefferman's paper).
+-/
+structure InitialVelocityConditionPeriodic {n : ℕ} (v₀ : ℝ^n → ℝ^n) : Prop
+  extends InitialVelocityCondition v₀ where
+  /-- The initial velocity is periodic with period 1 in each direction (condition 8) -/
```
grunweg: This is only part of condition 8 --- perhaps rephrase as "condition 8, part 1"? Same above.

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+  periodic : IsPeriodic v₀
+
+/--
+Basic smoothness condition on external forcing term.
```
grunweg: Please add a definite article here. Thanks!

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+  /-- The force is smooth on ℝⁿ × [0,∞) -/
+  smooth : ContDiffOn ℝ ∞ (↿f) (Set.univ ×ˢ Set.Ici 0)
+
+/--
+Force conditions for the Navier-Stokes problem on all of ℝⁿ.
+
+The force must be smooth and decay faster than any polynomial
+in both space and time (condition 5 in Fefferman's paper).
+-/
+structure ForceConditionRn {n : ℕ} (f : ℝ^n → ℝ → ℝ^n) : Prop
```
grunweg: style: I'd put the `extends` clause on the same line. (Also, please apply the style comments about doc-string formatting in the same way as above.)

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+  extends ForceCondition f where
+  /-- All derivatives of f decay faster than any polynomial in space and time (condition 5).
+
+      For any derivative orders m and decay rate K, there exists C such that
+      |∂ᵐ_{x,t} f(x,t)| ≤ C/(1+|x|+t)^K for t > 0.
+  -/
+  decay : ∀ m : ℕ, ∀ K : ℝ, ∃ C : ℝ, ∀ x, ∀ t > 0,
```
grunweg: I have the same comment as was made in the thread above: please use t >= 0 (and iteratedFDerivWithin if needed) --- or say clearly why you think this is a bad idea. I don't see why. Also below (and in the doc-string).

grunweg (commented):
Thanks for your PR. With Moritz Firsching sitting next to me (to answer any formal conjectures questions), I just took a first look, until line 130. Most of my comments are minor or stylistic --- but let me echo the comment about t > 0 vs t >= 0.

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+The Clay Millennium Problem asks for a proof of one of the following four statements:
+
+- `navier_stokes_existence_and_smoothness_R3`: (A) Global existence on ℝ³
+- `navier_stokes_existence_and_smoothness_periodic`: (B) Global existence on ℝ³/ℤ³
+- `navier_stokes_breakdown_R3`: (C) Existence of breakdown scenario on ℝ³
+- `navier_stokes_breakdown_periodic`: (D) Existence of breakdown scenario on ℝ³/ℤ³
+-/
+
+open ContDiff Set InnerProductSpace MeasureTheory EuclideanGeometry
+
```
grunweg, line 62: Should `n` be a global implicit `variable`?

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+/--
+Force conditions for the periodic Navier-Stokes problem on ℝⁿ/ℤⁿ.
+
+The force must be smooth, periodic in space, and decay in time
+(conditions 8-9 in Fefferman's paper).
+-/
+structure ForceConditionPeriodic {n : ℕ} (f : ℝ^n → ℝ → ℝ^n) : Prop
+  extends ForceCondition f where
+  /-- The force is periodic in space with period 1 for all times (condition 8) -/
+  periodic : ∀ t ≥ 0, IsPeriodic (f · t)
```
grunweg: I'd name this field `isPeriodic` (or `isOnePeriodic` if you rename `IsPeriodic` to `IsOnePeriodic`)

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+This structure captures the core requirements for a solution:
+1. The velocity and pressure satisfy the Navier-Stokes PDE (equation 1)
+2. The velocity remains divergence-free for all time (equation 2)
+3. The initial condition is satisfied (equation 3)
+4. The solution is smooth (C^∞) for all time t ≥ 0 (equations 6, 11)
+-/
+structure NavierStokesExistenceAndSmoothness {n : ℕ}
+    (nu : ℝ) (v₀ : ℝ^n → ℝ^n) (f : ℝ^n → ℝ → ℝ^n)
+    (v :  ℝ^n → ℝ → ℝ^n) (p : ℝ^n → ℝ → ℝ) : Prop where
+
```
grunweg:
```suggestion
```

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+In addition to the basic solution properties, we require:
+- The velocity is square-integrable at each time (finite kinetic energy)
+- The total energy remains bounded for all time (condition 7)
+-/
+structure NavierStokesExistenceAndSmoothnessRn {n : ℕ}
+    (nu : ℝ) (v₀ : ℝ^n → ℝ^n) (f : ℝ^n → ℝ → ℝ^n)
+    (v :  ℝ^n → ℝ → ℝ^n) (p : ℝ^n → ℝ → ℝ) : Prop
+  extends NavierStokesExistenceAndSmoothness nu v₀ f v p where
+
```
grunweg:
```suggestion
```

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+  /-- The velocity is square-integrable at each time t ≥ 0. -/
+  integrable : ∀ t ≥ 0, Integrable (‖v · t‖^2)
```
grunweg: Should this use `MemLp` instead, i.e.
```suggestion
  integrable : ∀ t ≥ 0, MemLp (‖v · t‖) 2
```

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+The velocity and pressure must be periodic with period 1 in each
+spatial direction for all times (condition 10).
+-/
+structure NavierStokesExistenceAndSmoothnessPeriodic {n : ℕ}
+    (nu : ℝ) (v₀ : ℝ^n → ℝ^n) (f : ℝ^n → ℝ → ℝ^n)
+    (v :  ℝ^n → ℝ → ℝ^n) (p : ℝ^n → ℝ → ℝ) : Prop
+  extends NavierStokesExistenceAndSmoothness nu v₀ f v p where
+
+  /-- The velocity is periodic in space with period 1 for all times t ≥ 0 (condition 10). -/
+  velocity_periodic : ∀ t ≥ 0, IsPeriodic (v · t)
```
grunweg: I'd name this as `isPeriodic_velocity` (following mathlib conventions). Similarly below.

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+structure NavierStokesExistenceAndSmoothnessRn {n : ℕ}
+    (nu : ℝ) (v₀ : ℝ^n → ℝ^n) (f : ℝ^n → ℝ → ℝ^n)
+    (v :  ℝ^n → ℝ → ℝ^n) (p : ℝ^n → ℝ → ℝ) : Prop
+  extends NavierStokesExistenceAndSmoothness nu v₀ f v p where
+
+  /-- The velocity is square-integrable at each time t ≥ 0. -/
+  integrable : ∀ t ≥ 0, Integrable (‖v · t‖^2)
+
+  /-- The kinetic energy ∫|v(x,t)|²dx remains uniformly bounded for all time (condition 7). -/
+  globally_bounded_energy : ∃ E, ∀ t ≥ 0, (∫ x : ℝ^n, ‖v x t‖^2) < E
```
grunweg: This is the Lebesgue integral, right? Then everything is fine. (For the Bochner integral, you'd have to worry about junk values.)

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+  ∃ (v₀ : ℝ³ → ℝ³) (f : ℝ³ → ℝ → ℝ³),
+    InitialVelocityConditionRn v₀ ∧ ForceConditionRn f ∧
+    ¬(∃ v p, NavierStokesExistenceAndSmoothnessRn (n:=3) nu v₀ f v p) := sorry
+
+/-- (D) Breakdown of Navier–Stokes Solutions on ℝ³/ℤ³. -/
+@[category research open, AMS 35]
+theorem navier_stokes_breakdown_periodic (nu : ℝ) (hnu : nu > 0) :
+  ∃ (v₀ : ℝ³ → ℝ³) (f : ℝ³ → ℝ → ℝ³),
+    InitialVelocityConditionPeriodic v₀ ∧ ForceConditionPeriodic f ∧
+    ¬(∃ v p, NavierStokesExistenceAndSmoothnessPeriodic (n:=3) nu v₀ f v p) := sorry
```
grunweg: Spaces around `:=`, please --- also above.

grunweg (commented):
I finished my review: another collection of minor items; no further big questions.

mo271:
@lecopivo it would be great to make some progress here: it essentially seems that this is in a very good state, just seems some polishing

Paul-Lez:
@mo271 It turns out that @lecopivo, @grunweg and I will all be at the same workshop this week so I'd been thinking of trying to have a quick meetup with all three of us to get this past the finish line!

PaulCharlton:
Overall:

Aggregator missing. Fefferman p.2 l.10–12 frames the prize as "we ask for a proof of one of the following four statements." That disjunction is the prize statement; File 1 only states the four routes as parallel theorems and never asserts A ∨ B ∨ C ∨ D.

Could use a lot of documentation for someone not coming from a formal PDE background -- explicit variable names and types, separate from parameters.

FormalConjectures/Millenium/NavierStokes.lean
```diff
@@ -0,0 +1,239 @@
...
+
+The force must be smooth and decay faster than any polynomial
+in both space and time (condition 5 in Fefferman's paper).
+-/
+structure ForceConditionRn {n : ℕ} (f : ℝ^n → ℝ → ℝ^n) : Prop
+  extends ForceCondition f where
+  /-- All derivatives of f decay faster than any polynomial in space and time (condition 5).
+
+      For any derivative orders m and decay rate K, there exists C such that
+      |∂ᵐ_{x,t} f(x,t)| ≤ C/(1+|x|+t)^K for t > 0.
```
PaulCharlton: should be  `t>=0` ?

FormalConjectures/Millenium/NavierStokes.lean
PaulCharlton: you already got this -- my comment can go away

PaulCharlton:
Errata not referenced. Two-part erratum: (a) p(x+e_j,t)=p(x,t) added to eq. (8), (b) sign correction to eq. (12). On (a), the condition is captured by pressure_periodic (line 211) but the docstring attributes it to eq. (10), which is velocity periodicity — wrong source, erratum uncited. On (b), eq. (12) is not in the file at all, so the sign correction has nothing to land on; whether eq. (12) belongs in a prize-statement file is debatable, but the silence is currently undocumented.

Paul-Lez (approved):
LGTM, let's merge this!

PR #1466, opened 2026-01-04
chore(Millenium): move P vs NP and GRH

YaelDillies:
I am moving them out of Wikipedia rather than duplicating them because Wikipedia isn't really a list of problems.


mo271:
Should we also 
FormalConjectures/Wikipedia/GeneralizedRiemannHypothesis.lean?

YaelDillies:
Good idea! Done

Issue #1104, opened 2025-10-14
Erdős Problem 1062


mo271:
### What is the conjecture

https://www.erdosproblems.com/1062

  Let $f(n)$ be the size of the largest subset $A\subseteq \\{1,\ldots,n\\}$ such that there are no three distinct elements $a,b,c\in A$ such that $a\mid b$ and $a\mid c$. How large can $f(n)$ be? Is $\lim f(n)/n$ irrational?

Status:  open

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

PR #1476, opened 2026-01-04
feat(ErdosProblems): 1062

conglu1997:
Closes https://github.com/google-deepmind/formal-conjectures/issues/1104 with the help of codex! :)

### DIFF
--- a/FormalConjectures/ErdosProblems/1062.lean
+++ b/FormalConjectures/ErdosProblems/1062.lean
@@ -36,10 +36,7 @@ def ForkFree (A : Set ℕ) : Prop :=
 /-- The extremal function from Erdős problem 1062: the largest size of a fork-free subset of
 `{1,...,n}`. -/
 noncomputable def f (n : ℕ) : ℕ :=
-  Nat.findGreatest
-    (fun k =>
-      ∃ A ⊆ Set.Icc 1 n, ForkFree A ∧ A.ncard = k)
-    n
+  Nat.findGreatest (fun k => ∃ A ⊆ Set.Icc 1 n, ForkFree A ∧ A.ncard = k) n
 
 /-- The interval `[m + 1, 3m + 2]` gives a construction showing that `f n` is asymptotically
 at least `⌊2n / 3⌋`. -/


FormalConjectures/ErdosProblems/1062.lean
```diff
@@ -0,0 +1,64 @@
...
+-/
+
+open Classical Filter
+
+namespace Erdos1062
+
+/-- A finite set `A` of positive integers is fork-free if no element divides
+two distinct other elements of `A`. -/
+def ForkFree (A : Finset ℕ) : Prop :=
+  ∀ a ∈ A, ((A.erase a).filter fun b => a ∣ b).card ≤ 1
```
YaelDillies:
```suggestion
  ∀ a ∈ A, {b \in A.erase a | a ∣ b}.Subsingleton
```
is more idiomatic

FormalConjectures/ErdosProblems/1062.lean
```diff
@@ -0,0 +1,64 @@
...
+
+/-- `Admissible A n` means that `A` is contained in `{1,...,n}` and is fork-free. -/
+def Admissible (A : Finset ℕ) (n : ℕ) : Prop :=
+  A ⊆ Finset.Icc 1 n ∧ ForkFree A
```
YaelDillies: This definition doesn't seem very useful: inline it?

FormalConjectures/ErdosProblems/1062.lean
```diff
@@ -0,0 +1,64 @@
...
+@[category research open, AMS 11]
+theorem erdos_1062.lower_bound (n : ℕ) : (2 * n) / 3 ≤ f n := by
+  sorry
+
+/-- Lebensold proved that for large `n`, the function `f n` lies between `0.6725 n` and
+`0.6736 n`. -/
+@[category research solved, AMS 11]
+theorem erdos_1062.lebensold_bounds :
+    ∃ N, ∀ ⦃n⦄, N ≤ n →
+      ((6725 : ℝ) / 10000) * n ≤ (f n : ℝ) ∧ (f n : ℝ) ≤ ((6736 : ℝ) / 10000) * n := by
```
YaelDillies: Unless I'm wholly mistaken, this works
```suggestion
      (0.6725 : ℝ) * n ≤ f n ∧ f n ≤ (0.6736 : ℝ) * n := by
```

FormalConjectures/ErdosProblems/1062.lean
```diff
@@ -31,14 +31,10 @@ two distinct other elements of `A`. -/
 def ForkFree (A : Finset ℕ) : Prop :=
   ∀ a ∈ A, {b \in A.erase a | a ∣ b}.Subsingleton
```
YaelDillies:
```suggestion
  ∀ a ∈ A, {b ∈ A.erase a | a ∣ b}.Subsingleton
```

FormalConjectures/ErdosProblems/1062.lean
```diff
@@ -31,14 +31,10 @@ two distinct other elements of `A`. -/
...
 
-/-- `Admissible A n` means that `A` is contained in `{1,...,n}` and is fork-free. -/
-def Admissible (A : Finset ℕ) (n : ℕ) : Prop :=
-  A ⊆ Finset.Icc 1 n ∧ ForkFree A
-
 /-- The extremal function from Erdős problem 1062: the largest size of a fork-free subset of
 `{1,...,n}`. -/
 def f (n : ℕ) : ℕ :=
-  Nat.findGreatest (fun k => ∃ A, Admissible A n ∧ A.card = k) n
+  Nat.findGreatest (fun k => ∃ A, A ⊆ Finset.Icc 1 n ∧ ForkFree A ∧ A.card = k) n
```
YaelDillies:
```suggestion
  Nat.findGreatest (fun k => ∃ A ⊆ Finset.Icc 1 n, ForkFree A ∧ A.card = k) n
```

FormalConjectures/ErdosProblems/1062.lean
CoolRmal: This will create a warning because we don't have`Finset.Subsingleton`...well why don't we just assume that `A` is a Set instead of a Finset in this definition?

YaelDillies: Sure, that's reasonable to me. Then you'll also need
```suggestion
  ∀ a ∈ A, {b ∈ A \ {a} | a ∣ b}.Subsingleton
```

FormalConjectures/ErdosProblems/1062.lean
```diff
@@ -0,0 +1,60 @@
...
+/-- The extremal function from Erdős problem 1062: the largest size of a fork-free subset of
+`{1,...,n}`. -/
+def f (n : ℕ) : ℕ :=
+  Nat.findGreatest (fun k => ∃ A, A ⊆ Finset.Icc 1 n ∧ ForkFree A ∧ A.card = k) n
+
+/-- The interval `[m + 1, 3m + 2]` gives a construction showing that `f n` is asymptotically
+at least `2n / 3`. -/
+@[category research open, AMS 11]
+theorem erdos_1062.lower_bound (n : ℕ) : (2 * n) / 3 ≤ f n := by
+  sorry
```
CoolRmal: This shouldn't be too hard to formalize. Also I believe a slightly better version should hold: 

```suggestion
theorem erdos_1062.lower_bound (n : ℕ) : ⌈(2 * n / 3 : ℝ)⌉₊ ≤ f n := by
  sorry
```

CoolRmal:
An important part of this conjecture is how large `f n` can be, which is not really formalized in this PR. Perhaps we can have sth like this?

```
theorem erdos_1062.asymptotics : (fun n => (f n : ℝ)) =Θ[atTop] (answer(sorry) : ℕ → ℝ) := by
  sorry
```
or 

```
theorem erdos_1062.asymptotics : (fun n => (f n : ℝ)) ~[atTop] answer(sorry) := by
  sorry
```

However, when an AI encounter this kind of problem, they will definitely cheat by picking `f` to be the answer, which is not really what we want (usually we expect sth `log n`, `log log n`, `n^2` to be our answer). I guess my real question is, how we should deal with this kind of growth problem in the future (because estimation of the growth rate appears frequently in other Erdos problems)?

YaelDillies:
Our approach so far has been to stick to the concrete conjectures that were formulated.

However, my personal opinion is that we can do better: As you say, we expect the growth rate to be a combination of elementary functions: +, -, *, /, constants, possibly combined with some function going to 0 or infinity. Such functions can be defined inductively on their depth, and my vision is that we could state those "Estimate the growth of blah" questions as `\exists f : ElementaryFun atTop, blah ~[atTop] elementary`, `\exists elementary : ElementaryFun atTop, blah =o[atTop] elementary`, etc...

To avoid vacuous statements of the form `blah ~[atTop] blahAsElementaryFun` (which would be possible if `blah` goes to infinity, for example), one could even require `elementary` to have depth at least 1 and to not be "simplifiable" (to avoid `blah ~[atTop] (fun n => blahAsElementaryFun n + n - n)`.

FormalConjectures/ErdosProblems/1062.lean
conglu1997: Adopted thank you!

conglu1997:
Thanks so much for all the feedback, resolved! :)

FormalConjectures/ErdosProblems/1062.lean
```diff
@@ -0,0 +1,60 @@
...
+def ForkFree (A : Finset ℕ) : Prop :=
+  ∀ a ∈ A, {b ∈ A \ {a} | a ∣ b}.Subsingleton
+
+/-- The extremal function from Erdős problem 1062: the largest size of a fork-free subset of
+`{1,...,n}`. -/
+def f (n : ℕ) : ℕ :=
+  Nat.findGreatest (fun k => ∃ A ⊆ Finset.Icc 1 n, ForkFree A ∧ A.card = k) n
+
+/-- The interval `[m + 1, 3m + 2]` gives a construction showing that `f n` is asymptotically
+at least `2n / 3`. -/
```
YaelDillies: The website says floor, not ceil
```suggestion
at least `⌊2n / 3⌋`. -/
```

FormalConjectures/ErdosProblems/1062.lean
```diff
@@ -0,0 +1,60 @@
...
+@[category research open, AMS 11]
```
YaelDillies:
```suggestion
@[category research solved, AMS 11]
```

FormalConjectures/ErdosProblems/1062.lean
```diff
@@ -0,0 +1,60 @@
...
+theorem erdos_1062.lower_bound (n : ℕ) : ⌈(2 * n / 3 : ℝ)⌉₊ ≤ f n := by
```
YaelDillies:
```suggestion
theorem erdos_1062.lower_bound (n : ℕ) : 2 * n / 3 ≤ f n := by
```
Natural number division floors

FormalConjectures/ErdosProblems/1062.lean
```diff
@@ -0,0 +1,60 @@
...
+  sorry
+
+/-- Lebensold proved that for large `n`, the function `f n` lies between `0.6725 n` and
+`0.6736 n`. -/
+@[category research solved, AMS 11]
+theorem erdos_1062.lebensold_bounds :
+    ∃ N, ∀ ⦃n⦄, N ≤ n →
+      (0.6725 : ℝ) * n ≤ f n ∧ f n ≤ (0.6736 : ℝ) * n := by
```
YaelDillies: We have notation for this:
```suggestion
    ∀ᶠ n in atTop, (0.6725 : ℝ) * n ≤ f n ∧ f n ≤ (0.6736 : ℝ) * n := by
```

FormalConjectures/ErdosProblems/1062.lean
YaelDillies: It certainly holds for large n because of the Lebensold bound, but the website really does state this using floor, not ceil, right?

conglu1997:
Fixing build errors, switched A to Set instead of finite set in ForkFree

FormalConjectures/ErdosProblems/1062.lean
```diff
@@ -23,18 +24,22 @@ import FormalConjectures.Util.ProblemImports
...
+  ∀ a ∈ A, ({b | b ∈ A \ {a} ∧ a ∣ b} : Set ℕ).Subsingleton
 
 /-- The extremal function from Erdős problem 1062: the largest size of a fork-free subset of
 `{1,...,n}`. -/
-def f (n : ℕ) : ℕ :=
-  Nat.findGreatest (fun k => ∃ A ⊆ Finset.Icc 1 n, ForkFree A ∧ A.card = k) n
+noncomputable def f (n : ℕ) : ℕ :=
+  Nat.findGreatest
+    (fun k =>
+      ∃ A : Set ℕ, A ⊆ (Finset.Icc 1 n : Set ℕ) ∧ A.Finite ∧ ForkFree A ∧ Nat.card A = k)
```
YaelDillies:
```suggestion
      ∃ A ⊆ Set.Icc 1 n, ForkFree A ∧ A.ncard = k)
```
`A` is necessarily finite if it is a subset of a finite set

conglu1997:
Fixed!

FormalConjectures/ErdosProblems/1062.lean
CoolRmal: The ceiling function is used in Guy's Unsolved Problems in Number Theory. Also it is not hard to see why the ceiling function is true. Consider the set

$$
\{\lfloor n / 3\rfloor+1,\lfloor n / 3\rfloor+2, \ldots, n\} .
$$

Its size is

$$
n-\lfloor n / 3\rfloor=\left\lceil2n/3\right\rceil .
$$

YaelDillies: Oh I see. Feel free to open a PR to change to the stronger bound!

In the future, if you claim something that goes against the website, please make it clear that you know you are doing so. My prior is that the website is correct

conglu1997: Thank you for the very kind advice, I have opened a PR now for this :) 

https://github.com/google-deepmind/formal-conjectures/pull/1495

CoolRmal: @YaelDillies Ok I'll make it more clear next time.

PR #1481, opened 2026-01-05
feat: functions of elementary growth

CoolRmal:
This PR intends to define the type of common functions used to estimate growth rate so that for example we can formalize a better statement in #1476 and potentially other Erdos problems.
I only include constants, identity function, log, exp, gamma (factorial), addition, multiplication, and composition because

1. `a^x = exp x * (log a)`
2. `x^a = exp a * (log x)`
3. `x^x = exp x * (log x)`
4. We certainly don't want other kinds of elementary functions (e.g. trigonometric functions) because they oscillate.
5. Subtraction is multiplying by -1.
6. `f/g = exp (log f - log g)`

This should include most of functions we are interested in.

I also only defined evaluation map for `ℝ → ℝ` and `ℕ → ℝ` instead of functions with more general domain/codomain because I think these are enough for Erdos problems.

### DIFF
--- a/FormalConjectures/ForMathlib/Analysis/Asymptotics/Basic.lean
+++ b/FormalConjectures/ForMathlib/Analysis/Asymptotics/Basic.lean
@@ -18,7 +18,6 @@ import Mathlib.Analysis.Asymptotics.Defs
 import Mathlib.Order.Filter.AtTopBot.Defs
 import Mathlib.Analysis.SpecialFunctions.Log.Basic
 
-open scoped Nat
 open Filter
 
 notation f " ≫ " g => Asymptotics.IsBigO Filter.atTop g f
--- a/FormalConjectures/ForMathlib/Analysis/Asymptotics/Basic.lean
+++ b/FormalConjectures/ForMathlib/Analysis/Asymptotics/Basic.lean
@@ -17,9 +17,9 @@ limitations under the License.
 import Mathlib.Analysis.Asymptotics.Defs
 import Mathlib.Order.Filter.AtTopBot.Defs
 import Mathlib.Analysis.SpecialFunctions.Log.Basic
-import Mathlib.Analysis.SpecialFunctions.Gamma.Basic
 
 open scoped Nat
+open Filter
 
 notation f " ≫ " g => Asymptotics.IsBigO Filter.atTop g f
 notation g " ≪ " f => Asymptotics.IsBigO Filter.atTop g f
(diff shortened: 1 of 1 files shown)


YaelDillies:
> gamma (factorial)

By Stirling's formula, this is unnecessary

FormalConjectures/ForMathlib/Analysis/Asymptotics/Basic.lean
```diff
@@ -16,6 +16,47 @@ limitations under the License.
...
+  | id       => fun x => x
+  | log      => fun x => x.log
+  | exp      => fun x => x.exp
+  | Gamma    => fun x => x.Gamma
+  | add f g  => fun x => eval f x + eval g x
+  | mul f g  => fun x => eval f x * eval g x
+  | comp f g => eval f ∘ eval g
+
+/-- The evaluation map of the type of common growth functions. -/
+noncomputable def Nat.eval : CommonFn → ℕ → ℝ
```
YaelDillies: Is this any different to `Real.eval n`? If not, I don't think we need it

FormalConjectures/ForMathlib/Analysis/Asymptotics/Basic.lean
```diff
@@ -16,6 +16,47 @@ limitations under the License.
...
+/-- The type of common functions used to estimate growth rate. -/
+inductive CommonFn : Type
+  | const : ℝ → CommonFn
+  | id    : CommonFn
+  | log   : CommonFn
+  | exp   : CommonFn
+  | Gamma : CommonFn
+  | add   : CommonFn → CommonFn → CommonFn
+  | mul   : CommonFn → CommonFn → CommonFn
+  | comp  : CommonFn → CommonFn → CommonFn
```
YaelDillies: I am rather sure you will need to allow functions tending to zero or infinity, so that we can state things like `~ x ^(1 + o(1))` or `>> x log x beta(x)` where `beta -> infty` (one Erdos problem has partial progress of this form where `beta` is reverse Ackermann)

YaelDillies: Could you define `sub`, `neg`, `div`, `inv`, `pow`, `zpow`, etc... from those primitives? Also as I said `Gamma` shouldn't be a primitive

FormalConjectures/ForMathlib/Analysis/Asymptotics/Basic.lean
```diff
@@ -16,6 +16,47 @@ limitations under the License.
...
+import Mathlib.Analysis.SpecialFunctions.Log.Basic
+import Mathlib.Analysis.SpecialFunctions.Gamma.Basic
+
+open scoped Nat
 
 notation f " ≫ " g => Asymptotics.IsBigO Filter.atTop g f
 notation g " ≪ " f => Asymptotics.IsBigO Filter.atTop g f
+
+/-- The type of common functions used to estimate growth rate. -/
+inductive CommonFn : Type
```
YaelDillies:
```suggestion
inductive ElementaryFun : Type
```
or
```suggestion
inductive ElementaryGrowth : Type
```
seems more appropriate

YaelDillies (commented):
Could you write an elaborator for such functions? Something that would eg make `elem(exp (x * x) * log x)` elaborate as `.mul (.comp .exp (.mul .id .id)) .log`.

CoolRmal:
@YaelDillies I think I get the idea but I am not sure how to realize this in Lean. What is the type of this elaborator? Do I need some knowledge in metaprogramming in order to define this?

YaelDillies:
Indeed, this would be metaprogramming. What you need to do is basically to follow [this section](https://leanprover-community.github.io/lean4-metaprogramming-book/main/05_syntax.html) of MetIL and possibly [this one](https://leanprover-community.github.io/lean4-metaprogramming-book/main/08_dsls.html) too.

BoltonBailey:
I guess by the same token that Gamma is definable in terms of the other growth rates via Stirling's formula, addition is definable in terms of exp, mul, and log. Perhaps that's a change worth making?

FormalConjectures/ForMathlib/Analysis/Asymptotics/Basic.lean
```diff
@@ -16,6 +16,46 @@ limitations under the License.
 
 import Mathlib.Analysis.Asymptotics.Defs
 import Mathlib.Order.Filter.AtTopBot.Defs
+import Mathlib.Analysis.SpecialFunctions.Log.Basic
+
+open Filter
 
 notation f " ≫ " g => Asymptotics.IsBigO Filter.atTop g f
 notation g " ≪ " f => Asymptotics.IsBigO Filter.atTop g f
+
```
Paul-Lez, line 25: I think it might be better to include this code in a file `../Asymptotics/ElementaryGrowth.lean` since I suspect more content will be added to it after! It would also be useful to add a module docstring to the file documenting what you've added, and the intended use (perhaps with some examples!).

Paul-Lez (commented):
Thanks for the PR!

mo271:
@CoolRmal do you plan to continue to work on this?

CoolRmal:
No not really. Please adopt this if someone is interested in working on it.

YaelDillies:
This is a cute idea but I am not yet convinced of the execution: asymptotic equivalence classes aren't preserved under eg exp, so the current constructors feel rough.

I will close for now, keeping in mind to come back to it in case it shows up in other formalisations again.

PR #1500, opened 2026-01-06
feat(ErdosProblems): 741

### LINKED ISSUE #924: Erdős Problem 741
### What is the conjecture

https://www.erdosproblems.com/741

  Let $A\subseteq \mathbb{N}$ be such that $A+A$ has positive density. Can one always decompose $A=A_1\sqcup A_2$ such that $A_1+A_1$ and $A_2+A_2$ both have positive density?

Is there a basis $A$ of order $2$ such that if $A=A_1\sqcup A_2$ then $A_1+A_1$ and $A_2+A_2$ cannot both have bounded gaps?

Status:  open

### Choose either option

CoolRmal:
Closes #924.

When Erdos wrote these two problems in his paper [Er94b], he assumed that `A` is a sequence of integers instead of a set, so I believe the website made a mistake. However, ultimately I still assume that `A` is a set in this PR because
1. We want `A + A` to be a set instead of a function.
2. `HasPosDensity` is defined a set.
3. It is obvious how one should define `OfBoundedGaps` for a sequence, but not very obvious for a set. I think the following definition gives the one that we want: https://en.wikipedia.org/wiki/Syndetic_set

I also think the definition of `IsMulBasisOfOrder` needs some modification. The definition in our repository is: `∀ a, a ∈ A ^ n`. The definition over here https://www.erdosproblems.com/definitions is : `A` is a basis of order `n` if any element can be expressed as sum of at most `n` integers. The sequence {1} ∪ {2, 4, 6, 8, 10, ...} is a basis or order 2 by the second definition but not a basis of order 2 by the first definition (because 1 is not sum two integers in this set), which is weird.

There's also an issue about whether 0 needs to representable as a sum. In the book Old and new problems and results in combinatorial number theory (https://mathweb.ucsd.edu/~ronspubs/80_11_number_theory.pdf), they only require positive integers to be representable as a sum.
Screenshot 2026-01-05 at 11 03 05 PM

### DIFF
--- a/FormalConjectures/ErdosProblems/741.lean
+++ b/FormalConjectures/ErdosProblems/741.lean
@@ -35,7 +35,7 @@ namespace Erdos741
 positive density? -/
 @[category research open, AMS 5]
 theorem erdos_741.density : answer(sorry) ↔ ∀ A : Set ℕ, HasPosDensity (A + A) → ∃ A₁ A₂,
-    A₁ ⊆ A ∧ A₂ ⊆ A ∧ A = A₁ ∪ A₂ ∧ Disjoint A₁ A₂ ∧ HasPosDensity (A₁ + A₁)
+    A = A₁ ∪ A₂ ∧ Disjoint A₁ A₂ ∧ HasPosDensity (A₁ + A₁)
     ∧ HasPosDensity (A₂ + A₂):= by
   sorry
 
@@ -43,7 +43,7 @@ theorem erdos_741.density : answer(sorry) ↔ ∀ A : Set ℕ, HasPosDensity (A 
 subsets `A₁` and `A₂` such that `A₁ + A₁` and `A₂ + A₂` cannot both have bounded gaps? -/
 @[category research open, AMS 5]
 theorem erdos_741.basis : answer(sorry) ↔ ∀ A : Set ℕ, IsAddBasisOfOrder (A ∪ {0}) 2 → ∃ A₁ A₂,
-    A₁ ⊆ A ∧ A₂ ⊆ A ∧ A = A₁ ∪ A₂ ∧ Disjoint A₁ A₂ ∧ ¬ (IsSyndetic A₁ ∧ IsSyndetic A₂):= by
+    A = A₁ ∪ A₂ ∧ Disjoint A₁ A₂ ∧ ¬ (IsSyndetic A₁ ∧ IsSyndetic A₂):= by
   sorry
 
 end Erdos741


YaelDillies:
> I also think the definition of `IsMulBasisOfOrder` needs some modification. The definition in our repository is: `∀ a, a ∈ A ^ n`. The definition over here https://www.erdosproblems.com/definitions is : `A` is a basis of order `n` if any element can be expressed as sum of at most `n` integers. The sequence {1} ∪ {2, 4, 6, 8, 10, ...} is a basis or order 2 by the second definition but not a basis of order 2 by the first definition (because 1 is not sum two integers in this set), which is weird.
> 
> There's also an issue about whether 0 needs to representable as a sum. In the book Old and new problems and results in combinatorial number theory (https://mathweb.ucsd.edu/~ronspubs/80_11_number_theory.pdf), they only require positive integers to be representable as a sum.

The current definition is correct for sets containing 0. The easy fix is therefore to artificially add 0 to your set

FormalConjectures/ForMathlib/Combinatorics/Basic.lean
```diff
@@ -29,6 +29,10 @@ if the equation $a + b = c$ has no solution with $a, b, c \in A$.
 -/
 def IsSumFree (A : Set α) : Prop := Disjoint (A + A)  A
 
+/-- A set $A$ of natural numbers is said to have bounded gaps if there exists an integer $p$ such
+that $A ∩ [n, n + 1, ..., n + p]$ is nonempty for all $n$. -/
+def OfBoundedGaps (A : Set ℕ) : Prop := ∃ p, ∀ n, A ∩ .Icc n (n + p) ≠ ∅
```
YaelDillies:
```suggestion
def IsSyndetic (A : Set ℕ) : Prop := ∃ p, ∀ n, (A ∩ .Icc n (n + p)).Nonempty
```

FormalConjectures/ErdosProblems/741.lean
```diff
@@ -0,0 +1,48 @@
...
+open Set
+
+namespace Erdos741
+
+/-- Let `A ⊆ ℕ` be a set such that `A + A` has positive density. Can one always decompose
+`A` as a disjoint union of two subsets `A₁` and `A₂` such that both `A₁ + A₁` and `A₂ + A₂` have
+positive density? -/
+@[category research open, AMS 5]
+theorem erdos_741.density : answer(sorry) ↔ ∀ A : Set ℕ, HasPosDensity (A + A) → ∃ B C,
+    B ⊆ A ∧ C ⊆ A ∧ A = B ∪ C ∧ Disjoint B C ∧ HasPosDensity (B + B) ∧ HasPosDensity (C + C):= by
```
YaelDillies: Can you use the same variable names?

FormalConjectures/ErdosProblems/741.lean
```diff
@@ -0,0 +1,48 @@
...
+  sorry
+
+/-- Let `A ⊆ ℕ` be a basis of order 2. Can one always decompose `A` as a disjoint union of two
+subsets `A₁` and `A₂` such that `A₁ + A₁` and `A₂ + A₂` cannot both have bounded gaps? -/
+@[category research open, AMS 5]
+theorem erdos_741.basis : answer(sorry) ↔ ∀ A : Set ℕ, IsAddBasisOfOrder A 2 → ∃ B C,
+    B ⊆ A ∧ C ⊆ A ∧ A = B ∪ C ∧ Disjoint B C ∧ (¬ (OfBoundedGaps B ∧ OfBoundedGaps C)):= by
```
YaelDillies:
```suggestion
    B ⊆ A ∧ C ⊆ A ∧ A = B ∪ C ∧ Disjoint B C ∧ ¬ (OfBoundedGaps B ∧ OfBoundedGaps C) := by
```

FormalConjectures/ErdosProblems/741.lean
```diff
@@ -0,0 +1,49 @@
...
+open Set
+
+namespace Erdos741
+
+/-- Let `A ⊆ ℕ` be a set such that `A + A` has positive density. Can one always decompose
+`A` as a disjoint union of two subsets `A₁` and `A₂` such that both `A₁ + A₁` and `A₂ + A₂` have
+positive density? -/
+@[category research open, AMS 5]
+theorem erdos_741.density : answer(sorry) ↔ ∀ A : Set ℕ, HasPosDensity (A + A) → ∃ A₁ A₂,
+    A₁ ⊆ A ∧ A₂ ⊆ A ∧ A = A₁ ∪ A₂ ∧ Disjoint A₁ A₂ ∧ HasPosDensity (A₁ + A₁)
```
YaelDillies: This is redundant
```suggestion
    A₁ ∪ A₂ ∧ Disjoint A₁ A₂ ∧ HasPosDensity (A₁ + A₁)
```

FormalConjectures/ErdosProblems/741.lean
```diff
@@ -0,0 +1,49 @@
...
+    ∧ HasPosDensity (A₂ + A₂):= by
+  sorry
+
+/-- Let `A ⊆ ℕ` be a basis of order 2. Can one always decompose `A` as a disjoint union of two
+subsets `A₁` and `A₂` such that `A₁ + A₁` and `A₂ + A₂` cannot both have bounded gaps? -/
+@[category research open, AMS 5]
+theorem erdos_741.basis : answer(sorry) ↔ ∀ A : Set ℕ, IsAddBasisOfOrder (A ∪ {0}) 2 → ∃ A₁ A₂,
+    A₁ ⊆ A ∧ A₂ ⊆ A ∧ A = A₁ ∪ A₂ ∧ Disjoint A₁ A₂ ∧ ¬ (IsSyndetic A₁ ∧ IsSyndetic A₂):= by
```
YaelDillies:
```suggestion
    A = A₁ ∪ A₂ ∧ Disjoint A₁ A₂ ∧ ¬ (IsSyndetic A₁ ∧ IsSyndetic A₂) := by
```

PR #1502, opened 2026-01-06
feat(ErdosProblems): 153

### LINKED ISSUE #390: Erdős Problem 153
### What is the conjecture
    
https://www.erdosproblems.com/153

  Let $A$ be a finite Sidon set and $A+A=\{s_1<\cdots<s_t\}$. Is it true that

$$\frac{1}{t}\sum_{1\leq i<t}(s_{i+1}-s_i)^2 \to \infty$$
as $\lvert A\rvert\to \infty$?

Status: open

### Prerequisites needed
None, Sidon sets are already in `ForMathlib`

### Choose either option

CoolRmal:
Closes #390

### DIFF
--- a/FormalConjectures/ErdosProblems/153.lean
+++ b/FormalConjectures/ErdosProblems/153.lean
@@ -29,31 +29,14 @@ import FormalConjectures.Util.ProblemImports
 open scoped Pointwise
 open Filter Finset Nat
 
-/-- Suppose `A` is a finite Sidon set of size `n`. Then `(A + A).card = (n + 1).choose 2`. -/
-@[category API, AMS 5]
-theorem IsSidon.add_card {A : Finset ℕ} {n : ℕ} (hn : A.card = n) (hA : IsSidon A) :
-    (A + A).card = (n + 1).choose 2 := by
-  have hs := Fintype.card_congr (Sym2.sortEquiv (α := A))
-  simp only [Sym2.card, Fintype.card_coe, hn] at hs
-  have ha := Fintype.card_coe (α := A × A) {s | s.1 ≤ s.2}
-  simp only [mem_filter, mem_univ, true_and] at ha
-  have h_card_eq : image (fun (p : A × A) => p.1.1 + p.2.1) {s | s.1 ≤ s.2} = A + A := by
-    ext; simp [Finset.mem_add]; grind
-  refine Eq.trans ?_ (hs.trans ha).symm
-  rw [← h_card_eq, Finset.card_image_of_injOn]
-  · intro p hp q hq hf
-    have := hA p.1 p.1.2 q.1 q.1.2 p.2 p.2.2 q.2 q.2.2 hf
-    simp [← Subtype.coe_le_coe] at hp hq
-    grind
-
 namespace Erdos153
 
 /-- Define `f n` to be the minimum of
 `∑ (i : Set.Ico 1 ((A + A).card), (s i - s (i - 1)) ^ 2 / n` as `A` ranges over all Sidon sets
 of size `n`, where `s` is an order embedding from `Fin n` into `A`. -/
 noncomputable def f (n : ℕ) : ℝ := ⨅ A : {A : Finset ℕ | A.card = n ∧ IsSidon A},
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/153.lean
```diff
@@ -0,0 +1,68 @@
...
+    Theory (1994), 329-347.
+-/
+
+open scoped Pointwise
+open Filter Finset Nat
+
+/-- Suppose `A` is a finite Sidon set of size `n`. Then `(A + A).card = (n + 1).choose 2`. -/
+@[category API, AMS 5]
+theorem IsSidon.add_card {A : Finset ℕ} {n : ℕ} (hn : A.card = n) (hA : IsSidon A) :
+    (A + A).card = (n + 1).choose 2 := by
```
YaelDillies: Can you move this to the earlier file?

FormalConjectures/ErdosProblems/153.lean
```diff
@@ -0,0 +1,68 @@
...
+  · intro p hp q hq hf
+    have := hA p.1 p.1.2 q.1 q.1.2 p.2 p.2.2 q.2 q.2.2 hf
+    simp [← Subtype.coe_le_coe] at hp hq
+    grind
+
+namespace Erdos153
+
+/-- Suppose `A` is a finite Sidon set of size `n`. We define `s` to be the increasing order
+isomorphism between `Fin (n.choose 2 + n)` and `A + A`. -/
+def s {A : Finset ℕ} {n : ℕ} (hn : A.card = n) (hA : IsSidon A) : Fin ((n + 1).choose 2) → A + A :=
```
YaelDillies: You don't need to know the size to write down the function and the statement
```suggestion
def s (A : Finset ℕ) : Fin (A + A).card → A + A :=
```

YaelDillies: In fact you don't even need to know that the set is of the form `A + A`!

FormalConjectures/ErdosProblems/153.lean
CoolRmal: May you be more specific about which file I should move this into? I think we can even create a new file named Sidon that contains API for Sidon sets.

YaelDillies: Oh yes, please create `ForMathlib.Combinatorics.Additive.Sidon`

FormalConjectures/ErdosProblems/153.lean
```diff
@@ -0,0 +1,63 @@
...
+    grind
+
+namespace Erdos153
+
+/-- Define `f n` to be the minimum of
+`∑ (i : Set.Ico 1 ((A + A).card), (s i - s (i - 1)) ^ 2 / n` as `A` ranges over all Sidon sets
+of size `n`, where `s` is an order embedding from `Fin n` into `A`. -/
+noncomputable def f (n : ℕ) : ℝ := ⨅ A : {A : Finset ℕ | A.card = n ∧ IsSidon A},
+  ∑ i : Set.Ico 1 ((A.1 + A).card), ((A.1 + A).orderIsoOfFin rfl ⟨i, i.2.2⟩ -
+  (A.1 + A).orderIsoOfFin rfl ⟨i - 1, by grind⟩) ^ 2 / (n : ℝ)
```
YaelDillies: What about
```suggestion
  let s := (A.1 + A).orderIsoOfFin rfl
  ∑ i \in {Fin.last (A.1 + A).card}^c, (s (i + 1) - s i) ^ 2 / (n : ℝ)
```

FormalConjectures/ErdosProblems/153.lean
YaelDillies: For the purpose of this PR, please simply remove this result as you don't need it anymore

FormalConjectures/ErdosProblems/153.lean
CoolRmal: `Fin.last (A.1 + A).card` has type `Fin (#(↑A + ↑A) + 1)`, while `s` needs an input of type `Fin #(↑A + ↑A)`.

This does not work even if we use `Fin.last ((A.1 + A).card - 1)` because `i` will be of type `Fin (#(↑A + ↑A) - 1 + 1)`.

PR #1503, opened 2026-01-06
CI: labeler

mo271:


mo271:
tested it in my fork,https://github.com/mo271/formal-conjectures/pull/6 and https://github.com/mo271/formal-conjectures/pull/7

.github/labeler.yml
```diff
@@ -0,0 +1,31 @@
...
+  - changed-files:
+    - any-glob-to-any-file: 'FormalConjectures/Arxiv/**'
+
+erdos-problems:
+  - changed-files:
+    - any-glob-to-any-file: 'FormalConjectures/ErdosProblems/**'
+
+ForMathlib:
+  - changed-files:
+    - any-glob-to-any-file: 'FormalConjectures/ForMathlib/**'
```
YaelDillies: Do we really need a label here?

mo271: Why not? Too much noise? 
I think this label might actually be useful if filtering out prs for upstreaming, but I'm also willing to skip it. 
We have ~never tagged anything `ForMathlib` manually, I suppose...

mo271: removed..

PR #1507, opened 2026-01-06
chore: split prerequisites to a separate Lean library

YaelDillies:
Create a new Lean library called `FormalConjecturesForMathlib` and move `ForMathlib` to it.

This will allow to selectively turn off warnings on `sorry` in problem statements while keeping them in the `ForMathlib` files.

### DIFF
--- a/FormalConjectures/Util/ForMathlib.lean
+++ b/FormalConjectures/Util/ForMathlib.lean
@@ -1,0 +1,18 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+import FormalConjecturesForMathlib
+
+deprecated_module (since := "2025-01-08")


FormalConjectures/Arxiv/0911.2077/Conjecture6_3.lean
```diff
@@ -13,8 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 -/
-
-import FormalConjectures.Util.ProblemImports
+import Prereqs
```
eric-wieser: I don't think we can reasonably claim this as a global module name

YaelDillies: This isn't public, eg it won't be on reservoir

eric-wieser: What do you mean? Anyone importing FormalConjectures will now not be able to make their own file called PreReqs.

YaelDillies: I can neither confirm nor infirm because it is hard to create a repo on v4.22.0 😐 I've renamed to `FCPrereqs`

eric-wieser:
Can you explain in the PR description what the motivation here is?

YaelDillies:
Done

eric-wieser:
Perhaps it's worth starting a Zulip thread about whether this is the best approach?

lakefile.toml
```diff
@@ -25,6 +25,11 @@ name = "mathlib"
 scope = "leanprover-community"
 rev = "v4.22.0"
 
+[[lean_lib]]
+name = "FCPrereqs"
```
eric-wieser: I think it might be better to either have:

* `FormalConjecturesPrereqs`

or:

* `FormalConjecturesForMathlib` (in case we want stricter mathlib-linting here)
* `FormalConjecturesUtils`

lakefile.toml
```diff
@@ -25,6 +25,11 @@ name = "mathlib"
...
+reservoir = false
```
eric-wieser: I don't think this does anything desirable.

FormalConjectures/Util/ForMathlib.lean
eric-wieser, line 1: Probably this should be a deprecated module, just in case.

lakefile.toml
```diff
@@ -25,6 +25,10 @@ name = "mathlib"
 scope = "leanprover-community"
 rev = "v4.22.0"
 
+[[lean_lib]]
+name = "FormalConjecturesForMathlib"
```
eric-wieser, line 30: Let's add a comment that notes we make this a separate library so that we can apply different lean options.

lakefile.toml
```diff
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name = "formal_conjectures"
+name = "formal-conjectures"
```
eric-wieser: Let's roll this back for now, it's out of scope and creates churn.

lakefile.toml
```diff
@@ -25,6 +25,10 @@ name = "mathlib"
 scope = "leanprover-community"
 rev = "v4.22.0"
 
+[[lean_lib]]
+name = "FormalConjecturesForMathlib"
+
 [[lean_lib]]
 name = "FormalConjectures"
 globs = ["FormalConjectures.+"]
+needs = ["FormalConjecturesForMathlib"]
```
eric-wieser: Does this do anything? I've not seen it before.

YaelDillies: It's used in `docbuild`

YaelDillies: But you are right that it's unnecessary, weird! Does that mean we can make cyclic dependencies between libraries in a single package?

PR #1510, opened 2026-01-06
chore(ErdosProblems/477): mark as disproved

YaelDillies:
and update to the new statement from https://www.erdosproblems.com/477

### DIFF
--- a/FormalConjectures/ErdosProblems/477.lean
+++ b/FormalConjectures/ErdosProblems/477.lean
@@ -24,22 +24,36 @@ import FormalConjectures.Util.ProblemImports
 
 namespace Erdos477
 
+-- TODO(lezeau): add the other (families of) counterexamples plus variants
+
+/--
+Let $f = X ^ 2$ be a polynomial of degree at least $2$. Then there is not set
+$A$ such that every $z\in \mathbb{Z}$ has exactly one representation as
+$z=a+f(n)$ for some $a\in A$ and $n > 0$
+-/
+@[category research solved, AMS 12]
+theorem erdos_477.variants.explicit_counterexample :
+    letI f := Polynomial.X ^ 2
+    ¬ ∃ A : Set ℤ, ∀ z, ∃! a ∈ A ×ˢ (f.eval '' {n | 0 < n}), z = a.1 + a.2 :=
+  sorry
+
 /--
 Let $f:\mathbb{Z}\to \mathbb{Z}$ be a polynomial of degree at least $2$. Is there a set $A$ such
 that every $z\in \mathbb{Z}$ has exactly one representation as $z=a+f(n)$ for some $a\in A$ and
 $n > 0$?
 -/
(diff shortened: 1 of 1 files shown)


mo271:
In the meantime the problem description on https://www.erdosproblems.com/forum/thread/477 has changed, but perhaps it is still good to push this version now, we can

mo271 (approved):

FormalConjectures/ErdosProblems/477.lean
```diff
@@ -19,27 +19,75 @@ import FormalConjectures.Util.ProblemImports
...
+Let $f = X ^ 2$ be a polynomial of degree at least $2$. Then there is no set
+$A$ such that every $z\in \mathbb{Z}$ has exactly one representation as
+$z=a+f(n)$ for some $a\in A$ and $n > 0$
+
+This is shown in [Sek59]
+-/
+@[category research solved, AMS 12]
+theorem erdos_477.S_sq :
+    letI f := X ^ 2
+    ∀ A : Set ℤ, ∃ z, ¬ ∃! a ∈ A ×ˢ (f.eval '' {n | 0 < n}), z = a.1 + a.2 := by
```
YaelDillies, line 50: The docstring and the statement don't match: what did you mean?

YaelDillies:
> In the meantime the problem description on https://www.erdosproblems.com/forum/thread/477 has changed, but perhaps it is still good to push this version now, we can

Was easy enough to update to the new ones

mo271 (approved):

FormalConjectures/ErdosProblems/477.lean
```diff
@@ -19,27 +19,75 @@ import FormalConjectures.Util.ProblemImports
...
+@[category research open, AMS 12]
+theorem erdos_477 : answer(sorry) ↔
+    ∃ f : ℤ[X], 2 ≤ f.degree ∧ ∃ A : Set ℤ,
+      ∀ z, ∃! ab ∈ A ×ˢ (f.eval '' {n | 0 < n}), z = ab.1 + ab.2 := by
+  sorry
+
+/--
+Let $f = X ^ 2$ be a polynomial of degree at least $2$. Then there is no set
+$A$ such that every $z\in \mathbb{Z}$ has exactly one representation as
+$z=a+f(n)$ for some $a\in A$ and $n > 0$
```
mo271:
```suggestion
Let $f = X ^ 2$ be a polynomial. Then for all set of integer $A$ there is a $z\in \mathbb{Z}$ that doesn't have exactly one representation as
$z=a+f(n)$ for some $a\in A$ and $n > 0$
```

FormalConjectures/ErdosProblems/477.lean
mo271: ah, must have manually `push_neg`ed the statement after writing the docstring...
Added a suggestion to improve it.

PR #1513, opened 2026-01-07
feat(ErdosProblems): 428

### LINKED ISSUE #725: Erdős Problem 428
### What is the conjecture
    
https://www.erdosproblems.com/428

  Is there a set $A\subseteq \mathbb{N}$ such that, for infinitely many $n$, all of $n-a$ are prime for all $a\in A$ with $0<a<n$ and
$$\liminf\frac{\lvert A\cap [1,x]\rvert}{\pi(x)}>0?$$
  

Status: open

### Choose either option

danielchin:
Closes #725

### DIFF
--- a/FormalConjectures/ErdosProblems/428.lean
+++ b/FormalConjectures/ErdosProblems/428.lean
@@ -1,0 +1,46 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


CoolRmal:
On https://www.erdosproblems.com/428, it also mentions the result that this conjecture is true if we assume the prime k-tuple conjecture and replace liminf by limsup. It would be great if we also have this. You can find out more information in FormalConjectures/Paper/PrimeTuples.lean

YaelDillies (approved):
Thanks!

PR #1515, opened 2026-01-07
chore(ErdosProblems/1064): mark as solved

mo271:


YaelDillies (commented):
Can you add the extra statement "for any function $f(n)=o(n)$, we have $\phi(n)>\phi(n-\phi(n))+f(n)$ for almost all $n$." and write the references in the docstrings?

YaelDillies (approved):

PR #1518, opened 2026-01-07
fix(ErdosProblems/330): fix the order of the minimal basis

mo271:
The definitions of `Rep`, `UnrepWithout`, and `MinAsymptoticAddBasis` are updated to include a parameter `h` representing the maximum number of terms in the sum. The main theorem statement now quantifies over such an order `h`.

### DIFF
--- a/FormalConjectures/ErdosProblems/330.lean
+++ b/FormalConjectures/ErdosProblems/330.lean
@@ -27,19 +27,19 @@ namespace Erdos330
 open Set
 open scoped BigOperators
 
-/-- `Rep A m` means `m` is a sum of **finitely many** elements of `A`
-    (i.e., representable by *some* finite number of terms, not a fixed order). -/
-def Rep (A : Set ℕ) (m : ℕ) : Prop :=
-  ∃ k : ℕ, ∃ f : Fin k → ℕ, (∀ i, f i ∈ A) ∧ (∑ i : Fin k, f i) = m
+/-- `Rep A m h` means `m` is a sum of at most `h` elements of `A`x. -/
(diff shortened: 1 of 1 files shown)


YaelDillies (commented):
That's not my interpretation of the problem. Can you give more context?

smmercuri:
> That's not my interpretation of the problem. Can you give more context?

An (asymptotic) [additive basis](https://github.com/google-deepmind/formal-conjectures/blob/3eac3cf9b5785b9caaac678ab27782523a492d12/FormalConjectures/ForMathlib/Combinatorics/Additive/Basis.lean#L33) `A` has an associated order `h` -- that is, almost all integers can be written as a sum of up to `h` terms from `A`. Presumably for `m` to be "represented" here means to express `m` as the sum of up to `h` terms from `A`.  The previous version allowed for an arbitrary number of terms in the sum representation of `m`. What is your interpretation?

YaelDillies:
I agree that "A is basis" means "exists h, A is basis of order h". But then "A is min basis" is "A is basis and forall a in A, A \ {a} isn't basis", ie "(exists h, A is basis of order h) and forall a in A, forall h, A \ {a} isn't basis of order h", while what you've written is "exists h, A is basis of order h and forall a in A, A \ {a} isn't basis of order h".

smmercuri:
No I think "A is min basis" means that there exists `h` such that `A` is a basis of order `h` and `A \ {a}` is not a basis _of order h_ for every `a` in `A`, which I think matches the formalisation in this PR? See [this article](https://www.sciencedirect.com/science/article/pii/S0022314X9892248X) for a literature example of this definition.

YaelDillies (approved):
Hmm okay, let's go with it

PR #1522, opened 2026-01-07
fix(ErdosProblems/659): make asymptotic

mo271:


FormalConjectures/ErdosProblems/659.lean
```diff
@@ -36,7 +36,7 @@ Is there a set of $n$ points in $\mathbb{R}^2$ such that every subset of $4$ poi
 least $3$ distances, yet the total number of distinct distances is $\ll \frac{n}{\sqrt{\log n}}$?
 -/
 @[category research open, AMS 52]
-theorem erdos_659 : answer(sorry) ↔ ∃ (a : ℕ → Finset ℝ²), ∀ n, #(a n) = n ∧
+theorem erdos_659 : answer(sorry) ↔ ∃ (a : ℕ → Finset ℝ²), ∀ᶠ n in Filter.atTop, #(a n) = n ∧
     3 ≤ minimalDistinctDistancesSubsetOfSize (a n) 4 ∧
     (fun n => (distinctDistances (a n) : ℝ)) ≪ fun (n : ℕ) => n / (n : ℝ).log.sqrt := by
```
YaelDillies: I don't understand what this is solving: `≪` already is asymptotic
```suggestion
theorem erdos_659 : answer(sorry) ↔ ∃ (a : ℕ → Finset ℝ²),
   (∀ n, #(a n) = n ∧ 3 ≤ minimalDistinctDistancesSubsetOfSize (a n) 4) ∧
    (fun n => (distinctDistances (a n) : ℝ)) ≪ fun (n : ℕ) => n / (n : ℝ).log.sqrt := by
```

mo271: @callesonne?

CoolRmal:
I just notice this PR and I'll close mine. I think we need to change `∀ n` into `∀ n ≥ 4` (as the problem only makes sense for ` n ≥ 4`). If we don't add this requirement, then one can prove the opposite by taking the advantage of the fact that `a 0` is an empty set and thus `3 ≤ minimalDistinctDistancesSubsetOfSize (a 0) 4 = 0`.

YaelDillies:
You're right. The issue ultimately is that `3 ≤ minimalDistinctDistancesSubsetOfSize (a n) 4` doesn't mean "every subset of $4$ points determines at least $3$ distances" if `a n` has size at most three. Let me fix this

FormalConjectures/ErdosProblems/659.lean
callesonne: I think it was a quick/lazy way of solving was an issue with `minimalDistinctDistancesSubsetOfSize`, where `n=0` was leading to issues. I thought that `n=1,2,3` would also give similar issues so to be safe I just put in `n` large enough (since like you say, the final statement is about asymptotics anyways). But your suggestion fixes this now in a cleaner way.

PR #1523, opened 2026-01-07
fix(ErdosProblems/74): assume subset of the edges

mo271:


FormalConjectures/ErdosProblems/74.lean
```diff
@@ -93,11 +93,10 @@ theorem SimpleGraph.subgraphEdgeDistsToBipartite_bddAbove (G : SimpleGraph V) (n
...
-  · use 0
-    rintro ⟨v, hv⟩ ⟨w, hw⟩ ⟨_, hvw⟩
-    aesop
-  · rfl
+  refine ⟨by rfl, ?_, rfl⟩
+  use fun _ => 0
+  simp
+
 
 /--
```
YaelDillies:
```suggestion
/--
```

PR #1528, opened 2026-01-07
fix(Wikipedia/WallSunSun): assume the discriminant is 0 or 1 mod 4

mo271:


FormalConjectures/Wikipedia/WallSunSun.lean
```diff
@@ -48,6 +48,8 @@ The discriminant of this number is the quantity $a^2 - 4b$. It is conjectured th
 infinitely many Lucas–Wieferich primes of any given discriminant.
 -/
 @[category research open, AMS 11]
-theorem infinite_isWallSunSunPrime_of_disc_eq (D : ℕ+) :
+theorem infinite_isWallSunSunPrime_of_disc_eq (A B : ℕ+) :
+    letI D : ℕ := A ^ 2 - 4 * B
+    D > 0 →
```
YaelDillies: Maybe assume instead D = 0 or 1 [4]?

mo271: @callesonne ?

callesonne: I'm happy with either

YaelDillies (approved):

FormalConjectures/Wikipedia/WallSunSun.lean
```diff
@@ -48,6 +48,6 @@ The discriminant of this number is the quantity $a^2 - 4b$. It is conjectured th
 infinitely many Lucas–Wieferich primes of any given discriminant.
 -/
 @[category research open, AMS 11]
-theorem infinite_isWallSunSunPrime_of_disc_eq (D : ℕ+) :
-    {p : ℕ | ∃ a b : ℕ, a ^ 2 - 4 * b = D ∧ IsLucasWieferichPrime a b p }.Infinite := by
+theorem infinite_isWallSunSunPrime_of_disc_eq {D : ℕ} (hD : D ≡ 0 [MOD 4] ∨ D ≡ 1 [MOD 4]) :
+    {p : ℕ | ∃ a b : ℕ, a ^ 2 - 4 * b = D ∧ IsLucasWieferichPrime a b p}.Infinite := by
```
callesonne, line 53: Now you are no longer assuming `D > 0`, have you checked that this is fine (requested change just so it didn't automerge)?

callesonne (changes requested):

FormalConjectures/Wikipedia/WallSunSun.lean
YaelDillies: Whoops yes you're right 😅

FormalConjectures/Wikipedia/WallSunSun.lean
```diff
@@ -48,6 +48,7 @@ The discriminant of this number is the quantity $a^2 - 4b$. It is conjectured th
 infinitely many Lucas–Wieferich primes of any given discriminant.
 -/
 @[category research open, AMS 11]
-theorem infinite_isWallSunSunPrime_of_disc_eq (D : ℕ+) :
-    {p : ℕ | ∃ a b : ℕ, a ^ 2 - 4 * b = D ∧ IsLucasWieferichPrime a b p }.Infinite := by
+theorem infinite_isWallSunSunPrime_of_disc_eq {D : ℕ} (hD₀ : 0 < D)
+    (hD : D ≡ 0 [MOD 4] ∨ D ≡ 1 [MOD 4]) :
+    {p : ℕ | ∃ a b : ℕ, a ^ 2 - 4 * b = D ∧ IsLucasWieferichPrime a b p}.Infinite := by
```
eric-wieser, line 53: Should this be integer subtraction instead of nat subtraction?

YaelDillies: That doesn't change the mathematical content since `D > 0`

eric-wieser: Why do we have `D>0` in the first place? I don't see it in the english.

YaelDillies: This is a mess. The Wikipedia conjecture is unsourced, and I cannot find back the source in an hour of searching... In the future we should enforce that conjectures are sourced

eric-wieser: My guess is that we added the condition ourselves to avoid this being incorrect when the subtraction truncates.

YaelDillies: No, this seems wrong. `D` is supposed to be the discriminant of a quadratic field, so it certainly cannot be a square

YaelDillies: #1813

PR #1533, opened 2026-01-07
feat(ErdosProblems): 683

### LINKED ISSUE #876: Erdős Problem 683
### What is the conjecture
    
https://www.erdosproblems.com/683

  Is it true that for every $0\leq k\leq n$ the largest prime divisor of $\binom{n}{k}$, say $P(\binom{n}{k})$, satisfies
$$P(\binom{n}{k})> \min(n-k+1, k^{1+c})$$
for some constant $c>0$?
  

Status:  open

### Choose either option

seewoo5:
Closes #876

### DIFF
--- a/FormalConjectures/ErdosProblems/961.lean
+++ b/FormalConjectures/ErdosProblems/961.lean
@@ -67,7 +67,7 @@ noncomputable def f (k : ℕ) : ℕ :=
 It is conjectured that $f(k) \ll (\log k)^O(1)$.
 -/
 @[category research open, AMS 11]
-theorem erdos_961 : answer(sorry) ↔ ∃ C > (0 : ℝ), ∀ᶠ k in atTop, f k < (log (k : ℝ)) ^ C := by
+theorem erdos_961 : answer(sorry) ↔ ∃ C : ℕ, ∀ᶠ k : ℕ in atTop, f k < log k ^ C := by
   sorry
 
 /--


FormalConjectures/ErdosProblems/683.lean
```diff
@@ -0,0 +1,75 @@
...
+| ⊥       => 0
+| (n : ℕ) => n
+
+/--
+Let $P(n, k)$ be the largest prime factor of $\binom{n}{k}$.
+Note that `(n.choose k).primeFactors.max` returns a `WithBot ℕ` to account for the empty set case,
+so we convert it to `ℕ` with `withBotToNat`. Note that we will only consider the case where
+$0 < k < n$, so $\binom{n}{k}$ is always greater than 1 and thus has at least one prime factor.
+-/
+def P (n k : ℕ) : ℕ := withBotToNat (n.choose k).primeFactors.max
```
YaelDillies: Why not
```suggestion
def P (n k : ℕ) : ℕ := (n.choose k).primeFactors.sup id
```

FormalConjectures/ErdosProblems/683.lean
```diff
@@ -0,0 +1,75 @@
...
+
+/--
+There exists $c > 0$ such that $P(n, k) > \min\{n-k+1, k^{1 + c}\}$ for all $0 \le k \le n$.}
+-/
+@[category research open, AMS 11]
+theorem erdos_683 : (∃ c > 0, ∀ n k : ℕ, 0 < k ∧ k < n → P n k > min (n - k + 1) (k ^ (1 + c))) ↔
+    answer(sorry) := by
```
YaelDillies: We have moved to putting `answer(sorry)` first, can you adapt?

YaelDillies:
Could you state https://www.erdosproblems.com/961 simultaneously?

FormalConjectures/ErdosProblems/683.lean
seewoo5: That's definitely way better, thanks!

seewoo5:
> Could you state [erdosproblems.com/961](https://www.erdosproblems.com/961) simultaneously?

Do you mean something like
```lean
/--
Erdos 961 is equivalent to Erdos 683.
-/
theorem erdos_683_equiv_erdos_961 :
    (∃ c > 0, ∀ n k : ℕ, 0 < k ∧ k < n → P n k > min (n - k + 1) (k ^ (1 + c))) ↔
    (∃ C > 0, ∀ᶠ k in atTop, f k < (log (k : ℝ)) ^ C) := by
  sorry
```
after importing Erdos961?

YaelDillies:
I was thinking you could *prove* one in terms of the other

FormalConjectures/ErdosProblems/683.lean
```diff
@@ -0,0 +1,68 @@
...
+Let $P(n, k)$ be the largest prime factor of $\binom{n}{k}$.
+-/
+def P (n k : ℕ) : ℕ := (n.choose k).primeFactors.sup id
+
+/--
+There exists $c > 0$ such that $P(n, k) > \min\{n-k+1, k^{1 + c}\}$ for all $0 \le k \le n$.}
+-/
+@[category research open, AMS 11]
+theorem erdos_683 : answer(sorry) ↔
+    (∃ c > 0, ∀ n k : ℕ, 0 < k ∧ k < n → P n k > min (n - k + 1) (k ^ (1 + c))) := by
```
smmercuri: I think `c` should have type `ℝ` here (currently it's `ℕ`). Similarly below

smmercuri: The informal has `le` here instead of `lt`
```suggestion
    (∃ c > 0, ∀ n k : ℕ, 0 ≤ k ∧ k ≤ n → P n k > min (n - k + 1) (k ^ (1 + c))) := by
```

FormalConjectures/ErdosProblems/683.lean
```diff
@@ -0,0 +1,87 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
```
YaelDillies:
```suggestion
Copyright 2026 The Formal Conjectures Authors.
```

FormalConjectures/ErdosProblems/683.lean
```diff
@@ -0,0 +1,87 @@
...
+
+open Filter Real Erdos961
+
+/--
+Let $P(n, k)$ be the largest prime factor of $\binom{n}{k}$.
+-/
+def P (n k : ℕ) : ℕ := (n.choose k).primeFactors.sup id
+
+/--
+There exists $c > 0$ such that $P(n, k) > \min\{n-k+1, k^{1 + c}\}$ for all $0 \le k \le n$.}
```
YaelDillies:
```suggestion
There exists $c > 0$ such that $P(n, k) > \min\{n-k+1, k^{1 + c}\}$ for all $0 < k < n$.}
```

FormalConjectures/ErdosProblems/961.lean
```diff
@@ -67,7 +67,7 @@ noncomputable def f (k : ℕ) : ℕ :=
 It is conjectured that $f(k) \ll (\log k)^O(1)$.
 -/
 @[category research open, AMS 11]
-theorem erdos_961 : answer(sorry) ↔ ∃ C > 0, ∀ᶠ k in atTop, f k < (log (k : ℝ)) ^ C := by
+theorem erdos_961 : answer(sorry) ↔ ∃ C > (0 : ℝ), ∀ᶠ k in atTop, f k < (log (k : ℝ)) ^ C := by
```
YaelDillies: This one was correct already

FormalConjectures/ErdosProblems/683.lean
YaelDillies: What's the largest prime divisor of 1?

smmercuri: ah I see, it was the informal docstring that had the error, the website is $1 \leq k ... $

mo271:
@seewoo5 also fine to do what @YaelDillies about 961 in a follow up or leave it as TODO....

FormalConjectures/ErdosProblems/961.lean
seewoo5: I double checked but it is `Nat` without change, where `^` is `@HPow.hPow ℝ ℕ ℝ instHPow (log ↑k) C : ℝ`.

seewoo5:
@mo271 I agree, I tried to add proofs but just had no time to do it these days...I could leave it as todo.
@YaelDillies What I'm not sure right now is if the equivalence statement is actually true or not. I think the correct one might be $e^{c\sqrt{k}}$ lower bound of $P(n, k)$ and $(\log k)^{c'}$ uppwer bound of $f(k)$, not in the current form. In the website it only says "This" is equivlent to 961 but not clear which upper bound of $f(k)$ is equivalent to "This".

YaelDillies:
> What I'm not sure right now is if the equivalence statement is actually true or not

Indeed, me too, that's why I suggested proving it. Happy to leave it as further work for now.

mo271:
@seewoo5 Is all that is left here to add some TODO?!

seewoo5:
@YaelDillies @mo271 Sorry, yes. I replace this as TODO for now. I also left comment on [Erdos961](https://www.erdosproblems.com/forum/thread/961) webpage - I don't think this is a hard problem, but also  didn't want to put wrong statement here. Feel free to resolve TODO later, or I can do it myself (when I become less busy..).

FormalConjectures/ErdosProblems/961.lean
YaelDillies: I agree, but the mathematical content is same (since `Nat` is cofinal in `Real`)

PR #1568, opened 2026-01-08
feat(ErdosProblems): 1068

### LINKED ISSUE #1109: Erdős Problem 1068
### What is the conjecture

https://www.erdosproblems.com/1068

  Does every graph with chromatic number $\aleph_1$ contain a countable subgraph which is infinitely connected?

Status:  open

### Choose either option

danielchin:
Closes #1109 

The InfinitelyConnected definition will be used in Erdos 1067 so we probably need to move this to a shared location in the future.

### DIFF
--- a/FormalConjectures/ErdosProblems/1068.lean
+++ b/FormalConjectures/ErdosProblems/1068.lean
@@ -47,8 +47,8 @@ infinitely connected?
 -/
 @[category research open, AMS 5]
 theorem erdos_1068 :
-  (∀ (V : Type*) (G : SimpleGraph V), G.chromaticNumber = aleph 1 →
-    ∃ (s : Set V), s.Countable ∧ InfinitelyConnected (G.induce s)) ↔ answer(sorry) := by
+  answer(sorry) ↔ ∀ (V : Type*) (G : SimpleGraph V), G.chromaticNumber = aleph 1 →
+    ∃ (s : Set V), s.Countable ∧ InfinitelyConnected (G.induce s) := by
   sorry
 
 end Erdos1068
--- a/FormalConjectures/ErdosProblems/1068.lean
+++ b/FormalConjectures/ErdosProblems/1068.lean
@@ -1,0 +1,54 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+import Mathlib.SetTheory.Cardinal.Ordinal
+
+/-!
+# Erdős Problem 1068
+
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/1068.lean
```diff
@@ -0,0 +1,53 @@
...
+
+open Cardinal
+
+namespace Erdos1068
+
+/--
+Two walks are internally disjoint if they share no vertices other than their endpoints.
+-/
+def InternallyDisjoint {V : Type*} {G : SimpleGraph V} {u v : V} (p q : G.Walk u v) : Prop :=
+  Disjoint ({x | x ∈ p.support} \ {u, v}) ({x | x ∈ q.support} \ {u, v})
```
YaelDillies: That definition is only correct for paths. Could you write the more generally correct one, eg by not coercing to a set?

FormalConjectures/ErdosProblems/1068.lean
```diff
@@ -0,0 +1,53 @@
...
+
+/--
+We say a graph is infinitely connected if any two vertices are connected by infinitely many
+pairwise disjoint paths.
+-/
+def InfinitelyConnected {V : Type*} (G : SimpleGraph V) : Prop :=
+  ∀ u v : V, u ≠ v → ∃ P : Set (G.Walk u v),
```
YaelDillies: I don't think you should exclude `u = v`?

FormalConjectures/ErdosProblems/1068.lean
danielchin: Rewrote this so it avoids using a set.

FormalConjectures/ErdosProblems/1068.lean
danielchin: I removed the exclusion, but (correct me I'm wrong) I don't think it matters? When `u = v` there's only one valid path which is a cycle to the same vertex which won't have infinite paths so that case will also just result to False.

YaelDillies: oh no sorry you're right. I had in mind that the definition was using `IsTrail`, not `IsPath`.

FormalConjectures/ErdosProblems/1068.lean
```diff
@@ -0,0 +1,53 @@
...
+-/
+
+open Cardinal List
+
+namespace Erdos1068
+
+/--
+Two walks are internally disjoint if they share no vertices other than their endpoints.
+-/
+def InternallyDisjoint {V : Type*} {G : SimpleGraph V} {u v : V} (p q : G.Walk u v) : Prop :=
```
YaelDillies: Can you make the walks be from different vertices?

FormalConjectures/ErdosProblems/1068.lean
```diff
@@ -0,0 +1,53 @@
...
+  Disjoint p.support.tail.dropLast q.support.tail.dropLast
+
+/--
+We say a graph is infinitely connected if any two vertices are connected by infinitely many
+pairwise disjoint paths.
+-/
+def InfinitelyConnected {V : Type*} (G : SimpleGraph V) : Prop :=
+  ∀ u v : V, ∃ P : Set (G.Walk u v),
```
YaelDillies:
```suggestion
  Pairwise fun u v \mapsto ∃ P : Set (G.Walk u v),
```

danielchin: Done.

FormalConjectures/ErdosProblems/1068.lean
danielchin: Done.

YaelDillies (approved):
Thanks!

PR #1569, opened 2026-01-08
fix `density_of_hyperbolicity_general_exponent`

mo271:


Paul-Lez (approved):

FormalConjectures/Wikipedia/Mandelbrot.lean
```diff
@@ -117,12 +117,12 @@ theorem MLC_general_exponent (n : ℕ) : LocallyConnectedSpace (multibrotSet n)
 `n`-periodic point (i.e. `f^[n] z = z`), `f^[n]` is differentiable at `z` and `‖deriv f^[n] z‖` is
 strictly less than one.-/
```
Paul-Lez, line 118:
```suggestion
`n`-periodic point (i.e. `f^[n] z = z`), `f^[n]` is differentiable at `z`, `‖deriv f^[n] z‖` is
strictly less than one, and `n > 0`. -/
```

mo271: #1571

PR #1573, opened 2026-01-08
fix(ErdosProblems/33)

### LINKED ISSUE #1543: Erdős Problem #33: sup/inf confusion?
https://github.com/google-deepmind/formal-conjectures/blob/234baea1a3a51f6f16689b06e9f77a1af8fcd552/FormalConjectures/ErdosProblems/33.lean#L64-L67

```
theorem erdos_33.variants.vanDoorn :
    ↑(2 * (φ ^ ((5 : ℝ) / 2))) ∈ lowerBounds { c : EReal | ∃ (A : Set ℕ), AdditiveBasisCondition A ∧
      Filter.atTop.limsup (fun N => (A.interIcc 1 N).ncard / (√N)) = c} := by
  -- Wait, there's a mistake. We can actually prove the opposite.
  negate_state;
  -- Proof starts here:
  rw [ mem_lowerBounds ] ; norm_num;
  use Set.univ;
  aesop;
  · exact fun k => ⟨ 0, k, by norm_num ⟩;
  · norm_num [ Set.ncard_eq_toFinset_card' ];
    rw [ Filter.limsup_eq ];
    rw [ show { a : ℝ | ∀ᶠ n : ℕ in Filter.atTop, Real.sqrt n ≤ a } = ∅ from _ ] <;> norm_num;
    · exact EReal.coe_pos.mpr ( by positivity );
    · exact Set.eq_empty_of_forall_notMem fun x hx => by rcases hx with ⟨ N, hN ⟩ ; exact absurd ( hN ( N + ⌊x ^ 2⌋₊ + 1 ) ( by linarith ) ) ( by push_cast; nlinarith [ Nat.lt_floor_add_one ( x ^ 2 ), Real.sqrt_nonneg ( N + ⌊x ^ 2⌋₊ + 1 : ℝ ), Real.mul_self_sqrt ( by positivity : ( 0 : ℝ ) ≤ N + ⌊x ^ 2⌋₊ + 1 ) ] ) ;
```

I think the issue is some kind of confusion about lower and upper bounds and the meaning thereof.  I briefly understood it at one point, but honestly I got super confused myself.

### Choose either option

CoolRmal:
Closes #1543

van Doorn's constructions implies that there exists a set `A` such that `lim sup |A ∩ {1, …, N}| / N^(1/2) ≤ (2 * (φ ^ (5 / 2)))`. Hence, the minimum value of `limsup` is less than `(2 * (φ ^ (5 / 2)))`.

Also fixes some typos.

### DIFF
--- a/FormalConjectures/ErdosProblems/33.lean
+++ b/FormalConjectures/ErdosProblems/33.lean
@@ -41,10 +41,8 @@ for some `a` in `A` and `n ≥ 0`. What is the smallest possible value of
 `lim sup n → ∞ |A ∩ {1, …, N}| / N^(1/2)`?
 -/
 @[category research open, AMS 11]
-theorem erdos_33 : IsLeast
-    { c : EReal | ∃ (A : Set ℕ), AdditiveBasisCondition A ∧
-      Filter.atTop.limsup (fun N => (A.interIcc 1 N).ncard / √N) = c}
-    answer(sorry) := by
+theorem erdos_33 : ⨅ A : {A : Set ℕ | AdditiveBasisCondition A}, Filter.atTop.limsup (fun N =>
+    (A.1.interIcc 1 N).ncard / (√N : EReal)) = answer(sorry) := by
   sorry
 
 /--


FormalConjectures/ErdosProblems/33.lean
```diff
@@ -38,13 +38,11 @@ def AdditiveBasisCondition (A : Set ℕ) : Prop :=
...
-`lim sup n → ∞ |A ∩ {1, …, N}| / N^(1/2) = 0`?
+`lim sup n → ∞ |A ∩ {1, …, N}| / N^(1/2)`?
 -/
 @[category research open, AMS 11]
-theorem erdos_33 : IsLeast
-    { c : EReal | ∃ (A : Set ℕ), AdditiveBasisCondition A ∧
-      Filter.atTop.limsup (fun N => (A.interIcc 1 N).ncard / √N) = c}
-    answer(sorry) := by
+theorem erdos_33 : ⨅ A : {A : Set ℕ | AdditiveBasisCondition A}, Filter.atTop.limsup (fun N =>
+    (A.1.interIcc 1 N).ncard / (√N : EReal)) = answer(sorry) := by
```
YaelDillies, line 45: I'm not sure I understand this change. Can you explain?

CoolRmal: The infimum of a set `s` may not belong to that set, while `IsLeast s ∈ s`. Based on the problem description, I don't think we know that the infimum is attained by some element.

YaelDillies: Ah yes I see. I was reading this as `IsGLB`

YaelDillies (approved):
Thanks!

PR #1577, opened 2026-01-09
feat(ErdosProblems): 194

danielchin:
I didn't actually see an issue created for Erdos 194, but nonetheless, there's no existing lean file so trying to formalize it regardless.


FormalConjectures/ErdosProblems/194.lean
```diff
@@ -0,0 +1,40 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 194
+
+*Reference:*
```
mo271:
```suggestion
*References:*
```

FormalConjectures/ErdosProblems/194.lean
```diff
@@ -0,0 +1,40 @@
...
+- [erdosproblems.com/194](https://www.erdosproblems.com/194)
+- [ABJ11] Ardal, H. and Brown, T. and Jungi\'C, V., Chaotic orderings of the rationals and reals. Amer. Math. Monthly (2011), 921-925.
+-/
+
+namespace Erdos194
+
+/--
+Let $k\geq 3$. Must any ordering of $\mathbb{R}$ contain a monotone $k$-term arithmetic progression?
```
mo271:
```suggestion
Let $k\geq 3$. Must any ordering of $\mathbb{R}$ contain a monotone $k$-term arithmetic progression,
that is, some $x_1 <\cdots < x_k$ which forms an increasing or decreasing $k$-term
arithmetic progression?
```

often good to stick with the formulation on erdosproblems.com verbatim, this way it is easier not to introduce new errors...

FormalConjectures/ErdosProblems/194.lean
mo271:
```suggestion
*References:*
```

mo271 (approved):
Thanks, LGTM!

mo271:
> I didn't actually see an issue created for Erdos 194, but nonetheless, there's no existing lean file so trying to formalize it regardless.

I only created issues for *open* Erdős Problems, but it is also welcome to tackle formatlizing the statement of solved ones!

FormalConjectures/ErdosProblems/194.lean
```diff
@@ -0,0 +1,40 @@
...
+
+The answer is no, even for $k=3$, as shown by Ardal, Brown, and Jungi\'{c} [ABJ11].
+-/
+@[category research solved, AMS 5]
+theorem erdos_194 :
+  answer(False) ↔ ∀ k ≥ 3, ∀ r : ℝ → ℝ → Prop, IsStrictTotalOrder ℝ r →
```
CoolRmal, line 38: Are you sure they want `r` to be a strict total ordering? It seems like that in [ABJ11] they deal with linear orderings.

FormalConjectures/ErdosProblems/194.lean
```diff
@@ -0,0 +1,40 @@
...
+- [erdosproblems.com/194](https://www.erdosproblems.com/194)
+- [ABJ11] Ardal, H. and Brown, T. and Jungi\'C, V., Chaotic orderings of the rationals and reals. Amer. Math. Monthly (2011), 921-925.
+-/
+
+namespace Erdos194
+
+/--
+Let $k\geq 3$. Must any ordering of $\mathbb{R}$ contain a monotone $k$-term arithmetic progression?
+
+The answer is no, even for $k=3$, as shown by Ardal, Brown, and Jungi\'{c} [ABJ11].
```
CoolRmal:
```suggestion
The answer is no, even for $k=3$, as shown by Ardal, Brown, and Jungić [ABJ11].
```

FormalConjectures/ErdosProblems/194.lean
```diff
@@ -0,0 +1,40 @@
...
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 194
+
+*Reference:*
+- [erdosproblems.com/194](https://www.erdosproblems.com/194)
+- [ABJ11] Ardal, H. and Brown, T. and Jungi\'C, V., Chaotic orderings of the rationals and reals. Amer. Math. Monthly (2011), 921-925.
```
CoolRmal:
```suggestion
- [ABJ11] Ardal, H. and Brown, T. and Jungić, V., Chaotic orderings of the rationals and reals. Amer. Math. Monthly (2011), 921-925.
```

FormalConjectures/ErdosProblems/194.lean
danielchin: Done.

FormalConjectures/ErdosProblems/194.lean
danielchin: Done.

FormalConjectures/ErdosProblems/194.lean
danielchin: Done.

FormalConjectures/ErdosProblems/194.lean
danielchin: Done.

FormalConjectures/ErdosProblems/194.lean
danielchin: I believe this is fine in terms of implementation. Even though the authors state linear ordering, I think we can still use `IsStrictTotalOrder` because we only care about orderings with distinct elements. If we included non-strict orderings, we would have to deal with constant sequences that are valid monotone arithmetic progressions.

danielchin:
@mo271 I believe this should be good to go if you want to merge.

mo271 (approved):
thanks!

PR #1581, opened 2026-01-09
chore: create a FormalConjecturesTest library

Paul-Lez:
This PR sets up a separate test library with tests for definitions and some of the metaprograms from `FormalConjectures/Util`. The intent here is to have a slightly clearer repo structure.

Drive by: few stylistic fixes, and changed `Util/Attributes.lean` -> `Util/Attributes/Basic.lean`.


FormalConjecturesTest/Util/Attributes/AttributesTest.lean
YaelDillies, line 1: Surely this should be `.Test`?

FormalConjectures/Util/Attributes/Attributes.lean
YaelDillies, line 1: Surely this should be `.Basic`?

YaelDillies (commented):
Are you intending this new library to allow warnings?

Paul-Lez:
> Are you intending this new library to allow warnings?

@YaelDillies Allowing warnings seems reasonable to me - what do you think?

YaelDillies:
I was hoping we would instead remove them with `#guard_msgs`

Paul-Lez:
> I was hoping we would instead remove them with `#guard_msgs`

Sounds reasonable. I've turned off sorry warnings - are there any other options I should turn off?

YaelDillies:
Wait, why would you turn off sorry warnings here? Will this library use sorry outside out tests (where they should be caught by `#guard_msgs`)?

Paul-Lez:
I think it's reasonable to turn sorry warnings off in tests with `guard_msg` because they create unnecessary noise in the docstring that is captured

YaelDillies:
Can you fix the conflicts?

YaelDillies:
You'll need to fix `overwrite_index`

FormalConjecturesTest/Util/Attributes/Test.lean
YaelDillies, line 1: Why not `FormalConjecturesTest/Util/Attributes.lean`? Same below

Paul-Lez: I was following the suggestion in https://github.com/google-deepmind/formal-conjectures/pull/1581#discussion_r2677094499 :)

Paul-Lez: But since we might be expanding the attributes soon (and not all functionalities are tested for currently), I think it could be worth keeping the Attributes subdirectory of FormalConjecturesTest (and maybe rename the file in there later when we expand tests)

YaelDillies: Ah sorry, I originally didn't spot you also moved it to the new sublibrary

All I want is to avoid several parts of the name carrying the same meaning. Right now, you have the `Test` file in the `Test` library

Paul-Lez: Renamed!

FormalConjecturesTest/Util/Attributes/Basic.lean
```diff
@@ -14,27 +14,37 @@ See the License for the specific language governing permissions and
...
 theorem an_open_problem : Transcendental ℝ (π + rexp 1) := by
   sorry
 
+
+#guard_msgs in
+@[category research formally solved using lean4 at "https://github.com/example/formal-proof"]
+theorem a_formally_solved_problem : 2 + 2 = 4 := by
+  rfl
+
+
```
YaelDillies: Too many linebreaks

FormalConjectures/Util/Attributes/Basic.lean
YaelDillies, line 1: Can't you do this rename when you actually add more `Attributes` material?

docbuild/scripts/overwrite_index.lean
```diff
@@ -146,6 +146,7 @@ classifications, please refer to the
 ## Repository growth
 "
     IO.println markdownBody
-    let .some newBody := MD4Lean.renderHtml (parserFlags := MD4Lean.MD_FLAG_TABLES ) markdownBody | throwError "Parsing failed"
+    let .some newBody := MD4Lean.renderHtml (parserFlags := MD4Lean.MD_FLAG_TABLES ) markdownBody
+      | throwError "Parsing failed"
```
YaelDillies, line 150: Unrelated?

FormalConjecturesTest/Util/Attributes/Basic.lean
YaelDillies, line 1: Similarly, this should probably stay `.Util.Attributes` for the time being (/forever, since a single test file could righly test the content of several non-test files)

FormalConjectures/Util/Attributes/Basic.lean
Paul-Lez: I think this isn't unreasonable because we already have some other attributes material in the file `AMS.lean`

YaelDillies: This looks like scope-creep in this PR, at least

FormalConjecturesTest/Util/Attributes/Basic.lean
Paul-Lez: One nice aspect of the organising files this way is that we get a structure that mimics that of the directory we're testing on - I think at the end of the day this makes things a little clearer/easier to find

FormalConjectures/Util/Attributes/Basic.lean
Paul-Lez: Indeed:D - I can mention that in the PR description if you want:)

docbuild/scripts/overwrite_index.lean
Paul-Lez: Oh yeah I just got a little annoyed by the long line:)

FormalConjecturesTest/Util/Attributes/Basic.lean
Paul-Lez: Fixed.

FormalConjectures/Util/Attributes/Basic.lean
YaelDillies: Yeah, would be nice

YaelDillies (approved):

PR #1585, opened 2026-01-10
fix(ErdosProblems/828): typo

### LINKED ISSUE #1537: Erdős Problem #828
https://github.com/google-deepmind/formal-conjectures/blob/234baea1a3a51f6f16689b06e9f77a1af8fcd552/FormalConjectures/ErdosProblems/828.lean#L48-L50

```
theorem erdos_828.variants.phi_dvd_self_iff_pow2_pow3 {n : ℕ} (hn : 1 > n):
    φ n ∣ n ↔ ∃ᵉ (a > 0) (b), n = 2^a * 3^b := by
  -- Wait, there's a mistake. We can actually prove the opposite.
  negate_state;
  -- Proof starts here:
  -- Let's choose any $n$ such that $n > 1$ and $\phi(n) \mid n - 1$.
  use 0; norm_num
```

### Choose either option

CoolRmal:
Closes #1537

Shall I also include the proof for this? I have a proof for the easy direction. The other direction is a bit long and but I can probably finish it within the next two days.

### DIFF
--- a/FormalConjectures/ErdosProblems/828.lean
+++ b/FormalConjectures/ErdosProblems/828.lean
@@ -42,7 +42,8 @@ theorem erdos_828.variants.lehmer_conjecture : answer(sorry) ↔ ∀ n > 1, φ n
   sorry
 
 /--
-It is an easy exercise to show that $\phi(n) | n$ if and only if $n = 0, 1$ or $n = 2^a 3^b$.
+It is an easy exercise to show that $\phi(n) | n$ if and only if $n = 0, 1$ or $n = 2^a 3^b$ for
+some $a > 0$.
 -/
 @[category undergraduate, AMS 11]
 theorem erdos_828.variants.phi_dvd_self_iff_pow2_pow3 {n : ℕ} :


FormalConjectures/ErdosProblems/828.lean
```diff
@@ -40,13 +40,12 @@ When $n > 1$, Lehmer conjectured that $\phi(n) | n - 1$ if and only if $n$ is pr
 @[category research open, AMS 11]
 theorem erdos_828.variants.lehmer_conjecture : answer(sorry) ↔ ∀ n > 1, φ n ∣ n - 1 ↔ Prime n := by
   sorry
-
 /--
```
YaelDillies, line 44:
```suggestion

/--
```

FormalConjectures/ErdosProblems/828.lean
```diff
@@ -40,13 +40,12 @@ When $n > 1$, Lehmer conjectured that $\phi(n) | n - 1$ if and only if $n$ is pr
...
-It is an easy exercise to show that, for $n > 1$, $\phi(n) | n$ if and only if $n = 2^a 3^b$.
+It is an easy exercise to show that $\phi(n) | n$ if and only if $n = 0, 1$ or $n = 2^a 3^b$.
```
YaelDillies: `a` must be non-zero too

CoolRmal: Actually `φ 0 = 0` and `0 ∣ 0`!

CoolRmal:
I'll provide a proof in another PR

FormalConjectures/ErdosProblems/828.lean
YaelDillies: You misread my message 😉 `φ 3` doesn't divide `3`

CoolRmal: Oh my bad, I just changed the docstring.

PR #1612, opened 2026-01-10
feat(GreensOpenProblems): 30

### LINKED ISSUE #1565: Green's Open Problems #30
### What is the conjecture
https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.30

Given a set $A \subset \mathbb{Z}$ with $D(A) \leq K$, find a large structured subset $A′$ which ‘obviously’ has $D(A′) \leq K + \varepsilon$ (Here, $D(A) := |A - A|/|A|$).

### Choose either option

akshayhat:
Adds Ben Green’s Open Problem #30.

Closes #1565.

### DIFF
--- a/FormalConjectures/ErdosProblems/828.lean
+++ b/FormalConjectures/ErdosProblems/828.lean
@@ -42,11 +42,12 @@ theorem erdos_828.variants.lehmer_conjecture : answer(sorry) ↔ ∀ n > 1, φ n
   sorry
 
 /--
-It is an easy exercise to show that, for $n > 1$, $\phi(n) | n$ if and only if $n = 2^a 3^b$.
+It is an easy exercise to show that $\phi(n) | n$ if and only if $n = 0, 1$ or $n = 2^a 3^b$ for
+some $a > 0$.
 -/
 @[category undergraduate, AMS 11]
-theorem erdos_828.variants.phi_dvd_self_iff_pow2_pow3 {n : ℕ} (hn : 1 > n):
-    φ n ∣ n ↔ ∃ᵉ (a > 0) (b), n = 2^a * 3^b := by
+theorem erdos_828.variants.phi_dvd_self_iff_pow2_pow3 {n : ℕ} :
+    φ n ∣ n ↔ n ≤ 1 ∨ ∃ᵉ (a > 0) (b), n = 2 ^ a * 3 ^ b := by
   sorry
 
 end Erdos828


rao107:
Hi, thanks for your contribution! Just a note, you should move your addition into a new file `30.lean` in the same folder to match the rest of the project. Also, how did you land on defining "large" and "structured" in a "Freiman-type sense"?

FormalConjectures/GreensOpenProblems/1.lean
```diff
@@ -39,3 +39,34 @@ theorem green_1 : answer(sorry) ↔ ∃ Ω : ℕ → ℝ, atTop.Tendsto Ω atTop
...
+@[category research open, AMS 5 11]
+theorem green_30 :
+  answer(sorry) ↔
+  ∀ (A : Finset ℤ) (K ε : ℝ),
+    0 < ε →
+    A.card ≠ 0 →
+    ((A.image (fun a => a)) \ A).card ≤ (K * A.card) →
+    ∃ (A' : Finset ℤ),
+      A' ⊆ A ∧
+      (True) := by
```
rao107, line 68: Why ∧ (True)?

akshayhat: Thanks for the feedback! I’ll move Green’s Open Problem #30 into a separate file `30.lean` to match the project structure.

By “large” and “structured” I had in mind a Freiman-type interpretation: “large” meaning |A'| is a positive proportion of |A| depending on K and ε, and “structured” meaning that A' is contained in a bounded-rank generalized arithmetic progression. I kept the statement weak for now, but I’m happy to adapt it to the preferred level of precision.

rao107: No, I'm not commenting on the math or anything. I just wanted to know why use that definition over any others. The original problem didn't mention Freiman (though the paper linked in the problem's discussion does). Just want to make sure we're formalizing the statement in the same spirit that Green wrote it.

akshayhat: That choice wasn’t meant to privilege Freiman’s formulation — it was simply the first concrete way I thought to make the informal terms “large” and “structured” precise, since small-doubling problems are often discussed in that language.
But I agree the original problem is deliberately noncommittal, so I’ll keep the statement closer to Green’s wording rather than fixing a specific interpretation.

Paul-Lez: The `∧ (True)` shouldn't be in the formalisation

YaelDillies:
This isn't going anywhere, we are talking straight to an LLM here.

PR #1614, opened 2026-01-10
feat(ErdosProblems): 238

### LINKED ISSUE #438: Erdős Problem 238
### What is the conjecture
    
https://www.erdosproblems.com/238

  Let $c_1,c_2>0$. Is it true that, for any sufficiently large $x$, there exist more than $c_1\log x$ many consecutive primes $\leq x$ such that the difference between any two is $>c_2$?

Status: open

### Choose either option

CoolRmal:
Closes #438

### DIFF
--- a/FormalConjectures/ErdosProblems/238.lean
+++ b/FormalConjectures/ErdosProblems/238.lean
@@ -41,7 +41,7 @@ theorem erdos_238 : answer(sorry) ↔ ∀ᵉ (c₁ > 0) (c₂ > 0), ∀ᶠ (x : 
 It is well-known that the conjecture above is true when `c₁` is sufficiently small.
 -/
 @[category research solved, AMS 11]
-theorem erdos_238.variant : ∀ᶠ c₁ in 𝓝[>] 0, ∀ c₂ > 0, ∀ᶠ (x : ℝ) in atTop, ∃ (k : ℕ),
+theorem erdos_238.variant : ∀ c₂ > 0, ∀ᶠ c₁ in 𝓝[>] 0, ∀ᶠ (x : ℝ) in atTop, ∃ (k : ℕ),
     c₁ * log x < k ∧ ∃ f : Fin k → ℕ, ∃ m, (∀ i, f i ≤ x ∧ f i = (m + i.1).nth Nat.Prime)
     ∧ ∀ i : Fin (k - 1), c₂ < primeGap (m + i.1) := by
   sorry


FormalConjectures/ErdosProblems/238.lean
```diff
@@ -0,0 +1,50 @@
...
+theorem erdos_238 : answer(sorry) ↔ ∀ᵉ (c₁ > 0) (c₂ > 0), ∀ᶠ (x : ℝ) in atTop, ∃ (k : ℕ),
+    c₁ * x.log < k ∧ ∃ f : Fin k → ℕ, ∃ m, (∀ i, f i ≤ x ∧ f i = (m + i.1).nth Nat.Prime)
+    ∧ ∀ i : Fin (k - 1), c₂ < primeGap (m + i.1) := by
+  sorry
+
+/--
+It is well-known that the conjecture above is true when `c₁` is sufficiently small.
+-/
+@[category research solved, AMS 11]
+theorem erdos_238.variant : ∀ᶠ c₁ in 𝓝 0, ∀ c₂ > 0, ∀ᶠ (x : ℝ) in atTop, ∃ (k : ℕ),
```
CoolRmal: I am not sure about the ordering of the quantifier here: `∀ᶠ c₁ in 𝓝 0, ∀ c₂ > 0` and `∀ c₂ > 0, ∀ᶠ c₁ in 𝓝 0` both seem plausible. The first one is the stronger one (there is a fixed `ε` such that if `c₁ < ε`, the statement is true for all `c₂ > 0`, while in the second one the choice of `ε` depends on `c₂`).

FormalConjectures/ErdosProblems/238.lean
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Erdos238
+
+/--
+Let `c₁, c₂ > 0`. Is it true that for any sufficiently large `x`, there exists more than
+`c₁ * log x` many consecutive primes `≤ x` such that the difference between any two is `> c₂`?
+-/
+@[category research open, AMS 11]
+theorem erdos_238 : answer(sorry) ↔ ∀ᵉ (c₁ > 0) (c₂ > 0), ∀ᶠ (x : ℝ) in atTop, ∃ (k : ℕ),
+    c₁ * x.log < k ∧ ∃ f : Fin k → ℕ, ∃ m, (∀ i, f i ≤ x ∧ f i = (m + i.1).nth Nat.Prime)
```
YaelDillies: Could we avoid using dot notation on these classical operators?
```suggestion
    c₁ * log x < k ∧ ∃ f : Fin k → ℕ, ∃ m, (∀ i, f i ≤ x ∧ f i = (m + i.1).nth Nat.Prime)
```
with `open Real`

FormalConjectures/ErdosProblems/238.lean
YaelDillies:
```suggestion
theorem erdos_238.variant : ∀ᶠ c₁ in 𝓝[>] 0, ∀ c₂ > 0, ∀ᶠ (x : ℝ) in atTop, ∃ (k : ℕ),
```

YaelDillies: I agree with your interpretation to the extent that the other one would have stated "... in terms of c_2", but maybe worth waiting for clarifications

CoolRmal: I eventually decided to use the second formulation because I think it is what Erdos intended to mean. You can see the discussion on https://www.erdosproblems.com/forum/thread/238

PR #1615, opened 2026-01-10
CI: warning for `research open` problems with proof

mo271:


YaelDillies:
From the README
> Solved research problem: a mathematical problem that has an accepted solution (in the sense that it is widely accepted by experts in the field). In particular this does not mean that the statement has been proven in the formal setting.

Is your rationale that we won't add formal proofs to this repo before posting them to the website?

mo271:
> From the README
> 
> > Solved research problem: a mathematical problem that has an accepted solution (in the sense that it is widely accepted by experts in the field). In particular this does not mean that the statement has been proven in the formal setting.
> 
> Is your rationale that we won't add formal proofs to this repo before posting them to the website?

Perhaps we should amend the README: obviously if there is a formal proof, then this particular formal statement is not open anymore (of course it could be a misformalisation, and I suppose in `solved` could then mean "it is accepted by experts in the community that the informal formulation is faithfully represented by the formal statement). 

I think we could possibly even adapt a more fine-grained categorization: 

1.   "formally solved here": the theorem has a sorry-free proof, included here
2.   "formally solved somewhere": the theorem has a formal proof that compiles at the commit where this comment is added, but the proof is in a different branch/ place
3.  "formally solved somehow": a different formalisation of the informal statement including proof exists (say in some different repo using Lean4, or Lean 3 or even some other formal language)
4.  "informally solved": the informal statement has an "accepted" informal proof

 
 No I suppose we can be a the curious situation that one obtains a formal proof without having an accepted informal proof...
 
 I will propose an edit to the README in this pull request
 
 In any case, is just seems wrong to put the label `research open` to a formal statement that comes with a complete formal proof, no?
 
 Also this warning is mostly useful when people copy and paste variants around during editing and forget to adjust the label for a variant they provide a proof for.

YaelDillies:
Yeah I think we could have `solved` and `formally solved`, with the former being stronger than the latter, although maybe that would be surprising

mo271:
> Yeah I think we could have `solved` and `formally solved`, with the former being stronger than the latter, although maybe that would be surprising

Make sense, but perhaps not in this pull request?!

FormalConjectures/Util/Attributes.lean
```diff
@@ -222,8 +222,14 @@ initialize Lean.registerBuiltinAttribute {
         let cat ← Syntax.toCategory s
         return (cat, "")
       | _ => throwUnsupportedSyntax
+    if status == Category.research ProblemStatus.open then
```
YaelDillies:
```suggestion
    if status == .research .open then
```
if you want

FormalConjectures/Util/Attributes.lean
```diff
@@ -222,8 +222,14 @@ initialize Lean.registerBuiltinAttribute {
...
+      let env ← getEnv
+      if let some info := env.find? decl then
+        if let some val := info.value? then
+          if !val.hasSorry then
+             logWarning "If a problem has a sorry-free proof, it should not be categorised as `open`."
```
YaelDillies:
```suggestion
      let some info := env.find? decl
      let some val := info.value?
      guard !val.hasSorry
      logWarning "If a problem has a sorry-free proof, it should not be categorised as `open`."
```
to respect the line of sight rule (untested)

mo271: let's test it...

Paul-Lez:
> > Yeah I think we could have `solved` and `formally solved`, with the former being stronger than the latter, although maybe that would be surprising
> 
> 
> 
> Make sense, but perhaps not in this pull request?!

I can update the attribute to have this extra category!

mo271:
@Paul-Lez 
> I can update the attribute to have this extra category!

Throwing out this idea: I think it would be useful to have a `formally solved here [link]` category with the following meaning: 
[link] here is a github-permalink (or link to github gist) to a proof that at the version where the commit where the category label was added could solve the sorry.

I think this would be most useful to keep track of which is the solved problems have been formalized without adding all the proofs to our repo.

Issue #1103, opened 2025-10-14
Erdős Problem 1061


mo271:
### What is the conjecture

https://www.erdosproblems.com/1061

  How many solutions are there to 
$$\sigma(a)+\sigma(b)=\sigma(a+b)$$
with $a+b\leq x$, where $\sigma$ is the sum of divisors function? Is it $\sim cx$ for some constant $c>0$?

Status:  open

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

PR #1618, opened 2026-01-11
feat(ErdosProblems): 1061

conglu1997:
Closes https://github.com/google-deepmind/formal-conjectures/issues/1103.

Taking the convention that (a, b) is an ordered pair, but please lmk if I should swap!

### DIFF
--- a/FormalConjectures/ErdosProblems/1061.lean
+++ b/FormalConjectures/ErdosProblems/1061.lean
@@ -36,7 +36,7 @@ In particular, `(a, b)` and `(b, a)` are counted separately; an unordered varian
 by additionally requiring `a ≤ b`. -/
 noncomputable abbrev S (x : ℝ) : ℝ :=
   ((Finset.Icc 1 ⌊x⌋₊ ×ˢ Finset.Icc 1 ⌊x⌋₊).filter fun (a, b) ↦
-      a + b ≤ ⌊x⌋₊ ∧ σ 1 a + σ 1 b = σ 1 (a + b)).card
+      a + b ≤ x ∧ σ 1 a + σ 1 b = σ 1 (a + b)).card
 
 /--
 How many (ordered) solutions are there to `σ(a) + σ(b) = σ(a + b)` with `a + b ≤ x`?


FormalConjectures/ErdosProblems/1061.lean
```diff
@@ -0,0 +1,49 @@
...
+namespace Erdos1061
+
+/-- Let `S x` count the number of **ordered** pairs of positive integers `(a, b)` with `a + b ≤ x`
+such that `σ(a) + σ(b) = σ(a + b)`, where `σ` is the sum of divisors function.
+
+In particular, `(a, b)` and `(b, a)` are counted separately; an unordered variant could be obtained
+by additionally requiring `a ≤ b`. -/
+noncomputable abbrev S (x : ℝ) : ℝ :=
+  ((Finset.Icc 1 ⌊x⌋₊ ×ˢ Finset.Icc 1 ⌊x⌋₊).filter fun (a, b) ↦
+      a + b ≤ ⌊x⌋₊ ∧ σ 1 a + σ 1 b = σ 1 (a + b)).card
```
callesonne:
```suggestion
      a + b ≤ x ∧ σ 1 a + σ 1 b = σ 1 (a + b)).card
```
The floor is unecessary here

FormalConjectures/ErdosProblems/1061.lean
```diff
@@ -0,0 +1,49 @@
...
+open Filter Asymptotics
+open scoped ArithmeticFunction
+
+namespace Erdos1061
+
+/-- Let `S x` count the number of **ordered** pairs of positive integers `(a, b)` with `a + b ≤ x`
+such that `σ(a) + σ(b) = σ(a + b)`, where `σ` is the sum of divisors function.
+
+In particular, `(a, b)` and `(b, a)` are counted separately; an unordered variant could be obtained
+by additionally requiring `a ≤ b`. -/
```
callesonne, line 36: I would say that we should use the unordered variant. But I would like someone elses opinion on this first @YaelDillies @mo271

callesonne (approved):
Generally LGTM! Thanks :)

conglu1997:
Suggestion adopted, thanks! Will follow advice on the ordered/unordered pair convention.

FormalConjectures/ErdosProblems/1061.lean
YaelDillies: I think we should use the ordered variant. Of course, it doesn't change anything, since `sigma (2 *  a) = 2 * sigma a` has no solution

YaelDillies:
Thanks!

PR #1624, opened 2026-01-11
feat(ErdosProblems): 855

### LINKED ISSUE #980: Erdős Problem 855
### What is the conjecture

https://www.erdosproblems.com/855

  If $\pi(x)$ counts the number of primes in $[1,x]$ then is it true that (for large $x$ and $y$)
$$\pi(x+y) \leq \pi(x)+\pi(y)?$$

Status:  open

### Choose either option

violet-hurricane:
fixes #980 

This PR adds a formal statement of Erdős Problem 855 (prime counting subadditivity).

I agree to the contribution guidelines.

### DIFF
--- a/FormalConjectures/ErdosProblems/855.lean
+++ b/FormalConjectures/ErdosProblems/855.lean
@@ -29,10 +29,8 @@ open scoped Nat.Prime
 namespace Erdos855
 
 @[category research open, AMS 11]
-theorem erdos_855 :
-    answer(sorry) ↔
-        (∀ᶠ x in atTop, ∀ᶠ y in atTop,
-        π (x + y) ≤ π x + π y) := by
+theorem erdos_855 : answer(sorry) ↔
+    ∀ᶠ x in atTop, ∀ᶠ y in atTop, π (x + y) ≤ π x + π y := by
   sorry
 
 end Erdos855


mo271:
Thanks for the contributions!

> I agree to the contribution guidelines.

Great, the only missing step is the CLA here: https://github.com/google-deepmind/formal-conjectures/pull/1624/checks?check_run_id=60052492069

mo271:
Note that we already have 
https://github.com/google-deepmind/formal-conjectures/blob/351e286843680a9fabf2d038e6f756903acd22e3/FormalConjectures/Wikipedia/HardyLittlewood.lean#L100

So this should probably be solved by importing the definition/file from the `Wikipedia` subdir, to avoid code duplications, see explanation here: https://github.com/google-deepmind/formal-conjectures/pull/1627#issuecomment-3737423822

violet-hurricane:
Thanks for the suggestion! I’ve updated Erdős 855 to import the Second Hardy–Littlewood conjecture from the Wikipedia directory and removed the duplicated statement.

FormalConjectures/ErdosProblems/855.lean
```diff
@@ -0,0 +1,42 @@
...
+
+/--
+If `π(x)` counts the number of primes in `[1, x]`, is it true that for sufficiently large `x` and `y`
+we have `π(x+y) ≤ π(x) + π(y)`?
+-/
+@[category research open, AMS 11]
+theorem erdos_855 :
+    answer(sorry) ↔
+      ∃ N : ℕ, ∀ x y : ℕ, x ≥ N → y ≥ N →
+        SecondHardyLittlewoodConjectureFor x y := by
```
YaelDillies: No need to obscure the statement with intermediate definitions
```suggestion
      ∀ᶠ x in atTop, ∀ᶠ y in atTop, π (x + y) ≤ π x + π y := by
```
You could also change `Wikipedia.HardyLittlewood` similarly

violet-hurricane: Thanks! I’ve updated Erdős 855 to use the eventual (∀ᶠ) formulation directly and removed the intermediate definition.

FormalConjectures/ErdosProblems/855.lean
```diff
@@ -0,0 +1,38 @@
...
+open Filter
+open scoped Nat.Prime
+
+namespace Erdos855
+
+@[category research open, AMS 11]
+theorem erdos_855 :
+    answer(sorry) ↔
+        (∀ᶠ x in atTop, ∀ᶠ y in atTop,
+        π (x + y) ≤ π x + π y) := by
```
YaelDillies:
```suggestion
theorem erdos_855 : answer(sorry) ↔
    ∀ᶠ x in atTop, ∀ᶠ y in atTop, π (x + y) ≤ π x + π y := by
```

YaelDillies (approved):
Thanks!

FormalConjectures/ErdosProblems/855.lean
violet-hurricane: Done. I removed the extra parentheses and wrote it as a single binder chain.

PR #1626, opened 2026-01-12
feat(ErdosProblems): 1067

danielchin:
This is similar to Erdos 1068 (https://github.com/google-deepmind/formal-conjectures/pull/1568). I'm reusing the definitions of `InternallyDisjoint` and `InfinitelyConnected`. These should now probably go into some shared location. Let me know where I should move these since I haven't really interacted with the shared library files yet.

### DIFF
--- a/FormalConjectures/ErdosProblems/1067.lean
+++ b/FormalConjectures/ErdosProblems/1067.lean
@@ -41,7 +41,7 @@ infinitely many edges. In other words, removing any finite set of edges leaves
 the graph connected.
 -/
 def InfinitelyEdgeConnected {V : Type*} (G : SimpleGraph V) : Prop :=
-  ∀ {s : Set (Sym2 V)}, s.Finite → (G.deleteEdges s).Connected
+  ∀ ⦃s : Set (Sym2 V)⦄, s.Finite → (G.deleteEdges s).Connected
 
 /--
 Does every graph with chromatic number $\aleph_1$ contain an infinitely connected subgraph with
--- a/FormalConjectures/ErdosProblems/1067.lean
+++ b/FormalConjectures/ErdosProblems/1067.lean
@@ -1,0 +1,77 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 1067
+
+*References:*
+- [erdosproblems.com/1067](https://www.erdosproblems.com/1067)
+- [BoPi24] N. Bowler and M. Pitz, A note on uncountably chromatic graphs. arXiv:2402.05984 (2024).
+- [ErHa66] Erdős, P. and Hajnal, A., On chromatic number of graphs and set-systems. Acta Math. Acad.
+  Sci. Hungar. (1966), 61-99.
+- [Ko13] Komjáth, Péter, A note on chromatic number and connectivity of infinite graphs. Israel
+  J. Math. (2013), 499--506.
+- [So15] Soukup, Dániel T., Trees, ladders and graphs. J. Combin. Theory Ser. B (2015), 96--116.
+- [Th17] Thomassen, Carsten, Infinitely connected subgraphs in graphs of uncountable chromatic
+  number. Combinatorica (2017), 785--793.
+-/
+
+open Cardinal List
+
+namespace Erdos1067
+
+/--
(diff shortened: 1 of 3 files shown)


CoolRmal:
You can put it somewhere in [FormalConjecturesForMathlib](https://github.com/google-deepmind/formal-conjectures/tree/main/FormalConjecturesForMathlib)/[Combinatorics](https://github.com/google-deepmind/formal-conjectures/tree/main/FormalConjecturesForMathlib/Combinatorics)/[SimpleGraph](https://github.com/google-deepmind/formal-conjectures/tree/main/FormalConjecturesForMathlib/Combinatorics/SimpleGraph)

FormalConjectures/ErdosProblems/1067.lean
```diff
@@ -0,0 +1,65 @@
...
+Komjáth [Ko13] proved that it is consistent that the answer is no. This was improved by
+Soukup [So15], who constructed a counterexample using no extra set-theoretical assumptions. A
+simpler elementary example was given by Bowler and Pitz [BoPi24].
+
+Thomassen [Th17] constructed a counterexample to the version which asks for infinite
+edge-connectivity (that is, to disconnect the graph requires deleting infinitely many edges).
+-/
+@[category research solved, AMS 5]
+theorem erdos_1067 :
+    False ↔ ∀ (V : Type) (G : SimpleGraph V), G.chromaticNumber = aleph 1 →
```
YaelDillies:
```suggestion
    answer(False) ↔ ∀ (V : Type) (G : SimpleGraph V), G.chromaticNumber = aleph 1 →
```

FormalConjectures/ErdosProblems/1067.lean
```diff
@@ -0,0 +1,65 @@
...
+
+/--
+In [ErHa66] Erdős and Hajnal asked the same question under the additional assumption that the graph
+has $\aleph_1$ many vertices. Komjáth [Ko13] proved that the answer is independent of ZFC.
+-/
+@[category research solved, AMS 5]
+theorem erdos_1067.variant.aleph_1_vertices :
+    False ↔ ∀ (V : Type) (G : SimpleGraph V), #V = aleph 1 → G.chromaticNumber = aleph 1 →
+      ∃ (H : G.Subgraph), H.coe.chromaticNumber = aleph 1 ∧ InfinitelyConnected H.coe := by
+  sorry
```
YaelDillies, line 68: "independent of ZFC" doesn't mean it's false! If you really want to state this, you will need to use model theory

FormalConjectures/ErdosProblems/1067.lean
```diff
@@ -0,0 +1,65 @@
...
+/--
+Does every graph with chromatic number $\aleph_1$ contain an infinitely connected subgraph with
+chromatic number $\aleph_1$?
+
+Komjáth [Ko13] proved that it is consistent that the answer is no. This was improved by
+Soukup [So15], who constructed a counterexample using no extra set-theoretical assumptions. A
+simpler elementary example was given by Bowler and Pitz [BoPi24].
+
+Thomassen [Th17] constructed a counterexample to the version which asks for infinite
+edge-connectivity (that is, to disconnect the graph requires deleting infinitely many edges).
```
YaelDillies, line 62: Can you state this separately?

danielchin: Stated this separately and added a def of infinite edge connectivity.

FormalConjectures/ErdosProblems/1067.lean
danielchin: Done.

FormalConjectures/ErdosProblems/1067.lean
danielchin: I just saw the comment on the Erdos thread and understand that this variant is independent but the rest of the problem is disproved. 

Though, I'm not entirely clear how to state this using model theory. Do you have any pointers? Otherwise, I could just leave a TODO here.

YaelDillies: In all honesty. stating this properly is quite difficult as you would need to translate everything this statement uses into first-order formulas in the language of graphs. I would just leave a TODO for now

danielchin: Replaced with a TODO

FormalConjectures/ErdosProblems/1067.lean
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Erdos1067
+
+/--
+A graph is infinitely edge-connected if to disconnect the graph requires deleting
+infinitely many edges. In other words, removing any finite set of edges leaves
+the graph connected.
+-/
+def InfinitelyEdgeConnected {V : Type*} (G : SimpleGraph V) : Prop :=
+  ∀ (s : Set (Sym2 V)), s.Finite → s ⊆ G.edgeSet → (G.deleteEdges s).Connected
```
YaelDillies: Just simplifying the definition slightly:
```suggestion
  ∀ ⦃s : Set (Sym2 V)⦄, s.Finite → (G.deleteEdges s).Connected
```

YaelDillies: Wrong brackets, read again

danielchin: Done.

danielchin: Oops, sorry about that! Fixed now.

YaelDillies (approved):
Thanks!

PR #1627, opened 2026-01-12
feat(ErdosProblems): Collatz conjecture

### LINKED ISSUE #1625: Collatz formalization doesn't show up on Erdos 1135
It is formalized at https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/Wikipedia/CollatzConjecture.lean

but on the Erdos Problems site:
https://www.erdosproblems.com/forum/thread/1135

it says
> Formalised statement? No ([Create a formalisation here](https://github.com/google-deepmind/formal-conjectures))

which links to the FormalConjectures repository.

What would be the best way to get this linked?

cyb3r17:
Import the Collatz conjecture for Erdős Problem 1135. This was done so that it appears as formalised on the erdos website, although it's the same as the Collatz Conjecture under the Wikipedia directory, due to the way the repo is setup, this had to be done.

Closes #1625

Note: I am aware this isn't a problem proposed by Erdos, however he did offer a prize for it and is also mentioned on the website as the 1135th problem.

### DIFF
--- a/FormalConjectures/ErdosProblems/1135.lean
+++ b/FormalConjectures/ErdosProblems/1135.lean
@@ -18,7 +18,7 @@ import FormalConjectures.Wikipedia.CollatzConjecture
 /-!
 # Erdős Problem 1135
 
-*Reference:*
+*References:*
 * [erdosproblems.com/1135](__https://www.erdosproblems.com/1135__)
 * [Gu04] Guy, Richard K., Unsolved problems in number theory.  (2004), xviii+437.
 * [La10] Lagarias, Jeffrey C., The {$3x+1$} problem: an overview.  (2010), 3--29.


mo271:
We probably want to avoid duplicating definitions and conjectures here. In a few other problems (Erdős 508 (Hadwiger Nelson), Erdős 424 (Ben Greens Open Problem 63), Erdős 107 (Happy Ending Problem), this is solved by just importing the file from `ErdosProblems` into a file in `Wikipedia`. This is how it could potentially be done here as well -- or importing the other way around, keeping `Wikipedia/CollatzConjecture` as is and then make a new file for Erdős 1135 and then import the `FormalConjecutures.Wikipedia.CollatzConjecture`and use that?

cyb3r17:
Alright, will implement asap. Thanks for the suggestion!

FormalConjectures/ErdosProblems/1135.lean
```diff
@@ -0,0 +1,36 @@
...
+* [La16] Lagarias, Jeffrey C., Erd\H os, {K}larner, and the {$3x+1$} problem. Amer. Math. Monthly (2016), 753--776.
+* [La85] Lagarias, Jeffrey C., The {$3x+1$} problem and its generalizations. Amer. Math. Monthly (1985), 3--23.
+-/
+
+namespace Erdos1135
+
+/-!The **Collatz conjecture** states that for any positive integer $n$, there exists a natural number $m$ such that the $m$-th term of the sequence is 1.
+-/
+@[category research open, AMS 11 37]
+alias collatzConjecture := CollatzConjecture.collatzConjecture
```
mo271:
```suggestion
alias erdos_1135 := CollatzConjecture.collatzConjecture
```

I'm not entirely sure if using `alias` here is a good idea, but if we do let's use the name we usually use for theorems.

FormalConjectures/ErdosProblems/1135.lean
```diff
@@ -0,0 +1,36 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+import FormalConjectures.Wikipedia.CollatzConjecture
+
+/-!
+# Erdős Problem 1135
+
+*Reference:*
```
mo271:
```suggestion
*References:*
```

FormalConjectures/ErdosProblems/1135.lean
cyb3r17: Will keep this in mind henceforth

cyb3r17:
Anything else left to be reviewed/changed?

PR #1635, opened 2026-01-12
feat(ErdosProblems): 318

### LINKED ISSUE #484: Erdős Problem 318
### What is the conjecture
    
https://www.erdosproblems.com/318

  Let $A\subseteq \mathbb{N}$ be an infinite arithmetic progression and $f:A\to \\{-1,1\\}$ be a non-constant function. Must there exist a finite non-empty $S\subset A$ such that $\sum_{n\in S}\frac{f(n)}{n}=0$?
What about if $A$ is an arbitrary set of positive density? What if $A$ is the set of squares excluding $1$?

Status: open

### Choose either option

CoolRmal:
Closes #484

A note about the TODO before `erdos_318.posDensity`. An obvious choice is to use the set of odd numbers union with {2} to prove this theorem. We need some lemmas:
1. The set of odd numbers has natural density equal to 1/2. More generally, an infinite AP has natural density equal to the gap.
2. Natural density is invariant under union with finite sets.

Hence, we need more APIs for `HasPosDensity` before proving this theorem.

### DIFF
--- a/FormalConjectures/ErdosProblems/318.lean
+++ b/FormalConjectures/ErdosProblems/318.lean
@@ -52,7 +52,7 @@ theorem erdos_318.univ : P₁ univ := by
 theorem erdos_318.odd : P₁ {n | Odd n} := by
   sorry
 
-/-- The set of nonzero squares does not have property `P₁`. -/
+/-- The set of squares does not have property `P₁`. -/
 @[category test, AMS 11]
 theorem erdos_318.squares : ¬ P₁ ({n | IsSquare n}) := by
   simp only [P₁, not_forall, not_exists, not_and]
@@ -116,7 +116,7 @@ theorem erdos_318.posDensity : ∃ A : Set ℕ, HasPosDensity A ∧ ¬ P₁ A :=
 theorem erdos_318.infinite_AP {A : Set ℕ} (hA : A.IsAPOfLength ⊤) : P₁ A := by
   sorry
 
-/-- Does the set of nonzero squares excluding 1 have property `P₁`? -/
+/-- Does the set of squares excluding 1 have property `P₁`? -/
 @[category research open, AMS 11]
 theorem erdos_318.square_excluding_one : answer(sorry) ↔  P₁ ({n | IsSquare n} \ {1}) := by
   sorry


FormalConjectures/ErdosProblems/318.lean
```diff
@@ -0,0 +1,122 @@
...
+    fun x ⟨y, hy⟩ => ?_, fun S h hs => ?_⟩
+  · have : (- 1 : ℝ) = 1 := by simpa using congr_fun h ⟨4, ⟨⟨2, by grind⟩, by grind⟩⟩
+    grind
+  · have : 1 = (- 1 : ℝ) := by simpa using congr_fun h ⟨1, ⟨IsSquare.one, by grind⟩⟩
+    grind
+  · by_cases x = 1 <;> grind
+  -- Consider two cases: `1 ∈ S` or `1 ∉ S`. In the first case, the finite sum over `S` is bounded
+  -- below by `1 - (π ^ 2 / 6 - 1)`, which is positive. In the second case, the finite sum over `S`
+  -- is negative.
+  · by_cases h1 : 1 ∈ S
```
YaelDillies: Can you unindent? This is the main case

FormalConjectures/ErdosProblems/318.lean
```diff
@@ -0,0 +1,122 @@
...
+  sorry
+
+/-- Every infinite arithmetic progression has property `P₁`. This is proved in [Sa82b]. -/
+@[category research solved, AMS 11]
+theorem erdos_318.infinite_AP {A : Set ℕ} (hA : A.IsAPOfLength ⊤) : P₁ A := by
+  sorry
+
+/-- Does the set of nonzero squares excluding 1 have property `P₁`? -/
+@[category research open, AMS 11]
+theorem erdos_318.square_excluding_one : answer(sorry) ↔  P₁ ({n | IsSquare n} \ {1}) := by
```
YaelDillies, line 121:
```suggestion
theorem erdos_318.square_excluding_one : answer(sorry) ↔  P₁ ({n | IsSquare n} \ {0, 1}) := by
```
although you could also prove directly that `P₁` doesn't care about zero

CoolRmal: Actually with my current definition any set `A` that contains `0` has the property `P₁` because `1/0 = 0`. I believe sometimes when they claim that `A` is a subset of `ℕ`, they actually mean that `A` is a subset of `ℕ+`. However, I think using PNat in Lean is a bit troublesome, so I eventually decide to use `A \ {0}` in my definition. May you take a look of the new definition?

YaelDillies: Yes, that looks good, thanks!

PR #1644, opened 2026-01-13
feat(ErdosProblems/386): more variants

mo271:


mo271:
@BIGfoot496, please have a look if this makes sense. I do think the intended statement might be the "exists" version, but haven't read all the comments nor the original sources.

BIGfoot496:
The commenters seem to assume that the question is whether this can happen infinitely often *for fixed k*, so I’d say this problem should really be phrased as “Figure out _which_ k work and which don’t”. The quote from Erdős and Graham (a proof that this cannot happen infinitely often for binomial(n,2) seems hopeless; probably this can never happen for binomial(n,k) if 3≤k≤n−3.) seems to imply they thought k=2 is infinite and k>2 is not

BIGfoot496:
But I haven’t been able to find the original paper by Erdős and Graham, so I’m not 100% sure I understand the problem correctly either

mo271:
> The commenters seem to assume that the question is whether this can happen infinitely often _for fixed k_, so I’d say this problem should really be phrased as “Figure out _which_ k work and which don’t”. The quote from Erdős and Graham (a proof that this cannot happen infinitely often for binomial(n,2) seems hopeless; probably this can never happen for binomial(n,k) if 3≤k≤n−3.) seems to imply they thought k=2 is infinite and k>2 is not

I suppose we could use a `answer(sorry)` formulation requiring the user to fill in the set of number where the statement is true?

mo271:
this is already better what was there before: we can add the `answer(sorry)` variant in a follow up

PR #1645, opened 2026-01-13
feat(OEIS): add A006697 - subword complexity of a → aab, b → b

### LINKED ISSUE #1475: math.CO/0409509 number 117
### What is the conjecture
Let $a_n$ the number of subwords of length $n$ in the word generated by a $\mapsto$ aab, b $\mapsto$ b. Then

$$
\begin{equation}\sum_{n\ge0}a_nx^n = 1+\frac{1}{1-x}+\frac{1}{(1-x)^2}(\frac{1}{1-x}-\sum_{k\ge1}x^{2^k+k-1}).\end{equation}
$$

See https://oeis.org/A006697, where there are also other forms conjectured. Are they equivalent to each other (more sub-hypotheses)?
 
### Prerequisites needed

Status: open

### Choose either option

jdhruv555:
## What is the conjecture

Let $a_n$ be the number of distinct subwords (contiguous factors) of length $n$ in the infinite word generated by the morphism $\sigma: a \mapsto aab, b \mapsto b$, starting from $a$.

The conjectured generating function is:
$$\sum_{n \geq 0} a_n x^n = 1 + \frac{1}{1-x} + \frac{1}{(1-x)^2}\left(\frac{1}{1-x} - \sum_{k \geq 1} x^{2^k + k - 1}\right).$$

## Formalization

- Defines the morphism σ: a → aab, b → b using `Bool`
- Constructs the infinite fixed-point word starting from 'a'
- Defines subword complexity a(n) as the cardinality of distinct factors of length n
- States the conjecture equating a(n) with coefficients of the generating function

## References

- [OEIS A006697](https://oeis.org/A006697)
- J.-P. Allouche and J. Shallit, "On the subword complexity of the fixed point of a → aab, b → b, and generalizations," arXiv:1605.02361 [math.CO], 2016.

Closes #1475

Let me know if there's any suggestion.


jdhruv555:
Lean 4 , Mathlib uses different method names than standard Lean (flatten instead of join or bind),hopefully it will pass the build check here this time , i have checked locally it's passing the case . 
If there is any better approach ,please let me know I will try to implement that .

jdhruv555:
@YaelDillies Hi, can you please review it once .

FormalConjectures/OEIS/006697.lean
```diff
@@ -0,0 +1,93 @@
...
+
+/-- The morphism σ on {a, b} defined by a ↦ aab, b ↦ b, represented on Bool where
+`false` = a and `true` = b. -/
+def morphism : Bool → List Bool
+  | false => [false, false, true]  -- a ↦ aab
+  | true => [true]                  -- b ↦ b
+
+/-- Apply the morphism to a list of symbols. -/
+def applyMorphism (w : List Bool) : List Bool :=
+  (w.map morphism).flatten
```
YaelDillies: This already exists as `List.flatMap`

FormalConjectures/OEIS/006697.lean
```diff
@@ -0,0 +1,93 @@
...
+  \sum_{k \geq 1} x^{2^k + k - 1}\right).$$
+
+Equivalently, a(n) equals the n-th coefficient of this generating function.
+-/
+@[category research open, AMS 68]
+theorem conjecture : ∀ n, (a n : ℚ) =
+    PowerSeries.coeff ℚ n (1 + (1 - PowerSeries.X)⁻¹ +
+      (1 - PowerSeries.X)⁻¹ ^ 2 * ((1 - PowerSeries.X)⁻¹ -
+        ∑ k ∈ (Finset.range (n + 1)).filter (isExceptionalExponent ·),
+          PowerSeries.X ^ k)) := by
```
YaelDillies: Why not
```suggestion
        ∑' k, .X ^ (2 ^ (k + 1) + k)) := by
```

YaelDillies: Maybe `open PowerSeries`

FormalConjectures/OEIS/006697.lean
```diff
@@ -0,0 +1,93 @@
...
+/-- The n-th iterate of the morphism applied to [a]. -/
+def iterateMorphism : ℕ → List Bool
+  | 0 => [false]
+  | n + 1 => applyMorphism (iterateMorphism n)
+
+/-- The infinite word w is the fixed point of the morphism starting from 'a'.
+We define it as the limit: w(i) is the i-th symbol, which stabilizes after
+sufficiently many iterations. -/
+noncomputable def infiniteWord (i : ℕ) : Bool :=
+  (iterateMorphism (i + 1)).getD i false
```
YaelDillies: Can you prove this `getD` is in bound? ie
```suggestion
  (iterateMorphism (i + 1))[i]'your_proof_here
```

jdhruv555:
@YaelDillies I have tried to implement all of your suggestions but when  I attempted to prove the bound `i < (iterateMorphism (i + 1)).length`, but the proof requires tracking the count of false/true values through flatMap iterations to establish L(n) = 2^(n+1) - 1. This was causing build failures.

Since the bound holds mathematically and `getD` with default `false` doesn't affect correctness (the index is always in bounds for the actual infinite word), I kept `getD` for simplicity.

Happy to revisit if you have a simpler approach in mind!

YaelDillies (approved):
Thanks!

PR #1648, opened 2026-01-14
feat(ErdosProblems): 770

### LINKED ISSUE #933: Erdős Problem 770
### What is the conjecture

https://www.erdosproblems.com/770

  Let $h(n)$ be minimal such that $2^n-1,3^n-1,\ldots,h(n)^n-1$ are mutually coprime. 

Does, for every prime $p$, the density $\delta_p$ of integers with $h(n)=p$ exist? Does $\liminf h(n)=\infty$? Is it true that if $p$ is the greatest prime such that $p-1\mid n$ and $p>n^\epsilon$ then $h(n)=p$?

Status:  open

### Choose either option

CoolRmal:
I'll prove TODOs in another PR.

Closes #933

### DIFF
--- a/FormalConjectures/ErdosProblems/770.lean
+++ b/FormalConjectures/ErdosProblems/770.lean
@@ -64,7 +64,7 @@ theorem erdos_770.prime : answer(sorry) ↔ ∀ ε > 0, ∀ᶠ n in atTop,
 
 /-- It is probably true that `h n = 3` for infinitely many `n`. -/
 @[category research open, AMS 11]
-theorem erdos_770.three : answer(sorry) ↔ {n | h n = 3}.Infinite := by
+theorem erdos_770.three : {n | h n = 3}.Infinite := by
   sorry
 
 end Erdos770


FormalConjectures/ErdosProblems/770.lean
```diff
@@ -0,0 +1,70 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 770
+
+*Reference:*
```
mo271:
```suggestion
*References:*
```

FormalConjectures/ErdosProblems/770.lean
```diff
@@ -0,0 +1,70 @@
...
+namespace Erdos770
+
+/-- Let `h n` be the minimal number such that `2 ^ n - 1, ..., (h n) ^ n - 1` are mutually
+coprime. -/
+noncomputable def h (n : ℕ) : ℕ∞ := sInf {m | 2 < m ∧ ∀ (i j : ℕ), (i : ℕ∞) ∈ Set.Icc 2 m →
+  (i : ℕ∞) ∈ Set.Icc 2 m → i ≠ j → (i ^ n - 1).Coprime (j ^ n - 1)}
+
+/-- `n` is prime iff `h n = n + 1`. #TODO: prove this theorem. -/
+@[category test, AMS 11]
+theorem Nat.Prime.h_eq_add_one {n : ℕ} : h n = n + 1 ↔ (n + 1).Prime := by sorry
```
YaelDillies: Can you turn one of the informal or formal statement around so that they match?

FormalConjectures/ErdosProblems/770.lean
```diff
@@ -0,0 +1,70 @@
...
+/-- Does `liminf h n = ∞`? -/
+@[category research open, AMS 11]
+theorem erdos_770.liminf : answer(sorry) ↔ liminf h atTop = ⊤ := by
+  sorry
+
+/-- Is it true that if `p` is the greatest prime such that `p - 1 ∣ n` and `p > n ^ ε`, then
+`h n = p`? -/
+@[category research open, AMS 11]
+theorem erdos_770.prime : answer(sorry) ↔ ∀ ε > 0, ∀ n : ℕ,
+    let p := sSup {m : ℕ | m.Prime ∧ (m - 1) ∣ n}
```
YaelDillies:
```suggestion
    let p := sSup {m : ℕ | m.Prime ∧ m - 1 ∣ n}
```

FormalConjectures/ErdosProblems/770.lean
```diff
@@ -0,0 +1,70 @@
...
+-/
+
+open Set ENat Filter
+
+namespace Erdos770
+
+/-- Let `h n` be the minimal number such that `2 ^ n - 1, ..., (h n) ^ n - 1` are mutually
+coprime. -/
+noncomputable def h (n : ℕ) : ℕ∞ := sInf {m | 2 < m ∧ ∀ (i j : ℕ), (i : ℕ∞) ∈ Set.Icc 2 m →
+  (i : ℕ∞) ∈ Set.Icc 2 m → i ≠ j → (i ^ n - 1).Coprime (j ^ n - 1)}
```
mo271:
```suggestion
noncomputable def h (n : ℕ) : ℕ∞ := sInf {m | 2 < m ∧ ∀ (i j : ℕ), (i : ℕ∞) ∈ Set.Icc 2 m →
  (j : ℕ∞) ∈ Set.Icc 2 m → i ≠ j → (i ^ n - 1).Coprime (j ^ n - 1)}
```

Also potentially could be reformulated with `Pairwise`. I would go with the equivalent 
```
noncomputable def h (n : ℕ) : ℕ∞ := sInf {m | 2 < m ∧
  (Set.Icc 2 m).Pairwise fun i j => i < ⊤ → j < ⊤ → (i.toNat ^ n - 1).Coprime (j.toNat ^ n - 1) }
```

mo271:
```lean
noncomputable def h (n : ℕ) : ℕ∞ := sInf {m | 2 < m ∧
  (Set.Icc 2 m).Pairwise fun i j => (i.toNat ^ n - 1).Coprime (j.toNat ^ n - 1) }
```
should also be equivalent

CoolRmal: Just updated my PR with your definition.

FormalConjectures/ErdosProblems/770.lean
```diff
@@ -0,0 +1,70 @@
...
+`h n = p`? -/
+@[category research open, AMS 11]
+theorem erdos_770.prime : answer(sorry) ↔ ∀ ε > 0, ∀ n : ℕ,
+    let p := sSup {m : ℕ | m.Prime ∧ m - 1 ∣ n}
+    p > (n : ℝ) ^ (ε : ℝ) → h n = p := by
+  sorry
+
+/-- It is probably true that `h n = 3` for infinitely many `n`. -/
+@[category research open, AMS 11]
+theorem erdos_770.three : answer(sorry) ↔ {n | h n = 3}.Infinite := by
```
YaelDillies:
```suggestion
theorem erdos_770.three : {n | h n = 3}.Infinite := by
```
since this isn't a question

FormalConjectures/ErdosProblems/770.lean
```diff
@@ -0,0 +1,70 @@
...
+
+/-- `n + 1` is prime iff `h n = n + 1`. #TODO: prove this theorem. -/
+@[category test, AMS 11]
+theorem Nat.Prime.h_eq_add_one {n : ℕ} : h n = n + 1 ↔ (n + 1).Prime := by sorry
+
+/-- If `n` is odd, then `h n = ∞`. #TODO: prove this theorem. -/
+@[category test, AMS 11]
+theorem Odd.h_unbounded {n : ℕ} (pn : Odd n) : h n = ⊤ := by sorry
+
+/-- For every prime `p`, does the density of integers with `h n = p` exists? -/
```
YaelDillies:
```suggestion
/-- For every prime `p`, does the density of integers with `h n = p` exist? -/
```

FormalConjectures/ErdosProblems/770.lean
```diff
@@ -0,0 +1,70 @@
...
+@[category research open, AMS 11]
+theorem erdos_770.liminf : answer(sorry) ↔ liminf h atTop = ⊤ := by
+  sorry
+
+/-- Is it true that if `p` is the greatest prime such that `p - 1 ∣ n` and `p > n ^ ε`, then
+`h n = p`? -/
+@[category research open, AMS 11]
+theorem erdos_770.prime : answer(sorry) ↔ ∀ ε > 0, ∀ n : ℕ,
+    let p := sSup {m : ℕ | m.Prime ∧ m - 1 ∣ n}
+    p > (n : ℝ) ^ (ε : ℝ) → h n = p := by
```
YaelDillies: That can't be correct. This is equivalent to
```suggestion
theorem erdos_770.prime : answer(sorry) ↔ ∀ n : ℕ,
    let p := sSup {m : ℕ | m.Prime ∧ m - 1 ∣ n}
    ∀ ε > 0, p > (n : ℝ) ^ (ε : ℝ) → h n = p := by
```
which is equivalent to
```suggestion
theorem erdos_770.prime : answer(sorry) ↔ ∀ n : ℕ,
    let p := sSup {m : ℕ | m.Prime ∧ m - 1 ∣ n}
    h n = p := by
```

FormalConjectures/ErdosProblems/770.lean
CoolRmal: I am not sure whether we want this change. It is still an open problem and we don't know whether Erdos' guess is correct.

FormalConjectures/ErdosProblems/770.lean
CoolRmal: You are correct, but I am also not sure what is Erdos's intention in his article (https://www.renyi.hu/~p_erdos/1974-27.pdf). He probably means that `∀ ε > 0, ∀ᶠ n in atTop,...`

Screenshot 2026-01-18 at 2 44 38 PM

YaelDillies: That seems very likely, yes

FormalConjectures/ErdosProblems/770.lean
YaelDillies: AI tools are usually able to revert a statement, negate it and then disprove it. The presence or absence of `answer(sorry) ↔` is merely communicating what we think the answer is

CoolRmal: I see, just commited your suggestions

PR #1652, opened 2026-01-14
Update 142.lean

SNPSHADY-92:
Hello, I found the asymptotic formula for EDMOS problem 124: Let rk(N) be the maximum possible size of a subset of {1, ..., N} containing no nontrivial elements. Prove an asymptotic formula for rk(N). Could someone please check it with me?


Paul-Lez:
@SNPSHADY-92 I think the best place to share this would be the comment section in the corresponding erdosproblems.com page! formal-conjectures aims to collect Lean statements (and occasionally short proofs).

Paul-Lez:
Closing since the PR is outside of the scope of the repository.

PR #1671, opened 2026-01-15
chore: avoid zero padding in filenames

mo271:
we don't do it in `ErdosProblems` nor `GreensOpenProblems`, also there are already non-zero padded files in `OEIS`


YaelDillies:
Actually, I would have done the opposite: OEIS entries are zero-padded, and Lean orders 10 before 2 when processing filenames.

mo271:
> Actually, I would have done the opposite: OEIS entries are zero-padded, and Lean orders 10 before 2 when processing filenames.

I suppose there is at least some support for non-zero padded entries: https://oeis.org/A1 just works.
It doesn't seem out of the question that the OEIS grows to 3 times its current size, and then one would need to add an additional zero. 

When we started the Erdős Problem would have needed one less zero if we had ever zero padded them. 
The problem of ordering is something that we should anyhow anyhow for Erdős problems, see #3.

I think these zero-padding approaches feel like workarounds from the 90ies, and instead we should fix the doc-gen4, or whatever else to handle them better. 
Another instance of mathoverflow-question numbers. 

We do agree that it should be done consistently within the `OEIS` dir, I suppose, right?

YaelDillies:
What does @eric-wieser think here?

mo271:
I chatted with @eric-wieser, he pointed out: 
"Does vscode support sane sorting of numeric files?
If not, I'd be inclined to continue padding"

I checked, vscode handles the sorting just fine. 
What do you mean by "Lean orders 10 before 2", @YaelDillies, doc-gen4?

YaelDillies:
I mean things like the order in which `#min_imports` sorts files, or the output of `lake exe mk_all`, but maybe these are the ones that should be fixed.

mo271:
yeah, the sorting for `#min_imports` used to be even much worse: https://github.com/leanprover-community/import-graph/pull/18

PR #1699, opened 2026-01-16
feat(ErdosProblems): 195

### LINKED ISSUE #416: Erdős Problem 195
### What is the conjecture
    
https://www.erdosproblems.com/195

  What is the largest $k$ such that in any permutation of $\mathbb{Z}$ there must exist a monotone $k$-term arithmetic progression $x_1<\cdots<x_k$?

Status: open

### Prerequisites needed
None, $k$-term arithmetic progressions are in `ForMathlib`.

### Choose either option

danielchin:
Closes #416 

I kept the formalization of the weaker bound of \leq 5 because I assume that we eventually want both the proof for \leq 4 and \leq 5 formalized as future training data, but let me know if it's not necessary and we should remove it.

### DIFF
--- a/FormalConjectures/ErdosProblems/195.lean
+++ b/FormalConjectures/ErdosProblems/195.lean
@@ -15,7 +15,6 @@ limitations under the License.
 -/
 
 import FormalConjectures.Util.ProblemImports
-import FormalConjecturesForMathlib.Combinatorics.AP.Basic
 
 /-!
 # Erdős Problem 195
--- a/FormalConjectures/ErdosProblems/195.lean
+++ b/FormalConjectures/ErdosProblems/195.lean
@@ -1,0 +1,61 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 195
+
+*References:*
+- [erdosproblems.com/195](https://www.erdosproblems.com/195)
+- [Ad22] Adenwalla, S., Avoiding Monotone Arithmetic Progressions in Permutations of Integers.
+  arXiv:2211.04451 (2022).
+- [Ge19] Geneson, Jesse, Forbidden arithmetic progressions in permutations of subsets of the
+  integers. Discrete Math. (2019), 1489-1491.
+-/
+
+namespace Erdos195
+
+open Function Set
+
+/-- A permutation `f` of `ℤ` has a monotone $k$-term AP if there is an AP `l`
+such that its image under `f` is sorted. -/
+def HasMonotoneAP (f : ℤ ≃ ℤ) (k : ℕ) : Prop :=
+  ∃ l : List ℤ, l.IsAPOfLength k ∧ ((l.map f).Sorted (· < ·) ∨ (l.map f).Sorted (· > ·))
+
+/--
+What is the largest $k$ such that in any permutation of $\mathbb{Z}$ there must exist a
+monotone $k$-term arithmetic progression $x_1 < \cdots < x_k$?
+-/
+@[category research open, AMS 5]
+theorem erdos_195 :
+    answer(sorry) = sSup {k : ℕ | ∀ f : ℤ ≃ ℤ, HasMonotoneAP f k} := by sorry
+
+/--
+Geneson [Ge19] proved that k ≤ 5.
+-/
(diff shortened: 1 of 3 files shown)


FormalConjectures/ErdosProblems/195.lean
```diff
@@ -0,0 +1,61 @@
...
+-/
+
+namespace Erdos195
+
+open Function Set
+
+/-- A permutation `f` of `ℤ` has a monotone $k$-term AP if there is an AP `l`
+such that its image under `f` is sorted. -/
+def HasMonotoneAP (f : ℤ ≃ ℤ) (k : ℕ) : Prop :=
+  ∃ l : List ℤ, l.IsAPOfLength k ∧ ((l.map f).Sorted (· < ·) ∨ (l.map f).Sorted (· > ·))
```
mo271: in https://github.com/google-deepmind/formal-conjectures/blob/f2182875a210d9616ab92a78a43fee4e12470a8a/FormalConjectures/ErdosProblems/196.lean#L29 this is just inlined.

Perhaps it would make sense to move this to `ForMathlib` and use it in both places?

FormalConjectures/ErdosProblems/195.lean
```diff
@@ -0,0 +1,61 @@
...
+
+/--
+What is the largest $k$ such that in any permutation of $\mathbb{Z}$ there must exist a
+monotone $k$-term arithmetic progression $x_1 < \cdots < x_k$?
+-/
+@[category research open, AMS 5]
+theorem erdos_195 :
+    answer(sorry) = sSup {k : ℕ | ∀ f : ℤ ≃ ℤ, HasMonotoneAP f k} := by sorry
```
mo271:
```suggestion
    answer(sorry) = sSup {k : ℕ | ∀ f : ℤ ≃ ℤ, HasMonotoneAP f k} := by 
  sorry
```

With a newline is the usual style. Also below

mo271 (commented):
Nice, thank you! Perhaps we could also add a variant `6 >= ...`, or `answer(sorry) \iff 6 \ge ...`?

FormalConjectures/ErdosProblems/195.lean
danielchin: Done.

FormalConjectures/ErdosProblems/195.lean
danielchin: Added to `Basic.lean` within the `Combinatorics` directory.

danielchin:
> Nice, thank you! Perhaps we could also add a variant `6 >= ...`, or `answer(sorry) \iff 6 \ge ...`?

I'm not entirely sure if it makes sense to add this? [Ad22] restricted this down to length 4 from length 5. Adding a variant of `6 \geq ...` is trivial given [Ad22] and [Ge19].

I do think the way the question is phrased on ErdosProblems is a bit confusing. It becomes a bit more clear after reading the abstract from [[Ad22]](https://arxiv.org/abs/2211.04451).

FormalConjectures/ErdosProblems/196.lean
```diff
@@ -15,6 +15,7 @@ limitations under the License.
 -/
 
 import FormalConjectures.Util.ProblemImports
+import FormalConjecturesForMathlib.Combinatorics.AP.Basic
```
YaelDillies:
```suggestion
```

FormalConjecturesForMathlib/Combinatorics/AP/Basic.lean
```diff
@@ -233,3 +233,11 @@ def ContainsMonoAPofLength {κ : Type} [Finite κ] {M : Set α}
...
   ∃ c : κ, ∃ ap : Set M, ((·.1) '' ap).IsAPOfLength k ∧
     ∀ m ∈ ap, coloring m = c
+
+/--
+A function `f : α → β` has a monotone `k`-term arithmetic progression if there exists an
+arithmetic progression `l` of length `k` in `α` such that its image under `f` is sorted
+(either strictly increasing or strictly decreasing).
+-/
+def HasMonotoneAP {β : Type*} [Preorder β] (f : α → β) (k : ℕ) : Prop :=
+  ∃ l : List α, l.IsAPOfLength k ∧ ((l.map f).Sorted (· < ·) ∨ (l.map f).Sorted (· > ·))
```
YaelDillies:
```suggestion
arithmetic progression `l` of length `k` in `α` such that its image under `f` is sorted.
-/
def HasMonotoneAP {β : Type*} [Preorder β] (f : α → β) (k : ℕ) : Prop :=
  ∃ l : List α, l.IsAPOfLength k ∧ (l.map f).Sorted (· < ·)
```
but you'll need to replace `List.IsAPOfLengthWith` to be `s = (range l).map fun n ↦ a + n • d \or s = (range l).reverse.map fun n ↦ a + n • d`. Right now, `[5, 4, 3]` isn't an AP, which is bad

FormalConjectures/ErdosProblems/196.lean
danielchin: Done.

FormalConjecturesForMathlib/Combinatorics/AP/Basic.lean
danielchin: Good catch! Changed `List.IsAPOfLengthWith`, but also had to change `length` to get it to build.

YaelDillies (approved):
Thanks!

PR #1707, opened 2026-01-17
feat(ErdosProblems): 239

danielchin:
Adding Erdos Problem 239: https://www.erdosproblems.com/239

### DIFF
--- a/FormalConjectures/ErdosProblems/239.lean
+++ b/FormalConjectures/ErdosProblems/239.lean
@@ -33,14 +33,6 @@ open scoped Topology
 namespace Erdos239
 
 /--
-Property that $f: \mathbb{N} \to \{-1, 1\}$ is a multiplicative function.
--/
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/239.lean
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Erdos239
+
+/--
+Property that $f: \mathbb{N} \to \{-1, 1\}$ is a multiplicative function.
+-/
+structure IsMultiplicativePlusMinusOne (f : ℕ → ℝ) : Prop where
+  range : ∀ n ≥ 1, f n = 1 ∨ f n = -1
+  mult  : ∀ m n, m.Coprime n → f (m * n) = f m * f n
+  one   : f 1 = 1
```
YaelDillies: Could you inline this?

danielchin: Done.

YaelDillies (approved):
Thanks!

PR #1708, opened 2026-01-17
feat(ErdosProblems): 275

danielchin:
Formalizing Erdos Problem 275: https://www.erdosproblems.com/275

### DIFF
--- a/FormalConjectures/ErdosProblems/275.lean
+++ b/FormalConjectures/ErdosProblems/275.lean
@@ -20,7 +20,7 @@ import FormalConjectures.Util.ProblemImports
 # Erdős Problem 275
 
 *References:*
-- [erdosproblems.com/270](https://www.erdosproblems.com/270)
+- [erdosproblems.com/275](https://www.erdosproblems.com/275)
 - [CrVE70] R.B. Crittenden and C.L. Vanden Eynden, *Any n arithmetic progressions covering the first
   2^n integers cover all integers*, Proc. Amer. Math. Soc. 24 (1970), 475-481.
 -/
@@ -30,13 +30,6 @@ open Set
 namespace Erdos275
 
 /--
-A system of $r$ congruences $\{(a_i, n_i)\}_{i=1}^r$ covers an integer $x$ if $x \equiv a_i
-\pmod{n_i}$ for some $i$.
--/
-def isCoveredBy {r : ℕ} (a : Fin r → ℤ) (n : Fin r → ℕ) (x : ℤ) : Prop :=
-  ∃ i, x ≡ a i [ZMOD n i]
-
-/--
 If a finite system of $r$ congruences $\{ a_i\pmod{n_i} : 1\leq i\leq r\}$ (the $n_i$ are not
 necessarily distinct) covers $2^r$ consecutive integers then it covers all integers.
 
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/275.lean
```diff
@@ -0,0 +1,52 @@
...
+open Set
+
+namespace Erdos275
+
+/--
+A system of $r$ congruences $\{(a_i, n_i)\}_{i=1}^r$ covers an integer $x$ if $x \equiv a_i
+\pmod{n_i}$ for some $i$.
+-/
+def isCoveredBy {r : ℕ} (a : Fin r → ℤ) (n : Fin r → ℕ) (x : ℤ) : Prop :=
+  ∃ i, x ≡ a i [ZMOD n i]
```
YaelDillies: Could you inline this?

FormalConjectures/ErdosProblems/275.lean
```diff
@@ -0,0 +1,52 @@
...
+necessarily distinct) covers $2^r$ consecutive integers then it covers all integers.
+
+This is best possible as the system $2^{i-1}\pmod{2^i}$ shows. This was proved indepedently by
+Selfridge and Crittenden and Vanden Eynden [CrVE70].
+-/
+@[category research solved, AMS 11]
+theorem erdos_275 (r : ℕ) (a : Fin r → ℤ) (n : Fin r → ℕ) :
+    (∃ k : ℤ, ∀ x ∈ Ico k (k + 2 ^ r), isCoveredBy a n x) →
+    (∀ x : ℤ, isCoveredBy a n x) := by
+  sorry
```
Paul-Lez: To stick closer to Mathlib's conventions
```suggestion
theorem erdos_275 (r : ℕ) (a : Fin r → ℤ) (n : Fin r → ℕ) 
    (H : ∃ k : ℤ, ∀ x ∈ Ico k (k + 2 ^ r), isCoveredBy a n x) (x : ℤ) :
    isCoveredBy a n x := by
  sorry
```

FormalConjectures/ErdosProblems/275.lean
danielchin: Done. For future reference, what's the heuristic for when to inline vs when not to inline? I usually like to break things out when I'm repeating code.

FormalConjectures/ErdosProblems/275.lean
danielchin: Done.

Also for future reference, is this the primary style guide I should be looking at to match Mathlib conventions? https://leanprover-community.github.io/contribute/style.html

FormalConjectures/ErdosProblems/275.lean
YaelDillies: I would tend to inline if it's not a named definition in paper mathematics, or if every variable appears exactly once (eg `def Foo (a : Nat) : Prop := a = a` vs `def Bar (a : Nat) : Prop := a = 0`)

FormalConjectures/ErdosProblems/275.lean
YaelDillies: Yes, this is the reference, although it certainly is incomplete compared to actual practice

FormalConjectures/ErdosProblems/275.lean
```diff
@@ -0,0 +1,45 @@
...
+
+open Set
+
+namespace Erdos275
+
+/--
+If a finite system of $r$ congruences $\{ a_i\pmod{n_i} : 1\leq i\leq r\}$ (the $n_i$ are not
+necessarily distinct) covers $2^r$ consecutive integers then it covers all integers.
+
+This is best possible as the system $2^{i-1}\pmod{2^i}$ shows. This was proved indepedently by
```
YaelDillies:
```suggestion
This is best possible as the system $2^{i-1}\pmod{2^i}$ shows. This was proved independently by
```

danielchin: Done.

YaelDillies (approved):
Thanks!

PR #1711, opened 2026-01-17
fix(ErdosProblems/961): assume `k` positive

seewoo5:
Fix misformalization, where the statement was false for $k = 0$. Found by Harmonic's Aristotle.

### DIFF
--- a/FormalConjectures/ErdosProblems/961.lean
+++ b/FormalConjectures/ErdosProblems/961.lean
@@ -41,13 +41,12 @@ theorem erdos_961.sylvester_schur_1_1 : Erdos961Prop 1 1 := by
   intro m hm
   use m
   constructor
-  · simp only [Set.mem_Ico]
-    exact ⟨le_refl m, Nat.lt_succ_self m⟩
+  · simp
   · rw [Nat.mem_smoothNumbers]
     push_neg
     intro hm0
     obtain ⟨p, hp, hpm⟩ := Nat.exists_prime_and_dvd (by omega : m ≠ 1)
-    exact ⟨p, (Nat.mem_primeFactorsList hm0).mpr ⟨hp, hpm⟩, Nat.Prime.two_le hp⟩
+    exact ⟨p, (Nat.mem_primeFactorsList hm0).mpr ⟨hp, hpm⟩, hp.two_le⟩
 
 @[category research solved, AMS 11]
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/961.lean
```diff
@@ -33,19 +33,33 @@ Sylvester and Schur [Er34] proved that every set of $k$ consecutive integers gre
...
+  intro m hm
+  use m
+  constructor
+  · simp only [Set.mem_Ico]
+    exact ⟨le_refl m, Nat.lt_succ_self m⟩
+  · rw [Nat.mem_smoothNumbers]
+    push_neg
+    intro hm0
+    obtain ⟨p, hp, hpm⟩ := Nat.exists_prime_and_dvd (by omega : m ≠ 1)
+    exact ⟨p, (Nat.mem_primeFactorsList hm0).mpr ⟨hp, hpm⟩, Nat.Prime.two_le hp⟩
```
YaelDillies:
```suggestion
    exact ⟨p, (Nat.mem_primeFactorsList hm0).mpr ⟨hp, hpm⟩, hp.two_le⟩
```

FormalConjectures/ErdosProblems/961.lean
```diff
@@ -33,19 +33,33 @@ Sylvester and Schur [Er34] proved that every set of $k$ consecutive integers gre
...
+theorem erdos_961.sylvester_schur (k : ℕ) (hk : 0 < k) : Erdos961Prop k k := by
   sorry
 
+@[category test, AMS 11]
+theorem erdos_961.sylvester_schur_1_1 : Erdos961Prop 1 1 := by
+  intro m hm
+  use m
+  constructor
+  · simp only [Set.mem_Ico]
+    exact ⟨le_refl m, Nat.lt_succ_self m⟩
```
YaelDillies:
```suggestion
  · simp
```
surely?

seewoo5: Yep it works!

seewoo5:
> Where is this talked about? Does it also concern our formalisation of 683?

It was found while I was trying to finish #1533 - I was curious to see if Aristotle can actually do that (i.e. prove equivalence between 683 and 961), and it proved half of it but also told me about this misformalization as a bonus. I think there's no issue with formalization of 683 in #1533.

YaelDillies (approved):
Great, let's go!

PR #1730, opened 2026-01-17
mark Erdős 591 as solved

mo271:
following a recent update on https://www.erdosproblems.com/591

### DIFF
--- a/FormalConjectures/ErdosProblems/591.lean
+++ b/FormalConjectures/ErdosProblems/591.lean
@@ -19,7 +19,9 @@ import FormalConjectures.Util.ProblemImports
 /-!
 # Erdős Problem 591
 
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/591.lean
```diff
@@ -31,9 +33,11 @@ universe u
 /--
 Let $α$ be the infinite ordinal $\omega^{\omega^2}$. Is it true that any red/blue colouring of the
 edges of $K_α$ there is either a red $K_α$ or a blue $K_3$.
```
Paul-Lez: (bikeshed)
```suggestion
edges of $K_α$ there is either a red $K_α$ or a blue $K_3$?
```

Paul-Lez (approved):

PR #1767, opened 2026-01-19
feat(ErdosProblems): 277

danielchin:
Formalizing Erdos Problem 277: https://www.erdosproblems.com/277


YaelDillies (changes requested):
Thanks!

YaelDillies (approved):

PR #1777, opened 2026-01-20
feat: Toronto problem

### LINKED ISSUE #1776: Toronto problem
### What is the conjecture
A Toronto space is a topological space homeomorphic to all of its subspaces of same cardinality.

It is conjectured that every T2, Toronto space is discrete. It is known this result holds under GCH, see Proposition 2.6 in the second reference.

- [Wikipedia](https://en.wikipedia.org/wiki/Toronto_space)
- [The Toronto problem](https://wrbrian.wordpress.com/wp-content/uploads/2012/01/thetorontoproblem.pdf)

### Prerequisites needed
Toronto spaces are not in mathlib but easy to define.

* ams-54

### Choose either option

felixpernegger:
Closes #1776.

### DIFF
--- a/FormalConjectures/Wikipedia/Toronto.lean
+++ b/FormalConjectures/Wikipedia/Toronto.lean
@@ -60,7 +60,7 @@ instance Finite.torontoSpace [Finite X] : TorontoSpace X where
 
 /-- Any T2, Toronto space is discrete. -/
 @[category research open, AMS 54]
-instance DiscreteTopology.of_t2_of_torontoSpace [T2Space X] [TorontoSpace X] :
+theorem DiscreteTopology.of_t2_of_torontoSpace [T2Space X] [TorontoSpace X] :
     DiscreteTopology X := by
   sorry
 


FormalConjectures/Wikipedia/Toronto.lean
```diff
@@ -0,0 +1,67 @@
...
+
+open Cardinal
+
+variable (X : Type*) [TopologicalSpace X]
+
+/-- A *Toronto space* is a topological space
+which is homeomorphic to all of its subspaces of same cardinality.
+-/
+class TorontoSpace where
+  toronto : ∀ Y : Set X, #Y = #X → Y ≃ₜ X
```
Paul-Lez:
```suggestion
  toronto : ∀ ⦃Y : Set X⦄, #Y = #X → Y ≃ₜ X
```

FormalConjectures/Wikipedia/Toronto.lean
```diff
@@ -0,0 +1,67 @@
...
+      exact rfl
+    rw [eq]
+    exact Homeomorph.Set.univ X
+
+/-- Any T2, Toronto space is discrete. -/
+@[category research open, AMS 54]
+instance DiscreteTopology.of_t2_of_torontoSpace [T2Space X] [TorontoSpace X] :
+    DiscreteTopology X := by
+  sorry
+
```
Paul-Lez, line 66:
```suggestion

-- TODO: Add implication that this holds under the GCH

```

FormalConjectures/Wikipedia/Toronto.lean
```diff
@@ -0,0 +1,67 @@
...
+      rw [this, hY]
+      exact rfl
+    rw [eq]
+    exact Homeomorph.Set.univ X
+
+/-- Any T2, Toronto space is discrete. -/
+@[category research open, AMS 54]
+instance DiscreteTopology.of_t2_of_torontoSpace [T2Space X] [TorontoSpace X] :
+    DiscreteTopology X := by
+  sorry
```
Paul-Lez, line 65: Actually I think this isn't quite right: what you want here is a theorem saying that the topology on `X` is the discrete one

Paul-Lez: So the theorem should say `IsDiscrete (Set.univ : Set X)`

felixpernegger: Are you sure? I copied this format from ```Finite.instDiscreteTopology``` from mathlib

felixpernegger: (```DiscreteTopology``` is actually a proposition confusingly enough)

Paul-Lez: Ah actually you're right my bad - I misread the docs!

felixpernegger: Thanks for the review :)

FormalConjectures/Wikipedia/Toronto.lean
```diff
@@ -0,0 +1,69 @@
...
+      have : Y.ncard = (#↑Y).toNat := by
+        exact rfl
+      rw [this, hY]
+      exact rfl
+    rw [eq]
+    exact Homeomorph.Set.univ X
+
+/-- Any T2, Toronto space is discrete. -/
+@[category research open, AMS 54]
+instance DiscreteTopology.of_t2_of_torontoSpace [T2Space X] [TorontoSpace X] :
```
Paul-Lez: Final small nitpick: our convention is to avoid stating open problems as instances (even though this would be sensible here!)
```suggestion
theorem DiscreteTopology.of_t2_of_torontoSpace [T2Space X] [TorontoSpace X] :
```

PR #1778, opened 2026-01-20
feat(ErdosProblems): 341

### LINKED ISSUE #500: Erdős Problem 341
### What is the conjecture
    
https://www.erdosproblems.com/341

  Let $A=\\{a_1<\cdots<a_k\\}$ be a finite set of integers and extend it to an infinite sequence $\overline{A}=\\{a_1<a_2<\cdots \\}$ by defining $a_{n+1}$ for $n\geq k$ to be the least integer exceeding $a_n$ which is not of the form $a_i+a_j$ with $i,j\leq n$. Is it true that the sequence of differences $a_{m+1}-a_m$ is eventually periodic?

Status: open

### Choose either option

danielchin:
Closes #500 

Formalizes Erdos Problem 341: https://www.erdosproblems.com/341

### DIFF
--- a/FormalConjectures/ErdosProblems/341.lean
+++ b/FormalConjectures/ErdosProblems/341.lean
@@ -37,11 +37,8 @@ This problem is discussed under Problem 7 on Green's open problems list.
 @[category research open, AMS 11]
 theorem erdos_341 :
     answer(sorry) ↔
-      ∀ (l : List ℤ) (a : ℕ → ℤ),
-        l.Sorted (· < ·) →
-        l ≠ [] →
-        (List.range l.length).map a = l →
-        (∀ n ≥ l.length - 1,
+      ∀ (a : ℕ → ℤ),
+        (∀ᶠ n in atTop,
           IsLeast { x | a n < x ∧ x ∉ { a i + a j | (i ≤ n) (j ≤ n) } } (a (n + 1))) →
         let d := fun i ↦ a (i + 1) - a i
         ∃ p > 0, ∀ᶠ m in atTop, d (m + p) = d m := by
--- a/FormalConjectures/ErdosProblems/341.lean
+++ b/FormalConjectures/ErdosProblems/341.lean
@@ -1,0 +1,49 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


FormalConjectures/ErdosProblems/341.lean
```diff
@@ -0,0 +1,50 @@
...
+@[category research open, AMS 11]
+theorem erdos_341 :
+    answer(sorry) ↔
+      ∀ (l : List ℤ) (a : ℕ → ℤ),
+        l.Sorted (· < ·) →
+        l ≠ [] →
+        (List.range l.length).map a = l →
+        (∀ n ≥ l.length - 1,
+          IsLeast { x | a n < x ∧ x ∉ { a i + a j | (i ≤ n) (j ≤ n) } } (a (n + 1))) →
+        let d := fun i ↦ a (i + 1) - a i
```
Paul-Lez: Note that here you don't really need the list `l`: you can simply quantify over `a` and the predicate that for sufficiently large `n`, the relation `IsLeast { x | a n < x ∧ x ∉ { a i + a j | (i ≤ n) (j ≤ n) } } (a (n + 1)))` holds

danielchin: Ah, thanks for pointing that out. Thanks!

FormalConjectures/ErdosProblems/341.lean
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Erdos341
+
+/--
+Let $A=\{a_1 < \cdots < a_k\}$ be a finite set of integers and extend it to an infinite
+sequence $\overline{A}=\{a_1 < a_2 < \cdots \}$ by defining $a_{n+1}$ for $n \geq k$ to be
+the least integer exceeding $a_n$ which is not of the form $a_i + a_j$ with $i,j \leq n$.
+Is it true that the sequence of differences $a_{m+1}-a_m$ is eventually periodic?
+
+This problem is discussed under Problem 7 on Green's open problems list.
```
YaelDillies, line 35: Can you add a file under `GreensOpenProblems` importing this file?

danielchin: Done.

danielchin:
Let me know if that was the proper way to import a theorem. 

Also to note, this is just a variant on Green's 7 and not the actual problem, so I left a TODO in the 7.lean file.

YaelDillies (approved):
Thanks!

PR #1781, opened 2026-01-20
feat(ErdosProblems): 417

### LINKED ISSUE #719: Erdős Probelm 417
### What is the conjecture
    
https://www.erdosproblems.com/417

  Let 
$$V'(x)=\\#\\{\phi(m) : 1\leq m\leq x\\}$$
and
$$V(x)=\\#\\{\phi(m) \leq x : 1\leq m\\}.$$
Does $\lim V(x)/V'(x)$ exist? Is it $>1$?

Status: open

### Choose either option

danielchin:
Closes #719 

Formalizes Erdos Problem 417: https://www.erdosproblems.com/417

### DIFF
--- a/FormalConjectures/ErdosProblems/417.lean
+++ b/FormalConjectures/ErdosProblems/417.lean
@@ -33,12 +33,15 @@ namespace Erdos417
 /--
 Let\[V'(x)=\#\{\phi(m) : 1\leq m\leq x\}\]and\[V(x)=\#\{\phi(m) \leq x : 1\leq m\}.\]
 Does $\lim V(x)/V'(x)$ exist?
+
+Formalization note: We formalize the limit of the inverse fraction V'(x)/V(x)
+to ensure the limit is finite (bounded between 0 and 1).
 -/
 @[category research open, AMS 11]
 theorem erdos_417.part.a :
     answer(sorry) ↔ ∃ L : ℝ, Tendsto (fun x ↦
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/417.lean
```diff
@@ -0,0 +1,56 @@
...
+/--
+Let\[V'(x)=\#\{\phi(m) : 1\leq m\leq x\}\]and\[V(x)=\#\{\phi(m) \leq x : 1\leq m\}.\]
+Does $\lim V(x)/V'(x)$ exist?
+-/
+@[category research open, AMS 11]
+theorem erdos_417.part.a :
+    answer(sorry) ↔ ∃ L : ℝ, Tendsto (fun x ↦
+      ({ k | k ∈ range totient ∧ (k : ℝ) ≤ x }.ncard : ℝ) /
+      ((totient '' { m | 1 ≤ m ∧ (m : ℝ) ≤ x }).ncard : ℝ))
+      atTop (𝓝 L) := by
```
YaelDillies, line 45: This won't hold true if the limit is infinite. Maybe reverse all the fractions, so that all possible limits are finite? Same below

danielchin: Good catch, I've flipped the fractions, changed the direction of the equality in part b, and left a note in the docstring to explain why it's formalized like that.

YaelDillies (approved):
Thanks!

PR #1786, opened 2026-01-20
feat: Infinitude of Woodall primes

### LINKED ISSUE #1782: Wikipedia: Woodall Primes
### What is the conjecture
There are infinite prime Woodall numbers, i.e., numbers of form $k2^k - 1$, $k>1$.
 
* https://en.wikipedia.org/wiki/Woodall_number#Woodall_primes
* https://oeis.org/A002234

* ams-11

### Choose either option

Paul-Lez:
Fixes #1782.

### DIFF
--- a/FormalConjectures/Wikipedia/WoodalPrimes.lean
+++ b/FormalConjectures/Wikipedia/WoodalPrimes.lean
@@ -1,0 +1,26 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+namespace WoodallPrimes
+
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/WoodalPrimes.lean
```diff
@@ -13,24 +13,22 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+-/
 
-namespace Erdos855
+namespace WoodallPrimes
 
+/-- There are infinitely many prime numbers of the form `k * 2 ^ k - 1` for `k > 1`. -/
 @[category research open, AMS 11]
-theorem erdos_855 : answer(sorry) ↔
-    ∀ᶠ x in atTop, ∀ᶠ y in atTop, π (x + y) ≤ π x + π y := by
+theorem infinitely_many_woodall_primes : {(k * 2 ^ k - 1) | (k : ℕ) (_ : 1 < k)}.Infinite := by
```
felixpernegger:
```suggestion
theorem infinitely_many_woodall_primes : {k : ℕ | Nat.Prime (k * 2 ^ k - 1)}.Infinite := by
```

Shouldnt this be something like this rather? I dont think your statement mentions primes. (Also note $k>1$ is redundant)

Paul-Lez: Indeed (what was I thinking?!). 

Re the `k > 1` condition - this is indeed redundant, but I personally prefer to keep things closer to the informal text when it doesn't make the formalisation much more cumbersome.

mo271: forgot to push?!

Paul-Lez: Indeed - pushed!

mo271 (approved):

PR #1788, opened 2026-01-20
feat(Mathoverflow): 235893

### LINKED ISSUE #1787: Mathoverflow/235893
### What is the conjecture
For topological spaces $X$ and $Y$ we say a function $f : X → Y$ is *connected* is it sends
connected sets to connected sets.

(When) Does there exist a connected bijection $f: \mathbb{R}^n \to \mathbb{R}^n$ such that the inverse is not connected?

See
https://mathoverflow.net/questions/235893/does-there-exist-a-bijection-of-mathbbrn-to-itself-such-that-the-forward-m

The motivation for this, is that this is the highest-upvoted (corresponding) question on mathstackexchange with no answers.

* ams-26
* ams-54

### Choose either option

felixpernegger:
Closes #1787.

### DIFF
--- a/FormalConjectures/Mathoverflow/235893.lean
+++ b/FormalConjectures/Mathoverflow/235893.lean
@@ -16,14 +16,14 @@ limitations under the License.
 
 import FormalConjectures.Util.ProblemImports
 
-open scoped EuclideanGeometry
-
 /-!
 # Mathoverflow 235893
 
 *Reference:* [mathoverflow/235893](https://mathoverflow.net/questions/235893)
 asked by user [*Willie Wong*](https://mathoverflow.net/users/3948/willie-wong)
 -/
+
+open scoped EuclideanGeometry
 
 namespace Mathoverflow235893
 


FormalConjectures/Mathoverflow/235893.lean
```diff
@@ -0,0 +1,66 @@
...
+theorem Continuous.isConnectedMap {X Y : Type*} [TopologicalSpace X] [TopologicalSpace Y]
+    {f : X → Y} (hf : Continuous f) : IsConnectedMap f :=
+  fun _ h ↦ IsConnected.image h f (Continuous.continuousOn hf)
+
+/--
+Does there exist a bijection $f : ℝ^n → ℝ^n$ such that $f$ is connected but the inverse is not?
+-/
+@[category research open, AMS 26 54]
+theorem mathoverflow_235893 (n : ℕ) :
+  ∃ (f : ℝ^n → ℝ^n) (hf : Function.Bijective f),
```
Paul-Lez: Here I'd just take `f` to be an Equiv directly to avoid messing around with `Equiv.ofBijective` below. I'd also suggest using the `answer(sorry)` convention since this is formulated as a question (that's the convention we usually follow in this repo). 
Do we also want `n` to be positive?

FormalConjectures/Mathoverflow/235893.lean
```diff
@@ -0,0 +1,66 @@
...
+def IsConnectedMap {X Y : Type*} [TopologicalSpace X] [TopologicalSpace Y] (f : X → Y) : Prop :=
+  ∀ ⦃s : Set X⦄, IsConnected s → IsConnected (f '' s)
+
+/--
+By a standard result, every continuous map is connected
+-/
+@[category test, AMS 54]
+theorem Continuous.isConnectedMap {X Y : Type*} [TopologicalSpace X] [TopologicalSpace Y]
+    {f : X → Y} (hf : Continuous f) : IsConnectedMap f :=
+  fun _ h ↦ IsConnected.image h f (Continuous.continuousOn hf)
```
Paul-Lez, line 44: This could make sense to add to `FormalConjecturesForMathlib`.

Paul-Lez:
If you have the appetite, it could also be nice to add some of the results mentioned in the remarks part of the MO post!

FormalConjectures/Mathoverflow/235893.lean
felixpernegger: Thank you for your comments, I hope I applied them correctly!

I dont think requiring ```n``` to be positive is necessary, since it's unclear what the answer would be in general anyways (could be dependend on ```n```), so that would just make the theorem uglier

FormalConjectures/Mathoverflow/235893.lean
```diff
@@ -0,0 +1,67 @@
...
+
+/--
+Does there exist a bijection $f : ℝ^n → ℝ^n$ such that $f$ is connected but the inverse is not?
+-/
+@[category research open, AMS 26 54]
+theorem mathoverflow_235893 (n : ℕ) :
+    answer(sorry) ↔ ∃ (f : ℝ^n ≅ ℝ^n), IsConnectedMap f.hom ∧ ¬ IsConnectedMap f.symm.hom := by
```
Paul-Lez: This is a slighly more standard way of writing this down 
```suggestion
    answer(sorry) ↔ ∃ (f : ℝ^n ≃ ℝ^n), IsConnectedMap f ∧ ¬ IsConnectedMap f.symm := by
```

FormalConjectures/Mathoverflow/235893.lean
```diff
@@ -0,0 +1,67 @@
...
+@[category test, AMS 54]
+theorem Continuous.isConnectedMap {X Y : Type*} [TopologicalSpace X] [TopologicalSpace Y]
+    {f : X → Y} (hf : Continuous f) : IsConnectedMap f :=
+  fun _ h ↦ IsConnected.image h f (Continuous.continuousOn hf)
+
+/--
+Does there exist a bijection $f : ℝ^n → ℝ^n$ such that $f$ is connected but the inverse is not?
+-/
+@[category research open, AMS 26 54]
+theorem mathoverflow_235893 (n : ℕ) :
```
Paul-Lez: Usually, we quantify _within_ the equivalence when using the answer sorry convention (so that replacing `sorry` with `False` allows us to reverse quantifiers). I do think you will need `n` positive though, since the result is trivially false when `n = 0`.
```lean
example :
    ∀ (f : EuclideanSpace ℝ (Fin 0) ≃ EuclideanSpace ℝ (Fin 0)), IsConnectedMap f ∧ IsConnectedMap f.symm := by
  intro f
  have : Subsingleton (EuclideanSpace ℝ (Fin 0)) := inferInstance
  have (F : EuclideanSpace ℝ (Fin 0) → EuclideanSpace ℝ (Fin 0)) : IsConnectedMap F := by  
    apply Subsingleton.set_cases
    · intro H
      exfalso
      apply Set.not_nonempty_empty H.nonempty
    · intro H
      have : F '' .univ = .univ := by
        suffices default ∈ F '' .univ by apply Subsingleton.eq_univ_of_nonempty (by use default)
        rw [Subsingleton.allEq default (F default)]
        exact Set.mem_image_of_mem F trivial
      rw [this]
      exact isConnected_univ
  exact ⟨this _, this _⟩
```

felixpernegger: Thanks, k = 0 does not matter, but I do have a more general question then:

In cases like this, where it is actually unclear what the answer might be, is it generally preferable to rule out trivial/known edge cases (like k=0 here)? 

Doing this *might* make the answer(sorry) (when it gets solved) a bit "nicer", but mentioning edge cases makes the problem statement longer, despite the conjecture (essentially) staying the same.

So what is the convention here?

Paul-Lez: Good questions! Generally speaking, I'd say it's better to remove edge cases that make a statement trivially provable or disprovable.

Paul-Lez: (Effectively, the goal is to avoid situations where one can easily "prove" an open statement by resorting to something trivial)

Paul-Lez: There are cases where this matters less (e.g if a conjecture about positive natural numbers also happens to hold for 0 by chance then it's generally fair to include that edge case)

felixpernegger: Okay this makes sense, thanks.
One more thing though, why would including edge cases make theorems trivially provable sometimes? If I understand correctly, ```answer(sorry)``` can be replaced by any (meaningful) statement.
So for example in this case, when ```k``` is in the context, the answer to the theorem could be (in theory) ```Odd k or k = 0```.

I do see that this is a problem when you have quantifiers in the goal, yeah.

Paul-Lez: Agreed - the above only applies when the quantifies are "within the goal" (i.e. when we can't quantify over the variables that have edge cases in the `answer(sorry)` block)

Paul-Lez: (but currently we tend to prefer having quantifiers within the goal when possible)

felixpernegger: okk

FormalConjectures/ErdosProblems/509.lean
```diff
@@ -119,4 +119,13 @@ theorem erdos_509.variants.Pommerenke_connected : answer(True) ↔ ∀ (f : ℂ[
     ∃ (ι : Type), Nonempty (BoundedDiscCover {z | ‖f.eval z‖ ≤ 1} 2 ι) := by
   sorry
 
+
+open MeasureTheory in
+@[category research open, AMS 28]
+theorem green_94 :
+   answer(sorry) ↔  ∀ A : Set ℝ,
```
YaelDillies:
```suggestion
   answer(sorry) ↔ ∀ A : Set ℝ,
```

FormalConjectures/ErdosProblems/509.lean
```diff
@@ -119,4 +119,13 @@ theorem erdos_509.variants.Pommerenke_connected : answer(True) ↔ ∀ (f : ℂ[
...
+    volume A > 0 → ∃ a b : ℝ, 0 < a ∧ ∀ n : ℕ, a * (2 ^ n)⁻¹ + b ∈ A := by
+  simp
```
YaelDillies:
```suggestion
```

FormalConjectures/ErdosProblems/509.lean
```diff
@@ -119,4 +119,13 @@ theorem erdos_509.variants.Pommerenke_connected : answer(True) ↔ ∀ (f : ℂ[
     ∃ (ι : Type), Nonempty (BoundedDiscCover {z | ‖f.eval z‖ ≤ 1} 2 ι) := by
   sorry
 
+
+open MeasureTheory in
```
YaelDillies:
```suggestion
open MeasureTheory in
```

FormalConjectures/ErdosProblems/509.lean
```diff
@@ -119,4 +119,13 @@ theorem erdos_509.variants.Pommerenke_connected : answer(True) ↔ ∀ (f : ℂ[
...
+@[category research open, AMS 28]
+theorem green_94 :
+   answer(sorry) ↔  ∀ A : Set ℝ,
+    volume A > 0 → ∃ a b : ℝ, 0 < a ∧ ∀ n : ℕ, a * (2 ^ n)⁻¹ + b ∈ A := by
```
YaelDillies: Can you add a docstring?

FormalConjectures/ErdosProblems/509.lean
YaelDillies, line 1: What are these changes doing here?

FormalConjectures/Mathoverflow/235893.lean
```diff
@@ -0,0 +1,66 @@
...
+  sorry
+
+/--
+There exists a connected bijection ℝ → ℝ^2 where the inverse is not connected,
+proven in [mathoverflow/260589](https://mathoverflow.net/questions/260589) by user
+[Gro-Tsen](https://mathoverflow.net/users/17064/gro-tsen).
+-/
+@[category research solved, AMS 26 54]
+theorem mathoverflow_260589 :
+    ∃ f : ℝ ≅ ℝ^2, IsConnectedMap f.hom ∧ ¬ IsConnectedMap f.symm.hom := by
```
YaelDillies: Wrong symbol?
```suggestion
    ∃ f : ℝ \equiv ℝ^2, IsConnectedMap f ∧ ¬ IsConnectedMap f.symm := by
```

FormalConjectures/ErdosProblems/509.lean
felixpernegger: Sorry I (again) accidently included something from wrong branch/PR. Will check this more carefully in the future.

FormalConjectures/ErdosProblems/509.lean
```diff
@@ -1,122 +0,0 @@
-/-
```
Paul-Lez, line 1: I think we shouldn't delete this file ideally;)

felixpernegger: yessss....

FormalConjectures/Mathoverflow/235893.lean
felixpernegger: thanks

FormalConjectures/Mathoverflow/235893.lean
```diff
@@ -0,0 +1,66 @@
...
+import FormalConjectures.Util.ProblemImports
+
+open scoped EuclideanGeometry
+
+/-!
+# Mathoverflow 235893
+
+*Reference:* [mathoverflow/235893](https://mathoverflow.net/questions/235893)
+asked by user [*Willie Wong*](https://mathoverflow.net/users/3948/willie-wong)
+-/
```
YaelDillies, line 24:
```suggestion

/-!
# Mathoverflow 235893

*Reference:* [mathoverflow/235893](https://mathoverflow.net/questions/235893)
asked by user [*Willie Wong*](https://mathoverflow.net/users/3948/willie-wong)
-/

open scoped EuclideanGeometry
```

PR #1811, opened 2026-01-21
feat: Infinitely many Pell primes

### LINKED ISSUE #1779: Wikipedia: Infinite Pell Primes
### What is the conjecture

There are infinite integer numbers that are both prime and in the Pell sequence.

* https://en.wikipedia.org/wiki/Pell_number#Primes_and_squares
* https://oeis.org/A086383
* https://oeis.org/A096650

* ams-11

### Choose either option

felixpernegger:
Closes #1779.

### DIFF
--- a/FormalConjectures/Wikipedia/Pell.lean
+++ b/FormalConjectures/Wikipedia/Pell.lean
@@ -1,5 +1,5 @@
 /-
-Copyright 2025 The Formal Conjectures Authors.
+Copyright 2026 The Formal Conjectures Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.


FormalConjectures/Wikipedia/Pell.lean
```diff
@@ -0,0 +1,69 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
```
mo271:
```suggestion
Copyright 2026 The Formal Conjectures Authors.
```
https://i.kym-cdn.com/entries/icons/original/000/007/784/what-year-is-it-robin-williams.jpg

FormalConjectures/Wikipedia/Pell.lean
```diff
@@ -0,0 +1,69 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Infinitude of Pell number primes
+
+*Reference:*
```
mo271:
```suggestion
*References:*
```

FormalConjectures/Wikipedia/Pell.lean
```diff
@@ -0,0 +1,69 @@
...
+ - [OEIS A086383](https://oeis.org/A086383)
+
+The Pell numbers $P_n$ are defined by $P_0 = 0$, $P_1 = 1$, $P_{n+2} = 2*P_{n+1} + P_n$
+
+The conjecture says that there are infinitely many prime Pell numbers.
+-/
+
+namespace PellNumbers
+
+/-- The *Pell numbers* $P_n$ are defined by $P_0 = 0$, $P_1 = 1$, $P_{n+2} = 2*P_{n+1} + P_n$ -/
```
mo271, line 34: potentially link https://oeis.org/A000129 here?

FormalConjectures/Wikipedia/Pell.lean
```diff
@@ -0,0 +1,69 @@
...
+def pell : ℕ → ℕ
```
mo271:
```suggestion
def pellNumber : ℕ → ℕ
```

In analogy with `fermatNumber`? On the other hand `fib` is for Fibonacci number, so it is already inconsistent, I guess..

FormalConjectures/Wikipedia/Pell.lean
```diff
@@ -0,0 +1,69 @@
...
+theorem pell_two : pell 2 = 2 := rfl
+
+@[category test, AMS 11]
+theorem pell_five : pell 5 = 29 := rfl
+
+/-- Similar to Fibonacci numbers, there exist numerous identites around Pell numbers, i.e.
+P_{2n+1} = P_n ^ 2 + P_{n+1} ^ 2 -/
+@[category high_school, AMS 11]
+theorem pell_sq_add_pell_succ_sq (n : ℕ) : pell (2 * n + 1) = pell n ^ 2 + pell (n + 1) ^ 2 := by
+  sorry
```
mo271, line 57:
```suggestion
  induction n using Nat.twoStepInduction with
  |zero
  |one => constructor
  |more =>
    simp_all! only [mul_add, mul_one, add_assoc, Nat.reduceAdd, Nat.reduceMul]
    linarith
```
perhaps add the proof?

FormalConjectures/Wikipedia/Pell.lean
felixpernegger: Thanks, I actually made this mistake multiple times

FormalConjectures/Wikipedia/Pell.lean
felixpernegger: nice

FormalConjectures/Wikipedia/Pell.lean
felixpernegger: Yes good idea. ```Pell``` is in mathlib for a related concept, which is not quite the same, so it is probably a good idea to differentiate.

FormalConjectures/Wikipedia/Pell.lean
felixpernegger: like this? (as it is now)

mo271: no worries, it is not really important at all..

PR #1813, opened 2026-01-21
fix(Wikipedia/WallSunSun): restrict `D` to fundamental discriminants

YaelDillies:
Not all integers appear as discriminants of a quadratic field.


FormalConjectures/Wikipedia/WallSunSun.lean
```diff
@@ -40,15 +40,26 @@ $p$-th Lucas number. It is conjectured that there are infinitely many Wall-Sun-S
 theorem infinite_isWallSunSunPrime : {p : ℕ | IsWallSunSunPrime p}.Infinite := by
   sorry
 
+/-- Fundamental discriminants are those integers `D` that appear as discriminants of quadratic
+fields.
```
eric-wieser: Does mathlib not have the notion of a discriminant of a quadratic field?

YaelDillies: Sure it does, but the notion of fundamental discriminant can be described very elementarily. That it corresponds to discriminants of quadratic fields is definitely something we should state, if that's what you mean

Paul-Lez: This could be a nice `category test` or `category undergradute` statement!

FormalConjecturesForMathlib/NumberTheory/NumberField/Quadratic.lean
```diff
@@ -0,0 +1,52 @@
...
+open Module NumberField
+
+namespace QuadraticAlgebra
+variable {d : ℕ}
+
+lemma discr_rat_of_modEq_one (hd₁ : d ≠ 1) (hd : Squarefree d) (hd₄ : d ≡ 1 [ZMOD 4]) :
+    discr (QuadraticAlgebra ℚ d 0) = d := sorry
+
+lemma discr_rat_of_not_modEq_one (hd : Squarefree d) (hd₄ : ¬ d ≡ 1 [ZMOD 4]) :
+    discr (QuadraticAlgebra ℚ d 0) = 4 * d := sorry
```
eric-wieser: I don't think we allow `sorry` in `ForMathlib`, do we?

YaelDillies: Yes, I have moved it to the Wall-Sun-Sun conjecture file, although that too is unsatisfactory

YaelDillies:
I could fill in more of the sorries using the material from https://github.com/sacerdot/QuadraticIntegers and https://github.com/leanprover-community/mathlib4/pull/36387. Let me know if you think it is worth it.

FormalConjectures/Wikipedia/WallSunSun.lean
```diff
@@ -22,7 +21,108 @@ import FormalConjectures.Util.ProblemImports
...
+/-- The discriminant of `ℚ[√d]` for `d ≥ 2` squarefree congruent to 1 mod 4 is `d`. -/
+@[category textbook, AMS 11, simp]
+lemma discr_rat_of_modEq_one (hd₄ : d ≡ 1 [ZMOD 4]) : discr (QuadraticAlgebra ℚ d 0) = d := by
+  sorry
+
+/-- The discriminant of `ℚ[√d]` for `d ≥ 2` squarefree not congruent to 1 mod 4 is `4 * d`. -/
+@[category textbook, AMS 11, simp]
+lemma discr_rat_of_not_modEq_one (hd₄ : ¬ d ≡ 1 [ZMOD 4]) :
+    discr (QuadraticAlgebra ℚ d 0) = 4 * d := by
+  sorry
```
eric-wieser, line 39: I don't think these should be `simp`, that just poisons agents that try to use the `simp` tactic.

eric-wieser:
I don't think it's worth backporting, but it's probably worth a link to the results existing elsewhere.

FormalConjectures/Wikipedia/WallSunSun.lean
YaelDillies: How so? The LHS is very specific

YaelDillies:
I don't think the results exist elsewhere yet. Riccardo's ItaLean project and Frankie's PRs get us most of the way there, but there is still some amount of missing glue.

Paul-Lez (approved):

PR #1836, opened 2026-01-21
feat: Carmichael's totient function conjecture

### LINKED ISSUE #1784: Wikipedia: Carmichael's totient function conjecture
### What is the conjecture
There is no integer $n$ such that the equation $\phi(m) = n$ has only one solution, $\phi(m)$ being the Euler totient function that counts the number of integers less than and coprime to $m$ (Robert D. Carmichael, 1907).

* https://en.wikipedia.org/wiki/Carmichael's_totient_function_conjecture
* https://oeis.org/A014197

* ams-11

### Choose either option

felixpernegger:
Closes #1784.

### DIFF
--- a/FormalConjectures/Wikipedia/CarmichaelTotient.lean
+++ b/FormalConjectures/Wikipedia/CarmichaelTotient.lean
@@ -36,7 +36,7 @@ $φ(n) = φ(m)$ where $φ$ is the Euler totient function.
 /-- Natural number $n$ for which there exists a $m ≠ n$ with $φ(m) = φ(n)$ -/
 def CarmichaelTotientFor (n : ℕ) : Prop := ∃ m : ℕ, m ≠ n ∧ φ m = φ n
 
-/-- n = 0 ↔ φ(n) = 0 -/
+/-- $n = 0 ↔ φ(n) = 0$ -/
 @[category test, AMS 11]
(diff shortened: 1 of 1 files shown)


Paul-Lez:
I think this is also in the Erdos Problems list! (or is very similar to a problem in that list - let me see if I can find out which one)

FormalConjectures/Wikipedia/CarmichaelTotient.lean
```diff
@@ -0,0 +1,67 @@
...
+/-- $n = 0 ↔ φ(n) = 0$ -/
+@[category test, AMS 11]
+theorem carchimichealTotientFor_zero : ¬ CarmichaelTotientFor 0 := by
+  simp [CarmichaelTotientFor]
+
+-- TODO: Version of this ↓ lemma to mathlib?
+
+/-- For every odd number $n$, $φ(2n) = φ(n)$ -/
+@[category high_school, AMS 11]
+theorem carmichealTotientFor_odd {n : ℕ} (hn : Odd n) : CarmichaelTotientFor n := by
```
YaelDillies, line 48: Sure, yes! Please add that to ForMathlib

felixpernegger: I just PR'ed this to mathlib directly:
https://github.com/leanprover-community/mathlib4/pull/34264

PR #1851, opened 2026-01-22
feat(Util): add formally solved here option to the category tag

Paul-Lez:
As discussed [here](https://github.com/google-deepmind/formal-conjectures/pull/1615#issuecomment-3747023568)

### DIFF
--- a/FormalConjectures/Util/Attributes.lean
+++ b/FormalConjectures/Util/Attributes.lean
@@ -40,9 +40,11 @@ The values of this attribute are
   The criterion for being solved is that there exists an informal solution
   that is widely accepted by experts in the area. In particular, this
   does *not* require a formal solution to exist.
-- `@[category research formally solved at "link"]` : a research level math problem
-  that has been formally solved in this repository, with "link" being a permalink
-  to the formal solution.
+- @[category research formally solved using <kind> at "link"]` : a formally solved research problem.
+   `<kind>` must be one of:
+  - `formal_conjectures` : solved in this repository (e.g. filling the `sorry` in the current file).
+  - `lean4` : solved in Lean 4 (e.g. Mathlib, or some other repo) as an equivalent statement.
+  - `other_system` : solved in another formal system (Roqc, Isabelle, Lean3, HOL, etc.).
 - `@[category test]` : a statement that serves as a sanity check (e.g. for a new definition).
 - `@[category API]` : a statement that constructs basic theory around a new definition
 
@@ -94,6 +96,21 @@ in the AMS classification for completeness. Some are not relevant to this reposi
 
 namespace ProblemAttributes
 
+/-- The type of formal proof that exists for a problem. This is meant to be used
+for the `research formally solved` category. -/
+inductive FormalProofKind
+  /-- The problem exactly as stated in formal-conjectures has a formal proof.
+  The link points to a commit that fills the `sorry` relative to the current
+  commit (i.e., the commit where this category is added, or the commit with the
+  latest fix for this statement). -/
+  | formalConjecturesProof
+  /-- The problem is solved in Lean 4 (e.g. in Mathlib or some other
+  repository), perhaps as an equivalent statement. -/
+  | lean4
+  /-- The problem is formally solved in a different system (Roqc, Isabelle, Lean 3, HOL, etc.). -/
+  | otherSystem
+  deriving Inhabited, BEq, Hashable, ToExpr
+
 inductive ProblemStatus
   /-- Indicates that a mathematical problem is still open. -/
   | open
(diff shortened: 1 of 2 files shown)


mo271:
thanks, that looks great already, will look at it more closely later.

somewhat related, currently we aren't checking against
```
@[category research open, AMS 11, category research solved]
```
right? 
Perhaps we can somehow ensure that exactly one category is provided. With the new one it might be conceivable that people try to apply both `solved` and `formallly solved here`?

Paul-Lez:
Opened an issue for this #1852

FormalConjectures/Util/Attributes.lean
```diff
@@ -40,6 +40,9 @@ The values of this attribute are
   The criterion for being solved is that there exists an informal solution
   that is widely accepted by experts in the area. In particular, this
   does *not* require a formal solution to exist.
+- `@[category research formally solved here "link"]` : a research level math problem
```
YaelDillies:
```suggestion
- `@[category research formally solved there "link"]` : a research level math problem
```
or even
```suggestion
- `@[category research formally solved at "link"]` : a research level math problem
```
?

FormalConjectures/Util/Attributes.lean
```diff
@@ -97,15 +100,19 @@ inductive ProblemStatus
   /-- Indicates that a mathematical problem is already solved,
   i.e., there is a published (informal) proof that is widely accepted by experts. -/
   | solved
+  /-- Indicates that a mathematical problem is formally solved in this repository,
+  with a link to the formal proof. -/
+  | formallySolvedHere : String → ProblemStatus
```
YaelDillies:
```suggestion
  /-- Indicates that a mathematical problem is formally solved somewhere,
  with a link to the formal proof. -/
  | formallySolvedAt : String → ProblemStatus
```
no?

FormalConjectures/Util/AttributesTest.lean
```diff
@@ -46,3 +46,8 @@ theorem test_success_3 : 1 = 1 := by
 @[category high_school]
 theorem test_ignore_other_categories : 1 = 1 :=
   rfl
+
+#guard_msgs in
```
YaelDillies, line 50: What's this guarding?

FormalConjectures/Util/Attributes.lean
```diff
@@ -40,6 +40,9 @@ The values of this attribute are
   The criterion for being solved is that there exists an informal solution
   that is widely accepted by experts in the area. In particular, this
   does *not* require a formal solution to exist.
+- `@[category research formally solved at "link"]` : a research level math problem
+  that has been formally solved in this repository, with "link" being a permalink
+  to the formal solution.
```
mo271:
```suggestion
- `@[category research formally solved using <kind> at "link"]` : a formally solved research problem.
   `<kind>` must be one of:
   - `this_repo` : solved in this repository (e.g. filling the `sorry` in the current file).
   - `lean4` : solved in Lean 4 (e.g. Mathlib) as an equivalent statement.
   - `other_system` : solved in another system (Roqc, Isabelle, Lean3, HOL, etc.).
```
and then define the kind of formal Proofs somewhere, like
```lean
/-- The type of formal proof that exists for a problem. -/
inductive FormalProofKind
  /-- The problem exactly as stated in this repository has a formal proof.
  The link points to a commit that fills the `sorry` relative to the current
  commit (i.e., the commit where this category is added). -/
  | thisRepo
  /-- The problem is solved in Lean 4 (e.g. in Mathlib), perhaps as an equivalent statement. -/
  | lean4
  /-- The problem is formally solved in a different system (Roqc, Isabelle, Lean 3, HOL, etc.). -/
  | other_system
  deriving Inhabited, BEq, Hashable, ToExpr
```
because I think it makes sense to distinguish further:
 - points to a commit that adds fills the `sorry` exactly on top of the current commit
 - has a formal proof of Lean 4 of a equivalent statement
 - has some formal proof written in some language

mo271: (Could also be done in a follow-up, but I think this distinction should be made before we apply this category widely!)

FormalConjectures/Util/AttributesTest.lean
Paul-Lez: This is guarding the fact that there is _no message_

YaelDillies: That would be unnecessary if we had CI as tight as mathlib's. Future work ;)

FormalConjectures/Util/Attributes.lean
Paul-Lez: Done!

FormalConjectures/Util/Attributes.lean
Paul-Lez: Now fixed:)

PR #1867, opened 2026-01-23
feat(OEIS): A056777

### LINKED ISSUE #1453: math.CO/0409509 number 33
### What is the conjecture

$$
\begin{equation}\text{composite $n$} \wedge \phi(n+12)=\phi(n)+12 \wedge \sigma(n+12)=\sigma(n)+12 \iff
n\equiv64\bmod72.\end{equation}
$$

### Prerequisites needed
Status: open

### Choose either option

rwst:
Resolves #1453. The proof seems doable on undergraduate level.

### DIFF
--- a/FormalConjectures/OEIS/56777.lean
+++ b/FormalConjectures/OEIS/56777.lean
@@ -16,6 +16,7 @@ limitations under the License.
 import FormalConjectures.Util.ProblemImports
 
 open scoped ArithmeticFunction
+open Nat
 
 /-!
 # Conjectures associated with A056777
@@ -34,7 +35,7 @@ namespace OeisA056777
 /-- A composite number $n$ is in the sequence A056777 if it satisfies both
 $\varphi(n+12) = \varphi(n) + 12$ and $\sigma(n+12) = \sigma(n) + 12$. -/
 def a (n : ℕ) : Prop :=
-  ¬n.Prime ∧ 1 < n ∧ Nat.totient (n + 12) = Nat.totient n + 12 ∧ σ 1 (n + 12) = σ 1 n + 12
+  ¬n.Prime ∧ 1 < n ∧ totient (n + 12) = totient n + 12 ∧ σ 1 (n + 12) = σ 1 n + 12
 
 /-- A number $n$ comes from a prime quadruple $(p, p+2, p+6, p+8)$ if
 $n = p(p+8)$ for some prime $p$ where $p$, $p+2$, $p+6$, $p+8$ are all prime. -/
@@ -46,7 +47,7 @@ def ComesFromPrimeQuadruple (n : ℕ) : Prop :=
 theorem a_65 : a 65 := by
   refine ⟨?_, by norm_num, ?_, ?_⟩
   · simp only [show (65 : ℕ) = 5 * 13 by norm_num]
-    exact Nat.not_prime_mul (by norm_num) (by norm_num)
+    exact not_prime_mul (by norm_num) (by norm_num)
   · decide
   · decide
 
@@ -56,29 +57,29 @@ theorem a_209 : a 209 := by
   set_option maxRecDepth 1000 in
   refine ⟨?_, by norm_num, ?_, ?_⟩
   · simp only [show (209 : ℕ) = 11 * 19 by norm_num]
-    exact Nat.not_prime_mul (by norm_num) (by norm_num)
+    exact not_prime_mul (by norm_num) (by norm_num)
   · decide
   · decide
 
 /-- Numbers coming from prime quadruples are in the sequence A056777. -/
 @[category undergraduate, AMS 11]
-theorem a_of_comesFromPrimeQuadruple {n : ℕ} (h : comesFromPrimeQuadruple n) : a n := by
+theorem a_of_comesFromPrimeQuadruple {n : ℕ} (h : ComesFromPrimeQuadruple n) : a n := by
   sorry
 
 /-- All members of the sequence A056777 come from prime quadruples. -/
 @[category research open, AMS 11]
-theorem comesFromPrimeQuadruple_of_a {n : ℕ} (h : a n) : comesFromPrimeQuadruple n := by
+theorem comesFromPrimeQuadruple_of_a {n : ℕ} (h : a n) : ComesFromPrimeQuadruple n := by
   sorry
 
 /-- Numbers coming from prime quadruples satisfy $n \equiv 65 \pmod{72}$. -/
 @[category undergraduate, AMS 11]
-theorem mod_72_of_comesFromPrimeQuadruple {n : ℕ} (h : comesFromPrimeQuadruple n) :
+theorem mod_72_of_comesFromPrimeQuadruple {n : ℕ} (h : ComesFromPrimeQuadruple n) :
     n % 72 = 65 := by
   sorry
 
 /-- Numbers coming from prime quadruples satisfy $n \equiv 9 \pmod{100}$. -/
 @[category undergraduate, AMS 11]
-theorem mod_100_of_comesFromPrimeQuadruple {n : ℕ} (h : comesFromPrimeQuadruple n) :
+theorem mod_100_of_comesFromPrimeQuadruple {n : ℕ} (h : ComesFromPrimeQuadruple n) :
     n % 100 = 9 := by
   sorry
 


FormalConjectures/OEIS/056777.lean
```diff
@@ -0,0 +1,45 @@
...
+- φ(n+12) = φ(n) + 12  (Euler's totient)
+- σ(n+12) = σ(n) + 12  (sum of divisors)
+
+then n ≡ 65 (mod 72).
+
+This conjecture, posed by R. Stephan, probably requires meticulous congruence case analysis
+of the φ and σ functions modulo 8 and 9.
+
+*References:*
+    * https://oeis.org/A056777
```
Paul-Lez: Let's follow the usual markdown style from other files
```suggestion
*References:* [oeis.org/A056777](https://oeis.org/A056777)
```

FormalConjectures/OEIS/056777.lean
```diff
@@ -0,0 +1,45 @@
...
+/-!
+# **Conjecture: A056777**:
+If n is composite and satisfies both:
+- φ(n+12) = φ(n) + 12  (Euler's totient)
+- σ(n+12) = σ(n) + 12  (sum of divisors)
+
+then n ≡ 65 (mod 72).
+
+This conjecture, posed by R. Stephan, probably requires meticulous congruence case analysis
+of the φ and σ functions modulo 8 and 9.
```
Paul-Lez:
```suggestion
```

FormalConjectures/OEIS/056777.lean
```diff
@@ -0,0 +1,45 @@
...
+
+*References:*
+    * https://oeis.org/A056777
+-/
+
+namespace OeisA056777
+
+@[category undergraduate, AMS 11]
```
Paul-Lez, line 75: This tag is reserved to already solved problems - if this is solved, could you provide a reference?

FormalConjectures/OEIS/56777.lean
```diff
@@ -0,0 +1,45 @@
...
+
+This conjecture, posed by R. Stephan, probably requires meticulous congruence case analysis
+of the φ and σ functions modulo 8 and 9.
+
+*References:*
+    * https://oeis.org/A056777
+-/
+
+namespace OeisA056777
+
```
Paul-Lez, line 34: Let's also add an informal docstring here

FormalConjectures/OEIS/056777.lean
rwst: It is not. I stopped Claude before it went deeper.

felixpernegger:
image
This seems to be the more interesting cnjecture on the webiste. This most likely also easily implies the other conjecture

FormalConjectures/OEIS/56777.lean
mo271, line 1: A few remarks:

- please define the sequence (or the Prop on Nat for the sequence separately and then use it:
```lean
def a (n : ℕ) : Prop := (¬ n.Prime ∧ 1 < n) ∧
  Nat.totient (n + 12) = Nat.totient n + 12 ∧
  σ 1 (n + 12) = σ 1 n + 12
```
fine to use `a` in the namespace. Then let's perhaps also add some test statement checking that this gets the values 65 and 209 right?

- use latex markdown in docstrings througout.

- let's not repeat the conjecture in two place, only at the theorem, not the main docstring

 -  let's add a predicate for the other (related conjecture)
```lean
def comesFromPrimeQuadruple (n : ℕ) : Prop := ∃ p,
  p.Prime ∧ (p + 2).Prime ∧ (p + 6).Prime ∧ (p + 8).Prime ∧ n = p * (p + 8)
```
(with docstring)

then then theorems like
```lean
theorem a_of_comesFromPrimeQuadruple (n : ℕ) : comesFromPrimeQuadruple n → a n := by
```
(should be easy to proof (`undergraduate`) and its converse (should be harder to prove)

as well as the implication
```lean
 theorem conjecture_for_prime_quadruples (n : ℕ) : comesFromPrimeQuadruple n → n % 72 = 65 := by
```
 as tagged `undergraduate` since this is also not hard to see.

I guess we could also make up other conjectures like `n % 100 = 9` and `n%200 = 9`, both of which can be proved for $n>65$ when assuming `comesFromPrimeQuadruple n`. In that case we also might have  `n % 1800 = 209`.

rwst: Done. Should any further (undergrad) proofs be added?

FormalConjectures/OEIS/56777.lean
```diff
@@ -0,0 +1,90 @@
...
+/-- Numbers coming from prime quadruples satisfy $n \equiv 9 \pmod{100}$. -/
+@[category undergraduate, AMS 11]
+theorem mod_100_of_comesFromPrimeQuadruple (n : ℕ) (h : comesFromPrimeQuadruple n) :
+    n % 100 = 9 := by
+  sorry
+
+/-- All members of the sequence satisfy $n \equiv 65 \pmod{72}$. -/
+@[category research open, AMS 11]
+theorem mod_72_of_a (n : ℕ) (h : a n) : n % 72 = 65 := by
+  sorry
```
felixpernegger: Change this into the same format as ```mod_100_of_comesFromPrimeQuadruple``` as well

rwst: Not sure, can you be more specific?

felixpernegger: also change it into undergraduate and formulate it as an implication from the main conjecture

felixpernegger: just like the other theorem/conjecture (the 100 mod thing)

FormalConjectures/OEIS/56777.lean
```diff
@@ -0,0 +1,90 @@
...
+@[category test, AMS 11]
+theorem a_209 : a 209 := by
+  refine ⟨?_, by norm_num, ?_, ?_⟩
+  · native_decide
+  · native_decide
+  · native_decide
+
+/-- Numbers coming from prime quadruples are in the sequence A056777. -/
+@[category undergraduate, AMS 11]
+theorem a_of_comesFromPrimeQuadruple (n : ℕ) (h : comesFromPrimeQuadruple n) : a n := by
```
felixpernegger:
```suggestion
theorem a_of_comesFromPrimeQuadruple {n : ℕ} (h : comesFromPrimeQuadruple n) : a n := by
```
Similar most of the other theorems.
Generally, if ```n``` can be inferrd from later statements use { } instead of ()

FormalConjectures/OEIS/56777.lean
```diff
@@ -0,0 +1,90 @@
...
+  sorry
+
+/-- All members of the sequence A056777 come from prime quadruples. -/
+@[category research open, AMS 11]
+theorem comesFromPrimeQuadruple_of_a (n : ℕ) (h : a n) : comesFromPrimeQuadruple n := by
```
felixpernegger: Phrase is with a forall quantifier and use the answer(sorry) format

mo271: not needed here, since it it not a phrased as a question!

FormalConjectures/OEIS/56777.lean
mo271: > also change it into undergraduate and formulate it as an implication from the main conjecture

but it could be true without going through the other conjecture, no?

FormalConjectures/OEIS/56777.lean
```diff
@@ -0,0 +1,90 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+import FormalConjectures.Util.ProblemImports
+
+open scoped ArithmeticFunction
+
+/-!
+# **Conjectures associated with A056777**
```
YaelDillies: No need for bold here. This is already a title
```suggestion
# Conjectures associated with A056777
```

FormalConjectures/OEIS/56777.lean
```diff
@@ -0,0 +1,90 @@
...
+  · native_decide
+  · native_decide
+
+/-- $209$ is in the sequence A056777. -/
+@[category test, AMS 11]
+theorem a_209 : a 209 := by
+  refine ⟨?_, by norm_num, ?_, ?_⟩
+  · native_decide
+  · native_decide
+  · native_decide
```
YaelDillies:
```suggestion
  refine ⟨?_, by norm_num, ?_⟩ <;> native_decide
```

FormalConjectures/OEIS/56777.lean
rwst: It is redundant. I removed it.

felixpernegger: > > also change it into undergraduate and formulate it as an implication from the main conjecture
> 
> but it could be true without going through the other conjecture, no?

in theory I guess yeah, but it's more of an afterthought, since the ifrst conjecture probably holds. But why not I suppose

rwst:
-awaiting-author

rwst:
awaiting-review

felixpernegger:
By the way way great your are adding your own conjectures from so long ago :)

FormalConjectures/OEIS/56777.lean
```diff
@@ -0,0 +1,82 @@
...
+  ∃ p : ℕ, p.Prime ∧ (p + 2).Prime ∧ (p + 6).Prime ∧ (p + 8).Prime ∧
+    n = p * (p + 8)
+
+/-- $65$ is in the sequence A056777. -/
+@[category test, AMS 11]
+theorem a_65 : a 65 := by
+  refine ⟨?_, by norm_num, ?_, ?_⟩
+  · native_decide
+  · native_decide
+  · native_decide
```
felixpernegger: Don't use ```native_decide```

FormalConjectures/OEIS/56777.lean
```diff
@@ -0,0 +1,82 @@
...
+
+/-- $209$ is in the sequence A056777. -/
+@[category test, AMS 11]
+theorem a_209 : a 209 := by
+  refine ⟨?_, by norm_num, ?_⟩ <;> native_decide
```
felixpernegger: Don't use ```native_decide```

rwst:
The fact that these were not proven in 20 years should make them valuable, I thought, so don't lose them.

FormalConjectures/OEIS/56777.lean
```diff
@@ -0,0 +1,88 @@
...
+def a (n : ℕ) : Prop :=
+  ¬n.Prime ∧ 1 < n ∧
+  Nat.totient (n + 12) = Nat.totient n + 12 ∧
+  σ 1 (n + 12) = σ 1 n + 12
+
+/-- A number $n$ comes from a prime quadruple $(p, p+2, p+6, p+8)$ if
+$n = p(p+8)$ for some prime $p$ where $p$, $p+2$, $p+6$, $p+8$ are all prime. -/
+def comesFromPrimeQuadruple (n : ℕ) : Prop :=
+  ∃ p : ℕ, p.Prime ∧ (p + 2).Prime ∧ (p + 6).Prime ∧ (p + 8).Prime ∧
+    n = p * (p + 8)
```
felixpernegger:
```suggestion
def a (n : ℕ) : Prop :=
  ¬n.Prime ∧ 1 < n ∧ Nat.totient (n + 12) = Nat.totient n + 12 ∧ σ 1 (n + 12) = σ 1 n + 12

/-- A number $n$ comes from a prime quadruple $(p, p+2, p+6, p+8)$ if
$n = p(p+8)$ for some prime $p$ where $p$, $p+2$, $p+6$, $p+8$ are all prime. -/
def comesFromPrimeQuadruple (n : ℕ) : Prop :=
  ∃ p : ℕ, p.Prime ∧ (p + 2).Prime ∧ (p + 6).Prime ∧ (p + 8).Prime ∧ n = p * (p + 8)
```

Also open the namespace ```Nat``` to be able to use \phi and so on

FormalConjectures/OEIS/56777.lean
```diff
@@ -0,0 +1,88 @@
...
+/-- A composite number $n$ is in the sequence A056777 if it satisfies both
+$\varphi(n+12) = \varphi(n) + 12$ and $\sigma(n+12) = \sigma(n) + 12$. -/
+def a (n : ℕ) : Prop :=
+  ¬n.Prime ∧ 1 < n ∧
+  Nat.totient (n + 12) = Nat.totient n + 12 ∧
+  σ 1 (n + 12) = σ 1 n + 12
+
+/-- A number $n$ comes from a prime quadruple $(p, p+2, p+6, p+8)$ if
+$n = p(p+8)$ for some prime $p$ where $p$, $p+2$, $p+6$, $p+8$ are all prime. -/
+def comesFromPrimeQuadruple (n : ℕ) : Prop :=
```
YaelDillies:
```suggestion
def ComesFromPrimeQuadruple (n : ℕ) : Prop :=
```

Issue #894, opened 2025-10-08
Green's Open Problems #6


rao107:
### What is the conjecture
https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.6

Fix an integer $d$. What is the largest sum-free subset of $[N]^d$?

### Choose either option
- [ ] I plan on adding this conjecture to the repository
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

cyb3r17:
@Paul-Lez are you working on this issue? If not I'd like to get started!

Paul-Lez:
@cyb3r17 I am - PR to arrive soon (I'll ping you if you fancy reviewing it:))

Sukumarsawant:
Is this still being worked on or already solved, since I'm looking for some good first issues to explore?

Paul-Lez:
@Sukumarsawant I'm still on this, but I'll open some extra good first issues if you want!

Sukumarsawant:
> [@Sukumarsawant](https://github.com/Sukumarsawant) I'm still on this, but I'll open some extra good first issues if you want!

Sure, thanks! I'm looking forward to work and explore those

Paul-Lez:
@Sukumarsawant I've added a `good-first-issue` tag to a few if you want to take a look!

PR #1882, opened 2026-01-24
Green's Open Problems #6 #894

Aman-codes0-0:
Formalizes Ben Green's Open Problem 6 in Lean.
The problem asks for the size of the largest sum-free subset of [N]^d (Box Power).


mo271:
issue is already assigend, let's avoid duplicate work

PR #1885, opened 2026-01-25
chore(GreensOpenProblems): Add aliases to open problems which are Erdos problems

felixpernegger:


mo271:
so far we haven't used `alias` at all, I believe. Also because that doesn't play well with our category attribute, but I suppose we could think about that more. 
What has been done so far:
- Files with just docstrings and imports, not even a theorem.  examples here:
```shell
grep -rl --include="*.lean" "import.*rdosProblem" .
./FormalConjectures/Wikipedia/BrocardProblem.lean
./FormalConjectures/Wikipedia/HappyEndingProblem.lean
./FormalConjectures/Wikipedia/SparseRuler.lean
./FormalConjectures/Wikipedia/HadwigerNelson.lean
./FormalConjectures/Wikipedia/MinimalOverlapProblem.lean
./FormalConjectures/ErdosProblems/40.lean
./FormalConjectures/GreensOpenProblems/7.lean
./FormalConjectures/GreensOpenProblems/81.lean
./FormalConjectures/GreensOpenProblems/45.lean
./FormalConjectures/GreensOpenProblems/63.lean
```
Sometimes using `type_of%` and then repeating the theorem with a different name:
Perhaps it is best to do it like it was done in `./FormalConjectures/ErdosProblems/1135.lean`?

felixpernegger:
> so far we haven't used `alias` at all, I believe. Also because that doesn't play well with our category attribute, but I suppose we could think about that more. What has been done so far:
> 
>     * Files with just docstrings and imports, not even a theorem.  examples here:
> 
> 
> ```shell
> grep -rl --include="*.lean" "import.*rdosProblem" .
> ./FormalConjectures/Wikipedia/BrocardProblem.lean
> ./FormalConjectures/Wikipedia/HappyEndingProblem.lean
> ./FormalConjectures/Wikipedia/SparseRuler.lean
> ./FormalConjectures/Wikipedia/HadwigerNelson.lean
> ./FormalConjectures/Wikipedia/MinimalOverlapProblem.lean
> ./FormalConjectures/ErdosProblems/40.lean
> ./FormalConjectures/GreensOpenProblems/7.lean
> ./FormalConjectures/GreensOpenProblems/81.lean
> ./FormalConjectures/GreensOpenProblems/45.lean
> ./FormalConjectures/GreensOpenProblems/63.lean
> ```
> 
> Sometimes using `type_of%` and then repeating the theorem with a different name: Perhaps it is best to do it like it was done in `./FormalConjectures/ErdosProblems/1135.lean`?

Yes I saw that, I thought just having a docstring in a file might be a bit confusing.
Using ```%type_of```` is certainly better, but it is still a little bit awkward to state the same theorem twice and it technically not being the same.

felixpernegger:
For anyone stumbling on this PR, also see
https://leanprover.zulipchat.com/#narrow/channel/524981-Formal-conjectures/topic/Problems.20that.20appear.20in.20two.20or.20more.20problem.20collection/with/570072889

mo271:
There was not much participation in that discussion, but instead of introducing a third way of doing this (`alias`), I propose to either 

1.  use `type_of%` together with the usual naming of theorem (preferred) or
2. just add docstring-only files, importing the other files, but other than that without theorems

felixpernegger:
Probably either way is fine, but we should be consistent.

Personally I would say
alias > type_of > just docstring

(As having just the docstring is a bit confusing)

I'll change this PR accordingly to type_of then

felixpernegger:
Changed format to ```type_of%``` now as discussed.

PR #1890, opened 2026-01-25
feat(ErdosProblems): 1139

danielchin:
Formalizing Erdos Problem 1139: https://www.erdosproblems.com/1139

### DIFF
--- a/FormalConjectures/ErdosProblems/1139.lean
+++ b/FormalConjectures/ErdosProblems/1139.lean
@@ -29,7 +29,7 @@ namespace Erdos1139
 
 /--
 Let $1\leq u_1 < u_2 < \cdots$ be the sequence of integers with at most $2$ prime factors.
-Is it true that \[\limsup_{k \to \infty} \frac{u_{k+1}-u_k}{\log k}=\infty?\]
+Is it true that $$\limsup_{k \to \infty} \frac{u_{k+1}-u_k}{\log k}=\infty?$$
 -/
 @[category research open, AMS 11]
 theorem erdos_1139 :


FormalConjectures/ErdosProblems/1139.lean
```diff
@@ -0,0 +1,41 @@
...
+namespace Erdos1139
+
+/--
+Let $1\leq u_1 < u_2 < \cdots$ be the sequence of integers with at most $2$ prime factors.
+Is it true that \[\limsup_{k \to \infty} \frac{u_{k+1}-u_k}{\log k}=\infty?\]
+-/
+@[category research open, AMS 11]
+theorem erdos_1139 :
+    answer(sorry) ↔
+      let u := Nat.nth (fun n ↦ 0 < n ∧ Ω n ≤ 2)
```
Paul-Lez:
```suggestion
      letI u := Nat.nth (fun n ↦ 0 < n ∧ Ω n ≤ 2)
```

FormalConjectures/ErdosProblems/1139.lean
```diff
@@ -0,0 +1,41 @@
...
+      atTop.limsup (fun k : ℕ ↦ (((u (k + 1) : ℝ) - (u k : ℝ)) / Real.log ↑k : EReal)) = ⊤ := by
```
Paul-Lez: I think the indexing may need to be shifted by 1 in some places here (in the denominator?): the `u` sequence in the informal version starts a `1` while in the formal version it starts at `0`.

FormalConjectures/ErdosProblems/1139.lean
danielchin: Done.

FormalConjectures/ErdosProblems/1139.lean
danielchin: Done.

FormalConjectures/ErdosProblems/1139.lean
```diff
@@ -0,0 +1,41 @@
...
+
+open Nat Filter
+open scoped ArithmeticFunction
+open scoped Topology
+
+namespace Erdos1139
+
+/--
+Let $1\leq u_1 < u_2 < \cdots$ be the sequence of integers with at most $2$ prime factors.
+Is it true that \[\limsup_{k \to \infty} \frac{u_{k+1}-u_k}{\log k}=\infty?\]
```
mo271:
```suggestion
Is it true that $$\limsup_{k \to \infty} \frac{u_{k+1}-u_k}{\log k}=\infty?$$
```

PR #1891, opened 2026-01-25
feat(ErdosProblems): 1141

danielchin:
Formalize Erdos Problem 1141: https://www.erdosproblems.com/1141

### DIFF
--- a/FormalConjectures/ErdosProblems/1141.lean
+++ b/FormalConjectures/ErdosProblems/1141.lean
@@ -36,21 +36,13 @@ The property that $n-k^2$ is prime for all $k$ with $(n,k)=1$ and $k^2 < n$.
 def Erdos1141Prop (n : ℕ) : Prop :=
   ∀ k, k ^ 2 < n → Coprime n k → (n - k ^ 2).Prime
 
-instance (n : ℕ) : Decidable (Erdos1141Prop n) := by
-  cases n with
-  | zero =>
-    apply isTrue
-    intro k h
-    contradiction
-  | succ n' =>
-    let bound := n'.sqrt
-    let P (k : ℕ) := Nat.Coprime (n' + 1) k → (n' + 1 - k ^ 2).Prime
-    have : DecidablePred P := inferInstance
-    apply decidable_of_iff (∀ k ∈ Finset.range (bound + 1), P k)
-    apply forall_congr'
-    intro k
-    dsimp [P, bound]
-    rw [Finset.mem_range, Nat.lt_succ_iff, Nat.lt_succ_iff, Nat.pow_two, ← Nat.le_sqrt]
+instance (n : ℕ) : Decidable (Erdos1141Prop n) :=
+  decidable_of_iff (∀ k ≤ .sqrt (n - 1), Coprime n k → (n - k ^ 2).Prime) <| by
+    cases n with
+    | zero => simp [Erdos1141Prop]
+    | succ n' =>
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/1141.lean
```diff
@@ -0,0 +1,47 @@
...
+- [erdosproblems.com/1141](https://www.erdosproblems.com/1141)
+- [A214583](https://oeis.org/A214583)
+- [Va99] Various, Some of Paul's favorite problems. Booklet produced for the conference "Paul Erdős
+  and his mathematics", Budapest, July 1999 (1999).
+-/
+
+open Nat Set
+
+namespace Erdos1141
+
```
Paul-Lez, line 32: I'd suggest the following: extract the predicate on `n` as a definition, and add two test statements for e.g. 968 and 1722.

danielchin: Split out the property and added the two cases for 968 and 1722. Let me know if using `example` is the correct method for listing specific test cases in these files.

I left the two cases as `sorry` because they were failing to compile with `native_decide` and `decide` due to `maximum recursion depth has been reached`.

FormalConjectures/ErdosProblems/1141.lean
```diff
@@ -0,0 +1,61 @@
...
+/--
+The property that $n-k^2$ is prime for all $k$ with $(n,k)=1$ and $k^2 < n$.
+-/
+def erdos_1141_property (n : ℕ) : Prop :=
```
YaelDillies:
```suggestion
def Erdos1141Prop (n : ℕ) : Prop :=
```

FormalConjectures/ErdosProblems/1141.lean
```diff
@@ -0,0 +1,61 @@
...
+    answer(sorry) ↔ Infinite { n | erdos_1141_property n } := by
+  sorry
+
+@[category research solved, AMS 11]
+example : ¬ erdos_1141_property 968 := by
+  sorry
+
+@[category research solved, AMS 11]
+example : erdos_1141_property 1722 := by
+  sorry
```
YaelDillies: Can `decide` or `decide +native` do these if you provide a `DecidablePred Erdos1141Prop` instance based on the fact that `Erdos1141Prop n` is equivalent to `∀ k \le .sqrt (n - 1) → Coprime n k → (n - k ^ 2).Prime` (cf https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Nat/Sqrt.html#Nat.le_sqrt)?

FormalConjectures/ErdosProblems/1141.lean
danielchin: Done.

FormalConjectures/ErdosProblems/1141.lean
danielchin: I think I got what you were implying. Let me know if this implementation is what you were thinking of.

FormalConjectures/ErdosProblems/1141.lean
danielchin: Done.

YaelDillies (approved):
Great, thanks!

FormalConjectures/ErdosProblems/1141.lean
YaelDillies: Perfect score for you :)

PR #1894, opened 2026-01-25
feat(OEIS): A087719

### LINKED ISSUE #1460: math.CO/0409509 number 46
### What is the conjecture
Define $\varsigma(n)$ the smallest prime factor of $n$. Let $a_n$ the least number such that the number of numbers $k\le a_n$ with $k>\varsigma(k)^n$ exceeds the number of numbers with $k\le\varsigma(k)^n$. Then $a_n = 3^n + 3\cdot2^n + 6.$

### Prerequisites needed

Status: open

### Choose either option

rwst:
Resolves #1460.

# Conjectures associated with A087719

Define $\varsigma(n)$ the smallest prime factor of $n$ (`Nat.minFac`). Let $a_n$ be the least
number such that the count of numbers $k \le a_n$ with $k > \varsigma(k)^n$ exceeds the count
of numbers with $k \le \varsigma(k)^n$.

The conjecture states that $a_n = 3^n + 3 \cdot 2^n + 6$ for $n \ge 1$.

*References:* [oeis.org/A087719](https://oeis.org/A087719)

### DIFF
--- a/FormalConjectures/OEIS/87719.lean
+++ b/FormalConjectures/OEIS/87719.lean
@@ -72,9 +72,6 @@ theorem a_three : a 3 = 57 := by
   intro m hm
   interval_cases m <;> decide
 
-/-- The formula value. -/
-def formula (n : ℕ) : ℕ := 3 ^ n + 3 * 2 ^ n + 6
-
 /-- Conjecture: a(n) = 3^n + 3 * 2^n + 6 for n ≥ 1. -/
 @[category research formally solved using lean4 at "https://github.com/google-deepmind/formal-conjectures/pull/1894/commits/7a286754f623759d69a3dd18f482c53c1d70959b", AMS 11]
 theorem a_formula {n : ℕ} (hn : n ≥ 1) : a n = 3 ^ n + 3 * 2 ^ n + 6 := by


FormalConjectures/OEIS/87719.lean
```diff
@@ -0,0 +1,80 @@
...
+  (Finset.Icc 1 m).filter (fun k => k ≤ k.minFac ^ n) |>.card
+
+/-- There exists m such that countExceeding n m > countNotExceeding n m. -/
+@[category research open, AMS 11]
+theorem a_exists {n : ℕ} : ∃ m, countExceeding n m > countNotExceeding n m := by
+    sorry
+
+/-- The sequence a(n): least m such that countExceeding n m > countNotExceeding n m. -/
+noncomputable def a (n : ℕ) : ℕ :=
+  Nat.find (@a_exists n)
```
Paul-Lez: Is `a_exists` really open?
```suggestion
/-- There exists m such that countExceeding n m > countNotExceeding n m. -/
@[category research open, AMS 11]
theorem a_exists (n : ℕ) : ∃ m, countExceeding n m > countNotExceeding n m := by
    sorry

/-- The sequence a(n): least m such that countExceeding n m > countNotExceeding n m. -/
noncomputable def a (n : ℕ) : ℕ :=
  Nat.find (a_exists n)
```

rwst: I'll try to give a proof for `a_exists`.

Paul-Lez: Note: you don't necessarily have to prove `a_exists` formally, but if this is an easy fact / already known fact then the `category research open` tag isn't the right choice.

rwst: I changed the category, thanks.

FormalConjectures/OEIS/87719.lean
```diff
@@ -0,0 +1,80 @@
...
+  intro m hm
+  interval_cases m <;> decide
+
+/-- a(3) = 57. -/
+@[category test, AMS 11]
+theorem a_three : a 3 = 57 := by
+  rw [a, Nat.find_eq_iff]
+  refine ⟨by decide, ?_⟩
+  intro m hm
+  interval_cases m <;> decide
```
mo271, line 73:
```suggestion
  unfold a
  norm_num +decide only [Nat.find_eq_iff]
```
same for the other analogous proofs

rwst: I'm currently trying to steer Claude for a proof of the main theorem. I guess, if this succeeds then this PR is no longer of interest?

rwst: Okay, I have an ungolfed 500-line proof which eventually will be published as preprint on zenodo. What to do with this PR?

mo271: Fantastic, this is great! Let's keep the PR and merge it when ready, no problem that is not contains a `research solved` theorem instead of a `research open` conjecture. I propose to 
 - create an intermediary commit, adding the unfolfed 500-line proof
 - delete that proof in the next commit and use our new `formally solved` tag to link to the commit with the ungolfed proof: https://github.com/google-deepmind/formal-conjectures/pull/1851
  - then update the docstring/references section once the preprint is out!
  
  
  Nice to hear that writing up the proof formally lead to its solution, even before it was added to our repo!

rwst: I'm however not sure about the `research` label, exactly because this was proved by LLM. I'm not publishing because of its importance, but just to have a DOI to refer to.

mo271: There have been some examples recently of research problems, i.e. problems that were mentioned in a research paper, which have been solved with the help of AI.
If you think this is rather high school level or graduate or whatever, we can just use this label. 
Perhaps we should expand our way to track formal proofs to all categories then. But as long as we don't have this it is also fine to provide the link to the proof in the docstring. 
In any case, I think it is still nice to have this conjecture/theorem in our repo, thanks for making the effort formalising it. 

Also: perhaps we can also add the mentioned generating function from this OEIS entry.

rwst: The g.f. simply reflects the formula. However, to add it would need some power series machinery, which I was planning to add for years already.

mo271: yeah, that makes sense. Perhaps the proof become easier using the g.f. 
But it is also fine not to add it here. 

I had thought everything is already in Mathlib, to at least state the power series identity? Using 
`PowerSeries.coeff ℤ` or whatever. What machinery is missing?

rwst: I think it's more the combinatorial proofs that benefit from a g.f. What is missing are the lemmas to prove a power series of a rational function (like the g.f. here) have coefficients of form $\sum_i c_i*a_i^n$, with the $a_i$ the zeros of the function denominator.

rwst: Finally, I refrained from writing that preprint and just added a proof sketch to the OEIS page.

FormalConjectures/OEIS/87719.lean
```diff
@@ -0,0 +1,83 @@
...
+
+/-- The formula value. -/
+def formula (n : ℕ) : ℕ := 3 ^ n + 3 * 2 ^ n + 6
```
mo271:
```suggestion
```

This not used anywhere, right?

I think it is better to inline this

mo271:
Thanks again!

PR #1895, opened 2026-01-26
chore: Remove zero padding for OEIS

felixpernegger:
per #1671


felixpernegger:
problem with compilation

FormalConjectures/GreensOpenProblems/58.lean
mo271, line 1: something went wrong here?

mo271:
I didn't remove the zero in docstrings and links in #1671, because I suppose on the OEIS the default way of naming them, but I didn't think it was a good idea to count zeros in file names and namespaces, but I suppose doing the same in the docstrings and links is more consisten..

FormalConjectures/GreensOpenProblems/58.lean
felixpernegger: my bad

felixpernegger:
in any case A81091 and A056777 still have wrong namespaces (outside of this PR)

PR #1896, opened 2026-01-26
chore: bump mathlib to v4.27.0

YaelDillies:


eric-wieser:
Does this build at every intermediate commit? Note that currently nothing is green.

YaelDillies:
`lake build` succeeded at v4.23, v4.26, v4.27. For v4.24, v4.25, I believe it succeeds as well but I haven't tried because I found the relevant fixes only after building to the version after (and subsequently hid the changes in the corresponding commits with a rebase). CI failing is due to the change in the `String` API, which I unfortunately don't know much about besides the fact that it is very disruptive...

Would you want `lake build` to succeed at each stage, or the whole CI? The latter promises to be annoying since the `String` API changed at least twice between v4.22 and v4.27.

eric-wieser:
I suppose all that really matters is that `lake build` succeeds at every stage, and the whole CI succeeds at the final stage. Let's see what @mo271 thinks.

mo271:
as @eric-wieser said, only `lake build` at every step (unless that would be an unreasonable amount of effort for some reason) and CI at the last step

FormalConjectures/Util/ProblemImports.lean
```diff
@@ -15,6 +15,7 @@ limitations under the License.
 -/
 
 -- A standard set of imports for open problems.
+import Counterexamples
```
eric-wieser: Why was this added?

Paul-Lez:
@YaelDillies what's the issue with the String API? (I suspect this might be showing up in code i wrote, so happy to look into this)

FormalConjectures/Util/ProblemImports.lean
YaelDillies: [`motzkin_polynomial_nonneg`](https://leanprover-community.github.io/mathlib4_docs/Counterexamples/Motzkin.html#motzkin_polynomial_nonneg) was moved to Counterexamples

eric-wieser: I'm not sure any care has been taken to make `Counterexamples`s reasonable to import (besides the fact that doc-gen checks names don't clash with mathlib); I'd be more comfortable just sorrying out

https://github.com/google-deepmind/formal-conjectures/blob/f01550cf4f910c7018f593f47979ab1cc9e8c70e/FormalConjectures/HilbertProblems/17.lean#L52-L56

eric-wieser: (or inlining the short proof with a comment)

YaelDillies:
I now locally have a sequence of bump commits that succeed on `lake build` over latest main, but Eric asked me privately to hold off pushing it.

@Paul-Lez, the issue with strings is that the substring API changed completely and most of the functions are now deprecated. Could you have a look at the CI failure and address it? Feel free to push to this branch, I can deal with the conflicts with my local version.

Paul-Lez:
@YaelDillies Sounds good, will take a look!

mo271:
@YaelDillies can you rebase this on current HEAD?

eric-wieser:
I don't think rebasing is worth the trouble, let's do a regular merge.

mo271:
We can also merge instead of rebase (but not squash).
Do the intermediate versions in this pull request currently `lake build`? From https://github.com/google-deepmind/formal-conjectures/pull/1896#issuecomment-3803395014 it seems like it currently might not, @YaelDillies?

I think we should whatever is easiest to get a main branch that contains commits where `lake build` works on the bumps to v4.23 through v4.26 and CI passes at v4.27.

YaelDillies:
Yes, all intermediate versions pass `lake build`

PR #1897, opened 2026-01-26
feat: add Green's Open Problem 82

### LINKED ISSUE #1737: Green's Open Problems #82
### What is the conjecture
https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.82

Let $A \subset \mathbb{Z}$ be a set of size $n$. For how many $\theta \in \mathbb{R}/\mathbb{Z}$ must we have $\sum\_{a \in A} \cos(a\theta) = 0$?

* ams-11
* ams-42

### Choose either option

krrish175-byte:
Closes #1737

This PR formalizes Green's Open Problem 82, which asks: for a finite set A ⊂ ℤ of size n, what is the minimum number of zeros that ∑_{a∈A} cos(2πaθ) must have on ℝ/ℤ?


FormalConjectures/GreensOpenProblems/82.lean
```diff
@@ -0,0 +1,52 @@
...
+/--
+Let $A \subset \mathbb{Z}$ be a set of size $n$. For how many $\theta \in \mathbb{R}/\mathbb{Z}$
+must we have $\sum_{a \in A} \cos(2\pi a\theta) = 0$?
+
+We formalize this as asking for the infimum of the number of zeros of
+$\theta \mapsto \sum_{a \in A} \cos(2\pi a \theta)$ in $[0, 1)$, over all finite sets
+$A \subset \mathbb{Z}$ with $|A| = n$.
+-/
+@[category research open, AMS 11 42]
+theorem green_82 (n : ℕ) (hn : 1 ≤ n) :
```
felixpernegger: ```n``` probably shouldnt be in the context as usual

FormalConjectures/GreensOpenProblems/82.lean
```diff
@@ -0,0 +1,52 @@
...
+    ⨅ (A : Finset ℤ) (_ : A.card = n),
+      ({θ : ℝ | θ ∈ Ico 0 1 ∧ ∑ a ∈ A, cos (2 * π * a * θ) = 0} : Set ℝ).ncard =
```
felixpernegger: Better write this as an inequality and not take sup?

felixpernegger:
The green problem list mentions known bounds, please include them.

FormalConjectures/GreensOpenProblems/82.lean
felixpernegger: This is not what the problem is asking. Why do you have to limit $\theta$ to ```Ico 0 1```?

krrish175-byte:
Thanks for the review. Sorry, I missed those. I have updated the code as you asked for:

- Moved n inside the theorem statement.
- Reformulated the statement using an inequality.
- Added the known lower bound $\Omega(\log n / \log \log n)$.

Regarding Ico 0 1: Since the trigonometric polynomial is 1-periodic, counting zeros in the fundamental domain [0, 1) seemed like the most direct way to represent the count on R / Z. Let me know if you would prefer using UnitAddCircle instead. Thanks for the review again!

felixpernegger:
> Thanks for the review. Sorry, I missed those. I have updated the code as you asked for:
> 
>     * Moved n inside the theorem statement.
> 
>     * Reformulated the statement using an inequality.
> 
>     * Added the known lower bound 
>         Ω
>         (
>         log
>         ⁡
>         n
>         
>           /
>         
>         log
>         ⁡
>         log
>         ⁡
>         n
>         )
>       .
> 
> 
> Regarding Ico 0 1: Since the trigonometric polynomial is 1-periodic, counting zeros in the fundamental domain [0, 1) seemed like the most direct way to represent the count on R / Z. Let me know if you would prefer using UnitAddCircle instead. Thanks for the review again!

Thank you ChatGPT. I mistook what R / Z meant in this context, yeah.

FormalConjectures/GreensOpenProblems/82.lean
```diff
@@ -0,0 +1,54 @@
...
+    answer(sorry) ↔ ∃ f : ℕ → ℕ, ∀ n > 0, ∀ A : Finset ℤ, A.card = n →
+      f n ≤ ({θ : ℝ | θ ∈ Ico 0 1 ∧ ∑ a ∈ A, cos (2 * π * a * θ) = 0} : Set ℝ).ncard := by
+  sorry
+
+/-- There must be at least $\Omega(\log n / \log \log n)$ zeros. -/
+@[category research solved, AMS 11 42]
+theorem green_82_lower_bound :
+    ∃ c > 0, ∀ n ≥ 2, ∀ A : Finset ℤ, A.card = n →
+      c * log n / log (log n) ≤
+        ({θ : ℝ | θ ∈ Ico 0 1 ∧ ∑ a ∈ A, cos (2 * π * a * θ) = 0} : Set ℝ).ncard := by
```
felixpernegger: Where did you get this estimate from? Its not mentioned on the website....
If you include known results, also make sure to reference them properly.

In theory, it is fine if you use AI assistance (sorry if im mistaken, but im pretty sure), but it is crucial you check the output.

On a nother note, there are also bounds in the other direction as mentioned in the problem list, which should be included as well.

felixpernegger (dismissed):

FormalConjectures/GreensOpenProblems/82.lean
krrish175-byte: I really really apologize for the confusion.

I went back to the original problem statement in Green's paper and I see now that I misread/misunderstood the bounds mentioned. The paper doesn't actually specify concrete bounds for Problem 82, so I shouldn't have included that theorem. I accept its my mistake.

Regarding the Ico 0 1 restriction, you make a good point. I was thinking about the periodicity, but the problem is really asking about the minimum number of zeros as we vary over all possible finite sets A, which is indeed a different question than counting in a fundamental domain.

I will update the PR to remove that lower bound theorem entirely, reformulate without the [0,1) restriction, keep it as an inequality as you suggested

Thanks again for the review, this is my first contribution to this project and I'm still learning the conventions so I request your patience and assistance. Thanks again...

felixpernegger: The Ico thing should be fine though.

krrish175-byte: Ah got it. Thanks for clarifying. Then I will keep the Ico 0 1 restriction then. So before I push any changes, can you please confirm if this approach is right or you need any more changes : 

1. Ico 0 1 kept (as you just confirmed)
2. The incorrect lower bound removed 
3. n ≥ 2 instead of n ≥ 1
4. The upper bound n^{5/6} included

let me know if there's anything else that needs adjustment

felixpernegger: According to the problem list, n^ 5/6 isnt the best known bound anymore.

I dont know why you need 1 leq n or 2 leq n. Everything is in Landau O notation, so this shouldnt matter.

There are known results both on the upper and lower bound. A complete foramilasation ought to nention both of them.

felixpernegger: According to the problem list, n^ 5/6 isnt the best known bound anymore.

I dont know why you need 1 leq n or 2 leq n. Everything is in Landau O notation, so this shouldnt matter.

There are known results both on the upper and lower bound. A complete foramilasation ought to nention both of them.

krrish175-byte: I have updated the code with the current best known bounds:

Lower bound: (log log n)^c for sufficiently large n
- wReference: Bedert (2025), arXiv:2407.16075

Upper bound: C(n log n)^(2/3) 
- Reference: Juškevičius & Sahasrabudhe (2020), arXiv:2005.01695

I've formalized both bounds as separate theorems as you suggested. The lower bound uses ∃ n0 to handle the asymptotic nature properly (since log log n isn't well-defined for small n).

Let me know if it is correct now.

FormalConjectures/GreensOpenProblems/82.lean
```diff
@@ -0,0 +1,71 @@
...
+must we have $\sum_{a \in A} \cos(2\pi a\theta) = 0$?
+
+It is known that there are sets $A$ for which the number of zeros is at most $C(n \log n)^{2/3}$
+(Juškevičius & Sahasrabudhe, 2020). For $n \ge 2$, every set $A$ has at least $(\log \log n)^c$
+zeros (Bedert, 2025).
+
+*Reference:*
+- [Gr24] [Ben Green's Open Problem 82](https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.82)
+- Benjamin Bedert, "An improved lower bound for a problem of Littlewood on the zeros of cosine polynomials", arXiv:2407.16075 (2025)
+- Juškevičius & Sahasrabudhe, "Cosine polynomials with few zeros", arXiv:2005.01695 (2020)
```
mo271: make those links, please 
look at some other files how we usually do it

FormalConjectures/GreensOpenProblems/82.lean
```diff
@@ -0,0 +1,71 @@
...
+
+/-!
+# Ben Green's Open Problem 82
+
+Let $A \subset \mathbb{Z}$ be a set of size $n$. For how many $\theta \in \mathbb{R}/\mathbb{Z}$
+must we have $\sum_{a \in A} \cos(2\pi a\theta) = 0$?
+
+It is known that there are sets $A$ for which the number of zeros is at most $C(n \log n)^{2/3}$
+(Juškevičius & Sahasrabudhe, 2020). For $n \ge 2$, every set $A$ has at least $(\log \log n)^c$
+zeros (Bedert, 2025).
```
mo271:
```suggestion
```

Let's not repeat docstrings from below here

FormalConjectures/GreensOpenProblems/82.lean
```diff
@@ -0,0 +1,71 @@
...
+-/
+
+open Real Set
+open scoped Finset
+
+namespace Green82
+
+/--
+Let $A \subset \mathbb{Z}$ be a set of size $n$. For how many $\theta \in \mathbb{R}/\mathbb{Z}$
+must we have $\sum_{a \in A} \cos(2\pi a\theta) = 0$?
```
mo271: I don't quite see how "For how many" is formalised here. Is this meant asymptocially or should we use the `answer(sorry)` mechanism?

can we just take `f := fun n => 0` here and it will be trivially true?

mo271 (commented):
@krrish175-byte are you confident that the formalisation reflects the informal statements? Do you understand the problem?

krrish175-byte:
@mo271 : I believe I understand it as asking for the minimum guaranteed zeros across all sets, but please correct me if I am off! Also I addressed all the reviews. Please check it now. Thanks!

mo271:
the statement of `green_82` still doesn't make sense. 
Also:
Likely parenthesization bug in the lower bound: Real.log (Real.log n) ^ c probably parses as Real.log ((Real.log n) ^ c) = $\log((\log n)^c)$, not the intended $(\log \log n)^c$. Should be (Real.log (Real.log n)) ^ c.

Paul-Lez (approved):

FormalConjectures/GreensOpenProblems/82.lean
```diff
@@ -0,0 +1,62 @@
...
+/-- The minimum number of zeros in $[0,1)$ of $\sum_{a \in A} \cos(2\pi a\theta)$
+over all sets $A \subset \mathbb{Z}$ of size $n$. -/
+noncomputable def minZeros (n : ℕ+) : ℕ∞ :=
+  ⨅ A : {A : Finset ℤ // A.card = n},
+    ({θ : ℝ | θ ∈ Ico 0 1 ∧ ∑ a ∈ A.val, cos (2 * π * a * θ) = 0} : Set ℝ).ncard
+
+/-- Let $A \subset \mathbb{Z}$ be a set of size $n$. For how many $\theta \in \mathbb{R}/\mathbb{Z}$
+must we have $\sum_{a \in A} \cos(2\pi a\theta) = 0$? The answer is the function `minZeros`. -/
+@[category research open, AMS 11 42]
+theorem green_82 : (answer(sorry) : ℕ+ → ℕ∞) = minZeros := by
```
mo271, line 42: I think we are in the process of adding a linter that will forbid having `answer(sorry)` in `research open` statements

mo271 (changes requested):

FormalConjectures/GreensOpenProblems/82.lean
Paul-Lez: Do you mean research open or research solved?

Paul-Lez: @mo271  shouldn't the ban be for `answer(sorry)` in `research solved`?

mo271: sorry, confused myselfs, looks ok now

mo271 (approved):

FormalConjectures/GreensOpenProblems/82.lean
Paul-Lez: No prob!

PR #1899, opened 2026-01-27
feat: add optimization constants

felixpernegger:
See the discussion at
https://leanprover.zulipchat.com/#narrow/channel/524981-Formal-conjectures/topic/Optimization.20constants/with/570016079

It is probably useful if there is a somewhat standard way to phrase the open question (what is lower and upper bound etc) and this is likely a good place to discuss this. Namely how many of the theorems/conjectures ```c1a_le```, ```c1a_ge```, ```c1a_eq``` do we want etc?

(Maybe also add optimization constants to the readme, next the green and erdos problems? as well as a tag for PRs)

Once/If this PR is merged, I will try to make the optimization constants [repository](https://github.com/teorth/optimizationproblems) link to this, similar as it is done for Erdos problems.

### DIFF
--- a/FormalConjectures/OptimizationConstants/1a.lean
+++ b/FormalConjectures/OptimizationConstants/1a.lean
@@ -13,7 +13,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 -/
-
 import FormalConjectures.Util.ProblemImports
 
 /-!


FormalConjectures/OptimizationConstants/1a.lean
```diff
@@ -0,0 +1,69 @@
...
+- [M2010] Matolcsi, Máté, and Carlos Vinuesa. "Improved bounds on the supremum of autoconvolutions."
+  Journal of mathematical analysis and applications 372.2 (2010): 439-447. [arXiv:0907.1379](https://arxiv.org/abs/0907.1379)
+- [Y2026] Yuksekgonul, Mert et al., "Learning to Discover at Test Time," 2026, [arXiv:2601.16175](https://arxiv.org/abs/2601.16175)
+-/
+
+open Set
+
+namespace Constant1a
+
+/-- A real number satisfying a certain inequality about integral. -/
```
Paul-Lez: I think this docstring could be a little more descriptive:)

felixpernegger: Thanks, I updated the docstring. I hope its better now.
I dont think putting the entire inequality in the docstring makes sense though as its quite technical.

FormalConjectures/OptimizationConstants/1a.lean
```diff
@@ -0,0 +1,71 @@
...
+  sorry
+
+/-- The best known upper bound, proven by Yuksekgonul et al. in [Y2026] -/
+@[category research solved, AMS 05 11 26]
+theorem c1a_upper_bound : C1a ≤ 1.5029 := by
+  sorry
+
+/-- How can the upper bound be improved? -/
+@[category research open, AMS 05 11 26]
+theorem c1a_le : C1a ≤ answer(sorry) := by
```
mo271:
```suggestion
theorem c1a_le : ∃ (c : ℝ), c < 1.5029 ∧ C1a ≤ c := by
```
Might be a version without `answer(sorry)` or 
```lean
theorem c1a_le : ∃ (c : ℝ), c = answer(sorry) ∧ c < 1.5029 ∧ C1a ≤ c := by
```
has the advantage, that no manually check is needed if a good answer was found. 
I think it would be worth the effort to think about how to best do this if we are doing this for many optimization problems!

Perhaps we could add some definitions that work analogously. 
Pinging also @Paul-Lez and @YaelDillies and @smmercuri and @callesonne and @eric-wieser 
for ideas. 

Here's an adhoc way I did it at the time: this is https://teorth.github.io/optimizationproblems/constants/1b.html 
(probably needs update of the constants, btw)

 https://github.com/google-deepmind/formal-conjectures/blob/169aaebeb73230c97bcf87e2a3b73ed2b202de46/FormalConjectures/ErdosProblems/36.lean#L181-L209

This is also a thing to consider: we might want to introduce a mechanism to pull the constants from 
https://teorth.github.io/optimizationproblems/constants
So perhaps they better go into an attribute?

Paul-Lez: I can have a go at implementing an attribute for those.

YaelDillies: > theorem c1a_le : ∃ (c : ℝ), c = answer(sorry) ∧ c \u003C 1.5029 ∧ C1a ≤ c := by\nhas the advantage, that no manually check is needed if a good answer was found.

I think this is a good idea but as I pointed out previously:
1. If you use numerics here, then an AI could "improve" the bound by just computing more digits 
2. Even if you use exact values, it might improve the bound infinitesimally by pushing some slack away in the argument

I believe these are minor drawbacks, but we should be aware of them

mo271: Agreed that this "computing more digits" might happen and somthing we should be aware of, but I think it is a little bit the point of the project [https://teorth.github.io/optimizationproblems](https://teorth.github.io/optimizationproblems/constants)

I don't think it is a major issue on the formal side. It it turns out to be annoying we could introduce stop-gaps for that, to require at least a 1% improvement or something like this, but don't let perfect be the enemy of good here...

mo271: > I can have a go at implementing an attribute for those.

@Paul-Lez that would be great

felixpernegger: > I don't think it is a major issue on the formal side. It it turns out to be annoying we could introduce stop-gaps for that, to require at least a 1% improvement or something like this, but don't let perfect be the enemy of good here...

A straight up threshold might be a bit problematic though as well, like for matrix multiplication complexity. But I agree "adding more digits" is likely not a big problem overall.

felixpernegger: > Might be a version without `answer(sorry)` or
> 
> ```lean
> theorem c1a_le : ∃ (c : ℝ), c = answer(sorry) ∧ c < 1.5029 ∧ C1a ≤ c := by
> ```

I changed to this format now, this seems reasonable.

I suggest we merge this PR to be able to link the optimization repo to formal conjectures now?
If number of decimals or attribute indeed become relevant, this can certainly be dealt with then.

felixpernegger:
Can someone take a look at this? :)
Would be nice, since we can then link the other repo to this.

FormalConjectures/OptimizationConstants/1a.lean
```diff
@@ -0,0 +1,71 @@
...
+  Journal of mathematical analysis and applications 372.2 (2010): 439-447. [arXiv:0907.1379](https://arxiv.org/abs/0907.1379)
+- [Y2026] Yuksekgonul, Mert et al., "Learning to Discover at Test Time," 2026, [arXiv:2601.16175](https://arxiv.org/abs/2601.16175)
+-/
+
+open Set
+
+namespace Constant1a
+
+/--
+A real number satisfying a certain inequality about (auto)convolutions and $L^2$-norms of functions.
```
eric-wieser: This docstring suggests that `IsL2Bound C : Real`, which is not true.

felixpernegger:
```suggestion
$\text{IsL2Bound $C$}$ means that a real number $C$ satisfies a certain inequality about (auto)convolutions and $L^2$-norms of functions.
```
Maybe like this? (Inspired by the docstring for ```IsOpen```, otherwise i really dont have an idea to be honest)

YaelDillies: Why don't you just inline the predicate?

FormalConjectures/OptimizationConstants/1a.lean
```diff
@@ -0,0 +1,71 @@
...
+Such numbers are related to the maximal size of Sidon sets in additive combinatorics. -/
+def IsL2Bound (C : ℝ) : Prop :=
+  ∀ ⦃f : ℝ → ℝ⦄, 0 ≤ f →  C * (∫ x in (- 1 / 4)..(1 / 4), f x) ^ 2
+    ≤ sSup {r | ∃ t ∈ Icc (1 / 2 : ℝ) 1, r = ∫ x, f (t - x) * f x}
+
+/-- **Tao's Optimization constant 1a / An autocorrelation constant related to Sidon sets**:
+The biggest real number satisfying a certain inequality about integral. -/
+noncomputable def C1a : ℝ := sSup {C : ℝ | IsL2Bound C}
```
eric-wieser: Should this be an `ENNReal` so that we don't have to worry about the bound being zero accidentally?

FormalConjectures/OptimizationConstants/1a.lean
```diff
@@ -0,0 +1,71 @@
...
+open Set
+
+namespace Constant1a
+
+/--
+A real number satisfying a certain inequality about (auto)convolutions and $L^2$-norms of functions.
+Such numbers are related to the maximal size of Sidon sets in additive combinatorics. -/
+def IsL2Bound (C : ℝ) : Prop :=
+  ∀ ⦃f : ℝ → ℝ⦄, 0 ≤ f →  C * (∫ x in (- 1 / 4)..(1 / 4), f x) ^ 2
+    ≤ sSup {r | ∃ t ∈ Icc (1 / 2 : ℝ) 1, r = ∫ x, f (t - x) * f x}
```
eric-wieser: I think this can be written more concisely as
```suggestion
    ≤ sSup {∫ x, f (t - x) * f x | t ∈ Icc (1 / 2 : ℝ) 1}
```
or
```suggestion
    ≤ sSup ((fun t => ∫ x, f (t - x) * f x) '' Icc (1 / 2 : ℝ) 1)
```

FormalConjectures/OptimizationConstants/1a.lean
felixpernegger: Since its proven 1.2 \leq C1a \leq 1.57, I dont think this matters?

FormalConjectures/OptimizationConstants/1a.lean
felixpernegger:
```suggestion
`IsL2Bound` $C$ means that a real number $C$ satisfies a certain inequality about (auto)convolutions and $L^2$-norms of functions.
```
do you mean like this? sorry I dont quite understand

FormalConjectures/OptimizationConstants/1a.lean
```diff
@@ -0,0 +1,71 @@
...
+  sorry
+
+/-- The best known upper bound, proven by Yuksekgonul et al. in [Y2026] -/
+@[category research solved, AMS 05 11 26]
+theorem c1a_upper_bound : C1a ≤ 1.5029 := by
+  sorry
+
+/-- How can the upper bound be improved? -/
+@[category research open, AMS 05 11 26]
+theorem c1a_le : ∃ (c : ℝ), c = answer(sorry) ∧ c < 1.5029 ∧ C1a ≤ c := by
```
eric-wieser:
```suggestion
theorem mem_Ico_c1a : answer(sorry) ∈ Set.Ico C1a 1.5029 := by
```
etc

felixpernegger: The point of this PR was mostly to have a somewhat stardardised method for formalising optimization constants, so this should stay as it is

mo271: arguably what eric proposes is nicer, so perhaps this should become the standard?

FormalConjectures/OptimizationConstants/1a.lean
YaelDillies: "inlining" is a general term for removing a definition and replacing all its occurrences by its body

felixpernegger: thanks a lot!

FormalConjectures/OptimizationConstants/1a.lean
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Constant1a
+
+/-- **Tao's Optimization constant 1a / An autocorrelation constant related to Sidon sets**:
+The biggest real number satisfying a certain inequality about (auto)convolutions
+and $L^2$-norms of functions.
+This number is related to the maximal size of Sidon sets in additive combinatorics. -/
+noncomputable def C1a : ℝ :=
+  sSup {C : ℝ | ∀ ⦃f : ℝ → ℝ⦄, 0 ≤ f →  C * (∫ x in (- 1 / 4)..(1 / 4), f x) ^ 2
+    ≤ sSup {∫ x, f (t - x) * f x | t ∈ Icc (1 / 2 : ℝ) 1}}
```
YaelDillies, line 38: Today there was a preprint about this constant on arxiv. Could you maybe state its results?

felixpernegger: Can you send me the link?

YaelDillies: https://arxiv.org/abs/2602.07292

mo271:
I'd love to make progress here, @felixpernegger @eric-wieser, @Paul-Lez and @YaelDillies! There are more and more problems on https://github.com/teorth/optimizationproblems and we should get to linking the two repos...
Can we get this merged here and then sequentially improve the integration?

PR #1901, opened 2026-01-27
Fix erdos_455 misformalization

mzhorvath1:
The previous formalization was false due to natural number subtraction underflow, causing an issue with the convexity condition.

### DIFF
--- a/FormalConjectures/OEIS/63880.lean
+++ b/FormalConjectures/OEIS/63880.lean
@@ -1,0 +1,111 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+import FormalConjectures.Util.ProblemImports
+
+open scoped ArithmeticFunction
+
+/-!
+# Conjectures associated with A063880
+
+A063880 lists numbers $n$ such that $\sigma(n) = 2 \cdot \text{usigma}(n)$, where $\sigma(n)$ is the
+sum of all divisors and $\text{usigma}(n)$ is the sum of unitary divisors.
+
+Equivalently, these are numbers whose unitary and non-unitary divisors have equal sum.
+
+The conjectures state that all members satisfy $n \equiv 108 \pmod{216}$, and that all
+primitive terms (those whose proper divisors aren't in the sequence) are powerful numbers,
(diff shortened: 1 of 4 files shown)


FormalConjectures/ErdosProblems/455.lean
```diff
@@ -28,20 +28,20 @@ open Filter
...
 
 /-- Let `q : ℕ → ℕ` be a strictly increasing sequence of primes such that
-`q (n + 1) - q n ≥ q n - q (n - 1)`. Must `lim q n / (n ^ 2) = ∞`? -/
+`q (n + 2) - q (n + 1) ≥ q (n + 1) - q n`. Must `lim q n / (n ^ 2) = ∞`? -/
 @[category research open, AMS 11]
 theorem erdos_455: answer(sorry) ↔ ∀ q : ℕ → ℕ, StrictMono q →
-    (∀ n, (q n).Prime ∧ q (n + 1) - q n ≥ q n - q (n - 1)) →
-    Tendsto (fun n : ℕ => (q n : ℝ) / n ^ 2) atTop atTop := by
+    (∀ n, (q n).Prime ∧ q (n + 2) - q (n + 1) ≥ q (n + 1) - q n) →
+    Tendsto (fun n : ℕ => (q n : ℝ) / (n ^ 2 : ℝ)) atTop atTop := by
```
mo271:
```suggestion
    Tendsto (fun n : ℕ => (q n : ℝ) / n ^ 2) atTop atTop := by
```
This is not necessary, the casting will be done in the correct way automatically, right?

FormalConjectures/ErdosProblems/455.lean
```diff
@@ -28,20 +28,20 @@ open Filter
...
 /--  Let `q : ℕ → ℕ` be a strictly increasing sequence of primes such that
-`q (n + 1) - q n ≥ q n - q (n - 1)`. Then `liminf q n / (n ^ 2) > 0.352`, and this is proved in
+`q (n + 2) - q (n + 1) ≥ q (n + 1) - q n`. Then `liminf q n / (n ^ 2) > 0.352`, and this is proved in
 [Ri76]. -/
 @[category research solved, AMS 11]
 theorem erdos_455.liminf : ∀ q : ℕ → ℕ, StrictMono q →
-    (∀ n, (q n).Prime ∧ q (n + 1) - q n ≥ q n - q (n - 1)) →
-    liminf (fun n : ℕ => (q n : ℝ) / n ^ 2) atTop > 0.352 := by
+    (∀ n, (q n).Prime ∧ q (n + 2) - q (n + 1) ≥ q (n + 1) - q n) →
+    liminf (fun n : ℕ => (q n : ℝ) / (n ^ 2 : ℝ)) atTop > 0.352 := by
```
mo271:
```suggestion
    liminf (fun n : ℕ => (q n : ℝ) / n ^ 2) atTop > 0.352 := by
```
same here

mo271 (approved):
good catch, thanks!

FormalConjectures/ErdosProblems/455.lean
mzhorvath1: Good point, removed the explicit casting.

mo271:
Thanks again!

PR #1904, opened 2026-01-28
feat(ErdosProblems): 1145

danielchin:
Formalizes Erdos Problem 1145: https://www.erdosproblems.com/1145

### DIFF
--- a/FormalConjectures/ErdosProblems/1145.lean
+++ b/FormalConjectures/ErdosProblems/1145.lean
@@ -29,7 +29,20 @@ open Set Filter Pointwise Topology AdditiveCombinatorics
 
 namespace Erdos1145
 
+/--
+Let $A=\{1\leq a_1 < a_2 < \cdots\}$ and $B=\{1\leq b_1 < b_2 < \cdots\}$ be sets of integers with
+$a_n/b_n\to 1$.
 
+If $A+B$ contains all sufficiently large positive integers then is it true that
+$\limsup 1_A\ast 1_B(n)=\infty$?
+
+Formalization note: There's some discussion in the comments of [erdosproblems.com/28] and
+[erdosproblems.com/1145] about whether or not $0$ should be included in $A$ or $B$ and has been
+left purposely ambiguous. Problem 1145 was originally written as $A + B = \mathbb{N}$, which
+would imply that $0$ would need to exist in $A$ or $B$ to include $1$ in $A + B$. However, it's been
+made more general and rewritten as "sufficiently large positive integers". The formalization below
+is the version that includes $0$.
+-/
 def Erdos1145Prop : Prop :=
   ∀ ⦃A B : Set ℕ⦄ (_ : A.Infinite) (_ : B.Infinite),
     Tendsto (fun n ↦ (Nat.nth (· ∈ A) n : ℝ) / (Nat.nth (· ∈ B) n : ℝ)) atTop (𝓝 1) →
--- a/FormalConjectures/ErdosProblems/1145.lean
+++ b/FormalConjectures/ErdosProblems/1145.lean
@@ -41,12 +41,12 @@ A stronger form of [erdosproblems.com/28].
 -/
 @[category research open, AMS 5]
 theorem erdos_1145 :
-  answer(sorry) ↔ (∀ (A B : Set ℕ) (hA : A.Infinite) (hB : B.Infinite),
-    A ⊆ Ici 1 →
-    B ⊆ Ici 1 →
+  answer(sorry) ↔ ∀ ⦃A B : Set ℕ⦄ (hA : A.Infinite) (hB : B.Infinite),
+    0 ∉ A →
+    0 ∉ B →
     Tendsto (fun n ↦ (Nat.nth (· ∈ A) n : ℝ) / (Nat.nth (· ∈ B) n : ℝ)) atTop (𝓝 1) →
     Ici 2 ⊆ A + B →
-    limsup (fun n ↦ ({p : ℕ × ℕ | p.1 ∈ A ∧ p.2 ∈ B ∧ p.1 + p.2 = n}.ncard : ℕ∞)) atTop = ⊤) := by
+    limsup (fun n ↦ ({p : ℕ × ℕ | p.1 ∈ A ∧ p.2 ∈ B ∧ p.1 + p.2 = n}.ncard : ℕ∞)) atTop = ⊤ := by
   sorry
 
 end Erdos1145
--- a/FormalConjectures/ErdosProblems/1145.lean
+++ b/FormalConjectures/ErdosProblems/1145.lean
@@ -1,0 +1,52 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 1145
+
+*References:*
+- [erdosproblems.com/28](https://www.erdosproblems.com/28)
+- [erdosproblems.com/1145](https://www.erdosproblems.com/1145)
+-/
+
+open Set Filter Pointwise Topology
+
+namespace Erdos1145
+
+
+/--
+Let $A=\{1\leq a_1 < a_2 < \cdots\}$ and $B=\{1\leq b_1 < b_2 < \cdots\}$ be sets of integers
+with $a_n/b_n\to 1$.
+
+If $A+B=\mathbb{N}$ then is it true that $\limsup 1_A\ast 1_B(n)=\infty$?
+
+A conjecture of Erdős and Sárközy.
+
+A stronger form of [erdosproblems.com/28].
+-/
+@[category research open, AMS 5]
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/1145.lean
```diff
@@ -0,0 +1,52 @@
...
+
+A conjecture of Erdős and Sárközy.
+
+A stronger form of [erdosproblems.com/28].
+-/
+@[category research open, AMS 5]
+theorem erdos_1145 :
+  answer(sorry) ↔ (∀ (A B : Set ℕ) (hA : A.Infinite) (hB : B.Infinite),
+    A ⊆ Ici 1 →
+    B ⊆ Ici 1 →
```
felixpernegger:
```suggestion
    0 \notin A →
    0 \notin B →
```

FormalConjectures/ErdosProblems/1145.lean
```diff
@@ -0,0 +1,52 @@
...
+
+If $A+B=\mathbb{N}$ then is it true that $\limsup 1_A\ast 1_B(n)=\infty$?
+
+A conjecture of Erdős and Sárközy.
+
+A stronger form of [erdosproblems.com/28].
+-/
+@[category research open, AMS 5]
+theorem erdos_1145 :
+  answer(sorry) ↔ (∀ (A B : Set ℕ) (hA : A.Infinite) (hB : B.Infinite),
```
felixpernegger:
```suggestion
  answer(sorry) ↔ (∀ ⦃A B : Set ℕ⦄ (hA : A.Infinite) (hB : B.Infinite),
```
Also, the outward brackets are redundant I think? (check this)

FormalConjectures/ErdosProblems/1145.lean
```diff
@@ -0,0 +1,52 @@
...
+    A ⊆ Ici 1 →
+    B ⊆ Ici 1 →
+    Tendsto (fun n ↦ (Nat.nth (· ∈ A) n : ℝ) / (Nat.nth (· ∈ B) n : ℝ)) atTop (𝓝 1) →
+    Ici 2 ⊆ A + B →
```
felixpernegger: why do you need this? its not mentioned on the website

felixpernegger: Actually you are right. There is a problem with the statement on the website, as we cant have $1 \in A + B$

felixpernegger: I asked about this on the website. Let's not merge this until this is cleared up.

felixpernegger:
Can you include the implication (and _maybe_ prove it) that 1145 => 28?

FormalConjectures/ErdosProblems/1145.lean
danielchin: Done.

FormalConjectures/ErdosProblems/1145.lean
danielchin: Yup, the outside parens can be removed.

FormalConjectures/ErdosProblems/1145.lean
danielchin: T. Bloom updated the question on erdosproblems.com so I've updated this as well.

FormalConjectures/ErdosProblems/1145.lean
```diff
@@ -0,0 +1,53 @@
...
+A stronger form of [erdosproblems.com/28].
+-/
+@[category research open, AMS 5]
+theorem erdos_1145 :
+  answer(sorry) ↔ ∀ ⦃A B : Set ℕ⦄ (hA : A.Infinite) (hB : B.Infinite),
+    0 ∉ A →
+    0 ∉ B →
+    Tendsto (fun n ↦ (Nat.nth (· ∈ A) n : ℝ) / (Nat.nth (· ∈ B) n : ℝ)) atTop (𝓝 1) →
+    (∀ᶠ n in atTop, n ∈ A + B) →
+    limsup (fun n ↦ ({p : ℕ × ℕ | p.1 ∈ A ∧ p.2 ∈ B ∧ p.1 + p.2 = n}.ncard : ℕ∞)) atTop = ⊤ := by
```
mo271:
```suggestion
    limsup (fun n => ↑(((𝟙_A ∗ 𝟙_B) : ℕ → ℕ) n)) atTop = (⊤ : ℕ∞) := by
```
Perhaps also works?

mo271 (approved):
Thanks, looks good!

Perhaps we could use

mo271:
Thanks especially for clarifying the problem by asking a question on erdosproblems.com!

FormalConjectures/ErdosProblems/1145.lean
danielchin: Yup, looks like this works and is equivalent.

mo271:
> Can you include the implication (and _maybe_ prove it) that 1145 => 28?

```lean
theorem erdos_1145 :
  ∀ ⦃A B : Set ℕ⦄ (hA : A.Infinite) (hB : B.Infinite),
    Tendsto (fun n ↦ (Nat.nth (· ∈ A) n : ℝ) / (Nat.nth (· ∈ B) n : ℝ)) atTop (𝓝 1) →
    (∀ᶠ n in atTop, n ∈ A + B) →
    limsup (fun n => ↑(((𝟙_A ∗ 𝟙_B) : ℕ → ℕ) n)) atTop = (⊤ : ℕ∞) := by
  sorry

/--
If $A ⊆ \mathbb{N}$ is such that $A + A$ contains all but finitely many integers then
 $\limsup 1_A ∗ 1_A(n) = \infty$.
-/
theorem erdos_28  (A : Set ℕ) (h : (A + A)ᶜ.Finite) :
    limsup (fun (n : ℕ) => (sumRep A n : ℕ∞)) atTop = (⊤ : ℕ∞) := by
  sorry

@[category test, AMS 11]
theorem test : type_of% erdos_1145 → type_of% erdos_28 := by
  delta sumRep
  use fun and K V=>V.exists_le.elim fun R L=>by_contra fun and' =>?_
  by_cases h:K.Finite
  · use(h.add h).infinite_compl V
  · bound[and h h (tendsto_atTop_of_eventually_const fun and p=>div_self (mod_cast (ne_of_gt (p.trans ((Nat.nth_strictMono h)).le_apply)))) ((Filter.eventually_gt_atTop R).mono fun and=>not_imp_comm.1 (L _) ∘not_le_of_lt)]

```
is a start for this (proof by AlphaProof)

I removed the `answer(sorry)`, which only 1145 should have, so needs to be put back there, but then it is probably a good idea to define a `Prop` for erdos 1145, to make it easier. 
Then one needs to decide which way around we import the definitions (probably better to do everything in the 1145 file?
Also I removed the hypothesis that those sets start at zero, not sure if this makes any difference, but in 28 we didn't have the condition!


In any case, @danielchin please either do that here, or add a `TODO`, so we can add it at some later point!

danielchin:
Added the implication here. The only change I had to make to the AlphaProof proof was changing `not_le_of_lt` to `not_le_of_gt` due to a deprecation warning of `not_le_of_lt`.

From an organizational standpoint, I would think it makes more sense to keep problem 28 in `28.lean` and just import it into `1145.lean`. It also looks like problem 40 is a stronger version of 28, so if we were to add the implication of 40 => 28, it would be easier to leave 28 where it is now instead of doing a roundabout way to get to 28 through `1145.lean`.

FormalConjectures/ErdosProblems/1145.lean
```diff
@@ -0,0 +1,67 @@
...
+theorem erdos_1145_implies_erdos_28 : Erdos1145Prop → type_of% Erdos28.erdos_28 := by
+  delta sumRep
+  use fun and K V=>V.exists_le.elim fun R L=>by_contra fun and' =>?_
+  by_cases h:K.Finite
+  · use(h.add h).infinite_compl V
+  · bound [and h h
+      (tendsto_atTop_of_eventually_const fun and p =>
+        div_self (mod_cast (ne_of_gt (p.trans ((Nat.nth_strictMono h)).le_apply))))
+      ((Filter.eventually_gt_atTop R).mono fun and =>
+        not_imp_comm.1 (L _) ∘ not_le_of_gt)]
```
mo271: Let's clean up the proof! for starters `and` is not a good name for a variable...

danielchin: Okay this was a bit of an exercise for me (assisted with Antigravity/Gemini).

The original AlphaProof proof was a bit hard for me to grok. I used Gemini to look up proper Lean style guidelines from the Mathlib documentation (i.e., add spaces before and after `:`, avoid using `$`, using proper variable names, etc.) And rewrote in a way that's more readable (or at least readable to me).

Let me know if you prefer this or the original AlphaProof proof with some minor tweaks.

mo271: LGTM! Thanks

mo271:
The only thing I'm not worried about is how we dropped the condition that A and B don't contain 0. Does this make any difference in the problem? In any case, now there is a difference between the docstring (which should also be duplicated for the Prop, btw) and the lean code. Either change the lean code back to include the condition (preferred!, but perhaps hard to make work with the relation to erdos28) or explain in the docstring if it doesn't matter.

mo271 (commented):
Thanks again!

danielchin:
I went back and forth on this one for a day. It looks like the implication from a zero excluded 1145 to 28 is possible, but it starts blowing up the proof in size to the point where it's probably not worth it to keep in this repository (I also never got to a point where it compiled properly).

There's also discussion on the Erdos problem website under the comments of both 1145 and 28 discussing whether should be included for both of the problems. And it seems like T. Bloom has made it a bit ambiguous on purpose. 1145 was originally written as $A + B = \mathbb{N}$ (according to Woett's comment in 28), which would imply 0 was included. And T. Bloom's comment in 1145, implies that 1145 could be interpreted as 0 could be included or we could treat it as "it holds for sufficiently large integers". So all in all ¯\_(ツ)_/¯

Left a docstring describing the above.

Issue #1407, opened 2025-12-23
Lint against `answer(sorry)` related errors.


Paul-Lez:
There are many ways to misuse the `answer(sorry)` convention and it would be good to lint against these.

PR #1922, opened 2026-01-28
fix(OEIS/41): `answer(sorry)` position

mo271:
was missed in the review of #1704 apparently?


mo271:
we should really lint against that!

mo271:
a case for #1407

Paul-Lez (approved):

PR #1926, opened 2026-01-29
update solution status for Erdős 705

mo271:
has recently been updated, see also https://www.erdosproblems.com/forum/thread/705


felixpernegger (approved):
Thanks

smmercuri (approved):

PR #1933, opened 2026-01-29
Update the year in the copyright header in the README

danielchin:
Updating this so I don't have to manually edit the year every time I copy and paste the header :^)


felixpernegger (approved):
good spot

mo271:
sure, why not, thanks! 

You can also start typing "copyr..." and then press tab in vscode and it will autocomplete it with the right year for you...

mo271 (approved):

PR #1934, opened 2026-01-29
chore: mention formal solution to Boxdot in the attribute

Paul-Lez:


Paul-Lez:
> Nice! Potentially also mention who formalised it, if it is easy to figure it out

Do you mean mentioning that in a docstring? Or in the attribute?

mo271:
> > Nice! Potentially also mention who formalised it, if it is easy to figure it out
> 
> Do you mean mentioning that in a docstring? Or in the attribute?

I meant docstring only. 

I think if we re-structure the attribute to take more metadata, then we should also do it for the normal `research solved` one, and so far we haven't even decided on a constistent way to handle reference to papers/books...

PR #1936, opened 2026-01-30
Refine formalization of Erdős Problem 503

### LINKED ISSUE #759: Erdős Problem 503
### What is the conjecture
    
https://www.erdosproblems.com/503

  What is the size of the largest $A\subseteq \mathbb{R}^n$ such that every three points from $A$ determine an isosceles triangle? That is, for any three points $x,y,z$ from $A$, at least two of the distances $\lvert x-y\rvert,\lvert y-z\rvert,\lvert x-z\rvert$ are equal.
  

Status: open

### Choose either option

Uday-maverick:
This PR refines the formalization of Erdős Problem 503.

Changes include:
- making the isosceles condition explicit,
- restricting attention to finite sets,
- clarifying the extremal formulation,
- and stating the known upper and lower bounds in a precise form.

These changes better reflect the original statement on
https://www.erdosproblems.com/503.

Closes #759.

### DIFF
--- a/FormalConjectures/ErdosProblems/503.lean
+++ b/FormalConjectures/ErdosProblems/503.lean
@@ -26,6 +26,13 @@ namespace Erdos503
 
 open scoped EuclideanGeometry
 
+open Set
+
+/-- A set is isosceles if every three points determine an isosceles triangle -/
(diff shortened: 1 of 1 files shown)


mo271:
Thanks, I don't think this improves the formalisation we already have

PR #1942, opened 2026-01-31
feat(ErdosProblems): 886

### LINKED ISSUE #1003: Erdős Problem 886
### What is the conjecture

https://www.erdosproblems.com/886

  Let $\epsilon>0$. Is it true that, for all large $n$, the number of divisors of $n$ in $(n^{1/2},n^{1/2}+n^{1/2-\epsilon})$ is $O_\epsilon(1)$?

Status:  open

### Choose either option

danielchin:
Closes #1003 

Formalizes Erdos Problem 886: https://www.erdosproblems.com/886

### DIFF
--- a/FormalConjectures/ErdosProblems/886.lean
+++ b/FormalConjectures/ErdosProblems/886.lean
@@ -32,9 +32,9 @@ namespace Erdos886
 /--
 The set of divisors of $n$ in the interval $(n^{1/2}, n^{1/2} + n^{1/2-\epsilon})$.
 -/
-noncomputable def Erdos886Divisors (n : ℕ) (ε : ℝ) : Finset ℕ :=
+noncomputable def Erdos886Divisors (n : ℕ) (ε : ℝ) (C : ℝ) : Finset ℕ :=
   (divisors n).filter (fun d =>
-    (n : ℝ) ^ (1/2 : ℝ) < d ∧ (d : ℝ) < (n : ℝ) ^ (1/2 : ℝ) + (n : ℝ) ^ (1/2 - ε))
+    (n : ℝ) ^ (1/2 : ℝ) < d ∧ (d : ℝ) < (n : ℝ) ^ (1/2 : ℝ) + C * (n : ℝ) ^ (1/2 - ε))
 
 /--
 Let $\epsilon>0$. Is it true that, for all large $n$, the number of divisors of $n$ in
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/886.lean
```diff
@@ -0,0 +1,59 @@
...
+-/
+
+open Nat Filter
+
+namespace Erdos886
+
+/--
+The set of divisors of $n$ in the interval $(n^{1/2}, n^{1/2} + n^{1/2-\epsilon})$.
+-/
+noncomputable def erdos_886_divisors (n : ℕ) (ε : ℝ) : Finset ℕ :=
```
mo271:
```suggestion
noncomputable def Erdos886Divisors (n : ℕ) (ε : ℝ) : Finset ℕ :=
```

FormalConjectures/ErdosProblems/886.lean
```diff
@@ -0,0 +1,59 @@
...
+  sorry
+
+/--
+Erdős and Rosenfeld [ErRo97] proved that there are infinitely many $n$ such that there are
+four divisors of $n$ in $(n^{1/2},n^{1/2}+n^{1/4})$.
+-/
+@[category research solved, AMS 11]
+theorem erdos_886.variants.rosenfeld :
+    Set.Infinite {n | 4 ≤ (erdos_886_divisors n (1/4)).card} := by
+  sorry
```
mo271, line 62: The statement as formalised can be disproved with AlphaProof: 
```
theorem erdos_886.variants.rosenfeld :
    ¬ Set.Infinite {n | 4 ≤ (erdos_886_divisors n (1/4)).card} := by
  norm_num [erdos_886_divisors]
  refine BddAbove.finite ⟨16,fun R L =>not_lt.1 fun and=> L.out.not_lt ?_⟩
  set α := R.divisors.filter fun and : ℕ=>(R^(1/2: ℝ): ℝ)<and∧ (and: ℝ)<R^(1/2:ℝ)+R^(1/4 : ℝ)
  replace L:α ⊆{sInf {x :ℕ|(R^(1/2: ℝ): ℝ)<x∧(x: ℝ)<R^(1/2:ℝ)+R^(1/4: ℝ)},sSup α}:=fun a s=>?_
  · exact ( (α.card_mono L).trans Finset.card_le_two).trans_lt (by decide)
  use Finset.mem_insert.2 (or_iff_not_imp_left.2 fun and=>List.mem_singleton.2<|symm (IsGreatest.csSup_eq ⟨s,fun A B=>not_lt.1 (( A.dvd_of_mem_divisors ( Finset.filter_subset _ _ B)).rec fun and K V=>? _)⟩))
  push_cast[←Real.sqrt_eq_rpow, K, α, Finset.mem_coe, Finset.mem_filter,Nat.mem_divisors]at*
  bound[(A*‹Nat›:ℝ).sq_sqrt<|by bound,(A*‹Nat›:ℝ).rpow_inv_natCast_pow (by bound) four_ne_zero,B.2]
  convert and (.symm (IsLeast.csInf_eq ⟨ (by use left_2),left.elim fun and _ _ _=>not_lt.1 fun and' => if I: A+and_1 ≤ a+ and then(? _)else _,⟩))
  · exact (Real.le_sqrt_of_sq_le (mod_cast‹_ = _›▸sq a▸mul_left_mono (by nlinarith only[I, V,‹_ = _›]))).not_lt left_2
  have:(A*and_1: ℝ) = a*and := by norm_cast
  have:(A: ℝ)≥a+1∧(A:ℝ)+and_1≥a+and+1:=mod_cast by valid
  nlinarith[sq_pos_of_neg (sub_neg.2 left_2),sq_nonneg (a-((A*and_1)^(1/4:ℝ))^2 : ℝ),mul_le_mul_of_nonneg_left this.2 a.cast_nonneg, mul_le_mul_of_nonneg_left B.2.2.le a.cast_nonneg]
```

I don't see a problem with the formalisation here, but suspect that the statement on erdosproblems is not to be take literally, perhaps best to ask there or inspect the original source https://eudml.org/doc/206983

mo271 (approved):

mo271:
Thanks, LGTM, I just wrote on erdosproblems.com to check for the additional statement..

FormalConjectures/ErdosProblems/886.lean
mo271: Given that there is still some discussion around this result, and the version on erdosproblems is buggy: 
https://www.erdosproblems.com/forum/thread/886
I rpopose to leave `erdos_886.variants.rosenfeld` as TODO instead and update when the problem is fixed on erdosproblems.com.

danielchin: Great catch on this! From the paper and the comments, it sounds like the proper statement might be: There exists a constant $C > 1$ such that there are infinitely many $n$ with four divisors in the interval $(n^{1/2}, n^{1/2} + C n^{1/4})$.

What do you reckon I do for now? Leave a TODO until T. Bloom updates the erdosproblems.com/886 page? Or just go ahead and update the statement?

FormalConjectures/ErdosProblems/886.lean
danielchin: Done.

FormalConjectures/ErdosProblems/886.lean
mo271: Both is fine! In case you add the statement explain the situation in the docstring

danielchin: Formalized the statement that I posted above and left a TODO. Edited Erdos886Divisors to take in parameter $C$.

mo271:
Thanks again!

PR #1946, opened 2026-01-31
fix reference in GreensOpenProblem 72

### LINKED ISSUE #1945: Incorrect references to Problem 62 in Green's Open Problem 72
The Ben GreensOpenProblem 72 has some misconfiguration in the reference section , it states # Ben Green's Open Problem 62 instead of 72.

### Choose either option

Aditi0522:
Closes #1945

### DIFF
--- a/FormalConjectures/GreensOpenProblems/72.lean
+++ b/FormalConjectures/GreensOpenProblems/72.lean
@@ -25,7 +25,6 @@ are there $3$ of the points on a common line?
 
 *References:*
 - [Ben Green's Open Problem 72](https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.72)
-- [Ben Green's Open Problem 62](https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.62)
 - [Wikipedia](https://en.wikipedia.org/wiki/No-three-in-line_problem)
 - [GK2025] Grebennikov, A. Kwan, M. No $(k + 1)$-in-line problem for large constant $k$.
   https://arxiv.org/abs/2510.17743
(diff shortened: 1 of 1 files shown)


FormalConjectures/GreensOpenProblems/72.lean
```diff
@@ -16,14 +16,15 @@ limitations under the License.
...
+# Ben Green's Open Problem 72
 
 More commonly known as the **no-three-in-line problem**.
 
 Given $N \lt 2$ and a more than $2 * N$ points on an $N \times N$-grid,
 are there $3$ of the points on a common line?
 
 *References:*
+- [Ben Green's Open Problem 72](https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.72)
 - [Ben Green's Open Problem 62](https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.62)
```
mo271:
```suggestion
```

mo271 (commented):
Thanks!

FormalConjectures/GreensOpenProblems/72.lean
Aditi0522: I also just checked this formalization in Green72(given below) and I think this is also an error, these should be 72, correct me if I am wrong. Let me know if I should change these too.

/-- **Green's Open Problem 62 / No-three-in-line problem**:
The no-k-in-line conjecture holds for $k = 3$. -/
@[category research open, AMS 05 52]
theorem green_62 {N : ℕ} (hN : 3 ≤ N) : NoKInLineFor 3 N := by
  sorry

alias no_three_in_line := green_62

/-- Does the no-three-in-line problem hold when $N$ is big enough? -/
@[category research open, AMS 05 52]
theorem green_62.variants.eventually : ∀ᶠ N in Filter.atTop, NoKInLineFor 3 N := by
  sorry

mo271: yes, I suppose so

mo271 (approved):

PR #1847, opened 2026-01-22
Add Sum of Two Squares Conjecture (Bambah-Chowla problem)

anshi282:
## Add Sum of Two Squares Conjecture (Bambah-Chowla problem)

This PR adds the formalization of the conjecture that the interval [x, x + 10x^(1/4)) always contains a sum of two squares for sufficiently large x.

### Changes:
- Add formalization of the main conjecture
- Include known Bambah-Chowla bound (2√2)
- Add weaker conjecture for intermediate constants
- Proper AMS classification (11 - Number Theory)
- References to Math StackExchange and Bambah-Chowla (1947) paper

### Mathematical Context:
This is a well-known open problem in number theory related to the distribution of sums of two squares. The Bambah-Chowla theorem provides a stronger bound (2√2 < 10), making this conjecture plausible but unproven.

PR #1951, opened 2026-01-31
Feat/heilbronn triangle 1948

anshi282:
Adds formalizations of two discrete geometry problems about minimal triangle areas in bounded point sets. Includes the unit disc conjecture and Green's Problem 77 (Heilbronn's triangle problem).


mo271:
Please don't open a pr with 3 different unrelated problems. 
Also the comments/problems already raised in #1847 still apply.

PR #1958, opened 2026-01-31
feat(Linter): catch dubious usages of `answer(sorry)`

mo271:
fixes parts of #1407

### DIFF
--- a/FormalConjectures/Util/Linters/AnswerLinter.lean
+++ b/FormalConjectures/Util/Linters/AnswerLinter.lean
@@ -32,8 +32,8 @@ register_option linter.style.answer_attribute : Bool := {
 }
 
 namespace AnswerLinter
--- Helper to check for arguments (h : True)
 
+-- Helper to check for arguments (h : True)
 def contains_early_args (stx : Syntax) : CommandElabM Bool := do
   let decl := stx[1]
   if decl.getKind != ``Lean.Parser.Command.theorem then return false


Paul-Lez (commented):
(Will give a more in depth review afterwards)
In the test file I think it would be worth adding some tests where answer(sorry) isn't a prop

mo271:
> (Will give a more in depth review afterwards) In the test file I think it would be worth adding some tests where answer(sorry) isn't a prop

done

FormalConjectures/Util/Linters/AnswerLinter.lean
```diff
@@ -0,0 +1,78 @@
...
+
+open Lean Elab Meta Linter Command Parser Term
+
+register_option linter.style.answer_attribute : Bool := {
+  defValue := true
+  descr := "enable the `answer(sorry)` style linter"
+}
+
+namespace AnswerLinter
+-- Helper to check for arguments (h : True)
```
Paul-Lez:
```suggestion
-- Helper to check for arguments (h : True)

```

FormalConjectures/Util/Linters/AnswerLinter.lean
```diff
@@ -0,0 +1,79 @@
...
+
```
mo271, line 35:
```suggestion
namespace AnswerLinter

-- Helper to check for arguments (h : True)
```

I think this was the intention?

PR #2013, opened 2026-02-02
fix(ErdosProblems): `B` should be a subset in `346`

smmercuri:
- `B` should be a subset of `range A` otherwise disjoint infinite set gives a contradiction.
- AlphaProof also found a disproof of `.example` with the subset fix, due to the `IsLacunary` condition, which contradicts the limsup and liminf properties. Presumably by "very irregular sequences" ([p. 57](https://mathweb.ucsd.edu/~ronspubs/80_11_number_theory.pdf)) Erdős and Graham mean not necessarily lacunary.
- Modify the definition of `IsLacunary` to apply only to sufficiently large indices (otherwise `f` is not lacunary).

### DIFF
--- a/FormalConjecturesForMathlib/NumberTheory/Lacunary.lean
+++ b/FormalConjecturesForMathlib/NumberTheory/Lacunary.lean
@@ -15,15 +15,19 @@ limitations under the License.
 -/
 
 import Mathlib.Analysis.Normed.Field.Lemmas
+import Mathlib.Order.Filter.Defs
 import Mathlib.Tactic.Rify
 
-/-- Say a sequence is lacunary if there exists some $\lambda > 1$ such that
-$a_{n+1}/a_n > \lambda$ for all $n\geq 1$. -/
-def IsLacunary (n : ℕ → ℕ) : Prop := ∃ c > (1 : ℝ), ∀ k, c * n k < n (k + 1)
+open Filter
 
-/-- Every lacunary sequence is strictly increasing. -/
-lemma IsLacunary.strictMono {n : ℕ → ℕ} (hn : IsLacunary n) : StrictMono n := by
-  refine strictMono_nat_of_lt_succ fun k => (Nat.cast_lt (α := ℝ)).mp ?_
-  obtain ⟨c, hc⟩ := hn
-  refine (hc.2 k).trans_le' ?_
-  grw [hc.1, one_mul]
+/-- Say a sequence is lacunary if there exists some $\lambda > 1$ such that
+$a_{n+1}/a_n > \lambda$ for all sufficiently large $n$. -/
+def IsLacunary (n : ℕ → ℕ) : Prop := ∃ c > (1 : ℝ), ∀ᶠ k in atTop, c * n k < n (k + 1)
+
+/-- Every lacunary sequence is eventually strictly increasing. -/
+lemma IsLacunary.eventually_lt {n : ℕ → ℕ} (hn : IsLacunary n) : ∀ᶠ k in atTop, n k < n (k + 1) := by
+  obtain ⟨c, hc, h⟩ := hn
+  obtain ⟨N, h⟩ := eventually_atTop.1 h
+  refine eventually_atTop.2 ⟨N, fun b hb ↦ ?_⟩
+  grw [hc] at h
+  simpa using h _ hb


FormalConjectures/ErdosProblems/346.lean
```diff
@@ -35,7 +35,7 @@ namespace Erdos346
 infinitely many terms are removed from it, `lim A (n + 1) / A n = (1 + √5) / 2`? -/
 @[category research open, AMS 11]
 theorem erdos_346 : answer(sorry) ↔ ∀ {A : ℕ → ℕ}, IsLacunary A → IsAddStronglyCompleteNatSeq A →
```
mo271, line 37: just to check, we don't need the fix for the first bullet point (for the finite subsets)?
> - $A\backslash B$ is complete for any finite subset $B$

smmercuri: This is fine in this PR as that condition is defined through shifting sequences in `IsAddStronglyCompleteNatSeq`, so we are automatically excluding subsets of the range. However, I did just notice that `IsAddStronglyComplete` also has the issue with subsets, but I don't think it is actually used anywhere

```lean
/-- A set `A ⊆ M` is complete if every sufficiently large element of `M` is a subset sum of `A`. -/
def IsAddStronglyComplete (A : Set M) : Prop :=
  ∀ ⦃B : Set M⦄, B.Finite → IsAddComplete (A \ B)

mo271: ok, great. then perhaps let's fix `IsAddStronglyComplete` (or remove it, if not used)

smmercuri: Oh actually, the condition is not needed here in the definition. It only causes an issue in `346` because of the negation of `IsAddComplete`.

mo271:
```lean
/-- We define a sequence `f` by the formula `f n = n.fib - (- 1) ^ n`. -/
def f (n : ℕ) : ℕ := if Even n then n.fib - 1 else n.fib + 1

/-- The sequence `f` is lacunary. -/
@[category test, AMS 11]
theorem erdos_346.f_isLacunary : ¬ IsLacunary f := by 
  delta f IsLacunary
  exact (by·norm_num+decide ∘fun ⟨a, R, H⟩=>((lt_mul_left · R|>.trans (H (1)))))
  ```
  
  There might be more problems?

smmercuri:
> ```lean
> /-- We define a sequence `f` by the formula `f n = n.fib - (- 1) ^ n`. -/
> def f (n : ℕ) : ℕ := if Even n then n.fib - 1 else n.fib + 1
> 
> /-- The sequence `f` is lacunary. -/
> @[category test, AMS 11]
> theorem erdos_346.f_isLacunary : ¬ IsLacunary f := by 
>   delta f IsLacunary
>   exact (by·norm_num+decide ∘fun ⟨a, R, H⟩=>((lt_mul_left · R|>.trans (H (1)))))
> ```
> 
> There might be more problems?

It's false that `f` is lacunary as is currently defined (i.e. at all indices), because e.g. `f 3 = 3` while `f 4 = 2` so there is no `c > 1` such that `c * f 3 = f 4`. I think it's meant to be "eventually lacunary"; indeed

```lean
theorem erdos_346.f_isLacunary : ∃ c > (1 : ℝ), ∀ᶠ k in atTop, c * f k < f (k + 1) := by
  simp_rw [Filter.eventually_atTop,f]
  use(1)+1/16,by norm_num, 9,fun R L=> R.even_or_odd'.elim fun and true => true.elim (·▸?_) (·▸? _)
  · norm_num[‹_›▸L.trans_lt']
    use and.strongRec fun A B=>match A with|0|1=>by norm_num | S+2=> (2 *_+1).fib_add_two▸Nat.fib_add_two.symm▸(@Nat.cast_add ℝ _ _ _).symm▸(@Nat.cast_add ℝ _ _ _).symm▸?_
    exact (@Nat.cast_add ℝ _ _ _).symm▸by linarith! only[(B _) (by constructor)]
  · norm_num[parity_simps,Nat.fib_add]
    use(4).le_induction (by norm_num) (fun F a s=> (2 *F+1).fib_add_two▸ (2 *F).fib_add_two▸(@Nat.cast_add ℝ _ _ _).symm▸(@Nat.cast_add ℝ _ _ _).symm▸?_) and (by valid)
    bound[(mod_cast (2 *F).fib_le_fib_succ: ( (2 *F).fib:ℝ) ≤ (2 *F+1).fib)]

mo271 (approved):
LGTM!

PR #2019, opened 2026-02-02
fix(ErdosProblems/99): `IsMinOn` footgun

mo271:
From the documentation:
```
/-- `IsMinOn f s a` means that `f a ≤ f x` for all `x ∈ s`. Note that we do not assume `a ∈ s`. -/
```

We noted that indeed now.

### DIFF
--- a/FormalConjectures/ErdosProblems/99.lean
+++ b/FormalConjectures/ErdosProblems/99.lean
@@ -43,9 +43,8 @@ that minimizes diameter must contain an equilateral triangle of side length 1? -
 @[category research open, AMS 52]
 theorem erdos_99 :
     answer(sorry) ↔ ∀ᶠ n in Filter.atTop, ∀ A : Finset ℝ²,
-      letI min_dist_one_set := {B : Finset ℝ² | B.card = n ∧ HasMinDist1 B}
-      A ∈ min_dist_one_set →
-      (IsMinOn (fun B: Finset ℝ² => diam (B : Set ℝ²)) min_dist_one_set A) →
+      A.card = n → HasMinDist1 A →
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/99.lean
```diff
@@ -37,11 +38,14 @@ def HasMinDist1 (A : Finset ℝ²) : Prop :=
...
-/-- For sufficiently large n, is it the case that any set of n points with minimum distance 1 that minimizes diameter must contain an equilateral triangle of side length 1? -/
+/-- For sufficiently large n, is it the case that any set of n points with minimum distance $1$
+that minimizes diameter must contain an equilateral triangle of side length 1? -/
 @[category research open, AMS 52]
 theorem erdos_99 :
     answer(sorry) ↔ ∀ᶠ n in Filter.atTop, ∀ A : Finset ℝ²,
-      (IsMinOn (fun B: Finset ℝ² => diam (B : Set ℝ²)) {B : Finset ℝ² | B.card = n ∧ HasMinDist1 B} A) →
+      letI min_dist_one_set := {B : Finset ℝ² | B.card = n ∧ HasMinDist1 B}
+      A ∈ min_dist_one_set →
+      (IsMinOn (fun B: Finset ℝ² => diam (B : Set ℝ²)) min_dist_one_set A) →
```
smmercuri: I guess we could avoid `letI` as follows. Personally I find it harder to parse with the `letI` in there, but up to you which you prefer!
```suggestion
      A.card = n → HasMinDist1 A →
      (IsMinOn (fun B: Finset ℝ² => diam (B : Set ℝ²)) {B : Finset ℝ² | B.card = n ∧ HasMinDist1 B} A) →
```

smmercuri (approved):

FormalConjectures/ErdosProblems/99.lean
mo271: That's nicer!

PR #2039, opened 2026-02-02
fix(ErdosProblems/1084): `le` not `lt` in `IsSeparated`

mo271:


FormalConjecturesForMathlib/Topology/MetricSpace/MetricSeparated.lean
```diff
@@ -0,0 +1,34 @@
...
+
+/-!
+### Metric-separated sets
+
+In this section we define the predicate `Metric.IsSeparated'` for `ε`-separated sets.
+-/
+
+/-- A set `s` is `≥ ε`-separated if its elements are pairwise at distance greater or equal to
+ `ε` from each other. -/
+def IsSeparated' (ε : ℝ≥0∞) (s : Set X) : Prop := s.Pairwise (ε ≤ edist · ·)
```
smmercuri, line 34: What's the reasoning behind this change from `<` to `≤`? Is it possible that the main issues in the theorems were the typos from `6` to `3`, and `n > 0` conditions, or is `≤` also required?

mo271: `≤` is probably the main problem, without it is is easy to disprove `erdos_1084_upper_d1` and some other, while with the corrected definition AlphaProof can show `erdos_1084_upper_d1`.
It also make sense: even in dimension 1 with two points: if you require `<` then there won't be a pair of distance `= 1`, right?

smmercuri: right, of course!

mo271:
https://github.com/leanprover-community/mathlib4/pull/34761

mo271:
the typo for the quadratic formula fixed here seems to be also have updated on https://www.erdosproblems.com/forum/thread/1084 now

PR #2133, opened 2026-02-03
fix(ErdosProblems): 886, fix TODO and update question

danielchin:
This is a follow up to https://github.com/google-deepmind/formal-conjectures/pull/1942#discussion_r2749752508

It looks like T. Bloom updated https://www.erdosproblems.com/886 to correct for the mistake that @mo271 found and added a new theorem.


mo271 (approved):
Thanks for following up on this!!

PR #2141, opened 2026-02-03
chore(FormalConjecturesForMathlib): remove `Set.interIcc` and `Set.interIio`

Paul-Lez:


FormalConjecturesForMathlib/Data/Set/Interval.lean
```diff
@@ -0,0 +1,38 @@
...
+
+noncomputable instance [LocallyFiniteOrderBot β] : Fintype (S ∩ Iio b : Set β) :=
+  Set.finite_inter_Iio.fintype
+
+theorem finite_inter_Icc [LocallyFiniteOrder β] [OrderBot β] :
+    (S ∩ Icc a b).Finite :=
+  Set.finite_Icc a b |>.inter_of_right S
+
+noncomputable instance [LocallyFiniteOrder β] [OrderBot β] : Fintype (S ∩ Icc a b : Set β) :=
+  Set.finite_inter_Icc.fintype
```
YaelDillies: Maybe inline? These will definitely not be accepted in mathlib and we're only one of them once

Paul-Lez: Inlined both theorems

mo271 (approved):
looks good, but needs some merging to pass the ci

PR #1906, opened 2026-01-28
feat(Kourovka): problem 20.76

robryk:

PR #2142, opened 2026-02-03
Added pro-orderable simple group conjecture (Kourovka 1.35c)

### LINKED ISSUE #1914: Kourovka 1.35
### What is the conjecture
The problem **"Do there exist simple pro-orderable groups?"** is a well-known unsolved problem in group theory. It appears as **Problem 1.35(c)** in the **Kourovka Notebook** (a famous collection of open problems in group theory), originally posed by **A. I. Mal'cev** and **L. Fuchs** in the 1960s.

The problem remains **open** (unsolved) as of the latest editions of the Kourovka Notebook (including the 2026 draft).

### Details and Context
*   **Definition:** A group is called **pro-orderable** (or sometimes an **$O^*$-group**) if every partial ordering of the group can be extended to a linear (total) ordering. In this context, "partial ordering" refers to a bi-invariant partial order (invariant under both left and right multiplication). Consequently, a pro-orderable group must be **bi-orderable** (also called "orderable" or an "O-group").
*   **Connection to Simple Groups:** The existence of **simple bi-orderable groups** is a major open problem in itself. Since every pro-orderable group is bi-orderable, a positive answer to Problem 1.35 would imply the existence of simple bi-orderable groups. Conversely, if simple bi-orderable groups do not exist, then simple pro-orderable groups do not exist either.
*   **Research Status:**
    *   The problem is discussed in literature concerning ordered groups, particularly in the works of **V. M. Kopytov**, **A. H. Rhemtulla**, and others.
    *   It is known that every **abelian torsion-free group** is pro-orderable.
    *   There are results characterizing pro-orderable groups within certain classes (e.g., nilpotent or solvable groups), but the case for simple groups remains elusive due to the difficulty of constructing simple bi-orderable groups.
    *   Researchers have constructed "O-simple" groups (groups with no non-trivial convex normal subgroups) that satisfy certain ordering properties, but these are not "simple" in the standard group-theoretic sense (having no non-trivial normal subgroups).

In summary, the problem is a specific and stronger variation of the famous open question regarding the existence of simple bi-orderable groups, and it is actively listed as unsolved in the primary research problem collection for the field.

### Prerequisites needed
Note, the following is from Gemini, so double-check.

To state Kourovka Problem 1.35 in Lean 4 using Mathlib, the following specific definitions are currently **missing** and would need to be defined by the user:

1.  **A bundled type for bi-invariant partial orders:** Mathlib handles orders (like `PartialOrder` and `LinearOrder`) primarily as *typeclasses* on the type itself (e.g., `[PartialOrder G]`). You cannot quantify over typeclasses directly (i.e., you cannot say `∀ [PartialOrder G]`). To state "for *every* partial ordering...", you must define a predicate or structure for a binary relation `r : G → G → Prop` that satisfies the partial order axioms and the bi-invariance property ($x \le y \implies gx \le gy \land xg \le yg$).
2.  **The definition of `ProOrderable`:** The specific property of a group being "pro-orderable" (or an $O^*$-group) is not defined in Mathlib.

The basic building blocks, however, **are present**:
*   `Group` (the structure of the group).
*   `IsSimpleGroup` (the predicate for simplicity).
*   `IsPartialOrder` and `IsLinearOrder` (predicates for binary relations).

### Lean 4 Code to State the Problem

Here is the code demonstrating the definitions you would need to add to state the question.

```lean
import Mathlib.Algebra.Order.Group.Defs
import Mathlib.GroupTheory.SimpleGroup.Basic

variable (G : Type*) [Group G]

/-- 
A relation `r` on a group `G` is bi-invariant if it is preserved under 
left and right multiplication. 
This definition is necessary because we need to quantify over *all* such relations,
rather than using the standard `[OrderedGroup G]` typeclass.
-/
def IsBiInvariant (r : G → G → Prop) : Prop :=
  (∀ g a b : G, r a b → r (g * a) (g * b)) ∧ 
  (∀ g a b : G, r a b → r (a * g) (b * g))

/--
A group is `ProOrderable` if every bi-invariant partial order extends to a 
bi-invariant linear order.
-/
def ProOrderable : Prop :=
  ∀ (r : G → G → Prop), 
    IsPartialOrder r → IsBiInvariant G r → 
    ∃ (s : G → G → Prop), 
      IsLinearOrder s ∧ IsBiInvariant G s ∧ (∀ x y, r x y → s x y)

/-- 
The Statement of Kourovka 1.35(c): 
"Do there exist simple pro-orderable groups?"
-/
def Kourovka_1_35_c : Prop :=
  ∃ (G : Type), Group G ∧ IsSimpleGroup G ∧ ProOrderable G
```

### Summary of Missing Definitions
*   `IsBiInvariant`: You need to define what it means for a specific relation to be compatible with the group operation (both left and right).
*   `ProOrderable`: You need to define the property that allows the extension of partial orders to linear orders.

* ams-16

### Choose either option

vikasprajapati-01:
Adds relation-level definitions for bi-invariant orders and pro-orderable groups. This lets the statement quantify over all partial orders without fixing an order typeclass.

States Kourovka Notebook Problem 1.35(c) as an open question: does a simple pro-orderable group exist?

Checked with `lake --wfail build FormalConjectures.Kourovka.«1_35c»`.

Fixes #1914


FormalConjectures/Kourovka/1_35c.lean
mo271, line 1: Please make sure to check if `IsOrderedMonoid` can be used here! It seems you are unnecessarily redefining things. 


Let's follow the convention for namespace, files for kourovka etc established in https://github.com/google-deepmind/formal-conjectures/pull/1906

vikasprajapati-01: `IsOrderedMonoid` could have been used here but I kept it relation-level because we need to quantify over all bi-invariant partial orders and extend them. `IsOrderedMonoid` fixes a single order via typeclass, so it doesn’t express that quantification.
The file now follows the Kourovka conventions from #1906 (path/namespace).

FormalConjectures/Kourovka/1_35c.lean
```diff
@@ -0,0 +1,58 @@
...
+    ∃ s : G → G → Prop,
+      IsLinearOrder G s ∧ IsBiInvariant (G := G) s ∧ ∀ x y, r x y → s x y
+
+/--
+Kourovka Notebook Problem 1.35(c): does there exist a simple pro-orderable
+(group with a bi-invariant order extending every bi-invariant partial order)?
+-/
+@[category research open, AMS 16]
+theorem kourovka_simple_proOrderable_group :
+    ∃ (G : Type*) (_ : Group G), IsSimpleGroup G ∧ ProOrderable G := by
```
YaelDillies, line 55:
```suggestion
    ∃ (G : Type) (_ : Group G), IsSimpleGroup G ∧ ProOrderable G := by
```
is (meta)equivalent to your statement, but simpler

FormalConjectures/Kourovka/1_35c.lean
```diff
@@ -0,0 +1,58 @@
...
+/--
+`ProOrderable G` means every bi-invariant partial order extends to a
+bi-invariant linear order.
+-/
+def ProOrderable (G : Type*) [Group G] : Prop :=
+  ∀ r : G → G → Prop,
+    IsPartialOrder G r →
+    IsBiInvariant (G := G) r →
+    ∃ s : G → G → Prop,
+      IsLinearOrder G s ∧ IsBiInvariant (G := G) s ∧ ∀ x y, r x y → s x y
```
YaelDillies, line 47:
```suggestion
      IsLinearOrder G s ∧ IsBiInvariant G s ∧ ∀ x y, r x y → s x y
```

FormalConjectures/Kourovka/1_35c.lean
YaelDillies: I agree that `IsOrderedMonoid` can't conveniently be used here

Paul-Lez (approved):

PR #2151, opened 2026-02-03
fix(ErdosProblems/1074): definition of `PillaiPrimes`

mo271:


FormalConjectures/ErdosProblems/1074.lean
```diff
@@ -30,7 +30,19 @@ abbrev Nat.EHSNumbers : Set ℕ := {m | 1 ≤ m ∧ ∃ p, p.Prime ∧ ¬p ≡ 1
 
 /-- The Pillai primes are those primes $p$ such that there exists an $m$ with
```
smmercuri: Could you also update the docstring with the condition on `m`

FormalConjectures/ErdosProblems/1074.lean
```diff
@@ -30,7 +30,19 @@ abbrev Nat.EHSNumbers : Set ℕ := {m | 1 ≤ m ∧ ∃ p, p.Prime ∧ ¬p ≡ 1
...
 $p\not\equiv 1\pmod{m}$ such that $m! + 1 \equiv 0\pmod{p}$-/
```
mo271:
```suggestion
/-- The Pillai primes are those primes $p$ such that there exists an $m \ge 1$ with
$p\not\equiv 1\pmod{m}$ such that $m! + 1 \equiv 0\pmod{p}$-/
```

FormalConjectures/ErdosProblems/1074.lean
mo271: good point, done!

smmercuri (approved):

PR #2154, opened 2026-02-03
feat(Wikipedia): Leinster Groups Conjecture

### LINKED ISSUE #2148: Are there infinitely many Leinster groups?
### What is the conjecture

 A finite group $G$ is a **Leinster group** if the sum of the orders of all its subgroups equals the square of the group's order:
$$\sum_{H \leq G} |H| = |G|^2$$

**Conjecture:** Are there infinitely many Leinster groups?

This is a problem in computational finite group theory asking whether the class of Leinster groups is infinite or finite.

 **Sources:**
 - https://en.wikipedia.org/wiki/Leinster_group

### Prerequisites needed

**Formalizability Rating:** 1/5 (0 is best) (as of 2026-02-03)

- `Fintype G` and `Subgroup G` from Mathlib for finite groups and their subgroups
- `Fintype.card` for computing the order of a group

- Definition of a Leinster group as a predicate on finite groups (not in Mathlib)
- Formalization of the set/class of all Leinster groups to express the infinitude conjecture

Rating justification (1-2 sentences): The core mathematical objects (finite groups, subgroups, cardinality) are well-established in Mathlib, so the statement can be written quickly with just a simple definition. Only a straightforward definitional wrapper is needed to introduce Leinster groups as a predicate on finite groups.

* ams-20

### Choose either option

---

franzhusch:
Closes #2148

### DIFF
--- a/FormalConjectures/Wikipedia/LeinsterGroup.lean
+++ b/FormalConjectures/Wikipedia/LeinsterGroup.lean
@@ -56,7 +56,7 @@ Formalized via the negation of "Does there exist an n such that all Leinster gro
 order less than n".
 -/
 @[category research open, AMS 20]
-theorem infinitely_many_leinster_groups :
+theorem infinitely_many_leinster_groups : answer(sorry) ↔
     ¬∃ n : ℕ, ∀ G : Type, ∀ (_ : Group G) (_ : Fintype G),
       IsLeinster G → Fintype.card G < n := by
   sorry


FormalConjectures/Wikipedia/LeinsterGroup.lean
```diff
@@ -0,0 +1,106 @@
...
+namespace LeinsterGroup
+
+open scoped Classical
+
+/--
+A finite group `G` is a **Leinster group** if the sum of the orders of all its normal subgroups
+equals twice the group's order.
+-/
+def IsLeinster (G : Type*) [Group G] [Fintype G] : Prop :=
+  ∑ H : {H : Subgroup G // H.Normal}, Nat.card H.val = 2 * Fintype.card G
```
mo271:
```suggestion
  ∑ H : {H : Subgroup G // H.Normal}, Nat.card H = 2 * Fintype.card G
```
should also work. Also you could take `Fintype.card` for `H` as well?

FormalConjectures/Wikipedia/LeinsterGroup.lean
```diff
@@ -0,0 +1,106 @@
...
+/--
+**Conjecture:** Are there infinitely many Leinster groups?
+
+This asks whether there exist infinitely many (non-isomorphic) finite groups that are
+Leinster groups.
+-/
+@[category research open, AMS 20]
+theorem infinitely_many_leinster_groups :
+    answer(sorry) ↔ ∀ n : ℕ, ∃ G : Type, ∃ (_ : Group G) (_ : Fintype G),
+      Fintype.card G > n ∧ IsLeinster G := by
```
mo271: I think this doesn't take the non-isomorphic into account. Instead I would just go with the formulation that implicitly uses that there are only finitely many groups of a fixed order: hence it is equivalent to ask if there is infinitely many `n` and a group of, so perhaps something like:
```lean
theorem infinitely_many_leinster_groups :
    ∃ᶠ n  in Filter.atTop, ∃ G : Type, ∃ (_ : Group G) (_ : Fintype G),
      Fintype.card G = n ∧ IsLeinster G := by
```

FormalConjectures/Wikipedia/LeinsterGroup.lean
```diff
@@ -0,0 +1,106 @@
...
+In particular, the existence of odd perfect numbers is equivalent to the existence of
+dihedral Leinster groups.
+
+Reference: Leinster, Tom (2001). "Perfect numbers and groups".
+-/
+@[category research solved, AMS 20]
+theorem dihedral_is_leinster_iff_odd_perfect (n : ℕ) [NeZero n] :
+    IsLeinster (DihedralGroup n) ↔ Nat.Perfect n ∧ Odd n := by
+  sorry
+
```
mo271, line 115: please add a TODO for the other properties mentioned on the wikipedia page: 
```
There are no Leinster groups that are symmetric or alternating.[3]
There is no Leinster group of order p2q2 where p, q are primes.[1]
No finite semi-simple group is Leinster.[1]
No p-group can be a Leinster group.[4]
All abelian Leinster groups are cyclic with order equal to a perfect number.[3]
```

mo271 (commented):
Looks good, some initial comments!

FormalConjectures/Wikipedia/LeinsterGroup.lean
franzhusch: Fintype is not needed because subgroup and Fintype G guanrantee that H is finite, so the definition is fine. Or do you want it for consistency? I dont know if it makes the definition more verbose for Lean as it would require the property of finiteness for H.

FormalConjectures/Wikipedia/LeinsterGroup.lean
franzhusch: This was also my plan with the formalization essentially "Are there infinite n such that we can find a Leinster Group of order bigger than n". But maybe the use of answer is unnecessary here, I should look more into the correct use of answer() I think...

franzhusch: Maybe even easier to read and formalize is "Does there exist a n, such that all Leinster Groups have order less than n".

franzhusch: I added that formalization now, should be simpler.

FormalConjectures/Wikipedia/LeinsterGroup.lean
```diff
@@ -0,0 +1,116 @@
...
+**Conjecture:** Are there infinitely many Leinster groups?
+
+This asks whether there exist infinitely many (non-isomorphic) finite groups that are
+Leinster groups.
+
+Formalized via the negation of "Does there exist an n such that all Leinster groups have
+order less than n".
+-/
+@[category research open, AMS 20]
+theorem infinitely_many_leinster_groups :
```
mo271:
```suggestion
theorem infinitely_many_leinster_groups : answer(sorry) ↔
```

no matter if we do it via the negation, it is still a question, hence should use the `answer(sorry)`

FormalConjectures/Wikipedia/LeinsterGroup.lean
mo271: I find the positive formulation more natural, but no strong opinion...

FormalConjectures/Wikipedia/LeinsterGroup.lean
mo271: not important

PR #2158, opened 2026-02-04
feat(ErdosProblems): 394

### LINKED ISSUE #2121: Erdős Problem 394: Asymptotic bounds for divisibility sums over consecutive integers
### What is the conjecture

 Let $t_k(n)$ denote the minimum value $m$ such that $n$ divides the product $m(m+1)(m+2)\cdots(m+k-1)$.

**Question 1:** Does $\sum_{n\leq x}t_2(n) \ll \frac{x^2}{(\log x)^c}$ hold for some $c>0$?

**Question 2:** For $k \geq 2$, is $\sum_{n\leq x}t_{k+1}(n) = o\left(\sum_{n\leq x}t_k(n)\right)$?

Erdős and Hall proved that $\sum_{n\leq x}t_2(n) \ll \frac{\log\log\log x}{\log\log x}x^2$, and they conjectured the sum equals $o(x^2/(\log x)^c)$ for any $c < \log 2$. A lower bound of $\sum_{n\leq x}t_2(n) \gg \frac{x^2}{\log x}$ follows from the fact that $t_2(p)=p-1$ for primes $p$.

 **Sources:**
 - https://www.erdosproblems.com/394

### Prerequisites needed

**Formalizability Rating:** 2/5 (0 is best) (as of 2026-02-02)

- Basic number-theoretic functions and divisibility properties available in Mathlib
- Asymptotic notation ($\ll$, $o$) and summatory function sums can be expressed
- Consecutive product constructions: $m(m+1)\cdots(m+k-1)$ is standard

- Formalization of $t_k(n)$ as the minimal element satisfying a divisibility property requires defining a specialized function and proving minimality properties
- Asymptotic bounds framework for summatory functions over specific arithmetic sequences; the exact relationship between growth rates of sums may require new lemmas

Rating justification (1-2 sentences): The statement itself can be written using existing Mathlib concepts for summation, divisibility, and asymptotic notation, though $t_k(n)$ requires a careful definition as a minimum over a predicate. The formalization of the question is straightforward, but proving properties about this function would require development of divisibility properties over consecutive integers.

* ams-11

### Choose either option

---

danielchin:
Closes #2121 

Formalizes Erdos Problem 394: https://www.erdosproblems.com/394

This one was a bit of a doozy since there's so many subcomponents, but maybe it's good fodder for AlphaProof.

### DIFF
--- a/FormalConjectures/ErdosProblems/394.lean
+++ b/FormalConjectures/ErdosProblems/394.lean
@@ -87,34 +87,8 @@ theorem erdos_394_lower_bound :
   sorry
 
 /--
-Erdős and Hall [ErHa78] note that $t_{n-1}(n!)=2$.
--/
-@[category research solved, AMS 11]
-theorem erdos_394_factorial_n_minus_1 :
-    ∀ (n : ℕ), n > 1 →
-    sInf { m : ℕ | 0 < m ∧ (n !) ∣ ∏ i ∈ range (n - 1), (m + i) } = 2 := by
-  sorry
-
-/--
-Erdős and Hall [ErHa78] note that $t_{n-2}(n!)\ll n$.
--/
-@[category research solved, AMS 11]
-theorem erdos_394_factorial_n_minus_2_upper :
-    (fun (n : ℕ) ↦ ((sInf { m : ℕ | 0 < m ∧ (n !) ∣ ∏ i ∈ range (n - 2), (m + i) } : ℕ) : ℝ)) =O[atTop]
-    (fun n ↦ (n : ℝ)) := by
-  sorry
-
-/--
-$n=2^r$ shows that $t_{n-2}(n!)\ll n$ is the best possible (implying the lower bound is also linear).
--/
-@[category research solved, AMS 11]
-theorem erdos_394_factorial_n_minus_2_lower_pow2 :
-    (fun (r : ℕ) ↦ (2 ^ r : ℝ)) =O[atTop]
-    (fun r ↦ ((sInf { m : ℕ | 0 < m ∧ (2 ^ r) ! ∣ ∏ i ∈ range (2 ^ r - 2), (m + i) } : ℕ) : ℝ)) := by
-  sorry
-
-/--
-Is it true that, for infinitely many $n$, $t_k(n!) < t_{k-1}(n!) - 1$ for all $1 \leq k < n$?
+They ask about the behaviour of $t_{n-3}(n!)$ and also ask ask whether, for infinitely many $n$,
+$t_k(n!)< t_{k-1}(n!)-1$ for all $1\leq k < n$.
 -/
 @[category research open, AMS 11]
 theorem erdos_394_factorial_gap_conjecture :
--- a/FormalConjectures/ErdosProblems/394.lean
+++ b/FormalConjectures/ErdosProblems/394.lean
@@ -33,15 +33,19 @@ open scoped Asymptotics Topology Nat
 namespace Erdos394
 
 /--
-Let $t_k(n)$ denote the least $m$ such that $n\mid m(m+1)(m+2)\cdots (m+k-1).$ Is it true that
-$\sum_{n\leq x}t_2(n)\ll \frac{x^2}{(\log x)^c}$ for some $c>0$?
+Let $t_k(n)$ denote the least $m$ such that $n\mid m(m+1)(m+2)\cdots (m+k-1).$
+-/
+noncomputable def t (k n : ℕ) : ℕ :=
+  sInf { m : ℕ | 0 < m ∧ n ∣ ∏ i ∈ range k, (m + i) }
+
+/--
+Is it true that $\sum_{n\leq x}t_2(n)\ll \frac{x^2}{(\log x)^c}$ for some $c>0$?
 -/
 @[category research open, AMS 11]
 theorem erdos_394_part_a :
     answer(sorry) ↔
       ∃ c > 0, (fun x ↦ ∑ n ∈ Ioc 1 ⌊x⌋₊,
-      ((sInf { m : ℕ | 0 < m ∧ n ∣ m * (m + 1) } : ℕ) : ℝ)) =O[atTop]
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/394.lean
```diff
@@ -0,0 +1,111 @@
...
+  theory. Monographies de L'Enseignement Mathematique (1980).
+- [ErHa78] Erdős, P. and Hall, R. R., On some unconventional problems on the divisors of integers.
+  J. Austral. Math. Soc. Ser. A (1978), 479--485.
+-/
+
+open Nat Filter Finset
+open scoped Asymptotics Topology Nat
+
+namespace Erdos394
+
```
mo271, line 34: I think it would be nice here to extract out the definition of $t_k(n)$ here, since it is used many times

FormalConjectures/ErdosProblems/394.lean
```diff
@@ -0,0 +1,111 @@
...
+/--
+Let $t_k(n)$ denote the least $m$ such that $n\mid m(m+1)(m+2)\cdots (m+k-1).$ Is it true that
+$\sum_{n\leq x}t_2(n)\ll \frac{x^2}{(\log x)^c}$ for some $c>0$?
+-/
+@[category research open, AMS 11]
+theorem erdos_394_part_a :
+    answer(sorry) ↔
+      ∃ c > 0, (fun x ↦ ∑ n ∈ Ioc 1 ⌊x⌋₊,
+      ((sInf { m : ℕ | 0 < m ∧ n ∣ m * (m + 1) } : ℕ) : ℝ)) =O[atTop]
```
mo271: can use `≪` here (and in other place where it is also used on erdosproblems.com
we also have  `≫` ...

FormalConjectures/ErdosProblems/394.lean
```diff
@@ -0,0 +1,111 @@
...
+Since $t_2(p)=p-1$ for prime $p$ it is trivial that $\sum_{n\leq x}t_2(n)\gg \frac{x^2}{\log x}$.
+-/
+@[category research solved, AMS 11]
+theorem erdos_394_lower_bound :
+    (fun x ↦ x ^ 2 / Real.log x) =O[atTop]
+    (fun x ↦ ∑ n ∈ Ioc 1 ⌊x⌋₊, ((sInf { m : ℕ | 0 < m ∧ n ∣ m * (m + 1) } : ℕ) : ℝ)) := by
+  sorry
+
+/--
+They ask about the behaviour of $t_{n-3}(n!)$ and also ask ask whether, for infinitely many $n$,
```
mo271:
```suggestion
They ask about the behaviour of $t_{n-3}(n!)$ and also ask whether, for infinitely many $n$,
```

mo271 (commented):
Thanks look pretty good!
A few comments...

FormalConjectures/ErdosProblems/394.lean
```diff
@@ -0,0 +1,111 @@
...
+$t_k(n!)< t_{k-1}(n!)-1$ for all $1\leq k < n$.
+-/
+@[category research open, AMS 11]
+theorem erdos_394_factorial_gap_conjecture :
+    answer(sorry) ↔
+      Set.Infinite { n : ℕ | ∀ k, 2 ≤ k → k < n →
+      sInf { m : ℕ | 0 < m ∧ (n !) ∣ ∏ i ∈ range k, (m + i) } <
+      sInf { m : ℕ | 0 < m ∧ (n !) ∣ ∏ i ∈ range (k - 1), (m + i) } - 1 } := by
```
mo271: probably fine here, becuase the sInf is never `0`, but out of habit, I would consider putting `- 1` on the other side of the equality?  On the other this way it is closer to the informal text..

FormalConjectures/ErdosProblems/394.lean
danielchin: Pulled this out into a def.

FormalConjectures/ErdosProblems/394.lean
danielchin: Replaced the `\ll` and `\gg` respectively

FormalConjectures/ErdosProblems/394.lean
danielchin: Done.

FormalConjectures/ErdosProblems/394.lean
danielchin: I left this one as is for now...

I've been getting different feedback from reviewers regarding `- 1`. In general, I try to turn it into `+ 1` to keep things safe, but I've also gotten comments on keeping it as close to the original question as possible.

So my general heuristic has been to it keep close to the original problem if it's safe, if not, then modify it.

FormalConjectures/ErdosProblems/394.lean
```diff
@@ -0,0 +1,114 @@
...
+noncomputable def t (k n : ℕ) : ℕ :=
+  sInf { m : ℕ | 0 < m ∧ n ∣ ∏ i ∈ range k, (m + i) }
+
+/--
+Is it true that $\sum_{n\leq x}t_2(n)\ll \frac{x^2}{(\log x)^c}$ for some $c>0$?
+-/
+@[category research open, AMS 11]
+theorem erdos_394_part_a :
+    answer(sorry) ↔
+      ∃ c > 0, (fun x ↦ ∑ n ∈ Ioc 1 ⌊x⌋₊,
```
mo271: Why letting the sum start at `2` here? I see that `t k 0 = 0`, so that doesn't make a difference. We probably have `t k 1 = 1` for all `k`, so compared to the version on erdosproblems.com you are loosing an additional summand `1`. It probably doesn't make a difference in the asymptotic statement, but why leave out the first summand? How about replacing `Ioc 1` with `Icc 1` or `Icc 0` everywhere?

danielchin: Oops, this is a mistake from my original commit. Swapped over to use `Icc 1` to match the statement more accurately.

mo271:
Thanks

mo271 (approved):

PR #2164, opened 2026-02-04
Add Herzog–Schönheim conjecture alias

### LINKED ISSUE #1956: Herzog-Schönheim conjecture
### What is the conjecture

 Let $G$ be a finite group and let $H_1, H_2, \ldots, H_s$ be subgroups of $G$ with indices $[G:H_i] = d_i$ for $i = 1, 2, \ldots, s$. Suppose the left cosets of these subgroups form a partition of $G$ into exactly $s$ parts, i.e., $G = \bigsqcup_{i=1}^{s} H_i a_i$ for some elements $a_i \in G$, where the union is disjoint. The Herzog–Schönheim conjecture states that if $s > 1$, then not all indices $d_1, d_2, \ldots, d_s$ can be distinct: at least two of them must be equal.

 **Sources:**
 - https://en.wikipedia.org/wiki/Herzog%E2%80%93Sch%C3%B6nheim_conjecture, https://arxiv.org/abs/1803.08301, https://arxiv.org/abs/1803.03569, https://pmc.ncbi.nlm.nih.gov/articles/PMC7247885/, https://arxiv.org/abs/1804.11103

### Prerequisites needed

**Formalizability Rating:** 1/5 (0 is best) (as of 2026-01-31)

Building blocks (from Mathlib):
- `Subgroup G`: subgroup type, standard in group theory
- `Subgroup.index`: function computing the index of a subgroup
- `Fintype`: finite types for working with finite groups and cosets
- Coset and partition definitions: available through `Subgroup.leftCosets` and lattice of sets

Missing pieces:
- Formal definition of "coset partition" (union of disjoint left cosets of multiple subgroups) as a standalone structure
- Auxiliary lemmas connecting partitions to index distinctness

Rating justification: All core group-theoretic objects (subgroups, indices, cosets) already exist in Mathlib. The statement can be formalized directly using existing `Fintype`, `Subgroup`, and set-theoretic concepts; only minor connecting definitions may be needed.

* ams-20

### Choose either option

---

Abhishek242004:
This is an alias for FormalConjectures/ErdosProblems/274.lean 

The alias uses `type_of%` to refer to the canonical statement in
`FormalConjectures/ErdosProblems/274.lean`, without duplicating any logical content. It has the same parameters and assumptions which are mapped to the parameters in the original theorem. Users can work with the Herzog–Schönheim conjecture using this file directly.

Closes #1956

### DIFF
--- a/FormalConjectures/Wikipedia/HerzogSchonheimConjecture.lean
+++ b/FormalConjectures/Wikipedia/HerzogSchonheimConjecture.lean
@@ -32,6 +32,7 @@ The actual formalization is in `FormalConjectures.ErdosProblems.«274»`.
 namespace HerzogSchonheimConjecture
 
 @[category research open, AMS 20]
-theorem herzog_schonheim_conjecture : type_of% @Erdos274.herzog_schonheim := by sorry
+theorem herzog_schonheim_conjecture : type_of% @Erdos274.herzog_schonheim := by 
+  sorry
 
 end HerzogSchonheimConjecture


mo271:
please look at the other files that use `type_of%` and do it analogously

abhi2oo4:
> please look at the other files that use `type_of%` and do it analogously

Thanks for the pointer. I looked at the existing uses of type_of% (e.g. ErdosProblems/1135.lean) and tried to apply the same minimal alias pattern here.

Screenshot from 2026-02-05 03-24-15

But I am getting the following error about the meta - variables where it cannot resolve `[Group G]` / `[Fintype ι]` instances, so it cannot infer how the parameters should be
mapped to those of the original theorem. This works in the case of `1135.lean`, likely because the statement does not involve
typeclass-dependent parameters, so there is no ambiguity for Lean to resolve.

Screenshot from 2026-02-05 03-23-51


Because of this, I see two reasonable options:

Keep the current version with explicit parameters, using type_of% + simpa, which avoids the metavariable issue.

Switch to a just import style (no theorem) like it is used in `FormalConjectures/Wikipedia/HappyEndingProblem.lean`

Please let me know what option would you prefer.

hanwenzhu:
Try `type_of% @Erdos274.herzog_schonheim`?

abhi2oo4:
Thanks a lot. It did work

FormalConjectures/Wikipedia/HerzogSchonheimConjecture.lean
```diff
@@ -0,0 +1,37 @@
...
+- [arXiv:1803.08301](https://arxiv.org/abs/1803.08301)
+- [arXiv:1803.03569](https://arxiv.org/abs/1803.03569)
+- [PMC7247885](https://pmc.ncbi.nlm.nih.gov/articles/PMC7247885/)
+- [arXiv:1804.11103](https://arxiv.org/abs/1804.11103)
+-/
+
+namespace HerzogSchonheimConjecture
+
+@[category research open, AMS 20]
+theorem herzog_schonheim_conjecture : type_of% @Erdos274.herzog_schonheim := by sorry
```
mo271:
```suggestion
theorem herzog_schonheim_conjecture : type_of% @Erdos274.herzog_schonheim := by 
  sorry
```

PR #2169, opened 2026-02-05
feat(ErdosProblems): 507

### LINKED ISSUE #760: Erdős Problem 507
### What is the conjecture
    
https://www.erdosproblems.com/507

  Let $\alpha(n)$ be such that every set of $n$ points in the unit disk contains three points which determine a triangle of area at most $\alpha(n)$. Estimate $\alpha(n)$.
  

Status: open

### Choose either option

jeangud:
# Description
Let $\alpha(n)$ be such that every set of $n$ points in the unit disk contains three points which determine a triangle of area at most $\alpha(n)$. Estimate $\alpha(n)$.

- Closes #760
- Added additional known bounds from the literature
- Tried to leverage `Finset` as much as possible to ligthen the proof burden

# Testing
:white_check_mark: Builds successfully
```shell
$ lake build FormalConjectures/ErdosProblems/507.lean 
Build completed successfully.
```

### DIFF
--- a/FormalConjectures/ErdosProblems/507.lean
+++ b/FormalConjectures/ErdosProblems/507.lean
@@ -32,7 +32,7 @@ import FormalConjectures.Util.ProblemImports
 -/
 
 open Asymptotics Filter Topology
-open scoped Classical EuclideanGeometry
+open scoped EuclideanGeometry
 
 namespace Erdos507
 
@@ -48,7 +48,7 @@ $\alpha(n)$ is the supremum of `minTriangleArea S` over all sets `S` of $n$ poin
 -/
 noncomputable def α (n : ℕ) : ℝ :=
   sSup (minTriangleArea '' { S : Finset ℝ² |
-    S.card = n ∧ ↑S ⊆ Metric.closedBall (0 : ℝ²) 1 ∧ ¬ Collinear ℝ (↑S : Set ℝ²) })
+    S.card = n ∧ ↑S ⊆ Metric.closedBall (0 : ℝ²) 1 ∧ ¬ Collinear ℝ (S : Set ℝ²) })
 
 /--
 Current best lower bound [KPS82].


jeangud:
Could an admin rename PR to **feat(ErdosProblems): 507** ? My bad, I did not check until after submitting :weary:

FormalConjectures/ErdosProblems/507.lean
```diff
@@ -0,0 +1,98 @@
...
+- [KPS81] Komlós, János, János Pintz, and Endre Szemerédi. "On Heilbronn's triangle problem."
+  Journal of the London Mathematical Society 2.3 (1981): 385-396.
+-/
+
+open Filter Topology
+open scoped EuclideanGeometry
+
+namespace Erdos507
+
+abbrev Point := EuclideanSpace ℝ (Fin 2)
```
smmercuri: I'm not sure there's a need for this `abbrev`, you can just use `EuclideanSpace ℝ (Fin 2)` where needed. In fact, since you've already `open scoped EuclideanGeometry` you can use the notation `ℝ²`

FormalConjectures/ErdosProblems/507.lean
```diff
@@ -0,0 +1,98 @@
...
+
+/-- The unit disk in the plane. -/
+def UnitDisk : Set Point := Metric.closedBall 0 1
```
smmercuri: Similar to my above comment, I think it's better to use `Metric.closedBall 0 1` where needed, especially in this case where it's only used once

FormalConjectures/ErdosProblems/507.lean
```diff
@@ -0,0 +1,98 @@
...
+
+/--
+The minimum area of a triangle determined by three distinct points in a set `S`.
+-/
+noncomputable def minTriangleArea (S : Finset Point) : ℝ :=
+  let triples := S ×ˢ S ×ˢ S
+  let distinctTriples := triples.filter fun ⟨p, q, r⟩ => p ≠ q ∧ q ≠ r ∧ p ≠ r
```
smmercuri: Certainly the problem requires the triangles to be non-degenerate. Unfortunately the condition that they are pairwise distinct is not quite enough, we also need that they are not collinear (otherwise triangle collapses to a line). You might find [`Affine.Triangle`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/LinearAlgebra/AffineSpace/Simplex/Basic.html#Affine.Triangle) useful from mathlib

FormalConjectures/ErdosProblems/507.lean
```diff
@@ -0,0 +1,98 @@
...
+noncomputable def α (n : ℕ) : ℝ :=
+  sSup (minTriangleArea '' { S : Finset Point | S.card = n ∧ ↑S ⊆ UnitDisk })
+
+/--
+Let $\alpha(n)$ be such that every set of $n$ points in the unit disk contains three points which
+determine a triangle of area at most $\alpha(n)$. Estimate $\alpha(n)$.
+-/
+@[category research open, AMS 51]
+theorem erdos_507 :
+    α = answer(sorry) := by
```
smmercuri: The phrase "Estimate foo" is always a bit tricky to formalise, but it surely is asking for asymptotic behaviour rather than exact behaviour. Similar to the solved variants below, one is interested in upper/lower bounds for sufficiently large `n`. I would suggest creating multiple versions of the main problem asserting different properties (with an `answer(sorry)` for the candidate function). For example:
1. `α` is asymptotically equivalent to `answer(sorry)`
2. `α` is `<< answer(sorry)` where `answer(sorry)` is asymptotically smaller than the best known upper bound
3. `α` is `>> answer(sorry)` where `answer(sorry)` is asymptotically bigger than the best known lower bound

mo271: look also for other examples in the repo how this can be done!

FormalConjectures/ErdosProblems/507.lean
jeangud: Good point, done

FormalConjectures/ErdosProblems/507.lean
jeangud: Thank you both for the pointers here. Done, drawing inspiration from [359.lean](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/359.lean) and [357.lean](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/357.lean)

FormalConjectures/ErdosProblems/507.lean
jeangud: Good catch! I actually struggled a bit to incorporate `Affine.Triangle` while keeping the tractability of the `Finset`.

I opted for `Collinear`, at the cost of including `Classical`:
```lean
noncomputable def minTriangleArea (S : Finset ℝ²) : ℝ :=
  let triples := S ×ˢ S ×ˢ S
  let validTriples := triples.filter fun ⟨p, q, r⟩ => p ≠ q ∧ q ≠ r ∧ p ≠ r ∧
    ¬ Collinear ℝ {p, q, r}
  let areas := validTriples.image fun ⟨p, q, r⟩ => EuclideanGeometry.triangle_area p q r
  (areas.min).getD 0 -- Return the minimum, defaulting to 0 if the set is empty (n < 3)
```

My other option was something like:
```lean
noncomputable def minTriangleArea (S : Finset ℝ²) : ℝ :=
  let triangles : Set (Affine.Triangle ℝ ℝ²) := { t | ∀ i, t.points i ∈ S }
  sInf ((fun t => EuclideanGeometry.triangle_area (t.points 0) (t.points 1) (t.points 2)) '' triangles)
```
which does not leverage the `Finset` that well...

Let me know if you have a preference?

FormalConjectures/ErdosProblems/507.lean
```diff
@@ -0,0 +1,122 @@
...
+    α ~[atTop] (answer(sorry) : ℕ → ℝ) := by
+  sorry
+
+/--
+Estimate a lower bound for$\alpha(n)$.
+-/
+@[category research open, AMS 51]
+theorem erdos_507.lower:
+    let ans := (answer(sorry) : ℕ → ℝ)
+    (lowerBest ≪ ans) ∧ (ans ≪ α) := by
```
smmercuri: The clause `lowerBest ≪ ans` means that this problem is not open, take `ans = lowerBest`. You will need to capture a strict notion of asymptotically improved magnitude. Note that if you use something `∀ᶠ n in atTop, lowerBest n < α n`, then this allows for improvements in constants which are somehow uninteresting. We should use little o here I think: `lowerBest =o[atTop] ans`

FormalConjectures/ErdosProblems/507.lean
```diff
@@ -0,0 +1,122 @@
...
+  sorry
+
+/--
+Estimate an upper bound for$\alpha(n)$.
+-/
+@[category research open, AMS 51]
+theorem erdos_507.upper:
+    let ans := (answer(sorry) : ℕ → ℝ)
+    (α ≪ ans) ∧ (ans ≪ upperBarrier) := by
```
smmercuri: Ditto previous comment on `ans << upperBarrier`, although perhaps this one is OK since it's not the true upper bound

FormalConjectures/ErdosProblems/507.lean
smmercuri: I think I prefer the second, mainly because it's better to make use of the existing API that has been tested in use. I suppose a third option, would be to use the `IsConvexPolygon` from the `ForMathlib` folder, that keeps the spirit of your original version. You might then say something like
```lean
(S ×ˢ S ×ˢ S).filter fun ⟨p, q, r⟩ => IsConvexPolygon ![p, q, r]

File: FormalConjectures/ErdosProblems/507.lean (hunk 5 above)
Comment (jeangud) [reply FormalConjectures/ErdosProblems/507.lean]: Oh yeah, I need to remember that edge cases are also a thing in Lean :sweat_smile: 
Thank you for catching that.

File: FormalConjectures/ErdosProblems/507.lean (hunk 3 above)
Comment (jeangud) [reply FormalConjectures/ErdosProblems/507.lean]: You got it, went for the second implementation with `Affine.Triangle` :+1:

File: FormalConjectures/ErdosProblems/507.lean [hunk 7]
```diff
@@ -0,0 +1,120 @@
...
+noncomputable def minTriangleArea (S : Finset ℝ²) : ℝ :=
+  let triangles : Set (Affine.Triangle ℝ ℝ²) := { t | ∀ i, t.points i ∈ S }
+  sInf ((fun t => EuclideanGeometry.triangle_area
+    (t.points 0) (t.points 1) (t.points 2)) '' triangles)
+
+/--
+$\alpha(n)$ is the supremum of `minTriangleArea S` over all sets `S` of $n$ points in the unit disk.
+-/
+noncomputable def α (n : ℕ) : ℝ :=
+  sSup (minTriangleArea '' { S : Finset ℝ² | S.card = n ∧ ↑S ⊆ Metric.closedBall (0 : ℝ²) 1 })
```
smmercuri: We also need to exclude "degenerate" sets `S` here, i.e., those sets `S` for which it is not possible to find a  triangle in. For example, take `S` to be any straight line inside the unit circle, then `minTriangleArea` is zero, so `α` is always zero.

FormalConjectures/ErdosProblems/507.lean
```diff
@@ -0,0 +1,120 @@
...
+
+namespace Erdos507
+
+/--
+The minimum area of a triangle determined by three distinct points in a set `S`.
+-/
+noncomputable def minTriangleArea (S : Finset ℝ²) : ℝ :=
+  let triangles : Set (Affine.Triangle ℝ ℝ²) := { t | ∀ i, t.points i ∈ S }
+  sInf ((fun t => EuclideanGeometry.triangle_area
+    (t.points 0) (t.points 1) (t.points 2)) '' triangles)
```
smmercuri: Btw, if you like, you could make use of set composition notation to make this a one-liner. But it's purely style so completely up to you which you prefer
```lean
noncomputable def minTriangleArea (S : Finset ℝ²) : ℝ :=
  sInf {EuclideanGeometry.triangle_area (t.points 0) (t.points 1) (t.points 2) | 
    (t : Affine.Triangle ℝ ℝ²) (_ : ∀ i, t.points i ∈ S)}

File: FormalConjectures/ErdosProblems/507.lean (hunk 7 above)
Comment (jeangud) [reply FormalConjectures/ErdosProblems/507.lean]: With the latest comment [here](https://github.com/google-deepmind/formal-conjectures/pull/2169#discussion_r2777292394), I believe we actually get $\text{minTriangleArea} = \text{sInf} (\varnothing) = \infty$? (we cannot find any `Affine.Triangle` in a problematic `S`, since `Affine.Triangle` already enforces for non-collinearity). However this is just as pathological because then $\alpha = \infty$.


This is where I liked [the formulation leveraging the `Finset`](https://github.com/google-deepmind/formal-conjectures/pull/2169#discussion_r2770909896) because it defaults to $0$ on pathological `S` inputs, not impacting the downstream `sSup` for $\alpha$. It also works indiscriminately for small sets $n < 3$, and for collinear sets and avoids doing the filtering in two places.

In any case, happy to go either way! To build on your last comments I just proposed the following with 663c955:
```lean
noncomputable def α (n : ℕ) : ℝ :=
  sSup (minTriangleArea '' { S : Finset ℝ² |
    S.card = n ∧ ↑S ⊆ Metric.closedBall (0 : ℝ²) 1 ∧ ¬ Collinear ℝ (↑S : Set ℝ²) })
```

FormalConjectures/ErdosProblems/507.lean
jeangud: I like it, added! Thank you

FormalConjectures/ErdosProblems/507.lean
smmercuri: In fact `sInf ∅ = 0` (see [`Real.sInf_empty`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Real/Archimedean.html#Real.sInf_empty)), because the reals in mathlib do not have an infinite element so it takes a junk value. 

To clarify: the issue is that if the input set `{ S : Finset ℝ² | S.card = n ∧ ↑S ⊆ Metric.closedBall (0 : ℝ²) 1 }` contains a "bad" set, then `minTriangleArea` will be `0`. So for any `n > 2` this function is constant at `0` because such a bad set always exists. This holds also in the `Finset` implementation, and indeed is caused precisely by a junk value of `0` for empty sets.

The proposed fix looks good to me. You can remove the coercion arrow from `S` inside `Collinear` and write `(S : Set ℝ²)`.

jeangud: Ah yes `sInf ∅ = 0` changes everything indeed, thank you for the clarification.

Coercion removed :heavy_check_mark:

PR #2178, opened 2026-02-06
feat(Arxiv/2602.05192): problem 4

mo271:


FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,92 @@
...
+$$
+c_k = \sum_{i+j=k} \frac{(n-i)! (n-j)!}{n! (n-k)!} a_i b_j
+$$
+for $k = 0, 1, \dots, n$.
+ -/
+noncomputable def finiteAdditiveConvolution (n : ℕ) (p q : F[X]) : F[X] :=
+  let c := fun k  => ∑ ij ∈ antidiagonal (k : ℕ),
+      ((n - ij.1)! * (n - ij.2)! : F) / (n ! * (n - k)! : F) *
+      (p.coeff (n - ij.1)) * (q.coeff (n - ij.2))
+  ∑ k ∈ range (n + 1),  (c k) • X^(n - k)
```
seewoo5:
```suggestion
  ∑ k ∈ range (n + 1), (c k) • X^(n - k)
```

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,92 @@
...
+
+/--
+For a monic polynomial $p(x)=\prod_{i\le n}(x- \lambda_i)$, define
+$$\Phi_n(p):=\sum_{i\le n}(\sum_{j\neq i} \frac1{\lambda_i-\lambda_j})^2$$
+and $\Phi_n(p):=\infty$ if $p$ has a multiple root.
+-/
+noncomputable def Φ (n : ℕ) (p : ℝ[X]) : ℝ≥0∞ :=
+  if p.Monic ∧ p.degree = n ∧ p.roots.Nodup then
+    let roots := p.roots.toFinset
+    (∑ i ∈ roots, (∑ j ∈ roots.erase i, (1 : ℝ) / (i - j))^(2 : ℝ)).toNNReal
```
smmercuri: Are these explicit casts needed?  I would expect them to coerce automatically since `i` and `j` are real
```suggestion
    (∑ i ∈ roots, (∑ j ∈ roots.erase i, 1 / (i - j)) ^ 2).toNNReal
```

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,92 @@
...
+(p \boxplus_n q)(x) = \sum_{k=0}^n c_k x^{n-k}
+$$
+where the coefficients $c_k$ are given by the formula:
+$$
+c_k = \sum_{i+j=k} \frac{(n-i)! (n-j)!}{n! (n-k)!} a_i b_j
+$$
+for $k = 0, 1, \dots, n$.
+ -/
+noncomputable def finiteAdditiveConvolution (n : ℕ) (p q : F[X]) : F[X] :=
+  let c := fun k  => ∑ ij ∈ antidiagonal (k : ℕ),
```
smmercuri:
```suggestion
  let c := fun k => ∑ ij ∈ antidiagonal (k : ℕ),
```

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,92 @@
...
+Note: while no proof of this is published yet, the authors of
+[arxiv/2602.05192](https://arxiv.org/abs/2602.05192) announced that a proof will be released
+on 2026-02-13
+
+TODO(firsching): update category and remove Note when proof is published.
+-/
+@[category research open, AMS 26]
+theorem four : answer(sorry) ↔
+    ∀ (p q : ℝ[X]) (n : ℕ), p.degree = n → p.roots.card = n → q.degree = n → p.Monic → q.Monic →
+    1 / Φ n (p (⊞_n) q) ≥ 1 / Φ n p + 1 / Φ n q := by
```
smmercuri: nit: I think we prefer `le` over `ge`
```suggestion
    1 / Φ n p + 1 / Φ n q ≤ 1 / Φ n (p (⊞_n) q) := by
```

mo271: I chose this order, because it was this way in the paper, but have no strong feelings about it...

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,102 @@
...
+$$
+c_k = \sum_{i+j=k} \frac{(n-i)! (n-j)!}{n! (n-k)!} a_i b_j
+$$
+for $k = 0, 1, \dots, n$.
+ -/
+noncomputable def finiteAdditiveConvolution (n : ℕ) (p q : F[X]) : F[X] :=
+  let c := fun k => ∑ ij ∈ antidiagonal (k : ℕ),
+      ((n - ij.1)! * (n - ij.2)! : F) / (n ! * (n - k)! : F) *
+      (p.coeff (n - ij.1)) * (q.coeff (n - ij.2))
+  ∑ k ∈ range (n + 1), (c k) • X^(n - k)
```
eric-wieser: nit:

```suggestion
  ∑ k ∈ range (n + 1), c k • X^(n - k)
```

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,102 @@
...
+where the coefficients $c_k$ are given by the formula:
+$$
+c_k = \sum_{i+j=k} \frac{(n-i)! (n-j)!}{n! (n-k)!} a_i b_j
+$$
+for $k = 0, 1, \dots, n$.
+ -/
+noncomputable def finiteAdditiveConvolution (n : ℕ) (p q : F[X]) : F[X] :=
+  let c := fun k => ∑ ij ∈ antidiagonal (k : ℕ),
+      ((n - ij.1)! * (n - ij.2)! : F) / (n ! * (n - k)! : F) *
+      (p.coeff (n - ij.1)) * (q.coeff (n - ij.2))
```
eric-wieser: nit:
```suggestion
      p.coeff (n - ij.1) * q.coeff (n - ij.2)
```

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,102 @@
...
+  ∑ k ∈ range (n + 1), (c k) • X^(n - k)
+
+local notation p " (⊞_"n ")" q:65  => finiteAdditiveConvolution n p q
+
+@[category test, AMS 26]
+theorem finiteAdditiveConvolution_comm (n : ℕ) (p q : F[X]) :
+    p (⊞_n) q = q (⊞_n) p := by
+  show ∑ a ∈_, _= ∑ a ∈_, _
+  exact sum_congr rfl fun m hm  =>
```
eric-wieser:
```suggestion
  exact sum_congr rfl fun m hm =>
```

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,102 @@
...
+
+/--
+For a monic polynomial $p(x)=\prod_{i\le n}(x- \lambda_i)$, define
+$$\Phi_n(p):=\sum_{i\le n}(\sum_{j\neq i} \frac1{\lambda_i-\lambda_j})^2$$
+and $\Phi_n(p):=\infty$ if $p$ has a multiple root.
+-/
+noncomputable def Φ (n : ℕ) (p : ℝ[X]) : ℝ≥0∞ :=
+  if p.Monic ∧ p.degree = n ∧ p.roots.Nodup then
+    let roots := p.roots.toFinset
+    (∑ i ∈ roots, (∑ j ∈ roots.erase i, 1 / (i - j)) ^ 2).toNNReal
```
eric-wieser, line 82: I think this is slightly nicer written as a sum over `roots.offDiag`?

eric-wieser: You also get the `infinity` value for free if you use `Multiset.offDiag`, as that gives you a `1/0` term.

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,102 @@
...
+    (congr_arg₂ _) (sum_equiv (.prodComm _ _) (by simp [add_comm]) fun _ _ => by ring!) rfl
+
+
+/--
+For a monic polynomial $p(x)=\prod_{i\le n}(x- \lambda_i)$, define
+$$\Phi_n(p):=\sum_{i\le n}(\sum_{j\neq i} \frac1{\lambda_i-\lambda_j})^2$$
+and $\Phi_n(p):=\infty$ if $p$ has a multiple root.
+-/
+noncomputable def Φ (n : ℕ) (p : ℝ[X]) : ℝ≥0∞ :=
+  if p.Monic ∧ p.degree = n ∧ p.roots.Nodup then
```
eric-wieser: I'd be tempted to drop the `n` argument and this entire condition.

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,102 @@
...
+    let roots := p.roots.toFinset
+    (∑ i ∈ roots, (∑ j ∈ roots.erase i, 1 / (i - j)) ^ 2).toNNReal
+  else
+    ⊤
+
+def FourProp (p q : ℝ[X]) (n : ℕ) : Prop :=
+    p.degree = n → p.roots.card = n → q.degree = n  → q.roots.card = n → p.Monic → q.Monic →
+    1 / Φ n p + 1 / Φ n q ≤ 1 / Φ n (p (⊞_n) q)
```
eric-wieser: Should this have a docstring?

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,102 @@
...
+
+/--
+Is it true that if $p(x)$ and $q(x)$ are monic real-rooted polynomials of
+degree $n$, then
+$$\frac{1}{\Phi_n(p\boxplus_n q)} \ge \frac{1}{\Phi_n(p)}+\frac{1}{\Phi_n(q)}?$$
```
eric-wieser, line 98: Given the conditions in the definition of $$\Phi_n$$, this question isn't well-defined unless we first postulate that `p\boxplus_n q` is itself monic and of degree `n`; could you add a test statement to that effect?

FormalConjectures/OEIS/358684.lean
```diff
@@ -25,6 +25,10 @@ $n$-th Fermat number exceeds $2^(2^n - k)$.
 -/
 
 
+
+namespace OeisA358684
```
eric-wieser: nit: split this to its own PR.

mo271: done, #2263

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
mo271: I had a version that used `Finset.offDiag`, see earlier commits, but reverted that precisely because we still needed the `if` because `Multiset.offDiag` doesn't exists yet: (not does List.offDiag` in the mathlib version we are using: 
https://github.com/leanprover-community/mathlib4/blame/dec6b9345707c57cf9f737626e9f337991e633c7/Mathlib/Data/Finset/Prod.lean#L246

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
mo271: sounds good, will do

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
mo271: left a TODO

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
mo271: yes, done, also for two theorems below that didn't yet have one

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
mo271: done, they can be proved, but I haven't bothered cleaning up the AlphaProof proofs, but deleted them instead.

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,131 @@
...
+@[category test, AMS 26]
+theorem finiteAdditiveConvolution_degree (n : ℕ) (p q : ℝ[X])
+  (hp : p.degree = n) (hq : q.degree = n):
+    (p (⊞_n) q).degree = n := by
+  sorry
+
+@[category test, AMS 26]
+theorem finiteAdditiveConvolution_monic' (n : ℕ) (p q : ℝ[X]) (hn : 0 < n)
+  (hp : p.degree = n) (hq : q.degree = n) (hp : p.Monic) (hq : q.Monic) :
+    (p (⊞_n) q).Monic := by
```
eric-wieser, line 71:
```suggestion
@[category test, AMS 26]
theorem finiteAdditiveConvolution_monic' (n : ℕ) (p q : ℝ[X]) (hn : 0 < n)
    (hp : p.degree = n) (hq : q.degree = n) (hp : p.Monic) (hq : q.Monic) :
    (p (⊞_n) q).Monic := by
```

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,131 @@
...
+  sorry
+
+/--
+For a monic polynomial $p(x)=\prod_{i\le n}(x- \lambda_i)$, define
+$$\Phi_n(p):=\sum_{i\le n}(\sum_{j\neq i} \frac1{\lambda_i-\lambda_j})^2$$
+and $\Phi_n(p):=\infty$ if $p$ has a multiple root.
+-/
+noncomputable def Φ (p : ℝ[X]) : ℝ≥0∞ :=
+  -- TODO: write this as
+  -- (∑ ij ∈ p.roots.offDiag, (1 : ℝ) / (ij.1 - ij.2)^(2 : ℝ)).toNNReal
```
eric-wieser: Why the `toNNReal`? I think the subtraction needs to happen in ENNReal to ensure the `⊤` return value

mo271: I moved it inside now, so the the result of the subtraction is in ENNReal.

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,131 @@
...
+  sorry
+
+/--
+For a monic polynomial $p(x)=\prod_{i\le n}(x- \lambda_i)$, define
+$$\Phi_n(p):=\sum_{i\le n}(\sum_{j\neq i} \frac1{\lambda_i-\lambda_j})^2$$
+and $\Phi_n(p):=\infty$ if $p$ has a multiple root.
+-/
+noncomputable def Φ (p : ℝ[X]) : ℝ≥0∞ :=
+  -- TODO: write this as
+  -- (∑ ij ∈ p.roots.toFinset.offDiag, 1 / ((ij.1 - ij.2).toNNReal)^2)
```
eric-wieser: Wait, I think this is wrong; the `^2` is between the two sums in the english?

mo271: it is a bit confusing, but now I have a valid version in the comment. I provide an adhoc definition of `Multiset.offDiag` here and prove equiavlence: https://github.com/mo271/formal-conjectures/blob/972c5ead991ed7f9f08ccc2d1aea1604a5fa98da/FormalConjectures/Arxiv/2602.05192/FirstProof4.lean#L118

It might be a bit surprising there the `^2` is, but I think it is correct. 
Of course we cannot take the subtraction  in `ℝ≥0∞`, because of the nonnegativity, we need to cast only after taking the square. Then it works...

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,131 @@
...
+  sorry
+
+/--
+For a monic polynomial $p(x)=\prod_{i\le n}(x- \lambda_i)$, define
+$$\Phi_n(p):=\sum_{i\le n}(\sum_{j\neq i} \frac1{\lambda_i-\lambda_j})^2$$
+and $\Phi_n(p):=\infty$ if $p$ has a multiple root.
+-/
+noncomputable def Φ (p : ℝ[X]) : ℝ≥0∞ :=
+  -- TODO: consider writing this as
+  -- `(p.roots.offDiag.map (fun ij => (1 : ℝ≥0∞) / ((ij.1 - ij.2) ^ 2).toNNReal)).sum`
```
eric-wieser: This still looks wrong to me, this implements

$$\Phi_n(p):=\sum_{i\le n}\sum_{j\neq i} (\frac1{\lambda_i-\lambda_j})^2$$

not

$$\Phi_n(p):=\sum_{i\le n}(\sum_{j\neq i} \frac1{\lambda_i-\lambda_j})^2$$

eric-wieser: I guess your claim is that these end up coinciding?

eric-wieser: Either way, this isn't a reasonable interpretation of the prose any more, so I think we should drop the TODO; sorry for the noise!

mo271: I think these two expressions are mathematically identical (even if this looks wrong)! The implementation is correct because the mixed terms in the expansion sum to zero.

If you expand the square of the sum, you get the sum of squares plus cross-terms. The cross-terms for any distinct triplet $i, j, k$ are:

$$
\frac{1}{(\lambda_i-\lambda_j)(\lambda_i-\lambda_k)} + \frac{1}{(\lambda_j-\lambda_k)(\lambda_j-\lambda_i)} + \frac{1}{(\lambda_k-\lambda_i)(\lambda_k-\lambda_j)} 
$$
If you combine these fractions over a common denominator $D = (\lambda_i-\lambda_j)(\lambda_j-\lambda_k)(\lambda_k-\lambda_i)$, the signs align such that the numerator becomes a telescoping sum:
$$
\frac{-(\lambda_j-\lambda_k) - (\lambda_k-\lambda_i) - (\lambda_i-\lambda_j)}{D} = \frac{0}{D} = 0
$$

Hence we are left with

$$
\sum_{i} \left( \sum_{j \neq i} \frac{1}{\lambda_i - \lambda_j} \right)^2 = \sum_{i} \sum_{j \neq i} \frac{1}{(\lambda_i - \lambda_j)^2}
$$

I feel like there should be a name for this identity, but couldn't find it.
This is the informal argument, linked in a commit above is already a formal (unreadable AlphaProof) proof. 
[Here](https://sagecell.sagemath.org/?z=eJyVlE1vGjEQhu-R8h9e4BBvVZZA6AUBl-TQQw8VkdpDlK4M6wUjr5f6g3aF-O8d7y6BpoG2llarGXveecYeOxUZNkZql3CdJlthZFYmMhXaSVcyHY2ur0Cjg36MB5FJLWDLfF4oucCWG8nnSlgons9TntwijuODsdPd_r6OJg8mUNI6RjEsuzkskfubCFlhICE1DNdLQTmj66tD1kGM-0JbZ_zC4dPHxxGszxMJVv1369ZE7tHvsUZQontIv44QfRs0-VeW8lMEq-0wWDDryCf5XMc9rZ9rnPUpDmRGfK0J1hEp4ijxB3g9dYJ_d4o_O-KfwLMz9C_w5g34YL0FH_gu8f8P-zDGZyOcK-lH_QG3EjDCeuVsvWRTzSZV97CVyxXL2uPVYHrPrYAm5p3ej3vkaEeN-rm49ng-pcMFe_Q5igz2u-dGpKFMG43GvflR4rdgOte_K5spqc9O1KVO5Vamnqsm0cUc5lWODj7E-FLdEwgKV3RPDjNfBTzVHluZb1S4SJlXikVwRdjvhcAjX4pgiZ8bum4I34LrhVBwwuTNvqYyy4QR5KY9DDXS-QaK17L_VvnDUY2FPe6GPozQaEmRXqj9SPIyG9ztbhvvMLytmqWDmddVO-nJHVgoqSx8KAvOlOQcvqfWKfxyVXVQ4d3GOyyFs6DXpsQPemxapHT-ERpGvwBHF2AW&lang=sage&interacts=eJyLjgUAARUAuQ==) is an exploration in symbolic algebra (only for fixed finite $n$ of course...)

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
```diff
@@ -0,0 +1,131 @@
...
+  -- when `Multiset.offDiag` becomes available.
```
mo271:
```suggestion
```

FormalConjectures/Arxiv/2602.05192/FirstProof4.lean
mo271: > Either way, this isn't a reasonable interpretation of the prose any more, so I think we should drop the TODO; sorry for the noise!

Yeah, that was also my conclusion when I first noticed that this formulation is also possible, hence I first had the commit with the `offDiag` and then changed back to the version without.

Now removing the comment from the code

PR #2181, opened 2026-02-06
feat: lint for duplicate category attributes

### LINKED ISSUE #1852: Lint for `category` attribute duplicates
We should lint to make sure we don't have statements of the form
```lean
@[category research solved, category research open]
theorem foo : ...
```
and so on.

### Choose either option
- [x] I plan on implementing this
- [ ] This issue is up for grabs

Paul-Lez:
fixes #1852


mo271 (approved):

mo271:
@Paul-Lez it might potentially be actually useful to allow two different `research formally solved using ... at ...` categories?

But we can allow this the first times it is useful...

PR #2183, opened 2026-02-06
chore: update README

Paul-Lez:


README.md
```diff
@@ -244,10 +250,11 @@ meaningful solution of the problem is outside of the scope of this repository.
     ```
     If the problem has been solved to the negative, then `P` should be replaced with
     `¬ P`.
-7.  Every file should start with the following copyright header:
+7.  Every file should start with the following copyright header (replace YYYY with
+    the current year):
     ```lean
     /-
-    Copyright 2026 The Formal Conjectures Authors.
+    Copyright YYYY The Formal Conjectures Authors.
```
felixpernegger, line 257: I think its better to keep it as it is. Changing the date once a year isnt much effort.

mo271 (approved):

README.md
mo271: I have opinion here.. whatever

PR #2207, opened 2026-02-08
fix(ErdosProblems): 1097

smmercuri:
This variant is false for small n (`< 3`) since `CommonDifferencesThreeTermAP` is empty there


FormalConjectures/ErdosProblems/1097.lean
Paul-Lez:
```suggestion
A trivial lower bound: for sufficiently large `n` there exist sets $A$ with $|A| = n$ that contain at least $\Omega(n)$
```

PR #2208, opened 2026-02-08
fix(Paper): Nat division in `StrongSensitivityConjecture`

smmercuri:


FormalConjectures/Paper/StrongSensitivityConjecture.lean
```diff
@@ -112,13 +112,13 @@ of the input. -/
 @[category test, AMS 68]
 def nisanExample (n : ℕ) (x : Fin n → Bool) : Bool :=
   let w := #{i | x i}
-  decide (w ∈ ({n / 2, n / 2 + 1} : Finset ℕ))
+  decide ((w : ℚ) ∈ ({(n / 2 : ℚ), (n / 2 : ℚ) + 1} : Finset ℚ))
 
 /-- Assuming `n` is a multiple of 4, the sensitivity of `nisanExample`
 is `n/2`, achieved by any `x` with Hamming weight `n/2`. -/
```
Paul-Lez, line 118:
```suggestion
/-- There exists an `n` that is a multiple of 4, such that the sensitivity of `nisanExample n` is not `n/2`. -/
```

smmercuri: oh oops, I forgot to un-negate! The old docstring is correct

PR #2219, opened 2026-02-08
Add determine Busy Beaver n=6

franzhusch:
I thought it makes sense to also have determining the next value of $BB(n)$ as a open problem.


mo271:
Thanks!

PR #2226, opened 2026-02-10
fix(ErdosProblems): 33 

smmercuri:
Needs `∧` instead of `→`


smmercuri:
> We have seen this pattern `∃ x, Foo x → ...` often enough that we might want to lint against it. I will open an issue, so we don't forget

agreed!

mo271:
see #2227

PR #2235, opened 2026-02-10
fix: disable copyright linter on FormalConjectures/All.lean

Paul-Lez:
This PR adds a check in the copyright linter to avoid running it on `FormalConjectures/All.lean` since this throws an error
message every time we run CI.


mo271 (approved):
a bit of a hack, but why not!

PR #2263, opened 2026-02-11
chore(OEIS/358684): add namespace

mo271:


Paul-Lez:
I guess we'll have a lot less of these issues once #2229 is merged!

PR #2005, opened 2026-02-01
feat(GreensOpenProblems): 14

jeangud:
# Description
We define the 2-colour van der Waerden numbers $W(k, r)$ to be the least quantities such that if $\{1, . . . , W(k, r)\}$ is coloured red and blue then there is either a red $k$-term progression or a blue $r$-term progression.

1) Is $W(k, r)$ a polynomial in $r$, for fixed $k$?
2) Is $W(3, r) \ll r^2$?

- Closes #1549 
- Literature presents a good amount of known values, and existing lower/upper bounds, which I tried to include.

# Testing
- Builds successfully
```shell
$ lake build FormalConjectures/GreensOpenProblems/14.lean 
Build completed successfully.
```

PR #2274, opened 2026-02-12
feat(GreensOpenProblems): 22

### LINKED ISSUE #1557: Green's Open Problems #22
### What is the conjecture
https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.22

If $\\{1, . . . , N\\}$ is $r$-coloured then, for $N \geq N_0(r)$, there are integers $x, y \geq 3$ such that $x + y, xy$ have the same colour. Find reasonable bounds for $N_0(r)$.

### Choose either option

jeangud:
# Description
If $\lbrace 1, . . . , N \rbrace$ is $r$-coloured then, for $N \geqslant N_0(r)$, there are integers $x, y \geqslant 3$ such that $x + y, xy$ have the same colour. Find reasonable bounds for $N_0(r)$.

- Closes #1557
- Green mentions their paper [GrSa25] " arguably addresses the original formulation of the problem" with $N_0(r) \leq \exp~\exp (r^{50})$. I include an additional version where the `answer` tries to beat that bound.
- Also added some of Moreira's results justifying the existence of $N_0(r)$.
- Not sure whether we want to expand further and include **Hindman's question** (mentioned in Green's comments). Likely for another PR.

# Testing
- I checked corner cases, especially $r = 0$ and $N < 9$ and looks like they do not render the theorems trivially True/False.
- Builds successfully
```shell
$ lake build FormalConjectures/GreensOpenProblems/22.lean 
Build completed successfully.
```


FormalConjectures/GreensOpenProblems/22.lean
```diff
@@ -0,0 +1,118 @@
...
+
+/--
+The monochromatic sum-product property: a colouring $c$ of $\lbrace 1, \ldots, N \rbrace$ has a
+pair $(x, y)$ with $x, y \geq 3$ such that $x + y$ and $xy$ are both in
+$\lbrace 1, \ldots, N \rbrace$ and receive the same colour.
+-/
+def HasMonochromaticSumProduct (N : ℕ) (r : ℕ) (coloring : ℕ → Fin r) : Prop :=
+  ∃ x y : ℕ, 3 ≤ x ∧ 3 ≤ y ∧
+    x + y ∈ Icc 1 N ∧ x * y ∈ Icc 1 N ∧
+    coloring (x + y) = coloring (x * y)
```
mo271: in the docstring this is about a coloring from $\{1, \dots, N\}$ but the code gives a coloring from `ℕ → Fin r`. What is really meant here? I don't undertand how the color of `x + y` and `x * y` is defined in the formal text.

FormalConjectures/GreensOpenProblems/22.lean
```diff
@@ -0,0 +1,118 @@
...
+  sorry
+
+/--
+[Mo17, Corollary 1.5] For any finite coloring of $\mathbb{N}$ there exist (infinitely many)
+$x, y \in \mathbb{N}$ such that $\{x, xy, x+y\}$ is monochromatic.
+-/
+@[category research solved, AMS 5 11]
+theorem green_22.variants.moreira_infinite :
+    ∀ (r : ℕ) (coloring : ℕ → Fin r),
+    Set.Infinite { p : ℕ × ℕ |
```
mo271, line 96: does this also need $3 \le p.1$ and $3 \le p.2$?

mo271 (commented):
Thanks! A few initial remarks..

FormalConjectures/GreensOpenProblems/22.lean
jeangud: Thank you for the initial thoughts, @mo271!

Fair point, reminiscent of your previous comment [here](https://github.com/google-deepmind/formal-conjectures/pull/2005#discussion_r2759510794). I had chosen a total coloring on ℕ by convenience, to avoid carrying the subtype around, with the values outside of $\lbrace 1,\ldots, N\rbrace$ being unused. I am actually unsure how much overhead that represents to carry around once on the proof side...

I pushed 195e2bf which now enforces the $r$-coloring to be over $\lbrace 1,\ldots, N\rbrace$ instead of ℕ.

Does this address your second point on the coloring of $x+y$ and $x * y$?

FormalConjectures/GreensOpenProblems/22.lean
jeangud: Technically no if we want to match the exact statement from [Mo17 Corollary 1.5 (p.2)](https://arxiv.org/pdf/1605.01469).

However, your comment made me realize Moreira uses the convention $\mathbb{N} = \lbrace 1, 2, \ldots \rbrace$ in their paper, so we need at least `0 < p.1 ∧ 0 < p.2` (added with f15a3e3). I also removed `moreira_existence` which was somewhat redundant, and inconsistent since it was using `HasMonochromaticSumProduct` which did required the $x,y \geq 3$ condition.

Thank you for forcing a second look here!

jeangud:
@mo271 any thoughts on the above? I added a couple commits trying to address your comments above :blush:

FormalConjectures/GreensOpenProblems/22.lean
```diff
@@ -0,0 +1,111 @@
...
+      0 < p.1 ∧ 0 < p.2 ∧
+      coloring p.1 = coloring (p.1 * p.2) ∧
+      coloring p.1 = coloring (p.1 + p.2) } := by
+  sorry
+
+/--
+Since $x, y \geq 3$, we must have $xy \geq 9$, so $N_0(r) \geq 9$.
```
mo271:
```suggestion
Since $x, y \geq 3$, we must have $xy \geq 9$, so $N_0(r) \geq 9$
(assuming $N_0(r)$ is well-defined, which follows from [Mo17]).
```

FormalConjectures/GreensOpenProblems/22.lean
```diff
@@ -0,0 +1,111 @@
...
+-/
+@[category research solved, AMS 5 11]
+theorem green_22.variants.lower_trivial : ∀ r : ℕ, r ≠ 0 → 9 ≤ N₀ r := by
```
mo271:
```suggestion
theorem green_22.variants.lower_nine : ∀ r : ℕ, r ≠ 0 → 9 ≤ N₀ r := by
```

mo271 (commented):
Thanks, LGTM now!
just some name and docstring suggestion

jeangud:
Sweet 🎉 Thank you @mo271

mo271 (approved):

PR #2283, opened 2026-02-13
fix(ErdosProblems): 786

smmercuri:
- Fix order of quantifiers in Selfridge example because the length `k` of the sequence is determined by the magnitude of the primes. If we take `k = 1` before the primes themselves, then very large primes will never straddle `1` in their sums of reciprocals, giving a contradiction.
- Redefine `consecutivePrimes` in order to formalise this interpretation.
- The condition that `A` is the set of numbers with only one prime from the sequence dividing it requires multiplicity one as well. 
- Fix constraints on `eps`
- Add a TODO since the updated website mentions that repetitions may be allowed in `Set.IsMulCardSet`, although both versions are open.


mo271 (approved):
Thanks, great work!

PR #2287, opened 2026-02-14
feat(ErdosProblems): 184

### LINKED ISSUE #410: Erdős Problem 184: decompose graph in linear number of cycles and edges
### What is the conjecture
    
https://www.erdosproblems.com/184

  Any graph on $n$ vertices can be decomposed into $O(n)$ many cycles and edges.

Status: open

### Choose either option

danielchin:
Closes: #410 

Formalizes Erdos Problem 184: https://www.erdosproblems.com/184

### DIFF
--- a/FormalConjectures/ErdosProblems/184.lean
+++ b/FormalConjectures/ErdosProblems/184.lean
@@ -14,7 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -/
 
-
 import FormalConjectures.Util.ProblemImports
 
 /-!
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/184.lean
```diff
@@ -0,0 +1,125 @@
...
+  (⋃ H ∈ D, H.edgeSet) = G.edgeSet
+
+/--
+Any graph on $n$ vertices can be decomposed into $O(n)$ many edge-disjoint cycles and edges.
+-/
+@[category research open, AMS 5]
+theorem erdos_184 :
+    ∃ C : ℝ, ∀ {V : Type} [Fintype V] [DecidableEq V] (G : SimpleGraph V),
+    ∃ (D : Finset (SimpleGraph V)),
+      (∀ H ∈ D, IsCycleOrEdge H) ∧
```
mo271: Maybe I'm mistakes, but I don't think having H here to be defined as a Graph on `V` plays well together with the definition of `IsCycleOrEdge`: will it not typically contain many isolated vertices and hence not fullfill `IsCycleOrEdge`? 
Perhaps a formulation using subgraphs might make more sense, also for the definition of `IsDecomposition`?

mo271 (commented):
Thanks!

FormalConjectures/ErdosProblems/184.lean
```diff
@@ -0,0 +1,125 @@
...
+        IsDecomposition (fromRel (fun (i j : Fin n) => (i : ℕ) < 3 ∧ 3 ≤ (j : ℕ))) D →
+        (1 + c) * n ≤ (D.card : ℝ) := by
+  sorry
+
+/--
+In [Er71] Erdős suggests that only $n-1$ many cycles and edges are required if we do not
+require them to be edge-disjoint.
+-/
+@[category research open, AMS 5]
+theorem erdos_184.variants.covering :
```
mo271, line 96: AlphaProof found a disproof here: 

```
theorem erdos_184.variants.covering :
    ¬ (answer(True) ↔
      ∀ {V : Type} [Fintype V] [DecidableEq V] (G : SimpleGraph V),
      ∃ (D : Finset (SimpleGraph V)),
        (∀ H ∈ D, IsCycleOrEdge H) ∧
        (⋃ H ∈ D, H.edgeSet) = G.edgeSet ∧
        D.card ≤ Fintype.card V - 1) := by
  -- AlphaProof found a disproof
  push_neg
  delta Erdos184.IsCycleOrEdge
  refine .inl ⟨trivial,ULift (Fin 6), inferInstance, inferInstance, {Adj:=fun A B=>A.1≠B.1∧A.1≠0∧B.1≠0}, fun and A B=>not_le.1 fun and' =>?_⟩
  replace B:and.biUnion (·.edgeFinset)=.image (fun(a, b)=>s(a, b)) {(a, b):_ ×_|a.1≠b.1∧a.1≠0∧b.1≠0}
  · simp_all [Sym2.forall, false,Set.ext_iff, Finset.ext_iff]
    trivial
  replace A : ∀ a ∈and,a.edgeFinset.card ≤ 1:=(A · ·|>.elim ( fun and=>not_lt.1 fun and=>? _) le_of_eq)
  · use (by decide ∘(B▸and.card_biUnion_le).trans) ( (and.sum_le_card_nsmul _ _ A).trans.comp (mul_one _).trans_le and')
  bound[2]
  rw[SimpleGraph.IsCycles]at*
  contrapose! B
  norm_num[Sym2.forall,Set.nonempty_def,Set.ncard_eq_toFinset_card',Finset.ext_iff]at A right⊢
  push_neg
  norm_num[exists_or, and_or_left]
  use(left 0 1).elim (by cases. with use .inl ⟨0,_, ⟨ _,x2,by valid⟩,by valid⟩)
```
I suspect a misformalisation, see other comment.

FormalConjectures/ErdosProblems/184.lean
danielchin: Great catch, fixed the formulation here.

FormalConjectures/ErdosProblems/184.lean
danielchin: I rewrote this a bit as well, there was a case where this could be vacuously false which I caught.

danielchin:
I actually rewrote this quite a bit so it matches the big O notation more closely and the rest of the statements more closely.

FormalConjectures/ErdosProblems/184.lean
```diff
@@ -0,0 +1,134 @@
...
+      (f =O[atTop] fun n : ℕ ↦ (n : ℝ)) ∧
+      ∀ {V : Type*} [Fintype V] [DecidableEq V] (G : SimpleGraph V),
+      ∃ (D : Finset G.Subgraph),
+        (∀ H ∈ D, IsCycleOrEdge H.coe) ∧
+        IsDecomposition G D ∧
+        (D.card : ℝ) ≤ f (Fintype.card V) := by
+  sorry
+
+/--
+Erdős and Gallai proved that $O(n \log n)$ many cycles and edges suffices.
```
mo271: let's add a reference to that!

mo271 (approved):
Thanks, LGTM now, modulo the missing reference

FormalConjectures/ErdosProblems/184.lean
danielchin: Had to do some digging to find this one. Left a comment on the Erdos Problems website as well with the source.

mo271:
Thanks!

PR #2156, opened 2026-02-04
feat(GreensOpenProblems): 15

jeangud:
# Description
Does there exist a Lipschitz function $f : \mathbb{N} \to \mathbb{Z}$ whose graph $\Gamma = \{(n, f(n)) : n \in \mathbb{N}\} \subseteq \mathbb{Z}^2$ is free of 3-term progressions?

- Closes #1550 
- Added the solved case for 4-term APs

# Testing
- Builds successfully
```shell
$ lake build FormalConjectures/GreensOpenProblems/15.lean 
Build completed successfully.
```

PR #2288, opened 2026-02-15
fix(ErdosProblems): 193

### LINKED ISSUE #415: Erdős Problem 193
### What is the conjecture
    
https://www.erdosproblems.com/193

  Let $S\subseteq \mathbb{Z}^3$ be a finite set and let $A=\\{a_1,a_2,\ldots,\\}\subset \mathbb{Z}^3$ be an infinite $S$-walk, so that $a_{i+1}-a_i\in S$ for all $i$. Must $A$ contain three collinear points?

Status: opend

### Choose either option

jeangud:
# Description

> Let $S \subseteq \mathbb{Z}^3$ be a finite set and let $A = \lbrace a_1, a_2,\ldots \rbrace$ be an infinite $S$-walk, so that $a_{i+1} - a_i \in S$ for all $i$. Must $A$ contain three collinear points?

I am not so confident about this implementation, the review will be helpful :blush: 
- Closes #415 
- A follow up from #2156 
- It turned out to be a bit more involved than I thought compared to #2156, although I tried matching the style for the main theorem statement.
- I believe edge cases are covered ($S = \varnothing$, singleton $S$, constant walk, duplicate points / trivial collinear points)
- I avoided using the obvious [**`Collinear`**](https://leanprover-community.github.io/mathlib4_docs/Mathlib/LinearAlgebra/AffineSpace/FiniteDimensional.html#Collinear) on $\mathbb{Z}$ as it required a [`DivisionRing`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Field/Defs.html#DivisionRing) or additional casting, but maybe there is an elegant way to do it.

# Testing
- Compiles successfully
```shell
$ lake build FormalConjectures/ErdosProblems/193.lean 
Build completed successfully.
```


FormalConjectures/ErdosProblems/193.lean
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Erdos193
+
+/-- An $S$-walk is a sequence where every difference is in $S$. -/
+def IsSWalk {V : Type*} [AddCommGroup V] (S : Set V) (a : ℕ → V) : Prop :=
+  ∀ n, a (n + 1) - a n ∈ S
+
+/-- True if set $A$ contains 3 distinct collinear points. -/
+def HasCollinearTriple {V : Type*} [AddCommGroup V] (A : Set V) : Prop :=
+  ∃ x ∈ A, ∃ y ∈ A, ∃ z ∈ A, x ≠ y ∧ y ≠ z ∧ x ≠ z ∧ ¬ AffineIndependent ℤ ![x,y,z]
```
Paul-Lez: I think `Colinear` makes more sense here. Note also that the problem probably also wants you to use the rationals as your base ring (you don't need to go to the reals here since your points have coords in Z)
```suggestion
  ∃ x ∈ A, ∃ y ∈ A, ∃ z ∈ A, x ≠ y ∧ y ≠ z ∧ x ≠ z ∧ Collinear ℤ ![x,y,z]
```

Paul-Lez (commented):
Thanks for the contribution!

FormalConjectures/ErdosProblems/193.lean
jeangud: Thank you for the review @Paul-Lez!

`Collinear` was also my first thought, but directly using `Collinear ℤ ![x,y,z]` results in some `failed to synthesize DivisionRing ℤ` errors, hence my later shift to `AffineIndependent`.

Following your idea of using ℚ then, I end up with something like
```lean
def HasCollinearTriple3D (A : Set (ℤ × ℤ × ℤ)) : Prop :=
  ∃ x ∈ A, ∃ y ∈ A, ∃ z ∈ A, x ≠ y ∧ y ≠ z ∧ x ≠ z ∧ 
    Collinear ℚ ({(x.1 : ℚ, x.2.1 : ℚ, x.2.2 : ℚ), 
                  (y.1 : ℚ, y.2.1 : ℚ, y.2.2 : ℚ), 
                  (z.1 : ℚ, z.2.1 : ℚ, z.2.2 : ℚ)} : Set (ℚ × ℚ × ℚ))
```
This works, but we lose polymorphism over the variant `erdos_193_z2` unless we duplicate the code for `ℤ × ℤ`.

You also mention using ℚ as the base ring (I understood for the whole problem?), definitely an option but then I am not sure we capture the exact statement of Erdos 193.

Happy to implement what you think is best, what is your preference?

Paul-Lez: Just to clarify, the suggestion here would be to use `ℚ` as the base ring when you need to use "vector space predicates" (e.g. `Collinear`). 

Re making things polymorphic, I agree that would be nice to keep - how about the following: work with `Fin 2 → ℤ` rather than `ℤ × ℤ` (and same for triples), state `HasCollinearTriple` for general `R`-modules, and then use `HasCollinearTriple` in your problem statements on the vector spaces `Fin 2 → ℚ` and `Fin 3 → ℚ` (and you can easily obtain a vector `Fin 2 → ℚ` from a vector `v : Fin 2 → ℤ` by writing `(↑) ∘ v`). 

Let me know if you run into any issues with this!

jeangud: This is great, much appreciate the guidance here! I need to get used to more advanced Lean syntax like this.

- Pushed your suggestion with 2745b74
- Check it still `lake build` fine :white_check_mark:

FormalConjectures/ErdosProblems/193.lean
```diff
@@ -0,0 +1,69 @@
...
+open Set
+
+namespace Erdos193
+
+/-- An $S$-walk is a sequence where every difference is in $S$. -/
+def IsSWalk {V : Type*} [AddCommGroup V] (S : Set V) (a : ℕ → V) : Prop :=
+  ∀ n, a (n + 1) - a n ∈ S
+
+/-- True if set $A$ contains 3 distinct collinear points. -/
+def HasCollinearTriple {R V : Type*} [DivisionRing R] [AddCommGroup V] [Module R V] (A : Set V) : Prop :=
```
Paul-Lez: Small nitpick here: you should make `R` into an explicit variable (i.e. using `( )`) since it can't be inferred from the type of `A`. This will avoid having to write `R := ...` in the conjectures below.

FormalConjectures/ErdosProblems/193.lean
```diff
@@ -0,0 +1,69 @@
...
+Let $S \subseteq \mathbb{Z}^3$ be a finite set and let $A = \lbrace a_1, a_2, \ldots \rbrace$ be
+an infinite $S$-walk, so that $a_{i+1} - a_i \in S$ for all $i$. Must $A$ contain three collinear
+points?
+-/
+@[category research open, AMS 5]
+theorem erdos_193 :
+    answer(sorry) ↔ ∀ S : Set (Fin 3 → ℤ), S.Finite →
+      /- The statement's $A = \lbrace a_1, a_2, \ldots \rbrace$ is an infinite set; we formalize
+      this as an infinite range. This allows for repeats but forces infinitely many distinct points,
+      thereby avoiding degenerate $S$-walks (e.g. constant sequences, non-collinear loops) -/
```
Paul-Lez: For this, one could actually avoid forcing infinitely distinct points since if the sequence only has finitely many values then one point is repeated infinitely many times and this should yield a collinear triple (for trivial reasons!), so the conjecture would hold for degenerate `S`-walks. That being said, I suspect this might not be what the authors of the conjecture had in mind so I think it's reasonable to stick to the current formalisation. 
Would you mind adding a small comment to that effect?

Paul-Lez (approved):

FormalConjectures/ErdosProblems/193.lean
jeangud: Lightens the syntax below indeed, thank you! Done :white_check_mark:

FormalConjectures/ErdosProblems/193.lean
jeangud: Expanded a bit on the comment, hopefully captures what you had in mind :+1:

jeangud:
@Paul-Lez thank you again for spending some time on the reviews. Anything more you'd like to see on this PR?

Paul-Lez:
@jeangud Will take another look today and hopefully merge:) (if I forget feel free to pester me!)

jeangud:
Before continuing on Green's conjectures, I was also hoping to get final eyes on #2005 and #2274, which should be just about ready to merge as well. Might you be able to check in with the team?

In any case, thanks for the help @Paul-Lez, really appreciate it.

Paul-Lez (approved):
Nice work!

Paul-Lez:
Will also take a look at your other two PRs:)

PR #2299, opened 2026-02-16
feat(CI): distinguish formally solved Erdős problem status

zond:
Map 'solved (Lean)', 'proved (Lean)', 'disproved (Lean)' to 'formally solved' as a distinct category from 'solved', and tags the issue "formalisation exists elsewhere".

### DIFF
--- a/.github/workflows/check_erdos_status.yml
+++ b/.github/workflows/check_erdos_status.yml
@@ -2,7 +2,7 @@ name: Check Erdos problem statuses
 
 on:
   schedule:
-    - cron: '37 * * * *' # once an hour, 37 min past the hour
+    - cron: '0 6 * * *' # Daily at 6 AM UTC
   workflow_dispatch: # Allow manual trigger
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/check_erdos_status.yml
```diff
@@ -2,7 +2,7 @@ name: Check Erdos problem statuses
 
 on:
   schedule:
-    - cron: '37 * * * *' # once an hour, 37 min past the hour
+    - cron: '0 6 * * *' # Daily at 6 AM UTC
```
mo271:
```suggestion
    - cron: '37 * * * *' # once an hour, 37 min past the hour
```

Let's keep this

scripts/check_erdos_status.py
mo271: Can you add the "formalisation exists elsewhere" label here in case we should put the tag `formally solved`.

mo271 (commented):
Thanks!

.github/workflows/check_erdos_status.yml
zond: Are you sure you want to run it hourly_

scripts/check_erdos_status.py
zond: Done

mo271 (approved):
Thanks!

PR #2327, opened 2026-02-17
feat(CI): handle multi-part Erdős problems in status check

zond:
For problems with no single `erdos_{N}` theorem (e.g. multi-part problems with `erdos_{N}.parts.i`, `erdos_{N}_cycles`, etc.), collect all non-variant theorems and report open if any part is open.


mo271:
I'd like to have @danielchin's opinion here!
Should we introduce a consistent naming for those parts of a main question that are in the box versus those that variants that are in the additional material below the box? 
One we have consistent naming it should be pretty clear when there is a mismatch between "solved" on the website as indicated by Tao's json and our "solved" categories, i.e. "solved" on the website should correspond to all parts in the box are marked `research solved` (or `research formally solved`)

danielchin:
I agree that we probably need to come up with a proper standard and document it in the README since it looks like the naming convention has evolved over time.

I'll call the two potential parts of an Erdos problem: the main part (whatever is in the red or green box) and the additional text (whatever is below the red or green box).

For the additional text, I think this is easy. We just always call it `erdos_{N}.variants.{name}`.

For the main part, we have to consider all the types of questions we have and whatever future ones T. Bloom might add in the future. 

1.  Single questions: This is simple, we just do `erdos_{N}`.
2. Multiple part question: We can maintain what we're doing right now with `erdos_{N}.parts.i` and `erdos_{N}.parts.ii`. There won't be an actual `erdos_{N}` in this case.
3. Estimate question: The common pattern right now is `erdos_{N}.lower_bound` and `erdos_{N}.upper_bound`.
4. Describe/Explain the behavior: I don't believe we have an example of this one yet.

Assuming that we're okay with that naming convention, we can just check for all `erdos_{N}` that do not contain with `variants` and if they are all solved (because some parts can be solved, but other parts unsolved, meaning that the website displays them as unsolved), then you can mark it as solved and compare it against the website.

I did previously ask something similar to this in the Zulip chat: [#Formal conjectures > Referencing OEIS @ 💬](https://leanprover.zulipchat.com/#narrow/channel/524981-Formal-conjectures/topic/Referencing.20OEIS/near/571175026)

Yael mentioned that we should replace the `.` notation with underscores, but I think it's cleaner to keep the `.`, but just state to not repeat subnames (which I assume that's what Yael is saying).

scripts/check_erdos_status.py
```diff
@@ -84,8 +83,19 @@ def lean_category(cat):
     return "solved"
 
 
+def is_variant(suffix):
+    """Check if a theorem name suffix indicates a variant (not a main part)."""
+    # ".variants." anywhere in the suffix means it's a variant
+    return ".variants." in suffix or ".variant." in suffix
```
danielchin: Can you leave a TODO here (you can assign it to me) to remove anything that's `variant`. We should just rename everything to `variants` and have the linter catch anything that's not one of our existing patterns.

mo271:
Sounds good, @danielchin! We could add this naming convention to a README.md in `./ErdosProblems/` and then do big pull request making sure we adhere to the proposed standard. Do you want to do that or shall somebody else?

danielchin:
Yup, I can do the write up and then do a massive clean up to ensure standards are being met. Feel free to assign me an issue.

scripts/check_erdos_status.py
danielchin: Fix the naming problem here: https://github.com/google-deepmind/formal-conjectures/pull/2408

zond:
So, do I need to change anything with this PR, or?

danielchin:
You can drop the condition on `.variant.` on line 89 since there aren't problems named with that anymore.

zond:
> You can drop the condition on `.variant.` on line 89 since there aren't problems named with that anymore.

Done

mo271 (approved):
This looks good to go now, 
Do we need anything else here, @danielchin ?

PR #2329, opened 2026-02-17
feat(FormalConjectures/Util/Linters): ExistsImplication linter

### LINKED ISSUE #2227: Lint against `∃ x, Foo x → ...`
because the intention here is likely  `∃ x, Foo x ∧ ...`

Paul-Lez:
Fixes #2227.

This linter detects misformalisations arising from statements of the form `∃ x, P x → Q` that should have been formalised instead as `∃ x, P x ∧ Q`.


mo271 (approved):
Thanks, beautiful!

mo271:
A bit surprised that it didn't catch any newones, but I had done a search for regular expressions before and fixed those....

PR #2331, opened 2026-02-17
fix(ErdosProblems): 434 

smmercuri:
Website now contains the condition that the set $A$ is pairwise coprime. If not then there are always infinitely many unrepresentable numbers (the non-multiples of the gcd).


mo271 (approved):
Thanks!

PR #2356, opened 2026-02-18
chore: bump to v4.28.1

YaelDillies:


mo271 (commented):
do we need to bump verso in `./docbuild` too?

mo271 (approved):

PR #2358, opened 2026-02-18
solve(GreensOpenProblems): upper_trivial in 24

theaustinhatfield:
Proof generated by Gemini 3 Flash and verified via Lean 4 compilation. The model solved this in 3 iterations (178 seconds) after failing the first two attempts on a wrong lemma name. Full info at https://asiprize.com/verified/greens-problem-24/

This is the only verified proof out of 297 formal conjectures evaluated across 4 models. All 7 verification layers passed. Axioms: propext, Classical.choice, Quot.sound.

Attribute changed from category research open to category research solved.

### DIFF
--- a/FormalConjectures/GreensOpenProblems/24.lean
+++ b/FormalConjectures/GreensOpenProblems/24.lean
@@ -60,10 +60,9 @@ namespace variants
 /-- From [Aa19] p.577: the trivial upper bound is $n^2$ (non asymptotic) -/
 @[category research solved, AMS 5 11]
 theorem upper_trivial {n : ℕ} : max013AffineTranslates n ≤ n ^ 2 := by
-  unfold max013AffineTranslates
   apply csSup_le
-  · obtain ⟨A, hA⟩ : ∃ A : Finset ℤ, A.card = n := ⟨(Finset.range n).image Int.ofNat, by rw [Finset.card_image_of_injective _ Int.ofNat_injective, Finset.card_range]⟩
-    exact ⟨_, ⟨A, hA, rfl⟩⟩
+  · exact ⟨_, ⟨(Finset.range n).image Int.ofNat, by
+      rw [Finset.card_image_of_injective _ Int.ofNat_injective, Finset.card_range], rfl⟩⟩
   · rintro k ⟨A, hA, rfl⟩
     apply le_trans (Finset.card_filter_le _ _)
     rw [Finset.card_product, hA, pow_two]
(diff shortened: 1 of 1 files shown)


FormalConjectures/GreensOpenProblems/24.lean
```diff
@@ -58,9 +58,15 @@ theorem green_24 : ∀ n, max013AffineTranslates n = answer(sorry) := by
...
+@[category research solved, AMS 5 11]
 theorem upper_trivial {n : ℕ} : max013AffineTranslates n ≤ n ^ 2 := by
-  sorry
+  unfold max013AffineTranslates
+  apply csSup_le
+  · obtain ⟨A, hA⟩ : ∃ A : Finset ℤ, A.card = n := ⟨(Finset.range n).image Int.ofNat, by rw [Finset.card_image_of_injective _ Int.ofNat_injective, Finset.card_range]⟩
+    exact ⟨_, ⟨A, hA, rfl⟩⟩
+  · rintro k ⟨A, hA, rfl⟩
+    apply le_trans (Finset.card_filter_le _ _)
+    rw [Finset.card_product, hA, pow_two]
```
Paul-Lez: (minigolf)
```suggestion
  apply csSup_le
  · exact ⟨_, ⟨(Finset.range n).image Int.ofNat, by
      rw [Finset.card_image_of_injective _ Int.ofNat_injective, Finset.card_range], rfl⟩⟩
  · rintro k ⟨A, hA, rfl⟩
    apply le_trans (Finset.card_filter_le _ _)
    rw [Finset.card_product, hA, pow_two]
```

FormalConjectures/GreensOpenProblems/24.lean
```diff
@@ -58,9 +58,15 @@ theorem green_24 : ∀ n, max013AffineTranslates n = answer(sorry) := by
 namespace variants
 
 /-- From [Aa19] p.577: the trivial upper bound is $n^2$ (non asymptotic) -/
-@[category research open, AMS 5 11]
+@[category research solved, AMS 5 11]
```
Paul-Lez:
```suggestion
@[category research formally solved using formal_conjectures at "", AMS 5 11]
```

theaustinhatfield:
Thanks again for the quick review. Applied both suggestions minigolf and category attribute updated with PR link. Proof re-verified locally (rc=0, clean axioms). Happy to make any other changes.

FormalConjectures/GreensOpenProblems/24.lean
mo271: we should perhaps add as a category `formally solved here` and then lint against sorry-free `research solved`, wdyt @Paul-Lez!
(In a separate pr of course)

Paul-Lez (approved):
Thanks!

FormalConjectures/GreensOpenProblems/24.lean
mo271: https://github.com/google-deepmind/formal-conjectures/issues/3399 issue for this

PR #2359, opened 2026-02-19
feat(ErdosProblems): 323

### LINKED ISSUE #488: Erdős Problem 323
### What is the conjecture
    
https://www.erdosproblems.com/323

  Let $1\leq m\leq k$ and $f_{k,m}(x)$ denote the number of integers $\leq x$ which are the sum of $m$ many nonnegative $k$-th powers. Is it true that $f_{k,k}(x) \gg_\epsilon x^{1-\epsilon}$ for all $\epsilon>0$? Is it true that if $m<k$ then $f_{k,m}(x) \gg x^{m/k}$ for sufficiently large $x$?

Status: open

### Choose either option

danielchin:
Closes #488 

Formalizes Erdos Problem 323: https://www.erdosproblems.com/323

### DIFF
--- a/FormalConjectures/ErdosProblems/323.lean
+++ b/FormalConjectures/ErdosProblems/323.lean
@@ -1,0 +1,76 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 323
+
+*Reference:* [erdosproblems.com/323](https://www.erdosproblems.com/323)
+-/
+
+open Filter
+open scoped Asymptotics
+
+namespace Erdos323
+
+/--
+Let $1\leq m\leq k$ and $f_{k,m}(x)$ denote the number of integers $\leq x$ which are the sum of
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/323.lean
```diff
@@ -0,0 +1,76 @@
...
+/--
+Is it true that $f_{k,k}(x) \gg_\epsilon x^{1-\epsilon}$ for all $\epsilon>0$?
+
+This would have significant applications to Waring's problem. Erd\H{o}s and Graham describe this as
+'unattackable by the methods at our disposal'.
+-/
+@[category research open, AMS 11]
+theorem erdos_323.parts.i :
+    answer(sorry) ↔ ∀ k ≥ 1, ∀ ε > 0,
+      (fun (x : ℕ) ↦ (x : ℝ) ^ (1 - ε)) =O[atTop] (fun (x : ℕ) ↦ (f k k x : ℝ)) := by
```
mo271, line 46: I think one problem here is that `ε` is a a natural number here. In any case, AlphaProof finds a proof:
```
theorem erdos_323.parts.i :
    answer(True) ↔ ∀ k ≥ 1, ∀ ε > 0,
      (fun (x : ℕ) ↦ (x : ℝ) ^ (1 - ε)) =O[atTop] (fun (x : ℕ) ↦ (f k k x : ℝ)) := by
  simp_all? (config := {singlePass:= 1}) -contextual[ f]
  use fun and R L M=>.of_bound (1) (.of_forall fun and=>mod_cast(1).sub_eq_zero_of_le M▸Nat.mul_pos (by decide) (Nat.card_pos_iff.2 ⟨⟨0,bot_le,0,by field_simp⟩,(Set.finite_le_nat and).sep _,⟩))
```

mo271 (commented):
Thank you! Looks good in general, but the at least one bug: the type of epsilon...

FormalConjectures/ErdosProblems/323.lean
danielchin: Changed `ε` to real.

mo271 (approved):
Thanks again, this LGTM now!

PR #2378, opened 2026-02-21
feat(Millennium): Add Riemann Hypothesis (from Mathlib)

franzhusch:
Adds the Riemann Hypothesis (from mathlib) as a open problem to the Millennium Folder.

I also wondered if we should let the GRH stay there, its technically not a official Millennium Problem but ofc it is closely related to the Riemann Hypothesis (and implies it).

### DIFF
--- a/.github/workflows/build-and-docs.yml
+++ b/.github/workflows/build-and-docs.yml
@@ -130,7 +130,7 @@ jobs:
         shell: bash
         run: |
           cd docbuild
-          lake exe overwrite_index ./out/index.html ./out/file_counts.html | tee -a "$GITHUB_STEP_SUMMARY"
+          lake exe overwrite_index ./out/index.html ./out/file_counts_dark.html ./out/file_counts_white.html | tee -a "$GITHUB_STEP_SUMMARY"
 
       - name: Upload docs
         id: deployment
--- a/.github/workflows/check_erdos_status.yml
+++ b/.github/workflows/check_erdos_status.yml
@@ -2,7 +2,7 @@ name: Check Erdos problem statuses
 
 on:
   schedule:
-    - cron: '0 6 * * *' # Daily at 6 AM UTC
+    - cron: '37 * * * *' # once an hour, 37 min past the hour
   workflow_dispatch: # Allow manual trigger
 
 jobs:
--- a/FormalConjectures/ErdosProblems/1043.lean
+++ b/FormalConjectures/ErdosProblems/1043.lean
@@ -18,7 +18,14 @@ import FormalConjectures.Util.ProblemImports
 /-!
 # Erdős Problem 1043
 
-*Reference:* [erdosproblems.com/1043](https://www.erdosproblems.com/1043)
+*References:*
+- [erdosproblems.com/1043](https://www.erdosproblems.com/1043)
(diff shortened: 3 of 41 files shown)


mo271:
We already have `FormalConjectures/Millenium/GeneralizedRiemannHypothesis.lean`, right?

franzhusch:
Yes but it technically doesnt ask the Riemann Hypothesis but a harder conjecture. This PR, was mainly meant to open this discussion. We could also potentially add the "regular" Riemann Hypothesis to the GRH File.

mo271:
Yes, let's just have both in one file!

franzhusch:
Should we then later also have the Extended Riemann Hypothesis in the Millenium Directory. The Extended Riemann Hypothesis is essentially the Riemann Hypothesis for Dedekind Zeta Functions.

mo271:
yes, I don't think it worth making an extra file in Wikipedia or wherever. Just explain in the docstring what the version from Clay is etc.
(Since we are just importing the statements from Mathlib anyhow they will be quite short in any case)

franzhusch:
Ok makes sense and what should be the name of the file be then? Just RiemannHypothesis or GeneralizedRiemannHypothesis?

mo271:
> Ok makes sense and what should be the name of the file be then? Just RiemannHypothesis or GeneralizedRiemannHypothesis?

no strong opinion, perhaps just RiemannHypothesis for brevity

franzhusch:
Should the copyright be 2025 or 2026?

FormalConjectures/Millenium/RiemannHypothesis.lean
```diff
@@ -17,11 +17,39 @@ limitations under the License.
...
+part $\frac{1}{2}$. That is, if $\zeta(s) = 0$, $s \neq 1$, and $s$ is not a trivial zero
+$-2(n+1)$ for some $n \in \mathbb{N}$, then $\operatorname{Re}(s) = \frac{1}{2}$.
+
+This is the official Millennium Prize Problem as posed by the
+[Clay Mathematics Institute](https://www.claymath.org/wp-content/uploads/2022/05/riemann.pdf).
+
+This uses the `RiemannHypothesis` type from Mathlib, which is defined as
+`∀ (s : ℂ), riemannZeta s = 0 → (¬∃ n : ℕ, s = -2 * (n + 1)) → s ≠ 1 → s.re = 1 / 2`. -/
+@[category research open, AMS 11]
+theorem Riemann_Hypothesis : RiemannHypothesis := by
```
smmercuri: This should be `riemannHypothesis` according to mathlib naming conventions

smmercuri:
> Should the copyright be 2025 or 2026?

I think we should use `2026` here

FormalConjectures/Millenium/RiemannHypothesis.lean
franzhusch: You mean instead of Riemann_Hypothesis?

smmercuri: Yes, see [point 5 under Capitalization](https://leanprover-community.github.io/contribute/naming.html). Since `RiemannHypothesis` is named in `UpperCamelCase` it becomes `lowerCamelCase` in theorem names.

Issue #2292, opened 2026-02-15
Will Antihydra halt?


MrBrain295:
### What is the conjecture
Will the turing machine [1RB1RA_0LC1LE_1LD1LC_1LA0LB_1LF1RE_---0RA](https://wiki.bbchallenge.org/wiki/1RB1RA_0LC1LE_1LD1LC_1LA0LB_1LF1RE_---0RA) halt?
### Prerequisites needed
None AFAIK. Its high level rules have been proven so one could probably just implement that.
### [AMS categories](https://github.com/google-deepmind/formal-conjectures/labels?q=ams-)

* ams-68

### Choose either option
- [ ] I plan on adding this conjecture to the repository
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

franzhusch:
Hello, thanks for opening a issue, unfortunately (or luckily?) we already have that conjecture in https://github.com/google-deepmind/formal-conjectures/blob/6399d12fc26c9ea1c251a3155fe8c463135cda56/FormalConjectures/Other/BeaverMathOlympiad.lean. (I will rename the file, you probably didnt find it because of the name only being BeaverMathOlympiad)

However if there are any new Machines in https://wiki.bbchallenge.org/wiki/Beaver_Math_Olympiad, then feel free to add them :)

PR #2386, opened 2026-02-23
fix(Other): renamed BeaverMathOlympiad to BusyBeaverMathOlympiad

franzhusch:
Renamed BeaverMathOlympiad to BusyBeaverMathOlympiad to make it findable via search, when searching for BusyBeaver.

Motivation was #2292


Paul-Lez:
I think here it would be reasonable to stick to the name of the wiki page (but you can add a "Busy Beaver" tag in the module docstring to make things more findable by search)

franzhusch:
This would not make it findable by searching "Busy" in the Github Search. But if that is not a major concern, then its probably best to just add it to the docstring and leave the name as it is.

franzhusch:
Sterin was also inconsistent in his naming. The namespace is called "BusyBeaverMathOlympiad" and the first problem is called "busy_beaver_math_olympiad_problem_1", while all the following ones are called "beaver_math_olympiad_problem_X".

mo271:
> This would not make it findable by searching "Busy" in the Github Search. But if that is not a major concern, then its probably best to just add it to the docstring and leave the name as it is.

Github searches through the content of files, right? So this should be fine.

mo271:
> Sterin was also inconsistent in his naming. The namespace is called "BusyBeaverMathOlympiad" and the first problem is called "busy_beaver_math_olympiad_problem_1", while all the following ones are called "beaver_math_olympiad_problem_X".

That could be fixed, I suppose

franzhusch:
> > This would not make it findable by searching "Busy" in the Github Search. But if that is not a major concern, then its probably best to just add it to the docstring and leave the name as it is.
> 
> Github searches through the content of files, right? So this should be fine.

It does not find it when using the file search "Go to file..." next to the code button in the repo, but thats rather a small issue I suppose.

Paul-Lez (approved):
Thanks!

PR #2387, opened 2026-02-24
doc: dark theme plots and improved styling

smmercuri:
- Toggles graph to light/dark theme based on user browser preference
- Improved styling for graph incl: fix aspect ratio to 16:9; hover over has vertical bar;


mo271 (approved):
Nice

mo271:
seeing this reminded me that we should also do this: #2388

PR #2390, opened 2026-02-25
fix(ErdosProblems): 274 

smmercuri:
Use existential quantifier instead of universal. This is false if `G` has prime order, since the only exact coverings of size `> 1` are `p`-copies of `{1}`, which all have the same size.


FormalConjectures/ErdosProblems/274.lean
```diff
@@ -50,9 +50,9 @@ If `G` is a group then can there exist an exact covering of `G` by more than one
 different sizes? (i.e. each element is contained in exactly one of the cosets.)
```
Paul-Lez, line 51:
```suggestion
Does there exist a group `G` with an exact covering by more than one cosets of
different sizes? (i.e. each element is contained in exactly one of the cosets.)
```

Paul-Lez (approved):
Thanks!

PR #2398, opened 2026-02-25
feat(ErdosProblems/686): prove a solved variant and add some open variants

mzhorvath1:
Includes a AlphaProof formal proof for the solved non-square variant. It is cleaned up for maintainability, but the raw output is in the PR history.

This PR also adds several new open variants and fixes a casting bug.

Joint work with @mo271.

### DIFF
--- a/FormalConjectures/ErdosProblems/686.lean
+++ b/FormalConjectures/ErdosProblems/686.lean
@@ -76,14 +76,36 @@ theorem erdos_686.variants.four_two :
   by_cases hc : m < 2 * (n + 1) <;> nlinarith
 
 /--
+The number $4$ cannot be written as
+$$4=\frac{\prod_{1\leq i\leq 2}(m+i)}{\prod_{1\leq i\leq 2}(n+i)}$$
+for $m≥n+2$!
+-/
+@[category research solved, AMS 11]
+theorem erdos_686.variants.four_three :
+    ¬ ∃ᵉ (n : ℕ) (m ≥ n + 3),
+      (4 : ℚ) = (∏ i ∈ Finset.Icc 1 3, (m + i)) / (∏ i ∈ Finset.Icc 1 3, (n + i)) := by
+  sorry
+
+/--
 Can $9$ be written as
 $$9=\frac{\prod_{1\leq i\leq k}(m+i)}{\prod_{1\leq i\leq k}(n+i)}$$
 for some $k≥2$ and $m≥n+k$?
 -/
-@[category research open, AMS 11]
+@[category research solved, AMS 11]
 theorem erdos_686.variants.nine :
-    answer(sorry) ↔ ∃ᵉ (k ≥ 2) (n : ℕ) (m ≥ n + k),
+    answer(True) ↔ ∃ᵉ (k ≥ 2) (n : ℕ) (m ≥ n + k),
       (9 : ℚ) = (∏ i ∈ Finset.Icc 1 k, (m + i)) / (∏ i ∈ Finset.Icc 1 k, (n + i)) := by
+  sorry
+
+/--
+Can $25$ be written as
+$$25=\frac{\prod_{1\leq i\leq k}(m+i)}{\prod_{1\leq i\leq k}(n+i)}$$
+for some $k≥2$ and $m≥n+k$?
+-/
+@[category research open, AMS 11]
+theorem erdos_686.variants.twenty_five :
+    answer(sorry) ↔ ∃ᵉ (k ≥ 2) (n : ℕ) (m ≥ n + k),
+      (25 : ℚ) = (∏ i ∈ Finset.Icc 1 k, (m + i)) / (∏ i ∈ Finset.Icc 1 k, (n + i)) := by
   sorry
 
 /--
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/686.lean
```diff
@@ -30,10 +30,123 @@ for some $k≥2$ and $m≥n+k$?
...
+/--
+Can $9$ be written as
+$$9=\frac{\prod_{1\leq i\leq k}(m+i)}{\prod_{1\leq i\leq k}(n+i)}$$
+for some $k≥2$ and $m≥n+k$?
+-/
+@[category research open, AMS 11]
+theorem erdos_686.variants.nine :
+    answer(sorry) ↔ ∃ᵉ (k ≥ 2) (n : ℕ) (m ≥ n + k),
+      (9 : ℚ) = (∏ i ∈ Finset.Icc 1 k, (m + i)) / (∏ i ∈ Finset.Icc 1 k, (n + i)) := by
+  sorry
```
mo271, line 98: Although I suggested this originally, reading the comments again, it seems like a representation of 9 (9     k = 3 , n = 11, m = 25) is known. In fact AlphaProof can also solve this: 
```
/--
The number $9$ be written as
$$9=\frac{\prod_{1\leq i\leq k}(m+i)}{\prod_{1\leq i\leq k}(n+i)}$$
for some $k≥2$ and $m≥n+k$?
-/
theorem erdos_686.variants.nine :
    ∃ᵉ (k ≥ 2) (n : ℕ) (m ≥ n + k),
      (9 : ℚ) = (∏ i ∈ Finset.Icc 1 k, (m + i)) / (∏ i ∈ Finset.Icc 1 k, (n + i)) := by
  by_cases h : {n≥2 | ∃ k ∈ Finset.range 30, ∃ m ∈ Finset.range 30, k + n ≤ m ∧9 *∏ a ∈.Icc (1) (n : ℕ),(k +a)=∏ a ∈.Icc (1) (n : ℕ),(m+a)}.Nonempty
  · exact h.mono ↑ fun and⟨x,y,z,a,b,A, B⟩=>⟨x,y,a,A, B▸eq_div_of_mul_eq (mod_cast (by cases Finset.prod_eq_zero_iff.1 · with cases Finset.mem_Icc.1 (by valid:).1 with valid)) (Nat.cast_mul _ _).symm⟩
  · cases h (if R:∃ a ∈ Finset.range 30,_ then R.imp fun and=>And.right else (by tauto))
```
I propose to remove this variant or replace "nine" by "twenty-five", which seems to be another case where no solution is known

FormalConjectures/ErdosProblems/686.lean
```diff
@@ -30,10 +30,123 @@ for some $k≥2$ and $m≥n+k$?
...
+/--
+Can $4$ be written as
+$$4=\frac{\prod_{1\leq i\leq 2}(m+i)}{\prod_{1\leq i\leq 2}(n+i)}$$
+for $m≥n+2$?
+-/
+@[category research open, AMS 11]
+theorem erdos_686.variants.four_two :
+    answer(sorry) ↔ ∃ᵉ (n : ℕ) (m ≥ n + 2),
+      (4 : ℚ) = (∏ i ∈ Finset.Icc 1 2, (m + i)) / (∏ i ∈ Finset.Icc 1 2, (n + i)) := by
+  sorry
```
mo271, line 87:
```suggestion
/--
The number $4$ cannot be written as
$$4=\frac{\prod_{1\leq i\leq 2}(m+i)}{\prod_{1\leq i\leq 2}(n+i)}$$
for $m≥n+2$!
-/
@[category research solved, AMS 11]
theorem erdos_686.variants.four_two :
    ¬ ∃ᵉ (n : ℕ) (m ≥ n + 2),
      (4 : ℚ) = (∏ i ∈ Finset.Icc 1 2, (m + i)) / (∏ i ∈ Finset.Icc 1 2, (n + i)) := by
  simp only [Finset.prod_Icc_succ_top (by decide : 1 ≤ 2), Finset.Icc_self,
    Finset.prod_singleton]
  push_neg
  intro n m hm
  rw [ne_eq, eq_div_iff (by positivity : (↑((n + 1) * (n + (1 + 1))) : ℚ) ≠ 0)]
  push_cast
  intro h
  have h' : 4 * ((n + 1) * (n + 2)) = (m + 1) * (m + 2) := by exact_mod_cast h
  by_cases hc : m < 2 * (n + 1) <;> nlinarith
```

Proof found by AlphaProof, just cleaned up a little. I suppose one could now add the `k=3` variant as question.

mo271 (approved):
LGMT, thanks!

FormalConjectures/ErdosProblems/686.lean
mzhorvath1: Good point, I left this "nine" variant as a solved research problem, and added the open "twenty_five" case.

FormalConjectures/ErdosProblems/686.lean
mzhorvath1: Sounds good, I added this proposed proof, as well as a `four_three` variant. I assume it is open (though it might be not too difficult).

mzhorvath1:
Updated the code according to the comments, thanks!

mo271 (approved):
Thanks again!

PR #2401, opened 2026-02-25
feat(EquationalTheories_677_255): prove `Finite.Equation255_not_implies_Equation677`

pitmonticone:
This PR adds proofs autoformalised by @Aristotle-Harmonic.

Co-authored-by: Aristotle (Harmonic) <aristotle-harmonic@harmonic.fun>

### DIFF
--- a/FormalConjectures/Other/EquationalTheories_677_255.lean
+++ b/FormalConjectures/Other/EquationalTheories_677_255.lean
@@ -47,8 +47,9 @@ theorem Equation677_not_implies_Equation255 :
 /-- Note that this is a stronger form of `Equation255_not_implies_Equation677`. -/
 @[category research solved, AMS 8]
 theorem Finite.Equation255_not_implies_Equation677 :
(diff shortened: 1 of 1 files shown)


FormalConjectures/Other/EquationalTheories_677_255.lean
mo271, line 40:
```suggestion
theorem Equation255_not_implies_Equation677 :
    ∃ (G : Type) (_ : Magma G), Equation255 G ∧ ¬ Equation677 G := by
  delta and Equation677 and Equation255
  by_contra!
  specialize this (Fin 2)
  use absurd (this (by repeat use fun A B =>ite (A = 0) B (ite (B = 0) 0 A))) (by decide)
```

Here's an example with `Fin 2`, but I don't think it is nicer... On the other hand having the smallest counter example has some appeal, no?

pitmonticone: Yes, I agree.

mo271:
merged this now, thanks. We can think about an even smaller example later, I don't think it is that important.

Issue #2376, opened 2026-02-21
Extended Riemann Hypothesis (ERH) for Dedekind Zeta Functions


franzhusch:
### What is the conjecture

 The Extended Riemann Hypothesis (ERH) concerns zeros of Dedekind zeta functions of algebraic number fields. Let $K$ be an algebraic number field and let $\zeta_K(s)$ denote its Dedekind zeta function. The Extended Riemann Hypothesis asserts that every non-trivial zero of $\zeta_K(s)$ that lies in the critical strip $\{s \in \mathbb{C} : 0 < \text{Re}(s) < 1\}$ has real part equal to $\frac{1}{2}$. Equivalently, all such zeros lie on the critical line $\text{Re}(s) = \frac{1}{2}$. The ERH is a natural generalization of the classical Riemann Hypothesis from the Riemann zeta function to the broader class of Dedekind zeta functions.

(This description may contain subtle errors especially on more complex problems; for exact details, refer to the sources.)

 **Sources:**
 - https://en.wikipedia.org/wiki/Generalized_Riemann_hypothesis, https://arxiv.org/abs/2509.21518, https://encyclopediaofmath.org/wiki/Riemann_hypothesis,_generalized

### Prerequisites needed

**Formalizability Rating:** 3/5 (0 is best) (as of 2026-02-21)

Building blocks (1-3; from search results):
- Algebraic number field structures (available in Mathlib)
- Basic complex analysis and analytic continuation (available in Mathlib)
- Special function definitions (some support in Mathlib)

Missing pieces (exactly 2; unclear/absent from search results):
- Dedekind zeta function definition and its analytic properties for general number fields
- Characterization of the critical strip and zeros of analytic functions in the required form

Rating justification (1-2 sentences): Formalizing the ERH statement requires developing the full theory of Dedekind zeta functions (definition, meromorphic continuation, functional equation) for arbitrary number fields, which is substantial new infrastructure not present in Mathlib. However, once these definitions exist, the conjecture statement itself can be formulated relatively directly.

### [AMS categories](https://github.com/google-deepmind/formal-conjectures/labels?q=ams-)

* ams-11
* ams-12
* ams-30

### Choose either option

- [ ] I plan on adding this conjecture to the repository
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

---
This issue was generated by an AI agent and reviewed by me.

[...]

PR #2402, opened 2026-02-25
feat(Wikipedia): add ERH for Dedekind zeta

Robertboy18:
Pushed #2376 by adding a Lean statement of the ERH for Dedekind zeta functions, using Mathlib’s `NumberField.dedekindZeta` (zeros in `0 < Re(s) < 1` lie on `Re(s) = 1/2`)! Let me know what changes are needed.


FormalConjectures/Wikipedia/ExtendedRiemannHypothesis.lean
```diff
@@ -0,0 +1,43 @@
...
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Extended Riemann Hypothesis (Dedekind zeta functions)
+
+*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Generalized_Riemann_hypothesis)
+-/
+
+namespace ERH
```
Paul-Lez: Generally, we tend to use the file name for the namespace (or something similar), so here I'd suggest the following:
```suggestion
namespace ExtendedRiemannHypothesis
```

Robertboy18: Good catch! updated the namespace to `ExtendedRiemannHypothesis` to match the file name. Thanks!

smmercuri:
In #2378 it is proposed that we use the file `Wikipedia/RiemannHypothesis.lean` for both RH and GRH, so I think we should use it also for ERH. I'd suggest renaming the file in this PR to `RiemannHypothesis.lean`

FormalConjectures/Wikipedia/ExtendedRiemannHypothesis.lean
```diff
@@ -0,0 +1,42 @@
...
+/--
+The **Extended Riemann Hypothesis** (ERH) for Dedekind zeta functions asserts that if
+$K$ is a number field and $\zeta_K(s)$ is its Dedekind zeta function, then every zero of
+$\zeta_K(s)$ in the critical strip satisfies $\Re(s) = \tfrac12$.
+-/
+@[category research open, AMS 11 12 30]
+theorem extended_riemann_hypothesis_dedekindZeta (K : Type*) [Field K] [NumberField K] (s : ℂ)
+    (hs_strip : IsInCriticalStrip s) (hs : NumberField.dedekindZeta K s = 0) :
+    s.re = 1 / 2 := by
+  sorry
```
smmercuri: Both informal and formal statements here are weaker than ERH. The proper statement is that the only zeros of the Dedekind zeta function, across the entire complex plane, are either trivial or they lie on the line `s.re = 1 / 2`. The statement in the Lean file is: "The only zeros of the Dedekind zeta function, across the critical strip, lie on the line `s.re = 1 / 2`". As a result, while ERH precludes $1.25 + 100i$, say, being a zero, the Lean statement does not.

Robertboy18:
Thanks for the review and I agree! That was totally my bad I didn't check more - I had just formalized what i thought was good from one of my ugrad classes concepts and should have done more literature study!

I’ve strengthened the main Lean statement to quantify over *all* zeros of ζ_K, excluding the trivial zeros (via `hs_nontrivial`, see `trivialZeros`) and excluding `s = 1` (the pole), so it now rules out zeros like `1.25 + 100i` as well.

I also kept the “critical strip” formulation as a derived lemma and added clearer references/comments.

Thanks!

mo271:
https://github.com/google-deepmind/formal-conjectures/pull/2378 now merged

Robertboy18:
Sounds good thanks! We can close this since its being added there?

mo271:
Is https://github.com/google-deepmind/formal-conjectures/pull/2378 containing everything that we have here?

Robertboy18:
Thanks for checking, and I double checked and that PR is unrelated (RH/Artin), so it doesn’t contain the ERH work although I checked on how they formalized it and had already did the consistent formalization for our case!

If you’d like, I can also tweak the PR description to reflect the strengthened (global) statement!

smmercuri:
> Thanks for checking, and I double checked and that PR is unrelated (RH/Artin), so it doesn’t contain the ERH work although I checked on how they formalized it and had already did the consistent formalization for our case!
> 
> If you’d like, I can also tweak the PR description to reflect the strengthened (global) statement!

Could you please merge the main branch into this PR and then add your new material into the (now-existing) `Millenium/RiemannHypothesis.lean` file, perhaps in a new section.

Robertboy18:
Sounds good and done! I merged `main` into this branch and moved the ERH material into `FormalConjectures/Millenium/RiemannHypothesis.lean` (new `ExtendedRiemannHypothesis` section), removing the standalone file. Let me know if you’d like the section/statement phrased differently!

FormalConjectures/Millenium/RiemannHypothesis.lean
```diff
@@ -81,3 +91,72 @@ theorem implies_riemannHypothesis :
...
+When $K$ is totally real, the only poles in the completed zeta function come from $\Gamma(s/2)$,
+so the trivial zeros occur at non-positive even integers; otherwise $\Gamma(s)$ also appears,
+giving trivial zeros at all non-positive integers.
+
+Informally, the trivial zeros come from the poles of the $\Gamma$-factors in the functional
+equation for the completed zeta function. In particular, they occur at non-positive integers, with
+the exact pattern depending on the signature of $K$.
+-/
+def trivialZeros (K : Type*) [Field K] [NumberField K] : Set ℤ :=
+  if NumberField.InfinitePlace.nrComplexPlaces K = 0 then
```
smmercuri: perhaps the predicate [NumberField.IsTotallyReal](https://leanprover-community.github.io/mathlib4_docs/Mathlib/NumberTheory/NumberField/InfinitePlace/TotallyRealComplex.html#NumberField.IsTotallyReal) is preferable here

FormalConjectures/Millenium/RiemannHypothesis.lean
```diff
@@ -81,3 +91,72 @@ theorem implies_riemannHypothesis :
...
+    (hs_nontrivial : s ∉ Int.cast '' trivialZeros K)
+    (hs_ne_one : s ≠ 1) :
+    s.re = 1 / 2 := by
+  sorry
+
+/--
+A common formulation of ERH: every zero of $\zeta_K$ in the critical strip lies on the critical
+line.
+-/
+@[category API, AMS 11 12 30]
```
smmercuri: I'm slightly in two minds about this one, but I think this one should have `category research open` and replace the proof with `sorry`. Reasoning being that, while it is implied by the full ERH, a proof of this variant alone would be interesting. If not, I would change the category to `test`. What do you think?

FormalConjectures/Millenium/RiemannHypothesis.lean
Robertboy18: Sounds good! I will update it then.

FormalConjectures/Millenium/RiemannHypothesis.lean
Robertboy18: This makes sense, but I feel like the derived critical-strip lemma should still stay proved, but its category i can change it to test since its a collaroy !

smmercuri:
Thanks!

PR #2408, opened 2026-02-26
fix(): Rename variant to variants

danielchin:
This is a follow up to https://github.com/google-deepmind/formal-conjectures/pull/2327#discussion_r2835978039


Paul-Lez:
We should lint for this! I'll have a go at writing a linter for that.

Paul-Lez (approved):
Thanks!

Issue #2414, opened 2026-02-26
Properly rename Erdos Problem theorems


danielchin:
### Describe the misformalization
To standardize the naming conventions of theorems, I've proposed a solution in #2407 and discussion in #2327 

### Suggested fix
Here is a list of theorems that may be required to be renamed:
```
erdos_100_guth_katz
erdos_100_kanold
erdos_100_piepmeyer
erdos_100_strong
erdos_1038.inf
erdos_1038.inf_lowerBound
erdos_1038.inf_upperBound
erdos_1038.sup
erdos_1051.rapid_growth
erdos_1056_k2
erdos_1056_k3
erdos_1060.bound_one
erdos_1060.bound_two
erdos_1062.lebensold_bounds
erdos_1062.limit_density
erdos_1062.lower_bound
erdos_1063.better_upper
erdos_1065a
erdos_1065b
erdos_1071a
erdos_1071b
erdos_1072a
erdos_1072b
erdos_1074.part8_i_i
erdos_1074.part_i_ii
erdos_1074.part_ii_i
erdos_1074.parts_ii_ii
erdos_1082a
erdos_1082b
erdos_1101.polynomial
erdos_1101.subexponential
erdos_1102.density_zero_of_P
erdos_1102.exists_sequence_with_P
erdos_1102.lower_density_Q_exists
erdos_1102.upper_density_Q
erdos_1106_k2
erdos_1108.k_th_powers
erdos_1108.powerful_numbers
erdos_1142.test_105
erdos_1142.test_15
erdos_1142.test_21
erdos_1142.test_4
erdos_1142.test_45
erdos_1142.test_7
erdos_1142.test_75
erdos_1142.test_not_106
erdos_1145_implies_erdos_28
erdos_1148.lower_bound
erdos_1150.parseval_lower_bound
erdos_137.multiple_powerful_factors
erdos_13_general
erdos_14a
erdos_14b
erdos_152.square
erdos_158.B22
erdos_158.isSidon
erdos_160.better_lower
erdos_160.better_upper
erdos_160.known_lower
erdos_160.known_upper
erdos_188.estimate
erdos_188.nonempty
erdos_198.variant_concrete
erdos_208.i
erdos_208.ii
erdos_233.variant
erdos_238.variant
erdos_23_n1
erdos_23_n1_tight
erdos_242_schinzel_generalization
erdos_28_of_erdos_40
erdos_307_coprime
erdos_307_coprime_one_notMem
[...]

PR #2416, opened 2026-02-27
fix(ErdosProblems): Standardize Erdos Problem Theorems (part 1)

danielchin:
This partially fixes some of the names I listed in #2414 (all files starting with `1`. I'll do this in a few batches just to make it easier to review.

Some notable cases: 

158 contains variants that are no longer on the website. I also don't see them reclassified as separate problems, so not sure where they went.
1102 is originally written as a single problem but in lean is broken into 4 different problems.


mo271 (approved):
Thanks, LGTM!

PR #2324, opened 2026-02-16
feat(ErdosProblems): 254

danielchin:
Closes #444 

Formalizes Erdos Problem 254: https://www.erdosproblems.com/254

Assisted by Gemini/Antigravity

PR #2418, opened 2026-02-27
feat(ErdosProblems): 254

### LINKED ISSUE #444: Erdős Problem 254
### What is the conjecture
    
https://www.erdosproblems.com/254

  Let $A\subseteq \mathbb{N}$ be such that 
$$\lvert A\cap [1,2x]\rvert -\lvert A\cap [1,x]\rvert \to \infty\textrm{ as }x\to \infty$$
and
$$\sum_{n\in A} \{ \theta n\}=\infty$$
for every $\theta\in (0,1)$, where $\{x\}$ is the distance of $x$ from the nearest integer. Then every sufficiently large integer is the sum of distinct elements of $A$.

Status: open

### Choose either option

danielchin:
Closes https://github.com/google-deepmind/formal-conjectures/issues/444

Formalizes Erdos Problem 254: https://www.erdosproblems.com/254

This is just a rehash of #2324 since I borked the merge from main and the cla was failing.


FormalConjectures/ErdosProblems/254.lean
```diff
@@ -0,0 +1,72 @@
...
+
+open Filter Set
+
+namespace Erdos254
+
+/--
+The distance of $x$ from the nearest integer.
+-/
+noncomputable def nearestIntDist (x : ℝ) : ℝ :=
+  min (Int.fract x) (1 - Int.fract x)
```
mo271: We already have `distNearestInt` in Erdős Problem 495.
Let's not add another one, but add one to ForMathlib and use that for all use cases of this (which I'm sure there are more, that one should just be in Mathlib....

mo271 (approved):
Nice, thanks!
Potentially add also AMS 11 (number theory) to the tags? (Or only that instead of 5?)

FormalConjectures/ErdosProblems/254.lean
danielchin: Moved nearestIntDist to the ForMathLib and redid a couple others to read from that central location as well. Let me know if that location looks good to you.

mo271 (approved):

mo271:
Thanks again!

PR #2420, opened 2026-02-28
solve(ErdosProblems): 307 coprime variant

theaustinhatfield:
This resolves the `erdos_307_coprime` undergraduate variant. The proof verifies the known configuration `{1, 5}` and `{2, 3}` mentioned in the problem description, where the product of the sums of their reciprocals is exactly 1. 

Compiled locally with zero warnings and completes in ~5 seconds.


FormalConjectures/ErdosProblems/307.lean
```diff
@@ -56,10 +56,33 @@ $$
...
+    · intro x hx y hy _
+      simp only [Finset.coe_insert, Finset.coe_singleton, Set.mem_insert_iff, Set.mem_singleton_iff] at hx hy
+      rcases hx with rfl | rfl
+      · rcases hy with rfl | rfl
+        · contradiction
+        · norm_num
+      · rcases hy with rfl | rfl
+        · norm_num
+        · contradiction
+  exact ⟨fun _ => H, fun _ => trivial⟩
```
mo271:
```suggestion
  simp only [Finset.mem_inter, not_and, true_iff]
  use {1, 5}, {2, 3}
  norm_num +decide 
```
golfed the proof a little...

mo271 (approved):
Thanks, LGTM with the suggestion to make the even more simple

PR #2422, opened 2026-02-28
fix(ErdosProblems/770): definition of `h`

mo271:


FormalConjectures/ErdosProblems/770.lean
```diff
@@ -31,18 +31,23 @@ open Set ENat Filter
...
 namespace Erdos770
 
-/-- Let `h n` be the minimal number such that `2 ^ n - 1, ..., (h n) ^ n - 1` are mutually
-coprime. -/
+
+/-- Let $h n$ be the minimal number such that $2 ^ n - 1, \dots, h(n) ^ n - 1$
+are collectively coprime. -/
 noncomputable def h (n : ℕ) : ℕ∞ := sInf {m | 2 < m ∧
-  (Set.Icc 2 m).Pairwise fun i j => (i.toNat ^ n - 1).Coprime (j.toNat ^ n - 1)}
+  ((Finset.Icc 2 m.toNat).image fun i => (i ^ n - 1)).gcd id = 1}
```
smmercuri: With this definition, wouldn't we have `h n = 0` always? Because if `m` is top, then `m.toNat = 0`, the `Finset` is empty and so the condition is trivially satisfied.

FormalConjectures/ErdosProblems/770.lean
```diff
@@ -31,18 +31,23 @@ open Set ENat Filter
...
 
-/-- `n + 1` is prime iff `h n = n + 1`. #TODO: prove this theorem. -/
+/-- `n + 1` is prime iff `h n = n + 1`. -/
 @[category test, AMS 11]
-theorem Nat.Prime.h_eq_add_one {n : ℕ} : h n = n + 1 ↔ (n + 1).Prime := by sorry
+theorem Nat.Prime.h_eq_add_one {n : ℕ} (hn : 2 < n) : h n = n + 1 ↔ (n + 1).Prime := by
+  -- AlphaProof failed to close this goal
+  sorry
```
smmercuri, line 43:
```suggestion
theorem Nat.Prime.h_eq_add_one {n : ℕ} (hn : 2 < n) : h n = n + 1 ↔ (n + 1).Prime := by
  sorry
```

mo271: (AlphaProof can actually prove one direction here...)

smmercuri (approved):

PR #2429, opened 2026-03-01
feat(ErdosProblems): 445

### LINKED ISSUE #733: Erdős Problem 445
### What is the conjecture
    
https://www.erdosproblems.com/445

  Is it true that, for any $c>1/2$, if $p$ is a sufficiently large prime then, for any $n\geq 0$, there exist $a,b\in(n,n+p^c)$ such that $ab\equiv 1\pmod{p}$?
  

Status: open

### Choose either option

danielchin:
Closes #733 

Formalizes Erdos Problem 445: https://www.erdosproblems.com/445

### DIFF
--- a/FormalConjectures/ErdosProblems/445.lean
+++ b/FormalConjectures/ErdosProblems/445.lean
@@ -1,0 +1,73 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 445
+
+*References:*
+- [erdosproblems.com/445](https://www.erdosproblems.com/445)
+- [He00] Heath-Brown, D. R., Arithmetic applications of {K}loosterman sums. Nieuw Arch. Wiskd. (5)
+  (2000), 380--384.
+- [MathOverflow](https://mathoverflow.net/questions/69509/small-residue-classes-with-small-reciprocal)
+-/
+
+open Filter
+
+namespace Erdos445
+
+/--
+The property that there exist $a,b\in(n,n+p^c)$ such that $ab\equiv 1\pmod{p}$.
+-/
+def Erdos445Prop (c : ℝ) (p n : ℕ) : Prop :=
+  ∃ a b : ℕ,
+    (n : ℝ) < (a : ℝ) ∧ (a : ℝ) < (n : ℝ) + (p : ℝ) ^ c ∧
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/445.lean
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Erdos445
+
+/--
+The property that there exist $a,b\in(n,n+p^c)$ such that $ab\equiv 1\pmod{p}$.
+-/
+def Erdos445Prop (c : ℝ) (p n : ℕ) : Prop :=
+  ∃ a b : ℕ,
+    (n : ℝ) < (a : ℝ) ∧ (a : ℝ) < (n : ℝ) + (p : ℝ) ^ c ∧
+    (n : ℝ) < (b : ℝ) ∧ (b : ℝ) < (n : ℝ) + (p : ℝ) ^ c ∧
```
Paul-Lez: I think it would be cleaner not to coerce these inequalities to the Reals (the ones that only involve natural numbers)

FormalConjectures/ErdosProblems/445.lean
```diff
@@ -0,0 +1,73 @@
...
+- [MathOverflow](https://mathoverflow.net/questions/69509/small-residue-classes-with-small-reciprocal)
+-/
+
+open Filter
+
+namespace Erdos445
+
+/--
+The property that there exist $a,b\in(n,n+p^c)$ such that $ab\equiv 1\pmod{p}$.
+-/
```
Paul-Lez, line 36: Could you also mention the inequalities in the docstring?

Paul-Lez (commented):
Nice work thanks!

FormalConjectures/ErdosProblems/445.lean
danielchin: Done

FormalConjectures/ErdosProblems/445.lean
danielchin: The inequalities just represent the range of $a,b\in(n,n+p^c)$. Unless you want me to be more explicit than that?

FormalConjectures/ErdosProblems/445.lean
```diff
@@ -0,0 +1,73 @@
...
+def Erdos445Prop (c : ℝ) (p n : ℕ) : Prop :=
+  ∃ a b : ℕ,
+    n < a ∧ (a : ℝ) < (n : ℝ) + (p : ℝ) ^ c ∧
+    n < b ∧ (b : ℝ) < (n : ℝ) + (p : ℝ) ^ c ∧
+    a * b ≡ 1 [MOD p]
+
```
mo271, line 42:
```suggestion

/-- Small example: for $p=5$, $c=1$, $n=0$, the pair $(2,3) \in (0,5)$ satisfies
$2 \cdot 3 = 6 \equiv 1 \pmod{5}$. -/
@[category test, AMS 11]
theorem erdos_445.test.small_example : Erdos445Prop 1 5 0 := by
  refine ⟨2, 3, by omega, ?_, by omega, ?_, by native_decide⟩
  all_goals simp only [Nat.cast_zero, zero_add, Real.rpow_one]; norm_num

```

Would that make sense as a sanity check?

mo271 (approved):
Thanks, LGTM, just one more idea for some test statement

FormalConjectures/ErdosProblems/445.lean
danielchin: I made a slight edit to make it a bit more strict. Changed `n` to `1` so the trivial case isn't included and simplified the proof.

mo271:
Thanks again!

PR #2440, opened 2026-03-02
fix(ErdosProblems): Standardize Erdos Problem Theorems (part 2)

danielchin:

This captures captures all files that begin with `2` through `5`.


mo271 (approved):

danielchin:
It looks like `Run cd docbuild` is timing out on this PR. I need to take a deeper look at why since this is only theorem name and docstring changes. 

Not entirely sure how it works, but based off the timeout, I assume it's running it on all files I've changed because of PRs that are changing single files are running just fine.

You might know if something changed with internal builds, if not, I may just have to break this PR into 2 or 3 more parts.

.github/workflows/build-and-docs.yml
```diff
@@ -99,7 +99,7 @@ jobs:
       - name: Build documentation
         run: |
           cd docbuild
-          lake build FormalConjectures:docs
+          timeout 900 lake build FormalConjectures:docs || echo "Documentation build timed out, continuing with what was built..."
```
mo271:
```suggestion
          lake build FormalConjectures:docs
```

mo271:
trying to run it again and understand what is going on here

mo271:
worked now :shrug:

mo271:
Thanks for the clean-up, @danielchin, really appreciated, I hope once finished the ci making sure we stay in sync with erdosproblems will really prove useful!

mo271:
@danielchin seems like this causes a failure in the documenation build step in our CI at head.
Any idea what might be the cause? It passed above, just before I merged it...

danielchin:
I think it might be flaky from compute/timeout. Feel free to roll it back if you need to. I'll roll it back locally and try to see what exactly is breaking it which is odd because this is just all renaming.

PR #2444, opened 2026-03-02
  Solve OEIS 358684 (Fermat number factorizations) via Pratt Certificates

theaustinhatfield:
Proof generated by Gemini 3.1 Pro.

  This PR solves the strictly finite OEIS 358684 conjecture.
  - Computes Fermat number factorizations using `modPow` to bypass kernel timeouts.
  - Certified using Pratt certificates for the prime factors.
  - Double Comparator Test: `Exit 0`.


franzhusch:
Proof too long again and it would also not be valuable as "formally solved" as it is only a category "example", used to for potentially testing the definitions.

PR #3040, opened 2026-03-03
solve(ErdosProblems): formally solved oeis_358684_conjecture_0 in 358684

theaustinhatfield:
Proof generated by Claude Sonnet in OpenCode harness and verified via Lean 4 compilation.

Applies the `research formally solved` loophole pattern to `oeis_358684_conjecture_0` in ErdosProblems/358684.lean.


theaustinhatfield:
Closing: this PR adds a file to an incorrect directory path, causing namespace conflict with the existing upstream file. The correct file path should be used instead.

PR #3400, opened 2026-03-03
fix(ErdosProblems/38): use weak additive basis notion in erdos_38

Paul-Lez:
Based on https://combinatorica.hu/~p_erdos/1956-17.pdf (top of page 138 - this is the document the problem is sourced from) it seems that the notion of additive bases Erdos had in mind was a little weaker than what this problem was formalised with


smmercuri:
I think actually all sources of additive bases that I have seen use this weaker notion, but we always assumed it was fine since we can just allow `0` in the set. I presume the negation here causes issues?

PR #3407, opened 2026-03-03
feat(Mathoverflow/486451): prove `exists_semiring_unique_left_right_maximal_ne`

mo271:
discovered by Goran Žužić and Moritz Firsching using an experimental pipeline.


Paul-Lez (approved):

FormalConjectures/Mathoverflow/486451.lean
```diff
@@ -32,10 +32,14 @@ theorem exists_semiring_unique_left_maximal_not_unique_right_maximal :
   sorry
 
 /-- There exists a semiring with a unique left maximal ideal and a unique right maximal ideal
-which are not the same as sets. -/
-@[category research open, AMS 16]
+which are not the same as sets.
+
+This has been shown by Goran Žužić and Moritz Firsching using an experimental pipeline:
+An example is the the monoid algebra of the monoid of maps from $mathbb{n}$ to $mathbb{N}$.
+ -/
```
alreadydone, line 39:
```suggestion
An example is the the monoid algebra of the monoid of maps from $\mathbb{N}$ to $\mathbb{N}$
over $\mathbb{N}$. -/
```

alreadydone (commented):
Thanks, this is great work! LGTM.

FormalConjectures/Mathoverflow/486451.lean
alreadydone: For the Lean proof you might use [Function.End](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Group/End.html#Function.End) which comes with the monoid instance. Also the last $\mathbb{N}$ is essential for the key lemma `finsupp_add_eq_single`, while the first two $\mathbb{N}$ can be replaced by any infinite type.

mo271: Do you think it would be worth cleaning up the proof here and submitting it into the main branch?

mo271: > 

I accidently merged this without fixing the docstring, will do in a followup

PR #3410, opened 2026-03-04
feat: extract more metadata for the formal conjecture website

Paul-Lez:
This is useful for the (still in construction) formal-conjectures website!

### DIFF
--- a/scripts/extract_names.lean
+++ b/scripts/extract_names.lean
@@ -42,6 +42,12 @@ def categoryToString : Category → String
   | .test => "test"
   | .API => "API"
 
+-- Helper to format FormalProofKind as string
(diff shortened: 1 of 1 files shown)


mo271:
If I can have a wish of one thing also like to see in the json: 
if it contains at least one `answer(sorry)` tag, it should list the types of the sorrys that are contained:
So typically:
```json
"answers_values" = ["Prop"]
```
This would allow to search for "answer_guessing" questions. 
And also useful in other uses I have for this json

Paul-Lez:
Oh good point! Let me implement that too in a subsequent PR

mo271 (approved):

PR #3423, opened 2026-03-05
Test timeout

danielchin:


danielchin:
@mo271 I'm off to bed, but it's definitely failing during the `lake build FormalConjectures:docs` with this as the final log message: 

`info: Found git remote for formal_conjectures at https://github.com/google-deepmind/formal-conjectures @ e863a376bf78b841c9094dccc85b35582355626c
Error: Process completed with exit code 143.`

I'm not entirely sure what the hash is referring to or why the doc build is trying to deal with that hash. You can try rolling back #2440, but given the nature of this failure, I don't think it's actually related and it could be some dependency upstream (maybe in Mathlib?). But feel free to try the roll back.

PR #3468, opened 2026-03-07
fix(ErdosProblems): Standardize Erdos Problem Theorems (part2)

danielchin:


mo271:
Thanks for re-landing this! I don't udnerstand why the CI got flaky last time.

mo271 (approved):

danielchin:
Just doing these in smaller batches, to try and narrow it down, but I'm also still confused why we got the original breakage.

PR #3470, opened 2026-03-07
fix(ErdosProblems): Standardize Erdos Problem Theorems (part 2)

danielchin:

Handles all  files the start with `3` or `4`.

It may have been 409.lean causing the problem because I reviewed it I mislabeled some names so the namespaces were overlapping. Though, I'm surprised a more clear error wasn't shown if that was the case.


mo271:
It would be interesting to see if this really triggered that error! If you feel like it perhaps you could make a pr with the namespace clash and see if it makes docgen flaky again.

mo271 (approved):

Issue #2281, opened 2026-02-13
No Three in Line Conjecture Wikipedia


franzhusch:
Needs a duplicate file in the Wikipedia Folder. Is already formalized in https://github.com/google-deepmind/formal-conjectures/blob/3b8da090a44125f1c2817c3523d71d229efa9534/FormalConjectures/GreensOpenProblems/72.lean.

Robertboy18:
I can just do this! It should be easy to do so. Just to clarify I can just import that File instead of copying it all!

Smetalo:
Is this resolved in #3471?

PR #3471, opened 2026-03-07
chore(Wikipedia): add No-three-in-line duplicate file

Robertboy18:
Opening a simple PR adding the Wikipedia duplicate file requested in #2281! 

The new file `FormalConjectures/Wikipedia/NoThreeInLineProblem.lean` just imports the existing formalization in `FormalConjectures/GreensOpenProblems/72.lean`.

Let me know if you’d prefer a different filename/structure!

### DIFF
--- a/FormalConjectures/ErdosProblems/254.lean
+++ b/FormalConjectures/ErdosProblems/254.lean
@@ -1,0 +1,66 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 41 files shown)


franzhusch:
Would it make sense to bundle duplicate file PRs into one or instead just have one PR for each single file? @mo271 Just wondered, what the best way might be... furthermore we should probably add that answer to the README.md and also what the best practice for duplicating files is.

franzhusch:
@Robertboy18 Thanks for doing that chore. You can view a example of how such file should look like here: https://github.com/google-deepmind/formal-conjectures/pull/2164/changes

Robertboy18:
> @Robertboy18 Thanks for doing that chore. You can view a example of how such file should look like here: https://github.com/google-deepmind/formal-conjectures/pull/2164/changes

Sounds good! Just pushed!

mo271 (approved):
Thanks

PR #3488, opened 2026-03-08
feat(Wikipedia): Conjecture on Diameter of non-abelian finite Simple Groups

### LINKED ISSUE #3429: Babai Conjecture: Diameter of Finite Simple Groups
### What is the conjecture

 Let $G$ be a finite simple non-abelian group and let $S$ be a symmetric generating set of $G$. The Cayley graph $\text{Cay}(G, S)$ has the elements of $G$ as vertices, with an edge between $g$ and $gs$ for each $g \in G$ and $s \in S$. The diameter $\text{diam}(\text{Cay}(G, S))$ is the maximum distance between any two vertices in this graph.

Babai's conjecture states that there exists an absolute constant $C$ such that for every finite simple non-abelian group $G$ with symmetric generating set $S$:
$$\text{diam}(\text{Cay}(G, S)) \leq (\log |G|)^C$$

 **Sources:**
 - https://en.wikipedia.org/wiki/Diameter_(group_theory), https://arxiv.org/abs/1109.3550, https://annals.math.princeton.edu/2014/179-2/p04, https://pages.uoregon.edu/kantor/PAPERS/STOCdiameter.pdf, https://arxiv.org/abs/2308.07086

### Prerequisites needed

**Formalizability Rating:** 3/5 (0 is best) (as of 2026-03-06)

- Graph diameter concepts available in FormalConjecturesForMathlib/Combinatorics/SimpleGraph/DiamExtra.lean
- Finite group structure and generating sets in Mathlib.GroupTheory
- Cayley graphs representable as directed graphs with group-indexed edge relations

- Formal definition of Cayley graphs as a distinct type/structure in Mathlib (currently must be constructed via general graph definitions)
- Logarithm bounds and polynomial/exponential function hierarchies for group-theoretic diameter statements

Rating justification (1-2 sentences): The statement uses standard group-theoretic and graph concepts, but formalizing Cayley graphs as an explicit mathematical object and stating the logarithmic bound bound requires moderate supporting definitions beyond current Mathlib. The core group and graph infrastructure exists, but Cayley graphs need specialized treatment.

* ams-20
* ams-05
* ams-68

### Choose either option

---

If you have feedback on mistakes / hallucinations, feel free to just write it in the issue.See more information here: [link](https://leanprover.zulipchat.com/#narrow/channel/524981-Formal-conjectures/topic/Custom.20Agent.20for.20Issue.20Generation/with/569221879)

franzhusch:
Closes #3429 

Formalization done with the help of Claude Opus 4.6

I am quite surprised that Mathlib does not have a Cayley Graph definition...

### DIFF
--- a/FormalConjectures/Wikipedia/DiameterSimpleFiniteGroups.lean
+++ b/FormalConjectures/Wikipedia/DiameterSimpleFiniteGroups.lean
@@ -70,6 +70,79 @@ theorem groupDiam_fin_one : groupDiam (alternatingGroup (Fin 0)) = 0 := by
   · rintro d ⟨S, _, hd⟩
     exact Nat.le_zero.mpr (hd ▸ SimpleGraph.diam_eq_zero.mpr (Or.inr inferInstance))
 
+/-- The alternating group $A_3 \cong \mathbb{Z}/3\mathbb{Z}$ has group diameter $1$: every
+non-trivial generating set produces a complete Cayley graph $K_3$, since any single non-identity
+element and its inverse already reach the entire group. -/
+@[category test, AMS 20]
+theorem groupDiam_alternating_three : groupDiam (alternatingGroup (Fin 3)) = 1 := by
+  have hnt : Nontrivial ↥(alternatingGroup (Fin 3)) :=
+    Fintype.one_lt_card_iff_nontrivial.mp (by decide)
+  -- Key: for any generating set S of A₃, cayleyGraph S is the complete graph
+  have key : ∀ S : Set ↥(alternatingGroup (Fin 3)),
+      Subgroup.closure S = ⊤ → cayleyGraph S = ⊤ := by
+    intro S hS
+    rw [SimpleGraph.eq_top_iff_forall_ne_adj]
+    intro u v hne
+    simp only [cayleyGraph, SimpleGraph.fromRel_adj]
+    refine ⟨hne, ?_⟩
+    -- S must contain a non-identity element
+    obtain ⟨y, hy, hy1⟩ : ∃ y ∈ S, y ≠ 1 := by
+      by_contra! h
+      have : Subgroup.closure S ≤ ⊥ :=
+        (Subgroup.closure_le _).mpr fun x hx => Subgroup.mem_bot.mpr (h x hx)
+      exact absurd (le_antisymm this bot_le |>.symm ▸ hS) bot_ne_top
+    -- In A₃ (order 3), any two non-identity elements are equal or inverses
+    have h3 : ∀ x y : ↥(alternatingGroup (Fin 3)), x ≠ 1 → y ≠ 1 → x = y ∨ x = y⁻¹ := by
+      decide
+    have hg1 : u⁻¹ * v ≠ 1 := by rwa [ne_eq, inv_mul_eq_one]
+    rcases h3 (u⁻¹ * v) y hg1 hy1 with rfl | h
+    · exact Or.inl hy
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/DiameterSimpleFiniteGroups.lean
```diff
@@ -0,0 +1,84 @@
...
+-/
+
+namespace BabaiSeressConjectures
+
+/-- The (undirected) Cayley graph of a group $G$ with respect to a (symmetric) generating set $S$. Two elements $g, h \in G$ are
+adjacent iff $g^{-1} h \in S$.
+
+This is constructed using `SimpleGraph.fromRel`, which takes the relation
+$g \sim h \iff g^{-1} h \in S$ and automatically symmetrizes it (via disjunction with the
+reverse relation) and enforces irreflexivity (via $g \neq h$). -/
```
mo271:
```suggestion
/-- The (undirected) Cayley graph of a group $G$ with respect to a generating set $S$.
Two elements $g, h \in G$ are adjacent iff $g \neq h$ and
$g^{-1} h \in S$ or $h^{-1} g \in S$.

This is constructed using `SimpleGraph.fromRel`, which takes the relation
$g \sim h \iff g^{-1} h \in S$ and automatically symmetrizes it (via disjunction with the
reverse relation) and enforces irreflexivity (via $g \neq h$). In particular, this definition
effectively uses the symmetrization $S \cup S^{-1}$, so it produces a standard undirected
Cayley graph even when $S$ is not itself symmetric. -/
```

Especially the "(symmetric)" is confusing since it auto-symmetrizes

FormalConjectures/Wikipedia/DiameterSimpleFiniteGroups.lean
```diff
@@ -0,0 +1,84 @@
...
+def cayleyGraph {G : Type*} [Group G] (S : Set G) : SimpleGraph G :=
+  SimpleGraph.fromRel (fun g h => g⁻¹ * h ∈ S)
+
+/-- The diameter of a finite group $G$, defined as the maximum diameter of the Cayley graphs
+$\Gamma(G, A)$ over all generating sets $A$ of $G$.
+-/
+noncomputable def groupDiam (G : Type*) [Group G] [Fintype G] : ℕ :=
+  sSup { d : ℕ | ∃ S : Set G, Subgroup.closure S = ⊤ ∧ (cayleyGraph S).diam = d }
+
```
mo271, line 60:
```suggestion

/-- For the trivial group (with one element), the group diameter is zero, since
every Cayley graph has only one vertex and hence diameter zero. -/
@[category test, AMS 20]
theorem groupDiam_fin_one : groupDiam (alternatingGroup (Fin 0)) = 0 := by
  unfold groupDiam
  apply Nat.le_zero.mp
  apply csSup_le
  · exact ⟨0, Set.univ, Subgroup.closure_univ,
      SimpleGraph.diam_eq_zero.mpr (Or.inr inferInstance)⟩
  · rintro d ⟨S, _, hd⟩
    exact Nat.le_zero.mpr (hd ▸ SimpleGraph.diam_eq_zero.mpr (Or.inr inferInstance))

```
Perhaps let's add a test statment like this? Perhaps even an more interesting one as well?

mo271:
> I am quite surprised that Mathlib does not have a Cayley Graph definition...

Me too, consider adding it to Mathlib? (Probably best to discuss on Zulip first...)

FormalConjectures/Wikipedia/DiameterSimpleFiniteGroups.lean
franzhusch: Good idea!

franzhusch:
Added some more test statements.

PR #3492, opened 2026-03-09
feat(GreensOpenProblems): 31

### LINKED ISSUE #1587: Green's Open Problems #31
### What is the conjecture
https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.31

Write $F(N)$ for the largest Sidon subset of $[N]$. Improve, at least for infinitely many $N$, the bounds $N^{1/2} + O(1) \leq F(N) \leq N^{1/2} + N^{1/4} + O(1)$.

* ams-05
* ams-11

### Choose either option

jeangud:
# Description
Write $F(N)$ for the largest Sidon subset of $[N]$. Improve, at least for infinitely many $N$, the bounds $N^{1/2} + O(1) \le F(N) \le N^{1/2} + N^{1/4} + O(1)$.

- Closes #1587.
- Leverages the existing implementation of #205.
- Adds the variants discussed by Green in the comments.
- Implemented with the help of Gemini 3.1 Pro: usually to get a first draft going, and then heavily steered and manually adjusted for style and simplicity.

# Testing
```shell
$ lake build FormalConjectures/GreensOpenProblems/31.lean 
Build completed successfully (7976 jobs).
```

### DIFF
--- a/FormalConjectures/GreensOpenProblems/31.lean
+++ b/FormalConjectures/GreensOpenProblems/31.lean
@@ -49,7 +49,7 @@ open Filter
 /-- Let $F(N)$ be the largest Sidon subset of $[N]$. -/
 noncomputable def F (N : ℕ) : ℝ := Finset.maxSidonSubsetCard (Finset.Icc 1 N)
 
-/-- Can we improve the lower bound $N^{1/2} + O(1)$? -/
+/-- Can we improve the lower bound $N^{1/2} + O(1)$, at least for infinitely many $N$? -/
 @[category research open, AMS 5 11]
 theorem green_31.lower :
     let ans := (answer(sorry) : ℕ → ℝ)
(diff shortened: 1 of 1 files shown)


FormalConjectures/GreensOpenProblems/31.lean
```diff
@@ -0,0 +1,133 @@
...
+- [CLZ01] Cohen, G.D., Litsyn, S., & Zémor, G. (2001). Binary B2-Sequences : A New Upper Bound.
+  J. Comb. Theory A, 94, 152-155.
+-/
+
+namespace Green31
+
+open Filter
+
+/-- Let $F(N)$ be the largest Sidon subset of $[N]$. -/
+noncomputable def F (N : ℕ) : ℝ := Erdos30.h N
```
mo271:
```suggestion
/--
Let $F(N)$ be the size of the largest Sidon subset of $\{1, \dots, N\}$.
-/
noncomputable abbrev F (N : ℕ) : ℕ := Finset.maxSidonSubsetCard (Finset.Icc 1 N)
```

Let's just repeat the definition here, using the Erdős 30 import seems like an unnecesasary redirection.

FormalConjectures/GreensOpenProblems/31.lean
```diff
@@ -0,0 +1,133 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+import FormalConjectures.ErdosProblems.«30»
```
mo271:
```suggestion
```
see comment below

FormalConjectures/GreensOpenProblems/31.lean
```diff
@@ -0,0 +1,133 @@
...
+-/
+
+namespace Green31
+
+open Filter
+
+/-- Let $F(N)$ be the largest Sidon subset of $[N]$. -/
+noncomputable def F (N : ℕ) : ℝ := Erdos30.h N
+
+/-- Can we improve the lower bound $N^{1/2} + O(1)$? -/
```
mo271:
```suggestion
/-- Can we improve the lower bound $N^{1/2} + O(1)$, at least
for infinitely many $N$? -/
```
I think this is a useful addition, then usually one might expect `∀ᶠ` here not just `∃ᶠ `. But I guess this sticks closely to the informal text. Potentially it would also be good to add the `∀ᶠ` variants, they will be harder to prove -- but easier to disprove. and in the "at least" it kinda implies both variants, no?
Same for `green_31_upper`

FormalConjectures/GreensOpenProblems/31.lean
jeangud: Done! :white_check_mark:

FormalConjectures/GreensOpenProblems/31.lean
jeangud: Done :white_check_mark:

FormalConjectures/GreensOpenProblems/31.lean
jeangud: @mo271 I agree yes, added the variants :+1:

jeangud:
Thank you for the review @mo271, this should be ready for another look :+1:

PR #2367, opened 2026-02-20
chore(FormalConjectureForMathlib): module-ize

mo271:

PR #3511, opened 2026-03-10
chore: change FormalConjecturesForMathlib to the module system

eric-wieser:
This will prevent us adding new files to ForMathlib that aren't themselves modules.


mo271:
Should have been part of #2367, I suppose

PR #3512, opened 2026-03-10
feat: elaboration options for answer(sorry)

Paul-Lez:
This PR gives users the option to change the behaviour of `answer(sorry)`. They can either
- have it replace `sorry` by `True` if it's a `Prop` (default behaviour for now), and attach metadata.
- do nothing except attach metadata
- create an auxiliary declaration which is then referenced by the `Expr` that the elaborator produces (together with a metadata annotation).

Note: this also changes the metadata attached to the expressions produced by `answer(sorry)` - I think it's reasonable for us to move to a format that is unique to `answer()` rather than using `mkSaveInfoAnnotation`.

### DIFF
--- a/FormalConjectures/Util/Answer.lean
+++ b/FormalConjectures/Util/Answer.lean
@@ -32,7 +32,7 @@ which is a job for human mathematicians, not Lean alone.
 -/
 namespace Google
 
-open Lean Elab Term
+open Lean Elab Meta Term
 
 /-- A type that captures the current setting for the `answer()` elaborator. -/
 inductive AnswerSetting
@@ -40,8 +40,7 @@ inductive AnswerSetting
   | alwaysTrue
   /-- Default mode for `answer(foo)`: just postpones elaboration. -/
   | postpone
-  /-- Elaborate `answer(foo)` by creating an auxiliary definition with value `foo`.
-    **For now, this is experimental** -/
+  /-- Elaborate `answer(foo)` by creating an auxiliary definition with value `foo`. -/
   | withAuxiliary
 deriving Inhabited, ToJson, BEq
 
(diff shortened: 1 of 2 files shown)


FormalConjectures/Util/Answer.lean
```diff
@@ -27,21 +27,89 @@ This file provides syntax for marking up answers in a problem statement.
...
+        type := exprType
+        value := expr
+        hints := .abbrev
+        safety := .safe
+      }
+      addDecl (.defnDecl answerAuxiliaryDecl) true
+      return mkAnswerAnnotation (.const answerName [])
+    | .alwaysTrue =>
+      -- If the answer is a `sorry` of type `Prop` then default to `True` in this setting
+      if expectedType? == some (Expr.sort .zero) && a == (← `(term| sorry)) then
```
eric-wieser, line 103: You might need to instantiate mvars in the expected type

Paul-Lez: (replied on Zulip DM).

Paul-Lez:
I'll wait for Eric to give me the green light and then merge

FormalConjectures/Util/Answer.lean
```diff
@@ -27,21 +27,89 @@ This file provides syntax for marking up answers in a problem statement.
...
+      elabTerm stx expectedType?
+
 /-- Indicates where the answer is in a problem statement. -/
 @[term_elab answer]
 def answerElab : TermElab := fun stx expectedType? => do
   match stx with
-  | `(answer($a:term)) => do
-    mkSaveInfoAnnotation <$> postponeElabTerm (← `(by exact $a)) expectedType?
+  | `(answer($a:term)) =>
+    match (← getOptions).get? `google.answer |>.get! with
```
mo271:
```suggestion
    match (← getOptions).get? `google.answer |>.getD .alwaysTrue with
```
Without that change I get horrible panics when building this locally (not sure why those don't appear in the ci)

Paul-Lez: Hm that's interesting - can I see the file these errors occurred in?

mo271: https://gist.github.com/mo271/f43302057cb54783fb59a43650b872f2

mo271: but it appears in many files. 
One thing that might be different than in the CI is that it builds with 128 threads

FormalConjectures/Util/Answer.lean
```diff
@@ -27,21 +27,89 @@ This file provides syntax for marking up answers in a problem statement.
...
+      elabTerm stx expectedType?
+
 /-- Indicates where the answer is in a problem statement. -/
 @[term_elab answer]
 def answerElab : TermElab := fun stx expectedType? => do
   match stx with
-  | `(answer($a:term)) => do
-    mkSaveInfoAnnotation <$> postponeElabTerm (← `(by exact $a)) expectedType?
+  | `(answer($a:term)) =>
+    match (← getOptions).get? `google.answer |>.getD .alwaysTrue with
```
eric-wieser:
```suggestion
    match google.answer.get (← getOptions) with
```
i think works?

Paul-Lez:
@mo271 The auxiliairy declaration mode is now compatible with universes:)

PR #3519, opened 2026-03-11
feat(GreensOpenProblems): 30

jeangud:
# Description
> Given a set $A \subset \mathbb{Z}$ with $D(A) \leq K$, find a large structured subset $A' \subset A$ which 'obviously' has $D(A') \leq K + \varepsilon$, with $D(A) := |A - A|/|A|$.

- Partially addresses #1565 
- The conjecture is open-ended on multiple fronts:
    - Multiple interpretations of "large structured subset" and "obviously" are possible, as highlighted in #1612 
    - The expected answer to the conjecture needs to be such example subsets, is finding one enough? Do we want construction rules or a full class of compatible subsets?
    - **The full conjecture is left as a TODO**
- As a starting point, I propose formalizing the helping example provided by Green in his comments: 
> If $|A - A| \leq (4 - \varepsilon)|A|$, then there is a progression $P$ of length $\gg_\varepsilon |A|$ on which $A$ has density $> 1/2$. Then $A' := A \cap P$ "obviously" has $D(A') \leq 4$
- First quick draft obtained with the help of Gemini 3.1 Pro, and then heavily rearranged for style and modularity

# Testing
:white_check_mark: Builds successfully.
```shell
$ lake build FormalConjectures/GreensOpenProblems/30.lean 
Build completed successfully (7975 jobs).
```

### DIFF
--- a/FormalConjectures/GreensOpenProblems/30.lean
+++ b/FormalConjectures/GreensOpenProblems/30.lean
@@ -1,0 +1,80 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


mo271:
I think we shouldn't add this file (which looks good so far!) unless we have a `research open` question representing the main question.

PR #3524, opened 2026-03-12
feat(Other): Large Steiner Systems

franzhusch:
Formalizes the question for a construction of a Steiner System under certain constraints, from https://epoch.ai/frontiermath/open-problems/large-steiner-systems. 

Formalization done with the help of Opus 4.6


FormalConjectures/Other/LargeSteinerSystems.lean
```diff
@@ -0,0 +1,90 @@
...
+
+*Reference:* [Large Steiner Systems](https://epoch.ai/frontiermath/open-problems/large-steiner-systems)
+
+Given a set $S$ of size $n$, an $(n, q, r)$-Steiner system is a collection of $q$-element subsets
+(called blocks) of $S$ such that every $r$-element subset of $S$ is contained in exactly one block.
+
+No example of a Steiner system with $r > 5$ is known, despite a theorem (Keevash, 2014) that
+Steiner systems exist for all sufficiently large $n$ satisfying the necessary divisibility
+conditions. The conjecture asks to explicitly construct one with $n > q > r > 5$, $r < 10$,
+and $n < 200$.
```
mo271:
```suggestion

```

FormalConjectures/Other/LargeSteinerSystems.lean
mo271, line 1: let's move to Wikipedia, use that as a main reference
https://en.wikipedia.org/wiki/Steiner_system

and also use their notation (including creating the S(t, k, n) as lean notation), which is 

Do mention this very specific question (with reference to FrontierMath) which is now under large LargeSteinerSystemWitness, with those specific constraints.
But also the known results for t = 5.
and also "also whether infinitely many have t = 4 or 5." (fine to leave those additional questions as TODO if this is too much
Nice to have an example, if it also include the question at least for t = 5 (perhaps mention the Witt design(s)?)

FormalConjectures/Other/LargeSteinerSystems.lean
```diff
@@ -0,0 +1,90 @@
...
+An $(n, q, r)$-Steiner system is a collection of $q$-element subsets (called blocks) of
+$\{0, \ldots, n-1\}$ such that every $r$-element subset is contained in exactly one block.
+-/
+structure SteinerSystem (n q r : ℕ) where
+  /-- The blocks of the Steiner system. -/
+  blocks : Finset (Finset (Fin n))
+  /-- Every block has exactly $q$ elements. -/
+  block_card : ∀ B ∈ blocks, B.card = q
+  /-- Every $r$-element subset is contained in exactly one block. -/
+  cover_unique : ∀ R : Finset (Fin n), R.card = r → (blocks.filter (R ⊆ ·)).card = 1
```
mo271: One could use `∃!` here, but then `native decide` wouldn't work, I think. If that is the case, let's add a comment explaning this design choice ("we avoid using `∃!` in order to stay decidable" or something to that effect)

mo271 (commented):
Thanks, looks pretty good, but I think we should make it a bit more general, also to open it up more towards related problems about Steiner systems. 
in Erdős Problesm 835 some comments mention that, that might potentially be another TODO to related the statement there (but if at all in a differenc pull request)

FormalConjectures/Other/LargeSteinerSystems.lean
```diff
@@ -0,0 +1,90 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Large Steiner Systems
+
+*Reference:* [Large Steiner Systems](https://epoch.ai/frontiermath/open-problems/large-steiner-systems)
```
mo271:
```suggestion
*Reference:* [Large Steiner Systems](https://epoch.ai/frontiermath/open-problems/large-steiner-systems) by Kunal Marwaha
```

FormalConjectures/Other/LargeSteinerSystems.lean
franzhusch: Does it really make sense to add it the t=5 case? Because this is about a witness / construction it would really only test if it was in the training or not? Or if he can just find it on the internet. Thats why I have not included it.

FormalConjectures/Other/LargeSteinerSystems.lean
franzhusch: It would indeed break (native) decide, I have left it as a comment.

joshuasteier:
Hi Franz(@franzhusch), I have computational results directly relevant to `large_steiner_systems`
that would be a natural addition to this file. Using the Kramer-Mesner method, I've
eliminated four candidate transitive automorphism groups for S(6,7,23) — the smallest
admissible case for t=6:

- M₂₃ (order 10,200,960): 3×4 KM matrix, structural obstruction
- AGL(1,23) (order 506): row 75 has zero coverage in usable columns
- D₂₃ (order 46): divisibility-2 obstruction from orbit size ratio
- Z₂₃ ⋊ Z₁₁ (order 253): exhaustive DLX search, 11,047,539 nodes, UNSAT

The Z₂₃ case (4389×10648 KM matrix) is currently running under CP-SAT. These are
partial progress results on `large_steiner_systems` — they eliminate specific
approaches for the smallest open case. I'd like to add them as `research solved`
variants following your file's structure, using your `SteinerSystem` definition
and `S(t, k, n)` notation. Happy to open a follow-up PR once yours merges, or
contribute directly to this branch if you prefer.

franzhusch:
Hi @joshuasteier, feel free to open a Follow-Up PR once this one has merged, where you add your computational results :)

FormalConjectures/Other/LargeSteinerSystems.lean
franzhusch: @mo271 I think this is the only open question at the moment, else we could merge after you have reviewed the new changes.

mo271 (approved):
Thanks, LGTM

PR #3529, opened 2026-03-13
fix(ErdosProblems): 488 

smmercuri:
Seems like we still need to exclude `0` even after the change in set condition. cf: https://github.com/google-deepmind/formal-conjectures/pull/256

### DIFF
--- a/FormalConjectures/ErdosProblems/488.lean
+++ b/FormalConjectures/ErdosProblems/488.lean
@@ -33,7 +33,7 @@ Is it true that, for every $m>n\geq \max(A)$,
 $$\frac{\lvert B\cap [1,m]\rvert }{m}< 2\frac{\lvert B\cap [1,n]\rvert}{n}?$$
 -/
 @[category research open, AMS 5 11]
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/488.lean
```diff
@@ -33,7 +33,9 @@ Is it true that, for every $m>n\geq \max(A)$,
 $$\frac{\lvert B\cap [1,m]\rvert }{m}< 2\frac{\lvert B\cap [1,n]\rvert}{n}?$$
 -/
 @[category research open, AMS 5 11]
-theorem erdos_488 : answer(sorry) ↔ ∀ (A : Finset ℕ), A.Nonempty →
+theorem erdos_488 : answer(False) ↔ ∀ (A : Finset ℕ), A.Nonempty →
```
mo271:
```suggestion
theorem erdos_488 : answer(sorry) ↔ ∀ (A : Finset ℕ), A.Nonempty →
```

PR #3557, opened 2026-03-13
feat: add module docstring to CopyrightLinter

Paul-Lez:
Add a module docstring to `FormalConjectures.Util.Linters.CopyrightLinter`.


Paul-Lez:
> I suppose we could also add a Linter that check that we have precisely 1 module docstring... @Paul-Lez

Done:)

PR #3559, opened 2026-03-13
feat: add module docstring to ProblemImports

Paul-Lez:
Add a module docstring to `FormalConjectures.Util.ProblemImports`.


mo271:
Can't have a modulo docstrings before the `import` commands...

PR #3583, opened 2026-03-17
fix: only replace dots outside guillemets

vEnhance:
ATM, the GitHub URL in pages like https://google-deepmind.github.io/formal-conjectures/theorem/?name=Arxiv.0911.2077.arxiv.id0911_2077.conjecture6_3 is broken. The site script just assumes all the dots need to be replaced in slashes, but in this case the folder is actually called `0911.2077`.

I think this should fix the issue both in the displayed link text and the link target itself. I did split the function into two to be a bit more DRY.

(To be upfront about LLM usage, I asked AI to write the regex, so that's lines 116-118. Everything else by hand. The CLA should be signed under the username `vEnhance`.)

### DIFF
--- a/site/build.js
+++ b/site/build.js
@@ -111,10 +111,17 @@ const GITHUB_BASE = 'https://github.com/google-deepmind/formal-conjectures/blob/
 // ---------------------------------------------------------------------------
 
 /** Convert a module name to a GitHub file URL. */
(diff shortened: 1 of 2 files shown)


mo271 (approved):
Thanks, good catch!

PR #3585, opened 2026-03-17
feat: add verso to website

mo271:


mo271:
This current state of this can be previewed here, I think is it ready to be merged
https://mo271.github.io/formal-conjectures/

Paul-Lez:
I think there's an issue with the spacing in the hover info when browsing conjectures. For example, on the ABC conjecture page I got:
```
ABC.radical (n : ℕ) : ℕThe radical of `n` denoted is the product of the distinct prime factors of `n`.
```

mo271:
good catch, @Paul-Lez there was something missing from the css. Fix looks good locally will take https://github.com/mo271/formal-conjectures/actions/runs/23205194636/job/67438492204 to finish until it becomse visible at https://mo271.github.io/formal-conjectures/

mo271:
@Paul-Lez should be fixed (and also some other things) and the build here is already done: 
https://mo271.github.io/formal-conjectures/
so with Ctrl-Shift-R refreshing the tooltips should render nicely!

lakefile.toml
```diff
@@ -26,6 +26,11 @@ name = "mathlib"
 scope = "leanprover-community"
 rev = "v4.27.0"
 
+[[require]]
+name = "verso"
+git = "https://github.com/leanprover/verso.git"
+rev = "v4.27.0"
+
```
eric-wieser: I think we should have a separate lakefile for the site and verso project? Can we put this in the docbuild folder?

site/build.js
```diff
@@ -291,6 +309,20 @@ function main() {
   const conjectures = rawData.map(processEntry);
   const stats = computeStats(conjectures);
 
+  // Load Verso literate fragments (module docstrings + const links)
+  let versoFragments = { moduleDocs: {}, constLinks: {} };
+  if (fs.existsSync('data/verso-fragments.json')) {
+    versoFragments = JSON.parse(fs.readFileSync('data/verso-fragments.json', 'utf8'));
+    // Strip guillemets from keys so lookups match guillemet-stripped theorem/module names
+    const stripG = s => s.replace(/[«»]/g, '');
```
eric-wieser: What are we looking up here? I know we don't care, but this will presumably cause a collision between `«A.B».C` and `A.«B.C»`, which in Lean are different names.

site/extract_verso_fragments.py
```diff
@@ -0,0 +1,163 @@
...
+        else:
+            depth -= 1
+            if depth == 0:
+                return html[block_start:close_pos + len('</code>')]
+            idx = close_pos + 7
+
+    return None
+
+
+def extract_from_html(html_path, base_dir):
```
eric-wieser, line 27: Would it not make sense to use beautifulsoup here?

lakefile.toml
mo271: I followed the documentation of verso 
https://github.com/leanprover/verso/blob/b565f0b0bcf322cf432e64d6f284fcb02612fd02/README.md?plain=1#L242
It doesn't mention separate lakefile like for docgen4.

I couldn't make the literate facet to work with a separate lakefile. Is there an anywhere, where this is done?
How exactly would the setup be?

site/build.js
mo271: That make sense. Should be fixed it even if it doesn't happen (yet) for any of our names.
Interestingly verso kinda avoids it by injecting the guillemets in the theorem source code view even if they are not written in this way in the source code. We do want to link to the files in github and hence need to do drop the guillemets at some point, but not when looking things up here.

site/extract_verso_fragments.py
mo271: It would make sense, doing this now.
It was actually a strictly positive change:
we now actually pick up 65 more docstrings thanks to more reliable HTML parsing!
There's no remaining regex or string-based HTML parsing anywhere.

mo271:
once 
https://github.com/mo271/formal-conjectures/actions/runs/23219439223/job/67488406106
finishes (assuming the BeatifulSoup installation in the runner script works)
the test site sits at
https://mo271.github.io/formal-conjectures/

lakefile.toml
mo271: asked here [#verso > separate lakefile/Project?](https://leanprover.zulipchat.com/#narrow/channel/576452-verso/topic/separate.20lakefile.2FProject.3F/with/580114888) for some guidance

mo271: I tried again, and I think it should work now with putting verso in the docbuild folder...

mo271:
@eric-wieser you want to take another look here or is this good to go?

mo271:
I assume this is good for now, we can improve it more later...
For instance we might want to style the theorems a bit nicer, like it is done in some other verso genres

PR #3593, opened 2026-03-17
updated to "formally solved" from open and included reference lean 1071

aryanamol10:
Marked as "formally solved" from open inference and included reference from erdosproblems


aryanamol10:
Have updated the docstring to include references from Erdos Problems as well as the author of formalization. Hopefully this can help fix the docstring consistency issue.

FormalConjectures/ErdosProblems/1071.lean
```diff
@@ -49,8 +49,14 @@ theorem erdos_1071.parts.i :
...
-/-- Is there a region $R$ with a maximal set of disjoint unit line segments that is countably infinite? -/
-@[category research open, AMS 52]
+/-- Is there a region $R$ with a maximal set of disjoint unit line segments that is countably infinite?
+Solved affirmatively by [Fo99], who gave an explicit construction.
+
+
+This was formalized in Lean by Alexeev using Aristotle and ChatGPT.
+-/
+@[category research formally solved using lean4 at
+"https://github.com/plby/lean-proofs/blob/main/src/v4.24.0/ErdosProblems/Erdos1071b.lean", AMS 52]
```
danielchin: Taking a look at this proof, it looks like this is actually the proof to `parts.i` and taking a look at the above proof, it looks like it's a proof of `parts.ii`, you may want to swap the two proofs because it looks like the original author (me) incorrectly added the wrong link.

smmercuri:
Thanks! Note that our current formalisation is incorrect and is fixed in #3630 but your link still proves the corrected version.

smmercuri (approved):

PR #3594, opened 2026-03-17
updated to "formally solved" from solved lean 1102

aryanamol10:
Changed "solved" addressing to "formally solved" from erdosproblems reference


Osalotioman (changes requested):
You should use the `formal_proof` attribute as specified in [contribute.html](https://github.com/google-deepmind/formal-conjectures/blob/main/site/src/templates/contribute.html), there is no `formally solved` category.

Smetalo:
Is this PR obsolete, since this is resolved in #3920?

PR #3598, opened 2026-03-17
feat(scripts/extract_names): more fine grained output control

mo271:


mo271:
This will make it easier to create a .json for the website (and other use cases) with only the stuff we really need.

scripts/extract_names.lean
```diff
@@ -68,6 +72,21 @@ structure TheoremInfo where
   formalProofLink : Option String
   deriving ToJson
 
+/-- Serialize `TheoremInfo` to JSON, omitting fields whose keys are in `exclude`. -/
+def TheoremInfo.toFilteredJson (info : TheoremInfo) (exclude : Std.HashSet String) : Json :=
```
Paul-Lez: nit: Should this just be _the_ `ToJson` instance on `TheoremInfo` (rather than deriving the instance when defining the structure, as is currently the case)?

mo271: done (needed a little extra work, to make it fly with the `exclude` arg)

PR #3599, opened 2026-03-18
fix(ErdosProblems): 277 

smmercuri:
- Previous formulation of `erdos_277` was missing the (often implicit) assumption that the moduli must all be distinct, as given in the strictly monotone condition in the [solution paper](https://www.cambridge.org/core/journals/mathematika/article/abs/covering-systems-of-congruences-a-negative-result/1428418132CBD5B6E87BAFE8A5B42640)
- Use `StrictCoveringSystem` from `ForMathlib` to cover this missing assumption
- `CoveringSystem` in `ForMathlib` itself was missing an additional non-triviality condition, that none of the moduli must be `1` (otherwise one gets the original ring as a trivial covering system)
- push negation through `erdos_277` to make it slightly nicer

### DIFF
--- a/FormalConjectures/ErdosProblems/277.lean
+++ b/FormalConjectures/ErdosProblems/277.lean
@@ -25,7 +25,7 @@ import FormalConjectures.Util.ProblemImports
   53--61.
 -/
 
-open ArithmeticFunction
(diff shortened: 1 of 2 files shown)


mo271 (approved):
Thanks!

PR #3605, opened 2026-03-20
Mark erdos_686.variants.four_three as solved

mzhorvath1:
Updates the category attribute from open to solved. 

Based on recent discussion N=4 is not representable for k=3.

Reference: https://www.erdosproblems.com/forum/thread/686#post-4599.

### DIFF
--- a/FormalConjectures/ErdosProblems/686.lean
+++ b/FormalConjectures/ErdosProblems/686.lean
@@ -80,7 +80,7 @@ The number $4$ cannot be written as
 $$4=\frac{\prod_{1\leq i\leq 2}(m+i)}{\prod_{1\leq i\leq 2}(n+i)}$$
 for $m≥n+2$!
 -/
-@[category research open, AMS 11]
+@[category research solved, AMS 11]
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/686.lean
```diff
@@ -80,7 +80,7 @@ The number $4$ cannot be written as
 $$4=\frac{\prod_{1\leq i\leq 2}(m+i)}{\prod_{1\leq i\leq 2}(n+i)}$$
 for $m≥n+2$!
```
mo271, line 81:
```suggestion
for $m≥n+2$!

See [comment section on erdosproblems.com](https://www.erdosproblems.com/forum/thread/686#post-4599)
```
Perhaps it is good to give some more concrete pointer?

mo271 (approved):

FormalConjectures/ErdosProblems/686.lean
mzhorvath1: Sounds good!

mo271 (commented):
Thanks!

PR #3609, opened 2026-03-21
feat(Books/UniformDistribution): add one to Equidistribution.lean

rwst:
This would help in the decidability question of some halting problems ;-)


mo271 (approved):
Thanks!

PR #2385, opened 2026-02-23
feat(Papers): Add Monochromatic Quantum Graph conjectures

MarioKrenn6240:
### Summary

This PR adds a formalization of the *monochromatic quantum graph* / *perfect-matching equation system* problem (see #2205) in the `formal-conjectures` style. The problem is motivated by quantum-optics constructions of high-dimensional multipartite GHZ-type states via linear optics, where interference amplitudes can be expressed as weighted sums over perfect matchings.

The setup uses a complete graph `K_N` whose edges carry both (i) endpoint “color indices” in `Fin D` and (ii) a coefficient/weight. For each vertex assignment `ι : V N → Fin D`, we define a perfect-matching sum `pmSumN N D W ι` (sum over perfect matchings, each contributing the product of its induced edge weights). The equation system `EqSystemN N D W` enforces:

* `pmSumN ... ι = 1` iff `ι` is constant (monochromatic inherited vertex coloring),
* `pmSumN ... ι = 0` otherwise.

This yields exponentially many constraints in `N` on only `O(N^2)` edge variables, matching the combinatorial formulation in the MathOverflow reference above.

---

### What this PR contains

* Core definitions:

  * edge/weight model (`EdgeN`, `WeightsN`)
  * perfect-matching sum (`pmSumN`)
  * the equation system (`EqSystemN`)
* Verified “sanity check” solutions (explicit witnesses):

  * `N = 4, D = 2`
  * `N = 4, D = 3`
  * `N = 6, D = 2`
* Open conjectures stating non-existence over several coefficient domains:

  * `ℂ`, `ℝ`, `ℤ`, and integer weights restricted to `{ -1, 0, 1 }`
* Recorded non-existence results over `ℝ≥0` attributed to Bogdanov’s lemma (as `research solved`, proof omitted)

All general statements are restricted to **even `N`**, since odd `N` admits no perfect matchings.

---

### Selected references:

* PRL (2017): [https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.119.240403](https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.119.240403)
* PNAS (2019): [https://www.pnas.org/doi/10.1073/pnas.1816096116](https://www.pnas.org/doi/10.1073/pnas.1816096116)
[...]

PR #3611, opened 2026-03-22
Add more monochromatic quantum graph special cases

MarioKrenn6240:
Follow-up to #2385.

This PR adds the following special cases to `MonochromaticQuantumGraph.lean`:

- `(n = 8, d = 3)`, `(n = 8, d = 10)`, `(n = 10, d = 3)`, `(n = 10, d = 10)` for `ℂ`, `ℝ`, `ℤ`, and `{-1, 0, 1}`
- `(n = 12, d = 3)`, `(n = 14, d = 3)`, `(n = 16, d = 3)` for `ℂ` only

These are special cases for which non-existence might be easy to be proven (first row) or where counter-examples might exist (second row).


MarioKrenn6240:
unintentionally added to the previous PR, will close and make a new one.

PR #3614, opened 2026-03-23
Fix tag for erdos_949.variants.sidon

mzhorvath1:
Updates tag to use the new feature of marking a research solved problem as solved using formal-conjectures.

### DIFF
--- a/FormalConjectures/ErdosProblems/949.lean
+++ b/FormalConjectures/ErdosProblems/949.lean
@@ -40,7 +40,7 @@ theorem erdos_949 : answer(sorry) ↔
 
 /-- Let $S\sub \mathbb{R}$ be a Sidon set. Must there be a set $A\sub \mathbb{R}∖S$ of cardinality
 continuum such that $A + A \sub \mathbb{R}∖S$? -/
-@[category research formally solved using formal_conjectures at "https://github.com/google-deepmind/formal-conjectures/pull/1509", AMS 5]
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/949.lean
```diff
@@ -40,7 +40,7 @@ theorem erdos_949 : answer(sorry) ↔
 
 /-- Let $S\sub \mathbb{R}$ be a Sidon set. Must there be a set $A\sub \mathbb{R}∖S$ of cardinality
 continuum such that $A + A \sub \mathbb{R}∖S$? -/
-@[category research solved, AMS 5]
+@[category research formally solved using formal_conjectures at "https://github.com/google-deepmind/formal-conjectures/pull/1509", AMS 5]
```
Paul-Lez:
```suggestion
@[category research formally solved using formal_conjectures at "", AMS 5]
```

Paul-Lez (approved):
Since we've already got this one in the repo, I'd suggest just using `""` for the "link" (I'm going to change the behaviour of the formally solved attribute anyway so that there a `here` mode for proofs that we've merged)

FormalConjectures/ErdosProblems/949.lean
mzhorvath1: Sounds good! Updated according to your suggestion.

PR #3621, opened 2026-03-24
fix(GreenOpenProblems): 2 

smmercuri:
- Existential quantifiers should be used instead of universal in some of the variants. In particular Rusza's example is existential (see [Theorem 1.1](https://arxiv.org/pdf/1804.03356)) as is Erdős' and Choi's (see [text above Theorem 1.1](https://arxiv.org/pdf/1804.03356) or Choi's [original paper](https://londmathsoc.onlinelibrary.wiley.com/doi/abs/10.1112/plms/s3-23.4.629) which involves a construction of an example)
- Some of the variants are missing the asymptotic bound in `n` (e.g., this appears in Choi's paper)


mo271 (approved):
Thanks!

jeangud:
Thank you @smmercuri for catching those and for the fixes!

PR #3629, opened 2026-03-25
fix(ErdosProblems): 197

smmercuri:
- The previous definition of `HasMonotoneAP` was correct but had a slightly unnatural ordering of types in the formalisation, with `f` being a map from the value type of the sequence to the index type.
- This led to a misformalisation in `erdos_197` where the permutations are maps from index type to value type.
- Rather than use `f.symm` in `erdos_197`, reverse ordering within `HasMonotoneAP` so that it is more natural.

### DIFF
--- a/FormalConjectures/ErdosProblems/197.lean
+++ b/FormalConjectures/ErdosProblems/197.lean
@@ -37,6 +37,4 @@ theorem erdos_197 :
       (∃ g : ℕ ≃ B, ¬HasMonotoneAP (Subtype.val ∘ g) 3) := by
   sorry
 
-
-
 end Erdos197
--- a/FormalConjecturesForMathlib/Combinatorics/AP/Basic.lean
+++ b/FormalConjecturesForMathlib/Combinatorics/AP/Basic.lean
@@ -241,8 +241,9 @@ def ContainsMonoAPofLength {κ : Type} [Finite κ] {M : Set α}
     ∀ m ∈ ap, coloring m = c
 
 /--
-A function `f : α → β` has a monotone `k`-term arithmetic progression if there exists an
-arithmetic progression `l` of length `k` in `α` such that its image under `f` is sorted.
+A function `f : β → α` has a monotone `k`-term arithmetic progression if there exists a choice
+of indices `b 1 < b 2 < ... < b k` such that the subsequence `f (b i)` forms an increasing or
+decreasing arithmetic progression of length `k`.
 -/
 def HasMonotoneAP {β : Type*} [Preorder β] (f : β → α) (k : ℕ) : Prop :=
   ∃ l : List β, (l.map f).IsAPOfLength k ∧ l.Pairwise (· < ·)


FormalConjectures/ErdosProblems/197.lean
```diff
@@ -33,8 +33,10 @@ Can $\mathbb{N}$ be partitioned into two sets, each of which can be permuted to
...
 theorem erdos_197 :
     answer(sorry) ↔ ∃ A B : Set ℕ, IsCompl A B ∧
-      (∃ f : ℕ ≃ A, ¬ HasMonotoneAP f 3) ∧
-      (∃ g : ℕ ≃ B, ¬ HasMonotoneAP g 3) := by
+      (∃ f : ℕ ≃ A, ¬HasMonotoneAP (Subtype.val ∘ f) 3) ∧
+      (∃ g : ℕ ≃ B, ¬HasMonotoneAP (Subtype.val ∘ g) 3) := by
   sorry
 
+
+
```
Paul-Lez:
```suggestion

```

FormalConjecturesForMathlib/Combinatorics/AP/Basic.lean
```diff
@@ -244,8 +244,8 @@ def ContainsMonoAPofLength {κ : Type} [Finite κ] {M : Set α}
 A function `f : α → β` has a monotone `k`-term arithmetic progression if there exists an
 arithmetic progression `l` of length `k` in `α` such that its image under `f` is sorted.
```
Paul-Lez:
```suggestion
A function `f : β → α` has a monotone `k`-term arithmetic progression if there exists an
sorted list in `β` whose image under `f` is an arithmetic progression `l` of length `k`.
```

FormalConjecturesForMathlib/Combinatorics/AP/Basic.lean
```diff
@@ -244,8 +244,8 @@ def ContainsMonoAPofLength {κ : Type} [Finite κ] {M : Set α}
...
 -/
-def HasMonotoneAP {β : Type*} [Preorder β] (f : α → β) (k : ℕ) : Prop :=
-  ∃ l : List α, l.IsAPOfLength k ∧ (l.map f).Pairwise (· < ·)
+def HasMonotoneAP {β : Type*} [Preorder β] (f : β → α) (k : ℕ) : Prop :=
```
Paul-Lez, line 248: Maybe I'm missing something, but don't we want `f` to be order preserving on that list?

smmercuri: The definition from [ErGr90] is "Let us say that a sequence $(a_1,a_2,…)$ has a monotone AP of length k if for some choice of indices $i_1<i_2<⋯<i_k$, the subsequence $a_{i_1},a_{i_2},…,a_{i_k}$ is either an increasing or decreasing AP. So we want the image of `f` to be either order preserving or anti order preserving. This condition is asserted inside `List.IsAPOfLength k`.

FormalConjectures/ErdosProblems/197.lean
```diff
@@ -33,8 +33,8 @@ Can $\mathbb{N}$ be partitioned into two sets, each of which can be permuted to
 @[category research open, AMS 5]
 theorem erdos_197 :
     answer(sorry) ↔ ∃ A B : Set ℕ, IsCompl A B ∧
-      (∃ f : ℕ ≃ A, ¬ HasMonotoneAP f 3) ∧
-      (∃ g : ℕ ≃ B, ¬ HasMonotoneAP g 3) := by
+      (∃ f : ℕ ≃ A, ¬HasMonotoneAP (Subtype.val ∘ f) 3) ∧
+      (∃ g : ℕ ≃ B, ¬HasMonotoneAP (Subtype.val ∘ g) 3) := by
```
mo271, line 37: I'd prefer to switch around the: A ≃ ℕ instead of the ugley Subtype.val, but either one is fine with me...

mo271 (approved):

FormalConjectures/ErdosProblems/197.lean
smmercuri: I think I'll keep the `Subtype.val`. It's ugly but it means this statement and `HasMonotoneAP` are a bit more readable with respect to the informal statement imo!

PR #3631, opened 2026-03-25
fix(ErdosProblems): 887 

smmercuri:
- `n` should be positive otherwise we are taking `Real.sqrt` of a negative values which junks to `0`. 
- Note also that the original problem assumes positivity (see e.g., this [arXiv paper](https://arxiv.org/pdf/1406.2230))
- About the variants on the infinitude of numbers with 4 divisors close to `sqrt n`:
  - There is a missing constant in the variants on the infinitude of numbers with 4 divisors (this is taken e.g, 2 in [the proof](http://matwbn.icm.edu.pl/ksiazki/aa/aa79/aa7944.pdf))
  - We should have `le 4` since the paper constructs a sequence with at least 4 such divisors.

### DIFF
--- a/FormalConjectures/ErdosProblems/887.lean
+++ b/FormalConjectures/ErdosProblems/887.lean
@@ -21,9 +21,10 @@ open Filter Finset Real
 /-!
 # Erdős Problem 887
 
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #3632, opened 2026-03-25
Prove 4 Sidon set theorems (Erdős 42, 44)

XC0R:
## Summary
- Prove `example_maximal_sidon`: `{1, 2, 4}` is maximal Sidon in `{1, ..., 4}` via case analysis
- Prove `example_difference_set`: difference set of `{1, 2, 4}` is `{0, 1, 2, 3}` via Set.sub extensionality
- Prove `maximal_sidon_contains_zero`: any maximal Sidon set has `0 ∈ A - A` via nonemptiness
- Prove `example_sidon_set` (44): `{1, 2, 4, 8, 13}` is Sidon via exhaustive case bash + omega

## Notes
- AI-assisted (Claude for tactic search), manually reviewed and submitted
- Techniques: case analysis on finite set membership, omega for arithmetic


mo271 (approved):
Thanks, looks good, 

Some proofs could likely be a bit simplified, but I guess we can also do that later or if they break during a version bump...

PR #3640, opened 2026-03-26
fix(OEIS): A006697 fix gen func and provide solution

vEnhance:
-  The Allouche-Shallit paper cited in the OEIS entry also provides a formula for a(n) given by $a_n = \sum_{i=0}^{n} \min(2^i,n-i+1)$. If one uses this instead as the definition of a(n), the conjecture becomes straightforward to prove.
- The OGF actually listed on the OEIS does not appear to match the sequence; it is off by a small normalization.  Indeed, letting X=0 we see the original OGF has value 3, but a(0)=1. This change corrects the listed OGF to match what I think is the correct generating function.

### DIFF
--- a/FormalConjectures/OEIS/6697.lean
+++ b/FormalConjectures/OEIS/6697.lean
@@ -117,7 +117,7 @@ for a formal proof of the generating function using this definition.
 Hence, a formalization of [arXiv:1605.02361](https://arxiv.org/abs/1605.02361)
 would complete a formal proof as below.
 -/
-@[category research solved, AMS 68]
+@[category research solved, formal_proof using lean4 at "https://github.com/AxiomMath/gdm-formal-conjectures/blob/main/OeisA6697/solution.lean", AMS 68]
 theorem conjecture (n : ℕ) :
     a n = coeff (R := ℚ) n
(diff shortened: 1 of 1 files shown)


FormalConjectures/OEIS/6697.lean
```diff
@@ -98,11 +98,27 @@ $$\sum_{n \geq 0} a_n x^n = 1 + \frac{1}{1-x} + \frac{1}{(1-x)^2}\left(\frac{1}{
   \sum_{k \geq 0} x^{2^{k+1} + k}\right).$$
 
 Equivalently, a(n) equals the n-th coefficient of this generating function.
+
+However, the reference quoted in the OEIS sequence
+
+> J.-P. Allouche and J. Shallit, "On the subword complexity of the fixed point of a → aab, b → b,
```
seewoo5, line 106: Maybe you want to add this to the `*References:*` above.

seewoo5: Nvm it is already there - I mean to add separate statement for the formula for a_n.

vEnhance:
I added the formula into the docstring at the top of the file, and also added a `formal_proof using lean4` pointer that was recently introduced in #3645 .

PR #3641, opened 2026-03-26
feat(Books): borwein_sine_series solved in 2005

vEnhance:
The conjecture is currently marked as open, but a complete proof was posted by Ravi Boppana in https://arxiv.org/abs/2007.11017, citing a discussion on AoPS in 2005.

We also link a partial formalization of a proof along these lines (taking irrationality measure of pi as given).


franzhusch:
Ah thanks for the report, might be good to also feedback that to mathworld, as I have taken them as a source for declaring the problem as open. Here is the link: https://mathworld.wolfram.com/HarmonicSeries.html

vEnhance:
@franzhusch Thanks for the tip on MathWorld. I sent the MathWorld Team a short note.

@Paul-Lez added `formal_proof using lean4` in to the metadata! (Just a note that the proof we posted does rely on an external citation that we include as a `Prop` hypothesis into the theorem statement.)

Paul-Lez:
> @Paul-Lez added `formal_proof using lean4` in to the metadata! (Just a note that the proof we posted does rely on an external citation that we include as a `Prop` hypothesis into the theorem statement.)

Good to know! In that case, would you mind adding a comment in the dosctring mentioning this for future reference?

vEnhance:
@Paul-Lez Should be such a note already:

```
Also see
https://github.com/AxiomMath/gdm-formal-conjectures/blob/main/docs/BorweinSineSeries.md
for a partial formalization of the conjecture,
conditional on such an irrationality measure of pi (cf https://arxiv.org/abs/1912.06345).
```

PR #3650, opened 2026-03-27
feat(Util/Attributes): Make things clearer for formal proof already in formal conjectures

Paul-Lez:


mo271 (approved):
Thanks!

mo271 (changes requested):
actually , we might not want to allow that:

When there is a proof, we don't really need a a label, because we can just see that the proof is there, no?

FormalConjectures/ErdosProblems/949.lean
```diff
@@ -40,7 +40,7 @@ theorem erdos_949 : answer(sorry) ↔
 
 /-- Let $S\sub \mathbb{R}$ be a Sidon set. Must there be a set $A\sub \mathbb{R}∖S$ of cardinality
 continuum such that $A + A \sub \mathbb{R}∖S$? -/
-@[category research solved, AMS 5, formal_proof using formal_conjectures at ""]
+@[category research solved, AMS 5, formal_proof using formal_conjectures]
```
mo271, line 43:
```suggestion
@[category research solved, AMS 5]
```

FormalConjectures/Wikipedia/InvariantSubspaceProblem.lean
```diff
@@ -66,7 +66,7 @@ lemma TopologicalSpace.nontrivial_of_not_separableSpace {H : Type*} [Topological
 Every bounded linear operator `T : H → H` on a non-separable Hilbert space `H` has a
 non-trivial closed `T`-invariant subspace. Such an invariant space is given by considering the
 closure of the linear span of the orbit of any single non-zero vector. -/
-@[category research solved, AMS 47, formal_proof using formal_conjectures at ""]
+@[category research solved, AMS 47, formal_proof using formal_conjectures]
```
mo271, line 69:
```suggestion
@[category research solved, AMS 47]
```

FormalConjecturesTest/Util/Attributes/Basic.lean
```diff
@@ -50,7 +50,7 @@ theorem a_graduate_problem_with_formal_proof : 1 + 1 = 2 := by
   rfl
 
 #guard_msgs in
-@[category test, formal_proof using formal_conjectures at ""]
+@[category test, formal_proof using formal_conjectures]
```
mo271, line 53:
```suggestion
@[category test]
```

mo271 (commented):
so just removing the `at ""` tags might be nicer?

FormalConjectures/ErdosProblems/949.lean
smmercuri: Are you suggesting this tag should then just be `@[category research solved, AMS 5]`? I guess one reason for keeping the `formal_proof` tag is that it contributes to the stats on the website

mo271: yeah, that is what I meant. 

We should adapt what the website counts (if it not already does that): if there is a sorry free proof (which our extract_names script for the website already detects, I believe), we should add it towards that count

mo271:
@Paul-Lez I think we don't need anythiing here anymore? For proofs in the repo, we don't want to label them, since the proof is right there, and it will just add superflous boilerplate. Perhaps in this pr we could just remove the unnessary usages of this that are left (already as pending changes)

PR #3438, opened 2026-03-06
feat(ErdosProblems/342)

aditya-ramabadran:
# Description

> With $a_1=1 $ and $a_2=2$  let $a_{n+1} $ for $n\ge 2 $ be the least integer greater than $a_n $ which can be expressed uniquely as $a_i + a_j $for $i < j <= n$.  Do infinitely many pairs $(a,a+2)$  occur in the sequence?  Does Ulam's sequence eventually have periodic differences?  Is the density of the sequence 0?

- Closes google-deepmind#2120
- Quantifies over all u with `IsUlamSequence u`, I think this is cleaner than defining a specific Ulam Sequence 

# Testing
- Compiles successfully
```shell
$ lake build FormalConjectures/ErdosProblems/342.lean
Build completed successfully.
```

Issue #1546, opened 2026-01-08
Green's Open Problems #7


rao107:
### What is the conjecture
https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#section.1 Problem 7

Define Ulam’s sequence 1, 2, 3, 4, 6, 8, 11, 13, 16, 18, 26, 28, 36, . . . as follows: $u_1 = 1$, $u_2 = 2$, and $u_{n+1}$ is the smallest number expressible as a sum $u_i + u_j$, $i < j ⩽ n$, in a unique way. Does this sequence have positive density? Can one explain the curious Fourier properties of this sequence?

See https://oeis.org/A002858 for Ulam's sequence.

See https://www.erdosproblems.com/342 for more questions about the same sequence.

Discussion under the problem mentions https://www.erdosproblems.com/341.

### Choose either option
- [ ] I plan on adding this conjecture to the repository
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

aditya-ramabadran:
Can do the first question today via #3438, second part will need to look into more to see if there are any specific formalizable claims about the "curious Fourier properties"

aditya-ramabadran:
I formalized the first part (positive density) at #3661 

The second part "Can one explain the curious Fourier properties of this sequence" is more arbitrary. It's unclear what to formalize, or if we would want to formalize this at all. 

https://arxiv.org/abs/1507.00267 - This paper says the following:
> We report the empirical discovery of a surprising global rigidity phenomenon: there seems to exist a real $α \sim 2.5714474995…$ such that $\set{\alpha a_n : n \in \mathbb N } \pmod{2\pi}$ generates an absolutely continuous non-uniform measure supported on a subset of $\mathbb T$. Indeed, for the first $10^7$ elements of Ulam's sequence, cos(2.5714474995 $a_n$)<0 for all $a_n \not\in$ {2,3,47,69}.

Some ideas on what exactly we could formalize, if we choose to formalize any of this:
* `∃ α, the sequence (α * a_n mod 2π) is not equidistributed on the circle.` This sort of captures the main idea, without hard coding that decimal near 2.57, which I don't think we should do. More difficult to formalize though
* We could also formalize something simpler like `∃ α, limsup_N (1/N) ∑_{n<N} cos(α a_n) < 0` which captures the cosine bias and is easy to do
* I think formalizing something to do with the specific decimal is weird and we probably shouldn't do that

Open to thoughts on this

PR #3661, opened 2026-03-29
feat(GreenOpenProblems): 7, positive density part

aditya-ramabadran:
# Description
> Does Ulam's sequence have positive density? Can one explain the curious Fourier properties of Ulam's sequence?

I've formalized the first part, basically borrowing what I did in #3438. I discuss the second part more in the issue #1546. 

# Testing
Compiles successfully.


mo271 (approved):
Thanks, LGTM!

PR #2404, opened 2026-02-25
feat(ErdosProblems/125): add two variants to erdos_125

Paul-Lez:
Fixes #2394

PR #3665, opened 2026-03-30
feat(ErdosProblems/125): additional variants

mo271:


mo271:
fixes follow-up mentioned in #2404

PR #3670, opened 2026-03-31
feat(ErdosProblems/978): solution for `parts.ii`

mo271:


Paul-Lez (commented):
Very cool! Do you want to merge this? (in which case maybe it would be worth golfing it a little?)

mo271:
> Very cool! Do you want to merge this? (in which case maybe it would be worth golfing it a little?)

I was planning on merging this one, since it seems like it can be golfed to a reasonable length. 
Note how the original proof didn't use the eisenstein criterion api directly...
Please do go ahead suggesting improvements!

We plan to post to the comment section of the erdosproblems.com soon, and perhaps there is yet something else wrong with the (informal) formulation of the problem: it has been changed before....

mo271:
decided against including the proof now: the problem has been changed on erdosproblems.com

Paul-Lez (approved):

PR #3674, opened 2026-03-31
fix(GreenOpenProblems): 72 

smmercuri:
Fix argument order in `AllowedSetSize`

### DIFF
--- a/FormalConjectures/GreensOpenProblems/72.lean
+++ b/FormalConjectures/GreensOpenProblems/72.lean
@@ -72,7 +72,7 @@ alias no_three_in_line := green_72
 
 /-- Does the no-three-in-line problem hold when $N$ is big enough? -/
 @[category research open, AMS 05 52]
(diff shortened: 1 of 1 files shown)


FormalConjectures/GreensOpenProblems/72.lean
```diff
@@ -72,7 +72,7 @@ alias no_three_in_line := green_72
 
 /-- Does the no-three-in-line problem hold when $N$ is big enough? -/
 @[category research open, AMS 05 52]
-theorem green_72.variants.eventually : ∀ᶠ N in Filter.atTop, NoKInLineFor 3 N := by
+theorem green_72.variants.eventually : ¬∀ᶠ N in Filter.atTop, NoKInLineFor 3 N := by
```
mo271: should this be using `answer(sorry)` since it is a question?

smmercuri: Yes, and I need to stop forgetting to remove the negation!

mo271 (approved):
Thanks!

PR #3676, opened 2026-03-31
feat(WrittenOnTheWallII/Test): add tests for `l` and `Ls`

mo271:


FormalConjecturesForMathlib/Combinatorics/SimpleGraph/GraphConjectures/Definitions.lean
```diff
@@ -81,10 +81,26 @@ Returned as a real number. -/
 noncomputable def b (G : SimpleGraph α) : ℝ :=
   (largestInducedBipartiteSubgraphSize G : ℝ)
 
+def is_indep_set {β : Type*} [Fintype β] [DecidableEq β] (G : SimpleGraph β) [DecidableRel G.Adj] (s : Finset β) : Bool :=
```
YaelDillies:
```suggestion
def isindepSet {β : Type*} [Fintype β] [DecidableEq β] (G : SimpleGraph β) [DecidableRel G.Adj] (s : Finset β) : Bool :=
```
and same everywhere else

mo271: done

YaelDillies (commented):
Although your code manifestly works, I am uneasy with how much duplication having a computable copy of each quantity in sight is. In particular, all these definitions will never make it to mathlib.

Have you considered using `cbv` instead? Here is a toy implementation for Bernoulli numbers: https://leanprover.zulipchat.com/#narrow/channel/116395-maths/topic/Fast.20computation.20of.20Bernoulli.20numbers/near/590487486

mo271:
@Danie-I if you feel like it you can take over this pr. I just wrote the computable implementations to get a sanity check for `conjecture 2`. Yaël's suggestions sound reasonable.

PR #3684, opened 2026-04-01
fix(HasMonotoneAP): values should form AP, not positions

mo271:
The original definition required positions to form an AP with monotone values. The correct mathematical concept (Erdős-Graham 1980) requires ordered positions with values forming an AP. This fixes Erdős 197. The other problems using the definition were not affected because source and target were identical.


smmercuri:
Does this duplicate #3629?

mo271:
yes, it does...

PR #3689, opened 2026-04-03
feat(ErdosProblems/890): solution for `parts.a`

mo271:


mo271:
in the meantime the problem statement was changed on erdosproblems.com, so this is not more fixing a misformalistaion

Paul-Lez (approved):

PR #3693, opened 2026-04-04
Add special cases to MonochromaticQuantumGraph

MarioKrenn6240:
Added special cases for N=10, D=4...9 and N=6, D=4,5,6.


mo271 (approved):
Thanks, LGTM!

(Probably fine to only add the complex version here for now, especially in light of the recent discoveries -- there was not a single case were the obstructions to solutions were found only in a not-complex case...)

PR #3698, opened 2026-04-06
feat(ErdosProblems/26): solution for `variants.tenenbaum`

mo271:


FormalConjectures/ErdosProblems/26.lean
```diff
@@ -104,13 +105,18 @@ theorem erdos_26.variants.rusza : ∃ A : ℕ → ℕ,
     StrictMono A ∧ ¬IsThick A ∧ ∀ k, ¬IsBehrend (A · + k) := by
   sorry
 
+
+
```
mo271:
```suggestion
```

Paul-Lez (approved):

PR #3706, opened 2026-04-08
feat(Mathoverflow/10799): Kahn-Kalai conjecture

mo271:


FormalConjectures/Mathoverflow/10799.lean
```diff
@@ -0,0 +1,218 @@
...
+  ∑ S ∈ F, μ p S
+
+/--
+Given a family $F$, for a subset $S$ of $X$, we write $h(S)$ as the number of subsets $T$ in $X$
+such that
+(1) $T$ differs from $S$ in exactly one element
+(2) Exactly one set among $S$ and $T$ belongs to $F$.
+-/
+def boundaryCount (n : ℕ) (F : Finset (Finset (Fin n))) (S : Finset (Fin n)) : ℕ :=
+  (Finset.univ.filter fun i : Fin n ↦ Xor' (S ∈ F) (symmDiff S {i} ∈ F)).card
```
GeorgeTsoukalas, line 72: This definition has a logical step over the informal definition. It is not "immediate" from the definition that the set of such T is exactly those S +- {i} for I \in Fin n. Is this problematic? I would say it makes it slightly more obfuscated to check the alignment

FormalConjectures/Mathoverflow/10799.lean
```diff
@@ -0,0 +1,218 @@
...
+-/
+@[category research solved, AMS 5 60,
+formal_proof using formal_conjectures at
+"https://github.com/mo271/formal-conjectures/blob/408f53dc0856c0882a5e77acd24fc83b978f0bc9/FormalConjectures/Mathoverflow/10799.lean#L252"]
+theorem mathoverflow_10799 : answer(False) ↔
+    ∀ (n : ℕ) (_ : 2 ≤ n)
+    (F : Finset (Finset (Fin n))) (_ : IsMonotoneIncreasing F)
+    (s t : ℝ) (_ : 0 < s) (_ : s ≤ t) (_ : t < 1)
+    (_ : t / s > 1000 * Real.log n),
+    ∃ p, s ≤ p ∧ p ≤ t ∧ IsOptimal p F := by
```
GeorgeTsoukalas, line 136: p \in [0,1] explicitly, or do you think it is fine? (I can see that it follows)

FormalConjectures/Mathoverflow/10799.lean
```diff
@@ -0,0 +1,218 @@
...
+/--
+Start with a set $X=\{1,2,...,n\}$ of $n$ elements and the family $2^X$ of all subsets of $X$.
+For a real number $p$ between zero and one, we consider a probability distribution $\mu_p$ on
+$2^X$ where the probability that $i \in S$ is $p$, independently for different $i$'s.
+Thus for $p=1/2$ we get the uniform probability distribution.
+
+For $S \subseteq \{0, \ldots, n-1\}$, its probability is $p^{|S|} (1-p)^{n - |S|}$.
+-/
+noncomputable def μ {n : ℕ} (p : ℝ) (S : Finset (Fin n)) : ℝ :=
+  p ^ #S * (1 - p) ^ (n - #S)
```
GeorgeTsoukalas, line 47: Q: Should we stipulate p \in Icc 0 (1 : \R)?

FormalConjectures/Mathoverflow/10799.lean
```diff
@@ -0,0 +1,218 @@
...
+(IR) $I^p(F) \ge (1/p) \mu_p(F) \cdot \log \mu_p(F)$
+This relation is true for every family $F$ and every $p$. It is especially famous and simple when
+$p=1/2$ and $\mu_p(F)=1/2$. In this case, it says that given a set of half the vertices of the
+discrete cube $2^X$, the number of edges between $F$ and its complement is at least $2^{n-1}$.
+-/
+@[category graduate, AMS 5 60]
+theorem discrete_isoperimetric_inequality (n : ℕ) (p : ℝ) (hp : 0 < p) (hp' : p < 1)
+    (F : Finset (Finset (Fin n))) :
+    let m := μFamily p F
+    edgeBoundary n p F ≥ m * Real.log (1 / m) / (p * Real.log (1 / p)) := by
```
GeorgeTsoukalas: I don't follow how this is equivalent to the isoperimetric relation, which doesn't have a log (1/p) term (at least from Gil's blog post?) Depending on if this is correct, then Is optimal should be fine, but otherwise that should change too

FormalConjectures/Mathoverflow/10799.lean
mo271: I'd say this is fine, we have `(_ : 0 < s) (_ : s ≤ t)` just above and it makes the statement harder to read if we add it.

FormalConjectures/Mathoverflow/10799.lean
mo271: good point, added a test statement to that effect

FormalConjectures/Mathoverflow/10799.lean
mo271: Yeah, that was a bit tricky, and I should have 
 - mentioned that it is easy to confuse natural log with base p log here (in the Kahn-Kalai paper they use $log_p$. 
 - just use `Real.logb p` instead --> doing that now.

FormalConjectures/Mathoverflow/10799.lean
mo271: I'd vote against it and just go with the dummy values, in the spirit that also in the definition of "division" we don't provide proof that the denominator is not zero..
I think it would make statements more ugly, but I suppose we could change it if you think it is necessary

FormalConjectures/Mathoverflow/10799.lean
```diff
@@ -0,0 +1,228 @@
...
+respect to $\mu_p$.
+
+This was a "missing lemma" in the work of Kahn and Kalai on threshold behavior of monotone
+properties. The related [Kahn–Kalai conjecture](https://arxiv.org/abs/math/0603218) was
+[settled by Park and Pham](https://arxiv.org/abs/2203.17207).
+
+**This conjecture is false** without the additional assumption $\mu_t(F) = 1/2$.
+A counterexample was found by Shlomo Perles (April 7, 2026).
+-/
+theorem mathoverflow_10799 : answer(False) ↔
```
mo271, line 131:
```suggestion
-/
@[category research solved, AMS 5 60, formal_proof using formal_conjectures at
"https://github.com/mo271/formal-conjectures/blob/408f53dc0856c0882a5e77acd24fc83b978f0bc9/FormalConjectures/Mathoverflow/10799.lean#L252"]
theorem mathoverflow_10799 : answer(False) ↔
```

GeorgeTsoukalas (approved):
Thanks Moritz!

PR #3713, opened 2026-04-10
feat(docs): Add Docstring Rule to AGENTS.md

franzhusch:
As the website uses the docstrings for additional information, I thought it would be good to add it as a rule to the AGENTS.md.

Let me know if the source part should not be there, then I will remove it again.


AGENTS.md
```diff
@@ -347,6 +347,18 @@ Some rare exceptions exist for consistency:
...
+  problem and a reference to a source:
+  ```lean
+  /--
+  Can every even integer greater than 2 be written as the sum of two primes?
+  See [Goldbach's conjecture](https://en.wikipedia.org/wiki/Goldbach%27s_conjecture).
+  -/
+  @[category research open, AMS 11]
+  theorem goldbach (n : ℕ) (hn : 2 < n) (hn_even : Even n) :
+      ∃ p q, Prime p ∧ Prime q ∧ n = p + q := by
+    sorry
```
mo271, line 381: This might not be the best example: this is using a question in the docstring and not the `answer(sorry)` tag. Perhaps it should be changed in that file and then can be used as example.

mo271 (commented):
Looks good to me in general, we could also add a Linter for checking if a docstring is present for theorems of those categories...
(either in addition to this or as an alternative)

Paul-Lez (approved):

FormalConjectures/Wikipedia/GoldbachConjecture.lean
```diff
@@ -28,10 +28,11 @@ namespace GoldbachConjecture
 
 /--
 Can every even integer greater than 2 be written as the sum of two primes?
+See [Goldbach's conjecture](https://en.wikipedia.org/wiki/Goldbach%27s_conjecture).
```
mo271:
```suggestion
```

we already have the link above

AGENTS.md
```diff
@@ -446,7 +459,8 @@ theorem myNewClass_sanity_check :
 - Follow existing file patterns in the repository
 - Keep formalisations clean and minimal
 - Add references to sources. These should all appear in the module docstring reference list, and
-  ideally each conjecture statement docstring should have a reference.
+  each `research open` and `research solved` theorem docstring MUST have a reference and a
+  concise description of the problem.
```
mo271: I think a file level reference should be fine, especially if there is a file with lots of variants.

AGENTS.md
```diff
@@ -347,6 +347,18 @@ Some rare exceptions exist for consistency:
   -/
   def left_factorial (n : ℕ) := ...
   ```
+- **`research open` and `research solved` docstrings MUST include** a concise description of the
+  problem and a reference to a source:
+  ```lean
+  /--
+  Can every even integer greater than 2 be written as the sum of two primes?
+  See [Goldbach's conjecture](https://en.wikipedia.org/wiki/Goldbach%27s_conjecture).
```
mo271:
```suggestion
```

AGENTS.md
```diff
@@ -347,6 +347,18 @@ Some rare exceptions exist for consistency:
   -/
   def left_factorial (n : ℕ) := ...
   ```
+- **`research open` and `research solved` docstrings MUST include** a concise description of the
```
mo271: and `textbook`

(This new rule, including a linter check was added in the meantime...)

mo271 (approved):
Thanks!

PR #3718, opened 2026-04-12
feat(Website): Total Problems Renaming

franzhusch:
Rename to Total Problems instead of Total Results and also count the number of Open + Solved + Formally Solved instead of all Statements.

Additionally some smaller README edits in regards to local development.


mo271 (approved):
Thanks, LGTM!

I assume this is tested? in general (and for bigger changes to the website) it would be great if you could serve the site with the new changes at your fork, so they can be viewed (happy to explain how to do this, if it gives trouble)

franzhusch:
Oh yeah, sorry I forgot to write that. Yes I have tested it, will look into github branch deployment, the next time that I make a PR on the website.

PR #3719, opened 2026-04-12
Prove monochromatic 3-AP with large difference (Erdős 645)

XC0R:
## Summary

- Prove that any 2-coloring of ℕ contains a monochromatic 3-term arithmetic progression x, x+d, x+2d with d > x.
- Proof by case analysis on c(1), c(3), c(5): if all equal, (1,3,5) works directly. Otherwise, show c becomes eventually constant, which forces a monochromatic triple.
- Based on the result of Brown and Landman (1999).

### DIFF
--- a/FormalConjectures/ErdosProblems/645.lean
+++ b/FormalConjectures/ErdosProblems/645.lean
@@ -28,6 +28,102 @@ import FormalConjectures.Util.ProblemImports
 
 namespace Erdos645
 
+private def hasMono (c : ℕ → Bool) : Prop :=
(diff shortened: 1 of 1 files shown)


XC0R:

mo271 (commented):
Let's use the `formal_proof` attribute for this long (and slow?) proof

Paul-Lez:
@XC0R Note that there's already a Lean proof for this (see the `formal_proof`) attribute. Since this is quite long, we can't merge it into the repo (though you can add your own proof as a link pointing to a specific commit in the `formal_proof` attribute)

PR #3721, opened 2026-04-13
Erdős 198: Sidon set intersects every infinite AP

XC0R:
## Summary

Proves `erdos_198` — there exists a Sidon set that intersects every infinite arithmetic progression.

Constructs A = {(n+1)! + n : n ∈ ℕ} (AlphaProof's construction) and proves:
- **IsSidon A**: via `sum_eq_of_le` helper leveraging factorial growth for contradiction
- **A meets every infinite AP**: for AP with common difference d, picks m = a+d so d | (m+1)!+d

The main theorem shows the answer is False: no Sidon set avoids all infinite APs.

### DIFF
--- a/FormalConjectures/ErdosProblems/198.lean
+++ b/FormalConjectures/ErdosProblems/198.lean
@@ -29,6 +29,13 @@ open Function Set Nat
 
 namespace Erdos198
 
+/-- Let $V$ be a vector space over the rationals and let $k$ be a fixed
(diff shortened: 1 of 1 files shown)


mo271 (changes requested):
Please don't remove the docstrings!

And let's use the `formal_proof using formal_conjectures` at mechanism

mo271 (approved):
Thanks, LGTM

PR #3722, opened 2026-04-13
Erdős 379: limsup S(n) = ∞

XC0R:
## Summary

Proves `erdos_379` — the limsup of S(n) is infinite, where S(n) is the largest r such that every binomial coefficient C(n,k) (1 ≤ k < n) is divisible by p^r for some prime p.

Proof via Tao's approach (Cambie–Kovač–Tao): binomial absorption lemma, Euler's theorem mod p^R, and Bertrand-style prime selection to construct arbitrarily large witnesses n = 2^φ(p^r).

### DIFF
--- a/FormalConjectures/ErdosProblems/379.lean
+++ b/FormalConjectures/ErdosProblems/379.lean
@@ -24,10 +24,93 @@ import FormalConjectures.Util.ProblemImports
 
 namespace Erdos379
 
-open Filter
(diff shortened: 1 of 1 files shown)


mo271 (commented):
Let's use the `formal_proof using formal_conjectures` mechanism here.

mo271 (approved):
Thanks, LGTM!

PR #3723, opened 2026-04-13
Erdős 1043: monic polynomial level set projection

XC0R:
## Summary

Proves `erdos_1043` — for the monic polynomial X^16 - 1, every direction u has projection measure > 2, so no line achieves measure ≤ 2.

Counterexample via X^16 - 1 (Pommerenke). Key steps:
- **Star-convexity at 0**: convexity of z ↦ ‖z-1‖ transfers to the level set
- **Symmetry**: (-z)^16 = z^16
- **Large projection**: for any unit u, a 16th root of unity within π/16 of u gives a level-set point z with Re(z·ū) > 1 (via 2^{1/16}·cos(π/16) > 1)
- **Measure bound**: star-convexity + symmetry ⟹ projection ⊇ closedBall of radius > 1, volume > 2. Transfer to ℝ via `stdOrthonormalBasis` + `LinearIsometryEquiv`.

### DIFF
--- a/FormalConjectures/ErdosProblems/1043.lean
+++ b/FormalConjectures/ErdosProblems/1043.lean
@@ -36,170 +36,6 @@ open MeasureTheory Polynomial
 def levelSet (f : Polynomial ℂ) : Set ℂ :=
   {z : ℂ | ‖f.eval z‖ ≤ 1}
 
-/- ## Counterexample helpers for X^16 - 1 -/
-
-private lemma levelSet_symmetric :
-    ∀ z ∈ levelSet (X ^ 16 - 1 : ℂ[X]), -z ∈ levelSet (X ^ 16 - 1 : ℂ[X]) := by
-  intro z hz
-  unfold levelSet at *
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/1043.lean
mo271:
```suggestion
@[category research solved, AMS 28 30, formal_proof using lean4 at "https://github.com/plby/lean-proofs/blob/main/src/v4.24.0/ErdosProblems/Erdos1043.lean", 
formal_proof uisng formal_conjectures at "https://github.com/XC0R/formal-conjectures/blob/7db17471701f15b125d1c36bc1fa5bb9b702d6be/FormalConjectures/ErdosProblems/1043.lean#L214"]
```

and then remove the long proof...

mo271 (commented):
Thanks, let's use the `formal_proof` mechanism...

mo271 (approved):
Thanks, LGTM

PR #3725, opened 2026-04-13
Erdős 350: distinct subset sums reciprocal bound

XC0R:
## Summary

Proves `erdos_350` — if A ⊂ ℕ has all distinct subset sums, then ∑ 1/n < 2.

**Proof technique:** Order-embed A as e : Fin |A| ↪o ℕ, show partial sums ≥ 2^k - 1 by subset-sum counting, then compare ∑ 1/e(i) ≤ ∑ 1/2^i < 2 via Abel summation by parts (custom lemma with Fin↔range bridge via Finset.Iic).


mo271 (approved):
Thanks, LGTM!

In order to make is easier in the review to see that it compiles at the provided link, it would be great to make a draft pull request at that place, let the CI finish, then add the commit that removes the proof. (But of course it is also possible to manually checkout that commit and check that it compiles)

PR #3727, opened 2026-04-13
Update totalCount calculation in build.js

Smetalo:
Removed _formalCount_ from _totalCount_ calculation, since it is already taken into account by _solvedCount_.


mo271 (approved):
Thanks, good catch!

PR #3728, opened 2026-04-13
fix(ErdosProblems/379): update formal proof reference

Smetalo:


FormalConjectures/ErdosProblems/379.lean
```diff
@@ -37,7 +37,7 @@ $\limsup S(n) = \infty$.
 This was formalized in Lean by Tao.
 -/
 @[category research solved, AMS 11,
-formal_proof using lean4 at "https://github.com/teorth/analysis/blob/main/analysis/Analysis/Misc/erdos_379.lean",
+formal_proof using lean4 at "https://github.com/teorth/analysis/blob/main/Analysis/Misc/erdos_379.lean",
```
mo271:
```suggestion
formal_proof using lean4 at "https://github.com/teorth/analysis/blob/4f623b0f4cacdb967f1f8132db0becaee0f1fb3d/Analysis/Misc/erdos_379.lean#L90",
```
Let's pin down to a specific version number and also give a line number

mo271 (commented):
Thanks!

mo271 (approved):
LGTM, thanks!

PR #3729, opened 2026-04-13
fix(ErdosProblems/397): update formal proof references

Smetalo:
Reinstated the link related to the comment above ("This was formalized in Lean by Wu using Aristotle.").


mo271 (approved):
Thanks, good catch!

PR #3737, opened 2026-04-14
Add formal proof for solved erdos_198.variants.concrete

mzhorvath1:


mo271 (approved):
Thanks, potentially mention that the proof was provided by AlphaProof?

mzhorvath1:
> Thanks, potentially mention that the proof was provided by AlphaProof?

Thanks. It said already that it was found by AlphaProof, but I made it now more precise to mention that it was also proved by AlphaProof.

PR #3740, opened 2026-04-14
Add solved tag for erdos_418.variants.conditional

mzhorvath1:
The problem has a proof within the formal-conjectures repo.


mzhorvath1:
As discussed, this tag is not needed, as the proof can be automatically detected without the tag.

PR #3749, opened 2026-04-15
Update tag to solved for union_closed.variants.singleton_mem

mzhorvath1:


mzhorvath1:
As discussed, this tag is not needed, as the proof can be automatically detected without the tag.

PR #3763, opened 2026-04-16
feat(ErdosProblems): 282

### LINKED ISSUE #471: Erdős Problem 282
### What is the conjecture
    
https://www.erdosproblems.com/282

  Let $A\subseteq \mathbb{N}$ be an infinite set and consider the following greedy algorithm for a rational $x\in (0,1)$: choose the minimal $n\in A$ such that $n\geq 1/x$ and repeat with $x$ replaced by $x-\frac{1}{n}$. If this terminates after finitely many steps then this produces a representation of $x$ as the sum of distinct unit fractions with denominators from $A$. Does this process always terminate if $x$ has odd denominator and $A$ is the set of odd numbers? More generally, for which pairs $x$ and $A$ does this process terminate?

Status: open

### Choose either option

smmercuri:
Fixes #471

### DIFF
--- a/FormalConjectures/ErdosProblems/282.lean
+++ b/FormalConjectures/ErdosProblems/282.lean
@@ -1,5 +1,5 @@
 /-
-Copyright 2026 The Formal Conjectures Authors.
+Copyright 2025 The Formal Conjectures Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/282.lean
```diff
@@ -0,0 +1,121 @@
...
+set of odd numbers? -/
+@[category research open, AMS 5]
+theorem erdos_282 {x : ℚ} (hx : x ∈ Set.Ioo 0 1) (hx_den : Odd x.den) :
+    greedyUnitFractionRem { n | Odd n } x =ᶠ[atTop] 0 := by
+  sorry
+
+/-- More generally, for which pairs $x$ and $A$ does this process terminate? -/
+@[category research open, AMS 5]
+theorem erdos_282.variants.general {x : ℚ} (hx : x ∈ Set.Ioo 0 1) (A : ℚ → Set ℕ) :
+    greedyUnitFractionRem (A x) x =ᶠ[atTop] 0 ↔ A ∈ (answer(sorry) : Set (ℚ → Set ℕ)) := by
```
mo271: Is is such that the set of functions A must work for all $x$ here? which is likely stronger than asking for which pairs $(x, A)$ it terminates. A more faithful representation might be to define a predicate on pairs (`ℚ × Set ℕ`) or to state it as a subset of the Cartesian product?

mo271 (approved):

mo271 (approved):
Thanks, LGTM!

PR #3767, opened 2026-04-16
feat(ErdosProblems): 750

### LINKED ISSUE #927: Erdős Problem 750
### What is the conjecture

https://www.erdosproblems.com/750

  Let $f(m)$ be some function such that $f(m)\to \infty$ as $m\to \infty$. Does there exist a graph $G$ of infinite chromatic number such that every subgraph on $m$ vertices contains an independent set of size at least $\frac{m}{2}-f(m)$?

Status:  open

### Choose either option

smmercuri:
Fixes #927


mo271 (approved):
Thanks, LGTM!

PR #3771, opened 2026-04-16
feat(ErdosProblems): add 15

### LINKED ISSUE #195: Erdős Problem 15: Convergence of a sum involving prime reciprocals
### What is the conjecture
    
https://www.erdosproblems.com/15

  Is it true that 
$$\sum_{n=1}^\infty(-1)^n\frac{n}{p_n}$$
converges, where $p_n$ is the sequence of primes?

Status: open

### Prerequisites needed
None needed. The solved implication from Hardy-Littlewood prime tuples conjecture should also be do-able, since we have https://github.com/google-deepmind/formal-conjectures/blob/ed8a8090f9b10a188daeec228150351831790395/FormalConjectures/Wikipedia/HardyLittlewood.lean

* ams-11

### Choose either option

mo271:
fixes #195

### DIFF
--- a/FormalConjectures/ErdosProblems/15.lean
+++ b/FormalConjectures/ErdosProblems/15.lean
@@ -34,8 +34,8 @@ Note: In the problem statement, $p_n$ is the $n$-th prime, indexed such that $p_
 We 0-index here to reflect how Nat.nth works.
 -/
 @[category research open, AMS 11]
-theorem erdos_15 : answer(sorry) ↔ (∃ m : ℝ, Tendsto (fun n ↦ ∑ i ∈ Finset.range n, (fun k : ℕ =>
-    (-1 : ℝ) ^ (k + 1) * ((k + 1) : ℝ) / (k.nth Nat.Prime : ℝ)) i) atTop (𝓝 m)) := by
+theorem erdos_15 : answer(sorry) ↔
+    Summable (fun k : ℕ => (-1 : ℚ) ^ (k + 1) * (k + 1) / (k.nth Nat.Prime)) := by
   sorry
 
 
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/15.lean
```diff
@@ -0,0 +1,45 @@
...
+
+/--
+Is it true that $\sum_{n=1}^\infty(-1)^n\frac{n}{p_n}$ converges,
+where $p_n$ is the sequence of primes?
+
+Note: In the problem statement, $p_n$ is the $n$-th prime, indexed such that $p_1=2, p_2=3, \ldots$.
+We 0-index here to reflect how Nat.nth works.
+-/
+@[category research open, AMS 11]
+theorem erdos_15 : answer(sorry) ↔ (∃ m : ℝ, Tendsto (fun n ↦ ∑ i ∈ Finset.range n, (fun k : ℕ =>
```
smmercuri: Perhaps `Summable` would be nicer here

FormalConjectures/ErdosProblems/15.lean
```diff
@@ -0,0 +1,45 @@
...
+    (-1 : ℝ) ^ (k + 1) * ((k + 1) : ℝ) / (k.nth Nat.Prime : ℝ)) i) atTop (𝓝 m)) := by
```
smmercuri: Can probably remove some of the coercions to real here

FormalConjectures/ErdosProblems/15.lean
mo271: agreed

FormalConjectures/ErdosProblems/15.lean
mo271: using less and coercing to Rat instead now

smmercuri (approved):

PR #3780, opened 2026-04-16
feat(OpenQuantumProblems): Added proofs for sanity-checked sorrys of OQP23

MarioKrenn6240:
This PR replaced all sanity-checks with formal proofs.


mo271:
@MarioKrenn6240 sorry I didn't see this and added some proofs for this myself. Can you resolved conflicts and just always pick the nicer proofs in the cases where we have two now...

mo271:
@MarioKrenn6240 I'm also happy to do so, just let me know

PR #3782, opened 2026-04-16
Add Erdős Problem 1128 (Prikry-Mills counterexample for monochromatic countable boxes)

henrykmichalewski:
Formalises [Erdős Problem 1128](https://www.erdosproblems.com/1128): whether every 2-colouring of $\omega \times \omega \times \omega_1$ admits a monochromatic product $A \times B \times C$ with $A, B, C$ countably infinite.

## Contents

- **Main theorem** (answer: `False`): the problem is resolved negatively via the Prikry-Mills (1978, unpublished) construction.
- **`prikryMills` lemma** (`sorry`): records the unpublished 1978 Prikry-Mills counterexample colouring.
- **`variants.two_dimensional_false`**: a formal disproof of the naturally-false two-dimensional variant using the ordering colouring on $\omega \times \omega_1$.
- **Four proved helper lemmas** about $\omega_1$ establishing countability and boundedness properties (`card_le_aleph0_of_lt_omega1`, `countable_toType_of_lt_omega1`, `countable_Iio_of_lt_omega1`, `countable_Iio_omega1`, `countable_subset_bdd`) used for the supporting infrastructure.


FormalConjectures/ErdosProblems/1128.lean
mo271, line 1: Per repo conventions, it should appear once - ideally only in the `erdos_1128` theorem docstring. The module docstring should be trimmed to title + reference link.

FormalConjectures/ErdosProblems/1128.lean
```diff
@@ -0,0 +1,359 @@
...
+
+## Note on the 2D analogue
+
+The file also contains a variant `two_dimensional` which claims that in 2D,
+any 2-colouring of $\omega_1 \times \omega_1$ has an uncountable monochromatic
+rectangle $A_1 \times B_1$. This statement appears to be **false** in ZFC:
+the ordering coloring $f(\alpha, \beta) = 0 \text{ iff } \alpha < \beta$
+provides a counterexample, since any uncountable subset of $\omega_1$ is cofinal
+and hence cannot be bounded away from any other uncountable set.
+-/
```
mo271, line 23:
```suggestion
/-!
# Erdős Problem 1128

*Reference:* [erdosproblems.com/1128](https://www.erdosproblems.com/1128)
-/
```

FormalConjectures/ErdosProblems/1128.lean
```diff
@@ -0,0 +1,359 @@
...
+def IsMonochromaticBox {A B C : Type*} (f : A → B → C → Fin 2)
+    (A₁ : Set A) (B₁ : Set B) (C₁ : Set C) : Prop :=
+  ∃ c : Fin 2, ∀ a ∈ A₁, ∀ b ∈ B₁, ∀ c' ∈ C₁, f a b c' = c
+
+/-
+Auxiliary lemmas for the Prikry–Mills construction.
+These establish key countability and boundedness properties of ω₁.
+-/
+
+private abbrev Omega1 := {o : Ordinal.{0} // o < (aleph 1).ord}
```
mo271: Let's not redefne `ω₁` with this dubious definition which perhaps give the wrong universe level. Mathib's already sits in `Type 0` which is probably what Prikry-Mills needs here. 

Also: this will probably alow us to elminate all the lemmas below: they are only thin wrappers around existing mathlib api already. 

this would als perhaps allow us to eliminate the two versions of Prikry Mills:
 Keep one and derive the other, or just keep one. Since the auxiliary lemmas are stated in terms of Omega1, the explicit version is more natural as the base case.

FormalConjectures/ErdosProblems/1128.lean
```diff
@@ -0,0 +1,359 @@
...
+$A_1$ would be strictly less than every element of $B_1$, making $A_1$ bounded above in
+$\omega_1$; but any bounded subset of $\omega_1$ is countable (since initial segments are
+countable), contradicting $A_1$ being uncountable. The colour-1 case is symmetric with
+the roles of $A_1$ and $B_1$ swapped.
+
+Note: The correct classical result for 2-colourings of pairs (not products) is the
+Erdős–Rado theorem $\omega_1 \to (\omega_1)^2_2$, which concerns unordered pairs.
+-/
+@[category research solved, AMS 3 5]
+theorem erdos_1128.variants.two_dimensional_false :
```
mo271, line 233: `two_dimensional_false` seems not super closely related to 1128, but I guess we can keep it...

mo271 (commented):
Thanks!

henrykmichalewski:
Applied mo271's docstring-convention review pass in [`663bcd6`](https://github.com/henrykmichalewski/formal-conjectures/commit/663bcd6):
- Lines 1, 63: trimmed the module docstring to title + reference link per the literal suggestion.
- Line 286 (`two_dimensional_false`): kept per mo271's "I guess we can keep it".

Line 81 follow-up (replace local `Omega1` redefinition with Mathlib's `ω₁`): I attempted the rewrite to `{o : Ordinal // o < ω₁}` but mixing `ω₁` and `(ℵ_ 1).ord` triggers definitional-equality issues across the auxiliary wrapper lemmas and the Prikry–Mills proof (despite `ord_aleph 1 : (ℵ_ 1).ord = ω_ 1`). Rather than risk a sorry-explosion in this PR I left a TODO at the `Omega1` declaration documenting mo271's intent (eliminate the wrappers, switch to Mathlib's `ω₁` directly), and reverted to the original `{o : Ordinal // o < (aleph 1).ord}` form. Happy to do the full refactor as a follow-up PR.

Build: `lake build FormalConjectures.ErdosProblems.«1128»` is green (pre-existing missing-attribute warnings on the private helper lemmas are unchanged).

mo271:
> Line 81 follow-up (replace local `Omega1` redefinition with Mathlib's `ω₁`): I attempted the rewrite to `{o : Ordinal // o < ω₁}` but mixing `ω₁` and `(ℵ_ 1).ord` triggers definitional-equality issues across the auxiliary wrapper lemmas and the Prikry–Mills proof (despite `ord_aleph 1 : (ℵ_ 1).ord = ω_ 1`). Rather than risk a sorry-explosion in this PR I left a TODO at the `Omega1` declaration documenting mo271's intent (eliminate the wrappers, switch to Mathlib's `ω₁` directly), and reverted to the original `{o : Ordinal // o < (aleph 1).ord}` form. Happy to do the full refactor as a follow-up PR.

Let's rather do the full refactor now instead of introducing the alternative `Omega1`

henrykmichalewski:
@mo271 — done in [`d094963`](https://github.com/henrykmichalewski/formal-conjectures/commit/d094963).

The previous TODO is removed. The local `Omega1 := {o // o < (aleph 1).ord}` is now `Omega1 := {o // o < ω_ 1}` (Mathlib's `ω_ 1`), and all 5 wrapper lemma signatures use `ω_ 1` directly.

The defeq gap between `(ℵ_ 1).ord` and `ω_ 1` (provably equal via `ord_aleph 1` but not definitionally so in this Mathlib version) is bridged at 4 specific sites using `lt_of_lt_of_eq` with `ord_aleph 1` / `(ord_aleph 1).symm`:
- `card_le_aleph0_of_lt_omega1` (calls `(isInitial_ord (aleph 1)).card_lt_card.mpr`)
- `countable_subset_bdd` (hf_lt' / hlt / hsucc_lt — calls `iSup_sequence_lt_omega_one` and `(isSuccLimit_ord _).succ_lt`)
- `countable_subset_bdd` empty case (uses `isRegular_aleph_one.ord_pos`)
- `prikryMills` `ha₀_succ_lt` (same `(isSuccLimit_ord _).succ_lt` pattern)

I kept the `Omega1` abbrev rather than fully inlining `{o : Ordinal // o < ω_ 1}` everywhere — the subtype appears in many use sites in `prikryMills` and the surrounding proof, and inlining would be a larger churn for no semantic gain. The wrapper lemmas are also retained (their signatures cleaned up), since they're called from the prikryMills proof. Happy to do the full inline if you'd prefer.

Builds cleanly:
```
lake build FormalConjectures.ErdosProblems.«1128»
```

mo271:
hanks for the cleanup, this is already much better! 

Using `ω_ 1` directly is definitely the right approach here. However, we can actually simplify these helper proofs significantly and completely eliminate the `ord_aleph 1` proposition bridging by using standard Mathlib lemmas:

1. **`countable_Iio_of_lt_omega1`**: Instead of using `card_le_aleph0` and `countable_toType`, we can prove this in a single line by leveraging `lt_omega_iff_card_lt` (which relates `< ω_ 1` to `< ℵ₁`):
   ```lean
   private lemma countable_Iio_of_lt_omega1 {γ : Ordinal} (hγ : γ < ω_ 1) :
       Countable (Set.Iio γ) := by
     rwa [countable_iff_lt_aleph_one, mk_Iio_ordinal, lift_lt_aleph_one, ← lt_omega_iff_card_lt]
   ```
   This makes `card_le_aleph0_of_lt_omega1` and `countable_toType_of_lt_omega1` redundant, so they can be safely deleted!

2. **`countable_subset_bdd`**: We can apply the regularity of `ℵ₁` directly to the subtype `↥S` using `iSup_lt_ord_of_isRegular` and regularity of `ℵ₁` without sequence indexation. For the empty case, `omega_pos 1` gives `0 < ω_ 1` directly.

3. **Successor Limit**: Since `ω_ 1` is a limit ordinal, we can prove `ha₀_succ_lt` in `two_dimensional_false` using `isSuccLimit_omega 1` directly:
   ```lean
   have ha₀_succ_lt : a₀.val + 1 < ω_ 1 := by
     rw [← succ_eq_add_one]
     exact (isSuccLimit_omega 1).succ_lt a₀.2
   ```

This completely avoids manual `ord_aleph 1` rewrites across the codebase. Does that make sense?

henrykmichalewski:
Pushed (`786e9e4`).

* `countable_Iio_of_lt_omega1` collapses to the single `rwa` you suggested; `card_le_aleph0_of_lt_omega1` and `countable_toType_of_lt_omega1` are gone.
* `isSuccLimit_omega 1` eliminates the limit-ordinal bridging in `countable_subset_bdd` and `two_dimensional_false`.
* `omega_pos 1` handles the empty case.

One detail: I tried `iSup_lt_ord_of_isRegular` on `↥S` directly but it pins the codomain to `Ordinal.{1}` (the iSup-universe TODO in Regular.lean), so I kept `iSup_sequence_lt_omega_one` for the supremum step. The `ord_aleph 1` rewrite to bridge `(ℵ_ 1).ord ↔ ω_ 1` is now the only one left in the file.

mo271 (approved):
Thanks again, LGTM!

PR #3784, opened 2026-04-16
Add Erdős Problem 70 (3-uniform partition relation on the continuum)

### LINKED ISSUE #303: Erdős Problem 70
### What is the conjecture
    
https://www.erdosproblems.com/70

  Let $\mathfrak{c}$ be the ordinal of the real numbers, $\beta$ be any countable ordinal, and $2\leq n<\omega$. Is it true that $\mathfrak{c}\to (\beta, n)_2^3$?

### Choose either option

henrykmichalewski:
Closes https://github.com/google-deepmind/formal-conjectures/issues/303

Formalises [Erdős Problem 70](https://www.erdosproblems.com/70): the 3-uniform partition relation $\mathfrak{c} \to (\beta, n)^3_2$ for every countable ordinal $\beta$ and every finite $n$.

## Contents

- **New definition `OrdinalCardinalRamsey3`**: uses `Set.Triplewise` to generalise the 2-uniform partition relation to triples.
- **Main open theorem**: $\mathfrak{c} \to (\beta, n)^3_2$ for all countable $\beta$ and finite $n$.
- **Erdős-Rado partial variant**.
- **Monotonicity lemma** (proved).


henrykmichalewski:
Closes #303

FormalConjectures/ErdosProblems/70.lean
```diff
@@ -0,0 +1,208 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
```
Paul-Lez: nit: Could you update the copyright year to 2026 for this new file?

FormalConjectures/ErdosProblems/70.lean
```diff
@@ -0,0 +1,208 @@
...
+
+Here:
+- `𝔠.ord = Cardinal.continuum.ord` is the least ordinal of cardinality the continuum.
+- `β` ranges over countable ordinals (i.e. `β.card ≤ ℵ₀`).
+- `n : ℕ` satisfies `2 ≤ n`.
+- `OrdinalCardinalRamsey3 α β c` is the 3-uniform partition relation `α → (β, c)³₂`.
+-/
+@[category research open, AMS 3]
+theorem erdos_70 :
+    answer(True) ↔
```
Paul-Lez: Should this use `answer(sorry)` rather than `answer(True)`? The Erdős page marks the general relation open, so `answer(True)` makes the file record a solved positive answer.

FormalConjectures/ErdosProblems/70.lean
```diff
@@ -0,0 +1,208 @@
...
+* There is a blue-monochromatic subset of cardinality `c`: a set `s ⊆ α.ToType` with `#s = c`
+  such that every three distinct elements of `s` are colored blue.
+
+The coloring is given as a function `col : α.ToType → α.ToType → α.ToType → Bool`
+(or equivalently, a predicate `isRed : α.ToType → α.ToType → α.ToType → Prop`) on
+unordered triples; we represent it symmetrically via `Set.Triplewise`.
+-/
+def OrdinalCardinalRamsey3 (α β : Ordinal.{u}) (c : Cardinal.{u}) : Prop :=
+  -- For any partition of 3-element subsets into red and blue:
+  ∀ (isRed : α.ToType → α.ToType → α.ToType → Prop),
```
Paul-Lez, line 57: Should this be a coloring of unordered 3-element subsets rather than an arbitrary predicate on ordered triples? As written, `isRed x y z` need not be symmetric and can also be queried on repeated vertices. I would suggest quantifying over `{s : Finset α.ToType // s.card = 3}` or adding the symmetry and distinctness constraints explicitly.

henrykmichalewski:
@Paul-Lez — applied the mechanical nits in [`bc14a7e`](https://github.com/henrykmichalewski/formal-conjectures/commit/bc14a7e), on top of an upstream/main merge: `@[category undergraduate]` → `@[category textbook]` (and copyright 2026 where applicable). The substantive content comments on this PR (open characterization vs. existence/universal phrasing, etc.) are tracked separately and will be addressed in follow-up commits per PR.

henrykmichalewski:
Thanks for the review @PaulLez. Pushed b0a7ad0 on `add-problem-70`: (1) added an explicit symmetry constraint requiring `isRed` to be invariant under permutation of three distinct arguments, faithfully encoding a coloring of unordered 3-element subsets; (2) switched the headline from `answer(True)` to `answer(sorry)` since the general relation is OPEN per the Erdős page.

mo271 (commented):
Thanks, LGTM mathematically!

A couple of minor recommendations to improve compliance with the repository's style guidelines (as mentioned in some other formalisations of yours already):

1. Avoid Redundant Status Mentions
The status of a problem (e.g., whether it is open or solved) should only be designated canonically via the category attributes (such as @[category research open] and @[category research solved]).

We recommend removing the explicit text mentions of the status (Open, OPEN, TRUE, Status: ...) from the comments and docstrings, as they are redundant and prone to getting out of sync. Specifically, please clean up:

Line 22: status O in **Verbatim statement (Erdős #70, status O):**
Line 27: **Notes:** OPEN
Lines 50–52: ## Status section
Line 108: (Open) in **Erdős Problem 70** (Open):
Line 140: **Status**: TRUE (Erdős–Rado).
Line 155: **Status**: OPEN ...
Line 171: **Status**: OPEN.
2. 📝 Docstring Redundancies / Repetition
To align with repository best practices, the module docstring (lines 19–59) should be kept extremely minimal (containing only the problem title, reference links, and a 1-2 sentence high-level context). Detailed descriptions of definitions, parameters, and variants should live exclusively directly on their respective code declarations to avoid duplication and keep them easier to maintain:

Definition Duplication: The explanation of OrdinalCardinalRamsey3 is duplicated in the module docstring (lines 54–58) and on the definition itself (lines 69–84).
Theorem Duplication: The description of the erdos_rado variant is repeated in both the module header (lines 45–48) and on the theorem itself (lines 131–141).

henrykmichalewski:
@mo271 — applied both style items in [`1cdda88`](https://github.com/henrykmichalewski/formal-conjectures/commit/1cdda88):

**Item 1 — redundant status mentions:** dropped from all 7 locations you flagged (lines 22, 27, 50–52, 108, 140, 155, 171). The `@[category research open/solved]` attributes are now the sole source of status truth.

**Item 2 — docstring deduplication:**
- Module docstring trimmed from a 5-section block to the standard `# Title` + `*Reference:*` + 2-sentence context.
- `OrdinalCardinalRamsey3` explanation now lives only on the definition (was duplicated module + def).
- `erdos_rado` variant context now lives only on the theorem (was duplicated module header + theorem).

Net: 212 → 163 lines, no semantic changes. Build green.

mo271 (commented):
Thanks! A two minor issues remain: 
 - Please use latex markdown for all docstrings
 - Just use `(𝔠).ord` directly instead of defining `continuumOrd`. This should be available after `open scoped Cardinal` (already present in the file)

henrykmichalewski:
Pushed (`9f71721`):

* Replaced `continuumOrd` everywhere with `(𝔠).ord` inline (parens needed: `𝔠.ord` parses as `«𝔠».ord`).
* Added `open scoped Cardinal` since plain `open Cardinal` doesn't bring the `𝔠` scoped notation.
* Converted docstring math (`α → (β, c)³₂`, `β.card ≤ ℵ₀`, ...) to LaTeX (`$\alpha \to (\beta, c)^3_2$`, etc.).

mo271 (commented):
Thanks!

both `omega_three` and `omega_one` are marked as `research open` while havin ghte `answer(True)` filled in:

- `omega_three` claims answer(True) without justification: The Erdős-Rado result proves $\mathfrak{c}\to (\omega+n, 4)_2^3$, but that does not imply $\mathfrak{c}\to (\omega, 3)_2^3$ (different blue side). No citation is given. Should be answer(sorry) / research open, right?

 - omega_one claims answer(True) without justification: $\omega_1$ is uncountable: this goes beyond the scope of Problem 70 and has no citation for being solved. either add a citation or potentially label `textbook` if it is clear. 

Slight non-verbatim quote: 
The original says "the ordinal of the real numbers" but the formalization writes "the cardinality of the continuum". The code ((𝔠).ord) is correct, but the docstring should match the source verbatim unless there is a good reason to deviate.

mo271 (approved):
Thanks again, LGTM now!

PR #3785, opened 2026-04-16
Add Erdős Problem 111 (bipartite edge deletion for uncountable chromatic graphs)

### LINKED ISSUE #341: Erdős Problem 111
### What is the conjecture
    
https://www.erdosproblems.com/111

  If $G$ is a graph let $h_G(n)$ be defined such that any subgraph of $G$ on $n$ vertices can be made bipartite after deleting at most $h_G(n)$ edges.What is the behaviour of $h_G(n)$? Is it true that $h_G(n)/n\to \infty$ for every graph $G$ with chromatic number $\aleph_1$?

### Choose either option

henrykmichalewski:
fixes #341 

Formalises [Erdős Problem 111](https://www.erdosproblems.com/111): for a graph $G$ with $\chi(G) = \aleph_1$, study the function $h_G(n)$ measuring the minimum edges that must be deleted from a subgraph of order $n$ to render it bipartite.

## Contents

- **Definitions**: `bipartiteDistance` (the minimum number of edges to delete to reach a bipartite graph) and `hG`.
- **Main open theorem**: $h_G(n)/n \to \infty$ as $n \to \infty$.
- **Lower and upper bound variants** (Erdős-Hajnal-Szemerédi).
- **Erdős's conjectured sharper bound**.


henrykmichalewski:
Closes #341

FormalConjectures/ErdosProblems/111.lean
```diff
@@ -0,0 +1,168 @@
...
+If $G$ is a graph with chromatic number $\aleph_1$, is it true that $h_G(n)/n \to \infty$?
+
+Formally: for every graph $G$ (on a type $V$) with chromatic cardinal $\aleph_1$, the ratio
+$h_G(n)/n$ is unbounded, i.e., for every constant $C$, there are arbitrarily large $n$ such
+that $h_G(n) > C \cdot n$.
+-/
+@[category research open, AMS 5]
+theorem erdos_111 : answer(sorry) ↔
+    ∀ (V : Type) (G : SimpleGraph V), G.chromaticCardinal = ℵ_ 1 →
+      ∀ C : ℝ, ∃ n : ℕ, 0 < n ∧ C * n < hG G n := by
```
Paul-Lez: Does `∀ C, ∃ n, C * n < hG G n` quite capture `h_G(n)/n → ∞`? It gives unboundedness of the ratio, but not eventual growth along `atTop`. Could you state this using a filter/Tendsto formulation, or at least as `∀ C, ∀ N, ∃ n ≥ N, C * n < ...`?

FormalConjectures/ErdosProblems/111.lean
```diff
@@ -0,0 +1,168 @@
...
+The stronger form: $h_G(n)/n$ is not bounded above by any constant — there exist arbitrarily
+large $n$ with $h_G(n) > n / 2$.
+
+This follows from the fact that every graph with uncountable chromatic number contains
+$\aleph_1$ vertex-disjoint odd cycles of some fixed odd length $\ell$: any induced subgraph
+on $n$ vertices containing $\gg n / \ell$ such cycles requires removing $\gg n$ edges to
+become bipartite.
+-/
+@[category research solved, AMS 5]
+theorem erdos_111.variants.lower_bound :
```
Paul-Lez, line 102: Is this lower-bound variant meant to be the same assertion as the main open question? It appears to have essentially the same conclusion as `erdos_111` but is marked solved. Could you either weaken it to the known theorem you want to record, or keep it open/omit it if the intended statement is exactly the conjecture?

henrykmichalewski:
Thanks for the review @PaulLez. Pushed 8b0c922 on `add-problem-111`: (1) the headline now uses `Filter.Tendsto … atTop atTop` to faithfully express $h_G(n)/n \to \infty$; (2) the `lower_bound` variant has been weakened to the actually-known fact $\exists c > 0,\, c \cdot n \leq h_G(n)$ eventually, so it is no longer indistinguishable from the open headline.

mo271 (commented):
Nice formalization with well-chosen definitions! A few issues:

1. Erdős conjecture variant uses = ℵ₁ but source says ≥ ℵ₁

The erdos_conjecture variant (line 167) states the upper bound for graphs with chromaticCardinal = ℵ_ 1, but the source explicitly says "for every graph $G$ with chromatic number $\geq \aleph_1$ and every $n$". This matters because a graph with larger chromatic number could potentially have denser subgraphs, so the bound for = ℵ₁ doesn't immediately imply the bound for ≥ ℵ₁. Consider using ℵ_ 1 ≤ G.chromaticCardinal.

2. Misleading docstring for lower_bound

Line 53 says "$h_G(n) \gg n$: this holds (with $h_G(n)/n \to \infty$)". But if $h_G(n)/n \to \infty$ were known, the main problem (which is labeled OPEN) would already be solved. The known result is just the linear lower bound $h_G(n) \geq cn$ for some constant $c > 0$, not super-linear growth. Please fix the parenthetical to avoid confusion.

3. Problem statement repeated

The problem appears in the verbatim quote (line 23) and again in the "Problem statement" section (lines 43–47). It should appear only once.

4. Missing cross-reference to Problem 74

The source page explicitly says "See also [74]", but the file doesn't mention this related

henrykmichalewski:
Thanks @mo271 for the review. Pushed `24ca0f9` addressing all four points:

1. **`erdos_conjecture` variant uses `≥ ℵ_1`.** Changed the hypothesis from `G.chromaticCardinal = ℵ_ 1` to `ℵ_ 1 ≤ G.chromaticCardinal` to match the source ("for every graph G with chromatic number ≥ ℵ_1 and every n"). Updated the docstring accordingly to note that `≥ ℵ_1` is the correct quantification (a graph with larger chromatic cardinal is still required to satisfy the bound).

2. **Clarified `lower_bound` docstring.** Reworded to state plainly that the *linear* lower bound `h_G(n) ≥ c·n` is the known result (no longer claims `h_G(n)/n → ∞`, which would solve the open problem). Added a sentence clarifying that the open question is precisely whether the linear lower bound can be improved to super-linear.

3. **Deduped problem statement.** Trimmed the module docstring to the title + `*References:*` block; removed the `## Problem statement` and `## Known results` sections (these were verbatim duplicates of content already in the `erdos_111` and variant docstrings).

4. **Cross-reference to Problem 74.** Added "See also Erdős Problem [74]." to the module docstring per the source.

Local `lake build FormalConjectures.ErdosProblems.«111»` succeeds.

FormalConjectures/ErdosProblems/111.lean
```diff
@@ -0,0 +1,144 @@
...
+$G$ with $\chi(G) \geq \aleph_1$, we have $h_G(n) \leq C(\varepsilon) \cdot n^{1 + \varepsilon}$
+for all $n$.
+
+This conjecture (if true) would mean that $h_G(n)$ grows only barely super-linearly. The
+quantification is over graphs with $\chi(G) \geq \aleph_1$ (matching the source); using
+$\chi(G) \geq \aleph_1$ is strictly stronger than $\chi(G) = \aleph_1$ as a hypothesis on the
+class of graphs since the bound must hold uniformly for any larger chromatic cardinal too.
+-/
+@[category research open, AMS 5]
+theorem erdos_111.variants.erdos_conjecture : answer(sorry) ↔
```
mo271, line 137: possibly too strong?

The source says: "[EHS82] proved that there is a $G$ with $\chi(G) = \aleph_1$ such that $h_G(n) \ll n^{3/2}$. Erdős conjectured that this can be improved to $\ll n^{1+\epsilon}$."

The word "this" most naturally refers to the existential example, i.e., there exists $G$ with $h_G(n) \ll n^{1+\epsilon}$. But the formalization uses a universal quantification: for all $G$ with $\chi(G) \geq \aleph_1$, $h_G(n) \leq C n^{1+\epsilon}$.

The universal version is strictly stronger. If Erdős indeed conjectured the universal bound, this is fine. The docstring explicitly notes: "The quantification is over graphs with $\chi(G) \geq \aleph_1$ (matching the source)." However, I cannot confirm from the source text alone that Erdős made this universal claim.

Can you do a careful check against the original paper [Er81]. If Erdős only conjectured the existential version, the statement should use ∃ instead of ∀ for the graph. The docstring's claim "matching the source" may be incorrect, maybe fine, just to double check...

FormalConjectures/ErdosProblems/111.lean
```diff
@@ -0,0 +1,144 @@
...
+noncomputable def bipartiteDistance {V : Type*} (G : SimpleGraph V) : ℕ :=
+  sInf {k : ℕ | ∃ H : SimpleGraph V, H ≤ G ∧ H.IsBipartite ∧
+    Set.ncard (G.edgeSet \ H.edgeSet) = k}
+
+/-- The function $h_G(n)$: for a graph $G$ on vertex type $V$, `hG G n` is the supremum over
+all sets $S \subseteq V$ with $|S| = n$ of the bipartite distance of the induced subgraph on $S$.
+
+This captures the problem's notion: "any subgraph of $G$ on $n$ vertices can be made bipartite
+after deleting at most $h_G(n)$ edges." -/
+noncomputable def hG {V : Type*} (G : SimpleGraph V) (n : ℕ) : ℕ :=
```
mo271, line 64:
```suggestion
noncomputable def hG {V : Type} (G : SimpleGraph V) (n : ℕ) : ℕ :=
```

mo271 (commented):
Thanks, LGTM, except for a question on `erdos_111.variants.erdos_conjecture`

PR #3788, opened 2026-04-17
Add Erdős Problem 601 (infinite path or large independent set, $500 prize)

### LINKED ISSUE #820: Erdős Problem 601
### What is the conjecture
    
https://www.erdosproblems.com/601

  For which limit ordinals $\alpha$ is it true that if $G$ is a graph with vertex set $\alpha$ then $G$ must have either an infinite path or independent set on a set of vertices with order type $\alpha$?
  

Status:  open

### Choose either option

henrykmichalewski:
## Problem
Erdős Problem 601: https://www.erdosproblems.com/601 ($500 prize)

> For which limit ordinals α is it true that if G is a graph with vertex set α, then G must have either an infinite path or an independent set on a set of vertices with order type α?

## Contents
- Main open theorem `erdos_601` ($500 prize note)
- New `HasPathOrIndepSetOfType α` definition
- Variants covering:
  - Erdős-Hajnal-Milner [EHM70]: TRUE for α < ω₁^(ω+2)
  - The specific $250 sub-case at α = ω₁^(ω+2)
  - Larson [La90] result under Martin's Axiom
  - Base case α = ω

Closes #820


FormalConjectures/ErdosProblems/601.lean
```diff
@@ -0,0 +1,201 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
```
Paul-Lez: nit: Could you update the copyright year to 2026 for this new file?

FormalConjectures/ErdosProblems/601.lean
```diff
@@ -0,0 +1,201 @@
...
+*Known partial results:*
+- Erdős–Hajnal–Milner proved the property holds for all `α < ω₁ ^ (ω + 2)`
+  (see `erdos_601.variants.erdos_hajnal_milner_1970`).
+- Larson proved the property holds for all `α < 2 ^ ℵ₀` assuming Martin's Axiom
+  (see `erdos_601.variants.under_martin_axiom`).
+
+**Status**: OPEN.
+-/
+@[category research open, AMS 5]
+theorem erdos_601 :
```
Paul-Lez, line 58: Does the source ask for a characterization of which limit ordinals have the property? This states that every limit ordinal has it, which is a much stronger concrete assertion. Could you put the class or criterion behind `answer(sorry)`, or otherwise phrase `erdos_601` as a characterization problem, with the all-ordinals statement as a named conjectural variant only if intended?

henrykmichalewski:
@Paul-Lez — applied the mechanical nits in [`69e9408`](https://github.com/henrykmichalewski/formal-conjectures/commit/69e9408), on top of an upstream/main merge: `@[category undergraduate]` → `@[category textbook]` (and copyright 2026 where applicable). The substantive content comments on this PR (open characterization vs. existence/universal phrasing, etc.) are tracked separately and will be addressed in follow-up commits per PR.

henrykmichalewski:
Thanks for the review @PaulLez. Pushed d11479d on `add-problem-601`: the headline now asserts pointwise equivalence between `HasPathOrIndepSetOfType α` and a conjectural predicate behind `answer(sorry)` for every limit ordinal `α` (capturing the source's *characterization* question); the previous universal "holds for every limit ordinal" form is moved to a named variant `erdos_601.variants.universal`.

mo271 (commented):
Thanks for the contribution

Same comment as many other pull request applies regarding the docstrings. 
One more severe problem here is the use of placeholders: we don't want that submitted in our repo. (Talking about `Prop`-ing out Martin's axiom.

PR #91, opened 2025-05-28
feat: formalise Lehmer totient

exekis:
hello! this PR adds the [Lehmer totient problem](https://en.wikipedia.org/wiki/Lehmer%27s_totient_problem) under `formal-conjectures/FormalConjectures/Wikipedia/LehmerTotient.lean` (closes issue #90), theorem uses sorry (unsolved question for benchmarking), included license header, import, docstrings, and @[category research open, AMS 11]. Builds with `lake build`. Please take a loot at the last commit: `feat: formalise Lehmer totient` :)

PR #3795, opened 2026-04-17
Add 16 WOWII open conjectures (batch 1)

henrykmichalewski:
Adds 16 open conjectures from the WOWII (Written on the Wall II) database by E. DeLaViña, following the existing pattern in FormalConjectures/WrittenOnTheWallII/.

Reference: http://cms.dt.uh.edu/faculty/delavinae/research/wowII/

## Scope
- **16 new conjecture files** (all status `@[category research open, AMS 5]`)
- **~900 lines** total Lean (statements + sanity-check tests)
- **8 primary invariants** covered: Ls, b, path, tree, Hamiltonian, well-TD, etc.
- **16 sanity-check tests**, all fully proved

## New definitions
- `largestInducedTreeSize` (local, used in 141 and 200)
- `averageEccentricity` (local, used in 198a)
- `pendantVertices` (shared across 315, 316)
- `IsTotalDominatingSet`, `IsMinimalTotalDominatingSet`, `isWellTotallyDominated` (shared across 315/316/322/327, defined in 315 and imported)

## Batch strategy
This is batch 1 of ~4 batches covering 40 open WOWII conjectures. This batch targets the easiest formalizations (existing or trivial-to-add invariants). Subsequent batches will add residue, distmin/distmax, σ (2-domination), etc.

## Checks
- Build passes cleanly
- All 16 sanity tests fully proved (zero new sorry in tests)
- Following existing Apache 2.0 license header convention

Assisted by Claude (Anthropic).

PR #85, opened 2025-05-28
chore: Move `GoldbachConjecture.lean` to the `Wikipedia` subdirectory

exekis:
This PR moves `GoldbachConjecture.lean` to the `Wikipedia` subdirectory to make it more visible and to stay consistent with convention of structuring the repo by problem source. 

~~hello! this PR adds Collatz conjecture stmt to Wikipedia/CollatzConjecture.lean (closes issue #83 ) with collatz_fn, the conjecture theorem (uses sorry ofc), license header, import, docstrings, and @[category research open, AMS 11]. Builds with `lake build` PTAL! :)~~

PR #3796, opened 2026-04-17
Add 10 WOWII open conjectures (batch 2)

henrykmichalewski:
Adds 10 open conjectures from WOWII (Written on the Wall II) by E. DeLaViña.
Follow-up to #3795 (batch 1, 16 conjectures).

Reference: http://cms.dt.uh.edu/faculty/delavinae/research/wowII/

## Conjectures

- **distEven-based (3):** Conj 23, 24, 63 — lower bounds on b(G), f(G) via min distEven
- **tree(G)-based (3):** Conj 72, 84, 85 — tree(G) lower bounds
- **Upper bounds (2):** Conj 91 on b(G), Conj 96 on α(G)
- **Hamiltonian sufficient (2):** Conj 189 (uses σ = 2-dom), Conj 199 (uses κ = connectivity)

## New definitions (local to each file, self-contained)

- `distEven G v` — count of vertices at even distance from v
- `largestInducedTreeSize` — reused from batch 1
- `twoDominationNumber` — σ(G), 2-domination number
- `vertexConnectivity` — κ(G), for Conj 199

Each file is self-contained (definitions copied as needed) to keep PRs atomic.

## Tests

All 10 files include 1-2 fully-proved `@[category test]` examples.

## Build

Passes cleanly.


mo271:

henrykmichalewski:
Thanks for flagging — re-audited all 10 files in this PR against the WOWII source. You were right to ask: two files in this PR exhibit the same off-by-one numbering shift you fixed in #3823 / #3824.

Pushed [6c5df66](https://github.com/henrykmichalewski/formal-conjectures/commit/6c5df66) addressing the off-by-one cascade:
- The previous `GraphConjecture23.lean` actually stated WOWII #24 verbatim (`b(G) ≥ max-l(v) + ⌈min dist_even(v)/3⌉`). Renamed to `GraphConjecture24.lean`, namespace and docstring updated, wrapped with `answer(False) ↔ ...` since #24 is status F.
- The previous `GraphConjecture24.lean` actually stated WOWII #25 verbatim (`b(G) ≥ 2⌈(1+min dist_even(v))/3⌉`). Renamed to `GraphConjecture25.lean`, same disproof pattern (#25 is also status F; WOWII notes "same counterexample as #24").
- The correct WOWII #23 is already in tree on `main` from your #3823, so we don't add a `GraphConjecture23.lean` here.

Full audit also surfaced one independent (non-numbering) issue I want to disclose for transparency, **not** addressed in this commit since it's outside the scope of your numbering question:
- `GraphConjecture84.lean` — WOWII #84 reads `tree(G) ≥ 2 rad(G) / δ(G)` where `δ(G)` is the minimum degree. The Lean uses `averageDistance G` as the denominator (interpreting δ as `d`). Happy to push a follow-up commit fixing this if you'd like it bundled in this PR; otherwise we'll file it separately.

`GraphConjecture189.lean` is on the borderline — it carries a `TODO` hedge about whether `σ(G)` should be the 2-domination number. The standard Fink–Jacobson reading is correct and consistent across WOWII #188/#190/#201, so the `TODO` is overcautious; we'll remove it in the same follow-up.

henrykmichalewski:
Follow-up: pushed [142c74c](https://github.com/henrykmichalewski/formal-conjectures/commit/142c74c) bundling the two non-numbering audit findings I disclosed earlier:

- **`GraphConjecture84.lean`** — replaced the denominator `averageDistance G` with `G.minDegree` to match the WOWII statement `tree(G) ≥ 2 rad(G) / δ(G)` (lowercase `δ` in the WOWII HTML, not `d`). Dropped the `0 < averageDistance G` hypothesis and added `0 < G.minDegree`.
- **`GraphConjecture189.lean`** — removed the σ-interpretation `TODO` hedge and added a Fink–Jacobson 1985 reference. WOWII uses `σ(G)` for the 2-domination number consistently across Conjectures 188/189/190/201.

Both files build cleanly. PR is now in `8/10 sorry` shape with 0 open audit issues.

FormalConjectures/WrittenOnTheWallII/GraphConjecture24.lean
```diff
@@ -0,0 +1,78 @@
...
+
+For a simple connected graph `G`,
+`b(G) ≥ λ(G) + ⌈(min_v dist_even(v)) / 3⌉`
+where `b(G)` is the size of a largest induced bipartite subgraph,
+`λ(G) := max_v l(v)` is the maximum over all vertices of the independence
+number of the open neighbourhood of `v`, and `dist_even(v)` is the number of
+vertices at even distance from `v`.
+-/
+@[category research solved, AMS 5]
+theorem conjecture24 : answer(False) ↔
```
Paul-Lez, line 51: Could you move the graph-type quantification under the `∀` after `answer(False)` here? Our convention for a solved negative yes/no problem is a single global statement, e.g. `answer(False) ↔ ∀ {α : Type*} [Fintype α] [DecidableEq α] [Nontrivial α], ∀ (G : SimpleGraph α), ...`, rather than fixing `α` as an outer theorem variable.

FormalConjectures/WrittenOnTheWallII/GraphConjecture25.lean
```diff
@@ -0,0 +1,75 @@
...
+WOWII [Conjecture 25](http://cms.uhd.edu/faculty/delavinae/research/wowII/all.html#conj25)
+(status F, disproved):
+
+For a simple connected graph `G`,
+`b(G) ≥ 2 · ⌈(1 + min_v dist_even(v)) / 3⌉`
+where `b(G)` is the size of a largest induced bipartite subgraph and
+`dist_even(v)` is the number of vertices at even distance from `v`.
+-/
+@[category research solved, AMS 5]
+theorem conjecture25 : answer(False) ↔
```
Paul-Lez, line 49: Same `answer(False)` shape issue here: the source is a global false answer over all finite connected graphs, so the `α`/instance quantification should live after `answer(False)` rather than as outer variables of the theorem.

henrykmichalewski:
@Paul-Lez — applied your `answer(False)` quantification convention in [`0418c4d`](https://github.com/henrykmichalewski/formal-conjectures/commit/0418c4d). Both `GraphConjecture24.lean` and `GraphConjecture25.lean` now read

```lean
theorem conjecture24 : answer(False) ↔
    ∀ {α : Type*} [Fintype α] [DecidableEq α] [Nontrivial α],
      ∀ (G : SimpleGraph α) (_ : G.Connected),
        ...
```

with the `α` and instance binders moved inside the `↔ ∀` block so it's a single global statement rather than a type-by-type one. `distEven` lost its dependence on the section variables (now takes `[Fintype α]` as an explicit argument). Builds cleanly.

Paul-Lez:
In WOWII isn't \sigma the second-smallest degree of the degree sequence rather than the 2-domination number?

henrykmichalewski:
@Paul-Lez — you were right and I was wrong. Verified against the WOWII definitions popup (`wowIIdefs.js` defEntry 65):

> "second smallest degree of the degree sequence: order the degree sequence in nondecreasing order d₁ ≤ d₂ ≤ … ≤ dₙ, the second smallest degree of the sequence is the 2nd entry"

Pushed [`510a95d`](https://github.com/henrykmichalewski/formal-conjectures/commit/510a95d) on this branch (and the companion fix [`0d790ef`](https://github.com/henrykmichalewski/formal-conjectures/commit/0d790ef) on `wowii-batches-3-9a` for `GraphConjecture143.lean`):

- Renamed local def `twoDominationNumber` → `secondSmallestDegree`
- Implementation: `(degreeSequence G).getD 1 0` using the upstream `degreeSequence` invariant
- Updated docstring to cite WOWII defEntry 65
- My commit `142c74c`'s confident "Fink–Jacobson 2-domination" claim was unsupported — the WOWII conjectures using σ are not Fink–Jacobson conjectures despite the σ notation overlap.

Both files build cleanly.

mo271 (commented):
This pr contains 6 definitions of `distEven`! Please move to FormalConjecturesForMathlib or re-use if the definition is already there. Same for all other definitions that are not super-specific to one particular conjecture

henrykmichalewski:
@mo271 — done in [`925d89d`](https://github.com/henrykmichalewski/formal-conjectures/commit/925d89d). New shared module `FormalConjecturesForMathlib/Combinatorics/SimpleGraph/DistEven.lean` defines `SimpleGraph.distEven` once. Removed the 6 local copies (in conjectures 24, 25, 63, 85, 96, 189). Added to the `FormalConjecturesForMathlib` umbrella. Build green.

mo271:
> @mo271 — done in [`925d89d`](https://github.com/henrykmichalewski/formal-conjectures/commit/925d89d). New shared module `FormalConjecturesForMathlib/Combinatorics/SimpleGraph/DistEven.lean` defines `SimpleGraph.distEven` once. Removed the 6 local copies (in conjectures 24, 25, 63, 85, 96, 189). Added to the `FormalConjecturesForMathlib` umbrella. Build green.

Same for all other definitions that are not super-specific to one particular conjecture
such as `largestInducedTreeSize`

Kuberwastaken:
@henrykmichalewski @mo271 before this merges, thought I'd add: I found that **Conjectures 63 and 85 are both false**, and one graph kills both. Since this branch already wraps the status-F items (#24/#25) in `answer(False)`, these two could get the same treatment?

**Counterexample:** G = C₅[K₄] - the 5-cycle with each vertex blown up into a K₄, adjacent blobs completely joined. n = 20, m = 110, simple, connected, vertex-transitive.

**Quantities** (per the [WOWII glossary](http://cms.dt.uh.edu/faculty/delavinae/research/wowII/) / `wowIIdefs.js` semantics; `distEven` counting v itself — though both conventions give the same verdicts):
- `tree(G) = f(G) = b(G) = 4`: none of the C(20,5) = 15,504 five-vertex subsets induces a tree / forest / bipartite subgraph — and each property is closed under vertex deletion, so the 5-subset exhaustion is conclusive. The lower bound is an induced P₄: one vertex from each of four consecutive blobs.
- `min_v distEven(v) = 9` (each vertex: 11 vertices at odd distance, 8 at even, plus itself).

So Conjecture 85: `tree(G) ≥ ⌈√(1 + 2·9)⌉ = ⌈√19⌉ = 5 > 4` ✗
and Conjecture 63: `f(G) ≥ ⌈(9 + 4 + 1)/3⌉ = ⌈14/3⌉ = 5 > 4` ✗

**Infinite family, hand proof.** In C₅[K_m]: three chosen vertices in one blob contain a triangle. Two chosen vertices in blob i are adjacent, and any chosen vertex in blob i±1 would complete a triangle — so blobs i±1 are empty, and the only remaining blobs i+2, i+3 have **no edges to blob i**; a connected induced subgraph containing the pair is therefore just the pair (2 vertices). Otherwise every blob holds ≤ 1 chosen vertex, so the induced graph is an induced subgraph of C₅ — and all five blobs would give the cycle C₅ itself. Hence tree = 4 for all m ≥ 2 (the P₄ above); forests are bipartite and the same blob analysis gives b = 4, so f = 4. Meanwhile min distEven = 2m + 1, so RHS₈₅ = ⌈√(4m+3)⌉ and RHS₆₃ = ⌈(2m+6)/3⌉ — both exceed 4 for every m ≥ 4 and grow without bound. At m = 2, 3 both conjectures hold **with equality** (presumably how they survived Graffiti.pc's database); exhaustive `geng` sweeps show no counterexample to #85 on ≤ 9 vertices and none to #63 on ≤ 8.

<details><summary>Self-contained stdlib verifier (~30 s)</summary>

```python
import itertools
V = [(i, a) for i in range(5) for a in range(4)]
E = {frozenset((u, v)) for u in V for v in V if u != v and
     (u[0] == v[0] or (u[0] - v[0]) % 5 in (1, 4))}
adj = {v: {u for u in V if frozenset((u, v)) in E} for v in V}

def bfs(s):
    d, q = {s: 0}, [s]
    while q:
        x = q.pop(0)
        for y in adj[x]:
            if y not in d: d[y] = d[x] + 1; q.append(y)
    return d

min_de = min(sum(1 for t in bfs(v).values() if t % 2 == 0) for v in V)

def ok(S, mode):
    ed = [(u, v) for u, v in itertools.combinations(S, 2) if frozenset((u, v)) in E]
    if mode == "bip":
        col = {}
        for s in S:
            if s in col: continue
            col[s], q = 0, [s]
            while q:
                x = q.pop()
                for y in adj[x] & set(S):
                    if y in col:
                        if col[y] == col[x]: return False
                    else: col[y] = 1 - col[x]; q.append(y)
        return True
    p = {v: v for v in S}
    def f(x):
        while p[x] != x: p[x] = p[p[x]]; x = p[x]
        return x
    for u, v in ed:
        ru, rv = f(u), f(v)
        if ru == rv: return False
        p[ru] = rv
    return mode == "forest" or len(ed) == len(S) - 1

assert len(E) == 110 and min_de == 9
for mode in ("tree", "forest", "bip"):
    assert not any(ok(S, mode) for S in itertools.combinations(V, 5)), mode
print("verified: m=110, tree=f=b=4, min distEven=9 -> C85 and C63 FALSE")
```
</details>

Can also send the `answer(False)` flip against this branch if that's easier as well
or open another PR if necessary

Kuberwastaken:
The complete Lean certificates for the `C₅[K₄]` counterexample are now public and reproducible:

- Tracking/integration issue: #4590
- Repository: https://github.com/Kuberwastaken/wowii-63-85-counterexample
- Conjecture 63 proof: https://github.com/Kuberwastaken/wowii-63-85-counterexample/blob/cba739842ec59adf7426c180009175b31935701d/lean/WOWII63.lean#L184-L195
- Conjecture 85 proof: https://github.com/Kuberwastaken/wowii-63-85-counterexample/blob/cba739842ec59adf7426c180009175b31935701d/lean/WOWII85.lean#L171-L183

Verification:

```sh
lake exe cache get
lake --wfail build
python3 verify_counterexample.py
```

The full Lean build passes with no warnings, `sorryAx`, or custom axioms. The finite certificates use `native_decide`, and the files print the resulting trust envelope (`Lean.ofReduceBool` and `Lean.trustCompiler`).

Because this PR currently owns the statement files for Conjectures 63 and 85, I have not opened a duplicate solved-status PR. Issue #4590 asks whether the proof links/status changes should land here, in a follow-up after this merges, or in a replacement PR.

Kuberwastaken:
Heads-up before these land as `research open`: **WOWII Conjectures 63 and 85 are false** under the imported WOWII semantics.

One counterexample kills both: `G = C₅[K₄]`, the 5-cycle with each vertex blown up into a `K₄` and adjacent blobs completely joined. It has `n = 20`, `|E| = 110`, and is vertex-transitive.

Verified quantities:

- `tree(G) = 4`
- `f(G) = 4`
- `b(G) = 4`
- `min_v distEven(v) = 9` when counting `v` itself at distance `0` (and the verdict also survives the non-self-counting convention)

So Conjecture 85 gives `tree(G) ≥ ceil(sqrt(1 + 2*9)) = ceil(sqrt(19)) = 5`, but `tree(G)=4`.

Conjecture 63 gives `f(G) ≥ ceil((9 + 4 + 1)/3) = 5`, but `f(G)=4`.

There is also a hand-checkable infinite family: for `C₅[K_m]`, `tree = f = b = 4` for every `m ≥ 2`, while the right-hand sides for both conjectures exceed `4` for every `m ≥ 4` and then grow unboundedly. The boundary cases `m=2,3` hold with equality.

Reproducible verification package:

- writeups: https://github.com/Kuberwastaken/breakthroughmaxxing/tree/main/04-wowii/findings
- C85 verifier: https://github.com/Kuberwastaken/breakthroughmaxxing/blob/main/04-wowii/verify_c85.py
- C63 verifier: https://github.com/Kuberwastaken/breakthroughmaxxing/blob/main/04-wowii/verify_c63.py
- stdlib-only inline verifier: https://github.com/Kuberwastaken/breakthroughmaxxing/blob/main/04-wowii/findings/c85-c63-inline-verify.py
- graph6/edge witnesses: https://github.com/Kuberwastaken/breakthroughmaxxing/tree/main/04-wowii/findings

I just reran the three local checks successfully:

```bash
uv run --no-project --with networkx python 04-wowii/verify_c85.py
uv run --no-project --with networkx python 04-wowii/verify_c63.py
python3 04-wowii/findings/c85-c63-inline-verify.py
```

Suggested import status: both `GraphConjecture63.lean` and `GraphConjecture85.lean` should be imported as false / `answer(False)`, not as open. Happy to follow with a PR if useful.

maxtalwar:
I found an explicit counterexample to the `GraphConjecture91.lean` statement in this batch. Let `G` be `K₃,₃ − a₁b₁` together with a `K₅`, joined only by `c₁a₂` and `c₁b₁`. Exhaustive enumeration gives `average λ(G) = 2`, `b(G) = 8`, and `f(G) = 6`, so the proposed bound is `8 ≤ 7`. graph6: `JBza_CB?wF_`.

Full certificate, hand proof, and standard-library verifier:
https://github.com/maxtalwar/wowii91-counterexample/releases/tag/v1.0.0

This is distinct from the [14-vertex K₃,₃–K₈ bridge construction](https://x.com/mattyhempstead/status/2080249051799588979) posted publicly on July 23.

AI assistance disclosure: the search and write-up were developed with ChatGPT assistance. The finite claim is independently reproducible from the construction and exhaustive enumeration of all `2¹¹ = 2048` induced vertex subsets.

Since Conjecture 91 is not yet on `main`, would you prefer to incorporate the negative resolution into this batch, or have me open a focused PR that extracts and resolves #91? I am happy to follow the maintainers' preferred route.

henrykmichalewski:
Thank you @Kuberwastaken and @maxtalwar — great finds, and much appreciated before these landed as `research open`!

I verified both counterexamples independently (re-ran the `C₅[K₄]` exhaustion for #63/#85, and decoded the `JBza_CB?wF_` graph6 witness for #91: `n=11`, `|E|=20`, connected, per-vertex λ = `[2,3,3,2,3,3,2,1,1,1,1]` so `l G = 2` exactly, `b = 8`, `f = 6`, bound `8 ≤ 7` ✗). The branch is updated in a707f488:

- **Conjectures 63, 85, 91** are now recorded as disproved via the `answer(False) ↔ …` pattern of Conjectures 23/24/25, with credit and links to your certificate repos in the module headers.
- Rebased onto current `main` (post-restructure): imports switched to `FormalConjecturesUtil`, and the local copies of `largestInducedTreeSize`/`averageEccentricity` are gone in favour of the shared `FormalConjecturesForMathlib.Combinatorics.SimpleGraph` definitions — this also resolves @mo271's earlier review comment. `distEven` stays as this PR's ForMathlib contribution (no upstream equivalent yet).

@Kuberwastaken @maxtalwar — once the statements merge, formal disproofs replacing the `sorry`s (per your Lean certificates) would make great follow-up PRs from you; deferring to @mo271 on whether `native_decide`-based certificates are acceptable in this repo.

-awaiting-author

Danie-I (commented):
I think usually we put definitions to ForMathlib and then reference them in the conjectures. Thus it is not necessary that the conjectures are self contained files, but maybe @mo271 can confirm.

henrykmichalewski:
Rebased onto current `main` (v4.33.1). Since @Kuberwastaken's #4592 landed Conjectures 63 and 85 with formal-proof links, and `distEven` now lives in `VertexDistance.lean`, this batch defers to upstream for both and drops its duplicate `DistEven.lean` — it is now the remaining eight files (24, 25, 72, 84, 91, 96, 189, 199), all definitions coming from ForMathlib as @Danie-I suggested. Builds clean with all linters locally.

PR #1844, opened 2026-01-22
feat(GreensOpenProblems): 85

jeangud:
# Description

Suppose that $A$ is an open subset of $[0, 1]^2$ with measure $\alpha$. Are there four points in
$A$ determining an axis-parallel rectangle with area $\gt c \alpha^2$?

Closes #1741.

*References*
- [Gr24] [Green, Ben. "100 open problems." (2024).](https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.85)
- [CCW99] Carbery, Anthony, Michael Christ, and James Wright. "Multidimensional van der Corput and sublevel set estimates." Journal of the American Mathematical Society 12.4 (1999): 981-1015 Section 6.
- [Ke00] Keleti, Tamás. "Density and covering properties of intervals of ℝn." Mathematika 47.1-2 (2000): 229-242.
- [KKM02] Katz, Nets Hawk, Elliot Krop, and Mauro Maggioni. "Remarks on the box problem." Mathematical Research Letters 9.4 (2002): 515-520.
- [Mu02] Mubayi, Dhruv. "Some exact results and new asymptotics for hypergraph Turán numbers." Combinatorics, Probability and Computing 11.3 (2002): 299-309 Conjecture 1.4.
- [CPZ20] Conlon, David, Cosmin Pohoata, and Dmitriy Zakharov. "Random multilinear maps and the Erd\H {o} s box problem." arXiv preprint arXiv:2011.09024 (2020).

# Testing
:white_check_mark: Builds successfully.
```shell
$ lake build FormalConjectures/GreensOpenProblems/85.lean 
Build completed successfully.
```

PR #3797, opened 2026-04-17
feat(GreensOpenProblems/57): solution for `z3_functional`

mo271:


Paul-Lez:
Very cool:)

FormalConjectures/GreensOpenProblems/57.lean
```diff
@@ -107,32 +92,56 @@ def functional (a : G → ℂ) (φ : G → ℂ) : ℝ :=
...
+The DeepMind prover agent provided a formal proof, showing that $\frac{183095}{30000}$ separates the
+support functions.
 -/
-@[category research open, AMS 5 11]
+@[category research solved, AMS 5 11, formal_proof using formal_conjectures at
+"https://github.com/mo271/formal-conjectures/blob/f5afe85e1e02611f63c32ae041b33c67b7938cba/FormalConjectures/GreensOpenProblems/57.lean#L1071"]
 theorem green_57.variants.z3_functional :
     let a : ZMod 3 → ℂ := ![(-1 : ℂ), -3, 3]
-    answer(sorry) ↔
+    answer(True) ↔
```
jeangud, line 110: I have seen [a variation](https://github.com/google-deepmind/formal-conjectures/pull/1844#discussion_r2717885373) where we just remove the redundant `answer(True)`. Looks like the repo is not uniform on this, wondering if you have a preference/recommendation?

mo271: preference is to keep `answer(True/False) ↔`. Especially as long the original source keeps the question.
Might be that we were not uniform on this in the beginning -- feel free to fix

Paul-Lez (approved):

PR #3800, opened 2026-04-18
Add D = N open special case for monochromatic quantum graph

MarioKrenn6240:
You have solutions for N=D, for N=4,6,8,10. Strongly indicates to a generalizable pattern.

### DIFF
--- a/FormalConjectures/Paper/MonochromaticQuantumGraph.lean
+++ b/FormalConjectures/Paper/MonochromaticQuantumGraph.lean
@@ -523,8 +523,8 @@ graph equation system over $\mathbb{C}$? -/
 @[category research open, AMS 5 14 81]
 theorem eqSystem_no_solution_even_ge4_d_eq_n_explicit :
     answer(sorry) ↔
-      ∀ N D : Nat, N ≥ 4 → Even N → D = N →
-        ¬ ∃ W : WeightsN N D ℂ, EqSystemN N D W := by
+      ∀ N : Nat, N ≥ 4 → Even N →
+        ¬ ∃ W : WeightsN N N ℂ, EqSystemN N N W := by
   sorry
 
 /-- For all even $N \geq 6$ and $D \geq 3$, does there exist no solution to the monochromatic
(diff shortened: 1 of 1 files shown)


FormalConjectures/Paper/MonochromaticQuantumGraph.lean
```diff
@@ -518,6 +518,15 @@ theorem eqSystem16_no_solution_d3 :
       ¬ ∃ W : WeightsN 16 3 ℂ, EqSystemN 16 3 W := by
   sorry
 
+/-- For all even $N \geq 4$ and $D = N$, does there exist no solution to the monochromatic quantum
+graph equation system over $\mathbb{C}$? -/
+@[category research open, AMS 5 14 81]
+theorem eqSystem_no_solution_even_ge4_d_eq_n_explicit :
+    answer(sorry) ↔
+      ∀ N D : Nat, N ≥ 4 → Even N → D = N →
+        ¬ ∃ W : WeightsN N D ℂ, EqSystemN N D W := by
```
mo271:
```suggestion
      ∀ N : Nat, N ≥ 4 → Even N →
        ¬ ∃ W : WeightsN N N ℂ, EqSystemN N N W := by
```

Is this more elegant?

mo271 (approved):
Thanks!
LGMT

FormalConjectures/Paper/MonochromaticQuantumGraph.lean
MarioKrenn6240: I had this before, but i thought it is more explicit the way i wrote it. Either way is fine for me.

MarioKrenn6240: Actually my statement is wrong as the sum should not be over D, as D is overwritten. I will change to yours (but remove the D in Weights by N.

MarioKrenn6240:
I added the suggestion, and added 3 special cases of N=6,D=5 for the real, integer and {-1,0,1} domain.

mo271 (approved):
Thanks!

PR #3810, opened 2026-04-21
feat(ErdosProblems): 1203

danielchin:
Formalizes Erdos Problem 1203: https://www.erdosproblems.com/1203

### DIFF
--- a/FormalConjectures/ErdosProblems/1203.lean
+++ b/FormalConjectures/ErdosProblems/1203.lean
@@ -1,0 +1,54 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/1203.lean
```diff
@@ -0,0 +1,54 @@
...
+open scoped ArithmeticFunction.omega
+
+namespace Erdos1203
+
+/--
+If $\omega(n)$ counts the number of distinct prime divisors of $n$ then let
+$F(n)=\max_k \omega(n+k)\frac{\log\log k}{\log k}.$
+-/
+noncomputable def F (n : ℕ) : ℝ :=
+  ⨆ k : ℕ, (ω (n + k) : ℝ) * (log (log (k : ℝ)) / log (k : ℝ))
```
mo271, line 36: ⨆ on ℝ could silently return 0 for unbounded sets; however, the set is likely bounded for each fixed n due to growth of $\omega$, so this is probably fine - just to double check

mo271 (approved):
Thanks, lgtm!

FormalConjectures/ErdosProblems/1203.lean
danielchin: Yup, took another look and I believe that it's always bounded for each fixed n, so it should fine.

PR #3811, opened 2026-04-21
Bugeaud Problem 4, move two defs to ForMathlib/Analysis/Equidistribution/

rwst:


Paul-Lez (approved):
Thanks!

PR #3823, opened 2026-04-22
fix(WOWII): Fix conjecture 22

jeangud:
# Description
I might have caught a couple quick fixes in `WrittenOnTheWallII`:
1. The implementation in `GraphConjecture22.lean` stated `b(G) ≥ ⌊α(G) + dist_avg(M, V)⌋`, which does not match 
Conjecture 22 described as `CEIL(2dist(B,V)))≥ CEIL(2distavg(V))` at http://cms.dt.uh.edu/faculty/delavinae/research/wowII/. The statement in the file looks closer (but not exactly) to conjecture 23 `b(G) ≥ ⌊α(G) + dist_avg(M, V) / 2⌋`. @henrykmichalewski this might have been the original intent?
2. The conjecture was actually disproved, so we need to either negate the statement as it is, or add `answer(False)`.
 
**Proposed changes:**
- Rename file, namespace, theorem to conjecture 23
- Added `answer(False)` to show conjecture was actually disproved

# Testing
```shell
$ lake build FormalConjectures/WrittenOnTheWallII/GraphConjecture23.lean 
Build completed successfully (7985 jobs).
```

### DIFF
--- a/FormalConjectures/WrittenOnTheWallII/GraphConjecture23.lean
+++ b/FormalConjectures/WrittenOnTheWallII/GraphConjecture23.lean
@@ -17,30 +17,34 @@ limitations under the License.
 import FormalConjectures.Util.ProblemImports
 
 /-!
-# Written on the Wall II - Conjecture 22
+# Written on the Wall II - Conjecture 23
 
 *Reference:*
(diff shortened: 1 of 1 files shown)


mo271 (approved):
thanks @jeangud !

PR #3751, opened 2026-04-15
Add Erdős Problem 602 (Komjáth's Property B for almost-disjoint families)

henrykmichalewski:
fixes #821 

## Problem

Erdős Problem 602: https://www.erdosproblems.com/602

> Let $(A_i)$ be a family of sets with $|A_i| = \aleph_0$, such that for any $i \neq j$ we have $|A_i \cap A_j|$ finite and $\neq 1$. Is there a 2-colouring of $\bigcup A_i$ such that no $A_i$ is monochromatic?

A problem of Komjáth. The 2-colouring property is sometimes known as Property B.

## What this PR adds

A new file `FormalConjectures/ErdosProblems/602.lean` containing:

- **Main open theorem** `erdos_602` — the original Komjáth question
- **Six proved variants** covering disjoint families, single sets, empty/unique/two-element index families
- **One open sub-variant** for the countable-index case
- **Five sanity-check examples** demonstrating non-vacuity and concrete witnesses
- **One formal disproof** of a natural misformalization (dropping the infiniteness hypothesis), with SafeVerify-style audit checks

The file is self-contained — it imports only `FormalConjectures.Util.ProblemImports`.

## Formulation choices

- **Ground set is arbitrary `Type*`** (not restricted to `ℕ`). The original problem doesn't specify a ground set; using arbitrary types covers families of countable subsets of uncountable spaces.
- **`|A_i| = ℵ_0`** is encoded as `Set.Countable s ∧ s.Infinite`, matching the style in `ErdosProblems/1071.lean` and `143.lean`.
- All variants accept arbitrary ground type via `{α : Type*}` implicit argument.

## Build status

Builds cleanly against the project's pinned Mathlib version. No new `sorry`s outside the explicitly open theorems.

## Acknowledgement

This formalization was developed with assistance from Claude (Anthropic), with iterative human review and verification.

PR #3837, opened 2026-04-25
fix category for `erdos_602.variants.countable_index`

mo271:


mo271:
@henrykmichalewski can you have a look here, this was added recently: #3751

Paul-Lez (approved):

PR #3838, opened 2026-04-25
feat(Wikipedia): Zaremba's conjecture

AlexKontorovich:
Zaremba's conjecture (wikipedia)


FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,73 @@
...
+@[category test, AMS 11]
+example : minmaxPartialQuotient 121 = 2 := by native_decide
+
+/--
+Zaremba's conjecture: there is an absolute constant `A` such that for every
+denominator `d`, there is a numerator `a` coprime to `d`, for which every partial
+quotient in the continued fraction of `a / d` is at most `A`.
+-/
+@[category research open, AMS 11]
+theorem zaremba_conjecture : ∃ A, ∀ d, minmaxPartialQuotient d ≤ A := by
```
mo271:
```suggestion
theorem zaremba_conjecture : ∃ (A : ℕ), ∀ d, minmaxPartialQuotient d ≤ A := by
```

Otherwise the type of `A` will be inferred to be `WithTop` and proving will become easy:

```lean
theorem zaremba_conjecture : ∃ A, ∀ d, minmaxPartialQuotient d ≤ A :=
  ⟨⊤, fun _ => le_top⟩
```

mo271 (commented):
Thanks! 
Great conjectures and nice contribution!

I'm not super familiar with continued fractions in Mathlib, but potentially we could re-use some of that instead of `partialQuotients` and Euclidian CF algorithm.(But I havn't checked what is there and what you have seems to make perfect sense and is well tests...
I assume mathlib's stuff doesn't work with `native_decide`? 

In any case, the main conjecture has a silly misformalisation, I tried to fix is but then could also prove 
```lean
theorem zaremba_conjecture' : ¬ (∃ A : ℕ, ∀ d, minmaxPartialQuotient d ≤ A) := by
  push_neg
  aesop( add safe forward True)
  delta minmaxPartialQuotient
  delta ZarembaConjecture.maxPartialQuotient ZarembaConjecture.coprimeResidues
  exact⟨0,by simp_all⟩
```

AlexKontorovich:
Argh the `Finset.min` adds `WithTop`. (And then when `d=0`, the min is `Top`, so the conjecture is false as stated.) This should fix it

AlexKontorovich:
As for Mathlib, neither the `Nat.gcd` nor `GenContFrac` API do what's needed here without significant patching back and forth, so I decided to do it by hand (which is already very simple).

AlexKontorovich:
And the `native_decide` was just to run some simple tests. We can remove those if you prefer

FormalConjectures/Wikipedia/ZarembaConjecture.lean
eric-wieser: I guess you could also write `∀ d, minmaxPartialQuotient d < ⊤`

AlexKontorovich: I guess we could write `(minmaxPartialQuotient '' univ).max < ⊤` (not `∀ d,`... because the whole question is about uniformity...). Let me know if you prefer that formulation? (I guess it's a little nicer, not to introduce `A` for no good reason?)

AlexKontorovich: I'll try a slightly different version, let me know what you think of this

eric-wieser: Am I right in thinking that there is a stronger statement which declares the lowest possible such A?

FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,101 @@
...
+-- Move to MATHLIB
+def Finset.max! {α : Type*} [LinearOrder α] [Inhabited α] (s : Finset α) : α :=
+  s.max.unbotD default
+
+def Finset.maxD {α : Type*} [LinearOrder α] (s : Finset α) (d : α) : α :=
+  s.max.unbotD d
+
+-- move to MATHLIB
+def Function.sup {α : Type*} {β : Type*} [SupSet (WithTop β)] (f : α → β) : WithTop β :=
+  ⨆ a, (f a : WithTop β)
```
mo271: can you move these to a suitable file in our `FormalConjecturesForMathlib`? 

This will make it easier not to forget to upstream it.

mo271 (approved):
LGTM now, not sure if @eric-wieser wants to chime in still

FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,101 @@
...
+  | a, d + 1 => a / (d + 1) :: partialQuotients (d + 1) (a % (d + 1))
+termination_by _ d => d
+decreasing_by exact Nat.mod_lt _ (Nat.succ_pos _)
+
+/-- For `333/106 = [3; 7, 15]`. -/
+@[category test, AMS 11]
+example : partialQuotients 333 106 = [3, 7, 15] := by native_decide
+
+/-- The maximum partial quotient of `a / d`, taking the value `0` when `d = 0`. -/
+def maxPartialQuotient (a d : ℕ) : ℕ := (partialQuotients a d).toFinset.max!
```
eric-wieser: Isn't this
```suggestion
def maxPartialQuotient (a d : ℕ) : ℕ := (partialQuotients a d).toFinset.sup id
```
I don't think we need a new definition here

FormalConjectures/Wikipedia/ZarembaConjecture.lean
eric-wieser: I'm pretty skeptical of this definition, and don't see it as something we would want in mathlib

FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,101 @@
...
+example : coprimeResidues 1 = {0} := by native_decide
+
+/-- Edge case, `d = 0`. -/
+@[category test, AMS 11]
+example : coprimeResidues 0 = {} := by native_decide
+
+/-- The least possible value, over positive coprime numerators `a < d`, of the maximum partial
+quotient in the continued fraction of `a / d`. -/
+def minmaxPartialQuotient (d : ℕ) : ℕ :=
+  ((coprimeResidues d).image fun a ↦ maxPartialQuotient a d).min.untopD 0
```
eric-wieser, line 64: Why the untop? Do things go wrong if we leave this ENat-valued?

AlexKontorovich: Yeah, this is the issue with `d = 0`, where, without the `untopD`, this would be `Top`. So to be able to state the conjecture "cleanly" as a max-min-max (which is kind of nice, I think -- though I understand your point elsewhere that this is a difference between the "style" of formal-conjectures vs Mathlib...), I thought it would be nice to "junk" value an empty min to 0. (And the suggested API at the top junk values to `default` if the Type is `Inhabited`)

FormalConjectures/Wikipedia/ZarembaConjecture.lean
AlexKontorovich: I was just thinking it'd be nice to be able to write `f.sup` for a function, even if `β` doesn't have a `Top`? This makes the statement of the main conjecture nice and clean, `minmaxPartialQuotient.sup < ⊤`? Happy to hear suggestions for a nicer way to do it?

eric-wieser: Does  `⨆ a, minmaxPartialQuotient a < (⊤ : ENat)` do the right thing?

AlexKontorovich: yeah, it certainly works; I just thought it looked a bit ugly, and it might be nice to have API to express `Function.sup`?

AlexKontorovich:
Ok I removed the for-Mathlib bits. Look ok? Thanks!

AlexKontorovich:
Just FYI, a proof of the conjecture has now been announced: https://arxiv.org/pdf/2605.02518. (Hence my desire to have the statement formalized...)

eric-wieser:
Repeating my question lost to the github UI from earlier:

> Am I right in thinking that there is a stronger statement which declares the lowest possible such A?

AlexKontorovich:
> Am I right in thinking that there is a stronger statement which declares the lowest possible such A?

Whoops sorry, missed that! Yes indeed. I added the strong form (`A = 5`), as well as Hensley's strongest "eventual" version with `A = 2`. Thanks!

mo271:
> Just FYI, a proof of the conjecture has now been announced: https://arxiv.org/pdf/2605.02518. (Hence my desire to have the statement formalized...)
Great!

Then perhaps a good idea to change the category from `research open` to `research solved` mentioning authors and the arxiv link in the docstring

AlexKontorovich:
> Then perhaps a good idea to change the category from `research open` to `research solved` mentioning authors and the arxiv link in the docstring

Well... it hasn't been fully refereed yet. So I don't know if "solved" is the right term, maybe "claimed"?

FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,109 @@
...
+theorem zaremba_conjecture : maxminmaxPartialQuotient < ⊤ := by
+  sorry
+
+/--
+Zaremba's conjecture in "strong" form: for every
+denominator `d`, there is a numerator `a` coprime to `d`, for which every partial
+quotient in the continued fraction of `a / d` is at most `5`.
+-/
+@[category research open, AMS 11]
+theorem zaremba_conjecture.variants.constant_five : maxminmaxPartialQuotient ≤ 5 := by
```
eric-wieser: The bound is tight, right? So perhaps
```suggestion
theorem zaremba_conjecture.variants.exact : maxminmaxPartialQuotient = answer(5) := by
```
or even something like
```suggestion
theorem zaremba_conjecture.variants.exact : IsLeast {minmaxPartialQuotient d | d > 0} answer(5) := by
```

mo271:
> > Then perhaps a good idea to change the category from `research open` to `research solved` mentioning authors and the arxiv link in the docstring
> 
> Well... it hasn't been fully refereed yet. So I don't know if "solved" is the right term, maybe "claimed"?

Unfortunately we don't have a category in between `research solved` and `research open` and here `open` seems more wrong. But definitely good to explain in the docstring that this result is still under review! We can change back to `open` it the proof turns out to have a bug. Unless you suspect there is a problem with the proof.

FormalConjectures/Wikipedia/ZarembaConjecture.lean
AlexKontorovich: sure, will do. Thanks!

AlexKontorovich:
> Unfortunately we don't have a category in between `research solved` and `research open` and here `open` seems more wrong. But definitely good to explain in the docstring that this result is still under review! We can change back to `open` it the proof turns out to have a bug. Unless you suspect there is a problem with the proof.

Got it. I set it to `solved` and added the reference. I don't suspect a problem with the proof; just a lot of heavy, technical machinery that needs to be checked carefully... Thanks!

FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,111 @@
...
+  sorry
+
+
+/--
+Hensley's conjecture in "strong" form: for every sufficiently large
+denominator `d`, there is a numerator `a` coprime to `d`, for which every partial
+quotient in the continued fraction of `a / d` is at most `2`.
+-/
+@[category research open, AMS 11]
+theorem hensley_conjecture : ∃ D : ℕ, ⨆ d ≥ D, minmaxPartialQuotient d = 2 := by
```
eric-wieser:
```suggestion
theorem hensley_conjecture : ∃ D : ℕ, (⨆ d ≥ D, minmaxPartialQuotient d) = 2 := by
```
otherwise I fear this takes the supremum (exists) of the proposition

FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,111 @@
...
+theorem zaremba_conjecture : maxminmaxPartialQuotient < ⊤ := by
+  sorry
+
+/--
+Zaremba's conjecture in "strong" form: for every
+denominator `d`, there is a numerator `a` coprime to `d`, for which every partial
+quotient in the continued fraction of `a / d` is at most `5`.
+-/
+@[category research open, AMS 11]
+theorem zaremba_conjecture.variants.exact : maxminmaxPartialQuotient = 5 := by
```
eric-wieser:
```suggestion
theorem zaremba_conjecture.variants.exact : maxminmaxPartialQuotient = answer(5) := by
```

FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,111 @@
...
+Shkredov https://arxiv.org/abs/2603.14116.
+-/
+@[category research solved, AMS 11]
+theorem zaremba_conjecture : maxminmaxPartialQuotient < ⊤ := by
+  sorry
+
+/--
+Zaremba's conjecture in "strong" form: for every
+denominator `d`, there is a numerator `a` coprime to `d`, for which every partial
+quotient in the continued fraction of `a / d` is at most `5`.
```
eric-wieser:
```suggestion
Zaremba's conjecture in "strong" form: _what_ is the absolute constant `A` such that for every
denominator `d`, there is a numerator `a` coprime to `d`, for which every partial
quotient in the continued fraction of `a / d` is at most `A`.
```

AlexKontorovich:
Just wanted to ping this - are you waiting on me for anything? Thanks!

mo271:
from my point of view this is good to go, do you agree @eric-wieser ?

FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,111 @@
...
+
+namespace ZarembaConjecture
+
+/-- The finite list of partial quotients in the simple continued fraction expansion of a
+positive rational `a / d`. -/
+def partialQuotients : ℕ → ℕ → List ℕ
+  | _, 0 => []
+  | a, d + 1 => a / (d + 1) :: partialQuotients (d + 1) (a % (d + 1))
+termination_by _ d => d
+decreasing_by exact Nat.mod_lt _ (Nat.succ_pos _)
```
YaelDillies, line 33: Is that not basically `GenContFract.convs`? Could you use that instead of rolling your own? We seem to be missing a statement of the form `(GenContFract.convs q).TerminatedAt q.den`, which you could then use to extract a list from the stream

FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,111 @@
...
+@[category test, AMS 11]
+example : coprimeResidues 1 = {0} := by native_decide
+
+/-- Edge case, `d = 0`. -/
+@[category test, AMS 11]
+example : coprimeResidues 0 = {} := by native_decide
+
+/-- The least possible value, over positive coprime numerators `a < d`, of the maximum partial
+quotient in the continued fraction of `a / d`. -/
+def minmaxPartialQuotient (d : ℕ) : ℕ :=
```
YaelDillies, line 63:
```suggestion
def minMaxPartialQuotient (d : ℕ) : ℕ :=
```

FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,111 @@
...
+example : minmaxPartialQuotient 0 = 0 := by native_decide
+
+/-- The best possible maximum partial quotient among coprime numerators for
+denominator `121` is `2`. -/
+@[category test, AMS 11]
+example : minmaxPartialQuotient 121 = 2 := by native_decide
+
+/-- The least possible value, over positive coprime numerators `a < d`, of the maximum partial
+quotient in the continued fraction of `a / d`. -/
+noncomputable def maxminmaxPartialQuotient : ℕ∞ :=
```
YaelDillies, line 78:
```suggestion
noncomputable def maxMinMaxPartialQuotient : ℕ∞ :=
```

FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,111 @@
...
+theorem zaremba_conjecture : maxminmaxPartialQuotient < ⊤ := by
+  sorry
+
+/--
+Zaremba's conjecture in "strong" form: _what_ is the absolute constant `A` such that for every
+denominator `d`, there is a numerator `a` coprime to `d`, for which every partial
+quotient in the continued fraction of `a / d` is at most `A`.
+-/
+@[category research open, AMS 11]
+theorem zaremba_conjecture.variants.exact : maxminmaxPartialQuotient = answer(5) := by
```
YaelDillies, line 98: I don't see `5` in the docstring. Can you make sure it is somewhere?

FormalConjectures/Wikipedia/ZarembaConjecture.lean
```diff
@@ -0,0 +1,111 @@
...
+  sorry
+
+
+/--
+Hensley's conjecture in "strong" form: for every sufficiently large
+denominator `d`, there is a numerator `a` coprime to `d`, for which every partial
+quotient in the continued fraction of `a / d` is at most `2`.
+-/
+@[category research open, AMS 11]
+theorem hensley_conjecture : ∃ D : ℕ, (⨆ d ≥ D, minmaxPartialQuotient d) = 2 := by
```
YaelDillies, line 108: The docstring says "at most 2" and the formal statement "exactly 2". I assume it is correct because the quantity is obviously bounded below by 2, but could you make that explicit?

PR #3840, opened 2026-04-26
feat(ErdosProblems): 1200

### LINKED ISSUE #3819: Erdős Problem 1200

https://www.erdosproblems.com/1200

There exists a constant $C$ such that for all large $x$ there is a collection of primes $p_1 < \dots < p_k < x$ with $\sum \frac{1}{p_i} < C$ together with a system of congruences $a_i \pmod{p_i}$ such that every integer $n < x$ satisfies at least one of these congruences.

Status: open

* ams-11

### Choose either option
- [x] I plan on adding this Erdős problem to the repository
- [ ] This issue is up for grabs: I would like to see this Erdős problem added by somebody else

zachary-feng:
Fixes #3819

Any feedback is greatly appreciated!


FormalConjectures/ErdosProblems/1200.lean
```diff
@@ -0,0 +1,104 @@
...
+  (fun (n : ℕ) ↦ (1 : ℝ)) =O[Filter.atTop] epsilon_function
+  := by sorry
+
+/--
+In [ErRu80] this is asked as a question: if $p_1 < \dots < p_k < x$ are primes with
+$\sum \frac{1}{p_i} \leq C$ and $a_i \pmod {p_i}$ are any residue classes then
+must there always be $\gg_C x$ many integers $n < x$ avoiding all of them?
+
+Of course if the answer is yes then this disproves [erdosproblems.com/1200].
+-/
```
Paul-Lez, line 68: Should this be asymptotic in `x`? Since the source says `≫_C x`, I would have expected an eventual statement, perhaps using `∀ᶠ x in Filter.atTop`; same question for the `ErRu80_theorem` variant below.

FormalConjectures/ErdosProblems/1200.lean
```diff
@@ -0,0 +1,104 @@
...
+
+@[category research open, AMS 11]
+theorem erdos_1200 :
+  ∃ (C : ℝ), (C > 0) ∧ ∀ᶠ (x : ℝ) in Filter.atTop,
+      ∃ (S : Finset ℕ), ∃ (a : ℕ → ℕ),
+        (∀ p ∈ S, Nat.Prime p) ∧
+        (∀ p ∈ S, p < x) ∧
+        (∑ p ∈ S, (1 : ℝ) / p < C) ∧
+        (∀ (n : ℕ), (1 ≤ n) ∧ (n < x) → ∃ p ∈ S, a p ≡ n [MOD p])
+  := by sorry
```
Paul-Lez: Since #3929 is adding the #688 formalisation too, could this import and reuse the canonical #688 predicate/epsilon function from there instead of duplicating them here? That should make the connection between the two problem files clearer.

FormalConjectures/ErdosProblems/1200.lean
```diff
@@ -0,0 +1,104 @@
...
+- [ErRu80] Erdős, P. and Ruzsa, I. Z., _On the small sieve. I. Sifting by primes_. J. Number Theory (1980), 385--394.
+-/
+
+open Classical
+
+namespace Erdos1200
+
+/--
+There exists a constant $C$ such that for all large $x$ there is a collection of primes
+$p_1 < \dots < p_k < x$ with $\sum \frac{1}{p_i} < C$ together with a system of congruences
```
Paul-Lez, line 35: Let’s use the `answer(sorry)` convention here since this is phrased as an open yes/no question. So I think this should start with something like

```lean
theorem erdos_1200 : answer(sorry) ↔
  ∃ (C : ℝ), (C > 0) ∧ ∀ᶠ (x : ℝ) in Filter.atTop,
    ...
```

rather than stating the conjecture as a bare proposition.

FormalConjectures/ErdosProblems/1200.lean
zachary-feng: Thanks for catching this, I completely missed it - will fix!

FormalConjectures/ErdosProblems/1200.lean
zachary-feng: Thanks for the suggestion, will fix pending #3929 (since I'll need to reference `688.lean`)

FormalConjectures/ErdosProblems/1200.lean
zachary-feng: I'm happy to use the `answer(sorry)` convention here, but I'm a bit unclear on the reason. My understanding was that they're simply asserting "There exists..." rather than posing the question "Does there exist..."?

zachary-feng:
@Paul-Lez Fixed all comments, except for the concern quoted below

> I'm happy to use the `answer(sorry)` convention here, but I'm a bit unclear on the reason. My understanding was that they're simply asserting "There exists..." rather than posing the question "Does there exist..."?

zachary-feng:
-awaiting-author

FormalConjectures/ErdosProblems/1200.lean
```diff
@@ -0,0 +1,96 @@
...
+-/
+@[category research open, AMS 11]
+theorem erdos_1200.variants.erdos_ruzsa_question : answer(sorry) ↔
+    ∀ C : ℝ, C > 0 → ∃ c : ℝ, c > 0 ∧
+      ∀ᶠ (x : ℝ) in atTop, ∀ S : Finset ℕ, ∀ a : ℕ → ℕ,
+        (∀ p ∈ S, p.Prime) ∧
+        (∀ p ∈ S, p < x) ∧
+        (∑ p ∈ S, (1 : ℝ) / p ≤ C) ∧
+        (filter (fun (m : ℕ) ↦ AvoidsCongruences S a m) <|
+          Icc (1 : ℕ) (Int.toNat <| Int.floor x)).card ≥ c * x := by
```
mo271, line 77:
```suggestion
      ∀ᶠ (x : ℝ) in atTop, ∀ S : Finset ℕ, ∀ a : ℕ → ℕ,
        (∀ p ∈ S, p.Prime) →
        (∀ p ∈ S, p < x) →
        (∑ p ∈ S, (1 : ℝ) / p ≤ C) →
        (filter (fun (m : ℕ) ↦ AvoidsCongruences S a m) <|
          Icc (1 : ℕ) (Int.toNat <| Int.floor x)).card ≥ c * x
```

is this what is intended here?

mo271 (commented):
Thanks, looks mostly good, except for a potential bug in `erdos_ruzsa_question`

FormalConjectures/ErdosProblems/1200.lean
zachary-feng: Fixed, thanks!

zachary-feng:
-awaiting-author

zachary-feng:
@mo271 I've applied your bug fix for `erdos_ruzsa_question`. Would you mind taking a look again? Thanks!

FormalConjectures/ErdosProblems/1200.lean
```diff
@@ -0,0 +1,96 @@
...
+$a_i \pmod{p_i}$ such that every integer $n < x$ satisfies at least one of these congruences.
+-/
+@[category research open, AMS 11]
+theorem erdos_1200 :
+    ∃ C : ℝ, C > 0 ∧ ∀ᶠ (x : ℝ) in atTop,
+        ∃ S : Finset ℕ, ∃ a : ℕ → ℕ,
+          (∀ p ∈ S, p.Prime) ∧
+          (∀ p ∈ S, p < x) ∧
+          (∑ p ∈ S, (1 : ℝ) / p < C) ∧
+          (∀ n : ℕ, 1 ≤ n ∧ n < x → ∃ p ∈ S, a p ≡ n [MOD p]) := by
```
YaelDillies: This doesn't matter, but why exclude `0`?

FormalConjectures/ErdosProblems/1200.lean
```diff
@@ -0,0 +1,96 @@
...
+theorem erdos_1200.variants.implied_by_erdos_688_variant :
+    (fun (n : ℕ) ↦ (1 : ℝ)) =O[atTop] Erdos688.epsilonFunction := by
+  sorry
+
+/--
+A predicate which decides whether (n : ℕ) avoids the congruences prescribed by
+(a : ℕ → ℕ) and (S : Finset ℕ), which will be used in erdos_rusza_question below.
+-/
+def AvoidsCongruences (S : Finset ℕ) (a : ℕ → ℕ) (n : ℕ) : Prop :=
+  ∀ p ∈ S, ¬(a p ≡ n [MOD p])
```
YaelDillies: Why not just inline that?

FormalConjectures/ErdosProblems/1200.lean
```diff
@@ -0,0 +1,96 @@
...
+Erdős and Ruzsa [ErRu80] proved that for any $C > 0$ there exists a set of primes $P$
+such that $\sum_{p ∈ P} \frac{1}{p} \leq C$ and the number of integers $n \leq x$ divisible
+by at least one $p \in P$ is $\gg_C x$.
+-/
+@[category research solved, AMS 11]
+theorem erdos_1200.variants.erdos_ruzsa_theorem :
+    ∀ C : ℝ, C > 0 → ∃ c : ℝ, ∃ P : Set ℕ, c > 0 ∧
+      (∀ p ∈ P, p.Prime) ∧
+      (∑' p : ℕ, Set.indicator P (fun (k : ℕ) ↦ (1 : ℝ) / k) p > 0) ∧
+      (∑' p : ℕ, Set.indicator P (fun (k : ℕ) ↦ (1 : ℝ) / k) p ≤ C) ∧
```
YaelDillies: Is this condition not just saying that $\sum_{p ∈ P} \frac{1}{p}$ exists? Why not rewrite the condition below as a sum over `ENNReal` to avoid this oddity?
```suggestion
      (∑' p : ℕ, Set.indicator P (fun (k : ℕ) ↦ (1 : \ENNReal) / k) p ≤ C) ∧
```

FormalConjectures/ErdosProblems/1200.lean
zachary-feng: I think I read integer to mean positive integer, but you're right, it doesn't matter!
(I removed `1 ≤ n`)

FormalConjectures/ErdosProblems/1200.lean
zachary-feng: Done

FormalConjectures/ErdosProblems/1200.lean
zachary-feng: Thanks for the suggestion, fixed!

zachary-feng:
-awaiting-author

zachary-feng:
@YaelDillies Thanks for the PR review!
I've addressed your comments, would you mind taking a look?

FormalConjectures/ErdosProblems/1200.lean
```diff
@@ -0,0 +1,95 @@
...
+by at least one $p \in P$ is $\gg_C x$.
+-/
+@[category research solved, AMS 11]
+theorem erdos_1200.variants.erdos_ruzsa_theorem :
+    ∀ C : ℝ, C > 0 → ∃ c : ℝ, ∃ P : Set ℕ, c > 0 ∧
+      (∀ p ∈ P, p.Prime) ∧
+      (∑' p : ℕ, Set.indicator P (fun (k : ℕ) ↦ (1 : ℝ≥0∞) / k) p ≤ ENNReal.ofReal C) ∧
+      ∀ᶠ (x : ℝ) in atTop,
+        (filter (fun (m : ℕ) ↦ ∃ p ∈ P, p ∣ m) <|
+          Icc (1 : ℕ) (Int.toNat <| Int.floor x)).card ≥ c * x := by
```
YaelDillies, line 92:
```suggestion
          Icc (1 : ℕ) \lfloor x\rfloor\_+).card ≥ c * x := by
```
Same above

PR #3882, opened 2026-04-28
Resolve Erdős 389 problem by confirming existence of k

Etalides:
Marcar el Problema de Erdős 389 como resuelto usando prueba externa (Etalides Bonillo, 2026)


mo271:
The reference for the proof and getting confirmation, for instance in the comment section of erdosproblems.com for that problem or providing a formal proof would be better than just having TODO there. Converting to draft for now

Deicyde (changes requested):
The solved status is still unsupported: `formal_proof` points to `TO_BE_SUPPLIED`, and the description explicitly says no formal proof is provided. Please supply a checkable proof reference and use the matching proof kind, or retain `research open`; an informal AI validation is not an `other_system` certificate. If an affirmative solution is established, also replace `answer(sorry)` with `answer(True)`.

PR #3894, opened 2026-04-28
fix(Arxiv): `MaximumAngle`

smmercuri:
Remove file since the original intention was unclear and has since been solved.


Paul-Lez (approved):

mo271:
there has been a major update on this paper -- also changing the status of the (intended) conjecture here of this to solved:

We could also just drop the file!

smmercuri:
> there has been a major update on this paper -- also changing the status of the (intended) conjecture here of this to solved:
> 
> We could also just drop the file!

Do you know where this conjecture is proved in the [updated paper](https://arxiv.org/pdf/1506.05785v5)? I can't seem to find it.

There are two new conjectures (Conjecture 3.3 and 3.4) in the updated paper we could replace it with. These look a bit more involved to formalise than the old one so I propose we add a TODO for now?

mo271 (approved):

PR #3895, opened 2026-04-28
fix(Wikipedia): `DedekindNumber`

### LINKED ISSUE #3490: Dedekind Numbers Closed Form
There is a closed form inefficient formula for the Dedekind Numbers https://en.wikipedia.org/wiki/Dedekind_number#Summation_formula. The formalization / docstring in https://github.com/google-deepmind/formal-conjectures/blob/18c761321903034f6272faa3a841e2bd1261f0d0/FormalConjectures/Wikipedia/DedekindNumber.lean#L231-L236 says the opposite. 

### Suggested fix
Change the docstring and / or the formalization.

### Additional context
This makes me wonder if the Dedekind Numbers aren't just trivially solvable anyway, atleast when formalized with answer(sorry), as one can just provide the arithmetic formula expression for $n=10$ with a proof that it equals $M(10)$. Or how does one force that the provided term is a "raw" number?

### Choose either option

smmercuri:
Fixes #3490

### DIFF
--- a/FormalConjectures/Wikipedia/DedekindNumber.lean
+++ b/FormalConjectures/Wikipedia/DedekindNumber.lean
@@ -228,8 +228,39 @@ theorem M_eq_M' : M  = M' := by
   ext n
   exact Fintype.card_congr (equivMonotoneSperner n)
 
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks

PR #3893, opened 2026-04-28
chore: link formal proof of erdos-1148

Paul-Lez:
Fixes #3595

PR #3897, opened 2026-04-29
fix(ErdosProblems): 1148

Smetalo:
Fixed formal proof link


mo271:
Thanks, potentially we can avoid those live lean links altogether and instead linking to a place that has them (this would allow that place to update), see
https://github.com/google-deepmind/formal-conjectures/pull/3893#discussion_r3159449947

mo271:
Also: and this is perhaps bigger issue that should be addressed more systematically, because it also affects some other problems: this is *not* a complete proof, but just a reduction to some other statement:
Let's remove the `formal_proof` here: https://www.erdosproblems.com/forum/thread/1148#post-4862

And also in other statements were the formalistaions are not complete!

Smetalo:
Since systematic changes are needed, it is best for one of the maintainers to make the necessary changes. Since those changes are out of scope for this PR, should I close it?

mo271:
Let's keep it open, I will just push to it.

and thanks, @Smetalo for the many comments on duplicate pr etc, btw...

mo271 (approved):
LGTM now

mo271:
Thanks again @Smetalo, keep up the good work!

PR #3904, opened 2026-04-30
feat: Green's Open Problem 54

### LINKED ISSUE #1659: Green's Open Problems #54
### What is the conjecture
Let $K \subset \mathbb{R}^N$ be a balanced compact set (that is, $\lambda K \subset K$ when $|\lambda| \leq 1$) and suppose that the normalised Gaussian measure $\gamma_\infty(K)$ is at least 0.99. Does $10K$ (say) contain a compact convex set $C$ with $\gamma_\infty(C) \geq 0.01$?

* ams-52
* ams-60

### Choose either option

Sfgangloff:
### Add formalization of Ben Green Open Problem 54

This PR adds a Lean formalization of Ben Green’s Open Problem 54, as stated in his list of open problems in additive combinatorics.

Changes:

FormalConjectures/GreensOpenProblems/54.lean

Closes #1659

### Original problem statement

Let $K \subset \mathbb{R}^N$ be a balanced compact set (that is,  $\lambda K \subset K$ when $|\lambda| \le 1$) and suppose that the normalised Gaussian measure  $\gamma_\infty(K)$ is at least 0.99. Does $10K$ (say) contain a compact convex set $C$ with $\gamma_\infty(C) \ge 0.01$?


FormalConjectures/GreensOpenProblems/54.lean
```diff
@@ -0,0 +1,46 @@
...
+  Measure.pi (fun _ => gaussianReal 0 1)
+
+/--
+Let $K \subset \mathbb{R}^n$ be a balanced compact set (that is, $\lambda K \subseteq K$ whenever
+$|\lambda| \leq 1$) and suppose that the normalised Gaussian measure $\gamma_n(K) \geq 0.99$.
+Does $10K$ contain a compact convex set $C$ with $\gamma_n(C) \geq 0.01$?
+-/
+@[category research open, AMS 46 52 60]
+theorem green_54 :
+    answer(sorry) ↔ ∀ n : ℕ, ∀ K : Set (Fin n → ℝ), IsCompact K → Balanced ℝ K → ENNReal.ofReal 0.99 ≤ gaussianMeasure n K → ∃ C : Set (Fin n → ℝ), IsCompact C ∧ Convex ℝ C ∧ C ⊆ (10 : ℝ) • K ∧
```
mo271: please make sure to observe the 100-chars line length limit

mo271: use `(0.99 : ℝ≥0∞)` instead of `ENNReal.ofReal 0.99`. This works after adding `ENNReal` to the `open scoped` above. Similar in similar places

FormalConjectures/GreensOpenProblems/54.lean
mo271, line 1: I'm a bit unsure about the following:

The source asks about $\gamma_\infty$ on an infinite-dimensional space (e.g. $\mathbb{R}^\mathbb{N}$), not about $\gamma_n$ on $\mathbb{R}^n$ for all finite $n$. The finite-dimensional reformulation may or may not be equivalent, but this should at minimum be documented. In Talagrand's framework, $K$ can be an infinite-dimensional set that doesn't factor through any finite-dimensional projection.

If the intent is to use the fact that $\gamma_\infty$ is the projective limit of $\gamma_n$, this equivalence should be stated or referenced, not silently assumed.

Also: perhaps good to add the $K=2$ as a solved variant and attirbute Talagrand (perhaps also add some reference, currently he is not mentioned at all in the file)

mo271 (commented):
Thanks! 
A few initial remarks..

FormalConjectures/GreensOpenProblems/54.lean
Sfgangloff: I think you are right, I confused $\mathbb{N}$ with integer N, I am correcting that. Also adding reference to Talagrand and the solved case.

Sfgangloff:
-awaiting-author

FormalConjectures/GreensOpenProblems/54.lean
```diff
@@ -0,0 +1,60 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-
```
mo271:
```suggestion
/-!
```

mo271 (approved):
Thanks again, LGTM now!

PR #3915, opened 2026-05-01
chore(Arxiv): update status of FirstProof problem 4 and link to formal proof

Paul-Lez:


mo271:
proposing #3922 instead

PR #3928, opened 2026-05-02
Added mod 8 ES solver

Cloudhabil:


mo271:
Please re-open when this described more clearly what is done and likely you only want to edit one file?

PR #3935, opened 2026-05-04
feat(Erdos/30): add Sidon bound variants and Singer witnesses

bengoechea:
Adds elementary Sidon-set bounds and finite examples around Erdős problem 30. The original `erdos_30` statement remains `category research open`, with its existing `answer(sorry)` and proof placeholder unchanged. This contribution does not solve the conjecture or claim an improved asymptotic bound.

**Scope**

- `FormalConjectures/ErdosProblems/30.lean`: positive-difference counting, cardinality and range of the pointwise sumset `A + A`, the elementary sum-counting bound, a weak square-root corollary, a variance identity, six concrete natural-number Sidon witnesses, and two explicitly conditional consequences.
- `FormalConjecturesForMathlib/Combinatorics/Basic.lean`: four supporting lemmas for ordered Sidon pairs, injectivity of positive differences, and triangular pair counts.

**Assumptions and source alignment**

- `BFRCoreBound` is an assumed rounded numerical inequality, including an assumed `N ≥ 10^12` threshold. [Balogh–Füredi–Roy, Theorem 1.1](https://arxiv.org/html/2103.15850) motivates the coefficient 0.998, but does not establish this threshold or rounded premise. `bfr` proves only an arithmetic consequence of the explicit assumption.
- `singer` assumes `SingerSidonExists` and returns a witness in `Finset.range (q*q+q+1)`, meaning `{0,…,q²+q}`. It does not establish a lower bound for `h`, which uses `{1,…,N}`. The general construction is not proved here.
- `lindstrom_weak` is the elementary `⌊√(2N)⌋ + 1` corollary, not Lindström’s sharper classical bound. Unsupported external compiled-proof claims have been removed.
- Pointwise sum notation and the `distinct_sums_card` proof are corrected. The elementary sum-counting result and conditional arithmetic consequences use `category textbook`; the sum-range comment agrees with the theorem's interval `{0,…,N}`.

**Verification**

- `lake --wfail build 'FormalConjectures.ErdosProblems.«30»'` passed on Lean `4.33.1` and Mathlib `0df444a360eaa60ab8c11dca51a86af692955474` (the pull request’s pinned environment). This is a scoped module build, not a full-project build claim.
- The axiom audit at `8f49d159` checked all 16 problem declarations and four shared lemmas: only the open conjecture depends on `sorryAx`; the nine other non-computational variants and four shared lemmas use only `propext`, `Classical.choice`, and `Quot.sound`. The six finite witnesses additionally depend on their generated native-computation axioms.
- `git diff --check` and the scoped build passed again at `7516662f`. This cleanup changes only one category tag and one range comment in `30.lean`; all theorem statements, proof bodies, and shared lemmas are unchanged.

**AI assistance**

Earlier revisions used Claude for Lean drafting and Perplexity for citation lookup. This update used Codex for source corrections, review, and local verification. This disclosure is confined to the pull-request description.


bengoechea:
Heads-up on a convention question before review: this PR adds two inline `axiom` declarations (`bfr_core_bound`, `singer_sidon_exists`), and I think it's the first time inline `axiom` shows up in formal-conjectures. I checked the rest of the repo with `grep -E '^axiom'` and other files handle missing-from-Mathlib literature results by describing them in the module docstring and using `sorry` in the theorem body — closest analogue I found is `ErdosProblems/659.lean` for Bernays' theorem.

The reason I went with `axiom`: both declarations are *premises* for derived theorems in the same file — `bfr_bound` follows from `bfr_core_bound`, `singer_lower_bound` follows from `singer_sidon_exists`. With `axiom`, `#print axioms` on the derived theorems names the literature dependency they rest on. With `sorry` in the source theorem's body, the consequences inherit an unnamed dependency. The 659 case uses the axiom-dependent result as the answer itself, so the audit-trail concern doesn't bite the same way there.

Happy to convert both to `sorry` + docstring if that's the project preference — it's a small mechanical change. Which form do you prefer?

Paul-Lez:
Hey @bengoechea thanks for the contribution! Usually we tend to avoid axioms and instead add the assumptions as hypotheses (or something along those lines). Here if they are needed several times, you could add two `Prop` valued definitions and then quote them as hypotheses whenever needed

bengoechea:
Done in 9cc31d6. Both axioms converted to `Prop`-valued definitions (`BFRCoreBound`, `SingerSidonExists`) threaded as explicit hypotheses on the two dependent theorems (`bfr_bound`, `singer_lower_bound`). Module docstring updated to match. No other proofs touched.

`lake build FormalConjectures.ErdosProblems.«30»` PASS locally on the pinned `v4.27.0` toolchain (7993/7993 jobs).

FormalConjectures/ErdosProblems/30.lean
```diff
@@ -20,6 +20,75 @@ import FormalConjectures.Util.ProblemImports
...
+- O'Bryant, K. (2004). *A complete annotated bibliography of work related to
+  Sidon sets.* Electron. J. Combin., DS11.
+
+### External formal-proof archive
+
+The full proof of the Lindström upper bound (≈1000 lines, residue-class
+pigeonhole + sorted-enumeration + telescoping argument) and additional
+supporting infrastructure are archived at
+`github.com/MendozaLab/erdos-experiments/tree/main/Erdos30`,
+DOI [`10.5281/zenodo.19444428`](https://doi.org/10.5281/zenodo.19444428).
```
Paul-Lez: Generally we try to avoid too much code bloat in the module docstring. I'd suggest merging the two reference blocks however.
```suggestion
```

FormalConjectures/ErdosProblems/30.lean
```diff
@@ -40,5 +109,525 @@ theorem erdos_30 : answer(sorry) ↔
...
+`IsSidonSet` is an ordered-pair specialisation of `IsSidon` (over `Finset ℕ`),
+used internally because several counting arguments below are cleaner stated
+with the ordering hypothesis `a_1 ≤ b_1`, `a_2 ≤ b_2` rather than the
+up-to-commutativity disjunction. The bridge `isSidon_iff_isSidonSet` shows
+the two forms are equivalent. -/
+
+/-- Ordered-pair form of the Sidon predicate, used internally. -/
+private abbrev IsSidonSet (A : Finset ℕ) : Prop :=
+  ∀ a₁ ∈ A, ∀ b₁ ∈ A, ∀ a₂ ∈ A, ∀ b₂ ∈ A,
+    a₁ ≤ b₁ → a₂ ≤ b₂ → a₁ + b₁ = a₂ + b₂ → (a₁ = a₂ ∧ b₁ = b₂)
```
Paul-Lez: Let's avoid duplicating definitions. If you need this additional property of `SidonSet`s, I'd recommend adding it as an API lemma in `FormalConjecturesForMathlib`.

FormalConjectures/ErdosProblems/30.lean
```diff
@@ -40,5 +109,525 @@ theorem erdos_30 : answer(sorry) ↔
...
+  · push_neg at h1
+    by_cases h2 : b₁ ≤ a₂
+    · have h_sidon := hS a₁ ha₁ b₂ hb₂ b₁ hb₁ a₂ ha₂ (Nat.le_of_lt h1) h2 (by omega)
+      exact absurd h_sidon.left.symm (Nat.ne_of_lt hlt₁)
+    · push_neg at h2
+      have h_sidon := hS a₁ ha₁ b₂ hb₂ a₂ ha₂ b₁ hb₁ (Nat.le_of_lt h1)
+        (Nat.le_of_lt h2) (by omega)
+      exact ⟨h_sidon.left, h_sidon.right.symm⟩
+
+/-- **Elementary upper bound.** For a Sidon set $A \subseteq \{0,\dots,N\}$,
```
Paul-Lez, line 51: I think this is too easy to be `research solved`. Perhaps `textbook`? Alternatively, you could put this in `FormalConjecturesForMathlib`, after having golfed the proof suitably.

FormalConjectures/ErdosProblems/30.lean
```diff
@@ -40,5 +109,525 @@ theorem erdos_30 : answer(sorry) ↔
...
+/- ## Variant 2: Distinct pairwise sums (Erdős–Turán counting)
+
+For a Sidon set A of size k, the multiset of sums {a+b : a,b ∈ A, a ≤ b}
+has all k(k+1)/2 values distinct. They lie in {2,...,2N}, giving the
+Erdős–Turán bound k(k+1)/2 ≤ 2N+1. -/
+
+/-- The set of distinct pairwise sums $a+b$ with $a \le b$ from $A$. -/
+def distinctSums (A : Finset ℕ) : Finset ℕ :=
+  ((A ×ˢ A).filter (fun p => p.1 ≤ p.2)).image (fun p => p.1 + p.2)
+
```
Paul-Lez, line 94: For all the research problems you've added here, could you in each case link in the docstring to the reference where you found the problem (ideally, adding page numbers is useful - just to document carefully where this was found). I would also suggest using the `.variants.insert_name_of_theorem_here` convention here (see in the repo documentation).

Paul-Lez (commented):
A few more comments! I see that this PR contains a few very long proofs and statements that whose labels don't match the difficulty (e.g. some basic computational results proved using `native_decide` are tagged as `research solved`). Could you fix the PR so it matches the repo guidelines for style and content?

bengoechea:
Thanks @Paul-Lez - I addressed the 5/15 review in `f6c3b43` and `34366c9`.

Summary:
- moved the ordered Sidon helper/bridge into `FormalConjecturesForMathlib`
- trimmed the module docstring and consolidated references
- renamed public variants under `erdos_30.variants.*`
- retagged elementary/computational items as `textbook`/`test`
- added reference/page details in the relevant docstrings

GitHub checks are green on the current head.

bengoechea:
@Paul-Lez friendly ping — the review comments from your 5/15 pass are all addressed: the ordered-Sidon helper + bridge moved into `FormalConjecturesForMathlib` (f6c3b43), docstring trimmed and references consolidated, public variants renamed under `erdos_30.variants.*`, and the computational items retagged `textbook`/`test` (34366c9). Checks are green. Whenever you have a moment for another look, I'd appreciate it — thanks!

FormalConjectures/ErdosProblems/30.lean
```diff
@@ -40,5 +83,519 @@ theorem erdos_30 : answer(sorry) ↔
...
   sorry
 
+/- ## Variant 1: Elementary upper bound k(k-1) ≤ 2N
+
+For a Sidon set A ⊆ {0,...,N}, the |A|(|A|-1) ordered pairs (a,b) with a ≠ b
+yield distinct positive differences in {1,...,N}, giving |A|(|A|-1)/2 ≤ N. -/
+
+/-- In a Sidon set, distinct pairwise differences identify their endpoints. -/
+@[category test, AMS 11]
+private lemma sidon_diff_injective (A : Finset ℕ)
```
Paul-Lez: Let's put this in the ForMathlib directory.

FormalConjectures/ErdosProblems/30.lean
```diff
@@ -40,5 +83,519 @@ theorem erdos_30 : answer(sorry) ↔
...
+      exact ⟨h_sidon.left, h_sidon.right.symm⟩
+
+/-- **Elementary upper bound.** For a Sidon set $A \subseteq \{0,\dots,N\}$,
+$|A|(|A|-1) \le 2N$. Hence $|A| \le \sqrt{2N} + O(1)$.
+
+**Reference:** Erdős, P., Turán, P. (1941). *On a problem of Sidon in additive
+number theory, and on some related problems.* J. London Math. Soc. **16**,
+212–215 (counting argument, p. 212). -/
+@[category textbook, AMS 11]
+theorem erdos_30.variants.elementary_difference_count (A : Finset ℕ) (N : ℕ)
```
Paul-Lez, line 58: This proof is a little long (above the 50 line limit), but I think one could extract intermediate results that would be useful in their own right and would shorten the proof. 
There's some golfing that can be done here too, e.g inlining have statements with short proofs that are only used once, and using automation a little more (e.g grind with the appropriate lemmas)

FormalConjectures/ErdosProblems/30.lean
```diff
@@ -20,6 +20,49 @@ import FormalConjectures.Util.ProblemImports
 # Erdős Problem 30
 
 *Reference:* [erdosproblems.com/30](https://www.erdosproblems.com/30)
+
+A *Sidon set* (or B₂-set) is a set of natural numbers all of whose pairwise
```
Paul-Lez: Let's remove this and the comment below about indexing since these already exist in the repo

FormalConjectures/ErdosProblems/30.lean
```diff
@@ -20,6 +20,49 @@ import FormalConjectures.Util.ProblemImports
...
+  Sidon sets.* Amer. Math. Monthly **130** (5), 437–445. arXiv:2103.15850.
+- O'Bryant, K. (2004). *A complete annotated bibliography of work related to
+  Sidon sets.* Electron. J. Combin., DS11.
+- Mendoza, K. A. (2026). *Sharper Sidon upper bound via residue-class
+  pigeonhole.* Sorted-enumeration + telescoping proof of the Lindström
+  $\sqrt N + N^{1/4} + 1$ form (~1000 lines of Lean), archived at
+  `github.com/MendozaLab/erdos-experiments/tree/main/Erdos30`,
+  DOI [`10.5281/zenodo.19444428`](https://doi.org/10.5281/zenodo.19444428).
+
+### AI disclosure
```
Paul-Lez: Let's also remove this (but keep it mentioned somewhere in the PR description)

FormalConjecturesForMathlib/Combinatorics/Basic.lean
```diff
@@ -148,6 +148,62 @@ instance (A : Finset α) [DecidableEq α] : Decidable (IsSidon (A : Set α)) :=
     i₁ + i₂ = j₁ + j₂ → (i₁ = j₁ ∧ i₂ = j₂) ∨ (i₁ = j₂ ∧ i₂ = j₁)) ?_
   rfl
 
+/-- Ordered-pair form of the Sidon predicate over `Finset ℕ`, where the
+quantification ranges over pairs $(a_1, b_1), (a_2, b_2)$ with $a_1 \le b_1$
+and $a_2 \le b_2$ rather than the up-to-commutativity disjunction in
+`IsSidon`. Useful for counting arguments that are cleaner stated with the
+ordering hypothesis. Equivalence with `IsSidon` over the coerced set is
+`Finset.isSidon_iff_isSidonOrdered`. -/
+def IsSidonOrdered (A : Finset ℕ) : Prop :=
```
Paul-Lez: Let's follow the Mathlib approach here: rather than having a second definition and showing it is equivalent to the first, let's just inline the second definition in the lemma that proves equivalence. This has the advantage of avoiding definition bloat.

bengoechea:
Thanks @Paul-Lez — addressed the 6/3 review in `03067ed`:

- inlined the ordered-Sidon characterisation into a single `Finset.isSidon_coe_iff` lemma and dropped the standalone `IsSidonOrdered` definition
- moved `sidon_diff_injective` into `FormalConjecturesForMathlib/Combinatorics/Basic.lean`, restated to take `IsSidon` directly
- extracted two reusable counting lemmas (`two_mul_card_product_filter_lt`/`_gt`) into ForMathlib; `elementary_difference_count` is now ~25 lines and `distinct_sums_card` reuses the same lemma, both under the proof-length limit
- trimmed the module docstring (dropped the Sidon-set/indexing prose and the reference + AI-disclosure blocks; the AI disclosure now lives in the PR description, and the external Lindström archive is referenced there too)

Full `lake build` passes locally on the pinned `v4.27.0` / mathlib `a3a10db0`. No `sorry` outside the open `erdos_30` statement and no `axiom`.

bengoechea:
Hi @Paul-Lez — wanted to check in on this one. All the threads from your previous round are resolved (axioms converted to `Prop`-valued hypotheses, helpers moved to `ForMathlib`, proofs shortened, AI disclosure in the PR description). Happy to make any further adjustments — just let me know.

bengoechea:
Hi @Paul-Lez — following up after a couple of weeks. Current head is mergeable and all GitHub checks are green; the prior review threads are addressed (axioms removed in favor of explicit hypotheses, helpers moved to ForMathlib, proofs shortened, AI disclosure kept in the PR description). Happy to make any remaining adjustments if there is another pass needed. Thanks again.

FormalConjectures/ErdosProblems/30.lean
```diff
@@ -40,5 +43,367 @@ theorem erdos_30 : answer(sorry) ↔
...
+
+/- ## Variant 2: Distinct pairwise sums (Erdős–Turán counting)
+
+For a Sidon set A of size k, the multiset of sums {a+b : a,b ∈ A, a ≤ b}
+has all k(k+1)/2 values distinct. They lie in {2,...,2N}, giving the
+Erdős–Turán bound k(k+1)/2 ≤ 2N+1. -/
+
+/-- The set of distinct pairwise sums $a+b$ with $a \le b$ from $A$. -/
+def distinctSums (A : Finset ℕ) : Finset ℕ :=
+  ((A ×ˢ A).filter (fun p => p.1 ≤ p.2)).image (fun p => p.1 + p.2)
```
YaelDillies, line 93: This is provably the same as `A + A`. Can you therefore remove this definition?

YaelDillies:
You're missing `open scoped Pointwise`, most likely

bengoechea:
Thanks for pointing out the missing scope. Added `open scoped Pointwise` and fixed the reversed `add_comm` step it exposed in `distinct_sums_card` in 8f49d159.

`lake --wfail build 'FormalConjectures.ErdosProblems.«30»'` now passes locally on Lean 4.33.1 and the pinned Mathlib revision. I also corrected the BFR assumption/threshold attribution, Singer interval wording, and proof-status notes; the description now records the axiom audit, including native-computation dependencies. The original open conjecture is unchanged.

PR #3936, opened 2026-05-05
feat(ErdosProblems): 96

### LINKED ISSUE #326: Erdős Problem 96
### What is the conjecture
    
https://www.erdosproblems.com/96

  If $n$ points in $\mathbb{R}^2$ form a convex polygon then there are $O(n)$ many pairs which are distance $1$ apart.

Status: open

### Prerequisites needed
While `convexHull` is in mathlib, the property "in convex position" is missing. Also see `Mathlib/Analysis/Convex/Exposed.lean`.

### Choose either option

flound1129:
fixes #326

formalizes erdős 96: is the max number of unit distances among the vertices of a convex n-gon O(n)?

uses `ConvexIndep` from `FormalConjecturesForMathlib/Geometry/2d.lean` (already there). includes a `test` lemma showing the count set is `BddAbove` so `sSup` is well-defined.

assisted by claude.


FormalConjectures/ErdosProblems/96.lean
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Erdos96
+open Finset
+
+/--
+Given a finite set of points, this function counts the number of **unordered pairs** of distinct
+points that are at a distance of exactly `1` from each other.
+-/
+noncomputable def unitDistancePairsCount (points : Finset ℝ²) : ℕ :=
+  (points.offDiag.filter (fun p => dist p.1 p.2 = 1)).card / 2
```
mo271: This definition should be extracted to FormalConjecturesForMathlib/Geometry/2d.lean, alongside the existing distinctDistances and distinctDistancesFrom definitions. Then both 90.lean and 96.lean should import and use it.

FormalConjectures/ErdosProblems/96.lean
```diff
@@ -0,0 +1,80 @@
...
+/--
+The **maximum number of unit distances** determined by the vertices of a convex `n`-gon.
+This function is often denoted as `U_c(n)` in combinatorics.
+-/
+noncomputable def maxConvexUnitDistances (n : ℕ) : ℕ :=
+  sSup (convexUnitDistanceCounts n)
+
+/--
+Does the maximum number of unit distances determined by the vertices of a convex `n`-gon
+grow linearly in `n`?
```
mo271: don't paraphrase, but use what is on erdosproblems verbatim. Also: please use latex markdown for docstrings  instead of backticks, here and in other places

mo271 (commented):
Thanks! 
LGTM, just some nits about docstrings and moving a definition to our `FormalConjecturesForMathlib`

flound1129:
All issues should be resolved.  Thanks for the review!

mo271 (approved):
Thanks, LGTM!

PR #3951, opened 2026-05-06
feat(GreensOpenProblems): 41

### LINKED ISSUE #1597: Green's Open Problems #41
### What is the conjecture
https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.41

How many rotated (about the origin) copies of the 'pyjama set' $\\{(x, y) \in \mathbb{R}^2 : \text{dist}(x, \mathbb{Z}) \leq \varepsilon\\}$ are needed to cover $\mathbb{R}^2$?

* ams-51
* ams-52

### Choose either option

jeangud:
# Description
> How many rotated (about the origin) copies of the "pyjama set" $\\{(x, y) \in ℝ^2 : \text{dist}(x, \mathbb{Z}) \leq \varepsilon\\}$ are needed to cover $ℝ^2$?
- Closes #1597.
- Proposed as an improvement over currently best-known bound.
- Added variants.
- :robot: Drafted with **Gemini 3.1 Pro** and patched up manually for style and modularity.

# Testing
:white_check_mark: Builds successfully.
```shell
$ lake build FormalConjectures/GreensOpenProblems/41.lean 
Build completed successfully (7993 jobs).
```

### DIFF
--- a/FormalConjectures/GreensOpenProblems/41.lean
+++ b/FormalConjectures/GreensOpenProblems/41.lean
@@ -1,0 +1,47 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Ben Green's Open Problem 41
+
+*Reference:* [Ben Green's Open Problem 41](https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.41)
+-/
+
+namespace Green41
+
+open Set
+
+/-- The pyjama set is the set of points in the plane whose x-coordinate is within ε of an integer. -/
+def pyjamaSet (ε : ℝ) : Set (ℝ × ℝ) :=
+  { p | ∃ z : ℤ, |p.1 - (z : ℝ)| ≤ ε }
+
+/-- A rotation about the origin by angle θ. -/
+noncomputable def rotate (θ : ℝ) (p : ℝ × ℝ) : ℝ × ℝ :=
+  (p.1 * Real.cos θ - p.2 * Real.sin θ, p.1 * Real.sin θ + p.2 * Real.cos θ)
+
+/--
+How many rotated (about the origin) copies of the 'pyjama set' $\\{(x, y) \in \mathbb{R}^2 : \text{dist}(x, \mathbb{Z}) \leq \varepsilon\\}$ are needed to cover $\mathbb{R}^2$?
+-/
+@[category research open, AMS 51 52]
+theorem green_41 :
+    ∀ ε > 0,
+      let ans := (answer(sorry) : ℕ)
+      IsLeast { n : ℕ | ∃ (Θ : Finset ℝ), Θ.card = n ∧ ∀ p : ℝ × ℝ, ∃ θ ∈ Θ, p ∈ rotate θ '' pyjamaSet ε } ans := by
+  sorry
+
+end Green41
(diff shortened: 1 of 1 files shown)


FormalConjectures/GreensOpenProblems/41.lean
```diff
@@ -0,0 +1,72 @@
...
+an integer.
+-/
+def pyjamaSet (ε : ℝ) : Set ℂ :=
+  { z | ∃ k : ℤ, |z.re - (k : ℝ)| ≤ ε }
+
+/-- The minimal number of rotated copies of the pyjama set of width ε needed to cover the plane. -/
+noncomputable def minCopies (ε : ℝ) : ℕ :=
+  sInf { n : ℕ | ∃ (Θ : Finset ℝ), Θ.card = n ∧
+    (⋃ θ ∈ Θ, exp (θ * I) • pyjamaSet ε) = univ }
+
```
Paul-Lez, line 58: Here it would be nice having a test statement that shows non-emptiness of the set we're taking a inf on

FormalConjectures/GreensOpenProblems/41.lean
```diff
@@ -0,0 +1,72 @@
...
+/--
+How many rotated (about the origin) copies of the 'pyjama set'
+$\\{(x, y) \in \mathbb{R}^2 : \text{dist}(x, \mathbb{Z}) \leq \varepsilon\\}$ are needed to cover
+$\mathbb{R}^2$?
+-/
+@[category research open, AMS 51 52]
+theorem green_41 :
+    ∃ C : ℝ, C > 0 ∧ ∃ ε₀ > 0, ∀ ε ∈ Ioc 0 ε₀,
+      let ans := (answer(sorry) : ℝ)
+      (minCopies ε : ℝ) ≤ ans ∧ ans < Real.exp (Real.exp (Real.exp (ε ^ (-C)))) := by
```
Paul-Lez, line 70: The upper bound on `ans` doesn't seem to be in the informal statement - where does it come from?

jeangud: Thank you @Paul-Lez for the review!

The upper bound is from [[KrLe25](https://arxiv.org/pdf/2510.17744)] as mentioned in [Green's comments](https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.41) (page 21):
> **Comments.** Manners [Ma15], solving the ‘pyjama problem’, proved that there is some finite set of rotations with this property. His proof uses topological dynamics and is ineffective. I am not aware of any nontrivial lower bounds: in particular, is $ε^{−C}$ rotations enough? 
> **Update 2025.** Kravitz and Leng [KrLe25] have established the first quantitative bound, showing via an analysis of Manners’s method that $\exp~\exp~\exp(ε^{−C})$ rotations suffice.

For `green_41`, I am proposing that the user suggests an `ans` bound, which would have to be tighter than the KrLe25 best-known bound.

This bound is also used below in `green_41.variants.kravitz_leng` as an intermediate variant.

Paul-Lez: Ok that sounds good! Would you mind mentioning it in the informal statement? (e.g. something along the lines of "can one find a better bound than [...]"). As an little extra variant that could be nice, I'd also suggest having an existential version of the main problem that just asks to prove that a better bound exists, without giving it explicitly!

FormalConjectures/GreensOpenProblems/41.lean
jeangud: Agreed. I explored a few paths and ended up adding a `sorry` test for now:
```
@[category test, AMS 51 52]
theorem minCopies_set_nonempty (ε : ℝ) (hε : 0 < ε) :
    (coveringCopies ε).Nonempty := by
  sorry
```

It is possible to obtain short-ish `sorry`-free proofs for reduced subsets like $\varepsilon = 1$, $\varepsilon > \frac{1}{2}$, or even $\varepsilon > \frac{1}{3}$. However, showing that the set is not empty for all $\epsilon > 0$ is exactly the scope of  [[Ma15](https://arxiv.org/pdf/1305.1514)] (27 pages), and would likely require many lines of code.

What do you think?

FormalConjectures/GreensOpenProblems/41.lean
jeangud: Good idea! Done, see https://github.com/google-deepmind/formal-conjectures/commit/ed4c8bdb923b5df05bfb258cd736f2e288e27422

jeangud:
-awaiting-author

FormalConjectures/GreensOpenProblems/41.lean
Paul-Lez: Ah that's more than I thought! In that case, we should add this as a `research solved` statement!

Paul-Lez (approved):
Nice!

FormalConjectures/GreensOpenProblems/41.lean
jeangud: Done :white_check_mark:

PR #3959, opened 2026-05-08
chore: update status of Rupert problem, adding link to paper and formalization

dwrensha:
Two updates:
1. In August 2025, [a paper](https://arxiv.org/abs/2508.18475) that resolved the conjecture was uploaded to arXiv.
2. As of this week, we have [a complete formalization](https://github.com/jcreedcmu/Noperthedron) of that paper. (Because it's a heavily computational proof, the formalization includes the Lean native code compiler in its trusted computing base.)

### DIFF
--- a/FormalConjectures/Paper/Rupert.lean
+++ b/FormalConjectures/Paper/Rupert.lean
@@ -109,6 +109,6 @@ not hold.
 -/
 @[category research solved, AMS 52, formal_proof using lean4 at "https://github.com/jcreedcmu/Noperthedron"]
 theorem is_every_convex_polyhedron_rupert :
-    ∃ vertices : Finset ℝ³,
-       (interior (convexHull ℝ vertices : Set ℝ³)).Nonempty ∧ ¬ IsRupert vertices := by
+    answer(False) ↔ ∀ (vertices : Finset ℝ³),
+       (interior (convexHull ℝ vertices : Set ℝ³)).Nonempty → IsRupert vertices := by
  sorry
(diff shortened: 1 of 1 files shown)


FormalConjectures/Paper/Rupert.lean
```diff
@@ -100,10 +104,11 @@ def IsRupert (vertices : Finset ℝ³) : Prop :=
...
-Does the Rupert property hold for every convex polyhedron with nonempty interior?
+There exists a convex polyhedron with nonempty interior for which the Rupert property does
+not hold.
 -/
-@[category research open, AMS 52]
+@[category research solved, AMS 52, formal_proof using lean4 at "https://github.com/jcreedcmu/Noperthedron"]
 theorem is_every_convex_polyhedron_rupert :
-    answer(sorry) ↔ ∀ (vertices : Finset ℝ³),
-       (interior (convexHull ℝ vertices : Set ℝ³)).Nonempty → IsRupert vertices := by
+    ∃ vertices : Finset ℝ³,
```
jcreedcmu: Question for the formal-conjectures maintainers: Would it be preferable to instead fill in `answer(sorry)` with `answer(False)`?

dwrensha: Either way is fine with me. I'll defer to the judgment of the project maintainers.

Paul-Lez: Usually we'd also fill the `answer()` with the actual solution!

Paul-Lez (approved):
Very cool!

FormalConjectures/Paper/Rupert.lean
dwrensha: Makes sense. I've updated this PR to fill in the `answer(sorry)` with `answer(False)`.

dwrensha:
The CI failure is
```
[FormalConjectures.ErdosProblems.«613»](warning: FormalConjectures/ErdosProblems/613.lean:32:0: AMS subject 05 should not have a leading zero.)
```
which I think should be fixed by https://github.com/google-deepmind/formal-conjectures/commit/99ae49df4b27d9d95877c562ab9e1fd9b7ff1dc7

PR #3977, opened 2026-05-10
Add Erdős 953 upper-bound formal proof link

### LINKED ISSUE #1032: Erdős Problem 953
### What is the conjecture

https://www.erdosproblems.com/953

  Let $A\subset \\{ x\in \mathbb{R}^2 : \lvert x\rvert <r\\}$ be a measurable set with no integer distances, that is, such that $\lvert a-b\rvert \not\in \mathbb{Z}$ for any distinct $a,b\in A$. How large can the measure of $A$ be?

Status:  open

### Choose either option

AllenGrahamHart:
fixes #1032 

This adds a new Erdős Problem 953 statement for the upper-bound half of the order-of-growth resolution.

The theorem formalizes the closed-ball extremal function M and states that M(R) is O(sqrt R). The file notes that this is not a formalization of the matching Sárközy lower bound, nor of the combined informal conclusion M(R) = R^{1/2+o(1)}.

The long Lean proof is hosted externally in FormalConjectures-Bench and linked through formal_proof using lean4.

Verification run locally:

lake env lean FormalConjectures/ErdosProblems/953.lean

This completed successfully, with only the expected sorry warning for the linked statement.

### DIFF
--- a/FormalConjectures/ErdosProblems/953.lean
+++ b/FormalConjectures/ErdosProblems/953.lean
@@ -1,0 +1,79 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


Smetalo:
Does this resolve #1032?

AllenGrahamHart:
> Does this resolve #1032?

It is debatable because 'How large can the measure of A be' is not that well defined.

I would argue 'substantively yes' but there might be disagreement on erdosproblems.com

PR #3987, opened 2026-05-11
feat(website): mention contributors per Lean file

Paul-Lez:
Adds file-level contributor metadata from git history to the website build and renders contributor chips with GitHub profile/avatar links when available.


mo271 (approved):
looks really nice in your fork: for instance
https://paul-lez.github.io/formal-conjectures/theorem/?name=ABC.abc.variants.quality!

I suppose this will work with the default token, and we don't need to create one for our repo here, right?

Paul-Lez:
I think so! Are you happy for us to merge and then fix if this doesn't work out?

PR #3988, opened 2026-05-11
feat(Bugeaud): Problem 10.5

rwst:
Problem 10.5 of Bugeaud's collection presents two connected conjectures by Dubickas.


FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_5.lean
mo271, line 1: Minor typo in directory name: "DistributionsModuloOne" the book title is "Distribution modulo one" (singular). Should be BugeaudDistributionModuloOne/

mo271, line 1: The filename Lacunary.lean is reasonable for a file about lacunary sequences. However, the source is Problem 10.5 from a specific chapter. If more problems from this book are expected, a filename like Problem10_5.lean or Chapter10.lean would be more conventional, especially if you plan to add more problems from this book.

FormalConjectures/Books/BugeaudDistributionsModuloOne/Lacunary.lean
```diff
@@ -0,0 +1,99 @@
...
+
+/--
+Problem 10.5 (first part). Let $\mathbb{K}$ be a real number field. Then, for any
+$\varepsilon > 0$, there exists a lacunary sequence $(t_n)_{n \ge 1}$ of positive numbers
+in $\mathbb{K}$ such that
+$$\limsup_{n \to \infty} \{\xi t_n\} \ge 1 - \varepsilon,$$
+for any real number $\xi$ not in $\mathbb{K}$.
+
+Lacunarity here is the real-valued analogue of `IsLacunary`: there is some $c > 1$ with
+$c \cdot t_k < t_{k+1}$ for all sufficiently large $k$.
```
mo271: This is a reasonable ad hoc choice. But the right fix would be to generalize `IsLacunary` to work for real-valued sequences (e.g., ℕ → ℝ), or add an `IsLacunaryReal` variant in that file. This would allow reuse across both contexts.

mo271 (commented):
Thanks!
Looks quite nice, just some minor comments

mo271 (approved):
Thanks, LGTM!

PR #3986, opened 2026-05-11
docs: centralize contribution guidance in CONTRIBUTING.md

Sfgangloff:
This is the first half of #3967, split per @Paul-Lez's review feedback. A follow-up PR will add the website-side rendering of `CONTRIBUTING.md` once this lands.

## Why

Contribution guidance previously lived in three places that drifted out of sync:

- `README.md` — high-level project description plus a long contribution section.
- `CONTRIBUTING.md` — only the CLA, community-guidelines, and label-management bits (auto-linked by GitHub from issues/PRs).
- `site/src/templates/contribute.html` — the website's contribute page, hand-written as static HTML.

This PR makes `CONTRIBUTING.md` the single source of truth on the repo side. The follow-up website PR will embed it into the site so GitHub and the website stay automatically in sync.

## What changed

`CONTRIBUTING.md` is now the canonical contribution guide. It now includes:

- "Ways to contribute" (including the literature/community/problem-list breakdown previously only in `README.md`).
- The full 9-step contribution process from `contribute.html`, including some previously absent: "add file to `FormalConjecturesForMathlib.lean` index", "verify embedded links", "link the PR to the issue".
- File structure and copyright header guidance.
- Reference docs for: `@[category]`, `@[formal_proof]`, `@[AMS]`, the `answer( )` elaborator.
- Style guidelines and the "Problems that require answers" caveat.
- A "Getting started with Lean 4" install snippet.
- The existing CLA, community-guidelines, code-review, and label-management sections.

New sections added:

- **Choosing a folder and category** — lists every top-level source directory under `FormalConjectures/`, gives a one-line "use X when source is Y" rule, and includes a quick decision rule for `@[category]` tags.
- **Mark a problem as solved** — explains how to flip the category tag and add an `@[formal_proof]` attribute when a previously-open problem has been solved (formally or informally).

`README.md` was simplified: the duplicated contribution/reference sections now redirect to `CONTRIBUTING.md`. It still contains the project description, install/build commands, directory structure, versioning, and licensing information.

## Test plan

- [x] `CONTRIBUTING.md` renders correctly on GitHub.
- [x] `README.md` still contains install/build/license info and links to `CONTRIBUTING.md`.

PR #3993, opened 2026-05-12
Rendering contribute webpage from CONTRIBUTE.md

### LINKED ISSUE #3969: Improve contributor onboarding and consolidate contribution documentation
The contribution documentation was previously duplicated across multiple locations (README.md, CONTRIBUTING.md, and site/src/templates/contribute.html), which made maintenance difficult and caused sections to drift out of sync.

This issue tracks the consolidation of all contribution guidance into CONTRIBUTING.md as the single source of truth, with the website contribution page rendered directly from it at build time.

The changes also improve onboarding for new contributors by:

clarifying the contribution workflow,
documenting folder/category selection,
explaining how to mark problems as solved,
adding Lean 4 setup guidance,
centralizing references and style conventions.

Additionally, external documentation links on the website now open in a new tab to improve navigation during the contribution process.

Implemented in: #3967

Sfgangloff:
Sequel of #3986

Closes #3969

The `/contribute/` page is now built from `CONTRIBUTING.md` instead of being hand-maintained in HTML, so contribution guidance lives in one place.

Includes a small Markdown→HTML renderer in `site/build.js` (no new dependencies). It handles only what `CONTRIBUTING.md` uses right now: headings, fenced code blocks, nested lists, inline code, bold/italic, links, and autolinks.


mo271:
Thanks! I like the general idea of only having CONTRIBUTE.md in one place and re-using it in the website -- however, it seems to me that having a custom Markdown->HTML renderer might be problematic. If we change the contributions page in the future, add more code examples, latex or something else we will have to remember also adapting hte HTML renderer. Also we should make sure that it is rendered identical or at least very similar to how it is rendered on Github. 
Perhaps we could consider some alternatives? On conceivably could try to use the GitHub REST API, to render the markdown, but likely this would require providng some token and introduce some latency.
Perhaps we can just look into whatever github uses to render markdown and re-use that? or use at least some standard markdown+latex rendering? Perhaps the one that we use for our doc-comments in the verso rendering anyhow?

Sfgangloff:
> Thanks! I like the general idea of only having CONTRIBUTE.md in one place and re-using it in the website -- however, it seems to me that having a custom Markdown->HTML renderer might be problematic. If we change the contributions page in the future, add more code examples, latex or something else we will have to remember also adapting hte HTML renderer. Also we should make sure that it is rendered identical or at least very similar to how it is rendered on Github.
> Perhaps we could consider some alternatives? On conceivably could try to use the GitHub REST API, to render the markdown, but likely this would require providng some token and introduce some latency.
> Perhaps we can just look into whatever github uses to render markdown and re-use that? or use at least some standard markdown+latex rendering? Perhaps the one that we use for our doc-comments in the verso rendering anyhow?

@mo271 Makes sense, I have not thought about this issue. Now I think relying on a standard version does not remove the problem entirely I believe (especially regarding visual identity). 

Another idea is not searching to make CONTRIBUTE.md and the contribute page identical, and keep CONTRIBUTE.md as it is with a link to the page which could be considered as full version (and can be augmented form now with any latex or else), including the copy of CONTRIBUTE.md. Another way would be to simply to replace CONTRIBUTE.md with the page and keep only this, but this means less discoverability. What do you think ?

mo271:
@Sfgangloff thanks for your ideas here, but I don't currently see a good way forward. I agree with you that it would be great to avoid duplication, just don't see how this would be done easily. Closing for now, feel free to re-open if there is a clear path forward

Issue #309, opened 2025-07-04
Erdős Problem 80


mo271:
### What is the conjecture
    
https://www.erdosproblems.com/80

  Let $c>0$ and let $f_c(n)$ be the maximal $m$ such that every graph $G$ with $n$ vertices and at least $cn^2$ edges, where each edge is contained in at least one triangle, must contain a book of size $m$, that is, an edge shared by at least $m$ different triangles. Estimate $f_c(n)$.  In particular, is it true that $f_c(n)>n^{\epsilon}$ for some $\epsilon>0$? Or $f_c(n)\gg \log n$?

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

zachary-feng:
Hi, I'd like to work on this!

PR #3996, opened 2026-05-12
feat(ErdosProblems): 600

### LINKED ISSUE #819: Erdős Problem 600
### What is the conjecture
    
https://www.erdosproblems.com/600

  Let $e(n,r)$ be minimal such that every graph on $n$ vertices with at least $e(n,r)$ edges, each edge contained in at least one triangle, must have an edge contained in at least $r$ triangles. Let $r\geq 2$. Is it true that
$$e(n,r+1)-e(n,r)\to \infty$$
as $n\to \infty$? Is it true that
$$\frac{e(n,r+1)}{e(n,r)}\to 1$$
as $n\to \infty$?
  

Status:  open

### Choose either option

zachary-feng:
Fixes #819

Any feedback is greatly appreciated!


FormalConjectures/ErdosProblems/600.lean
```diff
@@ -0,0 +1,69 @@
...
+
+/--
+Let $e(n,r)$ be minimal such that every graph on $n$ vertices with at least $e(n,r)$ edges,
+each edge contained in at least one triangle, must have an edge contained in at least
+$r$ triangles.
+-/
+def Erdos600Prop (n : ℕ) (e : ℕ) (r : ℕ) : Prop :=
+  ∀ G : SimpleGraph (Fin n), G.edgeFinset.card ≥ e ∧
+  (∀ e ∈ G.edgeFinset, ∃ t ∈ G.cliqueFinset 3, e.toFinset ⊆ t) →
+  (∃ e ∈ G.edgeFinset, ∃ S ⊆ G.cliqueFinset 3, ∀ t ∈ S, e.toFinset ⊆ t ∧ S.card ≥ r)
```
mo271: Rather than the awkward ∃ S ⊆ G.cliqueFinset 3, ∀ t ∈ S, ... formulation, use Finset.filter directly:

```lean
private def trianglesContaining (G : SimpleGraph (Fin n)) (uv : Sym2 (Fin n)) : Finset (Finset (Fin n)) :=
  (G.cliqueFinset 3).filter (fun t => uv.toFinset ⊆ t)
```
Then Erdos600Prop becomes:

```lean
def Erdos600Prop (n e r : ℕ) : Prop :=
  ∀ G : SimpleGraph (Fin n),
    G.edgeFinset.card ≥ e →
    (∀ uv ∈ G.edgeFinset, (trianglesContaining G uv).Nonempty) →
    ∃ uv ∈ G.edgeFinset, r ≤ (trianglesContaining G uv).card
```

This would also solve the confusing variable shadowing for the variable `e`

mo271 (commented):
Thanks, looks good generally, just one suggestion to make is cleaner.

Also: let's also mention the other (cross)references mention on erosproblems.com

FormalConjectures/ErdosProblems/600.lean
zachary-feng: Thanks for the review!

I rewrote `Erdos600Prop` using `trianglesContaining` per your suggestions.

Can I ask what is best practice if I want to reuse the prop `trianglesContaining` in #309?

mo271: the normal way would be to move the definition to an appropriate place in `FormalConjecturesForMathlib`

zachary-feng: @mo271 On second thought, I don’t think `trianglesContaining` has much potential for reuse, so I’m inclined to keep it local to this file instead. Would that be okay?

Other than that, I think everything else looks good. Would you mind taking another look? Thanks!

zachary-feng:
-awaiting-author

YaelDillies (approved):
Looks perfect, thank you!

PR #3997, opened 2026-05-13
prove(ErdosProblems/36): M(1) = 1

md-amer:
## Summary

Proves `M_one : M 1 = 1` in `FormalConjectures/ErdosProblems/36.lean`, replacing the `sorry` placeholder.

**Mathematical content:** M(1) = 1 is the base case of the minimum overlap problem. The set {1,2} has only one valid equal split (A={1}, B={2}), which achieves MaxOverlap = 1.

**Proof structure:**
- `overlap_singleton`: Overlap {a} {b} k = 1 if a−b=k, else 0
- `overlap_singleton_le_one`: bound ≤ 1 everywhere
- `overlap_singleton_self`: hits 1 at k = a−b
- `maxOverlap_singleton`: MaxOverlap {a} {b} = 1 (via ciSup_le + le_ciSup)
- `M_one`: the set of achievable MaxOverlap values equals {1}, so csInf_singleton closes the goal

**Verified** via leanprover-community online editor (Latest Mathlib).

### DIFF
--- a/FormalConjectures/ErdosProblems/36.lean
+++ b/FormalConjectures/ErdosProblems/36.lean
@@ -92,7 +92,7 @@ theorem M_one : M 1 = 1 := by
     constructor
     · rintro ⟨A, B, hdis, huni, hcard, rfl⟩
       have huni' : A ∪ B = {1, 2} := by
-        rw [huni]; decide
+        rw [huni]; ext x; simp [Finset.mem_Icc]; omega
       have hcard2 : A.card + B.card = 2 := by
         have := (Finset.card_union_of_disjoint hdis).symm
         rw [huni'] at this; simp at this; linarith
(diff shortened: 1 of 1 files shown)


mo271:
I think here the best step forward would be to prove all cases up to `M 5`, make a commit and if the proofs are too long or involved remove them in a follow up commit and use the `formal_proof using formal_conjectures at [link to previous commit]` mechanism

Paul-Lez:
Note also that commits co-authored by Claude Code may cause issues with the Google CLA.

md-amer:
Thanks for the feedback. Unfortunately I don't have the bandwidth to extend this to M(2)–M(5) right now. Closing for now — someone else can pick this up or I may return to it later.

PR #4002, opened 2026-05-13
chore(Wikipedia): harmonize module docstrings

Paul-Lez:
Various fixes:
- fixing `OddWeirdNumber.lean` to use a module docstring (`/-!`) so it appears in generated docs
- converting broken or literal-rendering references to normal Markdown links
- making reference headings and bullet styles more uniform across the Wikipedia files
- attaching continuation text such as authors and MathOverflow askers to the relevant reference bullets

### DIFF
--- a/FormalConjectures/Wikipedia/WoodalPrimes.lean
+++ b/FormalConjectures/Wikipedia/WoodalPrimes.lean
@@ -16,11 +16,12 @@ limitations under the License.
 
 import FormalConjectures.Util.ProblemImports
 
-/-! # Woodall Primes
+/-!
+# Woodall Primes
 
-References:
-* [Wikipedia/Woodall Number](https://en.wikipedia.org/wiki/Woodall_number#Woodall_primes)
-* [A2234](https://oeis.org/A2234)
+*References:*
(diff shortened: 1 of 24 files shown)


FormalConjectures/Wikipedia/WoodalPrimes.lean
```diff
@@ -16,11 +16,12 @@ limitations under the License.
...
-/-! # Woodall Primes
+/-!
+# Woodall Primes
 
-References:
-* [Wikipedia/Woodall Number](https://en.wikipedia.org/wiki/Woodall_number#Woodall_primes)
-* [A2234](https://oeis.org/A2234)
+*References:*
+- [Wikipedia: Woodall number](https://en.wikipedia.org/wiki/Woodall_number#Woodall_primes)
+- [OEIS A002234](https://oeis.org/A002234)
```
mo271, line 24: oeis numbers are incosistent with and without leading zeros even after fixes in this pull request. Perhaps worth harmonizing (I'm in favor of always leaving out leading zeros..)

mo271 (approved):

PR #4003, opened 2026-05-13
chore(ErdosProblems): Normalize Erdős problem docstring references

Paul-Lez:
- normalize several Erdős module docstring headings so the website gets a consistent `# Erdős Problem N` title
- clean malformed/raw reference links, singular/plural reference markers, and typoed source labels
- align reference keys with the Erdős Problems citation style and update in-file docstring cross-references where labels changed


mo271 (approved):
Thanks!

PR #3589, opened 2026-03-17
Add voting, truth predictions, and difficulty ratings to the website

zond:
## Summary

Adds a GitHub Discussions-backed voting system to the Formal Conjectures website:

- **Like** theorems (HEART reactions)
- **Predict true/false** (THUMBS_UP/DOWN reactions)
- **Rate difficulty** 0-9 (discussion comments)
- **Consent modal** on first use explaining that all activity is public on GitHub
- **Discussion links** on theorem detail pages (created lazily)
- **Sort by** most liked, difficulty, or prediction on the browse page

All data is stored as native GitHub Discussions features — no separate database.

## Architecture

- **Reads**: App Engine proxy (`formal-conjectures-web-worker`) fetches discussions using GitHub App installation tokens
- **Writes**: browser talks directly to GitHub GraphQL using the user's OAuth token
- **OAuth**: callback routed through the proxy so one registered URL works for all forks
- **Proxy validates** that requested repos are `google-deepmind/formal-conjectures` or forks of it
- **CORS** allows any `*.github.io` origin automatically

## What changed

- `site/src/js/voting.js` — complete rewrite: GitHub Discussions backend, consent modal, truth predictions
- `site/src/js/theorem.js`, `browse.js` — enable voting integration, new sort options
- `site/src/css/style.css` — truth widget, consent modal, discussion link styles
- `site/src/templates/` — uncomment voting.js, add sort options
- `site/appengine/` — new App Engine proxy (replaces `site/worker/` Cloudflare Worker)
- `site/build.js` — accept pre-processed JSON from the live site
- `.github/workflows/build-and-docs.yml` — `-webtest` branch support, voting config injection

## How to try it

The branch is deployed at https://zond.github.io/formal-conjectures/. Click any theorem, then try voting, predicting, or rating difficulty.

## Zero-config for forkers

Branches ending in `-webtest` skip the expensive Lean build, download the conjectures JSON from the live site, and deploy to GitHub Pages. To try on your own fork:

1. Enable GitHub Pages (Settings > Pages > GitHub Actions)
2. Enable Discussions (Settings > General > Features)
3. Install the GitHub App: https://github.com/apps/formal-conjectures-voting/installations/new
4. Push a branch ending in `-webtest`

No GCP project, secrets, or tokens needed. See `site/docs/voting.md` for full details.

## Test plan

- [x] Like/unlike a theorem, verify heart reaction appears on the GitHub Discussion
- [x] Predict true/false, verify thumbs reaction on the Discussion
[...]

PR #4008, opened 2026-05-15
feat(build-and-docs) quick `-webtest` branch

mo271:
and a script to quickly build the website locally from the .json that is served by the current website without running verso


mo271:
partially picked from #3589 (but updated to also support the verso fragments)

Paul-Lez (approved):

PR #4011, opened 2026-05-16
feat(ErdosProblems): 98

### LINKED ISSUE #328: Erdős Problem 98
### What is the conjecture
    
https://www.erdosproblems.com/98

  Let $h(n)$ be such that any $n$ points in $\mathbb{R}^2$, with no three on a line and no four on a circle, determine at least $h(n)$ distinct distances. Does $h(n)/n\to \infty$?

### Prerequisites needed
`Collinear` is in mathlib, needs a prop "no four on a Circle".

### Choose either option

flound1129:
fixes #328

formalizes erdős 98: does h(n)/n → ∞, where h(n) is the minimum number
of distinct distances determined by any n-point set in ℝ² in general
position (no three collinear, no four cocyclic)?

uses `distinctDistances` and `InGeneralPosition` from
`FormalConjecturesForMathlib/Geometry/2d.lean` (already there). h(n) is
defined via `sInf` over the natural-number distance counts of qualifying
finsets, so `Tendsto (h n / n) atTop atTop` is the cleanest spelling of
the limit question.

### DIFF
--- a/FormalConjectures/ErdosProblems/98.lean
+++ b/FormalConjectures/ErdosProblems/98.lean
@@ -1,0 +1,46 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

Two remarks: 

- Potentially also add the statement from the additional material or leave a TODO
- add references (the ones mentioned on erdosproblems.com, where the problem comes from

flound1129:
Added and updated!  Thanks for the review!

PR #4016, opened 2026-05-17
feat(Probabilistic): Sidorenko's conjecture (1993)

henrykmichalewski:
## Summary
- Adds `FormalConjectures/Probabilistic/SidorenkoConjecture.lean` (614 lines, 12 thm/lem) formalising Sidorenko's conjecture:
  > For every bipartite graph $H$ and every graph $G$, $t(H, G) \geq t(K_2, G)^{e(H)}$.
- Headline `sidorenko_conjecture` filed as `answer(sorry) ↔ ...` (open).
- **Three proved special cases**: $H = K_2$ (`rfl`), $H = K_{2,2}$ (Cauchy–Schwarz; closed in this file), and $H$ a tree (scaffolded with main inductive step `sorry`).
- Uses upstream `SimpleGraph.homDensity` from `FormalConjecturesForMathlib/Combinatorics/SimpleGraph/HomDensity.lean` (merged in #3834) — no new ForMathlib dependencies.
- Second of a planned probabilistic-content PR series; first is `#4015` (LovaszLocalLemma).

## Test plan
- [x] `lake build FormalConjectures.Probabilistic.SidorenkoConjecture` — green, no warnings.
- [x] No new sorrys outside the deliberate `answer(sorry)`-d headline and the tree-case induction.

PR #4015, opened 2026-05-17
feat(ForMathlib/Probability): Lovász Local Lemma (symmetric + asymmetric)

henrykmichalewski:
## Summary
- Adds `FormalConjecturesForMathlib/Probability/LovaszLocalLemma.lean` (268 lines) formalising the Erdős–Lovász 1975 Lovász Local Lemma in both forms:
  - **Symmetric form** — events with `Pr ≤ p`, dependency degree `≤ d`, hypothesis `e · p · (d+1) ≤ 1` ⇒ positive probability of avoiding all.
  - **General/asymmetric form** — dependency graph + reals `0 ≤ x_i < 1` with `Pr(A_i) ≤ x_i · ∏_{j ∈ Γ(i)} (1 - x_j)`.
- Sorry-free real-analytic core bound `(1 - 1/(d+1))^d ≥ 1/e` (the key step that lets the symmetric hypothesis imply the asymmetric one with `x_i := 1/(d+1)`).
- Two main LLL theorems stated with focused `sorry` pending Mathlib's probability-induction infrastructure.
- Will be used downstream by probabilistic-method problem files (Erdős 1959 triangle-free high χ + girth, Sidorenko special cases) — first of a planned probabilistic-content PR series.

## Test plan
- [x] `lake build FormalConjecturesForMathlib.Probability.LovaszLocalLemma` — green.
- [x] Reuses upstream `Mathlib.Probability.Independence.Basic`, `Mathlib.MeasureTheory.Measure.MeasureSpace`, `Mathlib.Analysis.SpecialFunctions.Exp` — no new dependencies outside Mathlib.

PR #4017, opened 2026-05-17
feat(Probabilistic): Erdős 1959 — graphs with arbitrarily high χ and girth

henrykmichalewski:
## Summary
- Adds `FormalConjectures/Probabilistic/ErdosHighChromaticHighGirth.lean` (124 lines, 3 thm/lem) formalising Erdős's 1959 deletion-method theorem:
  > For every $k \geq 2$ and every $g \geq 3$, there exists a graph $G$ with $\chi(G) \geq k$ and $\mathrm{girth}(G) \geq g$.
- Headline `erdos_1959_high_chromatic_high_girth` plus two trivial-boundary corollaries ($g = 3$, $k = 2$).
- Uses Mathlib's `SimpleGraph.chromaticNumber` and `SimpleGraph.girth` directly — **no new defs, no new ForMathlib dependencies**.
- Third of a planned probabilistic-content PR series (#4015 LLL, #4016 Sidorenko).

## Test plan
- [x] `lake build FormalConjectures.Probabilistic.ErdosHighChromaticHighGirth` — green.


mo271 (commented):
Thanks! 

A general comment before the actual review: 
Could you incorporate the feedback given on your previous prs in newly opened ones: 
Most notably there is probably around 10 previous prs where in review it was pointed out that we want to avoid docstring duplication, not repeat the "Status" in the docstring, etc.. 
See the reviews of previous prs. It would probably beneficial to copy these guidelines to your prompt when generating these formalisation. Also happy to add them to AGENTS.md if you think it is beneficial. It already says "Follow existing file patterns in the repository", perhaps one should need to tell the agent to follow what is in AGENTS.md?

henrykmichalewski:
@mo271 — meta-feedback applied retroactively across all 4 new probabilistic PRs (this one + #4015, #4016, #4018). In [`5243e79`](https://github.com/henrykmichalewski/formal-conjectures/commit/5243e79) on this PR:

- **Moved** to `FormalConjectures/Wikipedia/` per your Sidorenko hint.
- **Trimmed module docstring** to the standard `# Title` + `*References:*` block (Wikipedia link + 2 references). Dropped the verbatim quote, the multi-paragraph proof-sketch, and the Status header. Future probabilistic PRs will start from this template.

Sorry for the docstring-bloat repeats across PRs — happy to add the convention checklist to `AGENTS.md` if you'd like; that would surface naturally during the agent's planning step. Build green.

mo271 (commented):
Thank you for the contribution!

I'm not sure how this best fits in the repo: There aren't any conjectures added here. Is the plan to build upon that, i.e. add conjectures later which use these theorems/definitions? In that case `FormalConjecturesForMathlib` with `sorry`-free proofs is the way to go.

henrykmichalewski:
Same situation as #4015 — the file holds 3 theorems all `@[category research solved]` with `sorry` proofs (Erdős's 1959 high-chromatic-high-girth + two easy corollaries). They are *theorems* (Erdős 1959 + easy specialisations) not conjectures.

Two options I see:

1. **Keep here as solved-research-with-deferred-Lean-proof** — same pattern I'm proposing for #4015. Classifications stay accurate; `sorry` reflects only the unfinished Lean formalisation, not mathematical status.
2. **Move to `FormalConjecturesForMathlib/Combinatorics/SimpleGraph/HighChromaticHighGirth.lean`** once a sorry-free proof lands (probabilistic method using the LLL infrastructure from #4015, or alteration method as in the original paper).

Path (2) is structurally cleaner but blocks on either #4015 landing OR writing an independent proof here. Path (1) lets it land now and we can promote to ForMathlib later.

Happy to do (1) immediately if you agree, then track (2) as follow-up after #4015. Or if you'd prefer to keep this as a sketch-only file until the proof actually closes, I can mark the theorems differently (e.g. `@[category test]` — no, that's wrong too). What's your preference?

mo271:
let's first get #4015 in

PR #4019, opened 2026-05-18
fix(Mathoverflow/235893): quantifier fix

mo271:
otherwise $n=1$ is a counterexample

### DIFF
--- a/FormalConjectures/Mathoverflow/235893.lean
+++ b/FormalConjectures/Mathoverflow/235893.lean
@@ -43,7 +43,6 @@ theorem Continuous.isConnectedMap {X Y : Type*} [TopologicalSpace X] [Topologica
     {f : X → Y} (hf : Continuous f) : IsConnectedMap f :=
   fun _ h ↦ IsConnected.image h f (Continuous.continuousOn hf)
 
-
 /--
 A set in $\mathbb{R}$ is connected if and only if it is order-connected and non-empty.
 -/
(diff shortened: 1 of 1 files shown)


smmercuri:
So the formalization here is the negation of the original question on [math overflow](https://mathoverflow.net/questions/235893/does-there-exist-a-bijection-of-mathbbrn-to-itself-such-that-the-forward-m). Could you please add a comment to the effect that the docstring is negating the original source question for clarity?

mo271:
@smmercuri I now 
 - cite the original mathover question (almost) verbatim
 -  the only place where I don't cite it verbatim is adding "$n>1$", because for $n=1$ one can prove the statement, as pointed out in the original question
  - I also add a proof for the $n=1$ case

FormalConjectures/Mathoverflow/235893.lean
```diff
@@ -43,12 +43,168 @@ theorem Continuous.isConnectedMap {X Y : Type*} [TopologicalSpace X] [Topologica
     {f : X → Y} (hf : Continuous f) : IsConnectedMap f :=
   fun _ h ↦ IsConnected.image h f (Continuous.continuousOn hf)
 
+
+/--
```
YaelDillies:
```suggestion
/--
```

FormalConjectures/Mathoverflow/235893.lean
```diff
@@ -43,12 +43,168 @@ theorem Continuous.isConnectedMap {X Y : Type*} [TopologicalSpace X] [Topologica
...
+    simp [g]
+  rw [hf_eq]
+  have h1 : IsConnectedMap h := isConnectedMap_homeomorph h
+  have h2 : IsConnectedMap h.symm := isConnectedMap_homeomorph h.symm
+  change IsConnectedMap (h.symm ∘ g.symm ∘ h)
+  exact isConnectedMap_comp (isConnectedMap_comp h1 hg_symm_conn) h2
+
+
+
 /--
```
YaelDillies, line 198:
```suggestion

/--
```

YaelDillies (approved):

PR #4023, opened 2026-05-19
fix: correct arXiv shield text

pandaninjas:
It was previously 1234.56789, corrected to the actual arXiv identifier


mo271 (approved):

Paul-Lez:
~~Whelp, that was an embarrassing typo:)~~ (never mind)

eric-wieser:
At least it was the image not the link!

Paul-Lez:
Oh never mind I had misread the diff and thought I had put the wrong link in the original citation info I added:)

PR #3691, opened 2026-04-03
chore: bump to v4.29.1

YaelDillies:
Depends on #2356

PR #4025, opened 2026-05-20
chore: bump to v4.30.0

mo271:
depends on #2356 and #3691


docbuild/lakefile.toml
mo271, line 1: do we need to `lake update` here as well to get a fresh `lake-manifest.json`

YaelDillies (approved):

PR #4028, opened 2026-05-20
fix(ErdosProblems/50): missing assumption

mo271:


mo271:
otherwise we can disproof as follows:
```lean
theorem erdos_50 : answer(False) ↔ ∀ᵉ (f : ℝ → ℝ) (hf : IsDistributionOfPhiRatio f),
    ¬∃ x, ∃ y > 0, HasDerivAt f y x := by
  /-
  **Proof Outline:**
  The statement asks if it is true that for all $f$ satisfying `IsDistributionOfPhiRatio f`, there is no $x$ where $f'(x) > 0$.
  We prove the answer is False.
  The condition `IsDistributionOfPhiRatio f` only constrains the values of $f$ on the interval $[0, 1]$.
  By Schoenberg's theorem (`erdos_50_schoenberg`), there exists at least one such function $f_0$.
  We can construct a new function $f$ by setting $f(x) = f_0(x)$ for $x \le 1$ and $f(x) = x$ for $x > 1$.
  This $f$ still satisfies `IsDistributionOfPhiRatio f` because it agrees with $f_0$ on $[0, 1]$.
  However, for $x = 2$, $f(x) = x$ in a neighborhood of $2$, so $f'(2) = 1 > 0$.
  Thus, we have found an $f$ and an $x$ where the derivative exists and is positive, making the right-hand side false.
  -/
  constructor
  · intro h
    exfalso
    exact h
  · intro h
    have ⟨f_0, hf_0⟩ := erdos_50_schoenberg
    let f (x : ℝ) := if x ≤ 1 then f_0 x else x
    have hf : IsDistributionOfPhiRatio f := by
      intro c hc
      have : f c = f_0 c := by
        dsimp [f]
        rw [if_pos hc.2]
      rw [this]
      exact hf_0 c hc
    have h_not := h f hf
    apply h_not
    use 2, 1
    constructor
    · norm_num
    · have h1 : HasDerivAt id (1 : ℝ) (2 : ℝ) := hasDerivAt_id (2 : ℝ)
      refine HasDerivAt.congr_of_eventuallyEq h1 ?_
      filter_upwards [Ioi_mem_nhds (by norm_num : (1 : ℝ) < 2)] with x hx
      dsimp [f]
      rw [if_neg (not_le_of_gt hx)]
  ```

File: FormalConjectures/ErdosProblems/50.lean [hunk 1]
```diff
@@ -73,7 +73,7 @@ that the derivative $f'(x)$ exists and is positive?
 -/
 @[category research open, AMS 11]
 theorem erdos_50 : answer(sorry) ↔ ∀ᵉ (f : ℝ → ℝ) (hf : IsDistributionOfPhiRatio f),
-    ¬∃ x, ∃ y > 0, HasDerivAt f y x := by
+    ¬∃ x ∈ Ioo (0 : ℝ) 1, ∃ y > 0, HasDerivAt f y x := by
```
YaelDillies: This excludes the endpoints, right? Do we know anything about the derivative there mathematically?

I believe a more principled fix is the following
```suggestion
    ¬∃ x, ∃ y > 0, HasDerivWithinAt f y (Ioo 0 1) x := by
```
I believe works. What happens to `HasDerivWithinAt` when `x \notin Ioo 0 1`

mo271: I think this doesn't work: 
 if $x \notin [0, 1]$, then `HasDerivWithinAt f y (Ioo 0 1) x` holds for any $y > 0$.

```lean
/--
A buggy formulation of the conjecture, which is trivially false because
if $x \notin [0, 1]$, the derivative within $(0, 1)$ at $x$ is vacuously true.
-/
theorem erdos_50.variants.buggy :
    answer(False) ↔ ∀ (f : ℝ → ℝ), IsDistributionOfPhiRatio f →
      ¬∃ x, ∃ y > 0, HasDerivWithinAt f y (Ioo (0 : ℝ) 1) x := by
  constructor
  · intro h; exact False.elim h
  · intro h
    obtain ⟨f, hf⟩ := erdos_50_schoenberg
    have h_not := h f hf
    apply h_not
    use 2, 1
    refine ⟨by norm_num, ?_⟩
    apply HasFDerivWithinAt.of_notMem_closure
    rw [closure_Ioo (by norm_num)]
    simp only [mem_Icc, not_and, not_le]
    intro _
    norm_num
```

FormalConjectures/ErdosProblems/50.lean
```diff
@@ -73,7 +73,7 @@ that the derivative $f'(x)$ exists and is positive?
 -/
 @[category research open, AMS 11]
 theorem erdos_50 : answer(sorry) ↔ ∀ᵉ (f : ℝ → ℝ) (hf : IsDistributionOfPhiRatio f),
-    ¬∃ x, ∃ y > 0, HasDerivAt f y x := by
+    ¬∃ x, ∃ y > 0, HasDerivWithinAt f y (Ioo 0 1) x := by
```
mo271:
```suggestion
    ¬∃ x ∈ Ioo (0 : ℝ) 1, ∃ y > 0, HasDerivAt f y x := by
```

FormalConjectures/ErdosProblems/50.lean
YaelDillies: What about instead
```suggestion
    ¬∃ x ∈ Icc (0 : ℝ) 1, ∃ y > 0, HasDerivWithinAt f y (Icc 0 1) x := by
```

mo271: That works and is cleaner!

YaelDillies (approved):

PR #4029, opened 2026-05-20
feat(ErdosProblems): 696

### LINKED ISSUE #2124: Erdős Problem 696: Sequence Chains in Divisors
### What is the conjecture

 Define $h(n)$ as the largest $\ell$ such that there exists a sequence of primes $p_1 < \cdots < p_\ell$ all dividing $n$ where $p_{i+1} \equiv 1 \pmod{p_i}$ for each $i$. Define $H(n)$ as the largest $u$ such that there exists a sequence of integers $d_1 < \cdots < d_u$ all dividing $n$ where $d_{i+1} \equiv 1 \pmod{d_i}$ for each $i$.

The open problems are:
1. Estimate the growth rates of $h(n)$ and $H(n)$ as $n \to \infty$.
2. Determine whether $H(n)/h(n) \to \infty$ for almost all $n$.

Erdős conjectured that the typical order of $h(n)$ is the iterated logarithm $\log^*(n)$.

 **Sources:**
 - https://www.erdosproblems.com/696

### Prerequisites needed

**Formalizability Rating:** 2/5 (0 is best) (as of 2026-02-02)

Building blocks (from Mathlib):
- Basic divisibility and modular arithmetic structures
- Prime factorization and divisor sequences
- Asymptotic analysis framework for arithmetic functions

Missing pieces:
- Formal definitions of $h(n)$ and $H(n)$ as maximum chain lengths under the specified congruence conditions
- Infrastructure for reasoning about "almost all" $n$ (density arguments in number theory)

Rating justification: The statement involves standard number-theoretic concepts (divisors, congruences, primes) that exist in Mathlib, but the specific chains $h(n)$ and $H(n)$ require custom definitions. The growth-rate questions are asymptotic in nature and would benefit from existing Mathlib asymptotic analysis, but require careful formalization of the "almost all" quantification in number theory.

* ams-11

### Choose either option

---

MavenRain:
Adds `FormalConjectures/ErdosProblems/696.lean` formalising the headline open
  question of [Erdős Problem 696](https://www.erdosproblems.com/696): the chains
   $d_1 < \cdots < d_u$ of divisors of $n$ with $d_{i+1} \equiv 1 \pmod{d_i}$.

  ## Scope

  Only the **H/h → ∞ for almost all n** question is formalised here.  The Erdős
  sub-conjecture about $h(n)$ having typical order $\log^*(n)$ is deferred to a
  follow-up so that this PR does not have to ship a clean formalisation of the
  iterated logarithm.

  ## Definitions
  - `ValidChain n P s` — `s : List ℕ` is strictly increasing, consecutive
  elements satisfy `b % a = 1`, every element divides `n` and satisfies the
  auxiliary predicate `P`.
  - `h n = Nat.findGreatest (∃ s, s.length = k ∧ ValidChain n Nat.Prime s) n` —
  restricted to prime chain elements.
  - `H n = Nat.findGreatest (∃ s, s.length = k ∧ ValidChain n (fun _ => True) s)
   n` — no primality restriction.

  ## Note on prior attempts

  PRs #2140 (failed CI) and #2157 (passed CI, silently closed by author within
  an hour) previously attempted this problem; both bundled an unrelated
  `SumOfTwoSquares.lean` file and the latter contained a `def L (x : ℝ) : ℝ :=
  sorry`, i.e. `sorry` inside a definition rather than a theorem.  This v3 is
  single-file and contains no `sorry` in any definition; all `sorry` appears
  only in the `answer(sorry) ↔ ...` theorem body.

  Closes #2124.


FormalConjectures/ErdosProblems/696.lean
```diff
@@ -0,0 +1,75 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+import FormalConjecturesForMathlib.Data.Set.Density
```
mo271:
```suggestion
```

FormalConjectures/ErdosProblems/696.lean
```diff
@@ -0,0 +1,75 @@
...
+  Nat.findGreatest (fun k => ∃ s : List ℕ, s.length = k ∧ ValidChain n Nat.Prime s) n
+
+/--
+$H(n)$: the largest length of a `ValidChain` of $n$ with no primality restriction.
+-/
+noncomputable def H (n : ℕ) : ℕ :=
+  Nat.findGreatest (fun k => ∃ s : List ℕ, s.length = k ∧ ValidChain n (fun _ => True) s) n
+
+/--
+**Erdős Problem 696 (open).**  Does $H(n)/h(n) \to \infty$ for almost all $n$?
```
mo271:
```suggestion
Does $H(n)/h(n) \to \infty$ for almost all $n$?
```

FormalConjectures/ErdosProblems/696.lean
```diff
@@ -0,0 +1,75 @@
...
+$p_1 < \cdots < p_\ell$ all dividing $n$ where $p_{i+1} \equiv 1 \pmod{p_i}$ for each $i$.
+Define $H(n)$ as the largest $u$ such that there exists a sequence of integers
+$d_1 < \cdots < d_u$ all dividing $n$ where $d_{i+1} \equiv 1 \pmod{d_i}$ for each $i$.
+
+**Open problem.**  Is it the case that $H(n)/h(n) \to \infty$ for almost all $n$ (i.e., on a
+subset of $\mathbb{N}$ of density $1$)?
+
+Erdős also conjectured that the typical order of $h(n)$ is the iterated logarithm
+$\log^*(n)$, but this PR formalises only the headline ratio question; pinning down a clean
+formalisation of $\log^*$ is left to a follow-up.
```
mo271: let's not put the docstring here: instead put them to the defintoins and theorem statements directly. Also: please cite erdosproblems.com verbatim instead of rephrasing.

FormalConjectures/ErdosProblems/696.lean
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Erdos696
+
+/--
+A strictly-increasing chain of natural numbers $d_1 < d_2 < \dots$ such that consecutive
+elements satisfy $d_{i+1} \equiv 1 \pmod{d_i}$, and every element divides $n$ and satisfies
+the auxiliary predicate $P$.
+-/
+def ValidChain (n : ℕ) (P : ℕ → Prop) (s : List ℕ) : Prop :=
+  s.IsChain (fun a b => a < b ∧ b % a = 1) ∧ ∀ d ∈ s, d ∣ n ∧ P d
```
mo271:
```suggestion
  s.IsChain (fun a b => a < b ∧ b ≡ 1 [MOD a]) ∧ ∀ d ∈ s, d ∣ n ∧ P d
```
I think this is better to avoid some suble differences for `a=1`: 

`b % 1` is always 0 (since any number divided by 1 leaves a remainder of 0). This means b % 1 = 1 evaluates to False for all b.
`b ≡ 1 [MOD 1]` translates to "1 divides (b - 1)". Since 1 divides any integer, this evaluates to True for all b.
we want the latter here (although aymptotically is probably doesn't make a difference

mo271 (commented):
Thanks! Left a few comments..

MavenRain:
Thanks @mo271, addressed all four:
   - Dropped the explicit FormalConjecturesForMathlib.Data.Set.Density import (it's already pulled in via ProblemImports).
   - Removed the prose module docstring and moved the statements directly onto the definitions and theorem, citing erdosproblems.com/696 verbatim.
   - Switched the chain relation to b ≡ 1 [MOD a] per your a = 1 point (b % 1 = 1 is never true, whereas b ≡ 1 [MOD 1] holds, matching the intended pmod congruence).

mo271:
can you click "Resolve conversation" for those that are resolved?

MavenRain:
> can you click "Resolve conversation" for those that are resolved?

Done

MavenRain:
Hi @mo271 . . . friendly nudge here. I've resolved the conversations as requested, and your four review points are addressed in 8647b01.  I just synced main so it's green and mergeable.  It's ready for another look whenever you get the chance.  Thanks!

FormalConjectures/ErdosProblems/696.lean
```diff
@@ -0,0 +1,63 @@
...
+noncomputable def H (n : ℕ) : ℕ :=
+  Nat.findGreatest (fun k => ∃ s : List ℕ, s.length = k ∧ ValidChain n (fun _ => True) s) n
+
+/--
+Is it true that $H(n)/h(n)\to \infty$ for almost all $n$?
+
+Formalised as: for every threshold $M$, the set of $n$ with $h(n) > 0$ and $H(n)/h(n) > M$
+has natural density $1$.
+-/
+@[category research open, AMS 11]
```
mo271: can you check erdosproblems.com? It seems like this has been disproved in the meantime? hence `answer(False) ↔ ...` and possibly link a lean4 proof using the `formal_proof using lean4 at [link]` mechanism. You can also check one of those formalisations and compare with what is done here.

mo271 (commented):
Thanks!

Potentially add a test a category test lemma (e.g., calculating $h(1)$) to verify the definitions.

FormalConjectures/ErdosProblems/696.lean
MavenRain: Thanks @mo271, good catch.  I checked erdosproblems.com/696: it's now marked
**SOLVED (LEAN)**, with a full Lean solution by Jake Mallen and David Turturean
(Aristotle / Claude Code / Claude Opus 4.7 / GPT-5.5 Pro).  The resolution disproves
the conjecture — it is *not* the case that H(n)/h(n) → ∞ for almost all n.

I've updated the statement to `answer(False)` and kept the density RHS as the
formalisation of "the ratio tends to infinity for almost all n", so the theorem now
records that this fails:

```
answer(False) ↔ ∀ M : ℝ, {n : ℕ | (h n : ℝ) > 0 ∧ (H n : ℝ) / (h n : ℝ) > M}.HasDensity 1
```

I compared my definitions against that formalisation and they line up: h over the
primes dividing n, H over all divisors, chain relation `d_{i+1} ≡ 1 [MOD d_i]`.
Pushed to `feat/erdos-696`.   Ready for another look, thanks!

PR #4032, opened 2026-05-21
feat(ErdosProblems): 16

danielchin:
Formalizes Erdos Problem 16: https://www.erdosproblems.com/16

Might as well get my own formalization in :)

### DIFF
--- a/FormalConjectures/ErdosProblems/16.lean
+++ b/FormalConjectures/ErdosProblems/16.lean
@@ -1,0 +1,89 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/16.lean
```diff
@@ -0,0 +1,89 @@
...
+Is the set of odd integers not of the form $2^k+p$ the union of an infinite arithmetic progression
+and a set of density $0$?
+
+Erdős called this conjecture "rather silly".
+
+Chen [Ch23] has proved the answer is no.
+
+This was formalized in Lean by Chin using Aristotle.
+-/
+@[category research solved, AMS 11, formal_proof using lean4 at "https://github.com/danielchin/proofs/blob/main/Proofs/ErdosProblems/Erdos16.lean"]
```
mo271, line 57: That is great!
if it is easy, then potentially this could be upgraded to `formal_proof using formal_conjectures`
either by adapting the formal statement here or adapting the proof in your repo to fill the sorry here.
but perhaps best in a follow up

mo271 (approved):
Thanks, LGTM!

FormalConjectures/ErdosProblems/16.lean
danielchin: Ah I see. I have to take a look at how difficult it would be to adapt the proof in my personal repo. I prefer the variable names of this current PR over the variable names from Aristotle. 

But let's submit this for now and I'll follow up in a separate PR depending on what's easier.

PR #4033, opened 2026-05-21
fix(ErdosProblems): status for 90 

smmercuri:
fixes #4027 

Proved in the negative ([solution paper](https://cdn.openai.com/pdf/74c24085-19b0-4534-9c90-465b8e29ad73/unit-distance-proof.pdf))


FormalConjectures/ErdosProblems/90.lean
```diff
@@ -71,8 +71,8 @@ noncomputable def maxUnitDistances (n : ℕ) : ℕ :=
 Does every set of $n$ distinct points in $\mathbb{R}^2$ contain at most
 $n^{1+O(\frac{1}{\log\log n})}$ many pairs which are distance $1$ apart?
 -/
```
mo271, line 79:
```suggestion
$n^{1+O(\frac{1}{\log\log n})}$ many pairs which are distance $1$ apart?

This was [disproved](https://cdn.openai.com/pdf/74c24085-19b0-4534-9c90-465b8e29ad73/unit-distance-proof.pdf) by an internal model at OpenAI, which constructed (for infinitely many 𝑛) a set 𝑃 of 𝑛 points in ℝ2 such that the number of unit distance pairs in 𝑃 is at least 𝑛1+𝑐, where 𝑐 >0 is an absolute constant.
-/
```
Let's mention link to the disproof

mo271 (approved):

mo271:
in the suggestion, I unfortunately didn't use latex markdown, nor the 100 chars line limit, please fix...

PR #4038, opened 2026-05-22
feat(Bugeaud): Problem 10.6

rwst:
Problem 10.6 of Bugeaud's collection presents a conjecture on density of lacunary sequences mod 1.


FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_6.lean
```diff
@@ -0,0 +1,43 @@
...
+-/
+
+namespace Bugeaud
+
+/--
+Problem 10.6. Find a very rapidly increasing sequence $(m_n)_{n \ge 1}$ of positive
+integers such that $(\{\xi m_n\})_{n \ge 1}$ is dense modulo one for every irrational
+number $\xi$.
+
+Here "very rapidly increasing" is formalised by requiring that $(m_n)$ is lacunary,
```
Paul-Lez: Hey! Just to double check, is this the meaning that is used in the paper?

rwst: This problem has no paper associated so I assume it's posed by Bugeaud. It is he who wrote "very rapidly increasing". In problem 10.5 he used "lacunary sequence" which has a well-known definition. I don't know why, in 10.6, he didn't use "lacunary". It would have made sense because the problem is related to 10.5.

rwst: OK I see now what's needed.

rwst: +awaiting-author

rwst:
The "superpolynomial" condition used in commit https://github.com/google-deepmind/formal-conjectures/pull/4038/commits/029345428fae094807a8e2137b09645897096142:
```
(∀ k : ℕ, ∀ᶠ (n : ℕ) in Filter.atTop, (n : ℝ) ^ k ≤ (m n : ℝ)) 
```
is not enough. Aristotle constructed the sequence:
```
/-- Block starting values: grows roughly as `exp(k/log k)`. -/
def blockStart : ℕ → ℕ
  | 0 => 2
  | k + 1 => blockStart k + (k + 1) + blockStart k / (Nat.log 2 (k + 2) + 1)

/-- Triangular number: `tri k = k*(k+1)/2` -/
def tri (k : ℕ) : ℕ := k * (k + 1) / 2

/-- Block index for position `n`: the unique `k` with `tri k ≤ n < tri (k+1)` -/
def blockIdx (n : ℕ) : ℕ := (Nat.sqrt (8 * n + 1) - 1) / 2

/-- Offset within the block for position `n` -/
def blockOff (n : ℕ) : ℕ := n - tri (blockIdx n)

/-- The sequence: element at position `n` is `blockStart(blockIdx n) + blockOff n` -/
def mySeq (n : ℕ) : ℕ := blockStart (blockIdx n) + blockOff n
```
and proved the density theorem. This is clever but certainly not what Bugeaud had in mind with "very rapidly increasing". Opus proposes to use one of the two following conditions that would improve on Furstenberg's theorem:
```
-- (a) Force the ratios to be bounded below by something nontrivial:
def IsGenuinelySublacunary (m : ℕ → ℕ) : Prop :=
  ∃ c > 0, ∀ᶠ n in atTop, (1 + c / Real.log n) ≤ (m (n+1) : ℝ) / m n

-- (b) Or: force m_n itself to grow faster than Furstenberg's benchmark:
def BeatsFurstenberg (m : ℕ → ℕ) : Prop :=
  ∃ α > (1/2 : ℝ), ∀ᶠ n in atTop, Real.exp ((n : ℝ) ^ α) ≤ m n
```
I will include both versions of the open  problem.

FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_6.lean
```diff
@@ -0,0 +1,151 @@
...
+    Vol. 193. Cambridge University Press, 2012. Chapter 10.
+  - [Fur67] Furstenberg, H. "Disjointness in ergodic theory, minimal sets, and a problem
+    in diophantine approximation". Math. Systems Theory 1, 1–49 (1967).
+  - [Mat80] de Mathan, Bernard. "Numbers contravening a condition in density modulo 1."
+    Acta Mathematica Hungarica 36.3-4 (1980): 237-241.
+  - [Pol79] Pollington, Andrew Douglas. "On the density of sequence $\{n_ {k}\xi\} $."
+    Illinois Journal of Mathematics 23.4 (1979): 511-515.
+-/
+
+namespace Bugeaud06
```
mo271, line 35: in the other files for this book, it was just `Bugeaud`, perhaps best to change it in the other files for consistency.

mo271 (approved):
Thanks, looks good to me, just two nits: one about the namespace and the other: 
`IsGenuinelySublacunary` and `HasIntermediateGrowth` are missing docstrings.

rwst:
-awating-author

rwst:
-awaiting-author

mo271:
Thanks again!

PR #4039, opened 2026-05-22
 feat(extract_names): add `answerKind`

mo271:


Paul-Lez:
Presumably we don't want to add the massive json file here?

scripts/extract_names.lean
```diff
@@ -54,10 +77,34 @@ def nameAny (n : Name) (p : String → Bool) : Bool :=
 def isInternal (n : Name) : Bool :=
   nameAny n (fun s => s.startsWith "_" || s.startsWith "match_" || s.startsWith "proof_")
 
+/-- Recursively find the subexpression with the `answer` annotation, if any. -/
+partial def findAnswerExpr : Expr → Option Expr
```
Paul-Lez: I think this should go in the `Answer.lean` file.
Note also that you could implement this using [Lean.Expr.find?](https://leanprover-community.github.io/mathlib4_docs/Lean/Util/FindExpr.html#Lean.Expr.find?).
One other note: we discussed supporting several answers in a single statement - is this something you'd want here?

Paul-Lez (approved):

PR #4056, opened 2026-05-26
feat(Books/Bugeaud): Problem 10.7

rwst:
Problem 10.7 of Bugeaud's collection is an open question about Powers of Non-Pisot Numbers.


Paul-Lez (approved):
Thanks!

PR #4059, opened 2026-05-26
fix(Util/Answer): avoid module-dependent hygienic names in `answer(sorry)`

mo271:
When `answer(sorry)` was used with a non-Prop expected type (e.g. `ℕ → ℝ`), the elaborator fell through to ``postponeElabTerm (← `(by exact sorry)) ...``, which invokes Lean's built-in `sorry` macro. This macro expands to `@sorryAx _ false`, but the resulting term includes a hygienic binder name of the form `_sorry._@.FormalConjectures.Challenge._hyg.N`. Because hygiene embeds the current module name, this binder name becomes part of the theorem's *type* expression and differs between `FormalConjectures.Challenge` and `FormalConjectures.Solution` builds — causing the comparator to reject otherwise identical problem statements (e.g. Erdős problem 507).

Fix: detect the `answer(sorry)` pattern in `alwaysTrue` mode when a non-Prop expected type is already known, and instead construct the `sorryAx` application directly via `Meta.mkSorry`. This produces `mkApp2 (mkConst ``sorryAx [u]) type false` — a canonical, module-independent expression — so the theorem type is bit-for-bit identical regardless of which module the file is compiled as. The fallback to postponed elaboration is preserved for the case where no expected type is available yet.


eric-wieser:
I don't quite follow; either we use a version of comparator which understands `answer()` in which case the `sorry` contents aren't relevant, or we use a version that doesn't, in which case I don't understand why we would invoke comparator on a statement that still contains `answer(sorry)`.

mo271:
This is about using `answer(sorry)` with Comparator in the new `withAuxiliary` mode, see https://github.com/google-deepmind/formal-conjectures/pull/3512
which adds an auxiliary declaration, which in turn plays well with Comparator's definition holes -- except in some cases, described above.

mo271:
can you have a look again, @eric-wieser

Paul-Lez (approved):

PR #4060, opened 2026-05-27
feat(ErdosProblems): 1209

danielchin:
Formalizes Erdos Problem 1209: https://www.erdosproblems.com/1209


mo271 (approved):
Thanks, looks great!

PR #4075, opened 2026-05-27
Add Erdős problem 467

### LINKED ISSUE #744: Erdős Problem 467
### What is the conjecture
    
https://www.erdosproblems.com/467

  Prove the following for all large $x$: there is a choice of congruence classes $a_p$ for all primes $p\leq x$ and a decomposition $\\{p\leq x\\}=A\sqcup B$ into two non-empty sets such that, for all $n<x$, there exist some $p\in A$ and $q\in B$ such that $n\equiv a_p\pmod{p}$ and $n\equiv a_q\pmod{q}$.
  

Status: open

### Choose either option

putonthetableacat55:
Formalize Erdős problem 467 with the source’s quantifier interpretation stated explicitly.

Fixes #744.


Smetalo:
Does this resolve #744?

williamjblair (commented):

PR #4076, opened 2026-05-27
Add Erdős problem 468

### LINKED ISSUE #745: Erdős Problem 468
### What is the conjecture
    
https://www.erdosproblems.com/468

  For any $n$ let $D_n$ be the set of sums of the shape $d_1,d_1+d_2,d_1+d_2+d_3,\ldots$ where $1<d_1<d_2<\cdots$ are the divisors of $n$. 

What is the size of $D_n\backslash \cup_{m<n}D_m$?

If $f(N)$ is the minimal $n$ such that $N\in D_n$ then is it true that $f(N)=o(N)$? Perhaps just for almost all $N$?
  

Status: open

### Choose either option

putonthetableacat55:
Formalize Erdős problem 468 with the minimal-preimage condition restricted to representable values and the shared density API.

Fixes #745.


Smetalo:
Does this resolve #745?

williamjblair (commented):
`parts.ii` is unsatisfiable. It asks for `N ∈ D (f N)` at every `N`, but elements of `D n` are sums of divisors `> 1`, so `0` and `1` are in no `D n`, and nor is `4`: a lone divisor `4` would have to be the least divisor above `1`, but `2 ∣ 4`, and two or more sum to at least `5`.

So `Erdos468MinimalPreimageLittleO` is `False` and the statement reads `answer(sorry) ↔ False`. Restricting to represented `N` fixes it:

```lean
∀ N : ℕ, (∃ n, N ∈ D n) → IsMinimalDPreimage N (f N)
```

`variants.almost_all` is fine; a density-one set can drop `1` and `4`.

Note: `FormalConjectures.Util.ProblemImports` was deleted in #4433, so this needs `import FormalConjecturesUtil`, and `open Classical` trips `linter.style.openClassical`.

PR #4077, opened 2026-05-27
Add Erdős problem 472

### LINKED ISSUE #747: Erdős Problem 472
### What is the conjecture
    
https://www.erdosproblems.com/472

  Given some initial finite sequence of primes $q_1<\cdots<q_m$ extend it so that $q_{n+1}$ is the smallest prime of the form $q_n+q_i-1$ for $n\geq m$. Is there an initial starting sequence so that the resulting sequence is infinite?
  

Status: open

### Choose either option

putonthetableacat55:
Formalize Erdős problem 472 with a direct existence statement for the infinite prime extension.

Fixes #747.


Smetalo:
Does this resolve #747?

williamjblair (commented):

Issue #778, opened 2025-10-06
Erdős Problem 538


mo271:
### What is the conjecture
    
https://www.erdosproblems.com/538

  Let $r\geq 2$ and suppose that $A\subseteq\\{1,\ldots,N\\}$ is such that, for any $m$, there are at most $r$ solutions to $m=pa$ where $p$ is prime and $a\in A$. Give the best possible upper bound for
$$\sum_{n\in A}\frac{1}{n}.$$
  

Status: open

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

Paul-Lez:
Triage: Erdős Problem 538 is now represented in `FormalConjectures/ErdosProblems/538.lean`.

PR #4079, opened 2026-05-27
Add Erdős problem 538

putonthetableacat55:
Formalize Erdős problem 538 in FormalConjectures/ErdosProblems/538.lean.


Smetalo:
Does this resolve #778?

mo271:
@putonthetableacat55 you opened a lot of PRs and haven't been responsive in the reviews that they have received. Do you plan to work on them further?

williamjblair (commented):
Flagging a conflict rather than judging it: #4588 also adds `FormalConjectures/ErdosProblems/538.lean` and is approved. Mine states the headline as `answer(sorry) ↔ ...` over `∀ r ≥ 2` and adds a solved companion with a `formal_proof` link; this one asks for the exact extremal function `B r N` at every `(r, N)`, which is a finite optimum rather than the asymptotic bound the problem asks for. A maintainer should pick.

Either way this needs the two fixes below.

`FormalConjectures.Util.ProblemImports` was deleted in #4433, so `import FormalConjecturesUtil` instead, and `open Classical` trips `linter.style.openClassical`, which is on for the problem library.

mo271:
let's got with https://github.com/google-deepmind/formal-conjectures/pull/4588! Thanks @williamjblair

PR #4080, opened 2026-05-27
Add Erdős problem 555

### LINKED ISSUE #787: Erdős Problem 555
### What is the conjecture
    
https://www.erdosproblems.com/555

  Let $R(G;k)$ denote the minimal $m$ such that if the edges of $K_m$ are $k$-coloured then there is a monochromatic copy of $G$. Determine the value of
$$R(C_{2n};k).$$
  

Status:  open

### Choose either option

putonthetableacat55:
Formalize Erdős problem 555 with the correct domain for even cycles and a direct answer-function statement.

Fixes #787.


Smetalo:
Does this resolve #787?

williamjblair (commented):

One small thing on the statement: `EvenCycleRamseyFunction` allows `n = 1`, and `EvenCycleAdj 1` is a single edge rather than a cycle. Requiring `2 ≤ n` would keep it to the graphs the problem is about.

PR #4083, opened 2026-05-27
Add Erdős problem 676

### LINKED ISSUE #871: Erdős Problem 676
### What is the conjecture
    
https://www.erdosproblems.com/676

  Is every sufficiently large integer of the form
$$ap^2+b$$
for some prime $p$ and integer $a\geq 1$ and $0\leq b<p$?
  

Status:  open

### Choose either option

putonthetableacat55:
Formalize the main eventual representation question in Erdős problem 676 directly, with the source references and further variants recorded.

Fixes #871.


Smetalo:
Does this resolve #871?

FormalConjectures/ErdosProblems/676.lean
```diff
@@ -0,0 +1,53 @@
...
+        b < p ∧
+          n = a * p ^ 2 + b
+
+/-- The main conjecture from Erdős problem 676. -/
+def Erdos676Conjecture : Prop :=
+  ∀ᶠ n in atTop, PrimeSquareRemainderRepresentation n
+
+/-- Is every sufficiently large integer of the form `a p^2 + b`, `0 ≤ b < p`? -/
+@[category research open, AMS 11]
+theorem erdos_676 : answer(sorry) ↔ Erdos676Conjecture := by
```
danielchin, line 48: You probably just inline Erdos676Conjecture and PrimeSquareRemainderRepresentation.

FormalConjectures/ErdosProblems/676.lean
```diff
@@ -0,0 +1,53 @@
...
+  sorry
+
```
danielchin, line 50: Could you also include the variants of this problem listed on the Erdos Problem website?

I see these:
`The sieve of Eratosthenes implies that almost all integers are of this form, and the Brun-Selberg sieve implies the number of exceptions in [1,𝑥] is  ≪𝑥/(log⁡𝑥)𝑐 for some constant 𝑐 >0.`

`What if the condition that 𝑝 is prime is omitted? Selfridge and Wagstaff made a 'preliminary computer search' and suggested that there are infinitely many 𝑛 not of this form even without the condition that 𝑝 is prime. It should be true that the number of exceptions in [1,𝑥] is  <𝑥𝑐 for some constant 𝑐 <1.`

`Most generally, given some infinite set 𝐴 ⊆ℕ and function 𝑓 :𝐴 →ℕ one can ask for sufficient conditions on 𝐴 and 𝑓 that guarantee every large number (or almost all numbers) can be written as 𝑎⁢𝑚2+𝑏 for some 𝑚 ∈𝐴 and 𝑎 ≥1 and 0 ≤𝑏 <𝑓⁡(𝑚).`

`In another direction, one can ask what is the minimal 𝑐𝑛 such that 𝑛 can be written as 𝑛 =𝑎⁢𝑝2 +𝑏 with 0 ≤𝑏 <𝑐𝑛⁢𝑝 for some 𝑝 ≤√𝑛. This problem asks whether 𝑐𝑛 ≤1 eventually, but in [Er79d] Erdős suggests that in fact lim sup𝑐𝑛 =∞.`

`Is it true that 𝑐𝑛 <𝑛𝑜⁡(1)?`

FormalConjectures/ErdosProblems/676.lean
```diff
@@ -0,0 +1,53 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 676
+
+*Reference:* [erdosproblems.com/676](https://www.erdosproblems.com/676)
```
danielchin, line 22:
```suggestion
*References:* 
- [erdosproblems.com/676](https://www.erdosproblems.com/676)
- [Er79] Erdős, Paul, Some unconventional problems in number theory. Math. Mag. (1979), 67-70.
- [Er79d] Erdős, P., Some unconventional problems in number theory. Acta Math. Acad. Sci. Hungar. (1979), 71-80.
- [Er80] Erdős, Paul, A survey of problems in combinatorial number theory. Ann. Discrete Math. (1980), 89-115.
```

danielchin (changes requested):

FormalConjectures/ErdosProblems/676.lean
mo271: or leave a TODO

PR #4084, opened 2026-05-27
Add Erdős problem 679

### LINKED ISSUE #873: Erdős Problem 679
### What is the conjecture
    
https://www.erdosproblems.com/679

  Let $\epsilon>0$ and $\omega(n)$ count the number of distinct prime factors of $n$. Are there infinitely many values of $n$ such that
$$\omega(n-k) < (1+\epsilon)\frac{\log k}{\log\log k}$$
for all $k<n$ which are sufficiently large depending on $\epsilon$ only?

Can one show the stronger version with 
$$\omega(n-k) < \frac{\log k}{\log\log k}+O(1)$$
is false?
  

Status:  open

### Choose either option

putonthetableacat55:
Formalize Erdős problem 679 with Mathlib’s distinct-prime-factor API and the correct quantifier dependencies.

Fixes #873.


Smetalo:
Does this resolve #873?

williamjblair (commented):

PR #4089, opened 2026-05-27
Add Erdős problem 768

### LINKED ISSUE #931: Erdős Problem 768
### What is the conjecture

https://www.erdosproblems.com/768

  Let $A\subset\mathbb{N}$ be the set of $n$ such that for every prime $p\mid n$ there exists some $d\mid n$ with $d>1$ such that $d\equiv 1\pmod{p}$. Is it true that there exists some constant $c>0$ such that for all large $N$
$$\frac{\lvert A\cap [1,N]\rvert}{N}=\exp(-(c+o(1))\sqrt{\log N}\log\log N).$$

Status:  open

### Choose either option

putonthetableacat55:
Formalize the asymptotic question in Erdős problem 768 with source-faithful notation and the Er74b reference.

Fixes #931.


Smetalo:
Does this resolve #931?

mo271 (commented):
Mathematically this looks good!

- Please use latex markdown instead of backticks throughout. 
- Let's not duplicate the problem in the module docstring, once is enough! (and keep verbatim citing the erdosproblems.com site)
- Add [Er74b] to the reference list

PR #4090, opened 2026-05-27
Add Erdős problem 810

### LINKED ISSUE #955: Erdős Problem 810
### What is the conjecture

https://www.erdosproblems.com/810

  Does there exist some $\epsilon>0$ such that, for all sufficiently large $n$, there exists a graph $G$ on $n$ vertices with at least $\epsilon n^2$ many edges such that the edges can be coloured with $n$ colours so that every $C_4$ receives $4$ distinct colours?

Status:  open

### Choose either option

putonthetableacat55:
Formalize the rainbow-C₄ density question in Erdős problem 810 using Mathlib’s SimpleGraph and unordered edge APIs.

Fixes #955.


Smetalo:
Does this resolve #955?

mo271 (commented):
Thanks!
Mathematically this looks good!

- Please use latex markdown instead of backticks throughout. 
- potentially add [BEGS89] and [Er91] to the reference list in the module header.
- LEt's not duplicate the problem in the module docstring, once is enough!

PR #4091, opened 2026-05-27
Add Erdős problem 969

### LINKED ISSUE #1044: Erdős Problem 969
### What is the conjecture

https://www.erdosproblems.com/969

  Let $Q(x)$ count the number of squarefree integers in $[1,x]$. Determine the order of magnitude in the error term in the asymptotic 
$$Q(x)=\frac{6}{\pi^2}x+E(x).$$

Status:  open

### Choose either option

putonthetableacat55:
Formalize the order-of-magnitude question for the squarefree-counting error as explicit matching upper and lower asymptotic bounds, reuse Mathlib’s `Squarefree`, and record the known PNT upper and Evelyn–Linfoot lower estimates.

Fixes #1044.


Smetalo:
Does this resolve #1044?

FormalConjectures/ErdosProblems/969.lean
```diff
@@ -0,0 +1,64 @@
...
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 969
+
+*Reference:* [erdosproblems.com/969](https://www.erdosproblems.com/969)
+
+Let `Q(x)` count the squarefree integers in `[1, x]`.  Determine the order of
+magnitude of the error term in
+`Q(x) = (6 / π^2) x + E(x)`.
```
mo271, line 26:
```suggestion
```

let's not duplicate the problem formulation in the module docstring: the theorem/def doctrings are typically enough

Also: use latex markdown instead of backticks throughout

FormalConjectures/ErdosProblems/969.lean
```diff
@@ -0,0 +1,64 @@
...
+
+noncomputable section
+
+namespace Erdos969
+
+open Classical Asymptotics
+
+/-- A natural number is squarefree if no square of a prime divides it. -/
+def SquarefreeNat (n : ℕ) : Prop :=
+  ∀ p : ℕ, Nat.Prime p → ¬ p ^ 2 ∣ n
```
mo271, line 37: redundant duplication of definition that is already in MAthlib.

FormalConjectures/ErdosProblems/969.lean
```diff
@@ -0,0 +1,64 @@
...
+  squarefreeCountingError =O[Filter.atTop] g ∧
+    g =O[Filter.atTop] squarefreeCountingError
+
+/--
+Determine the order of magnitude of the squarefree-counting error term.
+-/
+@[category research open, AMS 11]
+theorem erdos_969 :
+    {g : ℕ → ℝ | SquarefreeErrorOrderCandidate g} = answer(sorry) := by
+  sorry
```
mo271, line 60: This is not typically how we would formalize a "Determine the order of magnitude" questions: there are many other examples asking for upper and lower bound.

mo271 (changes requested):

PR #4095, opened 2026-05-28
feat(ErdosProblems): 1201

danielchin:
Formalizes Erdos Problem 1201: https://www.erdosproblems.com/1201

Not sure if we have a more standard way of defining density.


mo271 (approved):
Thanks, looks great!

PR #4102, opened 2026-05-28
fix(OEIS/80170): sign error

mo271:


mo271:
@smmercuri has been fixed before here https://github.com/google-deepmind/formal-conjectures/pull/3679, but I think in the wrong direction....

smmercuri (approved):

PR #4103, opened 2026-05-28
fix(ErdosProblems/329): mark variant as solved

mo271:


mo271:
@callesonne also a misformalisation because of missing $n>0$ condition...

YaelDillies (commented):
It is a bit silly to leave implications whose antecedent/consequent we know to be false, no? Should we just remove them?

mo271:
perhaps a bit silly, but mathematically ok?
We do explain it in the docstring, and I would like to keep `converse_implication`, because, if possible beacuse it is  it is in the bench set

YaelDillies:
What about `of_sub_perfectDifferenceSet` then? Are you happy to drop that one?

Are we intending to freeze the parts of the benchmark that made it to the bench set?

mo271:
> What about `of_sub_perfectDifferenceSet` then? Are you happy to drop that one?

yes, good point dropping it now, especially since this void now.

> 
> Are we intending to freeze the parts of the benchmark that made it to the bench set?

Perhaps just a good idea to keep around the names of the theorems from the bench set -- fixing the statement is of course ok, versioning can be done via our tags.

YaelDillies (commented):
Did you accidentally push extra stuff?

mo271:
> Did you accidentally push extra stuff?

yes, should be fixed now

YaelDillies (approved):

PR #4107, opened 2026-05-28
feat(WrittenOnTheWallII): disprove 316

mo271:


mo271:
Seems unlikely that P5 is the intended counter example, rather perhaps something else of off?

mo271:
closing in favour of #4171

PR #4137, opened 2026-05-30
Prove erdos_1054 f_undefined_at_2 (f 2 = 0)

Ace3Z:
Fills the `sorry` in `Erdos1054.f_undefined_at_2`: `f 2 = 0`.

Here `f n` is the least `m` such that `n` is a sum of the `k` smallest divisors of `m` for some `k ≥ 1`. The smallest divisor of any positive `m` is `1` and every later divisor is `≥ 2`, so such a sum is `1` (for `k = 1`) or `≥ 3`, never `2`; for `m = 0` it is `0`. Hence no `(m, k)` works and `f 2` is the junk value `0`. Cross-checked numerically: no `m < 5000` has `2` as a sum of its `k` smallest divisors.

Adds two small `@[category API]` helper lemmas about `Nat.nth (· ∈ m.divisors)`. `f_undefined_at_3` (`f 5 = 0`) is left for a follow-up.

Verification:
- `lake build FormalConjectures.ErdosProblems.«1054» --wfail` passes.
- `#print axioms … f_undefined_at_2` → `[propext, Classical.choice, Quot.sound]` (no `sorryAx`).

### DIFF
--- a/FormalConjectures/ErdosProblems/1054.lean
+++ b/FormalConjectures/ErdosProblems/1054.lean
@@ -33,6 +33,26 @@ noncomputable def f (n : ℕ) : ℕ :=
     Nat.find h
   else 0
 
+/-- The smallest divisor of a positive number is `1` (`nth` index `0`). -/
+@[category API, AMS 11]
+lemma nth_divisors_zero {m : ℕ} (hm : m ≠ 0) : Nat.nth (· ∈ m.divisors) 0 = 1 := by
+  rw [Nat.nth_zero]
+  exact IsLeast.csInf_eq ⟨Nat.one_mem_divisors.mpr hm, fun y hy => Nat.pos_of_mem_divisors hy⟩
+
+/-- Every divisor enumerated after index `0` is at least `2`. -/
+@[category API, AMS 11]
+lemma two_le_nth_divisors {m : ℕ} (hm : m ≠ 0) {i : ℕ} (hi : i ≠ 0)
+    (h : Nat.nth (· ∈ m.divisors) i ≠ 0) : 2 ≤ Nat.nth (· ∈ m.divisors) i := by
+  have hfin : (setOf (· ∈ m.divisors)).Finite := Set.toFinite _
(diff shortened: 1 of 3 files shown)


FormalConjectures/ErdosProblems/1054.lean
```diff
@@ -33,6 +33,26 @@ noncomputable def f (n : ℕ) : ℕ :=
...
+  have hfin : (setOf (· ∈ m.divisors)).Finite := Set.toFinite _
+  have hpos : 1 ≤ Nat.nth (· ∈ m.divisors) i := Nat.pos_of_mem_divisors (Nat.nth_mem_of_ne_zero h)
+  rcases hpos.lt_or_eq with h2 | h1
+  · omega
+  · exact absurd (Nat.nth_injOn hfin
+      (Set.mem_Iio.mpr (Nat.lt_card_toFinset_of_nth_ne_zero h hfin))
+      (Set.mem_Iio.mpr (Nat.lt_card_toFinset_of_nth_ne_zero
+        (show Nat.nth (· ∈ m.divisors) 0 ≠ 0 by rw [nth_divisors_zero hm]; norm_num) hfin))
+      (by rw [nth_divisors_zero hm]; omega)) hi
+
```
mo271: those seem like they would be useful - let's move them to `FormalConjecturesForMathlib`

mo271 (approved):
Thanks, looks great! Just one suggestion to move the lemmas about divisors to a better place

FormalConjectures/ErdosProblems/1054.lean
Ace3Z: Done, thanks for the suggestion. I moved both lemmas into FormalConjecturesForMathlib/NumberTheory/Divisors.lean under namespace Nat (as Nat.nth_divisors_zero and Nat.two_le_nth_divisors), and the proof in 1054 now references them from there. The new file and 1054 both build cleanly under the strict lint gate, and #print axioms reports no sorry for either lemma.

mo271 (approved):
Thanks, LGTM!

PR #4166, opened 2026-06-01
feat(Wikipedia/SparseRuler) add Wichmann conjecture

rwst:
This fills the empty file and adds necessary Mathlib definitions.


mo271 (approved):
Thanks, LGTM!
Interesting problems...

PR #4167, opened 2026-06-01
chore(docs): reimplement replaceTag using String.Slice

jeangud:
# Description
Replaces manual substring offset calculations in `docbuild/scripts/overwrite_index.lean` with the safer `String.Slice` API. 

- **Why?** Raw `String.Pos` manipulations can accidentally split multi-byte UTF-8 characters, leading to invalid strings or runtime panics. `String.Slice.split` provides a safer, higher-level abstraction that guarantees valid UTF-8 boundaries.
- Addresses `TODO(lezeau): reimplement this using String.Slice API` from @Paul-Lez 

# Testing
- Added `docbuild/scripts/test_overwrite_index.lean` with `#eval` unit tests to verify basic replacements, multiple tags, missing tags, and inverted tag cases.
- The script was also successfully validated against a local documentation build (`lake build FormalConjectures:docs`).

image


docbuild/lakefile.toml
```diff
@@ -17,6 +17,11 @@ name = "verso"
 git = "https://github.com/leanprover/verso.git"
 rev = "v4.27.0"
 
+[[lean_lib]]
```
jeangud, line 20: Required to add the test. Perhaps overkill here..

docbuild/scripts/overwrite_index.lean
```diff
@@ -65,28 +65,31 @@ def replaceTag (tag : String) (inputHtmlContent : String) (newContent : String)
   let openTag := s!"<{tag}>"
   let closeTag := s!"</{tag}>"
 
-  -- TODO(lezeau): reimplement this using String.Slice API
```
jeangud, line 68: The PR addresses this TODO 👍

mo271 (approved):
Thanks, that looks good to me! 
On the other hand I propose to 
 (1) remove the entire docgen -- we already have verso now and it seems to me that docgen adds little and is super slow
 (2) hence we don't need to "overwrite index" - we already have the stats, except for the image with the growth. This one should be moved elsewhere, perhaps to the stats page that we already have
 
 but all that can just be a follow up/different issue..

PR #4168, opened 2026-06-02
[codex] Prove practicalH values for Erdős 18

AlperTheKing:
This adds finite `practicalH` certificates for Erdős problem 18.

Main additions:
- introduces a small reflected bounded subset-sum checker over divisor lists, with soundness/completeness lemmas connecting it back to `subsetSums`;
- proves `practicalH_fiftysix : practicalH 56 = 5`;
- proves `practicalH_7200 : practicalH 7200 = 6`.

For `7200`, the upper bound uses the decomposition `7200 = 60 * 120`: quotients up to `120` are represented with at most four divisors, and residues below `60` with at most two small divisors. The lower bound uses the hard value `6667`, checked by the reflected bounded search to rule out representations using at most five divisors.

Verification:
- `lake env lean FormalConjectures\\ErdosProblems\\18.lean`


mo271:
can we use the `formal_proof using formal_conjectures at [commit]` mechanism here? The proof is a bit long...

AlperTheKing:
done.

PR #4174, opened 2026-06-03
fix(ErdosProblems/1092): definition of `f`

mo271:


FormalConjectures/ErdosProblems/1092.lean
```diff
@@ -46,27 +33,37 @@ open Asymptotics
 open Filter
 
 /--
-$f_r(n)$ is maximal such that, if a graph $G$ on $n$ vertices has the property that every
-subgraph $H$ on $m$ vertices has chromatic number $\leq r$ once we remove $f_r(m)$ edges
-from it, then $G$ has chromatic number $\leq r+1$.
+Let $f_r(m)$ is maximal such that, if any graph $G$ has the property that every subgraph $H$ on $m$
```
smmercuri:
```suggestion
Let $f_r(m)$ be maximal such that, if any graph $G$ has the property that every subgraph $H$ on $m$
```

FormalConjectures/ErdosProblems/1092.lean
```diff
@@ -46,27 +33,37 @@ open Asymptotics
...
-noncomputable def f (r n : ℕ) : ℕ :=
+noncomputable def f (r m : ℕ) : ℕ :=
   sSup {k : ℕ |
-    ∀ G : SimpleGraph (Fin n),
-      (∀ H : Subgraph G,
+    ∀ (n : ℕ) (G : SimpleGraph (Fin n)),
+      (∀ H : Subgraph G, Fintype.card H.verts = m →
         ∃ E : Finset (Sym2 H.verts),
           E.card ≤ k ∧
           chromaticNumber (H.coe.deleteEdges E) ≤ (r : ℕ∞)) →
```
smmercuri, line 49: Shouldn't we take `E` to be a subset of the edges of `H`? Currently it is just a collection of ordered pairs of vertices from `H`, so this may not correspond to actual edges of `H`.

mo271: Since we are deleting them, I don't think it really matters. However it was a bit sloppy this way: now fixing to require `E` to be a subset.

smmercuri (approved):

PR #4176, opened 2026-06-03
fix(ErdosProblems/683): add `k ≤ n / 2`

mo271:


seewoo5 (approved):

FormalConjectures/ErdosProblems/683.lean
```diff
@@ -38,11 +38,17 @@ Let $P(n, k)$ be the largest prime factor of $\binom{n}{k}$.
 def P (n k : ℕ) : ℕ := (n.choose k).primeFactors.sup id
 
 /--
-There exists $c > 0$ such that $P(n, k) > \min\{n-k+1, k^{1 + c}\}$ for all $0 < k < n$.}
+There exists $c > 0$ such that $P(n, k) > \min\{n-k+1, k^{1 + c}\}$ for all $0 < k \le n/2$.
+
+Note: the original statement in [Er79d] used $1 \le k \le n$, but as discussed on
+[erdosproblems.com](https://www.erdosproblems.com/forum/discuss/683), Erdős intended the
+problem restricted to $k \le n/2$ (the case $k > n/2$ reduces by symmetry
```
YaelDillies, line 45: No, it doesn't. Should we just remove the min(n - k + 1, ...) part?

thepriceisright:
Data point for this fix: the current statement is refutable as written, not merely broader than intended — n = 4, k = 3 makes the required inequality 2 > 2 for every c > 0, so the existential is false. Machine-checked refutation and details in this [#4896](https://github.com/google-deepmind/formal-conjectures/issues/4896#issuecomment-5490112353) comment.

One consideration for the open review question here: erdosproblems.com displays the bound with ≥, under which the (4,3) equality case is not a counterexample — so changing > to ≥ is the more local repair matching the displayed source, independent of whether the k ≤ n / 2 restriction is also adopted. Worth checking the intended reading against the primary source before merging either form.

FormalConjectures/ErdosProblems/683.lean
```diff
@@ -38,11 +38,17 @@ Let $P(n, k)$ be the largest prime factor of $\binom{n}{k}$.
...
+$\binom{n}{k} = \binom{n}{n-k}$).
 -/
 @[category research open, AMS 11]
 theorem erdos_683 : answer(sorry) ↔
-    (∃ c > (0 : ℝ), ∀ n k : ℕ, 0 < k ∧ k < n → P n k > min (n - k + 1 : ℝ) (k ^ (1 + c))) := by
+    (∃ c > (0 : ℝ), ∀ n k : ℕ, 0 < k ∧ k ≤ n / 2 →
+      P n k > min (n - k + 1 : ℝ) (k ^ (1 + c))) := by
```
mo271:
```suggestion
    ∃ c > (0 : ℝ), ∀ n k : ℕ, 0 < k ∧ k ≤ n / 2 →
      (P n k : ℝ) ≥ k ^ (1 + c) := by
```

FormalConjectures/ErdosProblems/683.lean
```diff
@@ -38,11 +38,17 @@ Let $P(n, k)$ be the largest prime factor of $\binom{n}{k}$.
 def P (n k : ℕ) : ℕ := (n.choose k).primeFactors.sup id
 
 /--
-There exists $c > 0$ such that $P(n, k) > \min\{n-k+1, k^{1 + c}\}$ for all $0 < k < n$.}
+There exists $c > 0$ such that $P(n, k) > \min\{n-k+1, k^{1 + c}\}$ for all $0 < k \le n/2$.
+
+Note: the original statement in [Er79d] used $1 \le k \le n$, but as discussed on
+[erdosproblems.com](https://www.erdosproblems.com/forum/discuss/683), Erdős intended the
+problem restricted to $k \le n/2$ (the case $k > n/2$ reduces by symmetry
+$\binom{n}{k} = \binom{n}{n-k}$).
```
mo271:
```suggestion
Let $P(n, k)$ be the largest prime factor of $\binom{n}{k}$.
There exists $c > 0$ such that $P(n, k) \ge k^{1 + c}$ for all $0 < k \le n/2$.

Erdős originally stated this for $1 \le k \le n$ with a $\min(n-k+1, k^{1+c})$ bound [Er79d],
but as discussed on [erdosproblems.com](https://www.erdosproblems.com/forum/discuss/683),
the $\min$ term was introduced only to handle $k > n/2$; the problem is naturally stated
for $k \le n/2$ (cf. [#961](https://www.erdosproblems.com/961)).
```

FormalConjectures/ErdosProblems/683.lean
mo271: yes, done

mo271:
thanks for the reviews @YaelDillies and @thepriceisright please have a look now

FormalConjectures/ErdosProblems/683.lean
```diff
@@ -38,11 +38,18 @@ Let $P(n, k)$ be the largest prime factor of $\binom{n}{k}$.
...
+Erdős originally stated this for $1 \le k \le n$ with a $\min(n-k+1, k^{1+c})$ bound [Er79d],
+but as discussed on [erdosproblems.com](https://www.erdosproblems.com/forum/discuss/683),
+the $\min$ term was introduced only to handle $k > n/2$; the problem is naturally stated
+for $k \le n/2$ (cf. [#961](https://www.erdosproblems.com/961)).
 -/
 @[category research open, AMS 11]
 theorem erdos_683 : answer(sorry) ↔
-    (∃ c > (0 : ℝ), ∀ n k : ℕ, 0 < k ∧ k < n → P n k > min (n - k + 1 : ℝ) (k ^ (1 + c))) := by
+    ∃ c > (0 : ℝ), ∀ n k : ℕ, 0 < k ∧ k ≤ n / 2 →
+      (P n k : ℝ) ≥ k ^ (1 + c) := by
```
YaelDillies: nit:
```suggestion
    ∃ c > (0 : ℝ), ∀ n k : ℕ, 0 < k ∧ k ≤ n / 2 → (P n k : ℝ) ≥ k ^ (1 + c) := by
```

YaelDillies (approved):
Looks good now!

PR #4179, opened 2026-06-04
fix(LittProblems/1): add `q ≠ 0`

mo271:


mo271:
otherwise one can disproof, see intermediate commit

seewoo5 (approved):
Thank you for catching this!

YaelDillies (approved):

PR #4182, opened 2026-06-04
fix(GreensOpenProblems/72): fix `AllowedSet`

mo271:


felixpernegger (approved):
Nice, thanks for the fix

smmercuri (approved):

PR #4189, opened 2026-06-06
feat: add graceful tree conjecture (Ringel-Kotzig)

walidelkersh:
Adds the statement of the **graceful tree conjecture** (Ringel–Kotzig): every tree with `m` edges admits an injective vertex labeling into `{0, …, m}` whose induced edge labels are exactly `{1, …, m}`. Proved for many special classes of trees; open in general.

`FormalConjectures/Wikipedia/GracefulLabeling.lean`, `@[category research open, AMS 5]`. Addresses the Ringel–Kotzig entry in #104.

### DIFF
--- a/FormalConjectures/Wikipedia/GracefulLabeling.lean
+++ b/FormalConjectures/Wikipedia/GracefulLabeling.lean
@@ -28,7 +28,7 @@ namespace GracefulLabeling
 
 open SimpleGraph
 
-lemma graceful_tree_one_vertex :
+example :
     let T : SimpleGraph Unit := ⊥
     let m := T.edgeFinset.card
     ∃ f : Unit → ℕ,
@@ -41,10 +41,10 @@ lemma graceful_tree_one_vertex :
                   rw [← Int.natAbs_neg, neg_sub]⟩) = Finset.Icc 1 m := by
   intro T m
   use fun _ => 0
-  refine ⟨fun _ _ _ => rfl, fun _ => le_refl _, ?_⟩
+  refine ⟨fun _ _ _ => rfl, fun _ => Nat.zero_le _, ?_⟩
   simp [m, T]
 
-lemma graceful_tree_two_vertex :
+example :
     let T : SimpleGraph (Fin 2) := ⊤
     let m := T.edgeFinset.card
     ∃ f : Fin 2 → ℕ,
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/GracefulLabeling.lean
```diff
@@ -0,0 +1,54 @@
...
+namespace GracefulLabeling
+
+open SimpleGraph
+
+/--
+Every tree admits a graceful labeling.
+
+A graceful labeling of a tree `T` with `m` edges is an injective map `f : V → {0, …, m}`
+such that the multiset of absolute differences `|f(u) - f(v)|` over edges `{u,v}` of `T`
+equals `{1, …, m}`.
```
mo271: let's use latex markdown instead of backticks in the doc comment.

FormalConjectures/Wikipedia/GracefulLabeling.lean
```diff
@@ -0,0 +1,54 @@
...
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Graceful Tree Conjecture (Ringel–Kotzig conjecture)
+
+*Reference:* [Wikipedia/Graceful_labeling](https://en.wikipedia.org/wiki/Graceful_labeling)
+
+A graceful labeling of a graph `G` with `m` edges is an injective vertex labeling
+`f : V → {0, …, m}` such that the induced edge labels `{|f(u) - f(v)| : {u,v} ∈ E(G)}`
+are exactly `{1, …, m}`. Conjectured by Ringel (1963) and Kotzig; formalized by Rosa (1967).
```
mo271:
```suggestion
```

Avoiding to duplicate the docstring here, since it will be harder to maintain when it is in two places

mo271 (commented):
Thanks! Looks great mathematically, jsuta  few formatting nits. 
Also: Consider adding a simple test lemma, e.g., showing that a 1-vertex or 2-vertex tree has a graceful labeling, would be a nice enhancement to verify the definitions.

walidelkersh:
@mo271 I've updated the docstring to use LaTeX formatting and removed the duplicated definition.
I also added 1-vertex and 2-vertex test lemmas (using `example`) which verify the definitions and compile successfully.

-awaiting-author

mo271 (approved):
Thanks, LGTM!

PR #4191, opened 2026-06-06
fix(Paper/WeaklyFirstCountable): missing assumption

mo271:


mo271:
@felixpernegger can you check that the added Hausdorff assumption is indeed what is needed?

Ohterwise we can find such spaces for problem 2 and 3, see intermediate commit

felixpernegger (approved):
Yes it means compact + hausdorff as often assumed in the literature (when looking at the paper a few days ago I noticed this as well)

YaelDillies (approved):
Yep, looks good

mo271:
@felixpernegger perhaps it is good to have a look through other statements where potentially the same convention of compact = compact + Hausdorff is not yet implemented correctly

felixpernegger:
See #4242

PR #4194, opened 2026-06-07
Erdős 829, 470, 456, 17, 409: discharge test/textbook scaffolding sorries (not the open problems)

narinluangrath:
This PR discharges **test- and textbook-tier `sorry`s** (the scaffolding/sanity-check lemmas in the `formal-conjectures` repo), **not** the underlying open Erdős problems. Every `@[category research open]` statement in these files is left untouched. Each item below states explicitly which `@[category test]` or `@[category textbook]` declaration is being closed.

## Problem 829 (number of representations as a sum of two cubes)
Proves the four `test` representation-count facts:
- `sumRep cubes 0 = 1`, `sumRep cubes 2 = 1`, `sumRep cubes 3 = 0`
- `sumRep cubes 1729 = 4` — the Hardy-Ramanujan taxicab number ($1729 = 1^3 + 12^3 = 9^3 + 10^3$)

Adds a helper `mem_cubes_iff` (tagged `@[category API]`) rewriting membership in `cubes` (`∃ k, k^3 = n`) to the bounded, decidable form `∃ k ∈ Finset.range (n+1), k^3 = n` (justified by `k ≤ k^3`). The small cases close by `decide`; the taxicab case uses `native_decide` (a large `test` computation, per `AGENTS.md`).

## Problem 470 (weird numbers)
Proves the `textbook` fact that **70 is the smallest weird number**: `(∀ n < 70, ¬ n.Weird) ∧ (70).Weird`.

Adds `DecidablePred` instances for `Nat.Abundant`, `Nat.Pseudoperfect`, and `Nat.Weird` (the existential in `Pseudoperfect` ranges over the finite powerset of the proper divisors), using the existing `decidable_of_iff` pattern. The bounded check `∀ n < 70, ¬ n.Weird` is discharged by **kernel `decide`** (no `native_decide`); the `70.Weird` half reuses Mathlib's `Nat.weird_seventy`.

## Problem 456 (smallest prime / smallest totient-divisor)
Proves the `textbook` fact `m n ≤ p n`, where `p n` is the smallest prime `≡ 1 (mod n)` and `m n` is the smallest `k > 0` with `n ∣ φ(k)`.

Every prime `p ≡ 1 (mod n)` has `n ∣ φ(p) = p − 1`, so the set of such primes is contained in the set defining `m n`. For `n ≠ 0`, Dirichlet (`Nat.exists_prime_gt_modEq_one`) makes that set nonempty, so `p n` itself realises the infimum and lands in the `m n` set, giving `m n ≤ p n` via `Nat.sInf_le`. The `n = 0` case is handled separately (`m 0 = 0`).

## Problem 17 (cluster primes)
Proves the `test` fact that **97 is the smallest prime that is not a cluster prime**: `IsLeast {p | p.Prime ∧ ¬ IsClusterPrime p} 97`.

A prime `p` is a *cluster prime* if every even `n ≤ p − 3` is a difference of two primes `≤ p`. Adds a `Decidable (IsClusterPrime p)` instance via a bounded reformulation — the even `n` is bounded by `n ≤ p − 3` and the witnessing primes `q₁, q₂ ≤ p` range over finite sets — then discharges the `IsLeast` statement (97 is non-cluster; every prime `< 97` is a cluster prime) by `native_decide` (a `test` computation, per `AGENTS.md`).

## Problem 409 (iterating φ(n) + 1)
Proves the `test` fact that the iteration `n ↦ φ(n) + 1` always reaches a prime: `∃ i, ((φ · + 1)^[i] n).Prime` for `n > 0`.

By strong induction on `n`: if `n` is already prime, zero steps suffice; if `n = 1`, one step gives `φ(1) + 1 = 2`; if `n ≥ 2` is composite, then `φ(n) ≠ n − 1` (`Nat.totient_eq_iff_prime`), so together with `φ(n) < n` we get `φ(n) + 1 < n` and recurse on the strictly smaller value.

### Problem 409 — additional: a proven `O(√n)` upper bound (partial result, *not* the open problem)
Adds `erdos_409.variants.isBigO_sqrt` (tagged `research solved`): if `c(n)` is the number of iterations of `n ↦ φ(n)+1` needed to reach a prime, then `c(n) = O(√n)`. This is a correct improvement over the file's recorded-trivial `o(n)`, but it is **not** a solution to the open exact-asymptotic questions `erdos_409.parts.i` / `.isTheta` / `.isBigO`, which are left untouched as `research open`.

Proof: any composite `m` has `minFac m ≤ √m`, and `φ(m) ≤ m − m/minFac m` (the `m/minFac` multiples of `minFac m` in `[0,m)` are all non-coprime to `m`), so `φ(m) + √m ≤ m` — each iteration drops the value by at least `√(current)`. A real-valued descent (`√n − √(φ(n)+1) ≥ 1/6` for `n ≥ 9`) then bounds the step count by `6√n`. Supporting API lemmas: `aux.totient_le_sub_div_minFac`, `aux.totient_add_sqrt_le`, `aux.iter_count_le_sqrt`. No `native_decide`.

### Problem 409 — additional: basins are unbounded (partial result toward open part (ii), *not* the open problem)
Adds `erdos_409.variants.basins_unbounded` (tagged `research solved`): for every `k`, **some prime is reached by at least `k` distinct starting values** under `n ↦ φ(n)+1`. This is honest progress toward `erdos_409.parts.ii` (which asks whether some prime has an *infinite* basin) but is **strictly weaker** — unbounded finite basins, not a single infinite one — and does **not** resolve it; `parts.ii` is left untouched as `research open`.

Proof: every `n ≥ 2` flows to a unique terminal prime `≤ n` (primes are fixed points), so the `N−1` values in `[2,N]` map into the primes `≤ N`, of which there are only `π(N) = o(N)` (Chebyshev's `eventually_primeCounting_le`); the finite-pigeonhole `Finset.exists_lt_card_fiber_of_mul_lt_card_of_maps_to` then forces a fiber of size `> k`. Supporting API lemmas: `aux.iterate_mem_Icc`, `aux.exists_primeCounting_mul_lt`. No `native_decide`.

### Problem 409 — additional: basin size grows at least logarithmically (quantitative version of the above)
Adds `erdos_409.variants.basin_card_ge_log` (tagged `research solved`): there is an absolute constant `c > 0` such that, for all sufficiently large `N`, **some prime is reached by at least `c·log N` distinct starting values drawn from `[2,N]`**. This sharpens `basins_unbounded` (mere unboundedness) to a logarithmic lower bound on the maximal basin size within `[2,N]`. It remains a partial result toward `erdos_409.parts.ii` (a single prime with an *infinite* basin), not a solution.

Proof: the same terminal-prime pigeonhole as above with multiplier `k = ⌊log N / 5⌋`, fed by an eventual Chebyshev bound `π(N) ≤ 3N/log N` (new helper `aux.primeCounting_le_three_mul`, with `c = 1/6`). No `native_decide`.

### Problem 409 — additional: König-lemma reduction of the open part (ii)
Adds `erdos_409.variants.ii_reduction` (tagged `research solved` — a fully-proven *equivalence*, which reformulates but does **not** resolve the open `erdos_409.parts.ii`, left untouched): the basin `B(p) = {n | ∃ i, gⁱ(n) = p}` is **infinite** iff there is an infinite strictly increasing backward chain `p = m₀ < m₁ < ⋯` with `φ(m_{j+1}) = m_j − 1`.

Proof: `(⟸)` such a chain injects into `B(p)`, so the basin is infinite. `(⟹)` König's lemma, hand-rolled via an "infinite-descendants" choice recursion. The crux is that **totient fibers `{x | φ(x) = v}` are finite** (so the backward tree is finitely branching), which I prove from a new elementary bound `n ≤ 2·φ(n)²` (`aux.le_two_mul_totient_sq`, via a strong induction on `minFac n` with an odd-case refinement `n ≤ φ(n)²` for the `n ≡ 2 (mod 4)` step). Supporting lemmas: `aux.finite_totient_fiber`, and (private) `desc`/`pre`, `finite_pre`, `desc_sub`, `konig_step`, `lt_of_pre_infinite`. No `native_decide`.

All files build cleanly with no linter warnings.

---
- [ ] @narinluangrath has verified the content above

*Disclaimer: Claude (Anthropic) was used to help generate this report.*

### DIFF
--- a/FormalConjectures/ErdosProblems/17.lean
+++ b/FormalConjectures/ErdosProblems/17.lean
@@ -69,9 +69,46 @@ theorem erdos_17.variants.upper_Elsholtz :
         (fun x ↦ x * exp (-c * (log (log x)) ^ 2)) := by
   sorry
 
+/-- `IsClusterPrime` is decidable: both the universally quantified even `n` (which is bounded by
+`n ≤ p - 3`) and the witnessing primes `q₁, q₂ ≤ p` range over finite sets, so the whole predicate
+reduces to a bounded search. -/
+instance (p : ℕ) : Decidable (IsClusterPrime p) :=
+  haveI : Decidable (p.Prime ∧ ∀ n ∈ (Finset.range (p - 2)).filter (2 ∣ ·),
+      ∃ q₁ ∈ Finset.range (p + 1), ∃ q₂ ∈ Finset.range (p + 1),
+        q₁.Prime ∧ q₂.Prime ∧ (n : ℤ) = (q₁ : ℤ) - q₂) := inferInstance
+  decidable_of_iff
+    (p.Prime ∧ ∀ n ∈ (Finset.range (p - 2)).filter (2 ∣ ·),
+      ∃ q₁ ∈ Finset.range (p + 1), ∃ q₂ ∈ Finset.range (p + 1),
+        q₁.Prime ∧ q₂.Prime ∧ (n : ℤ) = (q₁ : ℤ) - q₂) <|
+    (by
+      constructor
+      · rintro ⟨hp, h⟩
+        refine ⟨hp, ?_⟩
+        intro n hev hle
+        have hn : n ∈ (Finset.range (p - 2)).filter (2 ∣ ·) :=
+          Finset.mem_filter.mpr ⟨Finset.mem_range.mpr (by omega), hev.two_dvd⟩
+        obtain ⟨q₁, hq₁, q₂, hq₂, hq₁p, hq₂p, heq⟩ := h n hn
+        simp only [Finset.mem_range] at hq₁ hq₂
+        exact ⟨q₁, q₂, hq₁p, hq₂p, by omega, by omega, heq⟩
+      · rintro ⟨hp, h⟩
+        refine ⟨hp, ?_⟩
+        intro n hn
+        obtain ⟨hnr, hdvd⟩ := Finset.mem_filter.mp hn
+        simp only [Finset.mem_range] at hnr
+        have hle : (n : ℤ) ≤ (p : ℤ) - 3 := by omega
+        obtain ⟨q₁, q₂, hq₁, hq₂, hq₁p, hq₂p, heq⟩ := h (even_iff_two_dvd.mpr hdvd) hle
+        exact ⟨q₁, Finset.mem_range.mpr (by omega), q₂, Finset.mem_range.mpr (by omega),
+          hq₁, hq₂, heq⟩)
+
 /-- $97$ is the smallest prime that is not a cluster prime. -/
 @[category test, AMS 11]
 theorem isClusterPrime_97_isLeast_non_cluster : IsLeast {p : ℕ | p.Prime ∧ ¬ IsClusterPrime p} 97 := by
-  sorry
+  constructor
+  · exact ⟨by norm_num, by native_decide⟩
+  · have key : ∀ p < 97, p.Prime → IsClusterPrime p := by native_decide
+    rintro p ⟨hpp, hnc⟩
+    by_contra hlt
+    push_neg at hlt
+    exact hnc (key p hlt hpp)
 
 end Erdos17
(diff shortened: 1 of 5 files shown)


mo271:
It would make review easier, if you could open one pull request per sorry (or per fille) -- this makes is much easier to review.

PR #4199, opened 2026-06-09
fix(Wikipedia/AgohGiuga): Fix IsCarmichael

danielchin:
The previous formalization of `IsCarmichael` allowed for all prime numbers to satisfy those conditions. This fixes it so that $n$ can't be prime. And fixes the formalizations of the proofs that depend on `IsCarmichael`.

Assisted by Antigravity/Gemini


danielchin:
+misformalization

mo271 (approved):
Thanks, LGTM!

PR #937, opened 2025-10-10
feat(erdos): add formalization scaffold for Problem 774 (dissociated sets)

ayush31010:
New file: 774.lean
Defines IsDissociated for finite sets and IsPropDissociated for infinite sets
States the main conjecture as erdos_774_statement with a placeholder proof
Includes references and documentation for context
This PR adds the basic formalization and statement for Erdős Problem 774, following the conventions of the project.
#935

PR #4202, opened 2026-06-09
feat(ErdosProblems): 937

AlperTheKing:
Formalises Erdős Problem #937 (Erdős [Er76d]): are there infinitely many four-term arithmetic progressions of coprime powerful numbers?

The answer is YES — Bajpai, Bennett and Chan [BBC24] (Int. J. Number Theory 20 (2024), 19-45) proved there are infinitely many four-term arithmetic progressions of pairwise coprime powerful numbers. Stated as `answer(True) ↔ {(a,d) | a,a+d,a+2d,a+3d pairwise coprime and powerful}.Infinite`, category `research solved`, using the existing decidable `Nat.Powerful` predicate. Proof body `sorry` (resolution is [BBC24], not yet in Lean).


FormalConjectures/ErdosProblems/937.lean
```diff
@@ -0,0 +1,59 @@
...
+
+*Reference:* [erdosproblems.com/937](https://www.erdosproblems.com/937)
+
+Are there infinitely many four-term arithmetic progressions of coprime powerful numbers?
+(A number `n` is *powerful* if `p ∣ n → p² ∣ n`; `Nat.Powerful`.)
+
+Erdős [Er76d] asked this; Bajpai, Bennett and Chan [BBC24] proved the answer is **yes** —
+there are infinitely many four-term arithmetic progressions of pairwise coprime powerful numbers.
+(Note: without coprimality this is easy, and by a theorem of Fermat there are no four *squares*
+in arithmetic progression.)
```
mo271: let's not repeat the problem in the module docstring.

FormalConjectures/ErdosProblems/937.lean
```diff
@@ -0,0 +1,59 @@
...
+powerful numbers_, Int. J. Number Theory 20 (2024), 19-45.
+-/
+
+namespace Erdos937
+
+open Nat
+
+/-- The four numbers `a, a+d, a+2d, a+3d` form a four-term arithmetic progression (`d > 0`) of
+pairwise coprime powerful numbers. -/
+def IsCoprimePowerfulAP4 (a d : ℕ) : Prop :=
```
mo271, line 35: potentiall add a test statement, (e.g., `¬ IsCoprimePowerfulAP4 0 1`)

mo271 (commented):
Thanks! Looks great mathematically, just some nits...

FormalConjectures/ErdosProblems/937.lean
AlperTheKing: Done in e7f8eac: removed the repeated problem statement from the module docstring (kept only the reference and the [BBC24] bibliography); the statement now lives on the `erdos_937` theorem.

FormalConjectures/ErdosProblems/937.lean
AlperTheKing: Done in e7f8eac: added `not_isCoprimePowerfulAP4_zero_one : ¬ IsCoprimePowerfulAP4 0 1` (the progression 0, 1, 2, 3 fails because 2 is not powerful), as a `@[category test]` lemma.

mo271:
Thanks again!

mo271 (approved):

PR #4203, opened 2026-06-09
feat(Bugeaud): Problem 10.8 and 10.7

rwst:
This adds the `sInf`-based formulation of the *p*-adic Littlewood conjecture and two related (solved) theorems.


FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_7.lean
mo271:
```suggestion
namespace Bugeaud07
```
for consistency with the other files

mo271:
```suggestion
end Bugeaud07
```

mo271 (approved):
Thanks, looks great, only minor nit about namespace naming

rwst:
-awaiting-author

mo271:
Thanks again!

PR #4204, opened 2026-06-09
feat(Bugeaud): Problem 10.9

rwst:
This adds (different from `Wikipedia/Mahler32.lean`) formulations of Mahler's Z-number problem and two related solved theorems.

### DIFF
--- a/FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_9.lean
+++ b/FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_9.lean
@@ -34,9 +34,8 @@ See also FormalConjectures/Wikipedia/Mahler32.lean.
 namespace Bugeaud09
 
 /--
-Problem 10.9. There are no real numbers $\xi$ such that
-$0 \le \{\xi (3/2)^n\} < 1/2$ for every positive integer $n$,
-i.e. no Z-number exists. Posed by Mahler [Mah68].
+Problem 10.9. There are no real numbers $\xi$ such that $0 \le \{\xi (3/2)^n\} < 1/2$
+for every positive integer $n$, i.e. no Z-number exists. Posed by Mahler [Mah68].
 -/
 @[category research open, AMS 11]
 theorem problem_10_9 : type_of% Mahler32.mahler_conjecture := by
--- a/FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_9.lean
+++ b/FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_9.lean
@@ -1,0 +1,68 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_9.lean
```diff
@@ -0,0 +1,68 @@
...
+-/
+
+namespace Bugeaud09
+
+/--
+A *Z-number* is a positive real number $\xi$ such that the fractional part of
+$\xi (3/2)^n$ lies in $[0, 1/2)$ for every positive integer $n$.
+-/
+def IsZNumber (ξ : ℝ) : Prop :=
+  0 < ξ ∧ ∀ n : ℕ, 1 ≤ n → Int.fract (ξ * (3 / 2 : ℝ) ^ n) < 1 / 2
```
mo271: We have a very similar defin in `Mahler32.lean`:

Both files independently define what a Z-number is, with slight differences in formulation:

`Mahler32.IsZNumber x: ∀ n > 0, Int.fract (x * (3 / 2 : ℝ) ^ n) < 1 / 2` (with $x \neq 0$ placed in the theorem).
`Bugeaud09.IsZNumber ξ: 0 < ξ ∧ ∀ n : ℕ, 1 ≤ n → Int.fract (ξ * (3 / 2 : ℝ) ^ n) < 1 / 2`.

Recommendation: Rather than defining `IsZNumber` twice across the repository, `IsZNumber` should be consolidated into a single definition in `FormalConjecturesForMathlib/` (e.g., in a Diophantine approximation file), allowing both problem files to share the exact same definition.

FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_9.lean
```diff
@@ -0,0 +1,68 @@
...
+
+/--
+Problem 10.9. There are no real numbers $\xi$ such that
+$0 \le \{\xi (3/2)^n\} < 1/2$ for every positive integer $n$,
+i.e. no Z-number exists. Posed by Mahler [Mah68].
+-/
+@[category research open, AMS 11]
+theorem problem_10_9 : ¬ ∃ ξ : ℝ, IsZNumber ξ := by
+  sorry
```
mo271, line 42: `problem_10_9` could be defined as `type_of% Mahler32.mahler_conjecture` (or vice versa), or the deviation should be explicitly accepted due to the differing binding structures and variant focuses.
At least pointing to the same file in the repo might make sense.

FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_9.lean
mo271: Adding a category test lemma (e.g., proving `¬ IsZNumber 1`)  might be a good sanity check.

rwst:
-awaiting-author

FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_9.lean
```diff
@@ -0,0 +1,74 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+import FormalConjectures.Wikipedia.Mahler32
```
mo271:
```suggestion
import FormalConjecturesUtil
import FormalConjectures.Wikipedia.Mahler32
```
was updated int he meantime

mo271 (approved):
Thanks, LGTM

took a bit longer to review, thanks for you patience

PR #4211, opened 2026-06-09
fix(ErdosProblems/1146): fix Schnirelmann sumset convention

mo271:


mo271:
@danielchin does this fix look good to you?

YaelDillies (approved):

PR #4212, opened 2026-06-09
feat(ErdosProblems/20): add Erdős–Rado sunflower bound variant

bengoechea:
Adds `erdos_20.variants.erdos_rado_bound`, recording the classical Erdős–Rado (1960) factorial upper bound on the k-sunflower threshold as a variant. This addresses the in-file `TODO(firsching): add the various known bounds as variants` — a first entry; the improved Alweiss–Lovett–Wu–Zhang and Rao bounds are left for follow-ups.

Statement: `f n k ≤ (k - 1) ^ n * n.factorial + 1`, for `n > 0` and `2 ≤ k`.

The `+ 1` is deliberate. Erdős–Rado guarantees a k-sunflower only for families *strictly* larger than `(k-1)^n * n!`, while `f` is defined here with a non-strict `m ≤ F.ncard` threshold, so the minimal guaranteeing cardinality is `(k-1)^n * n! + 1`. Stating it without the `+ 1` would be strictly stronger than the cited result.

Tagged `@[category research solved]` with a `sorry` body: the bound is the classical result cited externally (ErRa60), not a new Lean proof in this file.

AI disclosure is in the file docstring.


FormalConjectures/ErdosProblems/20.lean
```diff
@@ -22,6 +22,14 @@ import FormalConjectures.Util.ProblemImports
...
 * [erdosproblems.com/20](https://www.erdosproblems.com/20)
 * [Wikipedia](https://en.wikipedia.org/wiki/Sunflower_(mathematics))
+* [ErRa60] Erdős, Paul and Rado, Richard. Intersection theorems for systems of sets.
+  J. London Math. Soc. 35 (1960), 85--90.
+
+### AI disclosure
+
+Lean 4 code in this file was drafted with assistance from OpenAI Codex and
+Claude (Anthropic). The mathematical content and references are the author's
+own work.
```
mo271:
```suggestion
```

Let's not clutter our files with that: this should go to the pull request descriptions!

mo271 (approved):
Thanks, LGTM

the `n>0` guard seems good and necessary

mo271:
Thanks again!

PR #4213, opened 2026-06-09
Add Erdős Problem 474 (bracket partition relation 2^ℵ₀ ↛ [ℵ₁]²₃)

henrykmichalewski:
## Summary

Formalises [Erdős Problem 474](https://www.erdosproblems.com/474) — a $100 question on the **bracket partition relation** $2^{\aleph_0} \not\to [\aleph_1]^2_3$:

> Under what set-theoretic assumptions can $\mathbb{R}^2$ be 3-colored such that every uncountable $A \subseteq \mathbb{R}^2$ has $A^2$ containing a pair of each color?

Status: **NOT PROVABLE** (consistency open — the question is genuinely metamathematical).

This sits next to the just-merged [#3784 (Erdős 70)](https://github.com/google-deepmind/formal-conjectures/pull/3784) in the partition-relations track (round-bracket Ramsey on the continuum) — but uses square-bracket relations.

## What's in the file

* `CardinalBracketRamsey2 α β r`: the bracket partition predicate $\alpha \to [\beta]^2_r$ on cardinals.
* `erdos_474`: the headline open problem, `@[category research open, AMS 3]`.
* Four named variants:
  - `sierpinski_kurepa`: the ZFC negative result for 2 colors.
  - `erdos_under_ch`: Erdős's positive result under CH ($\mathfrak{c} = \aleph_1$).
  - `shelah_consistent` [Sh88]: positive bracket relation is consistent with very large $\mathfrak{c}$.
  - `vaughan_aleph_two` [Va99, 7.81]: the specific still-open question at $\mathfrak{c} = \aleph_2$.
* Two sanity-check theorems for monotonicity of the relation (one fully closed, one `sorry`).

## Test plan

- [x] `lake env lean FormalConjectures/ErdosProblems/474.lean` produces only the expected `sorry` warnings (6 total).
- [x] No new ForMathlib defs — uses only existing `Cardinal` API.
- [x] Single `*Reference:*` block per project style.


FormalConjectures/ErdosProblems/474.lean
```diff
@@ -0,0 +1,161 @@
...
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 474
+
+*Reference:* [erdosproblems.com/474](https://www.erdosproblems.com/474)
+
+*References:*
+- [Er95d] Erdős, P. *Some of my favourite problems which recently have been solved.*
```
mo271:
```suggestion
*References:*
- [erdosproblems.com/474](https://www.erdosproblems.com/474)
- [Er95d] Erdős, P. *Some of my favourite problems which recently have been solved.*
```

FormalConjectures/ErdosProblems/474.lean
```diff
@@ -0,0 +1,161 @@
...
+
+The bracket form is weaker than the round-bracket Ramsey property
+$\alpha \to (\beta)^2_r$ (which would require a *monochromatic* $\beta$-sized subset);
+here we only require that the coloring restricted to the $\beta$-sized subset is not
+surjective onto the $r$ colors.
+
+The coloring is given as a symmetric function `f : X → X → Fin r`; the value on a
+diagonal `f x x` is irrelevant (we never quantify over diagonal pairs).
+-/
+def CardinalBracketRamsey2 (α β : Cardinal.{u}) (r : ℕ) : Prop :=
```
mo271: seems like this should go to ForMathlib: it's a general bracket partition relation, not problem-specific.

mo271 (commented):
Thanks for the contribution!

This looks good mathematically, just two nits (one about moving a definition to `FormalConjecturesForMathlib`

FormalConjectures/ErdosProblems/474.lean
henrykmichalewski: Done in 7299936 — moved the definition into `FormalConjecturesForMathlib/Combinatorics/SetTheory/PartitionRelation.lean` as `Combinatorics.cardinalBracketRamsey2`, right alongside the existing `cardinalPartitionRel` (the round-bracket relation). The problem file now just `open Combinatorics` and uses it. Also consolidated the duplicated reference block into one `*References:*` section while here.

PR #4216, opened 2026-06-09
Add finite Erdos 23 n=5 consequence

AlperTheKing:
## Summary
- Adds the finite `n = 5` / 25-vertex consequence of Erdos problem 23 as a solved variant.
- Adds the matching tightness statement from the balanced `C_5` blow-up.
- Adds references to Balogh-Clemen-Lidicky and McKay's extremal graph catalogue.

## Mathematical note
The new upper bound can be proved from BCL high-density plus McKay/OEIS `a(23)=20` data: a hypothetical 25-vertex triangle-free graph with `beta >= 26` has `e <= 95`; deleting two degree-`<= 7` vertices gives a 23-vertex triangle-free graph with `beta >= 20`, hence one of McKay's six `minbip23_20x.g6` extremals, all of which have at least 100 edges, contradicting `e <= 95`.

## Verification
- `lake exe cache get`
- `lake build FormalConjectures.ErdosProblems.«23»`


mo271 (approved):
Thanks, LGTM!

PR #4217, opened 2026-06-10
feat(ErdosProblems): 1057

### LINKED ISSUE #1100: Erdős Problem 1057
### What is the conjecture

https://www.erdosproblems.com/1057

  Let $C(x)$ count the number of Carmichael numbers in the interval $[1,x]$. Is it true that $C(x)=x^{1-o(1)}$?

Status:  open

### Choose either option

danielchin:
Closes #1100  

Formalizes Erdos Problem 456: https://www.erdosproblems.com/456
Centralized `IsCarmichael` in Carmichael.lean.


danielchin:
I wanted to move the Carmichael related theorems to the `Carmichael.lean` file, but it looks like `korselts_criterion` is one of the problems in `FC100SolvedSet1.lean` and having a benchmark problem in a mathlib file feels a bit awkward, so I kept it in `AgohGiuga.lean` instead.

Also to note, since I fixed `IsCarmichael` in #4199, this means that `FC100SolvedSet1.lean` had an incorrect formalization in the problem set. Not sure if this affects any training or benchmarking this set was being used for.

FormalConjectures/ErdosProblems/1057.lean
```diff
@@ -0,0 +1,112 @@
...
+open Nat Real Filter Set
+open scoped Topology Asymptotics
+
+namespace Erdos1057
+
+
+/--
+Let $C(x)$ count the number of Carmichael numbers in the interval $[1,x]$.
+-/
+noncomputable def C (x : ℝ) : ℝ :=
```
BoltonBailey: I would suggest naming this `carmichaelCounting`, by analogy with [`Nat.primeCounting`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/NumberTheory/PrimeCounting.html#Nat.primeCounting), and because it would make it a bit less ambiguous to understand when we see it used.

BoltonBailey (approved):
One comment, but otherwise, these formalizations look fine to me.

FormalConjectures/ErdosProblems/1057.lean
danielchin: Great point, renamed to `carmichaelCounting`.

FormalConjecturesForMathlib/NumberTheory/Carmichael.lean
```diff
@@ -0,0 +1,38 @@
...
+*References:*
+ - [Wikipedia](https://en.wikipedia.org/wiki/Carmichael_number)
+-/
+
+/--
+A Carmichael number is a composite number `n` such that for all `b ≥ 1`,
+we have `b^n ≡ b (mod n)`.
+-/
+def IsCarmichael (n : ℕ) : Prop :=
+  1 < n ∧ ¬ n.Prime ∧ ∀ b ≥ 1, n.Coprime b → n.FermatPsp b
```
mo271: are `1 < n ∧ ¬ n.Prime` redundant here, because it is already implied by `n.FermatPsp`?

FormalConjecturesForMathlib/NumberTheory/Carmichael.lean
```diff
@@ -0,0 +1,38 @@
...
+
+*References:*
+ - [Wikipedia](https://en.wikipedia.org/wiki/Carmichael_number)
+-/
+
+/--
+A Carmichael number is a composite number `n` such that for all `b ≥ 1`,
+we have `b^n ≡ b (mod n)`.
+-/
+def IsCarmichael (n : ℕ) : Prop :=
```
mo271, line 36: perhaps a test like `IsCarmichael 561` might be a good addition here?

mo271 (approved):
Nice, thanks!
Just two minor (optional) suggestions

mo271:
Many thanks also @BoltonBailey for the review!!

FormalConjecturesForMathlib/NumberTheory/Carmichael.lean
danielchin: Shoot, you're completely right here; it looks like I completely overlooked that. The changes I made in #4199 were probably not necessary then.

I've removed the redundancies and fixed the proofs in AgohGiuga.

FormalConjecturesForMathlib/NumberTheory/Carmichael.lean
danielchin: I added the test to AgohGihua since the test of `IsCarmichael 561` is dependent on Korselt's Criteron. There also doesn't seem to be a way for us to add category tags within the Mathlib functions. Let me know if there's a better way of handling this.

mo271 (approved):
Thanks again, LGTM

PR #4218, opened 2026-06-10
Add Erdős Problem 1212 (visible lattice points, composite-protected i…

AlperTheKing:
First formalization of [Erdős Problem 1212](https://www.erdosproblems.com/1212) (Er80, p.114): does the visible-lattice graph (gcd(x,y)=1, unit steps) contain an infinite path all of whose vertices have both coordinates > 1 and at least one coordinate composite? The page lists the problem as open with no formalization.

Contents:
- The problem statement in the catalog pattern (research open, answer(sorry)).
- A `@[category test]` sanity check for the `Valid` vertex predicate.
- Six proven API lemmas: machine-checked cores of verified partial results obtained on this problem (2026): per-vertex validity of composite-anchored L-path legs, a roughness criterion (least prime factor exceeding the leg length forces coprimality along the leg), and three isolation lemmas underlying a no-periodic-certificate theorem.
- A discrete intermediate value theorem `Int.exists_eq_of_natAbs_sub_le_one` added to `FormalConjecturesForMathlib` (generalising the unit-step walk helper, per review): an integer-valued sequence with consecutive differences at most 1 attains every value between two of its terms.

Verification: lake build is clean; the only warnings are the expected sorry on the open catalog statement; all proven lemmas use only propext / Classical.choice / Quot.sound (the ForMathlib lemma uses only propext / Quot.sound).

Mathematical context (not in this PR): these cores support (i) a sufficient reduction (an infinite sequence of composites with a_{n+2} - a_n < min of the least prime factors at both ends yields the desired path); (ii) a theorem that no fixed periodic certificate can witness the answer; (iii) a conditional YES under an every-interval max-gap hypothesis for rough composites (open; closest known are the almost-all theorems of Matomäki, JLMS 2022 and Matomäki–Teräväinen, TAMS 2023). The problem itself remains open.


FormalConjectures/ErdosProblems/1212.lean
```diff
@@ -0,0 +1,187 @@
...
+/--
+**Erdős Problem 1212** [Er80, p.114]: is there an infinite path through visible lattice
+points, going to infinity, all of whose vertices have both coordinates `> 1` and at least
+one coordinate composite?
+-/
+@[category research open, AMS 11]
+theorem erdos_1212 :
+    (∃ f : ℕ → ℕ × ℕ, Function.Injective f ∧ (∀ n, Adj (f n) (f (n + 1))) ∧
+      (∀ n, Valid (f n)) ∧
+      Tendsto (fun n => (f n).1 + (f n).2) atTop atTop) ↔ answer(sorry) := by
```
mo271:
```suggestion
    answer(sorry) ↔ ∃ f : ℕ → ℕ × ℕ, Function.Injective f ∧ (∀ n, Adj (f n) (f (n + 1))) ∧
      (∀ n, Valid (f n)) ∧
      Tendsto (fun n => (f n).1 + (f n).2) atTop atTop := by
```

We usually put the `answer(sorry)` first

FormalConjectures/ErdosProblems/1212.lean
```diff
@@ -0,0 +1,187 @@
...
+
+/-- Isolation lemma, vertical neighbours: both coordinates even. -/
+@[category API, AMS 11]
+theorem vertical_neighbor_both_even {x y : ℕ} (h2x : 2 ∣ x) (h2y : ¬ 2 ∣ y) :
+    (2 ∣ x ∧ 2 ∣ (y + 1)) ∧ (1 ≤ y → 2 ∣ x ∧ 2 ∣ (y - 1)) := by
+  refine ⟨⟨h2x, by omega⟩, fun hy1 => ⟨h2x, by omega⟩⟩
+
+/-- Winding step: a `±1`-step walk attains every value between its first and last entries. -/
+@[category API, AMS 11]
+theorem walk_intermediate_value :
```
mo271: seems like this would be better suited in some more general form in `ForMathlib`? isn't it just a version of the intermediate value theorem for a suitably choosen topology?

FormalConjectures/ErdosProblems/1212.lean
```diff
@@ -0,0 +1,187 @@
...
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 1212
+
+*Reference:* [erdosproblems.com/1212](https://www.erdosproblems.com/1212)
+[Er80] Erdős, P., _Some notes on problems and results in number theory_ (1980), p. 114.
```
mo271:
```suggestion
*References:* 
- [erdosproblems.com/1212](https://www.erdosproblems.com/1212)
- [Er80] Erdős, P., _Some notes on problems and results in number theory_ (1980), p. 114.
```

FormalConjectures/ErdosProblems/1212.lean
```diff
@@ -0,0 +1,187 @@
...
+-/
+
+open Filter
+
+namespace Erdos1212
+
+/-- A vertex of the strengthened problem: a visible lattice point with both coordinates
+exceeding 1 and at least one coordinate composite. -/
+def Valid (p : ℕ × ℕ) : Prop :=
+  1 < p.1 ∧ 1 < p.2 ∧ Nat.gcd p.1 p.2 = 1 ∧ (¬ p.1.Prime ∨ ¬ p.2.Prime)
```
mo271, line 38: perhaps a short  sanity check could be added here for an example of a valid pair?

FormalConjectures/ErdosProblems/1212.lean
```diff
@@ -0,0 +1,187 @@
...
+theorem erdos_1212 :
+    (∃ f : ℕ → ℕ × ℕ, Function.Injective f ∧ (∀ n, Adj (f n) (f (n + 1))) ∧
+      (∀ n, Valid (f n)) ∧
+      Tendsto (fun n => (f n).1 + (f n).2) atTop atTop) ↔ answer(sorry) := by
+  sorry
+
+/- ### Verified partial results (2026): machine-checked cores -/
+
+/-- A natural number is composite. -/
+def Composite (n : ℕ) : Prop := 2 ≤ n ∧ ¬ n.Prime
```
mo271: use mathlib's `Nat.Composite` instead of redefining.

mo271 (commented):
Thanks for the contribution!

A few initial remarks... Looks good in general

AlperTheKing:
Thanks for the review, @mo271! Addressed in 0274462:

- `answer(sorry)` is now on the left of the `↔` in `erdos_1212`.
- References reformatted as a bulleted list.
- Added a `@[category test]` sanity check `valid_four_three : Valid (4, 3)`.
- Replaced the local `Composite` definition with the library's `Nat.Composite`.
- `walk_intermediate_value`: agreed — it's a general (discrete) intermediate-value statement that does not belong in a problem file, so I've removed it from this PR (it was not used by the statement or the other cores). Happy to contribute a generalized version to `FormalConjecturesForMathlib` / Mathlib in a separate PR if useful.

`lake env lean` exits 0 (only the expected `sorry` on the open statement); the proven lemmas keep `[propext, Classical.choice, Quot.sound]`.

mo271:
> Thanks for the review, @mo271! Addressed in [0274462](https://github.com/google-deepmind/formal-conjectures/commit/02744628724e47a2b5ede02df29b314ab16a4a46):
> 
> * `answer(sorry)` is now on the left of the `↔` in `erdos_1212`.
> * References reformatted as a bulleted list.
> * Added a `@[category test]` sanity check `valid_four_three : Valid (4, 3)`.
> * Replaced the local `Composite` definition with the library's `Nat.Composite`.
> * `walk_intermediate_value`: agreed — it's a general (discrete) intermediate-value statement that does not belong in a problem file, so I've removed it from this PR (it was not used by the statement or the other cores). Happy to contribute a generalized version to `FormalConjecturesForMathlib` / Mathlib in a separate PR if useful.
> 
> `lake env lean` exits 0 (only the expected `sorry` on the open statement); the proven lemmas keep `[propext, Classical.choice, Quot.sound]`.

can you click "Resolve conversation" for those conversations that are resolved? That makes reviewing easier

AlperTheKing:
Follow-up on the `walk_intermediate_value` point: rather than dropping it, I've **generalised** it into `FormalConjecturesForMathlib` and kept it in this PR (commit a451c41).

New lemma `FormalConjecturesForMathlib/Data/Int/IntermediateValue.lean`:
```
theorem Int.exists_eq_of_natAbs_sub_le_one (f : ℕ → ℤ)
    (hf : ∀ n, (f (n + 1) - f n).natAbs ≤ 1) {a b : ℕ} (hab : a ≤ b) {t : ℤ}
    (hta : f a ≤ t) (htb : t ≤ f b) : ∃ c, a ≤ c ∧ c ≤ b ∧ f c = t
```
i.e. a discrete intermediate value theorem: an integer-valued sequence whose consecutive terms differ by at most 1 attains every value between two of its terms. Proof is a short `Nat.le_induction` (no topology needed). `lake build` is clean and the lemma's axioms are `[propext, Quot.sound]`.

FormalConjectures/ErdosProblems/1212.lean
mo271, line 1: docstrings in this file should use latex markdown over backticks

AlperTheKing: Done in f619945: converted the mathematical notation in the per-declaration docstrings from backticks to LaTeX `$...$`; the only remaining backtick is the Lean identifier `Valid`.

FormalConjectures/ErdosProblems/1212.lean
```diff
@@ -0,0 +1,159 @@
...
+
+/-- Two lattice points are adjacent iff they differ by exactly 1 in exactly one coordinate. -/
+def Adj (p q : ℕ × ℕ) : Prop :=
+  (p.1 = q.1 ∧ (p.2 = q.2 + 1 ∨ q.2 = p.2 + 1)) ∨
+  (p.2 = q.2 ∧ (p.1 = q.1 + 1 ∨ q.1 = p.1 + 1))
+
+/--
+**Erdős Problem 1212** [Er80, p.114]: is there an infinite path through visible lattice
+points, going to infinity, all of whose vertices have both coordinates $> 1$ and at least
+one coordinate composite?
```
mo271:
```suggestion
Let $G$ be the graph with vertex set those pairs $(x,y)\in \mathbb{N}^2$ with $\mathrm{gcd}(x,y)=1$, in which we join two vertices if the differ in only one coordinate, and there by $\pm 1$.

Is there a path going to infinity on $G$, say $P$, such that for all $(x,y)\in P$ both $\min(x,y)>1$ and at least one of $x$ or $y$ is composite?
```
Let's cite verbatim

FormalConjectures/ErdosProblems/1212.lean
```diff
@@ -0,0 +1,159 @@
...
+- [Er80] Erdős, P., _Some notes on problems and results in number theory_ (1980), p. 114.
+
+Let $G$ be the graph on the visible lattice points $\{(x,y) \in \mathbb{Z}_{>0}^2 :
+\gcd(x,y) = 1\}$, where two points are joined if they differ by exactly $1$ in exactly one
+coordinate. Erdős asked: is there a path going to infinity in $G$ all of whose vertices
+$(x,y)$ satisfy $\min(x,y) > 1$ and at least one of $x, y$ composite?
+
+The weaker version (only $\min(x,y) > 1$) was solved by C. Stewart via the prime-pair path
+$(p_k, p_{k+1}) \to (p_{k+1}, p_{k+2})$, as recounted in [Er80]; the compositeness condition
+forbids those anchors and the question is open.
```
mo271:
```suggestion
```
Let's not duplicate the problem statement here

mo271 (commented):
Thanks, two more suggestions for the doc strings

FormalConjectures/ErdosProblems/1212.lean
AlperTheKing: Done in 6bef5cf: replaced the paraphrase on `erdos_1212` with the verbatim erdosproblems.com/1212 statement; kept the Stewart historical note below it.

FormalConjectures/ErdosProblems/1212.lean
AlperTheKing: Done in 6bef5cf: removed the duplicated problem statement from the module docstring (it now keeps only the references and the note on the file's machine-checked lemmas).

mo271 (approved):
Thanks, LGTM

PR #4221, opened 2026-06-10
Add Erdős Problem 1127 (ℝⁿ as countably many distinct-distance sets)

henrykmichalewski:
Formalizes [Erdős Problem 1127](https://www.erdosproblems.com/1127): can $\mathbb{R}^n$ be decomposed into countably many sets such that within each set all pairwise distances are distinct?

The answer is **independent of ZFC**:
- Under CH it is true: Kunen [Ku87] proved it for all $n$ (Davies [Da72] for $n = 2$; the $n = 1$ case follows from the Erdős–Kakutani theorem [ErKa43] that CH is equivalent to $\mathbb{R}$ being a countable union of $\mathbb{Q}$-linearly independent sets).
- Under $\neg$CH it fails: Erdős and Hajnal showed that any decomposition of $\mathbb{R}$ into finitely many sets contains four points in one part determining only four distances.

Contents (namespace `Erdos1127`, working in `EuclideanSpace ℝ (Fin n)`):
- `PairwiseDistinctDistances` and `Decomposable` definitions.
- Headline `erdos_1127` as `answer(sorry) ↔ ∀ n, Decomposable n` (answer independent of ZFC).
- Variants: `kunen` (CH → all $n$), `davies` (CH → $n = 2$), `erdos_kakutani` (CH ↔ countable union of $\mathbb{Q}$-independent sets, via `LinearIndepOn`), `erdos_hajnal` ($\neg$CH → no finite decomposition of $\mathbb{R}$).
- Test lemma `decomposable_zero` (`Decomposable 0`), proved without sorry.

Sorry count: 5 (the headline plus the four research variants; all are statements of known results or the independent headline question).

`lake env lean FormalConjectures/ErdosProblems/1127.lean` passes with only `declaration uses 'sorry'` warnings.


FormalConjectures/ErdosProblems/1127.lean
```diff
@@ -0,0 +1,141 @@
...
+
+/--
+A set $A$ in a metric space has **pairwise distinct distances** if any two pairs of
+distinct points of $A$ realising the same distance are equal as unordered pairs:
+whenever $x, y, z, w \in A$ with $x \neq y$, $z \neq w$ and
+$d(x, y) = d(z, w)$, we have $\{x, y\} = \{z, w\}$.
+-/
+def PairwiseDistinctDistances {X : Type*} [MetricSpace X] (A : Set X) : Prop :=
+  ∀ x ∈ A, ∀ y ∈ A, ∀ z ∈ A, ∀ w ∈ A,
+    x ≠ y → z ≠ w → dist x y = dist z w → (x = z ∧ y = w) ∨ (x = w ∧ y = z)
```
mo271: coordinate with  #4258: this has the same(?) definition and it should be in FormalConjecturesForMathlib.

mo271 (commented):
Thanks! 
Mathematically this looks quite good!

Other than the comment about `PairwiseDistinctDistances`, two more nits: 
 - LaTeX over backticks in docstrings.
 - Let's not repeat everything int he module doc string see #4296 and oinly *one* reference(s) section

FormalConjectures/ErdosProblems/1127.lean
henrykmichalewski: Done in 6a135de — moved `PairwiseDistinctDistances` into `FormalConjecturesForMathlib/Geometry/Metric.lean` (alongside `unitDistNum`) as a general metric-space predicate on `Set X`.

On coordinating with #4258: that PR's `HasDistinctDistances {d} (S : Finset (EuclideanSpace ℝ (Fin d)))` is the finite/Euclidean special case of this same notion. The version I've added is deliberately general (any `MetricSpace`, on `Set X`), so #4258 could reuse it — e.g. `HasDistinctDistances S := PairwiseDistinctDistances (S : Set _)` — rather than keeping a separate definition. Happy to adjust the name/shape if you'd prefer they be unified differently; whichever of the two PRs lands first can host the shared def.

henrykmichalewski:
Addressed the two general nits in 6a135de as well:
- **Single reference section / no repeated module docstring:** slimmed the `/-! ... -/` header to just the title + one `*References:*` block (per `ErdosProblems/README.md` and #4296); the statement now lives only in the `erdos_1127` theorem docstring.
- **LaTeX over backticks:** the docstrings already use LaTeX for all mathematical content; the one remaining backtick (`Decomposable n`) is a reference to the Lean definition, not math notation.

henrykmichalewski:
Rebased onto current `main` (v4.33.1) after the `MetricSpace` generalisation of `ForMathlib/Geometry/Metric.lean` (#5063): `PairwiseDistinctDistances` now sits alongside the upstream definitions. Builds clean locally.

PR #4225, opened 2026-06-10
ErdosProblems/349: add proven partial result `alpha_gt_two_not_isGoodPair`

cepadugato:
## Summary

Adds a **fully-proven partial result** to the (open) Erdős Problem 349 — the additive completeness of `⌊t·αⁿ⌋`:

> For every `t > 0` and every `α > 2`, the pair `(t, α)` is **not** good, i.e. the sequence `⌊t·αⁿ⌋` is not additively complete.

```lean
theorem alpha_gt_two_not_isGoodPair (t α : ℝ) (ht : 0 < t) (hα : 2 < α) : ¬ IsGoodPair t α
```

This is the "not complete" side of the problem, complementary to the existing conjectural `complete_for_alpha_in_Ioo_one_to_goldenRatio`. The proof is the classical gap argument: eventually each term exceeds `1 +` the sum of all previous terms (geometric-sum bound + `tendsto` to pick the index), leaving integers that no subset sum can reach.

## Honesty notes

- **This does not resolve the open problem.** `erdos_349`'s `answer(sorry)` and all existing theorems are untouched — the diff is a **pure addition** (+183 lines, 0 deletions).
- It is a partial result constraining the answer set (cf. W. van Doorn, *Completeness of exponentially increasing sequences*, arXiv:2602.23394, where this regime is part of Proposition 1).

## Verification

- Builds with the repo toolchain (Lean v4.27.0): `lake env lean FormalConjectures/ErdosProblems/349.lean` → exit 0; the only `sorry` warnings are the pre-existing open statements.
- `#print axioms Erdos349.alpha_gt_two_not_isGoodPair` → `[propext, Classical.choice, Quot.sound]` (no `sorryAx`, no `native_decide`).


cepadugato:
@googlebot I signed it!

mo271 (commented):
Thanks, looks like a good statement to have (even though it doesn't seem to be mentioned on erdosproblems.com). However its proof is too long: let's use the `formal_proof using formal_conjectures at [commit]` mechanism to refer to it instead of adding the completely written out proof. 

Also: the docstrings should use latex markdown.

mo271 (approved):
Thanks, LGTM

cepadugato:
Thanks for the review! Addressed both points:

- **Proof length** — the written-out proof is no longer inline. The statement now carries `@[… formal_proof using formal_conjectures at "…"]` pointing to a fixed commit on my fork (branch `erdos-349-alpha-gt-two-proof`) where the full proof lives, and the body is `by sorry` here, per the guideline.
- **Docstrings** — rewritten in LaTeX markdown (`$\alpha > 2$`, `$\lfloor t\alpha^n\rfloor$`, …).

I applied the same treatment to the stacked PRs #4233, #4234, #4235 (statements + `formal_proof` links; short helper lemmas keep their inline proofs). The file builds under the repo toolchain with only the expected `sorry` warnings on the statements.

PR #4230, opened 2026-06-11
Add Komlós conjecture (vector balancing / discrepancy)

henrykmichalewski:
This PR adds the **Komlós conjecture** from discrepancy theory, in `FormalConjectures/Wikipedia/KomlosConjecture.lean`.

**Statement.** There is a universal constant $K$ such that for all $n, m$ and all vectors $v_1, \dots, v_n \in \mathbb{R}^m$ with $\|v_i\|_2 \le 1$, there exist signs $\varepsilon_i \in \{-1, +1\}$ with $\left\|\sum_i \varepsilon_i v_i\right\|_\infty \le K$.

Vectors are encoded elementarily as `v : Fin n → Fin m → ℝ` with the hypothesis `∀ i, ∑ j, (v i j) ^ 2 ≤ 1` (the ℓ² unit ball) and the conclusion `∀ j, |∑ i, ε i * v i j| ≤ K` (the ℓ∞ bound), avoiding `PiLp` friction.

**Status.** The conjecture is open (`@[category research open, AMS 5]`). The best known bound is Banaszczyk's $O(\sqrt{\log n})$ (Random Structures & Algorithms 12 (1998), 351–360), included as the variant `komlos_conjecture.variants.banaszczyk` (`@[category research solved]`, statement left as `sorry`). The Beck–Fiala theorem is the special case for sparse set systems.

**Contents:**
- `komlos_conjecture` — the headline open conjecture (`sorry`)
- `komlos_conjecture.variants.banaszczyk` — Banaszczyk's $C\sqrt{\log(n+2)}$ bound, the best known (`sorry`; the $+2$ shift is a harmless normalization keeping the logarithm positive for $n \in \{0,1\}$)
- `komlos_conjecture.variants.zero_vectors` — sanity check for $n = 0$, proved without `sorry`

Sorry count: 2 (both research statements). `lake build FormalConjectures.Wikipedia.KomlosConjecture` passes with the project linters; the only warnings are `declaration uses 'sorry'`.

*Reference:* [Wikipedia: Discrepancy theory § Major open problems](https://en.wikipedia.org/wiki/Discrepancy_theory#Major_open_problems) (the Komlós conjecture has no dedicated article; it is listed there).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4243, opened 2026-06-11
Add Spencer's six-standard-deviations theorem (set-system discrepancy)

henrykmichalewski:
Adds `FormalConjectures/Wikipedia/SixStandardDeviations.lean`, a statement-only formalization of Spencer's celebrated 1985 theorem in discrepancy theory: for every $n$ and every family of $n$ subsets $S_1,\dots,S_n$ of $\{1,\dots,n\}$, there is a colouring $\chi : \{1,\dots,n\} \to \{-1,+1\}$ with $\left|\sum_{j \in S_i} \chi(j)\right| \le 6\sqrt{n}$ for every $i$.

Contents:
- `six_standard_deviations` (`research solved`): the headline theorem. A uniformly random colouring only gives $O(\sqrt{n \log n})$; Spencer's entropy-method proof removes the log factor, and $\Theta(\sqrt{n})$ is tight up to the constant. The proof is non-constructive; algorithmic versions are due to Bansal (2010) and Lovett–Meka (2012, SICOMP 2015), mentioned in the module docstring.
- `six_standard_deviations.variants.random_bound` (`research solved`): the easy Chernoff/union-bound benchmark $\exists K,\ |\chi(S_i)| \le K\sqrt{n\log(n+2)}$, included to contextualize what "six standard deviations" improves on.
- `six_standard_deviations.variants.zero_sets` (`test`): the $n = 0$ case, closed without `sorry`.

There is no dedicated Wikipedia article ("Six standard deviations" redirects to Six Sigma), so the *Reference:* line points at the section [Discrepancy of hypergraphs § General hypergraphs](https://en.wikipedia.org/wiki/Discrepancy_of_hypergraphs#General_hypergraphs), which states the theorem, following the section-anchored precedent of `BoundedBurnsideProblem.lean`.

Related: this complements the Komlós conjecture (#4230) — applied to the scaled incidence vectors of the sets, the Komlós conjecture would give an $O(\sqrt{n})$ bound for any number of sets, strengthening Spencer's theorem up to the constant. No statements are duplicated between the two files, and a content search found no existing formalization of this theorem in the repository (`ErdosProblems/67.lean` is the unrelated Erdős discrepancy problem).

Both `lake env lean` and `lake build FormalConjectures.Wikipedia.SixStandardDeviations` pass with only the expected `sorry` warnings (2).


mo271 (approved):
Thanks looks great, thanks!

PR #4245, opened 2026-06-11
feat(ErdosProblems/1084): prove upper_d1

Sanexxxx777:
`f 1 n = n - 1` is formally proved: among `n` points on the line with pairwise distance ≥ 1, the maximum number of pairs at distance exactly 1 is `n - 1`.

Following the repository convention for longer proofs (and @mo271's suggestion), the proof is **linked via the `formal_proof` mechanism** rather than inlined — this PR only adds the `formal_proof using formal_conjectures at …` attribute to the existing statement. The full Lean development (statement + 12 supporting lemmas) lives on the linked fork branch and builds cleanly.

**Proof outline:** upper bound via an injection of unit-distance pairs into `s.erase m` (`m` = coordinate-minimal point); lower bound via the witness `{0,1,…,n-1}`; the `iSup` over separated finsets is handled with `ciSup_le'` / `le_ciSup_of_le` and an explicit `BddAbove` bound.

Verification of the linked proof: `lake build` clean, `#print axioms` shows only `[propext, Classical.choice, Quot.sound]`.

### DIFF
--- a/FormalConjectures/ErdosProblems/1084.lean
+++ b/FormalConjectures/ErdosProblems/1084.lean
@@ -44,7 +44,7 @@ noncomputable def f (d n : ℕ) : ℕ :=
 -- TODO: Add erdos_1084.
 
 /-- It is easy to check that $f_1(n) = n - 1$. -/
-@[category research solved, AMS 52]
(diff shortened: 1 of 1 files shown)


mo271 (commented):
Thanks!
Proof looks good, but a little long, perhaps you could instead of adding if use the `formal_proof using formal_conjectures at` mechanism?!

Sanexxxx777:
Signed the CLA. Also restructured per your suggestion, @mo271 — the proof is now linked via the `formal_proof using formal_conjectures at …` mechanism instead of being inlined. Thanks for the review!

mo271 (approved):
Thanks, LGTM!

PR #4248, opened 2026-06-12
feat(Wikipedia): add Erdős–Hanani conjecture on asymptotic packings and coverings

henrykmichalewski:
Adds `FormalConjectures/Wikipedia/ErdosHananiConjecture.lean`, formalising the Erdős–Hanani conjecture (1963), proved by Rödl (1985) with the semi-random "nibble" method.

For fixed $2 \leq r < k$, let $m(n,k,r)$ be the maximum size of a family of $k$-element subsets of an $n$-element set in which every $r$-element subset lies in at most one member (a packing / partial Steiner system), and $M(n,k,r)$ the minimum size of a family containing every $r$-element subset in at least one member (a covering). Erdős and Hanani conjectured, and Rödl proved, that
$$\lim_{n\to\infty} \frac{m(n,k,r)}{\binom{n}{r}/\binom{k}{r}} = \lim_{n\to\infty} \frac{M(n,k,r)}{\binom{n}{r}/\binom{k}{r}} = 1.$$

Contents:
- `IsPacking` / `IsCovering` predicates and the extremal quantities `maxPacking` / `minCovering`;
- `erdos_hanani_packing` and `erdos_hanani_covering` (`research solved`, stated with `sorry`);
- three closed sanity tests (`category test`): the empty family is a packing, every packing has at most $2^n$ members (so `sSup` is over a nonempty bounded set), and for $r \leq k \leq n$ the family of all $k$-subsets is a covering.

Note: this statement is not in the erdosproblems.com database (problem [722](https://www.erdosproblems.com/722) is the related but distinct exact Steiner-system existence question, and the existing `Wikipedia/SteinerSystem.lean` concerns explicit large Steiner-system witnesses), so the file is placed under `Wikipedia/`, following the [Rödl nibble](https://en.wikipedia.org/wiki/R%C3%B6dl_nibble) article.

References: P. Erdős and H. Hanani, *On a limit theorem in combinatorial analysis*, Publ. Math. Debrecen **10** (1963), 10–13; V. Rödl, *On a packing and covering problem*, European J. Combin. **6** (1985), 69–78.

`lake build FormalConjectures.Wikipedia.ErdosHananiConjecture` passes with only the two intended `sorry`s.


mo271 (commented):
Thanks, looks great!
Everything seems mathematically correct and follows the style of the repo. 
The only potential improvement I see is to unify some definitions: 
FormalConjectures/Wikipedia/SteinerSystem.lean defines:

```lean
structure SteinerSystem (t k n : ℕ) where
  blocks : Finset (Finset (Fin n))
  block_card : ∀ B ∈ blocks, B.card = k
  cover_unique : ∀ R : Finset (Fin n), R.card = t → (blocks.filter (R ⊆ ·)).card = 1
```
Notice how cover_unique is ...card = 1. In ErdosHananiConjecture.lean, a packing requires ...card ≤ 1, and a covering requires ...card ≥ 1 (or ∃).

These are exactly the three sides of the same coin: combinatorial designs!

Packing Design (Partial Steiner System): ≤ 1
Exact Design (Steiner System): = 1
Covering Design: ≥ 1
Combining them into a unified file like FormalConjecturesForMathlib/Combinatorics/SetFamily/Design.lean is definitely the right architectural move. We can define generalized versions of IsPacking, IsCovering, and IsSteinerSystem over any [Fintype α] [DecidableEq α].

(One interesting note from the comments in SteinerSystem.lean is that they deliberately used (blocks.filter (R ⊆ ·)).card = 1 instead of ∃! to keep it computable for native_decide on the Fano plane test cases. When generalizing to Mathlib, preserving that computability by keeping Fintype and Decidable instances is a great pattern.)

I would either be great to go ahead and create FormalConjecturesForMathlib/Combinatorics/SetFamily/Design.lean here and unify these three concepts into it, and then update both Wikipedia benchmark files to use the new shared Mathlib definitions or at least leave a TODO and do it in a follow up...

PR #4255, opened 2026-06-12
feat(ErdosProblems): add 619

### LINKED ISSUE #4254: Erdős Problem 619

https://www.erdosproblems.com/619

For a triangle-free graph $G$ let $h_r(G)$ be the smallest number of edges that need to be added to $G$ so that it has diameter $r$ (while preserving the property of being triangle-free). Is it true that there exists a constant $c>0$ such that if $G$ is a connected graph on $n$ vertices then $h_4(G)<(1-c)n$?

Status: open

* ams-05

### Choose either option
- [x] I plan on adding this Erdős problem to the repository
- [ ] This issue is up for grabs: I would like to see this Erdős problem added by somebody else

### Additional context

A problem of Erdős, Gyárfás, and Ruszinkó [EGR98], who proved $h_3(G)\leq n$ and $h_5(G)\leq\frac{n-1}{2}$, and that there exist connected graphs on $n$ vertices with $h_3(G)\geq n-c$ for some constant $c>0$; I plan to include these as `research solved` variants. The formalisation uses `SimpleGraph.ediam` (so disconnected supergraphs are excluded by infinite diameter) and `Nat.sInf` for the minimum.

Note: I have a claimed disproof of this conjecture (in Lean, outside this repository). This issue and the upcoming PR concern only the statement, as `research open`; any status change would follow later via the `formal_proof` mechanism once the claim has been reviewed.

nick-kuhn:
## Summary

This PR adds Erdős Problem 619 ([erdosproblems.com/619](https://www.erdosproblems.com/619)): for a triangle-free graph $G$, let $h_r(G)$ be the smallest number of edges that need to be added to $G$ so that it has diameter at most $r$ while remaining triangle-free; is there a constant $c>0$ such that every connected triangle-free graph on $n$ vertices satisfies $h_4(G)<(1-c)n$?

- defines `minNewEdges r G` (the quantity $h_r(G)$ of Erdős–Gyárfás–Ruszinkó [EGR98]) as a `Nat.sInf` over triangle-free supergraphs on the same vertex set
- states the main question as `answer(sorry) ↔ ∃ c > 0, ...`, tagged `research open, AMS 5`
- records the partial results of [EGR98] as `research solved` variants: existence of a diameter-$\leq 3$ triangle-free extension, $h_3(G) \leq n$, $h_5(G) \leq \frac{n-1}{2}$, and the lower bound $h_3(G) \geq n - c$
- includes two proved `test` lemmas sanity-checking the definition

Closes #4254.

## Formalisation choices

- The diameter condition uses `SimpleGraph.ediam ≤ r` (valued in `ℕ∞`) rather than `diam`, so that disconnected supergraphs are excluded by their infinite diameter instead of slipping through `diam`'s junk value `0`.
- `Nat.sInf` returns `0` on the empty set; the `add_edges_diam_three` variant records the [EGR98] result that the defining set is nonempty for every connected triangle-free graph when `3 ≤ r`, so the junk case does not arise in the main statement.

## Scope

This is a statement addition. All `research` theorem bodies are `sorry`; the two `test` lemmas are proved.

I checked upstream `main` before opening: there is no existing `FormalConjectures/ErdosProblems/619.lean` and no `erdos_619` declaration.

## AI disclosure

The Lean file was drafted by Claude Fable (Claude Code), working from the problem statement and references on erdosproblems.com; I reviewed the statement and variants.

## Validation

- `lake build` (full project)
- references checked against the erdosproblems.com bibliography ([EGR98], [Er99])

### DIFF
--- a/FormalConjectures/ErdosProblems/619.lean
+++ b/FormalConjectures/ErdosProblems/619.lean
@@ -1,0 +1,132 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 619
+
+*References:*
+- [erdosproblems.com/619](https://www.erdosproblems.com/619)
+- [EGR98] Erdős, Paul and Gyárfás, András and Ruszinkó, Miklós, *How to decrease the
+  diameter of triangle-free graphs*. Combinatorica **18** (1998), 493-501.
+- [Er99] Erdős, Paul, *A selection of problems and results in combinatorics*.
+  Combin. Probab. Comput. **8** (1999), 1-6.
+-/
+
+open SimpleGraph
+
+namespace Erdos619
+
+/--
+For a graph $G$, `minNewEdges r G` is the smallest number of edges that need to be added
+to $G$ so that it has diameter at most $r$, while preserving the property of being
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

mo271:
The source says "has diameter $r$" (exact) but the code uses `ediam ≤ r` (at most). I think this is the intented interpretation: [EGR98] itself uses "at most", I guess the docstring for minNewEdges (line 36) could mention this, but on the other hand it seems quite clear already, so lgtm...

PR #4263, opened 2026-06-14
fix(ErdosProblems/619): mark as solved

### LINKED ISSUE #4262: Erdős Problem 619: mark as solved (negative resolution, Lean 4 proof)

https://www.erdosproblems.com/619

For a triangle-free graph $G$ let $h_r(G)$ be the smallest number of edges that need to be added to $G$ so that it has diameter $r$ (while preserving the property of being triangle-free). Is it true that there exists a constant $c>0$ such that if $G$ is a connected graph on $n$ vertices then $h_4(G)<(1-c)n$?

### Proposed status change

`research open` → `research solved` (negative resolution).

The answer is **no**: for every $\eta>0$ there exist connected triangle-free graphs on $n$ vertices with $h_4(G)\geq(1-\eta)n$, so no constant $c>0$ works. The statement was requested in #833 and added to the repository in #4255; this issue proposes flipping its status to solved.

### Evidence

A Lean 4 proof of the negation of the formalised statement, kernel-checked and using only the axioms `propext`, `Classical.choice`, `Quot.sound`:

- Proof: https://github.com/nick-kuhn/erdos-619/blob/7f65718b8c1019ecc24e6c9a6b04ec4c66a4e26f/Solution.lean#L5869
- The proof is stated against the `Erdos619.minNewEdges` definition and bound from #4255; a companion branch takes a Lake dependency on formal-conjectures at the merge commit `1a9fbee` and proves the same negation against the imported symbol, with a compile-time guard that the two statements are definitionally identical.

### Caveat (please scrutinise)

This is an AI-generated resolution of a previously-open Erdős problem, which has not been found in the prior literature. The faithfulness of the formal statement to the informal problem, and the correctness of the `comparator` setup that verifies the proof, both deserve additional scrutiny.

### Plan

- [x] I plan on opening the PR for this status change.

nick-kuhn:
## Summary

Marks Erdős Problem 619 as solved with a **negative** resolution, and links a kernel-checked Lean 4 proof.

Problem 619 asks whether there is a constant $c>0$ such that every connected triangle-free graph on $n$ vertices satisfies $h_4(G) < (1-c)n$, where $h_4(G)$ is the fewest edges that must be added to reach diameter $\leq 4$ while staying triangle-free. The answer is **no**: for every $\eta > 0$ there exist connected triangle-free graphs on $n$ vertices with $h_4(G) \geq (1-\eta)n$.

## Changes

- `FormalConjectures/ErdosProblems/619.lean`: `erdos_619` goes `@[category research open]` → `@[category research solved]`, `answer(sorry)` → `answer(False)`, and gains `@[formal_proof using lean4 at "<permalink>"]`. The `minNewEdges` definition, the right-hand side, and all `research solved`/`test` variants are unchanged. Proof body stays `sorry` per the proof-length policy in CONTRIBUTING.md.

## The proof

External Lean 4 repository, linked via `formal_proof`:
https://github.com/nick-kuhn/erdos-619/blob/7f65718b8c1019ecc24e6c9a6b04ec4c66a4e26f/Solution.lean#L5869

- ~5900 lines, no `sorry`; depends only on `propext`, `Classical.choice`, `Quot.sound` (audited).
- Kernel-checked by `comparator` (lean4export → Lean kernel) under a landrun sandbox, run reproducibly in CI.
- Proves the negation of **this repository's exact statement**: a companion branch adds a Lake dependency on formal-conjectures at the merge commit `1a9fbee`, imports `FormalConjectures.ErdosProblems.«619»`, and proves the refutation against the imported `Erdos619.minNewEdges`, with a compile-time guard (`guard_fc_rhs`) that fails unless the refuted statement is definitionally identical to the upstream one. The `main` branch linked above is the mathlib-only version (matching the convention of the other proof repos linked from this project).
- Mathematical route: a two-level counterexample family — a triangle-free core graph of small independence number with many pendant leaves per core vertex; a spanning-forest count lower-bounds the edges needed for diameter $\leq 4$, while triangle-freeness forces each new vertex's core-neighbourhood to be independent, capping "hub" capacity, so nearly every pendant needs its own edge.

## AI disclosure

The original proof was generated by Claude Fable 5; the formalization was sketched by Fable and implemented by GPT 5.5 with Codex. The result is verified by the Lean compiler and the comparator kernel check.

## Please scrutinise

This is an AI-generated resolution of a previously-open Erdős problem, which has not been found in the prior literature. The faithfulness of the formal statement to the informal problem, and the correctness of the `comparator` setup that verifies the proof, both deserve additional scrutiny.

Closes #4262.


mo271 (approved):
congrats!
Looks good to me!

Perhaps this could be updated to a `formal_proof using formal_conjectures` instead of `formal_proof using lean4` version without much effort? Seems like this statements are identical? Perhaps in a follow up

PR #4267, opened 2026-06-14
fix(ErdosProblems/619): link proof via formal_conjectures

nick-kuhn:
## Summary

Follow-up to #4263 (per @mo271's suggestion there): switch Erdős Problem 619's `formal_proof` from `lean4` (external repo) to `formal_conjectures`, with the proof carried in this PR's own history.

## How (two-commit scheme)

- **Commit 1** carries the full ~5900-line Lean proof of `erdos_619` inline (in an `Erdos619.Disproof` section). It builds green, and `#print axioms Erdos619.erdos_619` = `[propext, Classical.choice, Quot.sound]`.
- **Commit 2** (PR head) reverts `619.lean` to the `sorry` stub (per the >25–50-line proof-hosting policy in CONTRIBUTING.md) and points `formal_proof using formal_conjectures` at commit 1's blob.

**Net change to the live tree vs `main`: the `formal_proof` attribute line only.** The proof persists in commit 1's history (reachable from this PR's head ref), so the link resolves against `google-deepmind/formal-conjectures` — the same mechanism as existing `formal_conjectures` links (e.g. 1082). This survives squash-merge.

`minNewEdges`, the `erdos_619` statement, and all variants/tests are byte-identical to `main`.

Marked draft until the commit-1 link is confirmed live.


mo271 (approved):
Thanks, LGTM! 
It is also fine to keep *both* links

mo271 (approved):

PR #4268, opened 2026-06-14
feat(Paper): three-sunflower-free set systems

SproutSeeds:
## Summary
- add a paper-backed Formal Conjectures entry for the published v2 preprint "Three-sunflower-free set systems with bounded pairwise intersections"
- define M3/I3-style bounded-intersection three-sunflower extremal quantities and the restricted-threshold interface
- promote the reusable `IsUniform` predicate to `FormalConjecturesForMathlib`, preserving correct zero-cardinality semantics and adding the positive-cardinality equivalence
- clarify the paper's direct relationship to Erdős Problem 20 and its broader relationship to the different non-uniform model in Erdős Problem 857
- record the v2 solved statements, the Mantel-tight structural reduction, and the open t >= 3 / t = 2 constant questions, with Lean v2 formal-proof links where available

## Verification
- `lake --wfail build FormalConjecturesForMathlib.Combinatorics.SetFamily.Sunflower FormalConjectures.Paper.ThreeSunflowerFreeSetSystems` (8,055 jobs)
- `lake --wfail build` (9,152 jobs)
- GitHub Actions `Build project` passed at head `5f4f8b17`

### DIFF
--- a/FormalConjectures/Paper/ThreeSunflowerFreeSetSystems.lean
+++ b/FormalConjectures/Paper/ThreeSunflowerFreeSetSystems.lean
@@ -1,0 +1,230 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+import FormalConjecturesForMathlib.Combinatorics.SetFamily.Sunflower
+
+/-!
+# Three-sunflower-free set systems with bounded pairwise intersections
+
+*References:*
+* [Three-sunflower-free set systems with bounded pairwise intersections]
+  (https://doi.org/10.5281/zenodo.20693260), by *Cody Mitchell* (2026).
+* [sunflower-lean: paper-v2](https://doi.org/10.5281/zenodo.20693191),
+  companion Lean 4 formalization, release `paper-v2`.
+
+For `ℓ > t ≥ 1`, the paper studies the maximum size `M₃(ℓ,t)` of a
+family of distinct `ℓ`-sets with pairwise intersections of size at most `t`
+and no three-sunflower. The empty-core case is included, so three pairwise
+disjoint sets form a three-sunflower. The variant `I₃(ℓ,t)` imposes the
+additional condition that the family is intersecting.
+-/
+
+open Filter
+
+namespace ThreeSunflowerFreeSetSystems
+
+variable {α : Type}
+
+/-- A family is `ℓ`-uniform if every member is finite of cardinality `ℓ`. -/
+def IsUniform (ℓ : ℕ) (F : Set (Set α)) : Prop :=
+  ∀ A ∈ F, A.Finite ∧ A.ncard = ℓ
+
+/-- Every two distinct members of `F` have intersection size at most `t`. -/
+def HasPairwiseIntersectionsAtMost (t : ℕ) (F : Set (Set α)) : Prop :=
+  ∀ A ∈ F, ∀ B ∈ F, A ≠ B → (A ∩ B).ncard ≤ t
+
+/-- The family `F` contains no three-member sunflower. -/
+def ThreeSunflowerFree (F : Set (Set α)) : Prop :=
+  ¬ ∃ S : Set (Set α), S ⊆ F ∧ S.ncard = 3 ∧ IsSunflower S
+
+/--
+An admissible family for `M₃(ℓ,t)`: a distinct family of `ℓ`-sets, all
+pairwise intersections have size at most `t`, and no three members form a
+sunflower.
+-/
+def M3Admissible (ℓ t : ℕ) (F : Set (Set α)) : Prop :=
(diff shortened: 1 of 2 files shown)


mo271:
please mark as "Ready for review" when it is ready for review

SproutSeeds:
-awaiting-author

FormalConjectures/Paper/ThreeSunflowerFreeSetSystems.lean
```diff
@@ -0,0 +1,230 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+import FormalConjecturesForMathlib.Combinatorics.SetFamily.Sunflower
```
mo271:
```suggestion
```
Instead of `import`ing it here do it in `FormalConjecturesForMathlib.lean`

FormalConjectures/Paper/ThreeSunflowerFreeSetSystems.lean
```diff
@@ -0,0 +1,230 @@
...
+and at least `N` members contains a three-sunflower.
+-/
+noncomputable def restrictedThreshold (n t : ℕ) : ℕ :=
+  sInf {N : ℕ | ∀ {α : Type}, ∀ F : Set (Set α),
+    IsUniform n F → HasPairwiseIntersectionsAtMost t F → N ≤ F.ncard →
+      ∃ S ⊆ F, S.ncard = 3 ∧ IsSunflower S}
+
+/-- A natural number is a prime power. -/
+def IsPrimePower (q : ℕ) : Prop :=
+  ∃ p a : ℕ, p.Prime ∧ 0 < a ∧ q = p ^ a
```
mo271: Let's not duplicte definitions that are in mathlib: 
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/IsPrimePow.html#IsPrimePow

mo271 (commented):
Thanks! a few remarks:
 - Most of the docstrings in this file use backticks instead of standard LaTeX $ ... $.let's use Latex!
 - might it be possible to also link a few related references?

Generally this looks good mathematically, would potentially be good to relate it to other published questions/theorems

FormalConjectures/Paper/ThreeSunflowerFreeSetSystems.lean
SproutSeeds: Removed the file-local import. Current `main` already exports the sunflower module through `FormalConjecturesForMathlib.lean`, and the paper module now uses the current standard `FormalConjecturesUtil` import.

FormalConjectures/Paper/ThreeSunflowerFreeSetSystems.lean
SproutSeeds: Removed the duplicate predicate and switched the theorem hypothesis to Mathlib’s `IsPrimePow`.

SproutSeeds:
Addressed the June 21 review and rebased onto current `main`.

- removed the redundant file-local sunflower import and migrated to the current standard `FormalConjecturesUtil` import;
- reused Mathlib’s `IsPrimePow`;
- converted mathematical docstrings to LaTeX and added the related references;
- aligned the formal-proof annotations with the companion’s documented conditionality for the orthogoval-existence result.

Exact results: head `09c7f444`; one Lean file changed; the targeted module and full 8,893-job repository pass `lake --wfail build`; `git diff --check` passes. Both inline threads have direct replies and are resolved.

-awaiting-author

FormalConjectures/Paper/ThreeSunflowerFreeSetSystems.lean
```diff
@@ -0,0 +1,254 @@
...
+
+open Filter
+
+namespace ThreeSunflowerFreeSetSystems
+
+variable {α : Type}
+
+/-- A family is $\ell$-uniform if every member is finite of cardinality $\ell$. -/
+def IsUniform (ℓ : ℕ) (F : Set (Set α)) : Prop :=
+  ∀ A ∈ F, A.Finite ∧ A.ncard = ℓ
```
mo271: This seems like something very useful in general? Is it not already in (FormalConjecturesFor)Mathlib?
if not, let's put it there. 

I'm not entirely convinced that  requiring `A.Finite` is the right way to go here, I suppose if `0<ℓ` there would be no problem here?

mo271 (commented):
Thanks, looks good.

One more question: Is this related to anyof the sunflower conjectures that we already have in the repo (as Erdosproblems for instance). Perhaps you can quickly search and check...

FormalConjectures/Paper/ThreeSunflowerFreeSetSystems.lean
SproutSeeds: Good point. I could not find an existing predicate for uniform families of `Set α`. Mathlib has `Set.Sized` for families of `Finset α`, but it does not cover this representation.

I moved this definition to `FormalConjecturesForMathlib/Combinatorics/SetFamily/Sunflower.lean`. I kept the finiteness condition in the general definition because, at `ℓ = 0`, `ncard = 0` also holds for infinite sets. For `0 < ℓ`, you are right that finiteness follows from `A.ncard = ℓ` via `Set.finite_of_ncard_ne_zero`, so I added that as an equivalence lemma alongside the shared definition.

SproutSeeds:
Yes. The closest existing entry is Erdős Problem 20. It asks for the general uniform sunflower threshold, while this paper fixes three petals and adds the restriction that every pairwise intersection has size at most `t`. The `restrictedThreshold` definition here is the corresponding restricted version.

Erdős Problem 857 is related more broadly, but it uses non-uniform families of subsets of `Fin n` and studies asymptotics in the ground-set size. I updated the introductory text to make that distinction explicit.

SproutSeeds:
-awaiting-author

PR #4270, opened 2026-06-15
Fix conjecture 100

Danie-I:
Fix of conjecture 100. There was missing invariant length.


FormalConjectures/WrittenOnTheWallII/GraphConjecture100.lean
```diff
@@ -77,7 +77,7 @@ theorem conjecture100 (G : SimpleGraph α) [DecidableRel G.Adj] (h : G.Connected
     (hGc : Gᶜ.Connected) :
     let maxL := (Finset.univ.image (indepNeighborsCard G)).max' (by simp)
     let diamCompl := Gᶜ.ediam.toNat
```
mo271:
```suggestion
```

mo271 (commented):
Thanks, LGTM! generally, but let's delete the `let` statement -- it is not needed anymore, right?

Danie-I:
Yes, it is not needed anymore.

mo271 (approved):
Thanks!

PR #4271, opened 2026-06-15
fix(GreensOpenProblems/72): add `2 < k`

mo271:
Otherwise we can disprove `NoKInLine` for k=2

Drive-by: also fix docstring, should read  $(k - 1)N$ instead of  $k * N$


felixpernegger (approved):
nice

YaelDillies (approved):
Looks good!

PR #3644, opened 2026-03-26
Add Erdos problem 993

shutongliu-dev:
Adds a formal statement of Erdos Problem 993, asserting that the independent set sequence of every finite forest is unimodal.

References:
- https://www.erdosproblems.com/993
- Alavi, Erdos, Malde, Schwenk, "The vertex independence sequence of a graph is not constrained" (Congressus Numerantium 58, 1987).

Closes #1058

PR #4272, opened 2026-06-15
Add Erdős Problem 993

williamjblair:
Formalizes the statement of [Erdős Problem 993](https://www.erdosproblems.com/993).

**Conjecture (Alavi–Erdős–Malde–Schwenk, 1987).** The independent-set sequence $i_0, i_1, \dots$ of every finite forest is unimodal, where $i_k$ is the number of independent sets of cardinality $k$. The sequence is unconstrained for general graphs, but unimodality is open for forests (and trees).

The new file `FormalConjectures/ErdosProblems/993.lean` adds:
- `numIndepSets` — the $k$-th term of the sequence, via `Nat.card` of the size-$k$ independent sets (no `DecidablePred` needed);
- `Unimodal` — nondecreasing-then-nonincreasing predicate on `ℕ → ℕ`;
- `erdos_993` — the forest-unimodality statement, `@[category research open, AMS 5]`, kept as `sorry`.

`lake build` passes (only the expected `declaration uses 'sorry'`). 993 was absent upstream.

### DIFF
--- a/FormalConjectures/ErdosProblems/993.lean
+++ b/FormalConjectures/ErdosProblems/993.lean
@@ -1,0 +1,61 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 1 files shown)


williamjblair:
Closing as a duplicate — I didn't notice that #3644 (by the assignee @landauros) and #4192 were already open for Erdős 993. Apologies for the noise; deferring to the existing work.

PR #4275, opened 2026-06-16
feat: add preview of informal statements to the website

Paul-Lez:
Once the build is over, the results should be visible here: https://paul-lez.github.io/formal-conjectures/


mo271 (approved):
Thanks, LGTM. 

Could be made such that there is a little less space used in the header of those boxes, but I guess this is a question of taste and can also be done in a follow up or not at all.

Paul-Lez:
Cool, may have a go at that in a follow-up!

PR #4279, opened 2026-06-16
Add new graph invariants

Danie-I:


FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Eccentricity.lean
```diff
@@ -40,4 +40,12 @@ converted to a real number. Returns 0 if the graph has no vertices. -/
 noncomputable def averageEccentricity (G : SimpleGraph α) : ℝ :=
   (∑ v : α, (G.eccent v).toNat) / (Fintype.card α : ℝ)
 
+/-- The periphery of a graph: the set of vertices of maximum eccentricity (equal to diameter). -/
+noncomputable def periphery (G : SimpleGraph α) : Finset α :=
+  Finset.univ.filter (fun v => G.eccent v = G.ediam)
```
mo271: this is verbatim the same def as `boundaryVertices` above, let's keep at most one

FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Eccentricity.lean
```diff
@@ -40,4 +40,12 @@ converted to a real number. Returns 0 if the graph has no vertices. -/
...
+
+/-- The size of the periphery (number of vertices of maximum eccentricity). -/
+noncomputable def peripherySize (G : SimpleGraph α) : ℕ :=
+  G.periphery.card
```
mo271: and then potentially rename this!

FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Degrees.lean
```diff
@@ -98,4 +98,8 @@ noncomputable def numTrianglesAtVertex (G : SimpleGraph α) [DecidableRel G.Adj]
 noncomputable def length (G : SimpleGraph α) [DecidableRel G.Adj] : ℝ :=
   Real.sqrt (∑ v, (G.degree v : ℝ) ^ 2)
```
mo271, line 99: pre-existing, but still worth fixing: `length` seems like a very bad name for the L2-norm of the degree sequence:  it shadows `Walk.length` and the common graph theory meaning of "length" (= number of edges). Something like `degreeL2Norm` would be much clearer.

FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Degrees.lean
```diff
@@ -98,4 +98,8 @@ noncomputable def numTrianglesAtVertex (G : SimpleGraph α) [DecidableRel G.Adj]
...
 
+/-- The number of vertices of degree 2 in `G`. -/
+def countDegreeTwo (G : SimpleGraph α) [DecidableRel G.Adj] : ℕ :=
```
mo271: seems like it would be a good idea to generalise the "2" to "k", note that we already have the "1" case called `pendantVertices`

mo271 (commented):
Thanks! left a few comments

FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Eccentricity.lean
Danie-I: Probably peripherySize is still better name here.

Danie-I:
Thank you for the comments. I applied the fixes.

FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Degrees.lean
```diff
@@ -95,7 +95,11 @@ noncomputable def numTrianglesAtVertex (G : SimpleGraph α) [DecidableRel G.Adj]
   ((G.cliqueFinset 3).filter (fun s => v ∈ s)).card
 
 /-- The length of a graph: the square root of the sum of the squares of degrees. -/
-noncomputable def length (G : SimpleGraph α) [DecidableRel G.Adj] : ℝ :=
+noncomputable def degreeL2Norm (G : SimpleGraph α) [DecidableRel G.Adj] : ℝ :=
```
mo271, line 98: this renaming will probably require some fixes downstream if it is used anywhere?

FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Eccentricity.lean
```diff
@@ -40,4 +40,8 @@ converted to a real number. Returns 0 if the graph has no vertices. -/
 noncomputable def averageEccentricity (G : SimpleGraph α) : ℝ :=
   (∑ v : α, (G.eccent v).toNat) / (Fintype.card α : ℝ)
 
+/-- The size of the periphery (number of vertices of maximum eccentricity). -/
+noncomputable def peripherySize (G : SimpleGraph α) : ℕ :=
+  G.boundaryVertices.card
```
mo271: I'm not sure this definition is useful at all, can it just be inlined?

Danie-I: It is somewhat useful, that we will have more direct translation of python invariants to Lean.

FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Degrees.lean
Danie-I: This is just added here, and I did not include any uses so far.

FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Eccentricity.lean
mo271: can you not use use the string "G.boundaryVertices.card" in every place where you use the string "peripherySize" now?

Danie-I: Sure.

Danie-I: Removed

FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Degrees.lean
Danie-I: Can this conversation be resolved?

mo271 (approved):

PR #4282, opened 2026-06-17
fix conjecture 133

Danie-I:
Fixed definition of path. It is also a fix for other conjectures using path invariant. Also cC4 is a characteristic function.


mo271 (approved):
Thanks, LGTM!

mo271:
@Danie-I please fix the `lake build`

Danie-I:
> @Danie-I please fix the `lake build`

This is caused by misformalization. For conjecture 36 K_2 is not a valid counterexample. Although there is a known counterexample.

Danie-I:
Should I just remove the wrong proof?

mo271:
> Should I just remove the wrong proof?
> For conjecture 36 K_2 is not a valid counterexample. Although there is a known counterexample.

yes, this should have been caught when submitting this /during review of 

https://github.com/google-deepmind/formal-conjectures/pull/3820

@henrykmichalewski

mo271 (approved):

PR #4285, opened 2026-06-17
feat: add difficulty voting feature

Paul-Lez:


mo271:
should be rebased after #4292

PR #2260, opened 2026-02-11
feat(ErdosProblems): 1049

danielchin:
Closes #1093 

Formalizing Erdos Problem 1049: https://www.erdosproblems.com/1049

Assisted by Gemini/Antigravity

PR #4286, opened 2026-06-17
feat(ErdosProblems/1049): prove lambert_series_eq_num_divisor_sum

kukushking:
# Description

> The classical Lambert series identity: for $t \in \mathbb{Q}$ with $|t| > 1$,
> $$\sum_{n=1}^\infty \frac{1}{t^n - 1} = \sum_{n=1}^\infty \frac{\tau(n)}{t^n}$$
> where $\tau(n) = d(n)$ is the number of divisors of $n$.
> [[Source]](https://www.erdosproblems.com/1049)

- Closes the textbook `sorry` on `lambert_series_eq_num_divisor_sum` introduced in #2260 (cc @danielchin).
- Two `private lemma`s + the main theorem combining them:
    - `lambert_convergent` (|t| > 1) — applies Mathlib's `tsum_pow_div_one_sub_eq_tsum_sigma` at `k = 0` with `r := t⁻¹`, then algebraic massage on both `tsum_congr` side goals.
    - `lambert_divergent` (|t| ≤ 1) — four-case split (`t = 1`, `t = 0`, `t = -1`, `|t| < 1 ∧ t ≠ 0`) with a shared `key` non-summability helper. Uses `Summable.tendsto_cofinite_zero` because the `atTop` variant doesn't apply cleanly to `ℕ+`-indexed functions.
    - Main theorem: `by_cases ht : 1 < |(t : ℝ)|`.
- Inline proof, ~150 lines. Above the 25-50 guideline in CONTRIBUTING.md — most of which is bookkeeping forced by the universal `t : ℚ` statement. The classical Lambert identity (Erdős 1948) is stated for `|t| > 1` only, where the proof is just `lambert_convergent` (~30 lines, well under the 25-50 guideline). Amending the statement to add `(ht : 1 < |t|)` would let the `lambert_divergent` helper be deleted entirely. See "Note on the theorem statement" below.
- :robot: Generated by a custom Lean-kernel-verified theorem-proving harness I built — an attempt loop driving Claude Opus 4.8 as the LLM backend, with targeted Mathlib-idiom retrieval injected per target and compiler-error feedback threaded between attempts. The harness produced the structural proof from a hint pack tuned for the Lambert family; I manually patched ~5% leaf-level tactic mistakes (Nat.card_pos.mpr → Finset.card_pos.mpr, n.pos.ne' substituted for a hallucinated Nat.pos_iff.mp, one redundant ring removed after field_simp already closed the goal).

## Note on the theorem statement

The current statement quantifies universally over `t : ℚ`. The classical Lambert identity (Erdős 1948) requires `|t| > 1` for both sums to converge. In the divergent case `|t| ≤ 1`, both sides reduce to `0 = 0` via Lean's `tsum`-of-non-summable = 0 convention rather than via mathematical content. The proof handles these cases honestly (showing `tsum = 0` for each sub-case) but the divergent branch is bookkeeping, not the Lambert identity proper.

If you'd prefer the statement be tightened to add `(ht : 1 < |t|)`, the proof shrinks to ~30 lines (just `lambert_convergent`, the divergent helper deletable). Happy to amend in this PR or follow up separately — your call.

### References

- [[Er48]](https://www.erdosproblems.com/1049) Erdős, P. *On arithmetical properties of Lambert series*. J. Indian Math. Soc. (N.S.) **12** (1948), 63–66.
- [[#2260]](https://github.com/google-deepmind/formal-conjectures/pull/2260) Chin, D. Original formalization PR for Erdős Problem 1049.

# Testing

:white_check_mark: Builds cleanly:
```shell
$ lake build FormalConjectures.ErdosProblems.«1049»
Build completed successfully (8795 jobs).
```

:white_check_mark: Kernel axiom check — only Lean foundation axioms, no `sorryAx`, no user axioms:
```
'Erdos1049.lambert_series_eq_num_divisor_sum' depends on axioms: [propext, Classical.choice, Quot.sound]
```

:white_check_mark: Theorem statement is byte-identical to the original from #2260 (no signature drift).

### DIFF
--- a/FormalConjectures/ErdosProblems/1049.lean
+++ b/FormalConjectures/ErdosProblems/1049.lean
@@ -49,40 +49,75 @@ theorem erdos_1049.variants.geq_2_integer :
      ∀ t : ℤ, t ≥ 2 → Irrational (∑' n : ℕ+, 1 / ((t : ℝ) ^ (n : ℕ) - 1)) := by
   sorry
 
-/-- Convergent case (|t| > 1): direct from Mathlib's `tsum_pow_div_one_sub_eq_tsum_sigma`. -/
+/--
+Convergent case (`|t| > 1`).
+
+Substitute `r := t⁻¹` so `‖r‖ < 1`, then apply Mathlib's series identity
+`tsum_pow_div_one_sub_eq_tsum_sigma` at `k = 0`:
+$$\sum_{n \ge 1} \frac{r^n}{1 - r^n} = \sum_{n \ge 1} \sigma_0(n) \cdot r^n.$$
+After clearing denominators, both sides match the Lambert identity:
+LHS becomes `1/(t^n - 1)` and RHS becomes `τ(n) / t^n`.
+-/
 @[category API, AMS 11]
 private lemma lambert_convergent (t : ℝ) (ht : 1 < |t|) :
     ∑' n : ℕ+, 1 / (t ^ (n : ℕ) - 1) =
     ∑' n : ℕ+, ((n : ℕ).divisors.card : ℝ) / (t ^ (n : ℕ)) := by
+  -- `|t| > 1` implies `t ≠ 0`, hence `t^n ≠ 0` for all n.
   have ht0 : t ≠ 0 := fun h => by subst h; simp at ht; linarith [abs_nonneg (0:ℝ)]
   have htn : ∀ n : ℕ, t ^ n ≠ 0 := fun n => pow_ne_zero n ht0
+  -- Substitution `r := t⁻¹`, so `|r| < 1`.
   set r : ℝ := t⁻¹ with hr_def
   have hr_norm : ‖r‖ < 1 := by
     rw [Real.norm_eq_abs, hr_def, abs_inv]; exact inv_lt_one_of_one_lt₀ ht
+  -- Apply the Mathlib identity. Now reduce each side of our goal to its form.
   have h := tsum_pow_div_one_sub_eq_tsum_sigma (k := 0) hr_norm
   convert h using 1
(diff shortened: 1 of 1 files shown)


danielchin:
I think this proof is too large to be in this repo. Generally, it looks like the maintainers are okay with [proofs in the 25-50 line range](https://github.com/google-deepmind/formal-conjectures/blob/main/CONTRIBUTING.md#ways-to-contribute).

You could check this into a separate branch or in your own repo and provide a link to the proof [`@[formal_proof]` attribute](https://github.com/google-deepmind/formal-conjectures/blob/main/CONTRIBUTING.md#ways-to-contribute).

kukushking:
Thanks for taking a look! In the PR body I mentioned two alternatives for the length concern:

  1. **Tighten the statement to add `(ht : 1 < |t|)`** — most of the proof's bulk (~100 of 150 lines) is `lambert_divergent` handling `|t| ≤ 1`, which only exists because the current statement quantifies universally over `ℚ`. The classical Lambert identity (Erdős 1948) is stated for `|t| > 1` only. With the hypothesis added, `lambert_divergent` is deletable and the proof fits inline at ~30 lines (just `lambert_convergent`).

  2. **Fork+link migration** (the path you suggested) — straightforward to execute, keeps the statement universal.

  Since you authored the original formalization in #2260, context on whether the universal-over-`ℚ` scope was intentional would be useful. I'd lean toward (1) since it brings the formal statement closer to the source paper, but happy to pursue either path once a maintainer weighs in.

mo271 (approved):
Thanks looks great!

Perhaps it would be worth considering moving this to an appropriate place in `FormalConjecturesForMathlib`, and upstream the result to Mathlib eventually

But that can be done in a follow-up

kukushking:
Thank you! 🥰 Yes I will follow up with Mathlib!

PR #4289, opened 2026-06-18
fix(ErdosProblems/264): type the bounded perturbation b over the integers

### LINKED ISSUE #4264: Erdős Problem 264 — misformalisation: `IsIrrationalitySequence` types the bounded perturbation `b` as `ℕ → ℕ`, excluding the negative integers the problem intends
The current formalization of [`erdos_264`](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/264.lean) defines the central predicate with the bounded perturbation `b` typed `ℕ → ℕ`. The informal statement (and [erdosproblems.com/264](https://www.erdosproblems.com/264)) quantifies over a **bounded sequence of integers** `b`. Restricting `b` to `ℕ` excludes every negative perturbation, which makes the predicate **strictly weaker** than intended — and makes one of its two hypotheses vacuous.

## The current formalization

```lean
/--
A sequence $a_n$ of integers is called an irrationality sequence if for every bounded sequence
of integers $b_n$ with $a_n + b_n \neq 0$ and $b_n \neq 0$ for all $n$, the sum
$\sum \frac{1}{a_n + b_n}$ is irrational.
-/
def IsIrrationalitySequence (a : ℕ → ℕ) : Prop := ∀ b : ℕ → ℕ, BddAbove (Set.range b) →
  0 ∉ Set.range (a + b) → 0 ∉ Set.range b → Irrational (∑' n, (1 : ℝ) / (a n + b n))
```

The docstring says **"bounded sequence of integers `b_n`"** (twice). The formal `b : ℕ → ℕ` admits only non-negative `b`.

## Why this is a misformalisation (two independent reasons)

**1. The universal becomes strictly weaker.** `IsIrrationalitySequence a` is `∀ b … `. Shrinking `b` from `ℤ`-valued to `ℕ`-valued shrinks the family of admissible `b`, so the predicate becomes *easier* to satisfy. A proof for `Nat.factorial` (the open part `erdos_264.parts.ii`) would then establish irrationality only against non-negative perturbations — not the integer-perturbation statement the problem asks about.

**2. A hypothesis becomes vacuous — the statement's own structure betrays the intended type.** With `b : ℕ → ℕ`, the hypothesis `0 ∉ Set.range b` says `b n ≠ 0`, i.e. `b n ≥ 1`. Then for *every* `a : ℕ → ℕ`, `(a + b) n = a n + b n ≥ 1 > 0`, so `0 ∉ Set.range (a + b)` holds automatically — the second hypothesis is **redundant** under the current typing. This is machine-checkable in a few lines of **core Lean** (no imports; compiles `sorry`-free):

```lean
theorem second_hyp_redundant_over_nat
    (a b : Nat → Nat) (hb : ∀ n, b n ≠ 0) : ∀ n, a n + b n ≠ 0 := by
  intro n h
  have hbn := hb n
  omega
```

The informal problem lists `a_n + b_n ≠ 0` and `b_n ≠ 0` as **two independent conditions**, which is only meaningful when `b_n` may be **negative** (so `a_n + b_n` can hit `0` while `b_n` does not). This matches the canonical definition in Kovač–Tao 2024 (already cited in this file as `[KoTa24]`): the sum is irrational *"for every bounded sequence of integers `(b_n)` such that `b_n ≠ 0` and `a_n + b_n ≠ 0`"* — `b_n` ranges over `ℤ`. (Their *example constructions* use positive perturbations, but the **definition** is over `ℤ`.)

**Secondary:** `BddAbove (Set.range b)` captures "bounded" only because `ℕ` is bounded below for free; a genuine integer sequence needs both bounds (`BddAbove ∧ BddBelow`, i.e. `Bornology.IsBounded (Set.range b)`).

## Proposed fix

Type the perturbation over `ℤ`, add the lower bound, and coerce in the two arithmetic spots:

```lean
def IsIrrationalitySequence (a : ℕ → ℕ) : Prop := ∀ b : ℕ → ℤ,
    BddAbove (Set.range b) → BddBelow (Set.range b) →
    0 ∉ Set.range (fun n ↦ (a n : ℤ) + b n) → 0 ∉ Set.range b →
    Irrational (∑' n, (1 : ℝ) / ((a n : ℤ) + b n))
```

(Alternatively, type both `a b : ℕ → ℤ` to match "sequence of integers" literally; that cascades coercions into the example theorems, so the minimal targeted change above is proposed instead. The preferred encoding of "bounded integer sequence" is of course the maintainers' call.) 
The fix is the small change above — **up for grabs**; happy for a maintainer or contributor to apply it, and I may follow up with a PR once I'm set up to run the full FC build.

## How this was found

Surfaced by an automated formal-vs-informal fidelity screen over the open `ErdosProblems` set, then confirmed by manual reading and the elementary redundancy argument above (the load-bearing evidence is the manual/structural confirmation, not the screen).

## Scope (what this does and doesn't affect)

For the *solved negative* results this typing is harmless: disproving that a sequence is an irrationality sequence needs only *one* bad bounded perturbation, and a **positive** one already suffices — e.g. for `2^n` it is enough to range `b` over `{1,2,3,4}` (cf. the erdosproblems.com/264 discussion). So `erdos_264.parts.i` and the `variants` are unaffected by the `ℕ → ℤ` change.

The gap bites for (a) the **open** `parts.ii` (`n!`), whose answer under non-negative perturbations may differ from the integer-perturbation question Erdős poses, and (b) **fidelity** of the definition to the stated problem — verbatim *"for every bounded sequence of integers b_n (with a_n + b_n ≠ 0 and b_n ≠ 0)"* (erdosproblems.com/264).

Isthisreel:
# Fix erdos_264: type the bounded perturbation `b` over `ℤ` (currently `ℕ`)

The predicate `IsIrrationalitySequence` currently quantifies the bounded perturbation `b` over `ℕ → ℕ`. The informal statement and [erdosproblems.com/264](https://www.erdosproblems.com/264) quantify over a **bounded sequence of integers**. Restricting to `ℕ` excludes all negative perturbations, which (a) makes the `∀ b` strictly weaker than intended and (b) makes the hypothesis `0 ∉ Set.range (a + b)` vacuous — given `0 ∉ Set.range b` (so `b n ≥ 1`), we have `a n + b n ≥ 1 > 0` for free. The two conditions `a_n + b_n ≠ 0` and `b_n ≠ 0` are independent only when `b` may be negative.

## Change

```diff
@@ FormalConjectures/ErdosProblems/264.lean
-def IsIrrationalitySequence (a : ℕ → ℕ) : Prop := ∀ b : ℕ → ℕ, BddAbove (Set.range b) →
-  0 ∉ Set.range (a + b) → 0 ∉ Set.range b → Irrational (∑' n, (1 : ℝ) / (a n + b n))
+def IsIrrationalitySequence (a : ℕ → ℕ) : Prop := ∀ b : ℕ → ℤ,
+  BddAbove (Set.range b) → BddBelow (Set.range b) →
+  0 ∉ Set.range (fun n ↦ (a n : ℤ) + b n) → 0 ∉ Set.range b →
+  Irrational (∑' n, (1 : ℝ) / ((a n : ℤ) + b n))
```

This is the minimal change that fixes the reported bug (the type of `b`). An alternative is to type both `a b : ℕ → ℤ` to match "sequence of integers" literally; that cascades coercions into `parts.i`, `parts.ii`, `variants.example`, `variants.ko_tao_neg/pos`, so we propose the targeted change and defer the encoding choice to the maintainers.

## Why it matters

This is the same class of bug as the `erdos_274` quantifier issue and `erdos_707` (#1137): a small typing/quantifier choice yields logically valid Lean whose `answer` no longer matches the natural-language statement. Here the tell is internal: a hypothesis of the predicate is vacuous under the current type.

## Verification

- Source pinned: the change is against `264.lean` at sha256 recorded in the accompanying certificate.
- The redundancy that motivates the fix is captured as a machine-checkable lemma in `witness.lean` (`second_hypothesis_redundant_over_nat`), with an independence example over `ℤ`.
- **Compiles cleanly** under `lake --wfail build` (Lean / Mathlib v4.27.0, warnings-as-errors, 0 errors). All five theorems referencing `IsIrrationalitySequence` (`parts.i`, `parts.ii`, `variants.example`, `variants.ko_tao_neg`, `variants.ko_tao_pos`) are `sorry`, so retyping `b` breaks no proof.

## Detection

Surfaced by an automated formal-vs-informal fidelity screen, confirmed by manual reading and the elementary redundancy argument. Reported as a candidate, not a claim.

Closes #4264


mo271 (approved):
Thanks, LGTM!

PR #4290, opened 2026-06-18
fix(ErdosProblems/562): use same-order (IsTheta) not asymptotic equivalence

### LINKED ISSUE #4265: Erdős Problem 562 — misformalisation: `~[atTop]` (asymptotic equivalence) used where the statement says `≍` (same order / Θ)

[`erdos_562`](https://github.com/google-deepmind/formal-conjectures/blob/71d34fc2958f1010ee43ccd4d54a9631574d7ef9/FormalConjectures/ErdosProblems/562.lean#L38-L41) relates `log^[r-1] (R_r(n))` and `n` by **`~[atTop]`** (`Asymptotics.IsEquivalent`), but the informal statement asks for **`≍_r`** (Hardy's `\asymp`, i.e. same order of magnitude, `Θ`, with `r`-dependent implied constants). `~` is **strictly stronger** than `≍`, so the formal statement over-constrains the problem.

```lean
theorem erdos_562 : answer(sorry) ↔
    ∀ r ≥ 3, (fun n ↦ log^[r - 1] (hypergraphRamsey r n)) ~[atTop] (fun n ↦ (n : ℝ)) := by
  sorry
```

In Mathlib `f ~[l] g` (`Asymptotics.IsEquivalent`) is `(f - g) =o[l] g`, i.e. `f / g → 1` — strictly stronger than `f =Θ[l] g`. So `~[atTop]` demands the iterated logarithm be asymptotically *equal* to `n` with leading constant exactly `1`; the intended `≍` only asks it be of the same *order*. A solver proving the intended `Θ`-bound would not satisfy the formalized `~`, and the formal statement may be false as written (no reason the constant is `1`).

### Trivial proof/disproof

`~ ⟹ Θ` is a Mathlib lemma (`Asymptotics.IsEquivalent.isTheta`), `sorry`-free:

```lean
example (f g : ℕ → ℝ) (h : f ~[atTop] g) : f =Θ[atTop] g := h.isTheta
```

The converse fails — `(fun n ↦ 2 * n) =Θ[atTop] (fun n ↦ n)` but **not** `(fun n ↦ 2 * n) ~[atTop] (fun n ↦ n)` (ratio → `2`, not `1`). So `~` is *strictly* stronger than `≍`/`=Θ`: the current statement is a different (stronger, likely false) claim than the one Erdős poses.

### Suggested fix

Replace `~[atTop]` with `=Θ[atTop]` (`Asymptotics.IsTheta`), which is exactly `≍` (`=Θ[·]` is already in scope via `open scoped Asymptotics`):

```lean
    ∀ r ≥ 3, (fun n ↦ log^[r - 1] (hypergraphRamsey r n)) =Θ[atTop] (fun n ↦ (n : ℝ))
```

### Additional context

Related to — but distinct from — the open discussion in #33 (*`answer(sorry)` and asymptotic problems*): that asks whether an `answer(sorry)` captures *optimality*; this is the *relation* (`≍` vs `~`). Surfaced by an automated formal-vs-informal fidelity screen, then confirmed by reading the docstring's `\asymp` against the Mathlib meaning of `~[atTop]` plus the lemma above.

The fix is the single-token change above. Flagging it as **up for grabs** — happy for a maintainer or contributor to apply it; I may follow up with the PR once I'm set up to run the full FC build.

Isthisreel:
# Fix erdos_562: use `=Θ[atTop]` (the `≍` of the statement) instead of `~[atTop]`

The statement asks to prove `log_{r-1} R_r(n) ≍_r n` (`\asymp`, same order / `Θ`). The formalization uses `~[atTop]` = `Asymptotics.IsEquivalent` (ratio `→ 1`), which is **strictly stronger** than `≍`/`=Θ[atTop]`. This over-constrains the problem and may make the formal statement false (the leading constant need not be `1`).

## Change

```diff
@@ FormalConjectures/ErdosProblems/562.lean
 @[category research open, AMS 5]
 theorem erdos_562 : answer(sorry) ↔
-    ∀ r ≥ 3, (fun n ↦ log^[r - 1] (hypergraphRamsey r n)) ~[atTop] (fun n ↦ (n : ℝ)) := by
+    ∀ r ≥ 3, (fun n ↦ log^[r - 1] (hypergraphRamsey r n)) =Θ[atTop] (fun n ↦ (n : ℝ)) := by
   sorry
```

## Why it matters

`f ~[l] g` (`IsEquivalent`, `(f-g)=o[l] g`, i.e. `f/g→1`) implies but is not implied by `f =Θ[l] g` (`IsTheta`):
- `Asymptotics.IsEquivalent.isTheta` gives `~ ⟹ Θ`.
- Converse fails: `2•n =Θ n` yet `¬ (2•n ~ n)`.

The intended/known statements about iterated logs of hypergraph Ramsey numbers are order-of-magnitude (`≍`), so `~` is the wrong relation.

## Verification

- Source pinned: change is against `562.lean` at the sha256 in the accompanying certificate.
- `witness.lean` carries the load-bearing implication `IsEquivalent.isTheta` (clean) and the `2n`/`n` strictness counterexample (the routine limit step is marked `sorry`, with the argument spelled out — `isEquivalent_iff_tendsto_one`).
- **Compiles cleanly** under `lake --wfail build` (Lean / Mathlib v4.27.0, warnings-as-errors, 0 errors). The change is a single token (`~[atTop]` → `=Θ[atTop]`); `=Θ` is in scope via `open scoped Asymptotics`.

## Detection

Automated fidelity screen, confirmed by manual reading of `\asymp` vs `~[atTop]`. Candidate, not a claim.

Closes #4265


mo271 (approved):
Thanks, LGTM

PR #4299, opened 2026-06-22
ErdosProblems/17: prove isClusterPrime_97_isLeast_non_cluster

williamjblair:
Closes the `sorry` on `isClusterPrime_97_isLeast_non_cluster` (97 is the least prime that is not a cluster prime).

### Approach

For a prime `p`, `IsClusterPrime p` is reformulated as a fully bounded, decidable statement (`cluster_iff`): every even `n ∈ Finset.range (p - 2)` admits a prime `q₂ ≤ p` with `q₂ + n` prime and `≤ p`.

- **Membership** (`97 ∈ S`): `97` is prime and not a cluster prime. The even number `88 ≤ 94` is not a difference of two primes `≤ 97`, discharged by `decide` through `cluster_iff`.
- **Lower bound** (`97` is a lower bound of `S`): `interval_cases` over `b < 97`; each case is either non-prime (`decide`) or a cluster prime (`cluster_iff` + `decide`).

### Verification

Kernel-clean: `#print axioms isClusterPrime_97_isLeast_non_cluster` reports `[propext, Classical.choice, Quot.sound]` — no `sorryAx`. No `native_decide` (plain `decide` with a raised `maxRecDepth`). Built against this repo's pinned toolchain (`leanprover/lean4:v4.27.0`) and Mathlib rev.

The proof was drafted with AI assistance and machine-checked by the Lean kernel.

### DIFF
--- a/FormalConjectures/ErdosProblems/17.lean
+++ b/FormalConjectures/ErdosProblems/17.lean
@@ -72,6 +72,49 @@ theorem erdos_17.variants.upper_Elsholtz :
 /-- $97$ is the smallest prime that is not a cluster prime. -/
 @[category test, AMS 11]
 theorem isClusterPrime_97_isLeast_non_cluster : IsLeast {p : ℕ | p.Prime ∧ ¬ IsClusterPrime p} 97 := by
-  sorry
+  -- For a prime `p`, being a cluster prime is equivalent to a fully bounded
+  -- (hence decidable) statement over `ℕ`: every even `n ≤ p - 3` is `q₁ - q₂`
+  -- for primes `q₁, q₂ ≤ p`, which (since `n ≥ 0`) amounts to a prime `q₂ ≤ p`
+  -- with `q₂ + n` again prime and `≤ p`.
+  have cluster_iff : ∀ p : ℕ, p.Prime →
+      (IsClusterPrime p ↔
+        (∀ n ∈ Finset.range (p - 2), Even n →
+          ∃ q₂ ∈ Finset.range (p + 1), (Nat.Prime q₂) ∧ Nat.Prime (q₂ + n) ∧ q₂ + n ≤ p)) := by
+    intro p hp
+    constructor
+    · rintro ⟨-, h⟩ n hn hev
+      simp only [Finset.mem_range] at hn
+      have hn3 : (n : ℤ) ≤ (p - 3 : ℤ) := by omega
+      obtain ⟨q₁, q₂, hq1p, hq2p, hq1le, hq2le, heq⟩ := h hev hn3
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM.

I assume the `maxRecDepth 8000` doesn't make lake build on this file unreasonably slow...

williamjblair:
I don't believe so, but I'll check to make sure

williamjblair:
@mo271 Checked: It is ~7s of proof work over the ~8s Mathlib import (warm). The 8000 is far more headroom than needed and the real floor is ~1385. Happy to tighten it to 2000 if you'd prefer

While I was looking I noticed a few native_decide calls small enough to be plain decide (e.g. OEIS/306477 _3/_6, ErdosProblems/445), which makes them axiom-clean at no real cost. The heavier ones (387, 389, 730, 686, …) genuinely need native_decide, so I'd leave those. Could be worth a small PR for the cheap ones, or do you prefer keeping native_decide for consistency?

mo271:
> @mo271 Checked: It is ~7s of proof work over the ~8s Mathlib import (warm). The 8000 is far more headroom than needed and the real floor is ~1385. Happy to tighten it to 2000 if you'd prefer
> 
> While I was looking I noticed a few native_decide calls small enough to be plain decide (e.g. OEIS/306477 _3/_6, ErdosProblems/445), which makes them axiom-clean at no real cost. The heavier ones (387, 389, 730, 686, …) genuinely need native_decide, so I'd leave those. Could be worth a small PR for the cheap ones, or do you prefer keeping native_decide for consistency?


Thanks for checking! That is ok, no need to tighten those numbers, if we make them too small it might blow up when we bump to a new Lean/Mathlib version.

Every native-decide that can be replaced by a decide without unreasonably increasing the `build` time is a positive change. I don't think we strive for consistency in this aspect.

williamjblair:
> > @mo271 Checked: It is ~7s of proof work over the ~8s Mathlib import (warm). The 8000 is far more headroom than needed and the real floor is ~1385. Happy to tighten it to 2000 if you'd prefer
> > While I was looking I noticed a few native_decide calls small enough to be plain decide (e.g. OEIS/306477 _3/_6, ErdosProblems/445), which makes them axiom-clean at no real cost. The heavier ones (387, 389, 730, 686, …) genuinely need native_decide, so I'd leave those. Could be worth a small PR for the cheap ones, or do you prefer keeping native_decide for consistency?
> 
> Thanks for checking! That is ok, no need to tighten those numbers, if we make them too small it might blow up when we bump to a new Lean/Mathlib version.
> 
> Every native-decide that can be replaced by a decide without unreasonably increasing the `build` time is a positive change. I don't think we strive for consistency in this aspect.

Ok thanks!

PR #4301, opened 2026-06-23
feat(website/stats): growth plot

mo271:
This adds the growth plot to the stats page, preparing the removal of docgen, where we currently inject this plot.


jeangud (approved):
LGTM! Thank you @mo271 for preparing the follow up from #4167. We will be able to remove some extra build libs.

Tested locally:
image

PR #4302, opened 2026-06-23
Remove docgen

mo271:
This completes removes docgen: from the website and the build scripts and remvoes the entire `./doc` subdir. 

The build times for docgen have been unreasonably long (> ~2h typically), and it has been it by now provides little over verso which has a more acceptable build time (still long!) and *doesn't hide some information, like our category tags*!. (Also currently some pages in docgen are broken).
One thing that we potentially want to expand on: currently there is verso build missing for `FormalConjecturesForMathlib` (issue #4306 )


mo271:
depends on #4301

.github/workflows/build-and-docs.yml
```diff
@@ -67,17 +67,6 @@ jobs:
...
-      - name: Generate All.lean
-        if: steps.mode.outputs.website_only != 'true'
-        run: |
-          lake exe mk_all --lib FormalConjectures || true
-          # Avoid including Test files from `Util/` to avoid inflating the
-          # stats. This stop-gap should be removed after doing the TODO at
-          # `fetchStatsMarkdown` in `docbuild/overwrite-index.lean`.
-          grep -v -E "FormalConjectures\.Util\.|FormalConjectures\.All" FormalConjectures.lean > FormalConjectures/All.lean
-          echo "" >> FormalConjectures/All.lean
-          echo "set_option linter.style.moduleDocstring false" >> FormalConjectures/All.lean
```
eric-wieser, line 79: Note that this file arguably provides some value; it ensures that no two problem files define the same name.

lakefile.toml
```diff
@@ -29,6 +29,11 @@ name = "mathlib"
 scope = "leanprover-community"
 rev = "v4.27.0"
 
+[[require]]
+name = "verso"
+git = "https://github.com/leanprover/verso.git"
+rev = "v4.27.0"
```
eric-wieser: Why do we have this when we didn't before? If this is only a requirement for the website, I think we might still want to maintain a separate lakefile.

.github/workflows/build-and-docs.yml
mo271: I suppose we could keep that step and then lake build that file as well. Is there a more elegant way to check for not defining the same name twice?

lakefile.toml
mo271: I guess I moved it to here in order not to have a docbuild dir at all anyore, and it seems like the standard way of using verso, no?

I suppose we can keep it in a separate lakefile, is this best practice?

.github/workflows/build-and-docs.yml
eric-wieser: Not really, the only other option would be to run `importModules` from a lean script and see if that fails; but that's much harder to orchestrate than just running `mk_all`.

You can probably now drop the `grep` parts here.

lakefile.toml
eric-wieser: >  it seems like the standard way of using verso, no?

I think this is for when your project is primarily verso (it's a book / manual / blog / website)

But yes, I think a separately dir is best-practice for verso, especially since verso can even use a different version of lean to the code that it is rendering.

mo271:
another reason to remove it: flaky build failures: https://github.com/google-deepmind/formal-conjectures/actions/runs/29940416973/attempts/1

lakefile.toml
mo271: done

.github/workflows/build-and-docs.yml
mo271: put back a build step generating All.lean (but deleting it before verso is build)

mo271:
https://github.com/mo271/formal-conjectures/actions/runs/30005165766/job/89199330510 trying out a build on my fork, so also see if the website still renders nicely with those changes

Paul-Lez (approved):
Nice!

PR #4307, opened 2026-06-23
ErdosProblems/18: prove practicalH_six and practicalH_twelve

williamjblair:
Proves `practicalH_six` (`practicalH 6 = 2`) and `practicalH_twelve` (`practicalH 12 = 3`), the two small open values in the file.

Both are a straightforward `le_antisymm`. For the upper bounds, every `m` up to `n` is a sum of at most `h(n)` divisors of `n`, checked case by case with an explicit witness. For the lower bounds, the witnesses are `m = 4` (needs two divisors, `4 = 1 + 3`) for six and `m = 11` (needs three, `11 = 1 + 4 + 6`) for twelve.

Plain `decide` throughout, no `native_decide`, so `#print axioms` stays at `[propext, Classical.choice, Quot.sound]`. Builds on the pinned toolchain.

Complementary to the open #4168, which handles the large values (56, 7200) via `formal_proof` links; this closes the small ones.

Drafted with AI assistance and machine-checked by the Lean kernel.

### DIFF
--- a/FormalConjectures/ErdosProblems/18.lean
+++ b/FormalConjectures/ErdosProblems/18.lean
@@ -64,13 +64,97 @@ theorem practicalH_two : practicalH 2 = 1 := by
 m=4 or m=5, each requiring 2 divisors: 4=1+3, 5=2+3. -/
 @[category test, AMS 11]
 theorem practicalH_six : practicalH 6 = 2 := by
-  sorry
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

williamjblair:
Pushed an update after your approval to fold in practicalH_twelve so it closes both small values in one PR; the six proof is unchanged. Mind a quick second look at the twelve addition before merge?

mo271 (approved):

PR #4308, opened 2026-06-23
feat: Add anti-homogenous groups

ldct:
Add an open problem in group theory


mo271 (approved):
Thanks for the contribution, nice problem!

PR #4314, opened 2026-06-24
Wikipedia/IdonealCompleteness: prove knownIdonealNumbers_are_idoneal

williamjblair:
Proves `knownIdonealNumbers_are_idoneal` — all 65 known idoneal numbers are in fact idoneal.

The work is making the predicate decidable. `IsIdoneal n` is `0 < n ∧ ¬ ∃ a b c, 0 < a < b < c ∧ n = a*b + b*c + a*c`, and that existential ranges over unbounded `ℕ`. The helper `exists_triple_iff_bounded` reduces it to a bounded double-search: for a fixed pair `a, b ≤ n` the constraint pins `c = (n − a*b)/(a+b)` exactly, so only `a, b` are searched (over `Finset.range (n+1)`), with `a, b ≤ n` falling out of `nlinarith` on `0 < a < b < c`. The main proof then `fin_cases` over the 65 numbers, peels off the trivial `0 < n`, and discharges each bounded search.

This one needs `native_decide`: plain `decide` is infeasible — the kernel's `Finset` reduction hits `maxRecDepth` on a bounded search as small as 100×100, long before the `n = 1848` cases. So `#print axioms` includes `Lean.ofReduceBool` (the compiler-checked-computation axiom) alongside the usual `[propext, Classical.choice, Quot.sound]`; no `sorryAx`. The file builds in ~13s.

Only `knownIdonealNumbers_are_idoneal` is touched; the research-open `idoneal_numbers_completeness` keeps its `sorry`.

Drafted with AI assistance and machine-checked by the Lean kernel.

### DIFF
--- a/FormalConjectures/Wikipedia/IdonealCompleteness.lean
+++ b/FormalConjectures/Wikipedia/IdonealCompleteness.lean
@@ -54,10 +54,44 @@ def knownIdonealNumbers : Finset ℕ :=
    112, 120, 130, 133, 165, 168, 177, 190, 210, 232, 240, 253, 273, 280, 312,
    330, 345, 357, 385, 408, 462, 520, 760, 840, 1320, 1365, 1848}
 
+/--
+Reduces the unbounded search for a representation `n = a*b + b*c + a*c` (with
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM

Issue #3998, opened 2026-05-13
`plby` formalisations: sync status/update `formal_proof` tags + add new formalisations


mo271:
https://github.com/plby/lean-proofs/blob/main/data/sources.yaml

contains many lean formalistaions of Erdős Problems. Some of them we already link with `formal_proof`, some of them we should link with `formal_proof using lean4 ...` or `formal_proof using formal_conjectures` if the formal statement is verbatim what we have in our repo. Also it contains some formalisation of  (solved)  Erdős problems which haven't yet in our repository. Those should be added as well. 

A first step would be to run a script comparing `sources.yaml` against what is produced with our `extract_names` script, i.e. https://google-deepmind.github.io/formal-conjectures/data/conjectures.json comparing with the "formalProofLink" there and seeing what we are missing. 

Note: some of those formalisations are incomplete in the sense that they assume  some hyporthesis and only prove an implication. Those are marked in the file and also with `partial: yes` in the yaml. Those should not get a `formal_proof` attribute, but we should instead just link the proof manually in the docstring.

mo271:
cc-ing @Smetalo @franzhusch, who have been good at keeping track of these things...

Smetalo:
Just for completeness: the folder containing [sources.yaml](https://github.com/plby/lean-proofs/blob/main/data/sources.yaml) also contains [more_sources.yaml](https://github.com/plby/lean-proofs/blob/main/data/more_sources.yaml) and [urls.yaml](https://github.com/plby/lean-proofs/blob/main/data/urls.yaml).

plby:
FYI: `sources.yaml` is all the sources for Erdős problems.  `more_sources.yaml` has sources for other things in the repo, which are mostly "100 Theorems" and MathOverflow content.  `urls.yaml` are the raw URLs for when people post giant `live.lean-lang.org` links and I don't want to include them directly into the other files.

plby:
Oh, and I think `partial: yes` is slightly different from what @mo271 says?

`partial: yes` means it does not solve the "full" problem as listed on erdosproblems.com .  But often this corresponds to a specific statement in this project.  For example, `Erdos124` has `partial: yes` and it corresponds exactly to https://github.com/google-deepmind/formal-conjectures/blob/25c464cf57b537083c209d707a5bef4262554f3f/FormalConjectures/ErdosProblems/124.lean#L43-L46
I may start tracking which formulation of the problem here is covered.

[...]

Issue #993, opened 2025-10-11
Erdős Problem 871


mo271:
### What is the conjecture

https://www.erdosproblems.com/871

  Let $A$ be an additive basis of order $2$, and suppose $1_A\ast 1_A(n)\to \infty$ as $n\to \infty$. Can $A$ be partitioned into two disjoint additive bases of order $2$?

Status:  solved

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

plby:
Note that this problem has a formalized proof at https://www.erdosproblems.com/forum/thread/871#post-2733 so in particular it has some kind of statement.

On the one hand, this also this means the problem isn't labeled as open anymore, so probably of lower interest.  On the other hand, easier to write the statement?

plby:
In particular, this problem is no longer labeled as "open" and so can be removed from https://github.com/google-deepmind/formal-conjectures/milestone/1 I think?

Smetalo:
Is this resolved in #4319?

Paul-Lez:
Triage: Erdős Problem 871 is now represented in `FormalConjectures/ErdosProblems/871.lean`.

PR #4319, opened 2026-06-25
ErdosProblems: formalise 11 solved problems, linking plby/lean-proofs

### LINKED ISSUE #919: Erdős Problem 729
### What is the conjecture

https://www.erdosproblems.com/729

  Let $C>0$ be a constant. Are there infinitely many integers $a,b,n$ with $a+b> n+C\log n$ such that the denominator of 
$$\frac{n!}{a!b!}$$
contains only primes $\ll_C 1$?

Status:  solved

### Choose either option

williamjblair:
Part of #3998. Closes #919 (Erdős 729) and #993 (Erdős 871).

Adds statements for 11 solved Erdős problems FC doesn't have yet, each linking plby's proof in `src/v4.29.1` via `formal_proof using lean4`. Statement-only; the proofs stay in plby's repo. All files are ports of @plby's hosted formalizations.

A small first batch. I have ~100 more done locally and would rather you sign off on the conventions before I send the rest. I checked each statement against its erdosproblems.com page, not only that it compiles. One caveat: in 178 the sequences range over ℕ where the source says ℤ (non-weakening, but the one spot that isn't a literal transcription).

Who formalized each (from plby's `sources.yaml`):

```
134: "This was formalized in Lean by Boris Alexeev using Aristotle."
178: "This was formalized in Lean by Matteo Del Vecchio using Aristotle."
493: "This was formalized in Lean by Zheng Yuan using Seed-Prover 1.5, Aristotle, and ChatGPT."
532: "This was formalized in Lean by David Wärn."
729: "This was formalized in Lean by Kevin Barreto using Aristotle."
871: "This was formalized in Lean by Daniel Larsen using Claude Opus 4.5 and Gemini 3 Pro."
904: "This was formalized in Lean by Parcly Taxel using Aristotle."
923: "This was formalized in Lean by Parcly Taxel using Aristotle."
1090: "This was formalized in Lean by JoshuaB using Aristotle and Gemini 3.0 Flash."
1125: "This was formalized in Lean by Stefano Rocca using Aristotle."
1126: "This was formalized in Lean by JoshuaB using Aristotle."
```

The conditional-on-axiom problems will come separately, with a docstring note rather than `formal_proof`, per the discussion above.

### DIFF
--- a/FormalConjectures/ErdosProblems/1090.lean
+++ b/FormalConjectures/ErdosProblems/1090.lean
@@ -19,13 +19,17 @@ import FormalConjectures.Util.ProblemImports
 /-!
 # Erdős Problem 1090
 
-*Reference:* [erdosproblems.com/1090](https://www.erdosproblems.com/1090)
+*References:*
+- [erdosproblems.com/1090](https://www.erdosproblems.com/1090)
+- [Er75f] Erdős, Paul, *On some problems of elementary and combinatorial geometry*. Ann. Mat. Pura Appl. (4) (1975), 99-108.
 -/
 
 namespace Erdos1090
 
 /--
 Let $k\geq 3$. Does there exist a finite set $A\subset \mathbb{R}^2$ such that, in any $2$-colouring of $A$, there exists a line which contains at least $k$ points from $A$, and all the points of $A$ on the line have the same colour?
+
+Erdős [Er75f] says Graham and Selfridge proved the answer is yes when $k=3$. Hunter has observed that, for sufficiently large $n$, a generic projection of $[k]^n$ into $\mathbb{R}^2$ has this property, by the Hales-Jewett theorem.
 -/
 @[category research solved, AMS 5, formal_proof using lean4 at "https://github.com/plby/lean-proofs/blob/main/src/v4.29.1/ErdosProblems/Erdos1090.lean"]
 theorem erdos_1090 : answer(True) ↔ ∀ (k : ℕ), ∀ (hk : 3 ≤ k),
--- a/FormalConjectures/ErdosProblems/1090.lean
+++ b/FormalConjectures/ErdosProblems/1090.lean
@@ -25,13 +25,10 @@ import FormalConjectures.Util.ProblemImports
 namespace Erdos1090
 
 /--
-For every $k \ge 3$ there is a finite set $A \subseteq \mathbb{R}^2$ such that in every
-2-colouring of $A$ there is a line $S$ containing at least $k$ points of $A$, such that
-every point of $A$ on that line lies in $S$ and all of them have the same colour.
-The answer is yes (Hunter), via the Hales–Jewett theorem.
+Let $k\geq 3$. Does there exist a finite set $A\subset \mathbb{R}^2$ such that, in any $2$-colouring of $A$, there exists a line which contains at least $k$ points from $A$, and all the points of $A$ on the line have the same colour?
 -/
 @[category research solved, AMS 5, formal_proof using lean4 at "https://github.com/plby/lean-proofs/blob/main/src/v4.29.1/ErdosProblems/Erdos1090.lean"]
-theorem erdos_1090 (k : ℕ) (hk : 3 ≤ k) :
+theorem erdos_1090 : answer(True) ↔ ∀ (k : ℕ), ∀ (hk : 3 ≤ k),
     ∃ (A : Finset (Fin 2 → ℝ)), ∀ (C : A → Fin 2),
       ∃ (S : Finset (Fin 2 → ℝ)), ∃ (hSA : S ⊆ A),
         Collinear ℝ (S : Set (Fin 2 → ℝ)) ∧ S.card ≥ k ∧
--- a/FormalConjectures/ErdosProblems/1090.lean
+++ b/FormalConjectures/ErdosProblems/1090.lean
@@ -1,0 +1,42 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 1090
+
+*Reference:* [erdosproblems.com/1090](https://www.erdosproblems.com/1090)
+-/
+
+namespace Erdos1090
+
+/--
+For every $k \ge 3$ there is a finite set $A \subseteq \mathbb{R}^2$ such that in every
+2-colouring of $A$ there is a line $S$ containing at least $k$ points of $A$, such that
+every point of $A$ on that line lies in $S$ and all of them have the same colour.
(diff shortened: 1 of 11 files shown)


mo271 (commented):
Thanks that is great to see, @willblair0708!

A few general remarks (see also `AGENTS.md` and `ErdosProblems/README.md`): 
 - The style of the repo is not to duplicate the informal statement in the module docstring and then again in the docstring of the theorem: we only want to have the docstring once, namely in the docstring of the theorem. 
 - The docstring should use latex markdown and not backticks, generally. 
 - We typically don't record in the files who formalised it using what, but we still want to make this information available: please put in in the PR description, so it is visible there and also in the commit message.  Since this pr atypically adds more than one problems we, should in the pr description include a block like : 
```
1090: "This was formalized in Lean by JoshuaB using Aristotle and Gemini 3.0 Flash." 
1125: "This was formalized in Lean by Stefano Rocca using Aristotle."
```
and also mention that the files come from @plby's repo. 

- I will do a review of the mathematical content later today. If it turns out we need to fix some of the statements, I suppose the `formal_proof` links will go stale, so hopefully the statements are all good...

williamjblair:
Thanks @mo271! Added the attribution block and the @plby credit to the description. The single-docstring and LaTeX fixes I'll fold into the next push along with anything your review turns up

Edit: just made the formatting fixes

Smetalo:
Does this resolve #919 and #993?

williamjblair:
> Does this resolve #919 and #993?

Yes, should resolve both if all ok (729 is #919 and 871 is #993). Thanks for catching them!

FormalConjectures/ErdosProblems/871.lean
```diff
@@ -0,0 +1,47 @@
...
+      (∀ᶠ n in Filter.atTop, ∃ a ∈ A, ∃ b ∈ A, a + b = n) ∧
+      (∀ t, ∀ᶠ n in Filter.atTop, ∃ pairs : Finset (ℕ × ℕ),
+        pairs.card ≥ t ∧
+          ∀ p ∈ pairs, p.1 ∈ A ∧ p.2 ∈ A ∧ p.1 + p.2 = n ∧ p.1 ≤ p.2) ∧
+      ¬∃ (B C : Set ℕ),
+        (∀ x, x ∈ A ↔ x ∈ B ∨ x ∈ C) ∧
+        Disjoint B C ∧
+        (∀ᶠ n in Filter.atTop, ∃ a ∈ B, ∃ b ∈ B, a + b = n) ∧
+        (∀ᶠ n in Filter.atTop, ∃ a ∈ C, ∃ b ∈ C, a + b = n) := by
+  sorry
```
mo271, line 48:
```suggestion
/--
Let $A$ be an additive basis of order $2$, and suppose $1_A\ast 1_A(n)\to \infty$ as $n\to \infty$. Can $A$ be partitioned into two disjoint additive bases of order $2$?
-/
@[category research solved, AMS 11, formal_proof using lean4 at "https://github.com/plby/lean-proofs/blob/main/src/v4.29.1/ErdosProblems/Erdos871.lean"]
theorem erdos_871 :
    answer(False) ↔
      ∀ (A : Set ℕ),
        (∀ᶠ n in Filter.atTop, ∃ a ∈ A, ∃ b ∈ A, a + b = n) ∧
        (∀ t, ∀ᶠ n in Filter.atTop, ∃ pairs : Finset (ℕ × ℕ),
          pairs.card ≥ t ∧
            ∀ p ∈ pairs, p.1 ∈ A ∧ p.2 ∈ A ∧ p.1 + p.2 = n ∧ p.1 ≤ p.2) →
        ∃ (B C : Set ℕ),
          (∀ x, x ∈ A ↔ x ∈ B ∨ x ∈ C) ∧
          Disjoint B C ∧
          (∀ᶠ n in Filter.atTop, ∃ a ∈ B, ∃ b ∈ B, a + b = n) ∧
          (∀ᶠ n in Filter.atTop, ∃ a ∈ C, ∃ b ∈ C, a + b = n) := by
```

FormalConjectures/ErdosProblems/1090.lean
```diff
@@ -0,0 +1,42 @@
...
+
+*Reference:* [erdosproblems.com/1090](https://www.erdosproblems.com/1090)
+-/
+
+namespace Erdos1090
+
+/--
+For every $k \ge 3$ there is a finite set $A \subseteq \mathbb{R}^2$ such that in every
+2-colouring of $A$ there is a line $S$ containing at least $k$ points of $A$, such that
+every point of $A$ on that line lies in $S$ and all of them have the same colour.
```
mo271:
```suggestion
Let $k\geq 3$. Does there exist a finite set $A\subset \mathbb{R}^2$ such that, in any $2$-colouring of $A$, there exists a line which contains at least $k$ points from $A$, and all the points of $A$ on the line have the same colour?

```

FormalConjectures/ErdosProblems/1090.lean
```diff
@@ -0,0 +1,42 @@
...
+The answer is yes (Hunter), via the Hales–Jewett theorem.
+-/
+@[category research solved, AMS 5, formal_proof using lean4 at "https://github.com/plby/lean-proofs/blob/main/src/v4.29.1/ErdosProblems/Erdos1090.lean"]
+theorem erdos_1090 (k : ℕ) (hk : 3 ≤ k) :
+    ∃ (A : Finset (Fin 2 → ℝ)), ∀ (C : A → Fin 2),
```
mo271:
```suggestion
theorem erdos_1090 : answer(True) ↔ ∀ (k : ℕ), ∀ (hk : 3 ≤ k),
    ∃ (A : Finset (Fin 2 → ℝ)), ∀ (C : A → Fin 2),
```

mo271 (commented):
Thanks!

One more general comment that applie to all of those comments: 

Let's not rephrase the problem statements but rather cite verbatim from erdosproblems.com, unless the rephrasing fixes an error or inaccuray in the original formulation. 

The differences here seem to be that mainly on erdosproblems they are phrased as questions where as here we naturally state one direction. We prefer to keep the original question and then use the answer(sorry), where the sorry is filled in as True or False. I gave two examples how it should look like. (one for the first problem, one for the only one here that has a negative answer). Note that this includes taking some of the hypotheses of the theorem and instead move them begin a `∀` quantifier .

Mathematically all 11 of those formalisation look flawless! 

I anticipate that some other ones are more heavy on introducing definitions and then one has to be careful of unifying with the definitions we already have in `FormalConjecturesForMathlib`. 

As a follow-up and could promote the proofs from `formal_proof using lean4` to formal_proof using formal_conjectures` seeing if the proof can be made to compile with our lean version and answer(sorry) addition

williamjblair:
Great, thanks for the review! All 11 are now answer(True/False) ↔ ∀ … with the verbatim erdosproblems.com statement in each docstring. 871 and 1090 use your suggestions as written, while the other nine follow the same form. 

Will make sure to apply this convention to the rest, and reuse the FormalConjecturesForMathlib definitions where they exist, going forward

mo271 (commented):
Thanks! 
this looks pretty good to me now, almost ready to go. Further changes should be doc-string only: 
It would be nice to include the comments below the main problems that explain how the solution was obtained, i.e. typically copy verbatim the sentence explaining in what paper a solution was found/what the solution is (if it is there) and then add the reference to the references section in the file.

williamjblair:
Great, I added the solution provenance to all 11 so each docstring now quotes the erdosproblems.com sentence on how the problem was solved, with the cited papers in a `*References:*` list in the module docstring. Statements and `formal_proof` links are unchanged.

mo271 (approved):
Thanks for the patience, looks great now!

williamjblair:
> Thanks for the patience, looks great now!

Thanks for your quick work!

PR #4321, opened 2026-06-25
feat(ErdosProblems/686): prove variants.nine via explicit witness

williamjblair:
Closes the `erdos_686.variants.nine` sorry with an explicit witness.

Taking `k = 3`, `n = 11`, `m = 25` gives

```
(26·27·28) / (12·13·14) = 19656 / 2184 = 9,
```

with `m = 25 ≥ n + k = 14`. The proof supplies that witness, then `norm_num` discharges the rational identity after `Finset.prod_Icc_succ_top` unfolds the interval products.

The `@[category research solved, AMS 11]` attribute and the docstring stay as they were; only the proof body changes.

Verified against the pinned Mathlib: the file compiles, and `#print axioms erdos_686.variants.nine` returns `[propext, Classical.choice, Quot.sound]`. No `sorryAx`, no `native_decide`.

I used AI assistance (Claude) to draft the proof and verified it myself.


mo271 (approved):
Thanks, LGTM!

PR #4322, opened 2026-06-25
feat(Papers/DeGiorgi): De Giorgi's conjecture

jvanwinden:
This PR adds a conjecture of De Giorgi, stating that level sets of bounded solutions to $$\Delta u + u - u^3$$ for which $$\partial_1 u > 0$$ are necessarily hyperplanes, in spatial dimension $$n \leq 8$$.

Some details:
- Some authors assume $$\|u\| \leq 1$$ when stating the conjecture, others merely assume boundedness. I think that there is basically no difference between these due to the maximum principle, and I chose boundedness.
- I used `lineDeriv` to express the condition on the partial derivative. However, I am not sure if this is the most idiomatic way to state this.
- Most papers are not very explicit about the solution concept which is used. However, solutions should be smooth in any case, so I simply assumed $$C^2$$ from the start and formulated the conjecture for classical solutions.

### DIFF
--- a/FormalConjectures/Paper/DeGiorgi.lean
+++ b/FormalConjectures/Paper/DeGiorgi.lean
@@ -1,0 +1,95 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


jvanwinden:
It seems I am using the wrong email in my commit, which prevents the CLA check from working. I will reopen form a new branch.

PR #4323, opened 2026-06-25
feat(Papers/DeGiorgi): De Giorgi's conjecture

jvanwinden:
This PR adds a conjecture of De Giorgi, stating that level sets of bounded solutions to $$\Delta u + u - u^3$$ for which $$\partial_1 u > 0$$ are necessarily hyperplanes, in spatial dimension $$n \leq 8$$.

Some details:
- Some authors assume $$\|u\| \leq 1$$ when stating the conjecture, others merely assume boundedness. I think that there is basically no difference between these due to the maximum principle, and I chose boundedness.
- I used `lineDeriv` to express the condition on the partial derivative. However, I am not sure if this is the most idiomatic way to state this.
- Most papers are not very explicit about the solution concept which is used. However, solutions should be smooth in any case, so I simply assumed $$C^2$$ from the start and formulated the conjecture for classical solutions.


FormalConjectures/Paper/DeGiorgi.lean
```diff
@@ -0,0 +1,96 @@
...
+
+open ContDiff Set InnerProductSpace MeasureTheory EuclideanGeometry
+open scoped Laplacian
+
+namespace DeGiorgi
+
+variable {n : ℕ} [NeZero n]
+
+/--
+The function $u : ℝⁿ → ℝ$ is a bounded classical solution to $Δ u + u - u^3 = 0$.
```
mo271:
```suggestion
The function $u : ℝ^n → ℝ$ is a bounded classical solution to $Δ u + u - u^3 = 0$.
```

and similar in other places: not 100% sure, but I think otherwise it doesn't render nicely

FormalConjectures/Paper/DeGiorgi.lean
```diff
@@ -0,0 +1,96 @@
...
+$Δ u + u - u^3 = 0$ satisfying $∂₁u > 0$ everywhere, then the level sets of $u$ are hyperplanes.
+-/
+def DeGiorgi_conclusion (n : ℕ) [NeZero n] : Prop :=
+  ∀ u : ℝ^n → ℝ, (IsBoundedSolution u ∧ HasPositiveDeriv u) → HasHyperplaneLevelSets u
+
+/--
+De Giorgi's conjecture holds in dimension $n ≤ 8$.
+-/
+@[category research open, AMS 35]
+theorem DeGiorgi_le_eight (hn : n ≤ 8) : (DeGiorgi_conclusion n) := sorry
```
mo271: It would be great to give the more fine grained variants you mention above: 

like we should give the solved variant for $n=1$

```lean
/--
De Giorgi's conjecture vtrivially holds in dimension $n = 1$.
-/
@[category research solved, AMS 35]
theorem DeGiorgi_one : (DeGiorgi_conclusion 1) := by
  change (1) ∈ {s |_}
  delta IsBoundedSolution HasPositiveDeriv HasHyperplaneLevelSets and
  delta lineDeriv
  use fun and K V⟨A, B⟩=>B▸⟨affineSpan ℝ {A},Set.ext fun and=>⟨mem_affineSpan ℝ ∘(? _), by aesop⟩,by rw [direction_affineSpan,vectorSpan_singleton,finrank_bot]⟩
  replace B :StrictMono fun and:ℝ =>‹_ → ℝ› (and • EuclideanSpace.single (0 : Fin 01) (1: ℝ)) :=strictMono_of_deriv_pos fun and=>(? _)
  · use and.ext ∘ Fin.forall_fin_one.2.comp (B.injective<|.trans (congr_arg _ (eq_of_norm_sub_eq_zero (by norm_num[ EuclideanSpace.norm_eq])))<|·.trans (congr_arg _ (A.ext (by norm_num[ Fin.eq_zero]))))
  · exact (K.2 (and •.single 0 (↑1))).trans_eq ((symm ((add_zero and▸deriv_comp_const_add _ _ _).symm.trans (by. (simp_rw [add_smul])))))

```

(Either with a clean-up proof of `sorry`: this AlphaProof proof is too ugly/unmaintanable)

then list separate "solved" statemnet for $n=2$ and $n=3$ and then perhaps also even separate statement for each open case $4\le n \le 8$. 
This way smaller incremental progress can be made!

mo271:
```suggestion
@[category research open, AMS 35]
theorem DeGiorgi_le_eight (hn : n ≤ 8) : (DeGiorgi_conclusion n) := by
  sorry
```
(our convention..)

FormalConjectures/Paper/DeGiorgi.lean
```diff
@@ -0,0 +1,96 @@
...
+
+/--
+In dimension $n ≥ 9$, the conclusion of De Giorgi's conjecture does not hold.
+-/
+@[category research solved, AMS 35]
+theorem DeGiorgi_ge_nine (hn : n ≥ 9) : ¬(DeGiorgi_conclusion n) := sorry
```
mo271:
```suggestion
theorem DeGiorgi_ge_nine (hn : n ≥ 9) : ¬(DeGiorgi_conclusion n) := by
  sorry
```

mo271 (commented):
Thanks! 
This is a great contribution -- interesting problem. 

I left a few suggestion, but mathematically everything makes sense here

FormalConjectures/Paper/DeGiorgi.lean
jvanwinden: I added the cases $n =1...8$ as separate statements.

On further thought, perhaps the $n = 1$ case should be omitted completely since it is basically a degenerate case: the proof does not even use the PDE, but follows directly from the assumption that $u$ has positive derivative.

jvanwinden:
-awaiting-author

FormalConjectures/Paper/DeGiorgi.lean
mo271: I think either omitting or fixing the proof above are both good options here!

jvanwinden: In the end I decided to clean up the proof (or rather, rewrite it).

jvanwinden:
-awaiting-author

mo271 (approved):
Thanks, looks good to me now!

Fun problem!

PR #4329, opened 2026-06-26
feat(ErdosProblems/36): prove M 2 = 1

c0mpli:
Closes the test sorry `M_two` (`M 2 = 1`) with an explicit optimal partition witness. Verified clean under the project linters.


mo271 (changes requested):
same problem as in #4332, if the proof is longer than the one linked in `formal_proof at foo using bar` than we probably don't want to add it.

c0mpli:
You're right — same situation as #4332. `M` is noncomputable (no short `decide` route), and this proof doesn't improve on the existing `formal_proof`, so it shouldn't be added. Closing. Thanks for the careful review.

PR #4331, opened 2026-06-26
feat(ErdosProblems/456): prove m(n) ≤ p(n)

c0mpli:
Closes `erdos_456.variants.mn_leq_pn` via a case split plus Dirichlet. Verified clean under the project linters.


mo271 (approved):
Thanks, LGTM!

PR #4338, opened 2026-06-26
feat(ErdosProblems/221): add additive basis 2^k + a statement

williamjblair:
Adds Erdős 221, which FC didn't have a file for yet. The question: can a set `A ⊆ ℕ` of density `≪ N/log N` have every large integer in the form `2^k + a` with `a ∈ A`? Ruzsa proved it can, with an ingeniously simple construction; Lorentz had an earlier, weaker bound.

I kept the statement on Mathlib's `IsBigO` and `Set.ncard` so it introduces no new definitions, and checked it against the erdosproblems.com page rather than only that it compiles. Body is `sorry`; `formal_proof using lean4` points to the Lean proof.

Formalized in Lean by Wouter van Doorn using Aristotle, in his own repo (`Woett/Lean-files`), indexed via @plby's `sources.yaml`.

### DIFF
--- a/FormalConjectures/ErdosProblems/221.lean
+++ b/FormalConjectures/ErdosProblems/221.lean
@@ -1,0 +1,49 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Util.ProblemImports
+
+/-!
+# Erdős Problem 221
+
+*References:*
+- [erdosproblems.com/221](https://www.erdosproblems.com/221)
+- [Lo54] Lorentz, G. G., *On a problem of additive number theory*. Proc. Amer. Math. Soc. (1954), 838-841.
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/221.lean
```diff
@@ -0,0 +1,49 @@
...
+
+Lorentz [Lo54] proved there is such a set with, for all large $N$,
+\[\lvert A\cap\{1,\ldots,N\}\rvert \ll \frac{\log\log N}{\log N}N\]
+The answer is yes, proved by Ruzsa [Ru72].
+-/
+@[category research solved, AMS 11, formal_proof using lean4 at "https://github.com/Woett/Lean-files/blob/main/ErdosProblem221.lean"]
+theorem erdos_221 :
+    answer(True) ↔ ∃ A : Set ℕ,
+      (fun N => ({a ∈ A | a ≤ N}.ncard : ℝ)) =O[atTop] (fun N => (N : ℝ) / Real.log N) ∧
+      ∀ᶠ N in atTop, ∃ k a, 0 ≤ k ∧ a ∈ A ∧ N = 2 ^ k + a := by
```
mo271, line 46:
```suggestion
    answer(True) ↔ ∃ A : Set ℕ,
      (fun N => ({a ∈ A | a ≤ N}.ncard : ℝ)) ≪ (fun N => (N : ℝ) / Real.log N) ∧
      ∀ᶠ N in atTop, ∃ k a, 0 ≤ k ∧ a ∈ A ∧ N = 2 ^ k + a := by
```

We support this notation in our `FormalConjecturesForMathlib`

mo271 (approved):
Thanks, looks great, just one nit on using `≪` (should be mathematical equivalent...)

williamjblair:
Done, thanks @mo271. Switched to ≪ — it needed parentheses around the clause since ≪ binds looser than ∧, otherwise compiles clean.

mo271:
Thanks again!

PR #4339, opened 2026-06-26
feat(ErdosProblems/290): add harmonic-sum denominator statement

williamjblair:
Adds Erdős 290 (no FC file yet). For `a ≥ 1`, must there be some `b > a` where extending `∑_{a≤n≤b} 1/n` by the next term shrinks the reduced denominator? van Doorn proved yes, and in fact `b(a) ≪ a`.

The file needs one definition, `harmonicDen` over `Rat.den`, since Mathlib has nothing for the denominator of a partial harmonic sum. Body is `sorry`; `formal_proof using lean4` points to the Lean proof.

Formalized in Lean by Wouter van Doorn and Boris Alexeev using Aristotle (`Woett/Lean-files`), indexed via @plby's `sources.yaml`.

### DIFF
--- a/FormalConjectures/ErdosProblems/290.lean
+++ b/FormalConjectures/ErdosProblems/290.lean
@@ -1,0 +1,52 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4340, opened 2026-06-26
feat(ErdosProblems/154): add Sidon sumset equidistribution statement

williamjblair:
Adds Erdős 154 in the statement-oriented shape requested by the repo conventions.

- `erdos_154` is the boxed $A + A$ question, normalized by `|A + A|`, so the limiting proportion is `1 / m`.
- `erdos_154` links a hosted Lean proof of the full sumset statement: `Erdos154.erdos_154_sumset`.
- `erdos_154.variants.lindstrom` records Lindström's distribution theorem for $A$ itself, matching the linked Woett proof.
- The Lindström-to-sumset transfer proof is hosted externally rather than included in FC.

Attribution: Lindström theorem formalized by Wouter van Doorn using Aristotle (`Woett/Lean-files`), indexed via @plby. The hosted proof adds the transfer from Lindström's theorem to the $A + A$ statement. I opened plby/lean-proofs#4 to move it into the canonical proof index repo.

Hosted proof validation:
- `cd src/v4.29.1 && lake exe cache get && lake build ErdosProblems.Erdos154Sumset`
- `Erdos154.sidon_density_limit`: `[propext, Classical.choice, Quot.sound]`
- `Erdos154.erdos_154_sumset`: `[propext, Classical.choice, Quot.sound]`

FC validation:
- `lake env lean FormalConjectures/ErdosProblems/154.lean`
- `lake --wfail build 'FormalConjectures.ErdosProblems.«154»'`
- `lake --wfail build`
- `git diff --check`


FormalConjectures/ErdosProblems/154.lean
```diff
@@ -0,0 +1,61 @@
...
+@[category research solved, AMS 5 11, formal_proof using lean4 at "https://github.com/Woett/Lean-files/blob/main/ErdosProblem154.lean"]
+theorem erdos_154 : answer(True) ↔
+    ∀ (m : ℕ) (hm : 2 ≤ m) (N : ℕ → ℕ) (A : ℕ → Finset ℕ),
+      Tendsto (fun k => (N k : ℝ)) atTop atTop →
+      (∀ k, ∀ x ∈ A k, x ≤ N k) →
+      (∀ k, IsSidon (A k : Set ℕ)) →
+      Tendsto (fun k => ((A k).card : ℝ) / Real.sqrt (N k)) atTop (nhds 1) →
+      ∀ i < m, Tendsto (fun k => (((A k).filter (fun a => a % m = i)).card : ℝ) / Real.sqrt (N k))
+        atTop (nhds (1 / m)) := by
+  sorry
```
mo271, line 84: * **The Original Question (from `erdosproblems.com/154`):**
  > Let $A\subset \{1,\ldots,N\}$ be a Sidon set with $\lvert A\rvert\sim N^{1/2}$. Must **$A+A$** be well-distributed over all small moduli? In particular, must about half the elements of **$A+A$** be even and half odd?

  The question is about the **sumset** $A+A = \{a_1 + a_2 : a_1, a_2 \in A\}$, not about $A$ itself.

* **What Is Formalized:** The theorem `erdos_154` formalizes the distribution of **$A$ itself** over residue classes:
  ```lean
  ∀ i < m, Tendsto (fun k => (((A k).filter (fun a => a % m = i)).card : ℝ) / Real.sqrt (N k))
    atTop (nhds (1 / m))
  ```
  This filters elements of $A$ (not $A+A$) by their residue modulo $m$.

* **The Docstring Acknowledges This:** The author writes: *"We formalise Lindström's result for $A$ itself"* and argues that well-distribution of $A+A$ follows "immediately using the Sidon property." While this is mathematically reasonable (if $A$ is well-distributed mod $m$ and Sidon, then $A+A$ is also well-distributed mod $m$), the theorem named `erdos_154` should formalize the actual Erdős question about $A+A$, not a weaker auxiliary result.

* **Recommendation:** Either:
  1. **Rename** the current theorem to `erdos_154.variants.lindstrom` and add a new `erdos_154` that formalizes the $A+A$ distribution question directly, or
  2. **Refactor** the theorem to state the distribution of $A+A$ over residue classes (which is the actual Erdős question), keeping the Lindström result as a variant.

  The $A+A$ version would look something like:
  ```lean
  theorem erdos_154 : answer(True) ↔
      ∀ (m : ℕ) (hm : 2 ≤ m) (N : ℕ → ℕ) (A : ℕ → Finset ℕ),
        Tendsto (fun k => (N k : ℝ)) atTop atTop →
        (∀ k, ∀ x ∈ A k, x ≤ N k) →
        (∀ k, IsSidon (A k : Set ℕ)) →
        Tendsto (fun k => ((A k).card : ℝ) / Real.sqrt (N k)) atTop (nhds 1) →
        ∀ i < m, Tendsto (fun k =>
          (((A k + A k).filter (fun s => s % m = i)).card : ℝ) / (N k : ℝ))
          atTop (nhds (1 / m)) := by
    sorry
  ```
  (Note: the normalization changes from $\sqrt{N}$ to $N$ because $|A+A| \sim N$ for a Sidon set of size $\sim \sqrt{N}$.)

A good sanity check would be to actually prove that the lindström variant implies erdos_154 then.

mo271 (commented):
Thanks, looks great in general!
Usually we stick to verbatim statements and record more than one variant -- of course this makes more sense for open problems, because one does not know which variant is weaker and hence doesn't know which one should prove... 
Hence a detailied suggestion, but If this is too much effort I'm also happy to accept as is in this case..

williamjblair:
Good catch, thanks @mo271. Went with option 1: `erdos_154` is now the $A+A$ sumset question, and the previous $A$-distribution result is `erdos_154.variants.lindstrom` (what the linked Woett proof actually establishes).

One thing on the A+A statement: I normalized by $\lvert A+A\rvert$ rather than $N$. For a Sidon set of size $\sim \sqrt{N}$ the sumset has $\lvert A+A\rvert = \binom{\lvert A\rvert+1}{2} \sim N/2$, so dividing by $N$ would land on $1/(2m)$ (e.g. the even/odd case gives $1/4$, not the $1/2$ the problem asks for). Dividing by $\lvert A+A\rvert$ gives $1/m$ and avoids pinning the constant. Should be easy to switch to $/N$ with a $1/(2m)$ limit if you'd prefer.

I left `lindstrom → erdos_154` as a follow-up rather than rush it; the $A+A$-from-$A$ transfer via the Sidon property is a real proof. Will attempt this

williamjblair:
Update: pushed `91b85947`.

This keeps the FC file statement-sized, but `erdos_154` now has a `formal_proof` link to a hosted proof of the full A+A statement:
https://github.com/willblair0708/formal-conjectures/blob/f650ef01dc263a5628df936f2df4322f11aa5119/Proofs/Erdos154/ErdosProblems/Erdos154Sumset.lean

The hosted theorem is `Erdos154.erdos_154_sumset`. It combines the clean Lindström proof for `A` with the transfer to `A + A` outside this repo.

Axiom audit in the hosted project:
- `Erdos154.sidon_density_limit`: `[propext, Classical.choice, Quot.sound]`
- `Erdos154.erdos_154_sumset`: `[propext, Classical.choice, Quot.sound]`

Local checks passed: `lake env lean FormalConjectures/ErdosProblems/154.lean`, targeted `lake --wfail build`, full `lake --wfail build`, and `git diff --check`.

williamjblair:
Small follow-up: the hosted proof now lives in a dedicated repo, `willblair0708/lean-proofs`, linked at `main`:
https://github.com/willblair0708/lean-proofs/blob/main/ErdosProblems/Erdos154Sumset.lean

It is a standalone Lean project where CI runs `lake build` plus a `#print axioms` audit on every push, so the proof is continuously checked to be sorry-free with axioms `[propext, Classical.choice, Quot.sound]`. Same external-repo pattern as the Woett-hosted proofs FC already links.

The linked proof carries the Sidon argument through (the self-convolution of an equidistributed residue profile stays equidistributed, and Sidon injectivity turns the ordered-pair count into the sumset count), sorry-free with clean #print axioms.

One subtlety worth flagging: the sumset count is normalized by (A k + A k).card, not N. For a Sidon set with |A| ~ √N we have |A+A| ~ N/2, so dividing by N would send each class to 1/(2m); dividing by |A+A| gives the clean 1/m.

mo271 (approved):
Thanks, LGTM!

PR #4346, opened 2026-06-27
ErdosProblems: add 281, 419, 453, 476, 519, 540 (#3998 sync)

### LINKED ISSUE #470: Erdős Problem 281
### What is the conjecture
    
https://www.erdosproblems.com/281

  Let $n_1<n_2<\cdots$ be an infinite sequence such that, for any choice of congruence classes $a_i\pmod{n_i}$, the set of integers not satisfying any of the congruences $a_i\pmod{n_i}$ has density $0$. Is it true that for every $\epsilon>0$ there exists some $k$ such that, for every choice of congruence classes $a_i$, the density of integers not satisfying any of the congruences $a_i\pmod{n_i}$ for $1\leq i\leq k$ is less than $\epsilon$?

Status: open

### Choose either option

williamjblair:
fixes https://github.com/google-deepmind/formal-conjectures/issues/470

## Summary

Adds Formal Conjectures statement files for Erdős Problems 281, 419, 453, 476, 519, and 540: solved problems with a hosted Lean 4 proof but no FC statement yet. Each states the main `theorem erdos_n` from the boxed erdosproblems.com question and links the hosted proof.

Part of #3998.

## Verification

For each I confirmed the linked proof proves the stated theorem with the same quantifiers and no extra hypotheses:
- **281**: the `(∀ a, density 0) →` premise is the problem's own hypothesis and appears in the statement; the hosted proof discharges exactly `Hyp → Concl`.
- **419**: set equality `{x | MapClusterPt x atTop …} = …`, matching the hosted theorem directly.
- **453**: `answer(False)`, proven by the hosted negation.
- **476**: `∀ p prime, ∀ A, …`, matching the hosted `(p)[Fact p.Prime](A)` form.
- **519 / 540**: `∃ c > 0, ∀ …`, witnessed by the hosted constant.

`#print axioms` on each hosted proof reports only `[propext, Classical.choice, Quot.sound]`.

## Attribution

The external formal proofs are not mine; attribution follows the upstream proof headers, via @plby's hosted Lean proof index.

### DIFF
--- a/FormalConjectures/ErdosProblems/281.lean
+++ b/FormalConjectures/ErdosProblems/281.lean
@@ -41,16 +41,6 @@ def avoidAll (n : ℕ → ℕ) (a : ResidueChoice n) : Set ℤ :=
 def avoidPrefix (n : ℕ → ℕ) (a : ResidueChoice n) (k : ℕ) : Set ℤ :=
   {m | ∀ i : ℕ, i < k → (m : ZMod (n i)) ≠ a i}
 
-open Classical in
-/-- Two-sided natural density sequence for subsets of $\mathbb{Z}$. -/
-noncomputable def integerDensitySeq (S : Set ℤ) (N : ℕ) : ℝ :=
-  (((Finset.Icc (-(N : ℤ)) (N : ℤ)).filter (fun m => m ∈ S)).card : ℝ) /
-    (2 * (N : ℝ) + 1)
-
-/-- A subset of $\mathbb{Z}$ has two-sided natural density $d$. -/
-def HasIntegerDensity (S : Set ℤ) (d : ℝ) : Prop :=
-  Tendsto (integerDensitySeq S) atTop (nhds d)
-
 /--
 Let $n_1<n_2<\cdots$ be an infinite sequence such that, for any choice of congruence classes
 $a_i\pmod{n_i}$, the set of integers not satisfying any of the congruences $a_i\pmod{n_i}$ has
@@ -63,10 +53,10 @@ The answer is yes; the linked Lean proof formalizes Somani's argument.
 @[category research solved, AMS 11, formal_proof using lean4 at "https://github.com/plby/lean-proofs/blob/main/src/v4.29.1/ErdosProblems/Erdos281.lean"]
 theorem erdos_281 : answer(True) ↔
     ∀ (n : ℕ → ℕ), StrictMono n → (∀ i, 0 < n i) →
-      (∀ a : ResidueChoice n, HasIntegerDensity (avoidAll n a) 0) →
+      (∀ a : ResidueChoice n, Set.HasIntDensity (avoidAll n a) 0) →
         ∀ ε : ℝ, 0 < ε →
           ∃ k : ℕ, ∀ a : ResidueChoice n,
-            ∃ d : ℝ, HasIntegerDensity (avoidPrefix n a k) d ∧ d < ε := by
+            ∃ d : ℝ, Set.HasIntDensity (avoidPrefix n a k) d ∧ d < ε := by
   sorry
 
 end Erdos281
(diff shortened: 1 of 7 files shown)


Smetalo:
Does this resolve #470?

FormalConjectures/ErdosProblems/419.lean
```diff
@@ -0,0 +1,56 @@
...
+/-!
+# Erdős Problem 419
+
+*References:*
+- [erdosproblems.com/419](https://www.erdosproblems.com/419)
+- [EGIP96] Erdős, Paul and Graham, S. W. and Ivić, Aleksandar and Pomerance, Carl,
+  *On the number of divisors of $n!$*. (1996), 337--355.
+-/
+
+open Filter
```
mo271, line 30:
```suggestion
open Filter
open scoped ArithmeticFunction.sigma
```

FormalConjectures/ErdosProblems/419.lean
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Erdos419
+
+/-- The divisor-counting function $\tau(n)$. -/
+noncomputable def tau (n : ℕ) : ℕ := n.divisors.card
+
+/-- The ratio $\tau((n+1)!)/\tau(n!)$. -/
+noncomputable def factorialDivisorRatio (n : ℕ) : ℝ :=
+  (tau (n + 1).factorial : ℝ) / (tau n.factorial : ℝ)
```
mo271:
```suggestion
  (σ 0 (n + 1).factorial : ℝ) / (σ 0 n.factorial : ℝ)
```

or intro notation for it

FormalConjectures/ErdosProblems/419.lean
```diff
@@ -0,0 +1,56 @@
...
+  *On the number of divisors of $n!$*. (1996), 337--355.
+-/
+
+open Filter
+
+namespace Erdos419
+
+/-- The divisor-counting function $\tau(n)$. -/
+noncomputable def tau (n : ℕ) : ℕ := n.divisors.card
+
```
mo271:
```suggestion
```

FormalConjectures/ErdosProblems/281.lean
```diff
@@ -0,0 +1,72 @@
...
+  {m | ∀ i : ℕ, i < k → (m : ZMod (n i)) ≠ a i}
+
+open Classical in
+/-- Two-sided natural density sequence for subsets of $\mathbb{Z}$. -/
+noncomputable def integerDensitySeq (S : Set ℤ) (N : ℕ) : ℝ :=
+  (((Finset.Icc (-(N : ℤ)) (N : ℤ)).filter (fun m => m ∈ S)).card : ℝ) /
+    (2 * (N : ℝ) + 1)
+
+/-- A subset of $\mathbb{Z}$ has two-sided natural density $d$. -/
+def HasIntegerDensity (S : Set ℤ) (d : ℝ) : Prop :=
```
mo271: funny, I think our defintoin of `density` in FormalConjecturesForMathlib indeed doesn't cover the integer case. Perhaps it could be expanded to actually do?

mo271 (commented):
Thanks, looks good, just a few comments.

FormalConjectures/ErdosProblems/281.lean
williamjblair: Yep agreed, I think this would be useful going forward

williamjblair:
Thanks, addressed. 419 now uses Mathlib's `σ 0` for the divisor-counting function and drops the local `tau`. For 281 I moved the two-sided integer-density notion into `FormalConjecturesForMathlib.Data.Set.Density` as shared API and updated the statement to use it.

mo271 (approved):
Thanks, LGTM

PR #4355, opened 2026-06-29
chore(Wikipedia/EllipticCurveRank): don't assume `AddCommMonoid E⟮K⟯`

YaelDillies:
Assuming this is nonsense. One should instead use the existing typeclass provided by mathlib.


mo271 (approved):
Thanks, LGTM

PR #4359, opened 2026-06-29
Fix Erdős 686 four_three docstring

williamjblair:
## Summary

Fixes a stale docstring on `erdos_686.variants.four_three`: the statement is the `k = 3`, `m ≥ n + 3` case, but the prose still described the `k = 2` variant.

I also removed the earlier proof-experiment commits from this branch so this PR is just the docstring cleanup.

## Checks

```bash
lake env lean FormalConjectures/ErdosProblems/686.lean
git diff --check
```

The Lean check reports only the existing `sorry` warnings for statement bodies.


mo271 (approved):
Thanks, LGTM

PR #2276, opened 2026-02-12
feat: add Noncototient conjecture and related theorem

Brian-Nugent:
This is issue #2249

PR #4361, opened 2026-07-01
feat(ErdosProblems/418): state the Odd Noncototient Conjecture

### LINKED ISSUE #2249: Odd Noncototient Conjecture
### What is the conjecture

 A **noncototient** is a positive integer $k$ for which the equation $m - \phi(m) = k$ has no solution, where $\phi(m)$ denotes Euler's totient function.

The **Odd Noncototient Conjecture** states that all noncototients are even. Equivalently, every odd positive integer $k > 5$ can be expressed as $m - \phi(m)$ for some positive integer $m$.

This is supported by the observation that for odd primes $p$ and $q$ with $p + q$ even (i.e., one of them equals 2), we have $pq - \phi(pq) = pq - (p-1)(q-1) = p + q - 1$, which is odd. The conjecture is closely related to the Goldbach conjecture: if the strong Goldbach conjecture is true (every even integer $\geq 4$ is a sum of two primes), then every odd number $> 5$ can be written as $p + q - 1$ for primes $p, q$, thereby proving no odd noncototient exists.

 **Sources:**
 - https://en.wikipedia.org/wiki/Noncototient | https://oeis.org/A005278 | https://mathworld.wolfram.com/Noncototient.html | https://arxiv.org/pdf/math/0409231 | https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Nat/Totient.html

### Prerequisites needed

**Formalizability Rating:** 1/5 (0 is best) (as of 2026-02-10)

Building blocks (2; from search results):
- `Nat.totient n` (Mathlib.Data.Nat.Totient): Euler's totient function defined as the count of integers $< n$ coprime to $n$
- `Nat.Coprime a b` (Mathlib): standard definition of coprimality

Missing pieces (2; unclear/absent from search results):
- Definition of "noncototient" as a predicate on natural numbers (not in Mathlib)
- Helper lemmas relating totients of prime products to sums of primes (partial results likely exist, but full formalization would require new work)

Rating justification (1-2 sentences): The totient function is well-established in Mathlib, so the statement can be written immediately as a predicate or existential. However, one minor helper definition (the noncototient predicate itself) is needed, making this a rating of 1 rather than 0.

* ams-11

### Choose either option

---

Sanexxxx777:
## Summary

Closes #2249.

Adds the **Odd Noncototient Conjecture** — every non-cototient is even, equivalently every odd natural number is of the form $n - \phi(n)$ — as a variant of Erdős 418.

I stated it inside `ErdosProblems/418.lean` rather than as a new file because 418 already defines the non-cototient set `{ n - n.totient | n }ᶜ` and already contains `erdos_418.variants.conditional`, which derives the odd case from a strengthening of Goldbach. The new `erdos_418.variants.odd_noncototient` is precisely the **unconditional** form of that conditional variant, so keeping them together (and reusing the existing definition) follows the "state a problem in one place" convention. Happy to move it to `Wikipedia/` with a pointer if you'd prefer, since #2249 is tagged `wikipedia`.

## Changes

- `erdos_418.variants.odd_noncototient` (`category research open`): `{ (n - n.totient : ℕ) | n }ᶜ ⊆ { k | Even k }`.
- `erdos_418.variants.seven_mem_cototient` (`category test`): a proven sanity check that `7 = 15 - φ(15)` is a cototient (`⟨15, by decide⟩`).
- Added a Wikipedia: Noncototient reference to the module docstring.

## Verification

`lake --wfail build FormalConjectures.ErdosProblems.«418»` succeeds. The only `sorry`s are the research statements (as expected for benchmark statements); the `test` lemma is fully proved.


Smetalo:
Should these changes also be added to ‎_FormalConjectures/Wikipedia/Noncototient.lean_, just like it was done in #2276?

Sanexxxx777:
Good question — I looked into it. It turns out #2276 was closed without being merged (the CLA was never signed), so `FormalConjectures/Wikipedia/Noncototient.lean` doesn't actually exist on `main` — there's nothing to mirror. Since Erdős 418 *is* the noncototient problem, this statement lives naturally right next to the existing conditional (Goldbach-based) variant, so I've kept a single copy here rather than restating the same open conjecture in two files; this also addresses the still-open #2249. That said, if the maintainers would prefer the canonical home to be `FormalConjectures/Wikipedia/Noncototient.lean`, I'm happy to relocate it there instead of duplicating — just let me know.

Paul-Lez (approved):

PR #4153, opened 2026-05-31
feat(ErdosProblems/36): prove `M_two`

ChakshuGupta13:
Closes #4152. **Stacked on #4151** (M_one).

Proves `M_two : M 2 = 1` for the Erdős minimum-overlap problem at `n = 2`.

**Proof.** Reuses `one_le_overlap` introduced by #4151; adds two new helpers:
- `overlap_le_card_mul`: `Overlap A B k ≤ |A| * |B|` (card_filter_le + card_product).
- `maxOverlap_one_four_two_three`: the witness `A = {1, 4}, B = {2, 3}` has all four pairwise differences (`±1, ±2`) distinct, so `MaxOverlap = 1`.

Main bound: upper via `Nat.sInf_le` on the witness; lower via cardinality (`|A| + |B| = 4 ∧ |A| = |B|` forces both = 2, both nonempty), then `one_le_overlap` for any `(a, b) ∈ A × B`.

**Verification.**
- `lake build FormalConjectures.ErdosProblems.«36»` succeeds.
- `#print axioms Erdos36.M_two` → `[propext, Classical.choice, Quot.sound]`.

**Note.** Stacked on #4151 — please merge that first. Once #4151 lands I'll rebase this branch to upstream/main and it will be a clean delta.

PR #4151, opened 2026-05-31
feat(ErdosProblems/36): prove `M_one`

ChakshuGupta13:
Closes #4150.

Proves `M_one : M 1 = 1` for the Erdős minimum-overlap problem.

**Proof.** Two private API helpers:
- `one_le_overlap`: every `(a, b) ∈ A × B` witnesses `1 ≤ Overlap A B (a − b)`.
- `maxOverlap_singleton_one_two`: the only nonzero overlap between `{1}` and `{2}` is at `k = −1`, giving value 1.

Main bound: upper via `Nat.sInf_le` on the witness `A = {1}, B = {2}`. Lower via `le_csInf`: `|A| + |B| = 2 ∧ |A| = |B|` forces both singletons, then `Overlap {a} {b} (a − b) ≥ 1` lifts the iSup.

**Verification.**
- `lake build FormalConjectures.ErdosProblems.«36»` succeeds.
- `#print axioms Erdos36.M_one` → `[propext, Classical.choice, Quot.sound]`.

PR #4362, opened 2026-07-01
feat(ErdosProblems/36): prove M_three (M 3 = 2)

Sanexxxx777:
## Summary

Proves `M_three : M 3 = 2` for the Erdős minimum-overlap problem (Erdős 36) at `n = 3`. This is the first test value with `M(n) > 1` — `M_one` (#4151) and `M_two` (#4153) are both `1`, where the lower bound is trivial; at `n = 3` the lower bound requires actually ruling out `MaxOverlap ≤ 1`.

## Proof

- **Upper bound** (`M 3 ≤ 2`): the balanced partition `A = {1, 4, 5}`, `B = {2, 3, 6}` of `{1, …, 6}` has `MaxOverlap = 2` (`maxOverlap_145_236`). The `≤ 2` direction bounds `Overlap` at every `k` by splitting on whether `k` lies in the finite range of realizable differences `[-5, 3]` (`fin_cases … <;> decide`) or outside it (`Overlap = 0`).
- **Lower bound** (`M 3 ≥ 2`): every balanced partition has `MaxOverlap ≥ 2`. Since `A ∪ B = Icc 1 6`, `Disjoint A B`, `|A| = |B|` force `A ∈ (Icc 1 6).powersetCard 3` and `B = Icc 1 6 \ A`, I enumerate all `C(6,3)` partitions with `fin_cases` and discharge each by exhibiting a difference achieved twice (`two_le_maxOverlap_of_exists (by decide)`).

Adds three small API lemmas: `overlap_le_card_mul`, `two_le_maxOverlap_of_overlap`, `two_le_maxOverlap_of_exists`. (`overlap_le_card_mul` mirrors the one used behind the `M_two` `formal_proof` link, which is not present on `main`.)

The same machinery generalizes to `M_four = 2` and `M_five = 3` (with `powersetCard 4` / `powersetCard 5` enumerations); I kept this PR to `M_three` alone to match the incremental cadence of #4151 / #4153 and avoid a heavy `decide` over the larger partition sets. Happy to follow up with `M_four`/`M_five` if you'd like them.

## Verification

- `lake --wfail build FormalConjectures.ErdosProblems.«36»` succeeds (no linter warnings).
- `#print axioms Erdos36.M_three` → `[propext, Classical.choice, Quot.sound]`.

### DIFF
--- a/FormalConjectures/ErdosProblems/36.lean
+++ b/FormalConjectures/ErdosProblems/36.lean
@@ -121,9 +121,74 @@ Any balanced partition has both pieces nonempty, so `MaxOverlap \geq 1`.
 theorem M_two : M 2 = 1 := by
   sorry
 
+/-- A simple upper bound: `Overlap A B k ≤ |A| * |B|`. -/
+@[category API, AMS 5 11]
+private lemma overlap_le_card_mul {A B : Finset ℤ} (k : ℤ) :
+    Overlap A B k ≤ A.card * B.card := by
+  refine (Finset.card_filter_le _ _).trans ?_
+  rw [Finset.product_eq_sprod, Finset.card_product]
+
+/-- If some difference `k` is achieved at least twice, then `MaxOverlap` is at least `2`. -/
+@[category API, AMS 5 11]
+private lemma two_le_maxOverlap_of_overlap {A B : Finset ℤ} {k : ℤ}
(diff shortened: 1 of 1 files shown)


mo271:
in the meantime proofs of those have been added.

Thanks nonetheless for making the pr!

PR #4368, opened 2026-07-01
Add a `conditional` modifier and `assuming` clause to `formal_proof`

williamjblair:
Some solved problems only have formal proofs under an unproven assumption like GRH. This lets the repo record that assumption as a Lean statement instead of leaving it implicit in a declaration name, so the site can tell conditional formal proofs apart from unconditional ones and show what exactly is assumed.

For context on where this comes from: I audit hosted Lean proofs of Erdős problems at https://williamjblair.github.io/erdos-frontier/, picking up where the erdosproblems wiki's AI-contributions page left off when it was frozen on June 30, and the same audit feeds my #3998 statement batches here. It has so far turned up nine problems whose claimed formal proofs are conditional on unproven hypotheses, four of which `#print axioms` alone would not catch because the assumption rides as a theorem parameter. This attribute gives that information a home in the repo itself.

Following the review discussion, there is no separate attribute: `conditional` is a modifier on `formal_proof`, and the assumed hypotheses are stated in the same file (with `sorry` proofs) and referenced by name:

```lean
/-- The unproven hypothesis assumed by the conditional proof below (statement only). -/
@[category research open]
theorem grh : ... := by
  sorry

@[category research solved,
  conditional formal_proof using lean4 at "https://example.com/proof" assuming grh]
theorem a_conditionally_proved_problem : ... := by
  sorry
```

Behaviour worth knowing:

- names in `assuming` are resolved against the environment, so an unknown name is an error and hovering one jumps to the statement
- `conditional` without `assuming` is an error, as is `assuming` without `conditional`
- if a referenced hypothesis later gains a sorry-free proof, the attribute warns that the proof may no longer need to be `conditional`
- `extract_names` emits the declaration names under `proofConditions` (and supports `--exclude=proofConditions`); the theorem page shows a `Conditional` badge and an "Assumes" line

I left the existing Artin conditional declarations unannotated for now, since they don't carry a `formal_proof` link yet.

Tested with `lake build FormalConjectures.Util.Attributes.Basic`, a scratch file (not committed) covering the one- and two-hypothesis forms plus the three error paths under `#guard_msgs`, and `lake exe extract_names` on that file. Also rebuilt `Wikipedia.ArtinPrimitiveRootsConjecture`.


FormalConjectures/Util/Attributes/Basic.lean
```diff
@@ -83,6 +83,30 @@ theorem a_test_to_sanity_check_some_definition : ¬ FermatLastTheoremWith 1 := b
   sorry
 ```
 
+## The Proof Condition Attribute:
```
Paul-Lez: Hey! I think this is a nice idea! How about instead modifying the `formal_proof` attribute to have an optional `conditional` keyword in front - this would allow us to write e.g. `conditional formal_proof ...`.

mo271: in some sense it would also be nice to store exactly what the hypothesis used is, since that should be relatively short, perhaps we could add those statements to the file (with a sorry proof of course) and then add the declaration name to the attribute?!

williamjblair: Thanks @Paul-Lez , agreed! `conditional` is now a modifier on `formal_proof` itself (with no separate attribute), so it reads `@[conditional formal_proof using lean4 at "<link>" assuming grh]`. `conditional` without `assuming` errors, as does `assuming` without `conditional`.

williamjblair: @mo271 makes sense to me! `assuming` takes declaration names, so the hypothesis lives in the file as a statement with a `sorry` proof and the attribute references it. The idents resolve against the environment (an unknown name errors, hover jumps to the statement), and `extract_names` would emit the fully qualified names. This is useful as well bc if the hypothesis later gains a sorry-free proof, the attribute warns that the formal proof may no longer need to be `conditional`.

williamjblair:
@Paul-Lez @mo271 reworked along the lines you both suggested: `conditional` is now a modifier on `formal_proof`, and the assumed hypotheses are declarations in the file, referenced from an `assuming` clause. Details in the updated description.

Some context on the motivation: the erdosproblems wiki froze its AI-contributions page on June 30, and I have been continuing that tracking at https://williamjblair.github.io/erdos-frontier/ while auditing the hosted Lean proofs I sync for #3998. That audit has so far turned up nine claimed formal proofs that are conditional on unproven hypotheses, four of them invisible to `#print axioms` because the assumption rides as a theorem parameter. This attribute is meant to give those conditions a first-class home in the repo

Set to active for review now as well

FormalConjecturesUtil/Attributes/Basic.lean
```diff
@@ -207,9 +234,11 @@ initialize formalProofExt :
   }
```
Paul-Lez, line 235: Could you add a few tests in `FormalConjecturesTest`? There should already be a file with tests for attributes, so you could just add a few statements that demo the functionalities you've added.

williamjblair: Yes, just added. I added the tests in `FormalConjecturesTest/Util/Attributes/Basic.lean` covering conditional `formal_proof` usage, multiple assumed hypotheses, recorded `proofConditions` + the intended errors for `conditional` without `assuming` and `assuming` without `conditional`.

The tests also caught that the documented `@[conditional formal_proof ...]` form was not actually parsing as an attribute, so I fixed that parser registration in the same commit.

williamjblair:
Merged main in to clear the conflict. It was just the util rename moving `Attributes/Basic.lean`, so nothing else changed and the attribute tests still pass.

Ready for another look whenever you have time.

williamjblair:
Some fresh evidence for this, from working through the plby status-sync backlog in #3998 today. Of the eleven problems there with a Lean proof I could find, three could not be linked because the proof is conditional, and each fails a different way:

- **1148** — [`erdos_problem_1148 (h_duke : DukeTheoremStatement)`](https://github.com/plby/lean-proofs/blob/68da20b96673899166e94638f5a7fffeb7231d35/src/latest/ErdosProblems/Erdos1148.lean). The assumption rides as a theorem parameter, so `#print axioms` is clean. The file header says "Conditional on: duke_theorem", which is the only place it is written down.
- **659** — `#print axioms erdos_659` gives `[bernays, propext, Classical.choice, Quot.sound]`. Catchable, but only if you look.
- **694** — same shape, `mertens_product`.

Our own 678 is a fourth data point in the other direction. Its docstring said the formalisation was "conditional on asymptotic estimates for the prime counting function (specifically `pi_alt` from the PNT+ project)". That stopped being true at some point, and nobody noticed, because prose in a docstring is not checked against anything. I dropped the sentence in #4720.

All four are cases where the information exists and has nowhere to live. Happy to rebase again if it has gone stale.

mo271:
@Paul-Lez can you have another look?

FormalConjecturesTest/Util/Attributes/Basic.lean
```diff
@@ -49,14 +49,79 @@ theorem a_formally_solved_problem : 2 + 2 = 4 := by
 -- formal_proof on non-research categories
 #guard_msgs in
 @[category textbook, AMS 11, formal_proof using lean4 at "https://github.com/example/proof"]
+<<<<<<< HEAD
```
Paul-Lez: Seems to be a merge conflict here?

FormalConjecturesTest/Util/Attributes/Basic.lean
```diff
@@ -49,14 +49,79 @@ theorem a_formally_solved_problem : 2 + 2 = 4 := by
...
+  rfl
+
+#guard_msgs in
+@[category research solved,
+  conditional formal_proof using lean4 at "https://github.com/example/two-hypotheses"
+    assuming conditional_formal_proof_hypothesis another_conditional_formal_proof_hypothesis]
+theorem a_conditionally_formally_solved_problem_with_two_hypotheses : 5 + 5 = 10 := by
+  rfl
+
+#eval do
```
Paul-Lez: Small trick here: you can write something like run_meta or run_cmd/etc to control the monad you run the code in (I find it a little cleaner and better behaved than using #eval)

Paul-Lez (commented):
Small pass for now, will do a more in depth one tomorrow!

williamjblair:
Fixed in 5c9ee2b, thanks for catching it.

Those markers are from my merge on 22 July. First part was the graduate/textbook rename colliding, and I kept main's name since renaming a test theorem isn't what this PR is for. Second was main and this branch both adding tests in the same spot, so I kept both. Also switched to `run_meta do`.

It survived because of #4700: the test lib has no `globs` and no root module, so `lake build FormalConjecturesTest` builds nothing and no CI job touches it. I checked the fix by compiling the file directly with the lib's own options. Everything passes except the `#AMS` docstring, and main fails that too, so it's pre-existing.

Working on the lakefile fix now so this stops being possible.

williamjblair:
Worth knowing alongside this: #4735 makes `FormalConjecturesTest` actually build and adds `lake --wfail test` to CI. I checked and the two do not conflict, in either merge order.

The effect on this PR is that the conditional-proof tests here get compiled by CI for the first time. Until now nothing built that file, which is how the conflict markers survived. I verified the tests pass by compiling the module by hand, but after #4735 that stops being on trust.

FormalConjecturesUtil/Attributes/Basic.lean
```diff
@@ -293,39 +323,68 @@ This is independent of the `category` attribute and can be used with any categor
...
+  let env ← getEnv
+  -- Resolve each assumed hypothesis to a declaration (so it is checked to exist
+  -- and hovering it jumps to the statement).
+  let conditions ← conds.toList.mapM fun (c : TSyntax `ident) => do
+    let condName ← resolveGlobalConstNoOverload c.raw
+    Elab.addConstInfo c.raw condName
+    if (env.find? condName).bind (·.value?) |>.any (!·.hasSorry) then
+      logWarning m!"The assumed hypothesis `{condName}` has a sorry-free proof, \
+        so the formal proof may no longer need to be marked `conditional`."
+    return condName
```
Paul-Lez: nit: let's also add a test that exercises this codepath

FormalConjecturesUtil/Attributes/Basic.lean
```diff
@@ -293,39 +323,68 @@ This is independent of the `category` attribute and can be used with any categor
...
+  -- Warn if this is attached to a `research open` problem.
+  let catTags := categoryExt.getState env
+  if catTags.toArray.any fun tag => tag.declName == decl &&
+      tag.category == .research .open then
+    logWarning
+      "A `formal_proof` annotation on a `research open` problem is suspicious. \
+       If a formal proof exists, the problem should not be categorised as `open`."
```
Paul-Lez, line 379: I guess here there's a convention we're "hiding": we're happy to have conditional proofs of open problems (e.g. assuming the GRH) but we're saying we should not state those with a `formal_proof` annotation, and instead state those as `research_solved` problems that have the assumption as a hypothesis.
I think it's worth:
- documenting this explicitly in this file
- changing the docstring a little to reflect this!

Paul-Lez (approved):
Thanks for adding this!

mo271 (approved):
Thanks, LGTM!

FormalConjecturesUtil/Attributes/Basic.lean
williamjblair: Can't test this one yet: the check does nothing in a `module` file, and FormalConjecturesTest is all modules. Filed as #4780.

FormalConjecturesUtil/Attributes/Basic.lean
williamjblair: Agreed, that should be written down.

The warning never fires though: all 286 `formal_proof` links in the repo sit on `research solved` problems, none on open ones. The conditional ones too, [1148](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/1148.lean), 659 and 694 are all marked solved here.

So `conditional` is recording something slightly different from what that warning guards against. [plby's 1148](https://github.com/plby/lean-proofs/blob/68da20b96673899166e94638f5a7fffeb7231d35/src/latest/ErdosProblems/Erdos1148.lean) proves a problem we call solved, but takes `DukeTheoremStatement` as a parameter, so the proof is weaker than the statement it's attached to, and `#print axioms` comes back clean. That's what needs writing down, since you can't infer it.

I'll do that here and in the docstring.

Edit: Documented in ed7b9779.

FormalConjecturesUtil/Attributes/Basic.lean
williamjblair: Done in 60ebffe9, with a test either way: one assumption still open producing nothing, one since proved producing the warning.

The check had to move out of the attribute to be testable at all. It reads the hypothesis's proof term, and `env.find?` returns nothing useful inside a module, so it was silently inert in all ten files of `FormalConjecturesTest`. `FormalProofLinter` uses `findAsync?` instead, the same way `CategoryLinter.checkNotOpenIfSorryFree` does for the `category` version. Wrote it up as #4780.

PR #4369, opened 2026-07-01
ErdosProblems: add 24, 93, 164, 314, 315, 333, 369, 401, 429, 435 (#3998 sync)

### LINKED ISSUE #493: Erdős Problem 333
### What is the conjecture
    
https://www.erdosproblems.com/333

  Let $A\subseteq \mathbb{N}$ be a set of density zero. Does there exist a basis $B$ such that $A\subseteq B+B$ and
$\lvert B\cap \{1,\ldots,N\}\rvert =o(N^{1/2})$
for all large $N$?

Status: open

### Choose either option

### LINKED ISSUE #697: Erdős Problem 369
### What is the conjecture
    
https://www.erdosproblems.com/369

  Let $\epsilon>0$ and $k\geq 2$. Is it true that, for all sufficiently large $n$, there is a sequence of $k$ consecutive integers in $\{1,\ldots,n\}$ all of which are $n^\epsilon$-smooth?

Status: open

### Choose either option

### LINKED ISSUE #712: Erdős Problem 401
### What is the conjecture
    
https://www.erdosproblems.com/401

  Is there some function $\omega(r)$ such that $\omega(r)\to \infty$ as $r\to\infty$, such that for all large $n$ there exist $a_1,a_2$ with
$$a_1+a_2> n+\omega(r)\log n$$
such that $a_1!a_2! \mid n!2^n3^n\cdots p_r^n$?

Status: open

### Choose either option

williamjblair:
Adds statements for Erdős problems 24, 93, 164, 314, 315, 333, 369, 401, 429, 435.

Docstrings quote the boxed text on erdosproblems.com verbatim. I cross-checked each statement against the two hosted Lean proofs (plby and jayyhk). `formal_proof` links are pinned to a commit and appear only where the hosted proof passed an axiom and hypothesis audit as unconditional.

Departures from the hosted formalizations:

- **24**: uses Mathlib's `SimpleGraph.copyCount G (cycleGraph 5)` for the C₅ count. The hosted proofs count labelled embeddings divided by 10. The two counts agree on every graph.
- **93**: stated in FC's concrete `ℝ²`, the house style of the sibling problems 982 and 1082. The hosted proofs quantify over an abstract two-dimensional inner-product space, which is equivalent up to isometry.
- **164**: keeps the text's primitivity definition verbatim and adds `∀ a ∈ A, 2 ≤ a` as a separate hypothesis. Without it, Lean's junk value `1 / (1 * Real.log 1) = 0` lets `A = {1}` satisfy the bound even though the informal summand is undefined there.
- **314**: wraps the liminf equality in FC's `answer(True) ↔`. jayyhk states the bare equality, and plby proves a block-existence form that implies it via minimality. I followed the text's ε(n) formulation.
- **315**: the text says `a₁ < a₂ < ⋯`, so I used `StrictMono`. That matches jayyhk and departs from plby, which assumes only `Monotone`. The two hosted proofs disagree with each other here, so I followed the text.
- **333**: stated as `answer(False) ↔ P` per FC style for a disproved yes/no question (cf. 263.parts.ii). jayyhk proves `¬P` directly, and plby stops at the counterexample pair.
- **369**: the text as literally written is trivially true, as the site commentary notes. I formalized the commentary's non-trivial variant, with the k consecutive integers confined to [n/2, n]. That variant is what the "proved (Lean)" status refers to and what both hosted proofs establish. The docstring records this.
- **401**: follows the text and jayyhk's shape, `∃ f → ∞, ∀ r ≥ 1, …`. plby proves the property for one explicit witness ω(r) without the existential wrapper.
- **429**: states the question positively as `answer(False) ↔ P`. Both hosted proofs state the refuting construction, which is the negation of P. "Sparse enough" becomes a counting-function threshold with `f → ∞`; drop the Tendsto condition and `f = 0` degenerates the hypothesis class.
- **435**: renders "the largest integer not of the form …" as `IsGreatest`. The hosted proofs state membership plus an upper bound, which is equivalent. The two hosted final theorems are byte-identical on this problem.

Part of #3998. Fixes #493, fixes #697, fixes #712.


mo271:
please mark as "Ready for review" when ready for review

williamjblair:
@mo271 good for review now!

williamjblair:
Refreshed against main for the util rename. Still good for review.

mo271 (approved):
Thanks, LGTM

FormalConjectures/ErdosProblems/314.lean
```diff
@@ -0,0 +1,62 @@
...
+- [LiSt24] J. Lim and Steinerberger, S., *On differences of two harmonic numbers*.
+  arXiv:2405.11354 (2024).
+-/
+
+open Filter
+
+namespace Erdos314
+
+/-- The block harmonic sum $\sum_{n\leq k\leq m}\frac{1}{k}$. -/
+noncomputable def harmonicSum (n m : ℕ) : ℝ := ∑ k ∈ Finset.Icc n m, (1 : ℝ) / k
```
mo271: I think this sum is already defined multiple times (at least to Q). somtimes called Egytian somethings, perhaps good to move to ForMathlib, but can be done in a follow up..

williamjblair: Done in abbeb427.

There were two copies: this one over ℝ, and 290 over ℚ inside `harmonicDen`. 291 has `∑ i ∈ Finset.Icc 1 k, (1 : ℚ) / i`, but that one is Mathlib's `harmonic k`. The Egyptian fraction files you may have had in mind (283, 306) sum `1/(n i)` over an indexed sequence, so I left them alone.

New file `FormalConjecturesForMathlib/NumberTheory/Harmonic.lean`:

```lean
def harmonicBlock (K : Type*) [DivisionRing K] (n m : ℕ) : K := ∑ k ∈ Finset.Icc n m, (k : K)⁻¹
```

It is generic over the division ring so 290 and 314 both use it. I added `harmonicBlock_one` to connect it to `harmonic`, a cast lemma, and two small ones covering the empty and successor cases. 290 becomes `(harmonicBlock ℚ a b).den` and then 314 loses its local copy.

`Rat.cast_harmonicBlock` is `@[simp, norm_cast]`, and ForMathlib is imported everywhere, so it joins the global simp set. I can't see anything it would affect, but we can wait for CI

FormalConjecturesForMathlib/NumberTheory/Harmonic.lean
mo271, line 1: Thanks for reacting so thouroughly to the comment

not entirely sure that we really want that -- instead we could just ad-hoc cast to R when needed, but I'm fine either way (i.e. adding this new file or inlining the cast where it is needed..)

williamjblair: Got it, agreed we can simplify. I'll make the change

williamjblair: Done in 7a9944e5. Kept the shared definition but dropped the `DivisionRing` generality: `harmonicBlock` is ℚ-valued now, like `harmonic`, and 314 casts to ℝ in `epsilon`. `Rat.cast_harmonicBlock` is gone with it, so nothing new joins the global simp set.

If inlining is closer to what you meant, I can drop the file and put the sum back in 290 and 314.

mo271: Looks good as is now!

mo271: Thanks!

PR #4371, opened 2026-07-02
ErdosProblems: add 443, 484, 487, 497, 498, 502, 537, 582, 618, 646 (#3998 sync)

williamjblair:
Adds statements for Erdős problems 443, 484, 487, 497, 498, 502, 537, 582, 618, 646.

Docstrings quote the boxed text on erdosproblems.com verbatim. I cross-checked each statement against the two hosted Lean proofs (plby and jayyhk). `formal_proof` links are pinned to a commit and appear only where the hosted proof passed an axiom and hypothesis audit as unconditional.

Departures from the hosted formalizations:

- **443**: parametrises the set as `{k(m-k) : 1 ≤ k ≤ m/2}` per the text. The hosted proofs use `1 ≤ r ≤ k-1`, and the two images are equal by the symmetry `k ↦ m-k`. Carries `n < m` as both hosted theorems do, since at `m = n` the intersection is all of `A n` and the `(mn)^{o(1)}` bound is not intended.
- **484**: states "at least cN" as `c * N ≤ card` over ℝ, verbatim from the text. The hosted proofs use the floor `⌊c·N⌋₊`, which is equivalent up to rescaling the existential `c`.
- **487**: reads "positive density" literally as `Set.HasPosDensity`, the FC convention (cf. 741). Both hosted proofs assume only `lowerDensity A > 0`, a weaker hypothesis, so the pinned proof covers this statement.
- **497**: states the source's exponential form, antichain count `= 2^{(1+o(1))·C(n, n/2)}`, where plby proves the equivalent log-side asymptotic. Counts via `Nat.card` of a subtype instead of a classical `filter`.
- **498**: keeps the text's `1 ≤ ‖z_i‖` with the open unit disc. The closed-disc reading is false under that hypothesis: at `n = 1`, `z = 1`, the closed disc at 0 contains both ±1. Kleitman's proof is of the open-disc form.
- **502**: the problem asks for the exact largest 2-distance set, and that remains open between `C(n+1, 2)` and `C(n+2, 2)`. Bannai, Bannai and Stanton proved the upper bound, which is what the machine-verified proof covers, so I state `|A| ≤ C(n+2, 2)` rather than an `IsGreatest` form.
- **537**: stated as `answer(False) ↔ P` with the text's ground set `{1,…,N}` (`Finset.Icc 1 N`, agreeing with jayyhk). plby uses `Finset.range (N+1)`, which admits `0 ∈ A` and trivializes the inner pattern. Both refutations go through the same Ruzsa counterexample.
- **582**: requires `G.CliqueFree 4` per the text ("contains no K₄", agreeing with jayyhk). plby proves the stronger `cliqueNum = 3`, which implies this statement.
- **618**: two readings shared with both hosted proofs. "Has diameter 2" becomes diameter at most 2, since the readings coincide for triangle-free graphs on three or more vertices, and the site's `h(G)` is the `h₂(G)` the problem itself defines.
- **646**: reads "n! is divisible by an even power of each pᵢ" as "the exponent of pᵢ in n! is even", following Berend's paper and both hosted proofs. The literal reading is trivially true at `e = 0`.

Part of #3998.


williamjblair:
Merged main in and moved the formalization notes out of the Lean files into the description, per the AGENTS.md checklist. Good for review.

mo271 (approved):
Thanks, LGTM!

PR #4383, opened 2026-07-03
feat(OEIS): README.md

mo271:
Add a README.md decidated to specifics of files in in OEIS dir, and make the style in the files we currently already have follow that style.


FormalConjectures/OEIS/README.md
```diff
@@ -0,0 +1,83 @@
...
+- **References**: It must conclude with a standardized `*References:*` (or `*Reference:*` for a single item) section containing a Markdown link to the official OEIS page, along with any other papers or articles necessary to formulate the problem.
+
+```lean
+/-!
+# Four-square conjecture with powers of 2, 3, and 5
+
+Any integer $n > 1$ can be written as $(2^a \cdot 3^b)^2 + (2^c \cdot 5^d)^2 + x^2 + y^2$
+where $a, b, c, d, x, y$ are nonnegative integers.
+
+*References:*
```
eric-wieser, line 43: I think in mathlib we'd use `##` for a subheading

FormalConjectures/OEIS/README.md
```diff
@@ -0,0 +1,83 @@
...
+-/
+@[category research solved, AMS 11, formal_proof using formal_conjectures at ...]
+theorem conjecture (n : ℕ) (hn : 1 < n) : A n := by
+```
+
+## Term Theorems (`category test`)
+
+To ensure the formalized definition behaves correctly and matches the official OEIS sequence, every file **must include term theorems verifying the first few values of the sequence** (typically the first 5 values).
+
+- **Naming**: Every term verification theorem for sequence `a` (or predicate `A`) must be named strictly `a_0`, `a_1`, `a_2`, etc., according to the index (`a_[n]`). Note that even when testing an `UpperCamelCase` property definition like `A`, Mathlib naming rules mandate lowercasing it right inside `snake_case` theorem names (`a_0 : A 0`, `a_1 : A 1`).
```
eric-wieser, line 71: Perhaps you should mention `Set` here too, since there are examples of this. Do we want `aSet`?

FormalConjectures/OEIS/63880.lean
```diff
@@ -43,69 +43,69 @@ def usigma (n : ℕ) : ℕ :=
   ∑ d ∈ unitaryDivisors n, d
 
 /-- A number $n$ is in the sequence A063880 if $\sigma(n) = 2 \cdot \text{usigma}(n)$. -/
-def a (n : ℕ) : Prop :=
+def A (n : ℕ) : Prop :=
   0 < n ∧ σ 1 n = 2 * usigma n
 
 /-- The set of numbers in the sequence A063880. -/
-def A : Set ℕ := {n | a n}
+def a : Set ℕ := {n | A n}
```
eric-wieser: Perhaps best to inline this into `IsPrimitiveTerm`

FormalConjectures/OEIS/README.md
```diff
@@ -0,0 +1,83 @@
...
+```lean
+/-!
+# Four-square conjecture with powers of 2, 3, and 5
+
+Any integer $n > 1$ can be written as $(2^a \cdot 3^b)^2 + (2^c \cdot 5^d)^2 + x^2 + y^2$
+where $a, b, c, d, x, y$ are nonnegative integers.
+
+*References:*
+- [A308734](https://oeis.org/A308734)
+- Z.-W. Sun, "Refining Lagrange's four-square theorem," *J. Number Theory* **175** (2017), 167-190.
```
eric-wieser: Let's have an example of a non-OEIS reference with a URL as well.

FormalConjectures/OEIS/README.md
```diff
@@ -0,0 +1,83 @@
...
+-- definitions, helper lemmas, term theorems, and main conjecture
+
+end OeisA308734
+```
+
+## Module Docstrings & References
+
+Every file must include a descriptive module docstring (`/-! ... -/`) immediately following the imports.
+- **Content**: The docstring should clearly explain the sequence definition and the statement of the conjecture or theorem being formalized.
+- **References**: It must conclude with a standardized `*References:*` (or `*Reference:*` for a single item) section containing a Markdown link to the official OEIS page, along with any other papers or articles necessary to formulate the problem.
```
eric-wieser: I think just always make it plural

FormalConjectures/OEIS/358684.lean
```diff
@@ -97,33 +99,30 @@ theorem a_equiv_a' (n : ℕ) : a n = a' n := by
...
 
 @[category test, AMS 11]
-theorem six : a 6 = 46 := by
+theorem a_6 : a 6 = 46 := by
   decide +native
 
-@[category test, AMS 11]
-theorem seven : a 7 = 73 := by
-  sorry
-
```
eric-wieser, line 126: Deliberate?

FormalConjectures/OEIS/228828.lean
```diff
@@ -17,7 +17,9 @@ limitations under the License.
 import FormalConjectures.Util.ProblemImports
 
 /-!
-Numbers n such that $n^2 + \pi(n)$ is prime.
+# Numbers n such that n² + π(n) is prime
+
+Numbers $n$ such that $n^2 + \pi(n)$ is prime.
```
eric-wieser: Why have this twice, once with latex and once without?

eric-wieser (approved):

FormalConjectures/OEIS/228828.lean
mo271:
```suggestion
# Numbers $n$ such that $n^2 + \pi(n)$ is prime.
```

mo271: fixed

FormalConjectures/OEIS/358684.lean
mo271: no, good catch, let's keep it since this sequence has so many leading zeros..

FormalConjectures/OEIS/README.md
mo271: done here, will make a follow up pr to harmonize this across the entire repo

FormalConjectures/OEIS/README.md
mo271: perhaps not necessary to have explicit guidance here -- the one place where we had it is inlined now...

FormalConjectures/OEIS/README.md
mo271: done

FormalConjectures/OEIS/63880.lean
mo271: done

FormalConjectures/OEIS/README.md
mo271: yeah, we could do that, but this is currently the way we do it across all files, so I suggest to go with what we currently do for now

PR #4387, opened 2026-07-03
fix(ErdosProblems): 522 — make h_unif satisfiable (counting-measure reference for the discrete coefficient sets)

### LINKED ISSUE #4386: Erdős 522: `h_unif` is unsatisfiable by measurable coefficients — the defaulted `volume` reference measure degenerates on the null set {-1, 1}

`Erdos522.KacCoefficients` requires each coefficient to be uniformly
distributed via `pdf.IsUniform`, with the reference measure `μ` a structure
parameter defaulting to `volume` (through `volume_tac`):

https://github.com/google-deepmind/formal-conjectures/blob/0bf1250c42e675c5677186ca6ac7f4b07655b346/FormalConjectures/ErdosProblems/522.lean#L37-L43

All four statements in the file (`erdos_522`, `erdos_522.variants.zero_one`,
`erdos_522.variants.number_real_roots`, `erdos_522.variants.yakir_solution`)
instantiate the structure with `μ` elided — e.g.
`c : KacCoefficients ({-1, 1} : Set ℂ) Ω` — so the `volume` default fires.

That default makes `h_unif` degenerate. Mathlib defines

```text
pdf.IsUniform X s ℙ μ  :=  Measure.map X ℙ = ProbabilityTheory.cond μ s
cond μ s               :=  (μ s)⁻¹ • μ.restrict s
```

and every coefficient set used here (`{-1, 1} ⊆ ℂ`, `{0, 1} ⊆ ℂ`,
`{-1, 1} ⊆ ℝ`) is finite, hence `volume`-null, so `μ.restrict s = 0` and
`cond volume S` is the **zero measure**. `h_unif` therefore asserts
`Measure.map (toFun i) ℙ = 0`.

But `ℙ` is a probability measure, so `Measure.map (toFun i) ℙ` is a
probability measure — in particular nonzero — whenever `toFun i` is
a.e.-measurable (`Measure.map` returns the junk value `0` exactly when the
function is not a.e.-measurable). Hence:

- every honest coefficient (an a.e.-measurable ±1 Rademacher or {0,1}
  Bernoulli variable) **fails** `h_unif`;
- the only potential inhabitants of `KacCoefficients S Ω` have
  non-a.e.-measurable coefficient functions.

So the `∀`-quantified statements range over no intended instance (only
measure-theoretic pathologies, if any exist), and none of the four theorems
constrains actual Kac polynomials. In particular `erdos_522` as stated is not
a faithful formalization of the almost-sure problem, and the two
`research solved` variants assert their bounds vacuously with respect to the
intended ensemble.

This is independent of #3473 / PR #3896: that fix corrected the convergence
mode (in-probability → almost-sure) but kept `h_unif` and the defaulted `μ`.

### Trivial proof/disproof

Machine-checked against current Mathlib (verified on Lean/Mathlib v4.31.0;
both `#print axioms` report exactly `[propext, Classical.choice, Quot.sound]`):

```lean
import Mathlib.MeasureTheory.Measure.Lebesgue.Complex
import Mathlib.Probability.Distributions.Uniform
import Mathlib.Probability.UniformOn

open MeasureTheory

-- (1) The defaulted reference measure degenerates: conditioning `volume` on the
-- Lebesgue-null coefficient set {-1, 1} yields the ZERO measure.
theorem cond_volume_pm_one_eq_zero :
    ProbabilityTheory.cond (volume : Measure ℂ) ({-1, 1} : Set ℂ) = 0 := by
  have hS : (volume : Measure ℂ) ({-1, 1} : Set ℂ) = 0 :=
    ((Set.finite_singleton (1 : ℂ)).insert (-1 : ℂ)).measure_zero _
  unfold ProbabilityTheory.cond
  rw [Measure.restrict_eq_zero.mpr hS, smul_zero]

-- (2) Hence `h_unif` (with the defaulted μ = volume) excludes every honest
-- coefficient: ANY a.e.-measurable random variable on ANY probability space
-- fails it, ±1-valued or not. (`ℙ` in 522.lean instantiates `P`.)
theorem h_unif_fails_for_aemeasurable
    {Ω : Type*} [MeasurableSpace Ω] (P : Measure Ω) [IsProbabilityMeasure P]
    {f : Ω → ℂ} (hf : AEMeasurable f P) :
    ¬ MeasureTheory.pdf.IsUniform f ({-1, 1} : Set ℂ) P volume := by
  intro h
  have hmap : (Measure.map f P) Set.univ = 1 := by
    rw [Measure.map_apply_of_aemeasurable hf MeasurableSet.univ]
    simp
  unfold MeasureTheory.pdf.IsUniform at h
  rw [h, cond_volume_pm_one_eq_zero] at hmap
  simp at hmap

-- (3) The suggested fix is non-degenerate: with μ := Measure.count, `h_unif`
-- says exactly that the law of the coefficient is `uniformOn {-1, 1}` — the
-- honest Rademacher law, a genuine probability measure.
example {Ω : Type*} [MeasurableSpace Ω] (P : Measure Ω) (X : Ω → ℂ) :
    MeasureTheory.pdf.IsUniform X ({-1, 1} : Set ℂ) P Measure.count ↔
      Measure.map X P = ProbabilityTheory.uniformOn ({-1, 1} : Set ℂ) :=
  Iff.rfl

example : IsProbabilityMeasure (ProbabilityTheory.uniformOn ({-1, 1} : Set ℂ)) :=
  ProbabilityTheory.isProbabilityMeasure_uniformOn (Set.toFinite _) ⟨-1, by simp⟩
```

(The `{0, 1} ⊆ ℂ` and `{-1, 1} ⊆ ℝ` cases used by the variants are the same
proof with the set/field changed.)

This is a faithfulness refutation in the same sense as #3473: the point is
not that the formal statements are false in Lean, but that their hypotheses
exclude every instance the informal problem is about.

### Suggested fix

**Option A (minimal, no structure change).** Instantiate the reference
measure as the counting measure at the four statement sites:

```lean
c : KacCoefficients ({-1, 1} : Set ℂ) Ω Measure.count
```

Then `h_unif` unfolds definitionally (example (3) above) to
`Measure.map (c i) ℙ = ProbabilityTheory.uniformOn {-1, 1}` — the coefficient
law is the uniform distribution on the two points, as intended. A pleasant
side effect: since `Measure.map` junk-defaults to `0` for non-a.e.-measurable
functions and `uniformOn {-1, 1} ≠ 0`, the fixed `h_unif` automatically forces
each coefficient to be a.e.-measurable — no separate measurability field is
needed.

**Option B (structural).** Replace the `h_unif` field with an explicit
law-pinning field and drop the exposed `μ` parameter for these
discrete-coefficient problems:

```lean
h_law : ∀ i, Measure.map (toFun i) ℙ = ProbabilityTheory.uniformOn S
```

Equivalent to Option A by the `Iff.rfl` above, and removes the defaulted
parameter that is easy to elide by accident.

Either way, a cheap regression guard would be to exhibit one honest instance
(e.g. coordinate projections on a Bernoulli product space, mapped to ±1)
inhabiting the fixed structure, so the statements are known to quantify over
at least the intended objects.

### Additional context

- Mathlib defines `pdf.IsUniform` w.r.t. an arbitrary reference measure
  precisely so that discrete supports can use `Measure.count`; the defect here
  is only the elided `volume` default, not `pdf.IsUniform` itself.
- Other problem files that use `pdf.IsUniform` with a finite/countable support
  and the defaulted `volume` would degenerate the same way; it may be worth a
  repo-wide audit (we have not performed one).
- Found while building a conditional Lean development of Erdős 522 downstream;
  we machine-checked the exclusion there before reporting. Happy to submit the
  fix as a PR (Option A or B, plus the witness instance) if useful.

### Choose either option

widefox:
Fixes #4386.

`KacCoefficients` requires `h_unif : ∀ i, pdf.IsUniform (toFun i) S ℙ μ` with the
reference measure `μ` defaulting to `volume` via `volume_tac`, and all four statements
in `522.lean` elided `μ`, so the default fired. The coefficient sets here are finite,
hence `volume`-null, so `ProbabilityTheory.cond volume S` is the zero measure and
`h_unif` demanded `Measure.map (toFun i) ℙ = 0` — impossible for any a.e.-measurable
coefficient (the map of a probability measure under an a.e.-measurable function is a
probability measure). The statements therefore quantified over no intended instance.
See #4386 for the machine-checked exclusion.

Changes (Option A from the issue — no structure change):

- Instantiate `μ := Measure.count` at the four statement sites (`erdos_522`,
  `erdos_522.variants.zero_one`, `erdos_522.variants.number_real_roots`,
  `erdos_522.variants.yakir_solution`). With the counting measure, `h_unif` unfolds
  definitionally to `Measure.map (toFun i) ℙ = ProbabilityTheory.uniformOn S` — each
  coefficient's law is the uniform distribution on the (two-point) coefficient set, as
  intended. Since `Measure.map` junk-defaults to `0` for non-a.e.-measurable functions
  and `uniformOn S ≠ 0`, this also forces each coefficient to be a.e.-measurable — no
  extra measurability field needed.
- Add a note to the `KacCoefficients` docstring explaining the reference-measure choice.
- Add a regression witness (`RegressionWitness` section): an honest Rademacher
  coefficient sequence — coordinate signs on `ℕ → Bool` with the infinite product of
  fair coins (`Measure.infinitePi`) — inhabiting
  `KacCoefficients ({-1, 1} : Set ℂ) (ℕ → Bool) Measure.count`
  (`coinKacCoefficients`, with `kacCoefficients_count_nonempty`), so the statements are
  known to quantify over at least the intended objects. Independence is
  `iIndepFun_infinitePi`; the marginal law is `infinitePi_map_eval` pushed through the
  sign map; the supporting `uniformOn_pair` lemma computes `uniformOn {a, b}` as the
  average of two Dirac masses.

The four theorem statements' mathematical content is otherwise unchanged (same
convergence mode, same sets, same bounds); the two open problems keep
`answer(sorry)` and everything keeps its `sorry` proof.

Verified downstream against Mathlib v4.31.0 (statements elaborate; the witness and
supporting lemmas depend on exactly `[propext, Classical.choice, Quot.sound]`); all
Mathlib names used were checked present at this repo's v4.27.0 pin. Happy to iterate on
any CI/linter feedback.

https://claude.ai/code/session_011Hhp9yDLeqxGUEGDkXT5Mh

### DIFF
--- a/FormalConjectures/ErdosProblems/522.lean
+++ b/FormalConjectures/ErdosProblems/522.lean
@@ -33,6 +33,14 @@ of independent random variables, each uniformly distributed over `S`.
 
 Such a sequence determines a *Kac polynomial* of degree `n` for each `n`, which is the random
 polynomial given by `KacCoefficients.polynomial`.
+
+Note: the uniformity hypothesis `h_unif` is relative to the reference measure `μ`. For the
+discrete coefficient sets used in this file (`{-1, 1}`, `{0, 1}`), the statements below
+instantiate `μ := Measure.count`, so that `h_unif` states that the law of each coefficient is
+`ProbabilityTheory.uniformOn S`. The `volume` default would be degenerate here: a finite `S` is
+`volume`-null, conditioning `volume` on it yields the zero measure, and no a.e.-measurable
+coefficient has the zero measure as its law. See the regression witness
+`coinKacCoefficients` below.
 -/
 @[ext]
 structure KacCoefficients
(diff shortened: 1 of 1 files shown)


mo271:
Thanks for your contribution!
Seems like this got fixed by #5793?

PR #4393, opened 2026-07-04
linter: validate formal_proof links at build time

williamjblair:
Following @Paul-Lez's review: reworked from a Python check on the generated `conjectures.json` into a build-time check, so it throws in-file at elaboration rather than at a later CI stage. Also drops the checks Paul flagged as already enforced (the `formalProofKind` set is syntax-enforced; duplicate theorem names are already a Lean error).

What it checks, in the `formal_proof` attribute elaborator (next to the existing `research open` warning):
- a `lean4` or `other_system` proof lives outside this repo and must be locatable, so an empty link warns
- any link that is given must be an http(s) URL (catches typos and relative paths)
- a `formal_conjectures` proof is in this repository, so its link may be empty — consistent with the existing test `a_test_with_formal_proof`

I put the check in the attribute elaborator itself since that is where the link is parsed and where the `research open` warning already lives; happy to move it into a standalone linter under `Util/Linters` if you would rather have the toggle.

Verified locally: `FormalConjectures.Util.Attributes.Basic` builds, and the two new `#guard_msgs` tests (empty external link, malformed link) pass. Also fixes an invalid `category graduate` in the attribute test file (not a real category) that was blocking that file from building.


scripts/check_conjectures_json.py
```diff
@@ -0,0 +1,82 @@
...
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+from collections import Counter
+from urllib.parse import urlparse
+
+# The kinds accepted by the `formal_proof using <kind> at "..."` attribute.
+ALLOWED_KINDS = {"formal_conjectures", "lean4", "other_system"}
```
Paul-Lez: Note that these kinds should already be enforced at the build level (the syntax of the attribute shouldn't allow any other values)

Paul-Lez (commented):
Thanks for the contribution - this is a nice idea! I think these checks would make sense as lean linters - would you be down to implement them that way (such linters have the advantage that they throw a build error (and warnings inside the file) before even reaching other stages of ci)

Paul-Lez:
Happy to advise on the Lean linter implementation of this (assuming you have the appetite for that)

williamjblair:
Thanks, appreciate the offer to advise! I took an early attempt and reworked it into a build-time check so it throws in-file at elaboration rather than at a later CI stage.

I put it in the formal_proof attribute elaborator itself (right next to the existing research open warning), since that's where the elaborator already parses the link. It warns if a lean4/other_system proof has no link (the proof is external and needs locating), and if any given link isn't an http(s) URL. I left formal_conjectures links optional since the proof is in-repo, which matches the existing a_test_with_formal_proof test. Dropped the kind and duplicate-name checks per your note that those are already enforced.

One open design question: we could move it into a standalone linter under Util/Linters if you'd prefer that toggle over an always-on elaborator check. Also fixed an invalid `category graduate` in the test file that was blocking it from building.

mo271 (approved):
Thanks, LGTM!

PR #4396, opened 2026-07-05
feat(ErdosProblems/506): add circles-determined-by-n-points statement

### LINKED ISSUE #2128: Erdős Problem 506: Minimum circles determined by n points in the plane
### What is the conjecture

 Let $S$ be a set of $n$ points in $\mathbb{R}^2$ not all lying on a single circle. A **circle determined by $S$** is any circle passing through at least three points of $S$. The problem asks: what is the **minimum number of distinct circles determined by $S$**?

A classical result (Elliott, Purdy & Smith) establishes that for $n > 393$ points, the minimum number of circles is at least:
$$\binom{n-1}{2} + 1 - \left\lfloor\frac{n-1}{2}\right\rfloor$$

This bound is tight, as demonstrated by the configuration of $n-1$ points on a circle with one point positioned outside the circle. However, the problem remains **open for small values of $n$** (e.g., $n = 8$), where a counterexample has been discovered through projecting a cube onto a plane.

 **Sources:**
 - https://www.erdosproblems.com/506, Elliott, Purdy & Smith (research on circle configurations and incidence geometry)

### Prerequisites needed

**Formalizability Rating:** 3/5 (0 is best) (as of 2026-02-02)

- Basic point and circle types in $\mathbb{R}^2$ (available as `EuclideanSpace ℝ 2` or `ℝ × ℝ`)
- Finite set combinatorics and counting bounds

- Formal definition of "circle determined by a set of points" (incidence relation: circles passing through at least 3 specific points from a finite set)
- Combinatorial optimization/extremal result: machinery to express and prove lower bounds on the minimum number of circles in terms of set size and configuration constraints

Rating justification (1-2 sentences): Stating the conjecture requires defining point sets, circle incidence, and the minimum/counting operation over circles, which exist in Mathlib but need some gluing together; the main challenge is formalizing the extremal combinatorial statement itself (expressing "minimum number over all configurations"), not the underlying geometric objects.

* ams-51
* ams-52
* ams-05

### Choose either option

---

theebayuser:
Resolves #2128

## Summary
- Formalises [Erdős Problem 506](https://www.erdosproblems.com/506): the minimum number of circles determined by $n$ points in $\mathbb{R}^2$, not all on a circle.
- States Elliott's resolution for $n > 393$, with the Purdy–Smith correction $\binom{n-1}{2} + 1 - \lfloor (n-1)/2 \rfloor$, as an `IsLeast` statement.
- Adds the `segre_n_eq_eight` variant: Segre's cube-projection observation that the uncorrected bound $\binom{n-1}{2}$ fails at $n = 8$.

## Test plan
- [x] `lake --wfail build 'FormalConjectures.ErdosProblems.«506»'` passes with no errors or warnings.
- [x] Problem statement and formula checked against the live erdosproblems.com page.


Smetalo:
Does this resolve #2128?

theebayuser:
Yes, this formalises the problem tracked in #2128. I've added `Resolves #2128` to the PR description so it closes on merge.

mo271 (approved):
Thanks, LGTM!

FormalConjectures/ErdosProblems/506.lean
```diff
@@ -0,0 +1,75 @@
...
+
+There is a non-degeneracy condition intended here; following Elliott we assume the points are not
+all on a single line and not all on a single circle.
+
+For $n > 393$ the answer is $\binom{n-1}{2} + 1 - \left\lfloor \frac{n-1}{2} \right\rfloor$
+(Elliott [El67], with the correction of Purdy–Smith reported in [BaBa94]): this is both a lower
+bound for every such configuration and is attained, e.g. by a circle with $n - 1$ points together
+with a single point off the circle.
+-/
+@[category research solved, AMS 51 52]
```
mo271, line 74: can we add a statement `research open` for the small `n`?

theebayuser: i Added `erdos_506.variants.small_n` as `research open` for `4 ≤ n ≤ 393`, the range Elliott's result doesn't cover.

mo271 (commented):
thanks, the only remaining question I have is about which should be the main variant `erdos_506`: it seems to me verbatim the main question is the one for *all* `n` and then we have the `variants.large_n` which is `research open` and `erdos_506.variants.small_n` which is `research open` and would imply together with `large_n` the main question. 
Can we make sure we have three theorems like this?
(I find it a bit confusing that the probelm is makred in green on erdosproblems.com, even though the status is only "Decidable"... but this is also a quirk with some other problems..

theebayuser:
> thanks, the only remaining question I have is about which should be the main variant `erdos_506`: it seems to me verbatim the main question is the one for _all_ `n` and then we have the `variants.large_n` which is `research open` and `erdos_506.variants.small_n` which is `research open` and would imply together with `large_n` the main question. Can we make sure we have three theorems like this? (I find it a bit confusing that the probelm is makred in green on erdosproblems.com, even though the status is only "Decidable"... but this is also a quirk with some other problems..

I restructured it into 3 theorems

- `erdos_506`: the verbatim question for all `n`, `research open` with `answer(sorry)`
- `erdos_506.variants.large_n`: Elliott's `n > 393` answer with the Purdy–Smith correction, `research solved`
- `erdos_506.variants.small_n`: the open range `4 ≤ n ≤ 393`, `research open`

Together the two variants answer the main question by a case split at 393. I kept `large_n` as `research solved` rather than `research open` since Elliott's result is proved.

The `4 ≤ n` hypothesis on the main statement isn't a restriction of the problem - below four points nothing satisfies both non-degeneracy conditions, so the set is empty and `IsLeast` would be false.

mo271 (approved):
Thanks, LGTM!

PR #4398, opened 2026-07-05
Formalize Green's Open Problem 66

### LINKED ISSUE #1686: Green's Open Problems #66
### What is the conjecture
https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.66

Is there always a sum of two squares between $X - \frac{1}{10} X^{1/4}$ and $X$?

* ams-11

### Choose either option

xixifusi1213-gif:
## Summary
- Formalizes Ben Green's Open Problem 66 as an eventual statement over sufficiently large real X.
- Adds a local `IsSumOfTwoSquares` predicate for the statement.

Closes #1686.

## Validation
- `lake build FormalConjectures.GreensOpenProblems.«66»`

### DIFF
--- a/FormalConjectures/GreensOpenProblems/66.lean
+++ b/FormalConjectures/GreensOpenProblems/66.lean
@@ -39,7 +39,19 @@ theorem green_66 :
     answer(sorry) ↔
       ∀ᶠ X : ℝ in atTop,
         ∃ n : ℕ, IsSumOfTwoSquares n ∧
-          X - (1 / 10 : ℝ) * X ^ ((1 : ℝ) / 4) ≤ n ∧ (n : ℝ) ≤ X := by
+          (n : ℝ) ∈ Set.Icc (X - (1 / 10 : ℝ) * X ^ (1 / 4 : ℝ)) X := by
+  sorry
+
+/--
+Green notes that a well-known, almost trivial argument gives an $O(X^{1/4})$ bound on the left.
+-/
+@[category research solved, AMS 11]
+theorem green_66.variants.trivial_bound :
+    ∃ C > (0 : ℝ), ∀ᶠ X : ℝ in atTop,
+      ∃ n : ℕ, IsSumOfTwoSquares n ∧
+        (n : ℝ) ∈ Set.Icc (X - C * X ^ (1 / 4 : ℝ)) X := by
+  -- Subtract the greatest square `u ^ 2` below `X`, then the greatest square `v ^ 2`
+  -- below `X - u ^ 2`.
   sorry
 
 end Green66
--- a/FormalConjectures/GreensOpenProblems/66.lean
+++ b/FormalConjectures/GreensOpenProblems/66.lean
@@ -1,0 +1,45 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


FormalConjectures/GreensOpenProblems/66.lean
```diff
@@ -0,0 +1,45 @@
...
+/--
+Is there always a sum of two squares between $X - \frac{1}{10}X^{1/4}$ and $X$?
+We formalize this as an eventual statement for sufficiently large real $X$.
+-/
+@[category research open, AMS 11]
+theorem green_66 :
+    answer(sorry) ↔
+      ∀ᶠ X : ℝ in atTop,
+        ∃ n : ℕ, IsSumOfTwoSquares n ∧
+          X - (1 / 10 : ℝ) * X ^ ((1 : ℝ) / 4) ≤ n ∧ (n : ℝ) ≤ X := by
```
jeangud: We could directly leverage the Set intervals library here. This also helps harmonize type coercions a bit:
```suggestion
          (n : ℝ) ∈ Set.Icc (X - (1 / 10 : ℝ) * X ^ (1 / 4 : ℝ)) X := by
```

FormalConjectures/GreensOpenProblems/66.lean
```diff
@@ -0,0 +1,45 @@
...
+  sorry
+
```
jeangud, line 56: I propose adding the "trivial" variant mentioned in Green's comments as a stepping stone, how does this look?

```suggestion

/-- There is a wellknown, almost trivial, argument showing a bound of $O(X^{1/4})$ on the left. [Gr24] -/
@[category research solved, AMS 11]
theorem green_66.variants.trivial_bound :
    ∃ C > (0 : ℝ), ∀ᶠ X : ℝ in atTop,
      ∃ n : ℕ, IsSumOfTwoSquares n ∧
        (n : ℝ) ∈ Set.Icc (X - C * X ^ (1 / 4 : ℝ)) X := by
  -- Subtract off the greatest square $u^2$ less than $X$,
  -- then subtract off the greatest square $v^2$ less than $X - u^2$ [Gr24].
  -- [GrLi19] have a somewhat different argument, giving a sum of two almost equal squares within O(X1/4 of X. [Gr24]
  sorry

```

Builds fine:
```shell
$ lake build FormalConjectures/GreensOpenProblems/66.lean 
Build completed successfully (8037 jobs).
```

jeangud (commented):
Thank you @xixifusi1213-gif for the contribution!

This looks great, I see maybe two quick comments and we should be good to merge 👍

FormalConjectures/GreensOpenProblems/66.lean
xixifusi1213-gif: Thanks for the suggestion. I've replaced the two inequalities with membership in Set.Icc, which is the more idiomatic mathlib formulation for this closed interval. I also verified the change with lake --wfail build.

FormalConjectures/GreensOpenProblems/66.lean
xixifusi1213-gif: Thanks for the suggestion. I understand that this variant records the known O(X^(1/4)) bound: it only requires some positive constant C, whereas the main open problem fixes the coefficient at 1/10. I agree that it is a helpful solved stepping stone, and I have added it and verified the full build

jeangud (approved):
Thank you again!

PR #4399, opened 2026-07-05
Add OEIS A397433 reference to Erdős problem 448

hbakkaoui-hash:
Adds a reference to the OEIS entry for this problem's sequence: tau^+(n) is now published as A397433 (https://oeis.org/A397433), and the Ford constant alpha is A074738. Docstring-only change (no code changes).

I used Claude Code with the Claude Fable 5 model to add the two docstring lines to `448.lean`, linking to A397433 and A074738. No code or proofs were changed, and I personally reviewed and verified both links; A397433 is my own OEIS submission.


mo271 (approved):
Thanks, LGTM

PR #4401, opened 2026-07-06
feat(ErdosProblems): 793

### LINKED ISSUE #950: Erdős Problem 793
### What is the conjecture

https://www.erdosproblems.com/793

  Let $F(n)$ be the maximum possible size of a subset $A\subseteq\\{1,\ldots,N\\}$ such that $a\nmid bc$ whenever $a,b,c\in A$ with $a\neq b$ and $a\neq c$. Is there a constant $c$ such that
$$F(n)=\pi(n)+(c+o(1))n^{2/3}(\log n)^{-2}?$$

Status:  open

### Choose either option

danielchin:
Closes #950 

Formalizes https://www.erdosproblems.com/793

It's not explicit in the original statement (I also don't seem to have access to the original paper) that `b = c`, but I'm working under the assumption that they can't be the same element.

Assisted by Gemini

### DIFF
--- a/FormalConjectures/ErdosProblems/793.lean
+++ b/FormalConjectures/ErdosProblems/793.lean
@@ -1,0 +1,84 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


mo271:
seems like this was added by #6101

PR #660, opened 2025-09-07
Erdős Problem 92

mo271:
fixes #325

PR #4411, opened 2026-07-08
feat(ErdosProblems/92): prove possible_f_values_BddAbove

kukushking:
# Description

> Let $f(n)$ be maximal such that there exists a set $A$ of $n$ points in $\mathbb{R}^2$
> in which every $x \in A$ has at least $f(n)$ points in $A$ equidistant from $x$.
> [[Source]](https://www.erdosproblems.com/92)

- Closes the `category test` sanity-check `sorry` on `possible_f_values_BddAbove` introduced in #660 (cc @mo271 ) — the set of feasible values is bounded above, so the `sSup` defining `f` is well-defined.
- Single short proof (~13 lines): pick any `x ∈ A` (from `Nonempty`), then every equidistant class at `x` is a filter of `A.erase x`, so its card is ≤ `n` via `Finset.card_filter_le` → `Finset.card_erase_le` → `card = n`. The `sSup` bound closes with `csSup_le'`, which also absorbs the empty-distances corner case (`sSup ∅ = 0`).
- The proof bounds by `n` rather than the sharper `n - 1` mentioned in the original docstring: `BddAbove` only needs *some* bound, and `n` avoids the ℕ-subtraction edge case at `n = 0`. I adjusted the docstring wording to match. If you'd prefer to keep the `n - 1` framing, the same proof works via `Finset.card_erase_of_mem` — your call.
- :robot: Generated by a custom Lean-kernel-verified theorem-proving harness — an attempt loop with compiler-error feedback, driving **Claude Fable 5** as the LLM backend. This target closed on attempt 1 with no target-specific hints; I reviewed and verified locally (axiom check below).

### References

- [[erdosproblems.com/92]](https://www.erdosproblems.com/92) Erdős Problem 92.

# Testing

:white_check_mark: Builds cleanly:
```shell
$ lake build FormalConjectures.ErdosProblems.«92»
Build completed successfully (8795 jobs).
```

:white_check_mark: Kernel axiom check — only Lean foundation axioms, no `sorryAx`, no user axioms:
```
'Erdos92.possible_f_values_BddAbove' depends on axioms: [propext, Classical.choice, Quot.sound]
```

:white_check_mark: Theorem statement is byte-identical to the original (no signature drift); only the proof body and one docstring sentence changed.


FormalConjectures/ErdosProblems/92.lean
```diff
@@ -53,16 +53,23 @@ noncomputable def possible_f_values (n : ℕ) : Set ℕ :=
...
 @[category test, AMS 52]
 theorem possible_f_values_BddAbove (n : ℕ) : BddAbove (possible_f_values n) := by
-  use n - 1
-  rintro k ⟨points, h_card, h_prop⟩
-  unfold Erdos92.hasMinEquidistantProperty at *
-  unfold Erdos92.maxEquidistantPointsAt at *
-  sorry
+  refine ⟨n, fun k hk => ?_⟩
+  obtain ⟨points, hcard, hne, hall⟩ := hk
+  obtain ⟨x, hx⟩ := hne
```
Paul-Lez: (untested)
```suggestion
  obtain ⟨points, hcard, ⟨x, hx⟩, hall⟩ := hk
```

FormalConjectures/ErdosProblems/92.lean
```diff
@@ -53,16 +53,23 @@ noncomputable def possible_f_values (n : ℕ) : Set ℕ :=
...
+  refine (hall x hx).trans ?_
+  unfold maxEquidistantPointsAt
+  refine csSup_le' fun m hm => ?_
+  simp only [Finset.mem_coe, Finset.mem_image] at hm
```
Paul-Lez: Does this work
```suggestion
  rw [Finset.mem_coe, Finset.mem_image] at hm
```

kukushking:
Thanks @Paul-Lez Applied both suggestions — verified locally: build clean, axiom check unchanged.

mo271 (approved):
Thanks, LGTM

PR #4415, opened 2026-07-10
ErdosProblems/274: add formal_proof link for the abelian variant

Jostamon:
erdos_274.variants.abelian is tagged research solved but has no proof reference. This PR adds a formal_proof using lean4 attribute linking to a sorry-free Lean 4 proof of the finite abelian case of the Herzog–Schönheim conjecture, following a simplified Berger–Felzenbaum–Fraenkel argument.

The proof is hosted externally in accordance with the CONTRIBUTING guidance on proof length:
https://github.com/Jostamon/erdos274-hs-abelian. CI builds the project and checks its axioms. The linked theorem has the same statement up to renaming the unused binder hG to _hG.

The formalisation and proof were developed with assistance from Claude Fable 5 (Anthropic).


Jostamon:
Updated the proof reference and rebased on current main. The three fork workflows are awaiting maintainer approval.
Ready for review.

mo271 (approved):
Thanks, LGTM!

PR #4416, opened 2026-07-10
feat(Wikipedia): add Integer Factorization, Discrete Log, and Square-root Sum conjectures

### LINKED ISSUE #4414: Integer factorization in P?
### What is the conjecture

The integer factorization problem is (not) solvable in deterministic polynomial time on a classical computer. This is a well-known open question in computer science with implications for cryptography.

<https://en.wikipedia.org/wiki/Integer_factorization>

### Prerequisites needed

We already have infrastructure in mathlib and formal-conjectures on polynomial time computability. This should make it fairly easy to add the statement of this claim. All that really would be needed would be some additional infrastructure to encode lists of naturals as bitstrings.

* ams-68

### Choose either option

Paramveersingh-S:
Resolves #4414

### Description
This PR adds the formal statements for the Integer Factorization in P problem as well as related complexity theory open problems (Discrete Logarithm and Square-root Sum).

To support these conjectures within the existing `ComplexityClass.P` infrastructure (which expects polynomial time computable functions over `List Bool`), I've added a new encoding file `EncodingNat.lean`.

### Additions:
1. **`FormalConjecturesForMathlib/Computability/EncodingNat.lean`**
   - Added standard `FinEncoding` for `ℕ`, `List ℕ`, and related tuple types.
   - Leverages `encodeListNat` as a prefix-free boolean encoding to satisfy polynomial time computability definitions without exponential input inflation.

2. **`FormalConjectures/Wikipedia/IntegerFactorization.lean`**
   - Formalized `FactorizationDecision`, conjecturing `integer_factorization_not_in_P`.
   - Formalized `DiscreteLogDecision`, conjecturing `discrete_logarithm_not_in_P`.
   - Formalized `SquareRootSumDecision`, conjecturing `square_root_sum_in_P` (using `List ℕ × ℕ` mapped through `Real.sqrt`).


BoltonBailey:
Thanks for making this! I am actually starting to think that we should make a typeclass for bitstring encodings, so we don't have to define encoding after encoding for the different input and output types of these functions. This would maybe involve a change to the existing encoding infrastructure, but perhaps the formal conjectures review team might eventually find it overall easier. I'd be interested to hear other's thoughts on this.

mo271:
> Thanks for making this! I am actually starting to think that we should make a typeclass for bitstring encodings, so we don't have to define encoding after encoding for the different input and output types of these functions. This would maybe involve a change to the existing encoding infrastructure, but perhaps the formal conjectures review team might eventually find it overall easier. I'd be interested to hear other's thoughts on this.

Sounds great! I think this would indeed make things easier: there are many questions of that type, i.e. about algorithmic complexity, that are currently not straightforward to formalise

FormalConjectures/Wikipedia/IntegerFactorization.lean
```diff
@@ -0,0 +1,91 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+import FormalConjectures.Millenium.PvsNP
+import FormalConjecturesForMathlib.Computability.EncodingNat
+import Mathlib.Data.Real.Sqrt
```
mo271: we usually only have one import, see README.md

mo271 (commented):
Thanks!

PR #4417, opened 2026-07-10
fix(OEIS): correct a089026 definition to match true OEIS sequence

### LINKED ISSUE #4410: Autoformalization error on `auto_oeis`: `a089026` in `7406_a8289e8e.lean` encodes A000188, not A089026 (statement false at n=4)
Hi — while cross-checking OEIS autoformalizations I hit what looks like a
definition mismatch produced by the autoformalizer on the `auto_oeis` branch. Flagging it constructively; happy to send a PR if useful.

**Location**

- Branch: `auto_oeis` (not on `main`)
- File: `FormalConjectures/OEIS/Auto/7406_a8289e8e.lean`
- Commit: `67338a157bbb8d87e9a349d662f82a868bda6327`
- Permalink: https://github.com/google-deepmind/formal-conjectures/blob/67338a157bbb8d87e9a349d662f82a868bda6327/FormalConjectures/OEIS/Auto/7406_a8289e8e.lean

**The mismatch**

The file targets the Eldar–Ordowski conjecture on A007406 (numerators of
`Sum_{k=1..n} 1/k^2`), whose RHS is A089026, and formalizes:

```lean
def a089026 (n : ℕ) : ℕ := Nat.sqrt n.squarePart
theorem a007406_conjecture_0 (n : ℕ) (hn : n > 3) :
    Nat.gcd n (a (n - 1)) = a089026 n := by sorry
```

But OEIS **A089026** is `a(n) = n if n is prime, else 1`
(1, 2, 3, 1, 5, 1, 7, 1, 1, 1, 11, …). The Lean `a089026` instead computes
`Nat.sqrt (squarePart n)` = the largest `d` with `d^2 ∣ n`, which is
**A000188**, a different sequence. So the `sorry`'d statement is not the
intended conjecture, and it is false as written.

**Counterexample at n = 4 (checkable by hand)**

- `a` is the numerator of `H2(m) = Sum_{k=1..m} 1/k^2`. For n = 4 the LHS uses
  `a (4-1) = a 3`:
  `H2(3) = 1 + 1/4 + 1/9 = (36 + 9 + 4)/36 = 49/36`.
  `49 = 7^2` is coprime to `36`, so the fraction is reduced and `a 3 = 49`.
- **LHS:** `Nat.gcd 4 49 = 1` (since `49 = 7^2` and `gcd(4,7) = 1`).
- **RHS (Lean `a089026 4`):** `squarePart 4 = 4` (largest square dividing 4),
  so `Nat.sqrt 4 = 2`. Hence `a089026 4 = 2`.
- `1 ≠ 2`, so `a007406_conjecture_0` is **false at n = 4**.

For comparison, the *intended* RHS gives `A089026(4) = 1` (4 is not prime),
which matches the LHS `gcd = 1` — i.e. the genuine Eldar–Ordowski conjecture is
fine; only the formalized `a089026` definition is off. The same definition is
false at every prime `p > 3` as well (there `LHS = p` by Wolstenholme, but the
Lean RHS gives `sqrt(squarePart p) = 1`).

**Suggested fix**

Replace the definition with the actual A089026, e.g.
`def a089026 (n : ℕ) : ℕ := if n.Prime then n else 1`,
and re-check the statement (the intended conjecture is verified on OEIS to
n = 10^5). The `A000188`-style `Nat.sqrt n.squarePart` was likely substituted
for the true A089026 during autoformalization.

I checked `main` — the file is only on `auto_oeis`, so this is a WIP
autoformalizer-branch artifact rather than anything merged; posting in case the
branch feeds later merges or a dedup pass.

---

John Erlbacher (Independent Researcher) · Demonstrandum
Zenodo: https://doi.org/10.5281/zenodo.20673865 ·
Repo: https://github.com/demonstrandum-research/artifacts

Paramveersingh-S:
Resolves #4410

### Description
This PR fixes a definition mismatch produced by the autoformalizer for sequence `A089026` in `7406_a8289e8e.lean`.

The file targets the Eldar–Ordowski conjecture on A007406, but the previous formalized definition for A089026 computed `Nat.sqrt n.squarePart` (which is A000188). This caused the conjecture statement to be trivially false at `n=4` and every prime `p > 3`.

I have updated `a089026` to correctly reflect the true sequence:
`if n.Prime then n else 1`

*(This fix targets the `auto_oeis` branch where this file resides).*


mo271:
Hi, thanks for the contribution. The auto-oeis not part of `main` for a reason, but as a way forward here: it might be good to have a pr with only adding sequence A89026 and its conjecture(s). 
Currently your are adding 485 files -- this is not what we want. (Converting to draft for now)

PR #4418, opened 2026-07-10
feat(Complexity): add bounds on clique-detecting circuits (Issue #4404)

### LINKED ISSUE #4404: Bounds on clique-detecting circuits, in *tiny* graphs
### What is the conjecture

The question would be, for *tiny* values of $n$ and $k$:

How many NAND gates are needed to detect $k$-vertex cliques in an $n$-vertex graph?

The idea is to get bounds for, like, $n=8$ and $k=4$. This is (hopefully!) easier than resolving the
[P ?= NP](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/Millenium/PvsNP.lean) question.

### Prerequisites needed

To represent a circuit with unbounded-fan-in NAND gates, I found it convenient to define a trapezoidal bit array (backed by a BitVec). I don't think this should be added to MathLib, though. (And I'm not sure a BitVec is the best model for this.)

[lean-clique-nand](https://github.com/joshtburdick/lean-clique-nand) is my attempt at formalizing this. Although I plan to formalize this (by making a pull request to this repo), I'm new to Lean. (This is why I practiced, using a separate tiny repo.) Comments on this would be appreciated. Also, if someone else wants to do this, that would also be great.

* ams-03

### Choose either option

Paramveersingh-S:
Resolves #4404

### Description
This PR formalizes the problem of finding exact minimal bounds for unbounded fan-in NAND circuits detecting a $k$-vertex clique in an $n$-vertex graph. It ports and refines the draft implementation from the `lean-clique-nand` repository into the `formal-conjectures` project. 

The conjecture is formulated as a miniaturized version of circuit complexity lower bounds, explicitly targeting tiny graphs ($n=8, k=4$).

### Changes
1. **`FormalConjecturesForMathlib/Complexity/TrapezoidalBitArray.lean`**
   - Added `TrapezoidalBitArray` to cleanly represent upper-triangular adjacency matrices and variable inputs to gates as a flat `BitVec`.
2. **`FormalConjecturesForMathlib/Complexity/NandCircuit.lean`**
   - Added the base logic for evaluating unbounded fan-in NAND gates where arbitrary previous bits are accessible via trapezoidal bit indexing.
3. **`FormalConjectures/Complexity/CliqueCircuit.lean`**
   - Cleaned up the graph representation to idiomatically use Mathlib's `Finset (Fin n)`.
   - Defined `cliqueDetectorCircuit` and `IsMinCliqueCircuit`.
   - Formulated `clique_8_4_bound`, leaving the exact optimal size open to formalization by solvers.


FormalConjectures/Complexity/CliqueCircuit.lean
```diff
@@ -0,0 +1,52 @@
+import FormalConjecturesForMathlib.Complexity.TrapezoidalBitArray
+import FormalConjecturesForMathlib.Complexity.NandCircuit
+import Mathlib.Data.Finset.Basic
+
+/-
+Copyright (c) 2026 The Formal Conjectures Authors.
+Released under Apache 2.0 license as described in the file LICENSE.
+-/
```
mo271, line 8: that won't work like this

mo271 (commented):
Thanks! 
The problem looks certainly interesting!

Can you take a look at how our repo usually is structured and the convention we follow outlined in the README etc? 

Especially: we need to find a home for the conjecture: where is it coming from mentioned etc? PErhaps  there is some paper?

Paramveersingh-S:
Hi,
there actually isn't a specific published research paper for this exact formulation. I proposed this $n=8, k=4$ bound in Issue #4404 as a miniaturized, concrete testbed for circuit complexity lower bounds (it's something I was playing around with in my own side-repo to practice formalizing circuits). 
Since it doesn't come from the standard literature or a famous problem list, I've moved the file to the `FormalConjectures/Other/` directory per the `CONTRIBUTING.md` guidelines, and I added a reference link to the original issue in the file header.

Issue #4259, opened 2026-06-14
Add a machine-checked barrier (lower bound) for Erdős Problem 307


ElVec1o:
`ErdosProblems/307.lean` currently states the open problem and its coprime variants. I have a `sorry`-free Lean 4 (Mathlib) formalization of a **lower bound** on any solution: at least 59 primes in total, and `(∏ P)² ≥ 4·10¹¹²` (so `∏ P ≥ 2·10⁵⁶`).

I'd like to add it to that file as a `@[category research solved, AMS 11]` theorem with the proof hosted in my own repository and linked via `@[formal_proof using lean4 at …]`, following the pattern in `ErdosProblems/26.lean`. The main `erdos_307` statement stays `research open` and untouched.

- Proof + paper: https://github.com/ElVec1o/erdos307 (archived: https://doi.org/10.5281/zenodo.20684626)
- The proof builds on Lean v4.30; only the *statement* would live here (built by CI under v4.27). The linked `Closed.lean` (tag `v1.0.0`) proves both conjuncts sorry-free (`card_ge_59`, `erdos307_barrier_closed`); axioms `propext, Classical.choice, Quot.sound` + a `native_decide` evaluation of the first 59 primes.

I'll open a PR alongside this issue. Happy to reshape it however you prefer — standalone theorem vs. a reference on `erdos_307`. Thanks!

Smetalo:
Is this resolved by #4260?

PR #4429, opened 2026-07-14
feat(ErdosProblems): machine-checked lower-bound barrier for Erdős 307

SNAPKITTYWEST:
Closes part of #4259.

## Summary

Adds `FormalConjectures/ErdosProblems/307_Barrier.lean` — a machine-checked lower-bound certificate showing that any solution to Erdős Problem 307 requires a prime greater than 7.

**Key lemma:** `erdos_307_barrier_le_7` — if both sets P and Q consist only of primes ≤ 7, then (∑ 1/p)(∑ 1/q) ≠ 1.

**Proof sketch:** The sum of reciprocals of all primes ≤ 7 is 319/420 < 1, so the product of two such sums is < 1, verified by `norm_num`.

This is a concrete first step toward the machine-checked barrier requested in issue #4259. The bound can be extended by adding more primes to `primesUpTo7`.

## Test plan
- [ ] `lake build FormalConjectures.ErdosProblems.307_Barrier`


danielchin:
It looks like created a new file called `FormalConjectures/ErdosProblems/307_Barrier.lean`. Since this is related to Erdos Problem 307, can you just add this variant to `FormalConjectures/ErdosProblems/307.lean`?

SNAPKITTYWEST:
Thanks for the review @danielchin — happy to move the content into `307.lean` instead of a separate file.

My contribution is a simpler, elementary barrier: if both P and Q consist only of primes ≤ 7, then their product of reciprocal sums cannot equal 1. The proof uses only `norm_num` — no `native_decide`, no external dependencies. It is intentionally minimal so it can serve as a readable first example alongside the stronger Bonfioli barrier.

I will update the PR to add the two lemmas (`sum_recip_primes_le_7_lt_one` and `erdos_307_barrier_le_7`) directly into the `Erdos307` namespace in `307.lean`. Should I also add a `@[category research solved, AMS 11]` tag or leave it as an unlabeled supporting lemma?

Background: I built this as part of a formal verification project where I was catching AI-hallucinated theorems. The elementary barrier was the first machine-checked result I could produce cleanly with `norm_num` alone.

SNAPKITTYWEST:
Done @danielchin — content moved into `307.lean`. The separate `307_Barrier.lean` file has been deleted.

Two lemmas now live in the `Erdos307` namespace:
- `erdos_307.sum_recip_primes_le_7_lt_one` — norm_num proof that 1/2+1/3+1/5+1/7 < 1
- `erdos_307.barrier_le_7` — if P,Q ⊆ primes ≤ 7 then (∑1/p)(∑1/q) ≠ 1

Both proofs use only `norm_num`, `simp`, `interval_cases`, and `nlinarith`. No `native_decide`.

PR #4426, opened 2026-07-13
feat(WrittenOnTheWallII): mark Graph Conjecture 316 as solved

KitaKen1:
This PR marks Written on the Wall II Graph Conjecture 316 as `research solved` and adds a `formal_proof using lean4` link to an external Lean 4 proof.

The Lean proof was prepared by [KitaKen1 (Kenta Kitamura)](https://github.com/KitaKen1).

Formal proof: https://github.com/KitaKen1/wowii-graph-conjecture-316-lean/blob/3335e07151bc43e86d5c104dd30fee3596f06410/GraphConjecture316.lean

Lean4Web: https://live.lean-lang.org/#url=https://raw.githubusercontent.com/KitaKen1/wowii-graph-conjecture-316-lean/3335e07151bc43e86d5c104dd30fee3596f06410/GraphConjecture316Lean4Web.lean

The external proof establishes the same theorem statement as the one in Formal Conjectures and compiles without `sorry` or `admit`.

The proof and write-up were prepared with AI assistance, as disclosed in the external repository README.

This contribution was discussed in the Formal Conjectures Zulip stream, where opening a PR and using the `formal_proof` attribute was recommended.

PR #4431, opened 2026-07-14
Prove Written on the Wall II Conjecture 316

SamuelSchlesinger:
## What changed

- aligns Written on the Wall II Conjecture 316 with the source's nontrivial-order hypothesis
- replaces the placeholder with a complete counting and classification proof
- updates the source URL, category, and Formal Conjectures census annotations

## Proof idea

Split the vertices into leaves and the non-pendant core. Counting gives an upper bound on the number of edges, while the complement degree hypothesis gives a lower bound. Their combination forces the core to have at most three vertices; the remaining cases classify into families with well-defined total domination.

## Validation

- `lake --wfail build` passed on the reviewed combined branch (8,869 jobs)
- the changed file compiles directly with `lake env lean`
- the proof contains no `sorry` and does not depend on Conjecture 315
- an axiom audit found no `sorryAx`
- all connected simple graphs on 2 through 6 vertices were exhaustively checked against the statement


mo271:
Thanks!

Closing because it seems like it is an duplciate of #4426

PR #4432, opened 2026-07-14
Record proposed Voronovskaja formula for Bernstein-Bezier operators

SamuelSchlesinger:
## What changed

- defines the standard Gaussian CDF and the proposed Bernstein-Bezier bias term
- replaces four unspecified `answer(sorry)` values with a single explicit asymptotic formula
- records the proposed sufficient and sharp order for the case `m = 1`

## Status

This is deliberately a draft proposal for mathematical review. It does **not** claim a formal proof or field acceptance: the declarations remain `research open`, carry no `formal_proof` annotation, and their theorem bodies remain `sorry`.

The proposed formula comes from expressing the powered Bernstein tails through binomial probabilities and applying the central-limit transition near the evaluation point. The draft records the exact candidate answer consistently across the related declarations so it can be reviewed and formalized.

## Validation

- `lake --wfail build` passed on the reviewed combined branch (8,869 jobs)
- the changed file compiles directly with `lake env lean`
- the expected warnings are only the existing open-problem `sorry` placeholders

### DIFF
--- a/FormalConjectures/Paper/VoronovskajaTypeFormula.lean
+++ b/FormalConjectures/Paper/VoronovskajaTypeFormula.lean
@@ -93,6 +93,175 @@ noncomputable def bezierBernstein (n : ℕ) (α : ℝ) (f : ℝ → ℝ) (x : �
   ∑ k ∈ Finset.range (n + 1),
     f (k / n) * ((bernsteinTail n k).eval x ^ α - (bernsteinTail n (k + 1)).eval x ^ α)
 
+/-- The zeroth Bernstein tail is the constant polynomial one. -/
+@[category API, AMS 26 40 47]
+lemma bernsteinTail_zero (n : ℕ) : bernsteinTail n 0 = 1 := by
+  rw [bernsteinTail, ← Nat.range_succ_eq_Icc_zero]
+  exact bernsteinPolynomial.sum ℝ n
+
+/-- The Bernstein tail after the final index vanishes. -/
+@[category API, AMS 26 40 47]
+lemma bernsteinTail_succ_self (n : ℕ) : bernsteinTail n (n + 1) = 0 := by
+  simp [bernsteinTail]
+
+@[category API, AMS 26 40 47]
+private lemma bernsteinTail_eq_add {n k : ℕ} (hk : k ≤ n) :
+    bernsteinTail n k = bernsteinPolynomial ℝ n k + bernsteinTail n (k + 1) := by
+  rw [bernsteinTail, bernsteinTail, ← Finset.insert_Icc_add_one_left_eq_Icc hk,
+    Finset.sum_insert]
+  simp
+
+@[category API, AMS 26 40 47]
+private lemma bernsteinPolynomial_eval_nonneg {n k : ℕ} {x : ℝ} (hx : x ∈ I) :
+    0 ≤ (bernsteinPolynomial ℝ n k).eval x := by
+  rw [bernsteinPolynomial]
+  simp only [eval_mul, eval_natCast, eval_pow, eval_X, eval_sub, eval_one]
+  rcases hx with ⟨hx0, hx1⟩
+  exact mul_nonneg (mul_nonneg (Nat.cast_nonneg _) (pow_nonneg hx0 _))
+    (pow_nonneg (sub_nonneg.mpr hx1) _)
+
+/-- Evaluation identifies `bernsteinTail` with the usual binomial tail polynomial. -/
+@[category API, AMS 26 40 47]
+lemma bernsteinTail_eval_eq_binomial_tail (n k : ℕ) (x : ℝ) :
+    (bernsteinTail n k).eval x =
+      ∑ j ∈ Finset.Icc k n, (n.choose j : ℝ) * x ^ j * (1 - x) ^ (n - j) := by
+  simp [bernsteinTail, Polynomial.eval_finset_sum, bernsteinPolynomial]
+
+@[category API, AMS 26 40 47]
+private lemma bernsteinTail_eval_nonneg {n k : ℕ} {x : ℝ} (hx : x ∈ I) :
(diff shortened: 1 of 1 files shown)


mo271:
@shamEiNew can you have a look if this is a reasonable answer?

FormalConjectures/Paper/VoronovskajaTypeFormula.lean
```diff
@@ -114,43 +132,51 @@ theorem voronovskaja_theorem.bernstein_operators
...
+*Reference:* [Abel's source problem](https://www.math.bas.bg/mathmod/Proceedings_CTF/CTF-2010/files_CTF-2010/Open_problems.pdf).
 -/
 @[category research open, AMS 26 40 47]
 theorem voronovskaja_theorem.bezier_bernstein_operators
     (α : ℝ) (hα_pos : 0 < α) (hα : α ≠ 1)
     (f : ℝ → ℝ) (x : ℝ) (hx : x ∈ I)
     (hf : ContDiffOn ℝ 2 f I) :
     Tendsto (fun n : ℕ => Real.sqrt n * (bezierBernstein n α f x - f x)) atTop
-      (𝓝 answer(sorry)) := by
+      (𝓝 answer(bezierBias α * Real.sqrt (x * (1 - x)) *
```
shamEiNew: i guess its better to revert answer(sorry) and record the conjectured formula in the docstring or a comment? If the answer turns out to be wrong, these statements become false conjectures rather than open questions, which changes the meaning of the file.

FormalConjectures/Paper/VoronovskajaTypeFormula.lean
```diff
@@ -114,43 +132,51 @@ theorem voronovskaja_theorem.bernstein_operators
...
+$\alpha > 0$, $\alpha \neq 1$:
+\[
+\sqrt n\,(B_{n,\alpha}f(x)-f(x))
+\longrightarrow
+\mu_\alpha\sqrt{x(1-x)}\,f'(x).
+\]
+
+The source asks for sufficiently smooth functions. This concrete version keeps
+the existing formalization's $C^2$ baseline, although the proposed argument
+only requires $C^1$ regularity.
```
shamEiNew: the original source claims sufficient smoothness and this C1 might be correct based on heuristics but requires a proof. So better we go with original assumption

FormalConjectures/Paper/VoronovskajaTypeFormula.lean
```diff
@@ -93,6 +96,21 @@ noncomputable def bezierBernstein (n : ℕ) (α : ℝ) (f : ℝ → ℝ) (x : 
   ∑ k ∈ Finset.range (n + 1),
     f (k / n) * ((bernsteinTail n k).eval x ^ α - (bernsteinTail n (k + 1)).eval x ^ α)
 
+/-- The distribution function $\Phi$ of a standard normal random variable. -/
+noncomputable def standardGaussianCDF (t : ℝ) : ℝ :=
```
shamEiNew: probably could be stated using `mathlib`

FormalConjectures/Paper/VoronovskajaTypeFormula.lean
```diff
@@ -42,13 +42,16 @@ asymptotic formula
     = \tfrac{1}{2} x(1-x) f''(x).
 \]
 
-## Known Results
+## Known Results and Proposed Answer
```
shamEiNew: remove  proposed answer part, as this is again a claim with no proof
```suggestion
## Known Results
```

shamEiNew:
@mo271 i've checked the proposed limit and it looks good but there is no proof to assert it. Instead its better it kept as proposed solution to maintain the original conjecture.

SamuelSchlesinger:
Thanks for the review. I've made the requested revisions:

- Restored `answer(sorry)` in the open-question declarations, so the file continues to record the original unanswered problem rather than asserting the proposed limit.
- Moved the candidate formula out of the open theorem answers and into the main theorem's docstring, explicitly labeled as a proposed answer for future investigation.
- Removed the unsupported claim that `C^1` regularity suffices and restored the source's "sufficiently smooth" formulation, with the existing `C^2` baseline in the concrete theorem.
- Removed the local `standardGaussianCDF` definition and now use Mathlib's `cdf (gaussianReal 0 1)` directly.

The proved algebraic reductions and endpoint case remain, since they do not assert the unknown interior limit.

shamEiNew:
@SamuelSchlesinger thanks for the updates, it looks fine now.

mo271 (commented):
Thanks! Looks great in general, but I would propose to add an addtional `research open` statement with that concrete proposed solution: if it can be proven great, if it can be disproven, also great!

SamuelSchlesinger:
Ack, fixed in 28d01553. I added a separate `research open` statement for the concrete proposed asymptotic formula, while leaving the original open-question statement unchanged.

FormalConjectures/Paper/VoronovskajaTypeFormula.lean
```diff
@@ -166,6 +402,8 @@ theorem voronovskaja_theorem.bezier_bernstein_operators.variants.eventually_smoo
 Variant of the Bézier-Bernstein Voronovskaja problem with the required smoothness order itself
 left as an answer. Replacing `(answer(sorry) : ℕ × ((ℝ → ℝ) → ℝ → ℝ))` by a concrete value lets one
 state the conjecture for a chosen regularity threshold.
+
+*Reference:* [Abel's source problem](https://www.math.bas.bg/mathmod/Proceedings_CTF/CTF-2010/files_CTF-2010/Open_problems.pdf).
```
mo271, line 406: instead of repeating the reference often, just put it once on the top

SamuelSchlesinger: Done.

mo271 (approved):
Thanks, LGTM!

PR #4433, opened 2026-07-14
refactor: split out FormalConjectures.Util to a new library

YaelDillies:
... named FormalConjecturesUtil.

The goal here is to make all files in FormalConjectures be valid challenge files for Comparator.


mo271 (approved):
LGTM (but doesn't build yet..)

FormalConjecturesUtil.lean
eric-wieser, line 1: I would argue that we ought to deprecate `Util/ProblemImports` rather than deleting it entirely.

YaelDillies: Ah yes, good point

Issue #1685, opened 2026-01-16
Green's Open Problems #65


rao107:
### What is the conjecture
https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.65

Is there $c > 0$ with the following property: whenever $A \subset [N]$ is a set of size $N^{1-c}$, $A - A$ contains a nonzero square? What about $A - A$ containing a prime minus one?

### [AMS categories](https://github.com/google-deepmind/formal-conjectures/labels?q=ams-)

* ams-11
* ams-05

### Choose either option
- [ ] I plan on adding this conjecture to the repository
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

Robertboy18:
I can work on this.

AItoBit:
I am formalising this and will open a pull request.

mo271:
@AItoBit there is already a pull reqeust open for it, perhaps you could give it a review instead?

AItoBit:
Where can I find the code to review? @mo271 
offtopic: where can I find the unsolved open problems if  I've checked and there are "open"problems in the website which are already in the repo solved? thanks

mo271:
#4448

AItoBit:
@mo271 done, thanks

AItoBit:
Submitted a formalisation in https://github.com/google-deepmind/formal-conjectures/pull/6445 (current module header, both questions from the source, plus Ruzsa and the 100-fold variant). Related: #4448.

PR #4448, opened 2026-07-17
feat(GreensOpenProblems): formalize problem 65

Robertboy18:
Formalizes Green's Open Problem 65:)!

Here I add the:
- the nonzero-square difference-set question;
- the prime-minus-one variant;
- helper definitions for the integer interval, signed difference set, and shared large-difference-pattern shape.

It is related to this issue #1685! The build passes.

### DIFF
--- a/FormalConjectures/GreensOpenProblems/65.lean
+++ b/FormalConjectures/GreensOpenProblems/65.lean
@@ -1,0 +1,75 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# Ben Green's Open Problem 65
+
+*Reference:* [Ben Green's Open Problem 65](https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.65)
+-/
+
+open Filter
+
+namespace Green65
+
+/-- An integer is a nonzero square if it is the square of a nonzero natural number. -/
+def IsNonzeroSquare (d : ℤ) : Prop :=
+  ∃ n : ℕ, n ≠ 0 ∧ d = (n : ℤ) ^ 2
+
+/-- An integer is one less than a prime. -/
+def IsPrimeMinusOne (d : ℤ) : Prop :=
+  ∃ p : ℕ, p.Prime ∧ d = (p : ℤ) - 1
+
+/-- The interval `[N] = {1, ..., N}`, viewed as a finite set of integers. -/
+def interval (N : ℕ) : Finset ℤ :=
+  Finset.Icc (1 : ℤ) (N : ℤ)
+
+/-- The signed difference set `A - A`. -/
+def differenceSet (A : Finset ℤ) : Set ℤ :=
+  {d | ∃ a ∈ A, ∃ b ∈ A, d = a - b}
+
+/--
+There is a fixed density exponent `c > 0` such that every sufficiently large subset
+`A ⊆ [N]` with `|A| ≥ N^(1-c)` has a difference satisfying `P`.
+
+We model `[N]` as the integer interval `{1, ..., N}` and use sufficiently large `N`, since the
+literal small-`N` version is false for the nonzero-square question.
+-/
+def LargeDifferencePattern (P : ℤ → Prop) : Prop :=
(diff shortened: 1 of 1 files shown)


williamjblair:
Checked against the source PDF: Problem 65 asks both halves, so headline plus variant is right, and Green's "a set of size `N^(1-c)`" reads as "at least", which is what `≤` gives. Builds clean.

The one departure from the literal problem is `∀ᶠ N in atTop`, and the docstring asserts it's necessary without backing it. It is:

```lean
/-- Without "sufficiently large `N`" the question is false for every `c`:
`A = {1} ⊆ [1]` has `|A| = 1 = 1 ^ (1 - c)` and `A - A = {0}`. -/
@[category test, AMS 5 11]
theorem not_forall_of_isNonzeroSquare :
    ¬ ∃ c : ℝ, 0 < c ∧ c < 1 ∧ ∀ N : ℕ, ∀ A : Finset ℤ,
      A ⊆ interval N → (N : ℝ) ^ (1 - c) ≤ (A.card : ℝ) →
        ∃ d ∈ differenceSet A, IsNonzeroSquare d := by
  rintro ⟨c, -, -, h⟩
  obtain ⟨d, ⟨a, ha, b, hb, rfl⟩, n, hn, hdn⟩ :=
    h 1 {1} (by decide) (by norm_num [Real.one_rpow])
  simp only [Finset.mem_singleton] at ha hb
  subst ha; subst hb
  rw [eq_comm] at hdn
  simp at hdn
  exact hn hdn
```

Compiles as written. The file has no tests at present.

FormalConjectures/GreensOpenProblems/65.lean
```diff
@@ -0,0 +1,92 @@
...
+def interval (N : ℕ) : Finset ℤ :=
+  Finset.Icc (1 : ℤ) (N : ℤ)
+
+/-- The signed difference set `A - A`. -/
+def differenceSet (A : Finset ℤ) : Set ℤ :=
+  {d | ∃ a ∈ A, ∃ b ∈ A, d = a - b}
+
+/--
+There is a fixed density exponent `c > 0` such that every sufficiently large subset
+`A ⊆ [N]` with `|A| ≥ N^(1-c)` has a difference satisfying `P`.
```
mo271: please use latex docstring here and everywhere else

FormalConjectures/GreensOpenProblems/65.lean
```diff
@@ -0,0 +1,92 @@
...
+-/
+@[category research open, AMS 5 11]
+theorem green_65 :
+    answer(sorry) ↔ LargeDifferencePattern IsNonzeroSquare := by
+  sorry
+
+/--
+Green asks the analogous question with "nonzero square" replaced by "prime minus one".
+-/
+@[category research open, AMS 5 11]
```
mo271: is this really open? I thought Green mentiones some results in the update?

FormalConjectures/GreensOpenProblems/65.lean
Robertboy18: Yep, fixed — I switched the mathematical notation in the docstrings to `$...$` throughout the file.

FormalConjectures/GreensOpenProblems/65.lean
Robertboy18: Good catch, it is not open. Green’s Theorem 1.1 gives a power-saving bound for sets avoiding differences of the form $p - 1$; shrinking the exponent slightly absorbs the implied constant for sufficiently large $N$, so it implies this eventual formulation. I added the reference and marked the variant solved.

Robertboy18:
-awaiting-author

AItoBit (approved):
## Verdict

Approve. The Lean matches Green’s Problem 65, and the earlier review comments are addressed.

## Statements vs source

Green asks: is there \(c>0\) such that every \(A\subset[N]\) of size \(N^{1-c}\) has a nonzero square in \(A-A\)? Same question for \(p-1\).

- `green_65` is the square question, tagged `research open` with `answer(sorry)`. Correct.
- `green_65.variants.prime_minus_one` is the shifted-prime question, tagged `research solved` with `answer(True)`. Matches Green 2022, Theorem 1.1. The \(\ll N^{1-c}\) bound implies this eventual form after shrinking \(c\).
- `|A|\ge N^{1-c}` is the right reading of “a set of size \(N^{1-c}\)”.
- `∀ᶠ N in atTop` is necessary: for \(N=1\), \(A=\{1\}\) has size \(1^{1-c}\) and \(A-A=\{0\}\). `not_forall_of_isNonzeroSquare` proves this.
- `0<c<1` does not change the yes/no answer: no \(c\ge 1\) can work.

## Checks

- Categories, AMS 5 11, namespace, copyright, and `answer(sorry)` / `answer(True)` are in order.
- The small-\(N\) test is short and useful.
- CI was green on the last push (Aug 14).

## Pre-merge nits (not statement bugs)

Current `main` problem files start with `module`, `public import FormalConjecturesUtil`, and `@[expose] public section`. This file still uses a pre-module `import`. Please update the header so a re-run of CI on current `main` stays green.

Optional later: `differenceSet` is `A - A` under `open scoped Pointwise`; Green’s comments also mention Ruzsa’s obstruction and the \(100A-100A\) variant. Those can be follow-ups.

Fixes the two questions in #1685.

AItoBit (approved):

PR #4449, opened 2026-07-17
Formalize MOLS existence problem

Robertboy18:
This PR formalizes the MOLS existence problem using the existing `LatinSquare n` infrastructure!

  It adds:
  - `LatinSquare.Orthogonal`
  - `LatinSquare.MutuallyOrthogonal`
  - `LatinSquare.HasCompleteMOLS`

  The main statement, `molsExistenceProblem`, asks for the set of orders `n` that admit a complete family of `n - 1` mutually orthogonal Latin squares. I also included`molsOrder12`, since order 12 is the smallest unresolved case mentioned in the issue:)!

### DIFF
--- a/FormalConjectures/Paper/LatinSquare.lean
+++ b/FormalConjectures/Paper/LatinSquare.lean
@@ -33,6 +33,24 @@ namespace LatinSquare
 variable {n : ℕ}
 
 /--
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4457, opened 2026-07-18
Mark WOWII graph conjecture 142 solved

AlperTheKing:
## Summary

- mark WOWII / Graffiti.pc Conjecture 142 as solved
- link the immutable Lean proof commit in the formal proof annotation
- prove the stronger integral inequality `ecc(B) + girth - girth / 3 ≤ tree`, then derive the stated real-valued bound

## Verification

- `lake env lean -DwarningAsError=true FormalConjecturesForMathlib/WrittenOnTheWallII/GraphConjecture142Proof.lean`
- clean build of `FormalConjectures.WrittenOnTheWallII.GraphConjecture142`
- proof module contains no `sorry`, `admit`, or `native_decide`
- axiom audit reports only `propext`, `Classical.choice`, and `Quot.sound`

Formal proof commit: https://github.com/AlperTheKing/formal-conjectures/commit/46bf39015f5c3c3ba3bfcf9f752b4b1e49b584ac

### DIFF
--- a/FormalConjectures/WrittenOnTheWallII/GraphConjecture142.lean
+++ b/FormalConjectures/WrittenOnTheWallII/GraphConjecture142.lean
@@ -38,6 +38,41 @@ where $\mathrm{tree}(G)$ is the largest induced tree size, $\mathrm{girth}(G)$
 is the length of the shortest cycle ($0$ if acyclic), $B$ is the set of
 boundary vertices (those of maximum eccentricity), and $\mathrm{ecc}(B)$ is
 the eccentricity of the set $B$.
+
+**Proof sketch.** The proof first strengthens the real-valued target to the
+integral bound
+$$
+\mathrm{ecc}(B) + \mathrm{girth}(G)
+  - \lfloor \mathrm{girth}(G) / 3 \rfloor \leq \mathrm{tree}(G).
+$$
+For a cyclic graph, start with a shortest cycle $K$. It is chordless, so
+removing one vertex leaves an induced path on $\mathrm{girth}(G)-1$ vertices.
+Shortest paths from selected vertices to $K$ are called *descents*. Pairwise
+noninteracting descents attach to the retained cycle path as disjoint branches.
+If two descents interact, the proof cuts at the first interaction encountered
+from the outer endpoint toward the cycle and splices the paths there.
+Minimality of that interaction makes the retained prefix disjoint and
+nonadjacent to the first descent; geodesicity excludes chords, while girth at
+least five rules out extra cross-edges and cycle attachments.
+
+Choose a vertex realizing $\mathrm{ecc}(B)$ and two endpoints of a diametral
+geodesic. Either two of their descents interact, directly giving a sufficiently
+long spliced tree, or all three are separate, in which case a three-point
+distance inequality forces their total length to be large enough. This proves
+the integral bound in the main range. Girths three and four, small
+$\mathrm{ecc}(B)$, and the two remaining congruence cases are handled by
+separate geodesic/cycle certificates with one or two descents; the acyclic case
+follows from a diametral path. Finally,
+$$
+\mathrm{girth}(G)-\lfloor\mathrm{girth}(G)/3\rfloor
+  = \lceil 2\,\mathrm{girth}(G)/3\rceil
+  \geq 2\,\mathrm{girth}(G)/3,
+$$
+which gives the stated real-valued inequality.
+
+The argument was developed with assistance from ChatGPT Pro. The linked Lean 4
+formalization was produced with assistance from OpenAI Codex and checked by
+Lean.
 -/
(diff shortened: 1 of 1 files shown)


FormalConjectures/WrittenOnTheWallII/GraphConjecture142.lean
```diff
@@ -39,7 +39,8 @@ is the length of the shortest cycle ($0$ if acyclic), $B$ is the set of
 boundary vertices (those of maximum eccentricity), and $\mathrm{ecc}(B)$ is
 the eccentricity of the set $B$.
 -/
-@[category research open, AMS 5]
+@[category research solved, AMS 5,
+  formal_proof using formal_conjectures at "https://github.com/AlperTheKing/formal-conjectures/blob/46bf39015f5c3c3ba3bfcf9f752b4b1e49b584ac/FormalConjectures/WrittenOnTheWallII/GraphConjecture142.lean#L44"]
```
mo271: looks like the linked proof is
```
theorem conjecture142 (G : SimpleGraph α) [DecidableRel G.Adj] (h : G.Connected) :
    let B : Set α := (maxEccentricityVertices G : Set α)
    (2 : ℝ) / 3 * (G.girth : ℝ) + (eccSet G B : ℝ) ≤ (largestInducedTreeSize G : ℝ) := by
  exact FormalProofs.WrittenOnTheWallII.GraphConjecture142.conjecture142 G h
```

Does this compile for you?! it seems like it is using the theorem it is trying to prove in the proof??

AlperTheKing: Good catch — you're right that the pinned annotation URL was wrong: it pointed to the public wrapper and therefore made the proof look circular. Sorry about that. I fixed the annotation in [`60e4ec9`](https://github.com/google-deepmind/formal-conjectures/pull/4457/commits/60e4ec91a1d7a89d4f94216ec9655f9a3845472c), so it now points directly to the proof theorem:

https://github.com/AlperTheKing/formal-conjectures/blob/46bf39015f5c3c3ba3bfcf9f752b4b1e49b584ac/FormalConjecturesForMathlib/WrittenOnTheWallII/GraphConjecture142Proof.lean#L3927-L3932

There is no self-reference. The two declarations with the short name `conjecture142` are different fully qualified constants:

- target: `WrittenOnTheWallII.GraphConjecture142.conjecture142`
- proof: `FormalProofs.WrittenOnTheWallII.GraphConjecture142.conjecture142`

The proof module does not import the target module. Its proof chain derives `SimpleGraph.conjecture142_real_splice` at lines 3896–3916 from the integral strengthening at lines 3866–3892: [proof chain](https://github.com/AlperTheKing/formal-conjectures/blob/46bf39015f5c3c3ba3bfcf9f752b4b1e49b584ac/FormalConjecturesForMathlib/WrittenOnTheWallII/GraphConjecture142Proof.lean#L3866-L3932).

I rechecked commit `46bf39015f5c3c3ba3bfcf9f752b4b1e49b584ac`: `lake env lean` on the 3,934-line proof file and on the target wrapper separately both exit 0. The proof file contains no `sorry`, `admit`, `axiom`, `native_decide`, or `unsafe`; `#print axioms FormalProofs.WrittenOnTheWallII.GraphConjecture142.conjecture142` reports only `[propext, Classical.choice, Quot.sound]`.

Thanks for flagging this.

FormalConjectures/WrittenOnTheWallII/GraphConjecture142.lean
```diff
@@ -39,7 +39,8 @@ is the length of the shortest cycle ($0$ if acyclic), $B$ is the set of
 boundary vertices (those of maximum eccentricity), and $\mathrm{ecc}(B)$ is
 the eccentricity of the set $B$.
```
mo271, line 40: please add some comment on how the proof was obtained and how it roughly works (an informal sketch...)

AlperTheKing: Added this in [`3c2ab13`](https://github.com/google-deepmind/formal-conjectures/pull/4457/commits/3c2ab13bc1a10bd701dd9a5301e198009ec8159a). The theorem docstring now sketches the stronger integral bound and its proof: delete a vertex of a shortest cycle, attach noninteracting geodesic descents, splice interacting descents at the first outer-to-cycle interaction, use an eccentricity realizer together with a diametral pair and the three-point estimate, and then close the small-girth and residue cases separately before casting to `ℝ`. I also added an explicit ChatGPT Pro / OpenAI Codex assistance disclosure.

The targeted `lake --wfail build FormalConjectures.WrittenOnTheWallII.GraphConjecture142` succeeds (8038/8038). Thanks for the suggestion.

Danie-I (approved):
Congratulations on the solution. LGTM.

PR #4467, opened 2026-07-19
Create 934.lean

### LINKED ISSUE #1026: Erdős Problem 934
### What is the conjecture

https://www.erdosproblems.com/934

  Let $h_t(d)$ be minimal such that every graph $G$ with $h_t(d)$ edges and maximal degree $\leq d$ contains two edges whose shortest path between them has length $\geq t$.

Estimate $h_t(d)$.

Status:  open

### Choose either option

LezheGao:
Fixed #1026


mo271:
we don't want to have definitions with `sorry`

PR #4474, opened 2026-07-20
feat: add Jacobian disproof

Paul-Lez:
This PR adds the disproof of the Jacobian conjecture.

This makes a minor change to the statement, namely negating it (the original formalisation expected the conjecture to be true so couldn't be directly adapted by a disproof). In a subsequent PR (#4635), I'll update the statement so it aligns with the `answer(sorry)` convention that we often use in this repo, and state a disproof for all characteristics (see discussion and reviews below).

The tweet containing the original disproof can be found here: https://x.com/__alpoge__/status/2079028340955197566?s=46

Some independent formalisations were written by other people - see thread here https://leanprover.zulipchat.com/#narrow/channel/583339-AI-authored-projects/topic/Counterexample.20to.20the.20Jacobean.20conjecture/with/611589771. 

Many thanks to @deancureton for the golfing!

### DIFF
--- a/FormalConjectures/Wikipedia/JacobianConjecture.lean
+++ b/FormalConjectures/Wikipedia/JacobianConjecture.lean
@@ -132,8 +132,8 @@ theorem jacobian_conjecture {k : Type} [Field k] [CharZero k] :
 set_option linter.style.answer_attribute false in
 /-- Does the Jacobian conjecture hold in the two variable case? -/
 @[category research open, AMS 14]
-theorem jacobian_conjecture_two_variables {k : Type} [Field k] [CharZero k] :
-    answer(sorry) ↔ JacobianConjectureProp k (Fin 2) := by
+theorem jacobian_conjecture_two_variables :
+    answer(sorry) ↔ ∀ (k : Type) [Field k] [CharZero k], JacobianConjectureProp k (Fin 2) := by
   sorry
 
 end Conjecture
--- a/FormalConjectures/Wikipedia/JacobianConjecture.lean
+++ b/FormalConjectures/Wikipedia/JacobianConjecture.lean
@@ -52,14 +52,14 @@ noncomputable def id : RegularFunction k σ σ := MvPolynomial.X
 
 /-- The evaluation of a regular function `f` over `k` at some point `a`
 with coordinates in some algebra over `k`-/
-noncomputable def aeval {σ τ : Type*} [Fintype σ] {S₁ : Type*} [CommSemiring S₁] [Algebra k S₁]
+noncomputable def aeval {σ τ : Type*} {S₁ : Type*} [CommSemiring S₁] [Algebra k S₁]
     (F : RegularFunction k σ τ) : (σ → S₁) → τ → S₁ :=
   fun a t ↦ MvPolynomial.aeval a (F t)
 
 /--`aeval` is compatible with composition of regular functions. -/
 @[category API, AMS 14]
 lemma comp_aeval
-    {σ τ ι : Type*} [Fintype σ] [Fintype τ]
+    {σ τ ι : Type*}
     (F : RegularFunction k σ τ) (G : RegularFunction k τ ι)
     (a : σ → k) : (F.comp G).aeval a = G.aeval (F.aeval a) := by
   ext i
@@ -70,11 +70,9 @@ end RegularFunction
 
 end Prelims
 
-variable {k : Type*} [Field k] [CharZero k]
-
 section Conjecture
 
-open RegularFunction
+open RegularFunction MvPolynomial
 
 def JacobianConjectureProp (k σ : Type) [CommRing k] [Fintype σ] [DecidableEq σ] :
     Prop :=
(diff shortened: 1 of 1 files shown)


Paul-Lez:
~~This is currently a little messy, but should be highly golfable.~~
(Now cleaned up!)

bkim23 (approved):

FormalConjectures/Wikipedia/JacobianConjecture.lean
```diff
@@ -52,27 +50,88 @@ noncomputable def comp
...
+    F.comp G = id k σ
+
+/-- Alpöge/Fable's counterexample: a polynomial self-map of `k³` with Jacobian
+determinant `1` which is not injective. -/
+noncomputable def F (k : Type) [Field k] : RegularFunction k (Fin 3) (Fin 3) :=
+  ![(1 + C 2 * X 0 * X 1) ^ 3 * X 2
+      + C 4 * X 1 ^ 2 * (1 + C 2 * X 0 * X 1) * (C 2 + C 3 * (X 0 * X 1)),
+    X 1 + C 3 * X 0 * (1 + C 2 * X 0 * X 1) ^ 2 * X 2
+      + C 12 * X 0 * X 1 ^ 2 * (C 2 + C 3 * (X 0 * X 1)),
+    -X 0 + C 3 * X 0 ^ 2 * X 1 + X 0 ^ 3 * X 2]
```
alreadydone: [`name_poly_vars`](https://github.com/leanprover-community/mathlib4/blame/4cb03f453732be00b43b623661457188c10c1149/Mathlib/Tactic/Ring/NamePolyVars.lean#L20) has been there since one year ago. Are you unaware or unable to use it?

deancureton:
thanks @alreadydone for the pointer! was unaware of `name_poly_vars`

FormalConjectures/Wikipedia/JacobianConjecture.lean
```diff
@@ -52,27 +50,90 @@ noncomputable def comp
...
+  intro h
+  obtain ⟨G, -, hFG⟩ := h (F k) (det_jacobian_F k ▸ isUnit_one)
+  have hleft : Function.LeftInverse (G.aeval (S₁ := k)) ((F k).aeval) := fun a => by
+    rw [← RegularFunction.comp_aeval, hFG]
+    funext t
+    simp [RegularFunction.aeval, RegularFunction.id]
+  have h1 : (1 : k) = -1 := congrFun (hleft.injective (aeval_F_eq k)) 0
+  norm_num at h1
+
+set_option linter.style.answer_attribute false in
```
mo271:
```suggestion
```
why is this needed?

mo271 (approved):
Thanks, LGTM just one question about the linter

FormalConjectures/Wikipedia/JacobianConjecture.lean
deancureton: you can get rid of both occurrences of this line:

```lean
theorem jacobian_conjecture :
    answer(False) ↔ ∀ (k : Type) [Field k] [CharZero k] {σ : Type} [Fintype σ] [DecidableEq σ],
      JacobianConjectureProp k σ := by
  rw [false_iff]
  intro h
  obtain ⟨G, -, hFG⟩ := h ℚ (F ℚ) (det_jacobian_F ℚ ▸ isUnit_one)
  have hleft : Function.LeftInverse (G.aeval (S₁ := ℚ)) ((F ℚ).aeval) := fun a => by
    rw [← RegularFunction.comp_aeval, hFG]
    funext t
    simp [RegularFunction.aeval, RegularFunction.id]
  have h1 : (1 : ℚ) = -1 := congrFun (hleft.injective (aeval_F_eq ℚ)) 0
  norm_num at h1
```

(moving binders and replacing `k` with `ℚ` everywhere in the proof), and

```lean
theorem jacobian_conjecture_two_variables :
    answer(sorry) ↔ ∀ (k : Type) [Field k] [CharZero k], JacobianConjectureProp k (Fin 2) := by
```

(moving binders in the statement only)

i won't open a PR in case the original style was intentional, but this might be cleaner!

Paul-Lez: This was intentional - the idea was that the counter example works directly for any field of characteristic zero

Paul-Lez: Arguably for the unsolved problem (the two variable case) we should actually allow things to depend on the field. I'll add a commit for that tomorrow morning!

DrAlexHarrison:
This work was performed by Claude (Fable 5) under my direction. Fable also helped draft most of the following.

Two pointers that may be useful alongside this PR:

1. **An independent audit of [deancureton/jacobian](https://github.com/deancureton/jacobian)**: it proves `not_jacobianConjecture_all_char` (every field, including char 2 via a rescaled det ≡ 1 form). We audited it on our machine: `lake build` succeeds on the pinned toolchain, no `sorry`/`native_decide`/custom axioms, and `#print axioms` reports exactly `[propext, Classical.choice, Quot.sound]` for every theorem. 
**Full audit:**
     https://github.com/DrAlexHarrison/jacobian-anatomy/blob/main/audit/deancureton-audit.md

  2. **A reproducible computational suite** for the map's geometry (Jacobian certificates three independent ways, fiber structure, monodromy S₃, non-properness locus, exact image)
  is here, one command:
     https://github.com/DrAlexHarrison/jacobian-anatomy — includes an independent second Lean certificate (`JC.lean`) you're welcome to golf against.

Fable says it's happy to help reconcile statements if useful. lol. I certainly am incapable of that. 

Happy to donate tokens if this is actually useful information and contribution to the community. If unwanted, safely ignore.

FormalConjectures/Wikipedia/JacobianConjecture.lean
```diff
@@ -52,27 +50,90 @@ noncomputable def comp
...
+/-- Does the Jacobian conjecture hold in the two variable case? -/
 @[category research open, AMS 14]
-theorem jacobian_conjecture (F : RegularFunction k σ σ)
-    (H : IsUnit F.Jacobian.det) :
-    ∃ (G : RegularFunction k σ σ), G.comp F = id k σ ∧
-    F.comp G = id k σ := by
+theorem jacobian_conjecture_two_variables {k : Type} [Field k] [CharZero k] :
+    answer(sorry) ↔ JacobianConjectureProp k (Fin 2) := by
```
Paul-Lez:
```suggestion 
theorem jacobian_conjecture_two_variables :
    answer(sorry) ↔ ∀ (k : Type) [Field k] [CharZero k], JacobianConjectureProp k (Fin 2) := by
```

kbuzzard:
I think it is a little surprising that the selling point of this repo is supposed to be "we write down the conjectures so that you don't have to", but then this PR which resolves one of the conjectures also changes the statement of the conjecture.

FormalConjectures/Wikipedia/JacobianConjecture.lean
```diff
@@ -52,27 +50,89 @@ noncomputable def comp
...
+    ∃ (G : RegularFunction k σ σ), G.comp F = id k σ ∧
+    F.comp G = id k σ
+
+variable (k : Type) [Field k]
+
+name_poly_vars X, Y, Z over k
 
-variable {σ : Type*} [Fintype σ]
+/-- Alpöge/Fable's counterexample: a polynomial self-map of `k³` with Jacobian
+determinant `1` which is not injective. -/
```
kbuzzard: Alpöge's example did not have determinant 1. Is this example correctly attributed? I thought it was someone else who came up with the det=1 example? (maybe I'm wrong, I can't remember where I saw the det=1 example, but Alpöge had det=+-2)

mo271:
> I think it is a little surprising that the selling point of this repo is supposed to be "we write down the conjectures so that you don't have to", but then this PR which resolves one of the conjectures also changes the statement of the conjecture.

We often do exactly that, i.e. a new proof found just fills a `sorry`, not changing the statement at all. In some cases, for instance when a more general statement is proved or a weaker statement disproved we change the statement.
In this case we didn't use the `answer(sorry)` mechanism to start with, so in order to do the negation of the statement it needs to be changed.
It would be nice to have a proof of the statement where the only change is the negation (potentially moving quantifiers), and then in a follow up change refactor the statement to improve it if necessary. 
Here is happens to be the case that @Paul-Lez is both the author of the original statement and the person adding the formalisation of the disproof. It would be good to explain what changed in the statement and why this is needed/an improvement in the commit message.

FormalConjectures/Wikipedia/JacobianConjecture.lean
alreadydone: Should not this not say something like "regular functions from `k ^ σ` to `k ^ τ`?

FormalConjectures/Wikipedia/JacobianConjecture.lean
```diff
@@ -52,27 +50,89 @@ noncomputable def comp
...
+    pderiv_pow, pderiv_C, pderiv_X_self, pderiv_X_of_ne, ne_eq, Fin.reduceEq, not_false_eq_true]
+  simp only [map_ofNat]
+  ring
+
+/-- `F` identifies the two distinct points `(1, -3/4, 13/4)` and `(-1, 3/4, 13/4)`. -/
+@[category API, AMS 14]
+lemma aeval_F_eq [CharZero k] :
+    (F k).aeval ![1, -(3 / 4), (13 / 4 : k)] = (F k).aeval ![-1, 3 / 4, 13 / 4] := by
+  funext i
+  fin_cases i <;> simp [RegularFunction.aeval, F] <;> field_simp <;> ring
```
kbuzzard:
```suggestion
/-- `F` identifies the two distinct points `(1, 0, 1)` and `(0, 3, -71)`. -/
@[category API, AMS 14]
lemma aeval_F_eq :
    (F k).aeval ![1, 0, (1 : k)] = (F k).aeval ![0, 3, -71] := by
  funext i
  fin_cases i <;> simp [RegularFunction.aeval, F]; field_simp; ring
```
This breaks a proof below, but why not have a counterexample which works for all fields? (I don't know a reference for this, I just asked Claude to find an integer counterexample which works for all fields and it found this -- note that the points are distinct mod p for all p so this works for all fields)

FormalConjectures/Wikipedia/JacobianConjecture.lean
deancureton: After I posted a formalization of his example, [Alpöge gave me a hint](https://x.com/__alpoge__/status/2079061273237492157) that a simple rescaling would give determinant 1 to work in all characteristics. I found a rescaling and chose to use it here. In that sense the attribution isn't exactly correct (I'd be happy to fix), but I guess I thought it'd just count as the original, with a slight modification.

Paul-Lez: I'll make a few edits tonight to this PR - I'll add that clarification if you want (and link the extra tweet)

deancureton: sounds great, thanks!

FormalConjectures/Wikipedia/JacobianConjecture.lean
josephlr: After a quick search:
`F(-1, 1, 4) = F(0, 1, -8)  = (0,1,0)`
and
`F(1, 0, 1)  = F(0, 3, −71) = (1,3,0)`
seem like the minimal (w.r.t. L1-norm) integer counterexample points (up to sign).

I agree that using integer points for the counterexample is cleaner.

Paul-Lez:
Thanks all for the comments! 

Re changes in the statement:

Here three changes were made:
1) adding the `answer(False)`: the is one of the usual ways we deal with negations of statements
2) creating an auxiliary definition instead of inlining the statement (so it can be referred to and applied in other statements - directly using the theorem isn't possible anymore when you have the `answer(False)`)
3) removing `open Classical` and instead assuming decidable equality whenever needed.

All these things felt like very minor changes to the statement, so I felt like this was reasonable to have in this PR. If people disagree then happy to move the changes to another PR so they can be reviewed separately:). Generally I think we should be open to modifying statements when this is sensible, and I guess this is maybe a good opportunity to discuss the best procedure to do this! (IMO part of the value of this repo is that it allows us to document such changes when they occur, and is a good place for such discussions and reviews to be had!). 

Re counter examples that work in all characteristics: 

I've made a few local edits on my laptop to this branch without pushing just yet, so that we would have both:
1) the original counterexample with determinant -2 (the original content of this PR)
2) a variant suggested in the comments that works for all fields, and a proof of the negation of the Jacobian conjecture without the `CharZero` assumption. 
If we're sticking to the idea of not editing a statement in the PR that gives the proof, then I can open another PR with the content of part 2). 


I'll wait a bit for people (@kbuzzard, @deancureton, @mo271, and generally anyone else who wants to chime in) on whether they agree/disagree and then proceed with the splitting into two PRs.

FormalConjectures/Wikipedia/JacobianConjecture.lean
alreadydone: I thought you took the example [here](https://x.com/N8Programs/status/2079061392649302185?s=20) ...

FormalConjectures/Wikipedia/JacobianConjecture.lean
alreadydone: So `CommRing k` also works

FormalConjectures/Wikipedia/JacobianConjecture.lean
```diff
@@ -81,44 +141,24 @@ section Tests
...
-/--`aeval` is compatible with composition of regular functions. -/
-@[category API, AMS 14]
-lemma comp_aeval
-    (F : RegularFunction k σ τ) (G : RegularFunction k τ ι)
-    (a : σ → k) : (F.comp G).aeval a = G.aeval (F.aeval a) := by
-  ext i
-  rw [aeval, comp, MvPolynomial.aeval_bind₁, ←aeval]
-  rfl
-
+variable {k σ τ ι : Type} [Fintype σ] [Fintype τ] [DecidableEq σ] [DecidableEq τ] [Field k]
```
alreadydone:
```suggestion
variable {k σ : Type} [Fintype σ] [DecidableEq σ] [Field k]
```

FormalConjectures/Wikipedia/JacobianConjecture.lean
```diff
@@ -52,27 +50,89 @@ noncomputable def comp
...
+  norm_num at h1
+
+/-- Does the Jacobian conjecture hold in the two variable case? -/
 @[category research open, AMS 14]
-theorem jacobian_conjecture (F : RegularFunction k σ σ)
-    (H : IsUnit F.Jacobian.det) :
-    ∃ (G : RegularFunction k σ σ), G.comp F = id k σ ∧
-    F.comp G = id k σ := by
+theorem jacobian_conjecture_two_variables :
+    answer(sorry) ↔ ∀ (k : Type) [Field k] [CharZero k], JacobianConjectureProp k (Fin 2) := by
```
alreadydone: I think this is equivalent to the conjecture over C (the usual form), but it's not obvious: since any CharZero field is infinite, one can show the determinant is constant on the all points over the algebraic closure iff it's constant on the rational points, using [MvPolynomial.funext_set](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/MvPolynomial/Funext.html#MvPolynomial.funext_set). Then some model theory shows alg. closed char 0 fields behave the same way as C.

alreadydone (approved):

FormalConjectures/Wikipedia/JacobianConjecture.lean
deancureton: the example you linked is a different rescaling - the one below is from [here](https://github.com/deancureton/jacobian/blob/0d4a9212d874226ad81ce5a926becddfa94e6a88/Jacobian/Counterexample.lean#L74)

kbuzzard:
I've said all I have to say about this PR, I don't have any objections to it. In fact I learnt a little more about how this repo works as a result of this discussion, which is surely a good thing.

Paul-Lez:
Perf! Will make the edits and merge this PR as soon as I'm back from holidays (Sunday)

FormalConjectures/Wikipedia/JacobianConjecture.lean
Paul-Lez: Yeah that was what I was originally going for in a previous version of this statement. But I figured it would be nicer to have it this way in case someone ever comes up with a counterexample in some char zero that is not contained in C.

FormalConjectures/Wikipedia/JacobianConjecture.lean
Paul-Lez: Done in the following PR!

PR #4486, opened 2026-07-21
feat(JacobianConjecture): sorry-free refutation at n = 3 over ℚ (counterexample not yet peer-reviewed — see description)

techno-optimist:
## ⚠️ Read this first: the underlying counterexample is not yet peer-reviewed

This PR formalises a refutation that depends on a **very recent, externally-announced** counterexample:

- The Keller map is **Levent Alpöge's** (announced 2026-07-19, found with Claude Fable; question credited to Akhil). **It is not mine.**
- At the time of writing it is *"awaiting confirmation"* — widely verified informally, but **not peer-reviewed and not published**.
- **Maintainers may reasonably prefer to hold or decline this PR until the counterexample is confirmed.** I would rather flag that plainly up front than have it merged on the assumption that this is settled mathematics. If you'd like, treat this as a draft and I'm happy for it to sit until the situation clarifies.

If the map is ever retracted, this file proves nothing about the Jacobian Conjecture — only facts about three specific polynomials.

## What this adds

`FormalConjectures/Refutations/JacobianConjectureFalse.lean`: a `sorry`-free proof that the Jacobian Conjecture **as stated in this repo** is false for `k = ℚ`, `σ = Fin 3`.

The statement is copied **verbatim** from `FormalConjectures/Wikipedia/JacobianConjecture.lean` — same `RegularFunction`, `Jacobian`, `comp`, `id` — specialised to `ℚ` and `Fin 3`, so the goalposts are this repo's rather than mine:

```lean
theorem jacobian_conjecture_false :
    ¬ ∀ (F : RegularFunction ℚ (Fin 3) (Fin 3)), IsUnit F.Jacobian.det →
        ∃ G : RegularFunction ℚ (Fin 3) (Fin 3),
          G.comp F = RegularFunction.id ℚ (Fin 3) ∧
          F.comp G = RegularFunction.id ℚ (Fin 3)
```

## How it works

Two load-bearing facts, both machine-checked against mathlib's own `MvPolynomial.pderiv` and `Matrix.det`:

1. **`alpoge_jacobian_det`** — `det J_F = C (-2)`. Proved via nine explicit partial-derivative lemmas plus `ring`. (Computing the determinant in one `simp` does not terminate; splitting keeps `simp` off the degree-18 expression.)
2. **`alpoge_collision`** — `F` identifies `P = (0,0,-1/4)` and `Q = (1,-3/2,13/2)`.

The bridge is this repo's existing **`comp_aeval`** lemma: a regular right inverse forces the underlying point map to be injective, which the collision contradicts.

One small helper, `pderiv_ofNat_mv`, is needed because the constants `3` and `4` in the map are `OfNat` literals rather than `C _`, so `pderiv_C` does not fire on them (and `no_index` is required for `simp`'s discrimination tree to match). If you'd prefer that upstreamed to mathlib in a more general form rather than living here, happy to do that instead.

## Verification

```
lake build FormalConjectures.Refutations.JacobianConjectureFalse
```

builds clean (0 errors, 0 warnings) and the file ends with an axiom audit:

```
'JacobianConjecture.jacobian_conjecture_false' depends on axioms:
  [propext, Classical.choice, Quot.sound]
```

Lean's three standard axioms — **no `sorryAx`**.

## Deliberately *not* done

I did **not** touch `FormalConjectures/Wikipedia/JacobianConjecture.lean`. Its `jacobian_conjecture` is still `@[category research open]` with a `sorry`. Re-categorising a famous conjecture in your canonical file on the strength of a one-day-old unconfirmed announcement felt like your editorial call, not mine — I'm glad to follow up with whatever you decide.

## Housekeeping

- Uses your `@[category research solved, AMS 14]` / `@[category API, AMS 14]` conventions.
- Apache-2.0 header as per the repo.
- I understand a Google CLA is required; the signer will handle that via the CLA bot.

### DIFF
--- a/FormalConjectures/Refutations/JacobianConjectureFalse.lean
+++ b/FormalConjectures/Refutations/JacobianConjectureFalse.lean
@@ -1,0 +1,191 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjectures.Wikipedia.JacobianConjecture
+
+/-!
+# The Jacobian Conjecture is false (n = 3, over ℚ)
+
+CLAIM-TYPING — read before citing this file:
+
+* The counterexample map is **NOT ours**. It is Levent Alpöge's (announced 2026-07-19, found with
+  Claude Fable; question credited to Akhil). At the time of writing it is "awaiting confirmation":
+  widely verified, but not peer-reviewed. If the map were ever retracted this file would prove
+  nothing about the Jacobian Conjecture — only things about the specific polynomials below.
+* What this file contributes is a machine-checked proof, against the statement in
+  `FormalConjectures/Wikipedia/JacobianConjecture.lean` (restated verbatim below so the goalposts
+  are the community's, not ours), that the conjecture as stated is FALSE for `k = ℚ`, `σ = Fin 3`.
+* The two facts doing the work are `alpoge_jacobian_det` (`det J_F = C (-2)`) and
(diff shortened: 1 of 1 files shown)


danielchin:
There's already a WIP for the disproof of the Jacobian Conjecture in #4474

mo271:
closing, because the same counterexample is already being added in https://github.com/google-deepmind/formal-conjectures/pull/4474!
Thanks nonetheless! Feel free to comment on #4474 if there is anything from here which would improve that pr

PR #4455, opened 2026-07-18
Mark WrittenOnTheWallII.GraphConjecture314 as solved

glyaea:

PR #4496, opened 2026-07-21
Mark WOWII 314 solved

DomTheDeveloper:
Summary

Marks Written on the Wall II Graph Conjecture 314 as solved and links the complete immutable Lean 4 proof.

The theorem statement is unchanged. This PR only:

* changes the category from research open to research solved;
* adds a formal_proof using lean4 annotation pointing to the complete proof in the external Formal Conjectures fork.

Proof outline

The hypothesis that the largest induced path has at most four vertices excludes induced copies of (P_5).

The proof then classifies connected triangle-free (P_5)-free graphs into two cases:

1. a bipartite chain-graph case; or
2. a nonbipartite induced-(C_5) blow-up case.

In each case, the structure of the graph forces all minimal total dominating sets to have the same cardinality. Therefore the graph is well totally dominated, as required.

Formal verification

The external Lean development proves the exact current theorem

WrittenOnTheWallII.GraphConjecture314.conjecture314

without weakening or modifying its assumptions.

A focused isolated build and terminal axiom audit completed successfully. The proof does not depend on sorryAx.

Scope

The substantial proof remains in the external fork in accordance with the repository guidance for longer formal proofs. The upstream change is limited to the canonical GraphConjecture314.lean file.

AI assistance disclosure

The proof development and submission preparation used OpenAI GPT-5.6 Thinking. The mathematical argument and Lean artifact should receive normal maintainer review.

### DIFF
--- a/FormalConjectures/WrittenOnTheWallII/314.lean
+++ b/FormalConjectures/WrittenOnTheWallII/314.lean
@@ -79,7 +79,7 @@ argument and Lean formalization; all formal claims were checked by the pinned
 Lean compiler.
 -/
 @[category research solved, AMS 5,
-  formal_proof using lean4 at "https://github.com/DomTheDeveloper/formal-conjectures/blob/cf59008ef1cd432bf9803275dcf5d62ab1f094a3/FormalConjectures/WrittenOnTheWallII/GraphConjecture314.lean"]
+  formal_proof using lean4 at "https://github.com/DomTheDeveloper/formal-conjectures/blob/a5e35542007274589f7b88a1fd0244048abb7abd/FormalConjectures/WrittenOnTheWallII/314.lean"]
 theorem conjecture314 [Nontrivial α] (G : SimpleGraph α) [DecidableRel G.Adj]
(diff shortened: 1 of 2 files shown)


mo271 (commented):
Thanks! Congrats on solving the conjecture!

Can you add a paragraph outlining the proof informally in the docstring (like the one you have in the commit message), and potentially also something like "solved by [your name]  prompting [AI name]?

Danie-I:
From the informal proof I was not able to check the correctness, but the formalization of conjecture statement and the formal proof looks correct.

FormalConjectures/WrittenOnTheWallII/314.lean
mo271, line 1: I think the automatic merge is not working because of the rename?
let's keep `FormalConjectures/WrittenOnTheWallII/GraphConjecture314.lean`?

mo271: @DomTheDeveloper can we not move the file here?

mo271:
duplicate of #4455

PR #4503, opened 2026-07-21
feat(ErdosProblems/456): prove variants.infinitely_many_n via the witness family n = 2^(2j+3)

mkwatson:
Closes `erdos_456.variants.infinitely_many_n` using the witness family n = 2^k with k = 2j+3 odd.

For such n, m(n) <= 2n, since phi(2^(k+1)) = 2^k gives 2^k | phi(2^(k+1)). And p(n) > 2n, because the only candidate prime congruent to 1 mod 2^k that is at most 2n is 2^k + 1, which is divisible by 3 whenever k is odd and exceeds 3, hence composite. Together these give m(n) < p(n) for every n in the family, and j -> 2^(2j+3) is injective.

Sanity check at j = 0: n = 8, m(8) <= 16 and p(8) = 17, since 9 is composite.

Verified clean under the project linters (`lake build`, zero warnings). `#print axioms` on the target gives `[propext, Classical.choice, Quot.sound]`.

Assisted by OpenAI Codex; verified by me.

### DIFF
--- a/FormalConjectures/ErdosProblems/456.lean
+++ b/FormalConjectures/ErdosProblems/456.lean
@@ -111,7 +111,45 @@ Erdős [Er79e] writes it is 'easy to show' that for infinitely many $n$ we have 
 @[category research solved, AMS 11]
 theorem erdos_456.variants.infinitely_many_n :
     { n | m n < p n }.Infinite := by
-  sorry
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4511, opened 2026-07-21
Solve the N=6, D=3 integer quantum graph cases

infinityscroll:
## Summary

- Marks the N = 6, D = 3 integer equation system as having no solution.
- Marks the corresponding {-1, 0, 1}-restricted case as solved by the same result.
- Links a standalone Lean proof and its checksummed LRAT certificate bundle through formal_proof.

The proof reduces the integer system modulo two. The diagonal equations force odd perfect-matching parity; an exhaustive S_6 orbit classification leaves 47 Boolean support cases, each discharged by a reflected LRAT certificate.

## Verification

- lake build FormalConjectures.Paper.MonochromaticQuantumGraph
- Fresh standalone rebuild of all proof modules and 47 certificates
- Axiom audit contains no sorryAx; native reflection uses Lean.ofReduceBool and Lean.trustCompiler as documented

Certificate bundle SHA-256: c8855196dd1014ea59508078ed3e1f37f91dfff4e06e93317c98ef60a512f7c8

### DIFF
--- a/FormalConjectures/Paper/MonochromaticQuantumGraph.lean
+++ b/FormalConjectures/Paper/MonochromaticQuantumGraph.lean
@@ -654,10 +654,15 @@ theorem eqSystem4_no_solution_ge4_int :
   sorry
 
 /-- For $N = 6$ and $D = 3$, does there exist no solution to the monochromatic quantum graph
-equation system over $\mathbb{Z}$? -/
-@[category research open, AMS 5 14 81]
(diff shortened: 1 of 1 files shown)


mo271:
@MarioKrenn6240 
Does this look good to you?

mo271 (approved):
Thanks, LGTM!

mo271:
I think we now already have those proofs, please merge head and re-open if cases remain

PR #4514, opened 2026-07-21
Disprove WOWII Conjecture 65

SamuelSchlesinger:
## Summary

- correct `distMin` to minimize over distinct pairs, with `0` only as the fewer-than-two-vertices fallback
- add a verified computable bridge for evaluating `distMin` on finite vertex sets
- formalize a 17-vertex counterexample to WOWII Conjecture 65
- change the conjecture from `research open` to the proved answer `False`

## Counterexample

Start with the path `0-1-...-12` and attach disjoint triangles at vertices `1` and `11`. The only minimum-degree vertices are `0` and `12`, at distance `12`; the only maximum-degree vertices are `1` and `11`, at distance `10`. The proposed lower bound is therefore

`12 + ceil(10 / 3) = 16`.

Every induced forest must omit at least one vertex from each of the two vertex-disjoint triangles, so it contains at most `15` vertices. This contradicts the proposed bound.

## Formal verification

The proof checks connectivity explicitly, verifies the extremal-degree vertex sets, computes both distances through the proved BFS bridge, constructs a cycle for any induced subgraph containing either triangle, and derives the global forest-size upper bound by cardinality.

The solved theorem and its counterexample proof contain no `sorry` or native-decision proof.

## Validation

- `lake --wfail build` — passed all 8,876 jobs on the rebased branch
- branch rebased onto current `google-deepmind/formal-conjectures:main`


MrBrain295:
I don't think it is false. The conjecture as stated can be proven true. https://github.com/MrBrain295/formal-conjectures/blob/65/FormalConjectures/WrittenOnTheWallII/GraphConjecture65.lean is a proof of it being true. It looks like you changed the definition of the problem.

SamuelSchlesinger:
I'll check now if this one is or if I screwed up in reformulating it :)

SamuelSchlesinger:
Nope, I am still quite confident my reformulation was required and your proof validates that for me. By the way, discovering that proof was why I ended up reformulating it originally. What your proof basically says is:

in a connected graph, some outside vertex sits at distance 1 from S via a boundary dart, so old-distMin ≤ 1 for any nonempty proper S, and any two vertices induce a forest. The "conjecture" then reduces to 1 + ⌈1/3⌉ = 2 ≤ f(G).

Danie-I (commented):
Thank you for the correction of dist_min. In your counterexample dist_min(A) = 1, because you can take just an edge of a triangle for example.

SamuelSchlesinger:
A is not an arbitrary subset or induced forest. It is the set of all minimum-degree vertices. In this graph A={0,12}, since only the path endpoints have degree 1. The triangle vertices have degrees 2 or 4, so an edge of a triangle cannot be used to compute dist_min(A). Thus dist_min(A)=dist(0,12)=12.

SamuelSchlesinger:
I just re-reviewed this statement of the conjecture and I am quite confident it, and the disproof, are correct.

Danie-I:
Yes, I agree. I was checking different graph. Could you please add graph6 string of counterexample, so that it is easy to visualize? It should be PhCGGC@?G?_@?@O?G?G?G?@C

FormalConjectures/WrittenOnTheWallII/GraphConjecture65.lean
```diff
@@ -21,28 +21,189 @@ import FormalConjecturesUtil
 
 *Reference:*
 [E. DeLaVina, Written on the Wall II, Conjectures of Graffiti.pc](http://cms.dt.uh.edu/faculty/delavinae/research/wowII/)
+
+## Counterexample
+
+The conjecture is false. Start with the path $v_0-v_1-\cdots-v_{12}$ and attach one
+triangle at $v_1$ and another at $v_{11}$. The resulting graph has $17$ vertices.
```
Danie-I, line 28: Here you can add Graph_6 string: PhCGGC@?G?_@?@o?G?G?G?@C

SamuelSchlesinger: Done.

Danie-I (approved):
Thanks. LGTM

SamuelSchlesinger:
I believe it needs approval to run the rest of the workflows.

PR #4520, opened 2026-07-21
Solve wowii 145

DomTheDeveloper:
Marks Written on the Wall II Graph Conjecture 145 as solved and links the complete immutable Lean 4 proof. The exact theorem statement and hypotheses are unchanged, and the upstream change is limited to the solved status, the formal_proof reference, and provenance.

The complete proof is named 145.lean and is available at:

https://github.com/DomTheDeveloper/crl/blob/2ee448baa80c98f0c8b9a0c1c3d9421200f99aa5/math/wowii145/WOW145/145.lean

It was checked with the pinned Lean 4.27.0 environment, including direct compilation and #print axioms, with no sorryAx, admit, or native_decide.

The proof uses a diametral induced path for the general bound. In the local-independence-one case, a minimizing vertex forces radius at most two, and the final extremal case follows from the independently kernel-checked WOWII 146 exceptional theorem.

Solved by Dominic Dabish.

ProofOrchestrator, using OpenAI GPT-5.6 Thinking, assisted with the mathematical argument and Lean formalization; all formal claims were checked by the pinned Lean compiler.

### DIFF
--- a/FormalConjectures/WrittenOnTheWallII/GraphConjecture145.lean
+++ b/FormalConjectures/WrittenOnTheWallII/GraphConjecture145.lean
@@ -71,8 +71,17 @@ independence number of the complement graph.
 We state the inequality in the form
 $\mathrm{tree}(G) \cdot \mathrm{lMin}(\overline{G}) \ge 2 \cdot \mathrm{ecc}(B)$
 to avoid division.
+
+## Provenance
+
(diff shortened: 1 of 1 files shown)


Danie-I (approved):
Thanks for the contribution. Proof looks correct.

PR #4543, opened 2026-07-22
ci: point the linter label at the renamed util directory

williamjblair:
`.github/labeler.yml` still globs `FormalConjectures/Util/Linters/**`, a path the util rename removed, so the `linter` label stopped being applied to linter changes.

Checked the other eight globs while I was in there; they all still resolve.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -16,7 +16,7 @@ green-problems:
 
 linter:
   - changed-files:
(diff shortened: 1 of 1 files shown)


mo271:
see also #4498

mo271:
in fact it is already fixed there... closing in favour of https://github.com/google-deepmind/formal-conjectures/pull/4498

PR #4544, opened 2026-07-22
ErdosProblems: add 45, 46, 94, 105, 115, 150, 202, 205, 214, 337 (#3998 sync)

### LINKED ISSUE #336: Erdős Problem 105
### What is the conjecture
    
https://www.erdosproblems.com/105

  Let $A,B\subset \mathbb{R}^2$ be disjoint sets of size $n$ and $n-3$ respectively, with not all of $A$ contained on a single line. Is there a line which contains at least two points from $A$ and no points from $B$?

### Choose either option

### LINKED ISSUE #421: Erdős Problem 202
### What is the conjecture
    
https://www.erdosproblems.com/202

  Let $n_1<\cdots < n_r\leq N$ with associated $a_i\pmod{n_i}$ such that the congruence classes are disjoint (that is, every integer is $\equiv a_i\pmod{n_i}$ for at most one $1\leq i\leq r$). How large can $r$ be in terms of $N$?

Status: open

### Choose either option

### LINKED ISSUE #423: Erdős Problem 205
### What is the conjecture
    
https://www.erdosproblems.com/205

  Is it true that all sufficiently large $n$ can be written as $2^k+m$ for some $k\geq 0$, where $\Omega(m)<\log\log m$? (Here $\Omega(m)$ is the number of prime divisors of $m$ counted with multiplicity.) What about $<\epsilon \log\log m$? Or some more slowly growing function?

Status: solved

### Choose either option

williamjblair:
Adds statements for Erdős problems 45, 46, 94, 105, 115, 150, 202, 205, 214, 337.

Docstrings quote the boxed text on erdosproblems.com verbatim, with the solution sentence copied for solved problems. No `formal_proof` annotations: all ten show a green PROVED/DISPROVED (LEAN) badge, but the same page reports "Formalised statement? No" and gives no artifact URL, so there was nothing stable to pin.

Formalization choices:

- **45**: `D` is `{d ∈ n.divisors | 1 < d ∧ d < n}`. The `k`-colouring is a total `ℕ → Fin k` rather than a function on `D`, which avoids subtype coercions. Reciprocal sums use the existing `Finset.reciprocalSum`. `n = 0` gives an empty divisor set and sum `0 ≠ 1`, so I dropped the positivity hypothesis.
- **46**: colourings are `𝓒 : ℕ → ℕ` with `(Set.range 𝓒).Finite`, following 303. Sums are taken in `ℚ`, monochromaticity is `(𝓒 '' S).Subsingleton`. `variants.infinitely_many_disjoint` reads "infinitely many disjoint such monochromatic solutions" as each solution being monochromatic on its own, not all sharing one colour.
- **94**: the boxed text asserts rather than asks, so there is no `answer(...)` wrapper (cf. `erdos_1`). `≪` becomes `∃ C > 0`. `f(u)` counts unordered pairs, hence the `/2`. I inscribed `regularNGon` in the unit circle; the conjecture is scale-invariant, so the radius is my choice. Reuses `ConvexIndep` and `NonTrilinear`. The hosted SpringSense file assumes `ConvexPosition ∧ NoThreeCollinear`; I assume `ConvexIndep` alone, since their own `convexPosition_noThreeCollinear` makes the second conjunct redundant.
- **105**: `A.card = B.card + 3` replaces `n` and `n - 3`, which would use truncated `ℕ` subtraction and admit `n < 3`. A line is `line[ℝ, p, q]`, the affine span of two distinct points, rather than a bespoke `IsLine`. `variants.beck_szemerediTrotter` interprets "true with $n-3$ replaced by $cn$" rather than quoting it.
- **115**: the `o(1)` unfolds as `∀ ε > 0, ∀ᶠ n in atTop`, so the error term cannot depend on the polynomial. The `max` is a universal bound rather than a `sSup`. I left out the site's "equality iff $p(z) = z^n$" clause because it is false as written: $cz^n$ with $\lvert c\rvert = 1$ also attains the bound.
- **150**: worth a close look. The hosted gist defines a minimal cut as a minimal $u$-$v$ separator for a fixed pair. The problem text says "a minimal set of vertices whose removal disconnects the graph", inclusion-minimal among all disconnecting sets. Every minimal disconnecting set is a minimal separator for some pair, but not conversely, so the gist's $c(n)$ dominates the text's. I followed the text. "Disconnects" is `¬ Preconnected`, since `Connected` bundles nonemptiness and would give $K_n$ a spurious minimal cut.
- **202**: disjointness follows the site's "at most one $i$" phrasing rather than pairwise `Disjoint`. `StrictMono` on the moduli carries weight: drop it and `n₁ = n₂ = 2` with residues `0, 1` leaves `r` unbounded. I added `0 < n i` since "modulus" implies it. The eventual form is needed because `L(2) = 1` breaks small `N`. The hosted `Shashi456` file is named `Proof.lean`, but its `Erdos202Statement` is a `def : Prop` with no theorem asserting it.
- **205**: the three boxed questions become `parts.i/ii/iii`, with no standalone `erdos_205`. `Ω` is Mathlib's `ArithmeticFunction.cardFactors`. `Ω m < log log m` fails for `m ∈ {0, 1}`, so `m` needs no side condition. Romanoff's theorem lives at `Erdos851.erdos_851.variants.romanoff`, so I did not restate it.
- **214**: a blue set `S` with `Sᶜ` red, following the problem text, rather than the hosted files' `Color` type. `HasRedCopies n` carries `Function.Injective K`: the site defines `k` over "every set of `k` points", and an arbitrary family is not a set of `n` points. This also matches the literature's set-based counterexamples. `k := sSup {n | HasRedCopies n}` works because the predicate is downward closed in `n`, which I have not stated.
- **337**: uses `Set.IsAddBasis` where neighbouring files 330, 868 and 881 use `IsAsymptoticAddBasisOfOrder`. Over `ℕ` the exact-order version forces `0, 1 ∈ A`, which leaves the counting hypotheses alone. The `h`-fold sumset needed a local `sumsetOfOrder`, since Mathlib's `h • A` on `Set ℕ` is ambiguous between pointwise dilation and iterated sumset.

Part of #3998.

---
Closes #336
Closes #421
Closes #423


FormalConjectures/ErdosProblems/150.lean
```diff
@@ -0,0 +1,110 @@
...
+open Filter
+
+open scoped Topology
+
+namespace Erdos150
+
+/-- A *minimal cut* of a graph `G` is a minimal set `T` of vertices whose removal disconnects
+`G`, that is, `G - T` fails to be preconnected while `G - S` is preconnected for every `S ⊂ T`. -/
+def IsMinimalCut {V : Type*} (G : SimpleGraph V) (T : Set V) : Prop :=
+  ¬ (G.induce Tᶜ).Preconnected ∧ ∀ S ⊂ T, (G.induce Sᶜ).Preconnected
```
mo271, line 45: `IsMinimalCut` is a candidate for `FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Connectivity.lean` or something like this, but can also be done later.

FormalConjectures/ErdosProblems/202.lean
```diff
@@ -0,0 +1,75 @@
...
+/--
+Let $n_1<\cdots < n_r\leq N$ with associated $a_i\pmod{n_i}$ such that the congruence classes are
+disjoint (that is, every integer is $\equiv a_i\pmod{n_i}$ for at most one $1\leq i\leq r$). How
+large can $r$ be in terms of $N$?
+
+Let $f(N)$ be the maximum possible $r$, and let $L(N)=\exp(\sqrt{\log N\log\log N})$.
+
+This was proved by GPT-5.4 Pro (prompted by Ho Boon Suan), using the argument of [BFV13] together
+with the resolution of the Kahn-Kalai conjecture by Park and Pham [PaPh24], so that
+\[f(N)= N L(N)^{-1+o(1)}.\]
```
mo271: `\[` won't build anymore, please fix

mo271 (approved):
Thanks, LGTM!

FormalConjectures/ErdosProblems/150.lean
williamjblair: Left it in 150 for now. It is the only use in the repo and Mathlib has no equivalent, so there's nothing to dedupe yet. We could move it to Connectivity.lean if you'd rather have it there ahead of a second use

FormalConjectures/ErdosProblems/202.lean
williamjblair: Fixed in d12971c0. It was in six other files on this branch as well: 94, 115, 150, 205, 214, 337, all converted to `$$`. 150 also had the display jammed against the prose on both sides, so it moved onto its own line.

PR #4545, opened 2026-07-22
ErdosProblems: add 447, 481, 649, 650, 666, 692, 698, 751, 753, 756 (#3998 sync)

### LINKED ISSUE #751: Erdős Problem 481
### What is the conjecture
    
https://www.erdosproblems.com/481

  Let $a_1,\ldots,a_r,b_1,\ldots,b_r\in \mathbb{N}$ such that $\sum_{i}\frac{1}{a_i}>1$. For any finite sequence of $n$ (not necessarily distinct) integers $A=(x_1,\ldots,x_n)$ let $T(A)$ denote the sequence of length $rn$ given by
$$(a_ix_j+b_i) {1\leq j\leq n, 1\leq i\leq r}.$$
Prove that, if $A_1=(1)$ and $A_{i+1}=T(A_i)$, then there must be some $A_k$ with repeated elements.
  

Status: open

### Choose either option

### LINKED ISSUE #851: Erdős Problem 650
### What is the conjecture
    
https://www.erdosproblems.com/650

  Let $f(m)$ be such that if $A\subseteq \\{1,\ldots,N\\}$ has $\lvert A\rvert=m$ then every interval in $[1,\infty)$ of length $2N$ contains $\geq f(m)$ many distinct integers $b_1,\ldots,b_r$ where each $b_i$ is divisible by some $a_i\in A$, where $a_1,\ldots,a_r$ are distinct.

Estimate $f(m)$. In particular is it true that $f(m)\ll m^{1/2}$?
  

Status:  open

### Choose either option

williamjblair:
Adds statements for Erdős problems 447, 481, 649, 650, 666, 692, 698, 751, 753, 756.

Docstrings quote the boxed text on erdosproblems.com verbatim, with the solution sentence copied for solved problems. No `formal_proof` annotations: as in the previous batch, each of these shows a PROVED or DISPROVED (LEAN) badge while the same page reports "Formalised statement? No" and gives no artifact URL.

Formalization choices:

- **447**: two yes/no questions in the box, so `parts.i` and `parts.ii` with no standalone `erdos_447`. Neither Mathlib nor `FormalConjecturesForMathlib` has union-free families, so the file defines `UnionFree` and `maxUnionFree` locally. The ground set is `Fin n` rather than `Finset.range n` with a subset hypothesis. `o(2^n)` quantifies over the extremal function rather than over each family, which is what keeps the claim non-vacuous.
- **481**: the box says "Prove that", not a question, so there is no `answer(...)` wrapper. `A_{i+1} = T(A_i)` uses `Function.iterate` rather than a hypothesised sequence: writing `A 1 = [1]` alongside `∀ i, A (i+1) = T (A i)` would also constrain `A 0` and make the hypotheses contradictory. Elements are typed `ℕ`, since `A₁ = (1)` and `aᵢ, bᵢ ∈ ℕ` keep every `A_k` in `ℕ`.
- **649**: `P(m)` is the repo's `Nat.maxPrimeFac`, matching 648. No positivity hypothesis on `n`, since `maxPrimeFac 0 = maxPrimeFac 1 = 0` already forces `n ≥ 2` when `P(n)` is prime. One character in the quoted commentary is corrected: the site reads "this problem is false is a strong sense".
- **650**: the interval length follows the text's `2N` with `A ⊆ Finset.Icc 1 N`, where the hosted file uses `2·sup A`. The interval is also anchored by `1 ≤ x`, which the text's "every interval in $[1,\infty)$" requires and the hosted file omits. The binding case is `N = max A`, so the two agree, but the statements differ literally. `parts.i` is an equality rather than a bound, since van Doorn, Li and Tang settled the exact value.
- **666**: Mathlib has no hypercube graph, so the file generalises the `Q3` definition already in 567.lean. A subgraph is a spanning `H ≤ Q n`, and edges are counted with `edgeSet.ncard` so that no `DecidableRel H.Adj` instance is needed on a universally quantified `H`. The variant below the box is left out: its `k = 3` case appears false for the same reason the main problem is, since the four-part `C_6`-free edge partition already gives a `C_6`-free subgraph with `≥ n2^n/8` edges.
- **692**: `δ₁` is not defined as a function. Each statement quantifies over any `δ` satisfying the density characterisation, which avoids having to prove the density exists. `parts.ii` asks where the maximum is attained, so it uses `answer(sorry)` in term position, following 434.lean. The hosted gist has no main statement at all; its top-level theorem is Cambie's `n = 3` counterexample, which appears here as a variant.
- **698**: `h : ℕ → ℕ` with `Tendsto h atTop atTop` rather than a real-valued `h`, since gcd is a natural and the two forms are interchangeable. `≫` becomes an explicit absolute constant. The two Erdős–Szekeres variants were checked numerically (all `n < 90`, and `p` up to 101) before being written, so neither is vacuously wrong. The hosted file does not state the main question, only the quantitative bound.
- **751**: neither Mathlib nor ForMathlib has a set-of-cycle-lengths definition, so `cycleLengths` is local. "Min gap arbitrarily large" is stated as "every gap ≥ k" over distinct cycle lengths, which avoids truncated `ℕ` subtraction and building the enumeration. Bondy and Vince's "two cycles whose lengths differ by at most 2" is formalized as `m < m' ∧ m' ≤ m + 2`; read literally it is satisfied by taking one cycle twice.
- **753**: Mathlib has no choosability, so `IsKChoosable` and `listChromaticNumber` are local. Colours are drawn from `ℕ`, which is equivalent for graphs on `Fin n`. The guards `0 < n` and `2 ≤ n` stop `answer(False)` being discharged degenerately: at `n = 0` the empty graph makes the inequality fail, and `(n log n)^{1/2}` vanishes at `n = 1` while the sum is 2. The hosted gist states the already-negated asymptotic form; this follows the text's question form.
- **756**: points live in `ℝ²` where the hosted file works in `ℂ`. `≫` is rendered with `IsBigO`, following 89.lean. The hosted file's top-level theorem is Bhowmick's construction rather than the Erdős question, so it appears here as `variants.bhowmick` and the question takes the headline.

Part of #3998.

---
Closes #751
Closes #851


FormalConjectures/ErdosProblems/751.lean
```diff
@@ -0,0 +1,77 @@
...
+  Math. Pannon. (1994), 261--269.
+- [BoVi98] Bondy, J. A. and Vince, A., Cycles in a graph whose lengths differ by one or two.
+  J. Graph Theory (1998), 11--15.
+-/
+
+namespace Erdos751
+
+/-- The set $\{m_1, m_2, \ldots\}$ of lengths of the cycles in a graph `G`. -/
+def cycleLengths {V : Type*} (G : SimpleGraph V) : Set ℕ :=
+  {m | ∃ (v : V) (c : G.Walk v v), c.IsCycle ∧ c.length = m}
```
mo271: we could move this to `FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Circumference.lean`
and actually define Circumference using is via `sSup G.cycleLengths` (and girth via `sInf G.cycleLengths`

FormalConjectures/ErdosProblems/447.lean
```diff
@@ -0,0 +1,70 @@
...
+-/
+
+open Filter
+
+namespace Erdos447
+
+/-- A collection $\mathcal{F}$ of sets is *union-free* if there are no solutions to
+$A\cup B=C$ with distinct $A,B,C\in \mathcal{F}$. -/
+def UnionFree {α : Type*} [DecidableEq α] (F : Finset (Finset α)) : Prop :=
+  ∀ A ∈ F, ∀ B ∈ F, ∀ C ∈ F, A ≠ B → A ≠ C → B ≠ C → A ∪ B ≠ C
```
mo271:
```lean
def UnionFree {α : Type*} (F : Finset (Finset α)) : Prop :=
  (F : Set (Finset α)).Triplewise fun A B C ↦ A ∪ B ≠ C
```
might be more elegant (untested)

Also: it seems like a very common concept, perhaps let's move it to `FormalConjecturesForMathlib`?

FormalConjectures/ErdosProblems/692.lean
```diff
@@ -0,0 +1,110 @@
...
+-/
+def IsDelta₁ (n m : ℕ) (δ : ℝ) : Prop :=
+  (exactlyOneDivisorIn n m).HasDensity δ
+
+/--
+The sequence $f$ is unimodular on $(a, \infty)$ if it increases until some point and
+decreases thereafter.
+-/
+def UnimodularOn (f : ℕ → ℝ) (a : ℕ) : Prop :=
+  ∃ M > a, MonotoneOn f (Set.Ioc a M) ∧ AntitoneOn f (Set.Ici M)
```
mo271: seems like clearly this is a useful definition for many concepts, so we might consider moving it to ForMathlib?

mo271 (commented):
Thanks for the contribution!

Looks great, just a few nits around making stuff more re-usable...

FormalConjectures/ErdosProblems/447.lean
williamjblair: Moved to `FormalConjecturesForMathlib/Combinatorics/SetFamily/UnionFree.lean` as `Finset.UnionFree`.

The `Triplewise` version doesn't typecheck though, there's no `Set.Triplewise` in Mathlib, only `List.Triplewise`. Kept the explicit form for now.

FormalConjectures/ErdosProblems/666.lean
```diff
@@ -0,0 +1,63 @@
...
+def Q (n : ℕ) : SimpleGraph (Fin n → Bool) where
+  Adj u v := #{i | u i ≠ v i} = 1
+  symm _ _ := by simp [eq_comm]
+  loopless _ := by simp
+
+/--
+Let $Q_n$ be the $n$-dimensional hypercube graph (so that $Q_n$ has $2^n$ vertices and $n2^{n-1}$
+edges). Is it true that, for every $\epsilon>0$, if $n$ is sufficiently large, every subgraph of
+$Q_n$ with
+\[\geq \epsilon  n2^{n-1}\]
```
williamjblair: Already in, eb8d1d55 landed just before the review. All ten files on the branch are clean.

FormalConjectures/ErdosProblems/666.lean
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Erdos666
+
+/-- $Q_n$ is the $n$-dimensional hypercube graph: the vertices are the $n$-bit vectors, and two
+vertices are adjacent when they differ in exactly one coordinate. It has $2^n$ vertices and
+$n2^{n-1}$ edges. -/
+def Q (n : ℕ) : SimpleGraph (Fin n → Bool) where
+  Adj u v := #{i | u i ≠ v i} = 1
+  symm _ _ := by simp [eq_comm]
+  loopless _ := by simp
```
williamjblair: Added `SimpleGraph.hypercube` in `Combinatorics/SimpleGraph/Hypercube.lean`, 666 uses that now. Also pointed `Q3` in 567 at it, since that one was the same definition verbatim.

Left `Hypercube` in `Wikipedia/SnakeInTheBox.lean` alone. It's on `Finset (Fin n)` rather than `Fin n → Bool`, so folding it in means redoing the proofs there rather than swapping a definition. Can do that separately if you want all three unified.

FormalConjectures/ErdosProblems/751.lean
williamjblair: Done, `cycleLengths` lives in `Circumference.lean` now and `circumference` is `sSup G.cycleLengths`.

Left `girth` as it is. Mathlib already has one, and `girth = sInf cycleLengths` wants an actual proof rather than a definition swap. Can add it as a lemma if that's useful.

One wrinkle: `cycleLengths` can't pick up the file's `[Fintype α] [DecidableEq α]`, since 751 quantifies over a bare `V : Type`. So it sits above the `variable` line with its own binder.

FormalConjectures/ErdosProblems/692.lean
williamjblair: Moved to `FormalConjecturesForMathlib/Order/Unimodular.lean`.

FormalConjectures/ErdosProblems/447.lean
mo271: FormalConjecturesForMathlib/Data/Set/Triplewise.lean

williamjblair: My mistake, I only looked in Mathlib. Switched to it:

```lean
def UnionFree {α : Type*} [DecidableEq α] (F : Finset (Finset α)) : Prop :=
  (F : Set (Finset α)).Triplewise fun A B C ↦ A ∪ B ≠ C
```

`[DecidableEq α]` has to stay for the `A ∪ B`.

mo271 (approved):
Many thanks, LGTM now!

williamjblair:
-awaiting-author

All review changes are in and the build is green; ready to merge.

PR #4546, opened 2026-07-22
ErdosProblems: add 674 with plby formal_proof links (#3998 sync)

williamjblair:
Adds a statement for Erdős problem 674, with `formal_proof` links to plby's Lean proof.

The docstring quotes the boxed text on erdosproblems.com verbatim, along with the sentence below it giving Ko's partial result and the explicit counterexample.

`erdos_674` is the boxed yes/no question as `answer(True) ↔ solutionSet.Nonempty`. `variants.infinite` states the stronger claim from the text below the box, that there are infinitely many solutions.

Both carry a `formal_proof` link. plby's file proves `solutionSet.Nonempty` and `solutionSet.Infinite` over a `solutionSet` definition identical to the one here, with no `sorry` and no `axiom`, and `sources.yaml` marks 674 neither partial nor conditional.

Part of #3998.


mo271 (approved):
Thanks, LGTM!

PR #4547, opened 2026-07-22
doc: correct the cochromaticNumber docstring

williamjblair:
`cochromaticNumber` carried a docstring copy-pasted from Mathlib's `chromaticNumber`. It described the chromatic number rather than the cochromatic number, and told the reader that `ENat.toNat G.chromaticNumber` gives the ℕ-valued answer, which is a different definition.

Rewritten to describe the cochromatic number, keeping the original structure (what `⊤` means, and the `ENat.toNat` remark).

Also adds a docstring to `Cocolorable`, which had none.

Nothing referenced either definition until now, which is presumably how the docstring survived. Two Erdős problems I am formalising, 760 and 762, are the first consumers.


mo271 (approved):
Thanks, LGTM!

PR #4548, opened 2026-07-22
ErdosProblems: add 760, 762, 775, 785, 794, 798, 818, 844, 862, 867 (#3998 sync)

### LINKED ISSUE #938: Erdős Problem 775
### What is the conjecture

https://www.erdosproblems.com/775

  Is there a $3$-uniform hypergraph on $n$ vertices which contains at least $n-O(1)$ different sizes of cliques (maximal complete subgraphs)

Status:  solved

### Choose either option

### LINKED ISSUE #986: Erdős Problem 862
### What is the conjecture

https://www.erdosproblems.com/862

  Let $A_1(N)$ be the number of maximal Sidon subsets of $\\{1,\ldots,N\\}$. Is it true that
$$A_1(N) < 2^{o(N^{1/2})}?$$
Is it true that 
$$A_1(N) > 2^{N^c}$$
for some constant $c>0$?

Status:  solved

### Choose either option

williamjblair:
Adds statements for Erdős problems 760, 762, 775, 785, 794, 798, 818, 844, 862, 867.

Docstrings quote the boxed text on erdosproblems.com verbatim, with the solution sentence copied for solved problems. No `formal_proof` annotations, for the same reason as the previous two batches: these pages carry a PROVED or DISPROVED (LEAN) badge while reporting "Formalised statement? No" and giving no artifact URL.

Two small corrections to quoted text. 775's boxed question ends with no punctuation on the site, so a `?` was added. 862's quoted commentary reads "there number of Sidon sets", corrected to "the".

Formalization choices:

- **760**: uses the existing `SimpleGraph.cochromaticNumber` from ForMathlib rather than a bespoke definition. Subgraphs are `G.Subgraph` with `H.coe`, following 108.lean and 1175.lean, since the problem asks for arbitrary rather than induced subgraphs. The `ℕ∞`-valued cochromatic number is bridged to the real bound through an explicit natural witness rather than `ENat.toNat`, which would send `⊤` to `0` and corrupt the bound. The hosted gist uses `Nat.log 2` where the text says `\log`, so this uses `Real.log`.
- **762**: also uses `cochromaticNumber`. Restricted to finite graphs, which the original does not say but Steiner's counterexample is finite, so the `answer(False)` verdict is unaffected. Both `chromaticNumber` and `cochromaticNumber` are `ℕ∞`, so `χ ≤ ζ + 2` needs no casts.
- **775**: reuses the repo's `ThreeUniformHypergraph`. "Clique" is maximal complete subgraph, per the problem's own parenthetical, via Mathlib's `Maximal`. `n - O(1)` uses `∃ᶠ`, the weakest existence reading, which makes `answer(False)` the strongest refutation. The hosted gist is a full proof of Gao's paper and states only the refutation, with no question form.
- **785**: nine variants, all attested on the problem page and named after their authors. Flagging the density in case that is more than wanted. `IsAdditiveComplement` uses `∀ᶠ n in atTop`, the repo idiom for "contains all large integers", where the hosted file uses cofiniteness of the complement. Counting functions are indexed by `ℕ` rather than `ℝ`, which changes none of the limits. The hosted file's positivity hypotheses on `A` and `B` are dropped, since the problem text states none.
- **794**: does not reuse `ThreeUniformHypergraph`, unlike 775 in the same PR. That structure's `edges` field is a `Set (Finset V)` with no `card`, and this problem is entirely about edge counts, so it uses `Finset (Finset V)` with a local uniformity predicate. "Contains a subgraph on `m` vertices with `k` edges" is `k ≤`, not `k =`; the site's own Balogh remark reads "spanning" the same way, and `=` would break that implication.
- **798**: `t(n)` is an `sInf` over line covers, with lines rendered as `Collinear ℝ {p, q, x}` with `p ≠ q`. The hosted gist's covering relation permits a degenerate line through one point; the problem says the lines are determined by pairs. The bounds are named `variants.lower_bound` and `variants.upper_bound`: the README's estimate-question rule suggests bare `.lower_bound`, but `ErdosProblems/` uses the `variants.` form 30 times and the bare form never.
- **818**: `Finset ℤ` where the hosted gist works over `Finset ℝ` with positivity hypotheses, since the text says integers. `Real.log` where the gist uses `Nat.clog`, which differs by a constant the existential absorbs. `2 ≤ A.card` guards `log` positivity.
- **844**: `IsGreatest` rather than the gist's one-sided upper bound. "Achieved by" asserts both that the candidate set is admissible and that nothing admissible is larger.
- **862**: two boxed questions with opposite answers, so `parts.i` is `answer(False)` and `parts.ii` is `answer(True)`. Reuses `Set.IsMaximalSidonSetIn` from ForMathlib.
- **867**: the consecutive-sum condition uses an interval characterisation, `∀ m n, 2 ≤ (Icc m n ∩ A).card → (∑ a ∈ Icc m n ∩ A, a) ∉ A`, rather than the gist's sort-and-slice. The elements of `A` in an interval are exactly a contiguous run of the sorted listing, so the two agree.

Part of #3998.

---
Closes #938
Closes #986


FormalConjectures/ErdosProblems/794.lean
```diff
@@ -0,0 +1,103 @@
...
+  323-328.
+-/
+
+open Filter
+
+namespace Erdos794
+
+/-- A finite family `H` of sets of vertices is a *`3`-uniform hypergraph* if every edge of `H`
+has exactly `3` vertices. -/
+def IsThreeUniform {V : Type*} (H : Finset (Finset V)) : Prop := ∀ e ∈ H, e.card = 3
```
mo271: seems like we already have this? 
775 uses ThreeUniformHypergraph V while 794 uses Finset (Finset V) with a predicate. Providing coercion / equivalence bridges between Finset (Finset V) and ThreeUniformHypergraph V will unify future hypergraph problems.

I think there is already more (perhaps pending?) work on (three uniform) hypergraphs for ForMathlib, perhaps good to check what can be unified streamlined here.

FormalConjectures/ErdosProblems/775.lean
```diff
@@ -0,0 +1,93 @@
...
+  ∀ e : Finset V, e ⊆ S → e.card = 3 → e ∈ H.edges
+
+/-- The set of sizes of the cliques (maximal complete subgraphs) of a 3-uniform
+hypergraph `H`. -/
+def cliqueSizes {V : Type} (H : ThreeUniformHypergraph V) : Set ℕ :=
+  { k | ∃ S : Finset V, Maximal (IsCompleteSubgraph H) S ∧ S.card = k }
+
+/-- The set of sizes of the cliques (maximal complete subgraphs) of a simple graph `G`. -/
+def graphCliqueSizes {V : Type} (G : SimpleGraph V) : Set ℕ :=
+  { k | ∃ S : Finset V, Maximal (fun T : Finset V => G.IsClique (T : Set V)) S ∧ S.card = k }
```
mo271: look like candidate for FormalConjecturesForMathlib

FormalConjectures/ErdosProblems/785.lean
```diff
@@ -0,0 +1,195 @@
...
+  (A ∩ Set.Icc 1 x).ncard
+
+/-- The largest element of $A$ in $[1,x]$, that is $a^*(x)=\max A\cap [1,x]$
+(equal to $0$ when there is no such element). -/
+noncomputable def aStar (A : Set ℕ) (x : ℕ) : ℕ :=
+  sSup (A ∩ Set.Icc 1 x)
+
+/-- Two sets $A, B\subseteq \mathbb{N}$ are *additive complements* if $A+B$ contains all
+large integers. -/
+def IsAdditiveComplement (A B : Set ℕ) : Prop :=
```
mo271: Seems like this should already be in the repo somewhere?
A quick search only yielded `IsAdditiveComplementToPrimes`, but perhaps worth looking again, because it certainly looks liek something that woudl be a good fit for `FormalConjecturesForMathlib` in some form

mo271 (commented):
Thanks for the contribution!
Looks great, just again a few comments on re-usability.

FormalConjectures/ErdosProblems/794.lean
williamjblair: Unified them. `IsThreeUniform` and `ContainsSubgraph` now live in `Combinatorics/Hypergraph/ThreeUniform.lean` next to `ThreeUniformHypergraph`, in the `Finset` namespace so they read as `H.IsThreeUniform` / `H.ContainsSubgraph 4 3`. Added `ThreeUniformHypergraph.ofFinset` as the bridge.

Kept 794 on `Finset (Finset V)` rather than converting it to the structure, since its statements count edges (`n^3 + 1 ≤ H.card`) and the structure carries a `Set`. The `ofFinset` bridge means either side can move across when a proof needs it.

Didn't find other pending hypergraph work in ForMathlib beyond `ThreeUniform.lean` itself. If something is in flight elsewhere, point me at it and I'll reconcile.

FormalConjectures/ErdosProblems/775.lean
williamjblair: Moved. `IsCompleteSubgraph` and `cliqueSizes` are in `Combinatorics/Hypergraph/ThreeUniform.lean`, and the simple-graph one is `SimpleGraph.cliqueSizes` in `Combinatorics/SimpleGraph/Clique.lean` alongside `ecliqueNum`.

FormalConjectures/ErdosProblems/785.lean
williamjblair: Moved to `FormalConjecturesForMathlib/NumberTheory/AdditiveComplement.lean`.

`IsAdditiveComplementToPrimes` in 32 is the same shape specialised to the primes, so it could become `IsAdditiveComplement {p | p.Prime}`. Left 32 alone for now since it's outside this PR, happy to do it if you want them merged.

`IsExactAdditiveComplement` stayed in 785, it's defined in terms of the local `counting` function. That one also looks reusable if you want it pulled up.

mo271 (approved):
Thanks, LGTM now

PR #4549, opened 2026-07-22
GreensOpenProblems/27: link formal proof of variants.upper_be23

mkwatson:
Per @mo271's review, this drops the inline proof and links it instead. Now a two-line change.

```lean
@[category research solved, AMS 5 11,
  formal_proof using lean4 at "https://github.com/mkwatson/unique-sums-notes/blob/d4383f93a52e3a916ccce78d5fb7739c3bf3dabc/Green27Upper/Theorem.lean#L26"]
```

The linked project takes `formal_conjectures` as a Lake dependency (pinned at `9538d0ae57bd36ef5a5bd717dfe900401bbaee01`) and imports `FormalConjectures.GreensOpenProblems.«27»`, so `m`, `primesAtTop` and `upperBest` are the definitions from this repository rather than local copies. The proposition proved is the one stated in this file, so there is nothing to reconcile. It is named `upper_be23_proof` there only because the file imports the module that declares `green_27.variants.upper_be23`.

The route is Łuczak-Schoen ("On the number of unique sums"), via Straus's simple set, Lemmas 4 and 5, and the Q = 2 combine.

- Zero `sorry`. `#print axioms Green27.upper_be23_proof` reports `[propext, Classical.choice, Quot.sound]`.
- Clean `lake build Green27Upper` on `leanprover/lean4:v4.27.0`, against the same Mathlib revision this repository pins (`a3a10db0e9d66acbebf76c5e6a135066525ac900`).
- No `native_decide`, no `bv_decide`, no external certificate checking.
- Pinned by full SHA with a line anchor.

AI disclosure: the Lean proof bodies were filled using Codex GPT-5.6 Sol. Mathematical claims were checked against the cited sources, and the statements are checked by Lean.


mo271 (commented):
Thanks for the contribution, this looks great!
But since the the proof is a bit long (and adds 3 files, that don't look super re-useable), perhaps we could instead use the `formal_proof using formal_conjectures at [link]` mechanism here?

FormalConjectures/GreensOpenProblems/27.lean
```diff
@@ -80,11 +80,55 @@ theorem green_27.variants.lower_be23 :
...
+theorem repCount_pair_test : Finset.repCount ({0, 1} : Finset (ZMod 5)) 1 = 2 := by
+  decide
+
+/-- Sanity check: Straus's simple set in `ZMod 11` is `{0, ±5, ±2, ±1}`, of size
+`7 = 2 * Nat.log 2 11 + 1`. -/
+@[category test, AMS 5 11]
+theorem strausSet_eleven_test :
+    strausSet 11 = ({0, 5, 2, 1, -5, -2, -1} : Finset (ZMod 11)) := by
+  decide
+
```
mo271: those are great and shoud be checked, but prefer latex markdown over backticks

mkwatson:
Appologies for the slow reply, I was out of the country. Done as suggested: the proof lives in my notes repo and this PR is now just the `formal_proof` link, pinned to a commit. While reworking it I noticed `repCount` is definitionally `Finset.addConvolution` (the equality is proved by rfl in the linked repo), so the extra files really were redundant with Mathlib. The two test checks moved out with the definitions they exercise, so the docstring note no longer applies to this diff. From my side this is ready for another look. Thanks for the review!

mo271 (approved):
Thanks, LGTM!

mkwatson:
> Collaborator

Looks like the merge queue dropped this one even though the checks on the PR itself are green. Happy to update the branch if that would help.

PR #4550, opened 2026-07-22
ErdosProblems: add 898, 905, 907, 914, 958, 966, 967, 974, 990, 1000 (#3998 sync)

### LINKED ISSUE #1056: Erdős Problem 990
### What is the conjecture

https://www.erdosproblems.com/990

  Let $f=a_0+\cdots+a_dx^d\in \mathbb{C}[x]$ be a polynomial. Is it true that, if $f$ has roots $z_1,\ldots,z_d$ with corresponding arguments $\theta_1,\ldots,\theta_d\in [0,2\pi]$, then for all intervals $I\subseteq [0,2\pi]$
$$\left\lvert (\\# \theta_i \in I) - \frac{\lvert I\rvert}{2\pi}d\right\rvert \ll \left(n\log M\right)^{1/2},$$
where $n$ is the number of non-zero coefficients of $f$ and
$$M=\frac{\lvert a_0\rvert+\cdots +\lvert a_d\rvert}{(\lvert a_0\rvert\lvert a_d\rvert)^{1/2}}.$$

Status:  open

### Choose either option

### LINKED ISSUE #1062: Erdős Problem 1000
### What is the conjecture

https://www.erdosproblems.com/1000
Let $A=\{n_1<n_2<\cdots\}$ be an infinite sequence of integers, and let $\phi_A(k)$ count the number of $1\leq m\leq n_k$ such that the fraction $\frac{m}{n_k}$ does not have denominator $n_j$ for $j<k$ when written in lowest form; equivalently, $$\frac{n_k}{(m,n_k)}\neq n_j$$ for all $1\leq j<k$.

Is there a sequence $A$ such that $$\lim_{N\to \infty}\frac{1}{N}\sum_{k\leq N}\frac{\phi_A(k)}{n_k}=0?$$

Status:  solved

### Choose either option

williamjblair:
Adds statements for Erdős problems 898, 905, 907, 914, 958, 966, 967, 974, 990, 1000.

Docstrings quote the boxed text on erdosproblems.com verbatim, with the solution sentence copied for solved problems. No `formal_proof` annotations, as in the previous batches: these pages carry a PROVED or DISPROVED (LEAN) badge while reporting "Formalised statement? No" and giving no artifact URL.

Two places where the quoted text departs from the site, both because the site as written is wrong:

- **958** says the multiplicity counts **ordered** pairs. Under that reading the problem is vacuous: $n$ equidistant points on a line give ordered multiplicities $2n-2,\dots,2$, so $\{f(d_i)\}=\{n-1,\dots,1\}$ can never hold. The source paper counts unordered pairs, and the Clemen–Dumitrescu–Liu counterexample has unordered multiplicities exactly $n-1,\dots,1$. The docstring says unordered, so that it agrees with the Lean.
- **974**'s conclusion $z_j = e(j/n)$ is false for even $n$: at $n=2$, $z=(1,i)$ has $s_k=0$ whenever $k \equiv 2 \bmod 4$. The box hedges with "(essentially)" and Erdős declines to say what that means, so the statement uses the parity split quoted in the same docstring: roots of unity for odd $n$, and two regular $(n/2)$-gons on the unit circle for even $n$.

Formalization choices:

- **898**: perpendicular feet are given as quantified points constrained by membership in the side line plus orthogonality of directions, the idiom in 189.lean, rather than `EuclideanGeometry.orthogonalProjection`. The feet lie on the side *lines* rather than the closed segments, since for an obtuse triangle the foot from an interior point can fall outside the segment, and the Erdős–Mordell inequality is about distance to the lines.
- **905**: stated bare rather than as `answer(True) ↔`, since the boxed text is an assertion. Note 904.lean is also an assertion on the site but uses the `answer` wrapper, so this differs from its immediate neighbour; CONTRIBUTING's rule is what I followed. Both `n²/4` and `n/6` are over `ℝ`, since truncated `ℕ` division would weaken the statement.
- **907**: the problem overloads `h` for both the shift and the additive summand; the Lean calls the additive function `a` while the docstring keeps the original notation. The hypothesis is `∀ h : ℝ, 0 < h → …` rather than `∀ h > 0, …`, since the binder-sugar form risks elaborating `h` at `ℕ`.
- **914**: bare, per the same rule as 905. The equitable-colouring reformulation from below the box is a variant. The hosted file's headline is the classical max-degree form of Hajnal–Szemerédi, which is not the text on the page, so it does not appear.
- **958**: three bespoke definitions built on the `ℝ²` notation from ForMathlib. "Equidistant points on a circle" reads as equally spaced by angle rather than a full regular $n$-gon; a regular $n$-gon has only $\lfloor n/2 \rfloor$ distinct distances and could never satisfy $k = n-1$.
- **966**: reuses `Set.IsAPOfLengthFree` and `ContainsMonoAPofLength` from ForMathlib. Non-triviality comes from the length condition rather than an explicit `d ≠ 0`, since `IsAPOfLength` requires the cardinality to match.
- **967**: the finite-sequence case is a separate `research open` variant. The hosted gist folds finite and infinite sequences together, which would make Yip's disproof read as settling the finite case too, but the site says it remains open.
- **974**: see above. `z_1 = 1` is a real hypothesis rather than cosmetic, since scaling `z` preserves vanishing of the power sums.
- **990**: requires `f.coeff 0 ≠ 0` throughout. Without it `M` is `0/0 = 0` and `Real.log 0 = 0` in Lean, and the counterexample statement would be satisfiable by a junk polynomial. Intervals are `Set.Icc`, since the counterexample interval is written `[0, c/d]`.
- **1000**: the sequence is 0-indexed, so the problem's `1 ≤ j < k` becomes `j < k` and `∑_{k ≤ N}` becomes `∑ k ∈ Finset.range N`. The number of summands is unchanged, so the `1/N` normalisation still matches.

Part of #3998.

---
Closes #1056
Closes #1062


FormalConjectures/ErdosProblems/905.lean
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Erdos905
+
+open Finset
+
+/-- The triangles of `G` containing the edge `e`, that is, the `3`-cliques of `G` whose vertex
+set contains both endpoints of `e`. -/
+def trianglesContaining {n : ℕ} (G : SimpleGraph (Fin n)) [DecidableRel G.Adj]
+    (e : Sym2 (Fin n)) : Finset (Finset (Fin n)) :=
+  (G.cliqueFinset 3).filter (fun t ↦ e.toFinset ⊆ t)
```
mo271: this already exists in ErdosProblem/600.lean: let's not duplicate it here, but move to FormalConjecturesForMathlib?

FormalConjectures/ErdosProblems/990.lean
```diff
@@ -0,0 +1,123 @@
...
+
+open scoped Real
+
+namespace Erdos990
+
+/--
+The argument of a complex number, normalised to take values in the interval `[0, 2π)`.
+-/
+noncomputable def normArg (z : ℂ) : ℝ :=
+  if Complex.arg z < 0 then Complex.arg z + 2 * π else Complex.arg z
```
mo271: since it is only used once, perhaps better inline?

mo271 (commented):
Thanks, LGTM in general just two comments

FormalConjectures/ErdosProblems/905.lean
williamjblair: Moved to `SimpleGraph.trianglesContaining` in `Combinatorics/SimpleGraph/Clique.lean`, and 600 uses it too now instead of its own private copy.

FormalConjectures/ErdosProblems/990.lean
williamjblair: Inlined.

williamjblair:
-awaiting-author

Both review comments are addressed and the build is green; ready for re-review.

mo271 (approved):
Thanks, LGTM!

PR #4551, opened 2026-07-22
ErdosProblems: add 1007, 1008, 1014, 1022, 1023, 1026, 1028, 1034, 1036, 1037 (#3998 sync)

### LINKED ISSUE #1072: Erdős Problem 1014
### What is the conjecture

https://www.erdosproblems.com/1014

  Let $R(k,l)$ be the Ramsey number, so the minimal $n$ such that every graph on at least $n$ vertices contains either a $K_k$ or an independent set on $l$ vertices.

Prove, for fixed $k\geq 3$, that
$$\lim_{l\to \infty}\frac{R(k,l+1)}{R(k,l)}=1.$$

Status:  open

### Choose either option

### LINKED ISSUE #1078: Erdős Problem 1022
### What is the conjecture

https://www.erdosproblems.com/1022

  Is there a constant $c_t$, where $c_t\to \infty$ as $t\to \infty$, such that if $\mathcal{F}$ is a finite family of finite sets, all of size at least $t$, and for every set $X$ there are $<c_t\lvert X\rvert$ many $A\in \mathcal{F}$ with $A\subseteq X$, then $\mathcal{F}$ has chromatic number $2$ (in other words, has property B)?

Status:  open

### Choose either option

### LINKED ISSUE #1079: Erdős Problem 1023
### What is the conjecture

https://www.erdosproblems.com/1023

  Let $F(n)$ be the maximal size of a family of subsets of $\\{1,\ldots,n\\}$ such that no set in this family is the union of other members of the family. Is it true that there is a constant $c>0$ such that
$$F(n)\sim c \frac{2^n}{n^{1/2}}?$$

Status:  open

### Choose either option

### LINKED ISSUE #1080: Erdős Problem 1026
### What is the conjecture

https://www.erdosproblems.com/1026

  Let $x_1,\ldots,x_n$ be a sequence of distinct real numbers. Determine
$$\max\left(\sum x_{i_r}\right),$$
where the maximum is taken over all monotonic subsequences.

Status:  solved

### Choose either option

### LINKED ISSUE #1085: Erdős Problem 1034
### What is the conjecture

https://www.erdosproblems.com/1034

  Let $G$ be a graph on $n$ vertices with $>n^2/4$ many edges. Must there be a triangle $T$ in $G$ and vertices $y_1,\ldots,y_t$, where $t>(\frac{1}{2}-o(1))n$, such that every vertex is joined to at least two vertices of $T$?

Status:  solved

### Choose either option

williamjblair:
Adds statements for Erdős problems 1007, 1008, 1014, 1022, 1023, 1026, 1028, 1034, 1036, 1037.

Docstrings quote the boxed text on erdosproblems.com verbatim, with the solution sentence copied for solved problems. No `formal_proof` annotations, as in the previous batches.

None of these ten has a hosted Lean formalization, so all are drafted from the problem text alone.

Four places where the quoted text departs from the site, in each case because the site as written is wrong:

- **1028** defines `f` on `X^2` while quantifying `f` outside the maximum over `X`, so the domain depends on a variable not yet bound. It also sums over `x ≠ y`, i.e. ordered pairs; with an arbitrary `f` one can take `f(x,y) = -f(y,x)` and every subset sum vanishes, giving `H(n) = 0` and contradicting the quoted bounds. The docstring uses the fixed domain `{1,…,n}^2` and sums over `x < y`, which is the classical Erdős–Spencer imbalance problem.
- **1034** says "such that every **vertex** is joined to at least two vertices of $T$". That makes the list $y_1,\dots,y_t$ and the bound $t > (\tfrac12-o(1))n$ vacuous, and it is false as stated, since a graph with more than $n^2/4$ edges can have isolated vertices. Ma and Tang's note states it as "every $y_i$".
- **1014**'s solution sentence reads "This been solved"; corrected to "This has been solved".
- **1037** has "the number of distinct degree is" and "commment section"; both corrected.

Formalization choices:

- **1007**: "achieved solely by $K_{3,3}$" is false on the nose, since adjoining isolated vertices preserves both dimension 4 and 9 edges, so the uniqueness variant carries a no-isolated-vertices hypothesis. Injectivity in the embedding is load-bearing: without it $K_{3,3}$ would have dimension 1, by collapsing each side to a point.
- **1008**: the Folkman bound is stated on the site without proof, so it was checked before being written down. In a $C_4$-free subgraph of $K_{n,n^2}$ the $B$-side neighbourhoods pairwise meet in at most one vertex, so $\sum_b \binom{d_b}{2} \le \binom{n}{2}$, and $d - 1 \le \binom{d}{2}$ for $d \ge 2$ gives the $n^2 + \binom{n}{2}$ bound.
- **1014**: reuses `graphRamseyNumber` and its `R(k, l)` notation from `FormalConjectures/Wikipedia/RamseyNumbers.lean` rather than defining a local Ramsey number. `IsGraphRamsey.succ` supplies the monotonicity that makes the problem's "at least $n$ vertices" equivalent to the existing "on $n$ vertices" definition.
- **1022**: quantifies over nonempty `X`. Read literally the sparsity condition fails at $X = \emptyset$, since the count is 0 and $0 < 0$ is false, which would make the hypothesis unsatisfiable and the answer vacuously yes.
- **1023**: the union-free condition requires the subfamily to be nonempty. Without that the empty union is `∅`, so `∅` could never belong to a valid family, which the English does not assert.
- **1026**: the boxed question is, in the site's own words, "rather ambiguous" and has no formalizable content alone, so the statement is van Doorn's precise version of the constant $c$, which the site endorses. The clarifying text is quoted in the docstring.
- **1028**: `n/4 ≤ H(n)` holds eventually rather than for all `n`. It fails at small `n`: `H(1) = 0`, and at `n = 5` the pentagon colouring gives `H(5) = 1 < 5/4`.
- **1034**: `Y` is not required to be disjoint from the triangle. The box does not say "other vertices", and Ma and Tang bound the set over all of $V(G)$; the two readings differ by at most three vertices, which the $\varepsilon$ absorbs.
- **1036**: counts a family of pairwise non-isomorphic induced subgraphs rather than quotienting by isomorphism, which is equivalent and avoids the quotient.
- **1037**: "much larger than $\log n$" reads as $\omega(\log n)$, the reading consistent with the $O(\log n)$ refutation. Degrees use `(G.neighborSet v).ncard` rather than `G.degree v`, so no decidability instance needs threading through the existentially quantified graph in the counterexample variant.

Part of #3998.

---
Closes #1072
Closes #1078
Closes #1079
Closes #1080
Closes #1085


FormalConjectures/ErdosProblems/1028.lean
```diff
@@ -0,0 +1,84 @@
...
+theorem erdos_1028.variants.lower_bound :
+    ∀ᶠ n : ℕ in atTop, (n : ℝ) / 4 ≤ (H n : ℝ) := by
+  sorry
+
+/-- Erdős [Er63d] proved
+$$\frac{n}{4}\leq H(n) \ll n^{3/2}.$$
+-/
+@[category research solved, AMS 5]
+theorem erdos_1028.variants.upper_bound :
+    (fun n => (H n : ℝ)) =O[atTop] (fun n : ℕ => (n : ℝ) ^ (3 / 2 : ℝ)) := by
```
mo271: could use the notation we have for $\ll$ in our repo..

FormalConjectures/ErdosProblems/1023.lean
```diff
@@ -0,0 +1,83 @@
...
+open scoped Asymptotics
+
+namespace Erdos1023
+
+/--
+A family `S` of finite sets is union-free when no set in `S` is the union of other
+members of `S`.
+-/
+def IsUnionFree {α : Type*} [DecidableEq α] (S : Finset (Finset α)) : Prop :=
+  ∀ A ∈ S, ∀ T ⊆ S.erase A, T.Nonempty → T.sup id ≠ A
```
mo271: Generalizes 3-set UnionFree from 
447.lean
 to arbitrary non-empty subfamilies ($A \neq \bigcup T$ for $T \subseteq S \setminus {A}$).
could be unified in `Combinatorics/SetFamily/UnionFree.lean`

mo271 (commented):
Thanks, looks good to me, just two minor comments

FormalConjectures/ErdosProblems/1028.lean
williamjblair: Switched both this one and the Erdős-Spencer variant below it to `≪`.

FormalConjectures/ErdosProblems/1023.lean
williamjblair: Unified. It's `Finset.SubfamilyUnionFree` in `Combinatorics/SetFamily/UnionFree.lean` now, sitting next to the pairwise `UnionFree` from 447 that it strengthens.

That file arrives with #4545, so whichever of the two merges second will hit a small conflict there. Happy to rename either if `UnionFree` / `SubfamilyUnionFree` isn't the split you'd want.

williamjblair:
-awaiting-author

Both review comments are addressed and the build is green; ready for re-review.

williamjblair:
Merged main in to clear the conflict. It was an add/add on `FormalConjecturesForMathlib/Combinatorics/SetFamily/UnionFree.lean`: main landed the same `Finset.UnionFree` definition, character for character, so the resolution just keeps it with this branch's `SubfamilyUnionFree` on top. Nothing else changed.

`lake build --wfail FormalConjectures FormalConjecturesForMathlib` passes on the merge.

mo271 (approved):
Thanks, LGTM now!

PR #4552, opened 2026-07-22
ErdosProblems: add 1044, 1047, 1048, 1098, 1121, 1136, 1190, 1193 (#3998 sync)

### LINKED ISSUE #1091: Erdős Problem 1044
### What is the conjecture

https://www.erdosproblems.com/1044

  Let $f(z)=\prod_{i=1}^n(z-z_i)\in\mathbb{C}[x]$ where $\lvert z_i\rvert\leq 1$ for all $i$. If $\Lambda(f)$ is the maximum of the lengths of the boundaries of the connected components of 
$$\\{ z: \lvert f(z)\rvert<1\\}$$
then determine the infimum of $\Lambda(f)$.

Status:  solved

### Choose either option

williamjblair:
Adds statements for Erdős problems 1044, 1047, 1048, 1098, 1121, 1136, 1190, 1193.

This exhausts the unformalized problems in plby's `sources.yaml` that are not flagged `partial` or `conditional`. Docstrings quote the boxed text on erdosproblems.com verbatim, with the solution sentence copied for solved problems. No `formal_proof` annotations.

Three corrections where the site is wrong, and one bibliography substitution:

- **1047** gives the referee's constant as `5.6^{-6/5}`, where the `.` is a multiplication dot rather than a decimal point: it is $5 \cdot 6^{-6/5}$. For $f(z) = z^6 - z$ we have $f' = 6z^5 - 1$, so at $z^5 = 1/6$ the value is $6^{-1/5} \cdot \tfrac56 = 5 \cdot 6^{-6/5}$. Read as the decimal 5.6 the statement is false.
- **1190** states $\epsilon_m < 1$ without restriction, which fails at $m = 0$: the single congruence $x \equiv 0 \pmod 1$ is admissible, giving $\epsilon_0 = 1$. The variant carries `1 ≤ m`.
- **1044** has "then the the infimum"; corrected.
- **1121** cites `[Ha47]` for Hadwiger's generalisation to convex bodies, but `erdosproblems.com/bibs/Ha47` resolves to Hall, "Cyclic projective planes", Duke Math. J. (1947) — an unrelated paper. The reference list carries the intended Hadwiger, *Nonseparable convex systems*, Amer. Math. Monthly (1947) instead. Reverting to the site's key would mean citing the wrong paper.

Formalization choices:

- **1044**: imports 1041 and reuses its `length` (the 1-dimensional Hausdorff measure) rather than redefining it; both are [EHP58] problems. `IsAdmissible` keeps the product form of the text, with `0 < n` forced by $\prod_{i=1}^n$ — without it $f = 1$ gives an empty sublevel set and the infimum would collapse to 0. "Determine the infimum" uses `IsGLB S answer(sorry)` with the known value 2 in a separate variant.
- **1047**: Pommerenke's $a = (1+\tfrac1k)k^{1/(k+1)}$ is exactly the critical value, so "sufficiently close" is formalized as arbitrarily close rather than throughout a neighbourhood; on one side the two components merge, which would make the neighbourhood reading false. The component count in the referee's example is stated as 6, which the site omits, so that the statement is an instance of the problem's hypothesis rather than a vacuous non-convexity claim.
- **1048**: imports 1043 and reuses its `levelSet`. The box uses the open sublevel set while the text below uses the closed one, and both are followed verbatim: the bullets speak of "the component which contains 0", and at $r = 1$ one can have $|f(0)| = 1$, so 0 would not lie in the open set at all. A nonconstancy hypothesis is added, since $f = 1$ is monic with an empty sublevel set and would falsify the statement for an unrelated reason. `Metric.ediam` is used rather than `EMetric.diam`, which is deprecated in Mathlib v4.27 and would emit a warning that CI treats as a failure.
- **1098**: builds the non-commuting graph as a real `SimpleGraph`, so "complete subgraph" is Mathlib's `IsClique`, and bounds clique size with `Finset.card` rather than `Set.ncard` — the latter is 0 on infinite sets and would make the bound spuriously satisfiable. Carries the batch's one proved declaration, a `simp` lemma unfolding the adjacency by `Iff.rfl`.
- **1121**: circles are closed discs. A line meets the open disc iff it meets the bounding circle, so the separation hypothesis is unchanged, and a convex covering region contains the circle iff it contains the disc, so the conclusion is unchanged. Non-separability is stated as all centres lying strictly on one side, which is equivalent since each disc is connected and misses the line.
- **1136**: reuses `Set.lowerDensity` and `Set.HasDensity`. Müller's set is written in the site's congruence form $3\cdot 2^i \pmod{2^{i+2}}$ rather than the hosted file's odd-part form; the two are equivalent. The upper bound is stated for lower density, following the site, where the hosted file proves the stronger upper-density version.
- **1190**: defines its own `L m = exp(√(log m · log log m))`. Note 202.lean, in the batch 9 PR, defines the same scale as `Erdos202.L`; the namespaces differ so there is no conflict, but they could be unified once both land.
- **1193**: needs no bespoke definitions, reusing `AdditiveCombinatorics.sumRep` (which is exactly $1_A \ast 1_A$) and `Set.lowerDensity`/`upperDensity`. The hosted gist states only the counterexample identity and never formalizes either question, so that identity is a variant here.

Part of #3998.

---
Closes #1091


mo271 (approved):
Thanks, LGTM!

PR #4554, opened 2026-07-22
Fix antiRamseyNum being identically zero, and CDSColorable's implicit graph

williamjblair:
`antiRamseyNum` returns `0` for every input.

Its defining set `{k | ∃ c : Sym2 (Fin n) → Fin k, ∀ f, ¬IsRainbow f c}` is upward closed in `k`: compose a witness with `Fin.castSucc` and you get one for `k+1`, since `castSucc` is injective and `IsRainbow` survives it. The set is therefore empty or unbounded, and `sSup` on `ℕ` gives `0` either way. Requiring `c` to be surjective bounds the set by the edge count, and the supremum becomes the maximum the docstring describes.

That matters beyond the docstring. `ErdosProblems/1105.lean` states specific values of `antiRamseyNum`, and those statements were vacuous while the function was constant.

`CDSColorable` takes `G` implicitly, but `G` never appears in the resulting `Prop`, so unification cannot infer it and the only way to use the definition is `CDSColorable (G := G)`. It is now explicit.

`ErdosProblems/1105.lean` and `Paper/LatinTableau.lean` both still typecheck.

### DIFF
--- a/FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Coloring.lean
+++ b/FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Coloring.lean
@@ -148,7 +148,7 @@ noncomputable def indepNumK (G : SimpleGraph V) (k : ℕ) : ℕ :=
 by natural numbers such that for all `k > 0`, the number of
 vertices with color `< k` equals the maximum size of
 the union of `k` independent sets. -/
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, good catch! Those fixes are super valuable of course

PR #4577, opened 2026-07-23
fix(ErdosProblems/424,822): use lower density for "positive density"

### LINKED ISSUE #4553: Erdős 424: "positive density"
At https://www.erdosproblems.com/424 it says
> As with many of Erdős' questions, by 'positive density' he most likely meant 'positive lower density' - in other words, does there exist 𝑐 >0 such that for all large 𝑥 the number of values of 𝑎_𝑖 in [1,𝑥] is at least 𝑐⁢𝑥?

but it seems the formalization here still uses "the density exists and it is positive", ie "the lower density is positive AND equals the upper density".

### Suggested fix
Obviously you can just say that the lower density is positive.  But perhaps a new predicate can be defined specifically for this situation, which occurs more than once.

### Additional context
It would be good to scan the remaining problems for the correct interpretations of "positive density".

### Choose either option

kid-ye:
### Summary

erdosproblems.com/424 notes that when Erdős asks about "positive density," he almost
always means *positive lower density* — i.e. `∃ c > 0` such that for all large `x`,
at least `c·x` of the elements up to `x` are in the set — not that the natural density
exists (lower = upper) and is positive.

Both `erdos_424` and `erdos_822` used `Set.HasPosDensity`, which requires the density
to *converge* to a positive value. That's a strictly stronger statement than the source
intends, and in `erdos_822`'s case it directly contradicted its own docstring, which
already said "positive (**lower**) density."

### Changes

- **`FormalConjectures/ErdosProblems/424.lean`**: the main `erdos_424` statement now uses
  `0 < generatedSet.lowerDensity`. The old literal ("density exists and is positive")
  interpretation is preserved as `erdos_424.variants.exact_density`, following the same
  pattern already used in `erdos_125` and `erdos_741` for this exact ambiguity.
- **`FormalConjectures/ErdosProblems/822.lean`**: statement updated to
  `0 < (Set.range fun n => n + Nat.totient n).lowerDensity`, matching what its docstring
  already claimed. No change to `answer(True)` — [GIL24] proves the lower-density
  statement.

No new definitions were introduced; `Set.lowerDensity` already exists in
`FormalConjecturesForMathlib/Data/Set/Density.lean` and is used the same way elsewhere.

### Validation

- `lake build FormalConjectures.ErdosProblems.«424» FormalConjectures.ErdosProblems.«822»`
  completes successfully 

Fixes #4553.


mo271 (approved):
Thanks, LGTM!

PR #4578, opened 2026-07-23
ErdosProblems 130 254 267 394 489 521 769 796 1188: Star Fleet Math formal_proof links

williamjblair:
Adds `formal_proof` links to Lean 4 proofs by Star Fleet Math (Colin Snyder, starfleetmath.com), mirrored at [williamjblair/lean-proofs](https://github.com/williamjblair/lean-proofs) under `starfleet/` and pinned to `4f915a3`. Not my work. Colin has confirmed the solutions are meant to be open for the community and is adding a licence to his bundles.

Where FC had no statement, this adds one mirroring the proof's own definitions so the link is faithful.

| Problem | Star Fleet calls it | What the proof gives |
|--:|---|---|
| 254 | full solution | `answer(True)` |
| 267 | full solution | `answer(True)`: the reciprocal-Fibonacci sum is irrational |
| 394 | full solution | `answer(True)`, both parts |
| 489 | full solution | `answer(True)`: the squared-gap average converges |
| 796 | full solution | `answer(True)`: asymptotic for `g_3(n)` |
| 1188 | full solution | `loglog(F(x))/log x → 1` |
| 130 | partial | `answer(True)`: the chromatic number can be infinite |
| 769 | partial | `answer(False)`: `c(n) ≫ n^n` is false |
| 521 | proof of a result others got first | `answer(False)` |

Those three classes are Colin's own. Partial means "real, machine-checked results on the problem, where the problem as posed asks for more". Two of them show up in the docstrings: 130 answers only the chromatic half of a question that also asks about the clique number, and 521 credits the earlier solvers, with the link described as an independent proof.

## Verification

I read each linked theorem against the FC statement it sits on, unfolding any named `Prop` and comparing every supporting definition. Nine of the ten match exactly, including 130's seven definitions and 769's eight. Where the two sides differ it is notation or an indirection that unfolds the same way: `×ˢ` against `.product`, `Finset.card` against `fun A => A.card`, and helpers that FC inlines. 267's hosted theorem carries an explicit `0 < n k` that FC's statement omits, and FC does not need it: `n 0 = 0` makes the ratio hypothesis read `c ≤ n 1 / 0 = 0` against `c > 1`, and `StrictMono` handles the rest.

Two agreed by argument rather than by unfolding, so this PR proves them.

- **521**: the proof builds `rademacherMeasure` from `bernoulliMeasure true false half`; FC's `fairCoin` is `(2 : ENNReal)⁻¹ • (dirac true + dirac false)`. `bernoulliMeasure` is not in FC's Mathlib, so `fairCoin_apply` and `fairCoin_isProbabilityMeasure` pin `fairCoin` to mass `1/2` on each sign instead, which identifies it uniquely.
- **254**: the proof's second hypothesis is `Tendsto (phasePartialSum A θ) atTop atTop` where FC has `¬ Summable (…)` over the subtype. `not_summable_subtype_iff_tendsto_sum_indicator`, new in ForMathlib, relates the two for a nonnegative summand; `not_summable_iff_tendsto_partial_sums` instantiates it.

Links are pinned, every linked file resolves, none contains `sorry` or declares an `axiom`. That check reads the linked file only, not its imports, and the `#print axioms` audit is the proofs repo's CI rather than something rerun here.

## One proof filled in

`erdos_394.variants.factorial_gap_10` was a `sorry` and is now proved: the Erdős–Selfridge result that `t k (10!) < t (k-1) (10!) - 1` for `2 ≤ k < 10`. It needs all nine values of `t k (10!)`, which are `3628800, 512000, 6398, 5373, 348, 160, 30, 9, 2`, each with a witness for the upper bound and a check that every smaller positive `m` fails for the lower one. `t_eq_of` packages that shape and `t_one` proves `t 1 n = n` directly, since checking 3.6M cases to establish that the least positive multiple of `n` is `n` would be silly.

`k = 5..9` go by `decide`. `k = 2..4` need `native_decide`, the `k = 2` case ruling out all 512000 values below the answer. That puts `Lean.ofReduceBool` and `Lean.trustCompiler` in the axiom set for that one theorem, so it is trusted-by-compiler rather than kernel-checked. `native_decide` is used elsewhere in the repo, but say the word and I will drop it back to `sorry`.

## Status

These are all `open` in [problems.yaml](https://github.com/teorth/erdosproblems/blob/main/data/problems.yaml), so `scripts/check_erdos_status.py` will flag a mismatch for each until upstream picks them up. Raising it so it is not a surprise, and happy to follow whatever convention FC wants for machine-checked results that upstream has not caught up with.


mo271:
still marked as draft, are they ready for review?

williamjblair:
Yes, ready. Sorry for the lag, I took it out of draft but did not say so here.

Build is green and all ten pinned links still resolve. The one thing worth a reviewer's eye is the two statements in the Verification section that agreed by argument rather than by unfolding, 521 and 254. Those are where I would expect disagreement if there is any.

FormalConjectures/ErdosProblems/130.lean
```diff
@@ -0,0 +1,76 @@
...
+equidistant from all four. -/
+def Concyclic4 (p q r s : Point) : Prop :=
+  ∃ o : Point, sqDist o p = sqDist o q ∧ sqDist o p = sqDist o r ∧ sqDist o p = sqDist o s
+
+/-- General position: no three points collinear and no four points concyclic. -/
+def GeneralPosition (A : Set Point) : Prop :=
+  (∀ ⦃p q r⦄, p ∈ A → q ∈ A → r ∈ A →
+      p ≠ q → p ≠ r → q ≠ r → ¬ Collinear p q r) ∧
+  (∀ ⦃p q r s⦄, p ∈ A → q ∈ A → r ∈ A → s ∈ A →
+      p ≠ q → p ≠ r → p ≠ s → q ≠ r → q ≠ s → r ≠ s → ¬ Concyclic4 p q r s)
```
mo271: Let's change that to use these definitions we already have on ℝ²

FormalConjectures/ErdosProblems/130.lean
```diff
@@ -0,0 +1,76 @@
...
+
+/-- Two points are adjacent in the integer-distance graph iff they are distinct
+and their distance is a positive integer (equivalently `sqDist` is a positive
+perfect square). -/
+def Adjacent (p q : Point) : Prop :=
+  p ≠ q ∧ ∃ n : ℕ, 0 < n ∧ sqDist p q = (n : ℝ) ^ 2
+
+/-- The integer-distance graph on `A` has a proper colouring with `k` colours. -/
+def HasKColoring (A : Set Point) (k : ℕ) : Prop :=
+  ∃ color : A → Fin k, ∀ x y : A, Adjacent x.1 y.1 → color x ≠ color y
```
mo271: could also need more concepts from `FormalConjecturesForMathlib`

FormalConjectures/ErdosProblems/769.lean
```diff
@@ -0,0 +1,80 @@
...
+/--
+Let $c(n)$ be minimal such that if $k\geq c(n)$ then the $n$-dimensional unit
+cube can be decomposed into $k$ homothetic $n$-dimensional cubes. Give good
+bounds for $c(n)$ — in particular, is it true that $c(n)\gg n^n$?
+
+The `c(n) \gg n^n` conjecture is **false**: for odd `n`, one can tile the unit
+`n`-cube into `k` homothetic cubes for every `k ≥ U(n)` with `U(n)/n^n → 0`, so
+`c(n)/n^n → 0` along odd dimensions and no absolute constant lower-bounds it.
+
+Determining good bounds for `c(n)` in general remains open.
```
mo271, line 78: could we add some reasonable `research open` statement (also fine to leave a todo or do in a follow-up).

mo271 (commented):
Thanks for the contribution

FormalConjectures/ErdosProblems/130.lean
williamjblair: Done in 2a15d000. 130 uses `InGeneralPosition` and `IntegerDistancePlaneGraph` now, so `Point`, `sqDist`, `Collinear`, `Concyclic4`, `GeneralPosition`, `Adjacent` and `HasKColoring` are gone:

```lean
∃ A : Set ℝ², A.Infinite ∧ InGeneralPosition A ∧
  (IntegerDistancePlaneGraph A).chromaticNumber = ⊤
```

`InGeneralPosition` took a `Finset ℝ²` and 130 needs an infinite set, so it takes a `Set ℝ²` now; #98 and #655 get the coercion for free. `IntegerDistancePlaneGraph` is new, next to `UnitDistancePlaneGraph`, with `1` replaced by any positive integer.

FormalConjectures/ErdosProblems/130.lean
williamjblair: Same commit: adjacency and colouring are `IntegerDistancePlaneGraph` and `chromaticNumber` now.

FormalConjectures/ErdosProblems/769.lean
williamjblair: Added `erdos_769.variants.growth_rate`. Refuting $c(n)\gg n^n$ leaves "give good bounds" open, so it asks for the order on the problem's own scale:

```lean
∃ c : ℕ → ℕ, (∀ n, IsCutoff n (c n)) ∧
  ∃ γ : ℝ, Tendsto (fun n : ℕ => Real.log (c n) / (n * Real.log n)) atTop (𝓝 γ)
```

`c` is existential so the statement isn't vacuous if no cutoff function exists. The refuted conjecture would have forced $\gamma\geq 1$. Happy to make it a two-sided bound instead if you prefer.

mo271 (approved):
Thanks, LGTM!

PR #4587, opened 2026-07-23
ErdosProblems 486, 788: ShouqiaoW formal_proof links (486 statement correction)

### LINKED ISSUE #945: Erdős Problem 788
### What is the conjecture

https://www.erdosproblems.com/788

  Let $f(n)$ be maximal such that if $B\subset (2n,4n)\cap \mathbb{N}$ there exists some $C\subset (n,2n)\cap \mathbb{N}$ such that $c_1+c_2\not\in B$ for all $c_1\neq c_2\in C$ and $\lvert C\rvert+\lvert B\rvert \geq f(n)$. 

Estimate $f(n)$. In particular is it true that $f(n)\leq n^{1/2+o(1)}$?

Status:  open

### Choose either option

williamjblair:
Adds `formal_proof` links to Lean 4 proofs by ShouqiaoW ([github.com/ShouqiaoW/erdos](https://github.com/ShouqiaoW/erdos)), pinned to `f2ae0ed`. Not my work; I found and verified them.

| Problem | Change | What the proof gives |
|--:|---|---|
| 486 | statement rewrite + link | `answer(False)`: a delayed congruence system whose surviving set has no logarithmic density |
| 788 | new statement + link | `f(n) = n^{1/2+o(1)}` |

## 486 needed rewriting

The problem asks about `B = {m : m ∉ X_n (mod n) for all n ∈ A with m > n}`. The existing `erdos_486` dropped the modulus set `A` and the `m > n` activation, writing plain `∀ n`. Without the activation `n = 1` constrains every `m`, so the statement was degenerate.

The rewrite puts both back: `survivors` carries the `n < m` activation and `0 ∉ A`. It is `answer(False)` because the surviving set of ShouqiaoW's system oscillates, with `liminf ≤ 177/200` and `limsup ≥ 49/50`, so no logarithmic density exists. `answer(False) ↔ Erdos486Assertion` is definitionally `¬Erdos486Assertion`, which is the linked `erdos486_negative`.

## 788

New file. `f(n)` is `findGreatest` over the sumset-avoidance guarantee, matching "maximal such that", and `HasExponentOneHalf` says `f(n) = n^{1/2+o(1)}`. The linked proof gives the sharper `(1/2000)·√(n log n) ≤ f(n) ≤ n^{1/2 + O((loglog n / log n)^{1/3})}`; the exponent statement is one of its conjuncts.

## Verification

Both projects use FC's toolchain (`v4.27.0`) and Mathlib pin (`a3a10db`), so I rebuilt them against FC's own cache rather than taking the author's word. Clean builds, no `sorry`, and `#print axioms` on `Erdos486.erdos486_negative` and `Erdos788.erdos788` gives `[propext, Classical.choice, Quot.sound]`.

For faithfulness I compared each linked theorem against the FC statement it sits on, unfolding the named `Prop` and every supporting definition. `survivors`, `logSum`, `logAverage`, `HasLogDensity` and `Erdos486Assertion` are identical on both sides; so are `I`, `J`, `Admissible`, `Guarantees`, `scoreBound`, `fNat`, `f` and `HasExponentOneHalf` for 788.

## Notes

These are AI-assisted formalizations, per the source repo's README. The links point at that repo rather than vendoring the code, and it has no licence file yet; I've asked the author about licensing and intent.

Problem 1038 in the same repo is not included. It is a ~3200-file machine-generated proof and needs its own axiom and faithfulness review.

Closes #945


danielchin (commented):
I would probably hold off of these for now until we get some more vetting from Bloom on the Erdos Problems website.

I took a look at the 486 lean files and those seem okay, but I'd wait for someone to view the actual proof.

For 788, I would request that Shouqiao actually refactors or golfs the lean files. There's 2000+ lean files in the directory and I don't think it's reasonable to accept that as an accepted formalization.

williamjblair:
Agreed, will keep as draft. I think the lean files for some are pretty heavy brute force right now and could be cleaned up first

williamjblair:
Closing this draft so the active queue reflects work that is actually reviewable. Erdős 486 should return only after the mathematical proof has been vetted against the corrected statement. Erdős 788 should return only after the 2,000-plus-file proof package has been reduced to a reviewable, reproducible boundary. These are two different blockers and should be resubmitted separately when resolved; closing does not make any negative claim about the underlying mathematics.

ValeZAA:
390 is 390k of Lean, btw

williamjblair:
Yeah, that can probably be at least 10x smaller with Mathlib. Would be hard to support any of these unless someone rewrote them ...

ValeZAA:
> Yeah, that can probably be at least 10x smaller with Mathlib. Would be hard to support any of these unless someone rewrote them ...

How do you mean? The proof is 117 pages, so of course the mathlib use is gonna be large and original code is gonna be enormous. I mean the proof of Erdos 90 is a million of lines of Lean4.  https://github.com/plby/Erdos90/

williamjblair:
Woah really, didn't know it was that long! Just meant usually first pass of autonomous proofs results in a lot of marginal theorem-specific code that can be better abstracted. I think our caution with these on first pass was it seemed like some could be better consolidated or shortened, but not something I've looked seriously at

PR #4588, opened 2026-07-23
ErdosProblems 450, 538, 959: Star Fleet Math proved bounds (open headline + solved companion)

williamjblair:
Companion to #4578. Three machine-checked results from Star Fleet Math (Colin Snyder, starfleetmath.com), hosted at [williamjblair/lean-proofs](https://github.com/williamjblair/lean-proofs) and pinned to `4f915a3`. Not my work.

Each follows FC's idiom for an estimation problem (cf. `961.lean`): the headline `erdos_N` stays `research open` with `answer(sorry)`, and the proved bound is a `research solved` companion carrying the link. Nothing here claims to resolve its problem.

| Problem | Open headline asks | `research solved` companion |
|--:|---|---|
| 450 | how large must the window scale `y(ε,n)` be? | `erdos_450.linear_scale_suffices` — a linear scale `Y ε n ≤ C(ε)·n` suffices |
| 538 | best possible upper bound for `∑ 1/n` | `erdos_538.matching_order` — `∑ 1/n = Θ_r(log N / loglog N)` |
| 959 | does `extremalGap n` have a growth exponent? | `erdos_959.lower_bound` — `≥ n^(1 + c/loglog n)` |

Star Fleet classifies 450 and 538 as full solutions rather than partial. The headlines stay open anyway, since both problems ask for the best possible bound and the proofs give the order rather than the optimal constant. 959 it classifies as partial, which matches the treatment here.

## Verification

Each linked theorem was read against the FC statement it sits on, unfolding intermediate definitions on both sides. All three match exactly:

- **450** — FC inlines `localCount`; the proof routes it via `badIntegers`, which unfolds to the same `(Finset.Ioo x (x+y)).filter (HasMediumDivisor n)`
- **538** — `×ˢ` against `.product` is notation for the same operation
- **959** — all ten supporting definitions identical

Links are pinned, every linked file resolves, none contains `sorry` or declares an `axiom`. That check reads the linked file only, not its imports, and `#print axioms` is audited by the proofs repo's CI rather than rerun here.


mo271:
is this ready for review?

williamjblair:
Yes, ready. Build is green and the three pinned links resolve.

It is the companion to #4578, so it may be easier to read that one first. The thing to check here is the split: all three headlines stay `research open` with `answer(sorry)`, and the proved bound sits in a separate `research solved` companion that carries the link. Nothing in it claims to resolve its problem.

FormalConjectures/ErdosProblems/959.lean
```diff
@@ -0,0 +1,110 @@
...
+noncomputable section
+
+/-- A point of the real affine plane, represented by Cartesian coordinates. -/
+abbrev Point := ℝ × ℝ
+
+/-- Squared Euclidean distance. Squaring is injective on nonnegative reals, so
+equality of `sqDist` is equivalent to equality of ordinary Euclidean distance,
+and the two determine the same distance classes and multiplicities. -/
+def sqDist (p q : Point) : ℝ := (p.1 - q.1) ^ 2 + (p.2 - q.2) ^ 2
+
```
mo271: Duplicated across 130 and 959 should be unified.

also more generally use our standard notation for points ℝ²

FormalConjectures/ErdosProblems/959.lean
```diff
@@ -0,0 +1,110 @@
...
+-/
+@[category research open, AMS 52]
+theorem erdos_959 : answer(sorry) ↔
+    ∃ γ : ℝ, Tendsto (fun n : ℕ => Real.log (extremalGap n) / Real.log n) atTop (𝓝 γ) := by
+  sorry
+
+/--
+A superlinear lower bound: there is a `c > 0` such that for all large `n`,
+`extremalGap n ≥ n^{1 + c / log log n}`. In particular `max (f(d₁) - f(d₂))` grows
+faster than any linear function of `n`. This is a proved lower bound; determining
```
mo271: please prefer Latex markdown over backticks

mo271 (commented):
Thanks!

FormalConjectures/ErdosProblems/959.lean
williamjblair: Done in 4732dfd7. `Point` and `sqDist` are gone; the file uses `ℝ²` and `dist`.

Ordinary distance rather than squared costs nothing here since squaring is injective on nonnegatives, and it drops the paragraph that justified the choice. 130 loses the same pair in #4578.

FormalConjectures/ErdosProblems/959.lean
williamjblair: Done, that one and the five above it.

mo271 (approved):
Thanks, LGTM!

PR #4594, opened 2026-07-23
fix(Other): use integer-valued sequences for Beaver Math Olympiad 8

### LINKED ISSUE #4591: incorrect formalization of BeaverMathOlympiad
https://github.com/google-deepmind/formal-conjectures/blob/e751934294a381afd2d5fc1124c5953c8e25f9fa/FormalConjectures/Other/BeaverMathOlympiad.lean#L231
In the original problem statement, a_n, b_n are expected to be integers, not just natural numbers (note that a_n may become -1).

DomTheDeveloper:
## Summary

Corrects the formalization of Beaver Math Olympiad Problem 8 by changing the value type of the sequences `a` and `b` from `ℕ` to `ℤ`.

The original problem permits negative sequence values, so natural-number subtraction incorrectly truncates at zero. The original two-branch recurrence is otherwise unchanged.

Fixes #4591.

### DIFF
--- a/FormalConjectures/Other/BeaverMathOlympiad.lean
+++ b/FormalConjectures/Other/BeaverMathOlympiad.lean
@@ -228,7 +228,7 @@ using `answer(sorry) ↔`.
 -/
 @[category research open, AMS 5 11 68]
 theorem beaver_math_olympiad_problem_8 : answer(sorry) ↔
-    ∀ᵉ (a : ℕ → ℕ) (b : ℕ → ℕ)
+    ∀ᵉ (a : ℕ → ℤ) (b : ℕ → ℤ)
     (a_ini : a 0 = 10)
     (a_rec : ∀ n, a (n + 1) =
       if b n / 2 < a n then a n - b n / 2 - 3 else 3 * a n + 5)


FormalConjectures/Other/BeaverMathOlympiad.lean
```diff
@@ -231,10 +238,16 @@ theorem beaver_math_olympiad_problem_8 : answer(sorry) ↔
     ∀ᵉ (a : ℕ → ℕ) (b : ℕ → ℕ)
     (a_ini : a 0 = 10)
     (a_rec : ∀ n, a (n + 1) =
-      if b n / 2 < a n then a n - b n / 2 - 3 else 3 * a n + 5)
+      if a n = b n / 2 + 1 then a n
```
danielchin: I'm not entirely clear what you're doing here. It looks like you're stating a different problem?

FormalConjectures/Other/BeaverMathOlympiad.lean
```diff
@@ -210,19 +210,26 @@ theorem beaver_math_olympiad_problem_5 : answer(sorry) ↔
...
 
-for all positive integers $n$.  Does there exist a positive integer $i$ such that
-$a_i = \lfloor b_i/2 \rfloor + 1$?
+The first case makes the halting state absorbing so that the recurrence is total. Does there exist a
+positive integer $i$ such that $a_i = \lfloor b_i/2 \rfloor + 1$?
 
 [BMO#8](https://wiki.bbchallenge.org/wiki/Beaver_Math_Olympiad#8._1RB0LD_0RC1RB_0RD0RA_1LE0RD_1LF---_0LA1LA_(bbch)) is equivalent to asking whether the 6-state Turing machine
 [`1RB0LD_0RC1RB_0RD0RA_1LE0RD_1LF---_0LA1LA`](https://wiki.bbchallenge.org/wiki/1RB0LD_0RC1RB_0RD0RA_1LE0RD_1LF---_0LA1LA) halts or not.
 
+The former two-branch `ℕ` recurrence omitted the exceptional difference-two transition; truncated
```
danielchin: This comment isn't necessary.

danielchin (changes requested):
From the PR description, it sounds like you're trying to fix the issue with subtraction truncation which is also mentioned in #4591.

If that's the case, then it should just be matter of changing the type from $\mathbb{N}$ to $\mathbb{Z}$ since it looks like $a_n$ and $b_n$ can go negative.

FormalConjectures/Other/BeaverMathOlympiad.lean
DomTheDeveloper: Fixed.

DomTheDeveloper:
> From the PR description, it sounds like you're trying to fix the issue with subtraction truncation which is also mentioned in #4591.
> 
> If that's the case, then it should just be matter of changing the type from N to Z since it looks like a n and b n can go negative.

Fixed, thank you.

mo271 (approved):
Thanks, LGTM!

PR #4595, opened 2026-07-23
Fix the Hilbert fifth problem formulation

Deicyde:
## Summary

- Introduce `LieGroupPresentation`, packaging a continuous group isomorphism from `G` to a Hausdorff, second-countable carrier with its own real-analytic manifold atlas and analytic group operations.
- Define `AdmitsLieGroupStructure` through such a presentation.
- Correct `hilbert_fifth_problem` to produce a Lie-group presentation instead of asserting analyticity for the arbitrary input atlas.
- Update the supporting API to construct presentations and transfer local compactness across the continuous group equivalence.

## Related changes

- [PR #4598: Demonstrate a counterexample to the previous Hilbert fifth formulation](https://github.com/google-deepmind/formal-conjectures/pull/4598)
- [PR #4599: Formalize Pardon’s three-dimensional theorem](https://github.com/google-deepmind/formal-conjectures/pull/4599)

## Validation

- `lake --wfail build`


FormalConjectures/HilbertProblems/5.lean
```diff
@@ -40,28 +42,48 @@ variable {G : Type*} [Group G] [TopologicalSpace G]
...
 /-- Every Lie group trivially admits a Lie group structure. -/
 @[category API, AMS 22]
 theorem admitsLieGroupStructure_of_lieGroup
-    [ChartedSpace (EuclideanSpace ℝ (Fin n)) G] [LieGroup (𝓡 n) ⊤ G] :
+    [T2Space G] [SecondCountableTopology G]
+    [ChartedSpace (EuclideanSpace ℝ (Fin n)) G]
+    [IsManifold (𝓡 n) ω G] [LieGroup (𝓡 n) ω G] :
     AdmitsLieGroupStructure G :=
-  ⟨n, inferInstance, inferInstance⟩
+  ⟨n, ⟨{ carrier := G, equiv := ContinuousMulEquiv.refl G }⟩⟩
```
Deicyde: why can't this still be by `inferInstance`, we should try to make this line better

FormalConjectures/HilbertProblems/5.lean
```diff
@@ -40,28 +42,48 @@ variable {G : Type*} [Group G] [TopologicalSpace G]
...
 /-- A group admitting a Lie group structure is locally compact. -/
 @[category API, AMS 22]
 theorem locallyCompact_of_admitsLieGroupStructure
     (h : AdmitsLieGroupStructure G) : LocallyCompactSpace G := by
-  obtain ⟨k, cs, _⟩ := h
-  haveI := cs
+  obtain ⟨k, ⟨p⟩⟩ := h
+  letI := p.topologicalSpace
+  letI := p.group
+  letI := p.chartedSpace
```
Deicyde: Remove `letI` spam please try to find out why these don't synthesize

FormalConjectures/HilbertProblems/5.lean
Deicyde: Correction: I reverted the same-carrier class change in 3988f53. That approach reintroduced the original semantic bug: it required the selected `LieGroup` atlas on `G` instead of presenting `G` as continuously isomorphic to a separately structured Lie group. The explicit isomorphic-carrier witness is restored; the remaining `Nonempty` packaging cannot simply be discharged by `inferInstance` without a semantics-preserving presentation instance.

FormalConjectures/HilbertProblems/5.lean
Deicyde: Correction: I reverted this cleanup in 3988f53 because collapsing the presentation onto `G` was not semantics-preserving. The `letI` assignments install the topology, group, and atlas belonging to the separate carrier before transferring local compactness back along the continuous group equivalence. I am leaving this thread open rather than replacing that correct formulation with the invalid same-carrier version.

mo271:
@Paul-Lez do you want to review this?

Paul-Lez:
Hello yes! Sorry I've been a little slow on this - will do it this week hopefully (if I don't, feel free to nag me)

Paul-Lez (approved):
LGTM, thanks!

Paul-Lez (approved):

PR #4604, opened 2026-07-24
Disprove Green31 Sidon and Erdos128 literal statements

Kuberwastaken:
This PR marks two literal formal statements as false.

1. `Green31.green_31.variants.sidon_01n`

The formal predicate `IsSidon` includes diagonal pairs. In `𝔽₂ n`, every diagonal sum is zero, so `x + x = y + y` forces any literal Sidon finset to have at most one element. Therefore the claimed eventual lower bound by `((2 : ℝ) ^ n) ^ 0.51` is false.

External Lean proof artifact:
https://github.com/Kuberwastaken/formal-conjectures-counterexamples/blob/d2af9fe927ef93d6402f4f7dbb3943898a255047/lean/Green31F2SidonCheck.lean#L44-L61

2. `Erdos128.erdos_128`

The literal formal statement has the triangle conclusion inside the implication for each single dense majority subset. `K₂`, with `Vprime` equal to the full vertex set, satisfies the displayed premises but is triangle-free.

External Lean proof artifact:
https://github.com/Kuberwastaken/formal-conjectures-counterexamples/blob/d2af9fe927ef93d6402f4f7dbb3943898a255047/lean/Erdos128Check.lean#L20-L45

Related: #4501 already identifies this Erdős 128 formalization issue and proposes a corrected intended statement. This PR records the current literal statement as false.

Scope: these are corrections to the current literal formal statements, not claims about the intended informal conjectures.

PR #4605, opened 2026-07-24
fix(GreensOpenProblems/31): formalise binary Sidon sums

Kuberwastaken:
The two changes from #4604 are split as requested. This PR fixes only the Green 31 binary Sidon variants.

The intended binary B₂ notion uses ordinary coordinatewise sums in `{0,1,2}^n`. The existing `IsSidon` application instead uses addition in `𝔽₂^n`, where every diagonal sum is zero, making the displayed asymptotic statements inconsistent with the cited binary B₂ literature.

This introduces `IsBinarySidon`, comparing coordinatewise sums through the canonical representatives in `ℕ`, and uses it in both the open `0.51` question and the related CLZ01 upper-bound variant.

Validation:
- `git diff --check`

Split from and supersedes the Green 31 portion of #4604.

### DIFF
--- a/FormalConjectures/GreensOpenProblems/31.lean
+++ b/FormalConjectures/GreensOpenProblems/31.lean
@@ -132,13 +132,22 @@ theorem green_31.variants.abelian : answer(sorry) ↔
   sorry
 
 /--
+A set of binary vectors is Sidon if its pairwise sums, taken coordinatewise in `ℕ`, determine the
+unordered pair of summands.
+-/
+def IsBinarySidon {n : ℕ} (S : Set (𝔽₂ n)) : Prop :=
+  ∀ ⦃a b c d : 𝔽₂ n⦄, a ∈ S → b ∈ S → c ∈ S → d ∈ S →
+    (fun i ↦ (a i).val + (b i).val) = (fun i ↦ (c i).val + (d i).val) →
+      (a = c ∧ b = d) ∨ (a = d ∧ b = c)
+
+/--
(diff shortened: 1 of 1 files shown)


williamjblair:
Yeah this is right, and slightly worse than the description says. `x + x = 0` in `𝔽₂ n`, so `a + a = b + b` for any two elements, and `IsSidon` collapses to "at most one element":

```lean
example {n : ℕ} (S : Set (𝔽₂ n)) (hS : IsSidon S) : S.Subsingleton := by
  intro a ha b hb
  have h : a + a = b + b := by
    have : ∀ x : 𝔽₂ n, x + x = 0 := fun x => by ext i; simp [CharTwo.add_self_eq_zero]
    rw [this a, this b]
  rcases hS a ha b hb a ha b hb h with ⟨h1, -⟩ | ⟨h1, -⟩ <;> exact h1
```

So both statements on main are incorrect:

- `sidon_01n_clz01` is `research solved`, and `C = 1` proves it. Card is at most one and `(2^n)^0.5753 ≥ 1`, so Sidon sets never come into it.
- `sidon_01n` is `research open`, but the answer is no. Subsingletons can't reach `(2^n)^0.51`.

Your `IsBinarySidon` doesn't have the problem: `{0, 1} ⊆ 𝔽₂ 1` satisfies it, checked by `decide`. You could keep the subsingleton lemma in the file as a test, so the reason for the change doesn't only live in this PR description. Also, looks like CI is stuck, you may have to rebase with main

mo271 (approved):
Thanks, LGTM!

PR #4606, opened 2026-07-24
fix(ErdosProblems/128): quantify all dense induced subgraphs

Kuberwastaken:
The two changes from #4604 are split as requested. This PR fixes only the Erdős 128 quantifier placement.

The source assumes the density bound for every sufficiently large induced subgraph. The existing statement instead quantifies one subset outside the implication, so a single dense subset can trigger the triangle conclusion. This moves `V′` into the antecedent and leaves the problem open.

Validation:
- `git diff --check`

Split from and supersedes the Erdős 128 portion of #4604.


mo271:
thanks!
Closing as it seems to be a duplicate of #4501

PR #4617, opened 2026-07-24
fix(Wikipedia): require induced snake paths

### LINKED ISSUE #4521: fix(Wikipedia/SnakeInTheBox): require an induced path
The definition `IsSnakeInGraphOfLength` requires `G′.IsInduced` and identifies `G′.verts` with the support of a path, but it never identifies the edges of `G′` with the edges of that path. Consequently, a path with chords is accepted: `G′` contains every ambient edge among the path vertices while `P` need only be a simple path through them. In particular, a Hamiltonian Gray-code path through a hypercube can be paired with `G′ = ⊤`.

The source statement asks for an induced path. Requiring `G′ = P.toSubgraph` makes `G′.IsInduced` assert exactly that the path has no ambient chords.

References:
- https://en.wikipedia.org/wiki/Snake-in-the-box
- https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/SimpleGraph/Subgraph.html

kernelpanic888:
## Summary
- require the witness subgraph in `IsSnakeInGraphOfLength` to equal `P.toSubgraph`
- preserve `G'.IsInduced` so the definition now excludes ambient chords
- update the zero-dimensional example to use the empty path's own subgraph

## Why
Equality of vertex sets alone allowed a path with chords: `G'` could contain every ambient edge among the path vertices while `P` used only a subset. Identifying `G'` with `P.toSubgraph` makes the existing induced-subgraph condition express the intended induced-path condition.

## Verification
- `lake build FormalConjectures.Wikipedia.SnakeInTheBox`

Closes #4521.


kernelpanic888:
@googlebot rescan

mo271 (approved):
Thanks, LGTM!

Issue #1849, opened 2026-01-22
Some conjectures in general topology


felixpernegger:
### What is the conjecture
A nice list of open 20 problems can be found in
https://www.math.md/files/basm/y2013-n2-3/y2013-n2-3-(pp37-46).pdf.pdf

As far as I can tell (by going over citations in Google scholar), so far all are still open except Problem 11, which has been solved in
https://arxiv.org/pdf/2006.12675

Like usual in set theory related topics, many of the problems are known to be true under CH or some other additional axiom. A formalisation should probably reflect that.

Problem 1 is PR #1810.

I think this would be nice to have, as right now the repository basically only contains one conjecture in topology at all (#1776).

### Prerequisites needed
Most likely all problems need some sort of additional definition, but this is typically a fairly easy task.

### [AMS categories](https://github.com/google-deepmind/formal-conjectures/labels?q=ams-)

* ams-54

### Choose either option
- [ ] I plan on adding this conjecture to the repository
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

Paul-Lez:
Very nice! I'd suggest opening sub-issues so that people can claim the problems individually (and we can track progress on this issue)

felixpernegger:
> Very nice! I'd suggest opening sub-issues so that people can claim the problems individually (and we can track progress on this issue)

I dont think I have perms to create subissues directly?

mo271:
While we could use subissues, I'd prefer just normal issue and then we can create a milestone for the collection, like we have for Ben Green's problem for instance

felixpernegger:
> While we could use subissues, I'd prefer just normal issue and then we can create a milestone for the collection, like we have for Ben Green's problem for instance

I think a milestone sounds great! Maybe 20 problems is a bit little though?
In any case, I think milestones can bring down the current bias towards number theory and combinatorics (80.1%) a little bit, since they set clear goals to work towards.
A similar but *much* longer list is for example
https://staff.fnwi.uva.nl/j.vanmill/papers/papers1990/opit.pdf

Paul-Lez:
I'll claim problems 5 - 7 (the ones in the _Symmetrizable spaces_  section)

felixpernegger:
> I'll claim problems 5 - 7 (the ones in the _Symmetrizable spaces_  section) 

FYI I checked this before and if I remember correctly one or two of those actually have a simple counterexample

I recommed you to use the "Explore" tab of this website to check:
[...]

PR #4619, opened 2026-07-25
feat(Paper): formalize weak first-countability Problem 4

kernelpanic888:
## Summary

- define countable Souslin number for topological spaces
- prove the separable-space API test using Mathlib
- formalize Problem 4 from Arhangel’skii (2013) as an open conjecture

## Verification

```text
lake env lean FormalConjectures/Paper/WeaklyFirstCountable.lean
```

Progress on #1849.


FormalConjectures/Paper/WeaklyFirstCountable.lean
```diff
@@ -49,6 +49,12 @@ class WeaklyFirstCountableTopology (X : Type*) [TopologicalSpace X] : Prop where
       (∀ (x : X), Antitone (V x) ∧ ∀ (n : ℕ), x ∈ V x n) ∧
         ∀ O : Set X, IsOpen O ↔ ∀ x ∈ O, ∃ k : ℕ, V x k ⊆ O
 
+/-- A space has countable Souslin number if every pairwise-disjoint family of nonempty open
+sets has a countable index set. -/
+def HasCountableSouslinNumber (X : Type*) [TopologicalSpace X] : Prop :=
+  ∀ {ι : Type*} (U : ι → Set X) (a : Set ι), a.PairwiseDisjoint U →
```
felixpernegger: You should use `Type*` after forall quantifiers. Rather do : `(X : Type u)` for some universe and then `forall {iota : Type u} ...` 

But in any case, the iota here is a bit odd since you arent using it anyways. Why not define `U` as `Set (Set X)` directly? (and thus also inline the `a`)

FormalConjectures/Paper/WeaklyFirstCountable.lean
```diff
@@ -110,6 +123,12 @@ theorem CH.existsWeaklyFirstCountableCompactNotFirstCountable
       WeaklyFirstCountableTopology X ∧ CompactSpace X ∧ T2Space X ∧
         ¬ FirstCountableTopology X := by sorry
 
--- TODO: add Problem 4 in [Ar2013]
+/-- Problem 4 in [Ar2013]: If a Tychonoff weakly first-countable space has countable
+Souslin number, then does its cardinality not exceed the continuum? -/
+@[category research open, AMS 54]
+theorem cardinalMk_le_continuum_of_weaklyFirstCountable_of_countableSouslinNumber :
+    answer(sorry) ↔ ∀ (X : Type) (_ : TopologicalSpace X) (_ : T35Space X),
+      WeaklyFirstCountableTopology X → HasCountableSouslinNumber X → #X ≤ 𝔠 := by
```
felixpernegger: Why are you using forall quantifier for `T35Space` but not `WeaklyFirstCountableTopology`? Both ought to be fine but lets try to be consistent

FormalConjectures/Paper/WeaklyFirstCountable.lean
felixpernegger: 1. I would probably make this a class
2. You can make a `def` for the Souslin number more generally (see i.e. https://www.sciencedirect.com/science/article/pii/0166864194900469), then `HasCountableSouslinNumber` can be stated very nicely as `SouslinNumber X = aleph0` basically

FormalConjectures/Paper/WeaklyFirstCountable.lean
```diff
@@ -77,6 +83,13 @@ instance FirstCountableTopology.weaklyFirstCountableTopology (X : Type*) [Topolo
   · obtain ⟨n, hn⟩ := h x hx
     exact mem_of_superset (HasAntitoneBasis.mem (hU x) n) hn
 
+/-- Every separable space has countable Souslin number. -/
+@[category test, AMS 54]
+theorem hasCountableSouslinNumber_of_separable (X : Type*) [TopologicalSpace X]
+    [SeparableSpace X] : HasCountableSouslinNumber X := by
+  intro ι U a hdisj hopen hnonempty
+  exact hdisj.countable_of_isOpen hopen hnonempty
```
felixpernegger:
```suggestion
theorem hasCountableSouslinNumber_of_separable (X : Type*) [TopologicalSpace X]
    [SeparableSpace X] : HasCountableSouslinNumber X :=
  fun _ _ _ hdisj hopen hnonempty ↦ hdisj.countable_of_isOpen
```
this should work (didnt test)

felixpernegger (commented):
cool!

felixpernegger:
If you define the Souslin number more egenrally as I suggested, i can approve this :)

kernelpanic888:
Implemented the general invariant as suggested: `souslinNumber X` is the supremum of cardinalities of cellular families (with the standard `ℵ₀` lower bound), and `HasCountableSouslinNumber X` is now stated as `souslinNumber X = ℵ₀`. The separable-space instance is proved against this definition.

Validated with `lake env lean FormalConjectures/Paper/WeaklyFirstCountable.lean`.

FormalConjectures/Paper/WeaklyFirstCountable.lean
```diff
@@ -49,6 +51,19 @@ class WeaklyFirstCountableTopology (X : Type*) [TopologicalSpace X] : Prop where
       (∀ (x : X), Antitone (V x) ∧ ∀ (n : ℕ), x ∈ V x n) ∧
         ∀ O : Set X, IsOpen O ↔ ∀ x ∈ O, ∃ k : ℕ, V x k ⊆ O
 
+/-- A cellular family is a pairwise-disjoint collection of nonempty open sets. -/
+def IsCellularFamily (X : Type u) [TopologicalSpace X] (F : Set (Set X)) : Prop :=
+  F.PairwiseDisjoint id ∧ ∀ U ∈ F, IsOpen U ∧ U.Nonempty
+
+/-- The Souslin number of a topological space, with the standard convention that it is at
+least `ℵ₀`. -/
```
felixpernegger: Where is this "standard convention from" (genuine question)? The only reference I can find explicitly does not use this.

kernelpanic888: You're right—the cited definition uses the unmodified supremum. I removed the ℵ₀ lower bound and now define `HasCountableSouslinNumber X` by `souslinNumber X ≤ ℵ₀`. The separable-space instance is updated accordingly, and the file passes Lean.

felixpernegger (approved):

mo271 (approved):
Thanks, LGTM!

PR #4620, opened 2026-07-25
feat(Arxiv): formalize Talagrand's convexity problem

kernelpanic888:
## Summary

- define the canonical standard Gaussian measure on `Fin n → ℝ` as a finite product
- define an unambiguous recursive q-fold Minkowski sum
- formalize Talagrand's dimension-independent convexity statement
- mark the result `research solved` and cite Hua–Song–Tudose (arXiv:2605.10908), which resolves the problem affirmatively

Closes #1853

## Validation

- `lake env lean FormalConjectures/Paper/TalagrandConvexity.lean`

PR #4618, opened 2026-07-24
Formalize the Dittert conjecture

kernelpanic888:
Closes #2267

This PR formalizes the Dittert conjecture from arXiv:2312.00464. It defines the constrained matrix class `Kₙ`, row and column sums, Dittert's objective function, and the uniform matrix `Jₙ`, while reusing Mathlib's existing `Matrix.permanent`.

The research-open theorem states that `Jₙ` is the unique maximizer of the objective on `Kₙ` for every positive `n`. A one-dimensional API test checks the bespoke constraint definitions.

Validation:
- `lake env lean FormalConjectures/Arxiv/2312.00464/DittertConjecture.lean`

PR #4621, opened 2026-07-25
feat(Arxiv): formalize spectral set problems

### LINKED ISSUE #3520: Bounded nowhere dense spectral set existence
### What is the conjecture
Let $\Omega$ be a bounded nowhere dense set of positive measure in $\mathbb{R}^d$ $(d \ge 2)$. Can it be spectral?

* Answer known to be negative in dimension 1 (see Section 1.6 in paper below) but open in dimensions 2 and higher

* Problem 7.1 in https://arxiv.org/pdf/2209.04540 

* For definition of spectral, see https://en.wikipedia.org/wiki/Fuglede%27s_conjecture

### Prerequisites needed
Mathematical context that needs to be formalized so that this conjecture can be added.
Consider linking any open Mathlib issues / PRs here.

**Depends on** #3497 (Fuglede's conjecture, helper definitions/lemmas for spectral sets in FormalConjecturesForMathlib)

* ams-42
* ams-46

### Choose either option

### LINKED ISSUE #3523: Spectral Set, Product spectrality with convex factor
### What is the conjecture
If $\Omega = A \times B$ is spectral, where $A$ is a convex body in $\mathbb{R}^n$ and $B$ is a bounded measurable set in $\mathbb{R}^m$, must $B$ also be spectral? 

* Problem 7.2 in https://arxiv.org/pdf/2209.04540 
* It is already known that $A$ must be spectral by Theorem 1.6 in the above paper. The conjecture above itself about $B$ is known only for $n=1,2$.
* The general theorem that doesn't assume $A$ is a convex body has been proven to be false, by https://arxiv.org/abs/2410.19138v2. (Can perhaps add this as `research solved`) 
* For definition of spectral, see https://en.wikipedia.org/wiki/Fuglede%27s_conjecture 

### Prerequisites needed
Mathematical context that needs to be formalized so that this conjecture can be added.
Consider linking any open Mathlib issues / PRs here.

**Depends on** #3497 (Fuglede's conjecture, helper definitions/lemmas for spectral sets in FormalConjecturesForMathlib)

* ams-42
* ams-46

### Choose either option

kernelpanic888:
## Summary
- formalize Problems 7.1 and 7.2 from Kolountzakis–Lev–Matolcsi in the paper's Arxiv directory
- state the open cases with `answer(sorry) ↔ ...`
- split Problem 7.2 into solved `n = 1`, `n = 2` cases and open `n = 3`, general positive-dimensional cases
- reuse existing `isSpectral`, `ConvexBody`, boundedness, measurability, nowhere-density, and volume interfaces

Closes #3520
Closes #3523

## Validation
- `lake env lean FormalConjectures/Arxiv/2209.04540/SpectralSetsAndWeakTiling.lean`
- GitHub Actions `Build project`


williamjblair:
Problem 7.1 in #4622 and 7.2 here come from the same paper but are formalised differently. #4622 uses `answer(sorry) ↔ ...`; this one asserts the conjecture as a bare theorem. 7.2 reads as a question in the source ("must `B` be spectral?") rather than a claim, so `answer(sorry) ↔` seems the better fit, and it would match its neighbour.

Both shapes exist here (741 `research open` statements use `answer(...)`, 502 don't), so this is a consistency point rather than a rule.

For the batch as a whole, #4618 #4620 #4621 #4622 #4623 #4624 #4625: I built all seven against current main and they are clean, only the expected `sorry` warnings and nothing else. CI has run `Build project` on #4622 and #4624 only, so the other five have not been checked there.

FormalConjectures/Paper/SpectralSetProduct.lean
```diff
@@ -0,0 +1,50 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+import Mathlib.Analysis.Convex.Body
```
mo271:
```suggestion
```

FormalConjectures/Paper/SpectralSetProduct.lean
```diff
@@ -0,0 +1,50 @@
...
+
+namespace SpectralSetProduct
+
+/-- The product of two Euclidean sets, represented by consecutive coordinate blocks. -/
+def productSet {n m : ℕ} (A : Set (Fin n → ℝ)) (B : Set (Fin m → ℝ)) :
+    Set (Fin (n + m) → ℝ) :=
+  {x | (fun i ↦ x (Fin.castAdd m i)) ∈ A ∧ (fun j ↦ x (Fin.natAdd n j)) ∈ B}
+
+/-- Problem 7.2 in [KLM2023]: if `A × B` is spectral, where `A` is a convex body and
+`B` is bounded and measurable, then `B` is spectral. -/
```
mo271: please use latex markdown instead of backticks in doc comments

mo271 (commented):
Thanks for the contribution!

I suggest to put this in the `Arxiv` subdir!

Also: seems like for PRoblem 7.2, the cases n=1 and n=2 are sovled and the first open case is n=3. hence I suggest you add 4 theorems here: two `research solved` (case n=1 and case n=2) as well as two `research open`: n=3 and the case for general positive n.

Also, it might make sense to unify with the changes in #4622 in one file?

kernelpanic888:
Addressed the review in 2e4e80c: moved the formalization to the Arxiv subdirectory, consolidated Problems 7.1 and 7.2 in one file, removed the redundant import, switched doc comments to LaTeX, and split Problem 7.2 into solved n = 1 / n = 2 cases plus open n = 3 / general positive-dimensional cases. The updated module compiles locally, and the full GitHub Actions project build passes.

-awaiting-author
-paper

kernelpanic888:
-paper

mo271 (commented):
Instead of repeating the statement with different `n` four times or so, it might be better to have
```lean
∀ (n m : ℕ), 0 < n → 0 < m →
        ∀ (A : ConvexBody (Fin n → ℝ)) (B : Set (Fin m → ℝ)),
          Bornology.IsBounded B → MeasurableSet B →
            isSpectral (productSet (A : Set (Fin n → ℝ)) B) → isSpectral B 
```
as a definition with a suitable name and use that in the statements

kernelpanic888:
Addressed in a34a5e9: introduced `spectralProductImpliesRightSpectral n m` for the shared implication and rewrote all four statements to use it. `lake env lean FormalConjectures/Arxiv/2209.04540/SpectralSetsAndWeakTiling.lean` passes locally with only the expected `sorry` warnings.

mo271 (approved):
Thanks, LGTM!

PR #4630, opened 2026-07-25
chore: avoid heartbeat modifications

felixpernegger:
Remove heartbeat modifications in a file by simplifying before splitting up the goals.

This adds a few nonterminal simps but makes the file much faster; I'd say the tradeoff is worth it (if needed i can unsqueeze the simps of course; at leats the first one)


felixpernegger:
!bench
EDIT: would be cool if this was available :(

mo271 (approved):
nice, thanks!

PR #4635, opened 2026-07-26
feat: rephrase the Jacobian conjecture and add bivariate open case

Paul-Lez:
This PR rephrases the statement of the Jacobian conjecture in order to
- allow arbitrary fields in the disproof
- use the `answer(sorry)` convention
- use `DecidableEq` instead of `open Classical`
- have a separate predicate saying the conjecture holds for a given field and index type.

It also adds:
- the still open bivariate case
- the general disproof (which is a variant of the Alpolge one). This holds for any non-trivial commutative ring.
- one of the sanity checks that had been removed in the previous PR on this (this check needed a separate predicate asserting that the conjecture holds for a given index type to be able to state an implication).

Co-authored-by: Dean Cureton <me@deancureton.com>


mo271 (approved):
Thanks, LGTM!

PR #4638, opened 2026-07-26
feat(Kourovka): formalize Problem 1.74 (Tarski monster topologizability)

### LINKED ISSUE #1931: Kourovka 1.74 - Topologizability of Tarski Monsters
### What is the conjecture
Does there exist a Tarski monster group that admits a non-discrete Hausdorff group topology?

This is one of the questions to answer to resolve **Problem 1.74** from the **Kourovka Notebook** (Unsolved Problems in Group Theory), proposed by **V. P. Platonov** in the 1st edition (1965):
**"Describe all minimal topological groups, that is, non-discrete groups all of whose closed subgroups are discrete."**

*   **Terminology Note:** The term "minimal topological group" in this problem is defined specifically as *a non-discrete group having no proper non-discrete closed subgroups*. This should not be confused with the more common modern definition of "minimal topological groups" (attributed to R.M. Stephenson), which refers to groups whose topology is minimal among all Hausdorff group topologies. To avoid confusion, these are sometimes referred to as **Platonov minimal groups** or **groups with discrete closed subgroups**.

### Prerequisites needed (by Gemini)
Based on the current state of **Mathlib (Lean 4)**, **no definitions or structures are missing** to state this question.

While "Tarski Monster" is not a pre-bundled definition in Mathlib (like `Group` or `TopologicalSpace`), the primitives required to define it—subgroups, infinite types, prime numbers, and cardinality—are all available and robust.

### The Formal Statement in Lean 4

You can define the property of being a Tarski Monster and then state the existence question as follows:

```lean
import Mathlib.Topology.Algebra.Group.Basic
import Mathlib.Topology.Separation
import Mathlib.GroupTheory.Subgroup.Basic
import Mathlib.Data.Nat.Prime.Basic
import Mathlib.Data.Nat.Card

/-- 
Definition of a Tarski Monster Group:
An infinite group where every proper non-trivial subgroup has a fixed prime order p.
-/
def IsTarskiMonster (G : Type*) [Group G] : Prop :=
  Infinite G ∧ 
  ∃ p : ℕ, p.Prime ∧ 
  ∀ (H : Subgroup G), H ≠ ⊥ → H ≠ ⊤ → Nat.card H = p

/-- 
The Formal Question: 
Does there exist a Tarski monster group that admits a non-discrete Hausdorff group topology?
-/
def TarskiMonsterProblemStatement : Prop :=
  ∃ (G : Type*) 
    (grp : Group G) 
    (top : TopologicalSpace G),
    IsTarskiMonster G ∧ 
    TopologicalGroup G ∧ 
    T2Space G ∧ 
    ¬ DiscreteTopology G
```

### Explanation of Used Structures
1.  **`Nat.card H`**: This function (from `Mathlib.Data.Nat.Card`) returns the cardinality of a type as a natural number.
    *   If `H` is infinite, it returns `0`.
    *   If `H` is finite, it returns the exact size.
    *   In the definition above, `Nat.card H = p` (where `p` is prime and thus $>0$) implicitly enforces that `H` is a finite subgroup.
2.  **`H ≠ ⊥` and `H ≠ ⊤`**: `⊥` (bot) represents the trivial subgroup $\{1\}$, and `⊤` (top) represents the whole group $G$. This correctly targets "proper non-trivial subgroups."
3.  **`Infinite G`**: Ensures the group itself is not one of the finite cyclic groups of prime order (which would otherwise trivially satisfy the subgroup condition).
4.  **`IsCyclic` (Implicit)**: You do not need to explicitly require `IsCyclic H`. A group of prime order is automatically cyclic, which is a standard theorem in Mathlib (`IsCyclic.of_prime_card` or similar).

This formalization captures the exact algebraic constraints of a Tarski Monster combined with the topological constraints of the problem.

* ams-22

### Choose either option

walkthroughwonder:
Formalizes the Tarski-monster sub-problem of Kourovka Notebook Problem 1.74 (Platonov, 1st edition, 1965): does there exist a Tarski monster group that admits a non-discrete Hausdorff group topology?

Closes #1931.

- Defines `IsTarskiMonster`: an infinite group in which every non-trivial proper subgroup has order a fixed prime `p`, built from Mathlib primitives only (no new infrastructure needed).
- States the existence question as `answer(sorry)` with `@[category research open, AMS 20 22]`, following the conventions of `FormalConjectures/Kourovka/19_25.lean`.
- Typechecks locally with `lake env lean` (only the expected `sorry` warning).

### DIFF
--- a/FormalConjectures/Kourovka/1_74.lean
+++ b/FormalConjectures/Kourovka/1_74.lean
@@ -1,0 +1,55 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# Conjecture 1.74 (Tarski monster topologizability)
+
+by V. P. Platonov
+
+Problem 1.74 asks to describe all "minimal topological groups" in Platonov's
+sense: non-discrete Hausdorff topological groups all of whose proper closed
+subgroups are discrete. A natural test case: does there exist a Tarski monster
+group admitting a non-discrete Hausdorff group topology? A Tarski monster
+would be a minimal group in this sense, since all its proper subgroups are
+finite (hence discrete in any Hausdorff group topology).
+
+*Reference:* [The Kourovka Notebook](https://arxiv.org/abs/1401.0300v40)
+-/
+
+namespace Kourovka.«1.74»
+
+/--
+A Tarski monster group: an infinite group in which every non-trivial proper
+subgroup has order a fixed prime $p$.
+-/
+def IsTarskiMonster (G : Type*) [Group G] : Prop :=
+  Infinite G ∧ ∃ p : ℕ, p.Prime ∧
+    ∀ H : Subgroup G, H ≠ ⊥ → H ≠ ⊤ → Nat.card H = p
+
+/--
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

Issue #1940, opened 2026-01-30
Kourovka 2.5


rwst:
### What is the conjecture
**"According to Plotkin, a group is called an NR-group if the set of its nil-elements coincides with the locally nilpotent radical, or, which is equivalent, if every inner automorphism of it is locally stable. Can an NR-group have a nil-automorphism that is not locally stable?"**

**Terminology:**
*   **NR-group:** A group where the set of nil-elements (Engel elements) is exactly the locally nilpotent radical (Hirsch-Plotkin radical). Equivalently, every inner automorphism is locally stable.
*   **Nil-automorphism:** An automorphism $\phi$ of a group $G$ such that for every $g \in G$, there exists an integer $n > 0$ (depending on $g$) such that $[\dots[g, \phi], \dots, \phi] = 1$ ($n$ times).
*   **Locally stable automorphism:** An automorphism $\phi$ is locally stable if for every finitely generated $\phi$-invariant subgroup $H$, $\phi$ acts as a stable automorphism (i.e., stabilizes a normal series of $H$).

The problem essentially asks if the property of "inner automorphisms being locally stable" (which characterizes NR-groups) implies that *all* nil-automorphisms are locally stable in such groups.

See also #2014.

### Prerequisites needed
To state the problem from Kourovka Notebook 2.5 in Lean 4, the following definitions and structures are currently **missing** from Mathlib and would need to be defined manually:

1.  **Engel Elements (Nil-elements):**
    Mathlib has `IsNilpotent`, but it does not have a bundled definition for the set of **Right Engel elements** (often denoted $R(G)$ or the set of nil-elements). You would need to define:
    $$ \{ g \in G \mid \forall x \in G, \exists n \in \mathbb{N}, [x, {_n} g] = 1 \} $$
    where $[x, {_n} g]$ denotes the right-normed commutator $[...[[x, g], g], ... g]$.

2.  **Hirsch-Plotkin Radical (Locally Nilpotent Radical):**
    While `IsNilpotent` exists, the concept of **local nilpotency** (every finitely generated subgroup is nilpotent) and the **Hirsch-Plotkin radical** (the unique maximal normal locally nilpotent subgroup) are not defined as standard objects in Mathlib. You would need to define the radical as the join (supremum) of all normal locally nilpotent subgroups.

3.  **Stability of Automorphisms on Series:**
[...]

PR #4639, opened 2026-07-26
feat(Kourovka): formalize Problem 1.40 (products of normal Engel subgroups)

### LINKED ISSUE #1915: Kourovka 1.40
### What is the conjecture
**"Is a group a nilgroup if it is the product of two normal nilsubgroups?"**

The problem stated is **Problem 1.40** in the **Kourovka Notebook** (a famous collection of unsolved problems in group theory), posed by **Sh. S. Kemkhadze** (likely in 1965, the first issue).

Here, a **nilgroup** (or **Engel group**) is defined as a group where every element is an Engel element (specifically, a left Engel element, though for normal subgroups the distinction is often less critical in the context of the problem statement's intent). That is, for every $x, y \in G$, there exists an integer $n = n(x, y) \ge 1$ such that $[x, _n y] = [x, y, \dots, y] = 1$.

### Status of the Problem
The problem is discussed in research papers and remains **unsolved** (open) as of the recent editions of the Kourovka Notebook (including the 20th edition in 2022 and reports on the 21st edition in 2026).

While the **Fitting Theorem** guarantees that the product of two normal **nilpotent** subgroups is nilpotent, the analogue for **Engel groups** (nilgroups) is not known to be true or false in the general case.

### Key Points from Research
*   **Context:** The problem seeks to determine if the class of Engel groups is closed under forming products of normal subgroups. This is a fundamental question in the theory of generalized nilpotent groups.
*   **Partial Results:**
    *   If the group $G$ satisfies certain additional conditions (e.g., solubility or finiteness conditions), the answer is often positive because Engel groups with those conditions are locally nilpotent.
    *   For **bounded Engel groups** (where the Engel degree $n$ is bounded for the whole group), the problem is also related to whether the product of normal bounded Engel subgroups is bounded Engel (which is also a non-trivial question).
*   **Current State:** There is no known counterexample, nor is there a proof for the general case. It stands as one of the long-standing open questions from the very first issue of the Kourovka Notebook.

### Prerequisites needed
To state Kourovka Problem 1.40 in Lean 4 using Mathlib, you are primarily missing the specific definitions related to **Engel elements** and **Nilgroups** (also known as Engel groups). While Mathlib has extensive support for groups, normal subgroups, and commutators, it does not currently (as of early 2026) have the high-level predicates for "being an Engel element" or "being an Engel group" in its main `GroupTheory` library (though they exist for Lie Algebras).

Here are the specific missing definitions and how you would implement them to state the problem:

### Missing Definitions
1.  **Iterated Commutator (Left Engel sequence):** The sequence defined recursively by $[x, _0 y] = x$ and $[x, _{n+1} y] = [[x, _n y], y]$.
2.  **Engel Element:** An element $y$ is an Engel element if for every $x$, there exists some $n$ such that $[x, _n y] = 1$.
3.  **Nilgroup:** A group consisting entirely of Engel elements.

### Lean 4 Code to State the Problem
You can define the missing structures and state the problem as follows:

```lean
import Mathlib.Algebra.Group.Subgroup.Basic
import Mathlib.GroupTheory.Commutator.Basic
import Mathlib.GroupTheory.Subgroup.Pointwise

variable {G : Type*} [Group G]

-- 1. Missing Definition: Iterated Commutator
-- We define the left normed commutator sequence [x, _n y]
def iteratedCommutator (n : ℕ) (x y : G) : G :=
  match n with
  | 0 => x
  | k + 1 => ⁅iteratedCommutator k x y, y⁆

-- 2. Missing Definition: Engel Element
-- An element g is an Engel element if [x, _n g] = 1 eventually for all x.
def IsEngelElement (g : G) : Prop :=
  ∀ x : G, ∃ n : ℕ, iteratedCommutator n x g = 1

-- 3. Missing Definition: Nilgroup (Engel Group)
-- A group is a Nilgroup if all its elements are Engel elements.
-- (Note: We define this as a predicate on a Group type)
def IsNilgroup (G : Type*) [Group G] : Prop :=
  ∀ g : G, IsEngelElement g

-- 4. Stating Kourovka Problem 1.40
-- "Is a group a nilgroup if it is the product of two normal nilsubgroups?"
-- We use the coercion from Subgroup to Type to apply IsNilgroup to N and M.
-- The condition `N * M = ⊤` represents G being the product of N and M.
theorem Kourovka_1_40 (G : Type*) [Group G] (N M : Subgroup G)
  [N.Normal] [M.Normal]
  (h_prod : N * M = ⊤)       -- The product of subgroups is the whole group
  (h_N : IsNilgroup N)       -- N is a nilgroup
  (h_M : IsNilgroup M)       -- M is a nilgroup
  : IsNilgroup G :=          -- Question: Is G a nilgroup?
sorry
```

### Summary of what was strictly "missing":
*   `iteratedCommutator` (specifically the sequence $[x, _n y]$).
*   `IsEngelElement` predicate.
*   `IsNilgroup` predicate.

The concepts of **Groups**, **Normal Subgroups**, and **Subgroup Products** (`Subgroup.pointwiseMul` which allows `N * M`) are already present in Mathlib.

* ams-16

### Choose either option

walkthroughwonder:
Formalizes Kourovka Notebook Problem 1.40 (Kemkhadze, 1st edition, 1965): is a group a nilgroup if it is the product of two normal nilsubgroups?

Closes #1915.

- Defines `engelCommutator` (the iterated commutator $[x, {}_n y]$), `IsEngelElement`, and `IsEngelGroup` — none of these exist in Mathlib's group theory library yet.
- Since $H$ and $K$ are normal, the product $HK$ equals the join $H \sqcup K$, so the hypothesis "$G$ is the product of $H$ and $K$" is stated as `H ⊔ K = ⊤`.
- States the question as `answer(sorry)` with `@[category research open, AMS 20]`, following `FormalConjectures/Kourovka/19_25.lean`.
- Typechecks locally with `lake env lean` (only the expected `sorry` warning).

Note: the same Engel machinery is needed for Kourovka 2.5 (#1940), which I plan to tackle next if this shape looks right.

### DIFF
--- a/FormalConjectures/Kourovka/1_40.lean
+++ b/FormalConjectures/Kourovka/1_40.lean
@@ -1,0 +1,67 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# Conjecture 1.40
+
+by Sh. S. Kemkhadze
+
+Is a group a nilgroup if it is the product of two normal nilsubgroups?
+
+Here a nilgroup (Engel group) is a group in which every element is an Engel
+element. This is the Engel-group analogue of Fitting's theorem, which
+guarantees that the product of two normal nilpotent subgroups is nilpotent.
+
+*Reference:* [The Kourovka Notebook](https://arxiv.org/abs/1401.0300v40)
+-/
+
+namespace Kourovka.«1.40»
+
+variable {G : Type*} [Group G]
+
+/-- The iterated commutator $[x,\, {}_n y]$, defined by $[x,\, {}_0 y] = x$ and
+$[x,\, {}_{n+1} y] = [[x,\, {}_n y], y]$. -/
+def engelCommutator (x y : G) : ℕ → G
+  | 0 => x
+  | n + 1 => ⁅engelCommutator x y n, y⁆
+
+/-- An element $y$ of a group $G$ is a (left) Engel element if for every
+$x \in G$ there is some $n$ with $[x,\, {}_n y] = 1$. -/
+def IsEngelElement (y : G) : Prop :=
+  ∀ x : G, ∃ n : ℕ, engelCommutator x y n = 1
+
+/-- A nilgroup (Engel group) is a group in which every element is an Engel
+element. -/
+def IsEngelGroup (G : Type*) [Group G] : Prop :=
+  ∀ y : G, IsEngelElement y
+
+/--
+Is a group a nilgroup if it is the product of two normal nilsubgroups?
+
+Since $H$ and $K$ are normal, the product $HK$ coincides with the join
+$H \sqcup K$, so "$G$ is the product of $H$ and $K$" is stated as
+$H \sqcup K = G$.
+-/
+@[category research open, AMS 20]
+theorem kourovka.«1.40» : answer(sorry) ↔
+    ∀ (G : Type) [Group G] (H K : Subgroup G),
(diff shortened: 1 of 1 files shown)


williamjblair:
Builds clean here against current main, one `sorry` warning for the open statement and nothing else.

The definitions match the standard left-Engel ones: `engelCommutator x y 0 = x` with `⁅·, y⁆` iterated, and `IsEngelElement y` quantifying over all `x`. Using `H ⊔ K = ⊤` for the product is the right move given both are normal, and the docstring says why.

mo271 (approved):
Thanks, LGTM!

PR #4641, opened 2026-07-26
feat(Paper/BranchingVAS): Decidability of BVAS Reachability

### LINKED ISSUE #4640: Decidability of BVAS Reachability
### What is the conjecture
It has been an open question for some time whether reachability in branching vector addition systems is decidable. This is equivalent to the question of whether entailment in multiplicative-exponential linear logic (MELL) is decidable.

A *branching vector addition system* (BVAS) of dimension `d` is given by a finite
list of *axioms*, a finite list of *unary rules*, and a finite list of *binary
rules*, each of which is a vector in `ℤ^d`. A *configuration* is a vector in `ℕ^d`,
which here is represented as a vector `v ∈ ℤ^d` subject to `0 ≤ v`. The set of
*reachable* configurations is defined inductively:

* every axiom that is a valid configuration (i.e. lies in `ℕ^d`) is reachable;
* if `v` is a reachable configuration, `r` is a unary rule and `v + r ∈ ℕ^d`, then
  `v + r` is reachable;
* if `v₁, v₂` are both reachable configurations, `r` is a binary rule and
  `v₁ + v₂ + r ∈ ℕ^d`, then `v₁ + v₂ + r` is reachable.

Modelling axioms as vectors in `ℤ^d` and only counting the non-negative ones as
reachable yields the same set of reachable configurations as the usual definition
in which axioms are required to lie in `ℕ^d`.

Branching vector addition systems are distinguished from ordinary vector addition systems (VAS) by allowing binary rules. VAS reachability is known to be decidable.

*References:*
- [The covering and boundedness problems for branching vector addition systems](https://drops.dagstuhl.de/entities/document/10.4230/LIPIcs.FSTTCS.2009.2317)
  (*Stéphane Demri, Marcin Jurdziński, Oded Lachish, Ranko Lazić*, FSTTCS 2009)
  Gives a similar definition of BVAS, and compares it to related equivalent definitions.
- [On the Reachability Problem for Two-Dimensional Branching VASS](https://drops.dagstuhl.de/entities/document/10.4230/LIPIcs.MFCS.2025.22)
  by *Clotilde Bizière, Thibault Hilaire, Jérôme Leroux, Grégoire Sutre*, MFCS 2025, which
  settles the two-dimensional case and states that "the decidability status of the reachability
  problem for BVASS remains open in higher dimensions".
- [The General Vector Addition System Reachability Problem by Presburger Inductive
  Invariants](https://arxiv.org/abs/1009.1076) by *Jérôme Leroux* (2010), for the decidability
  of reachability for ordinary VAS.
- [Solving the Reachability Problem for Branching Vector Addition Systems via Semilinear
  Inductive Invariants](https://arxiv.org/abs/2607.09558) by *Clotilde Bizière, Jérôme Leroux,
  Grégoire Sutre* (2026), a recent preprint claiming a positive resolution to the conjecture:
  reachability for branching vector addition systems is decidable.

### Prerequisites needed
The definition of BVAS reachability, but it has a very short formalization given what is in Mathlib already.

* ams-68 ams-03

### Choose either option

jcreedcmu:
Fixes #4640.


jcreedcmu:
I made a judgment call to consider this `category research open` since there is a [claimed resolution in an ArXiv paper](https://arxiv.org/abs/2607.09558) (cited in issue and in the doc comment in the PR above), and I didn't know if that met the bar of "widely accepted by experts in the field" per the guidelines CONTRIBUTING.md. Happy to change if the maintainers judge differently!

williamjblair:
Builds clean here against current main, one `sorry` warning for the open statement and nothing else.

`answer(sorry) ↔ ∀ {d}, ComputablePred ...` reads right for a decidability question, and `reachable_of_mem_axioms` pins that the base case needs the axiom to be non-negative.

FormalConjectures/Paper/BranchingVAS.lean
```diff
@@ -0,0 +1,163 @@
...
+    (hmem : v ∈ b.axioms) (hcfg : 0 ≤ v) : b.Reachable v :=
+  .base hmem hcfg
+
+/-- Reachable vectors are always non-negative. -/
+@[category test, AMS 3 68]
+theorem reachable_imp_pos {d : ℕ} (v : Fin d → ℤ) (b : Bvas d) :
+    b.Reachable v → 0 ≤ v := by
+  intro h; cases h; all_goals assumption
+
+/-- A small BVAS of dimension3 used for a test below. -/
```
mo271:
```suggestion
/-- A small BVAS of dimension 3 used for a test below. -/
```

mo271:
> I made a judgment call to consider this `category research open` since there is a [claimed resolution in an ArXiv paper](https://arxiv.org/abs/2607.09558) (cited in issue and in the doc comment in the PR above), and I didn't know if that met the bar of "widely accepted by experts in the field" per the guidelines CONTRIBUTING.md. Happy to change if the maintainers judge differently!

Since this is very new and not peer review, it would be a close call, but I would tend to marking this `research solved` with mentioned some caveat in the docstring (that the result is quite recent, not yet peer reviewed or whatever), unless there is a strong reason why you would believe there is a problem with the proof in the preprint.

mo271 (commented):
Thanks, LGTM modulo the categorization (see my reply above)

jcreedcmu:
Sounds good, changed category to `solved`. I do not have any specific reason to believe there is a problem with the proof in the preprint. Thanks for the review!

mo271 (approved):
Thanks, LGTM!

PR #4644, opened 2026-07-27
feat(Mathoverflow/507128): mark as formally solved

KitaKen1:
I present a Lean proof for MathOverflow 507128 (located at `FormalConjectures/Mathoverflow/507128.lean`) and propose changing the status of this Formal Conjectures entry from "open" to "solved."

The Lean proof provided is by KitaKen1 (Kenta Kitamura).

GitHub: https://github.com/KitaKen1/mo507128-lean
Main Lean proof: https://github.com/KitaKen1/mo507128-lean/blob/e9507429c01c4288089e4af1c92a03b7d1e17f74/MO507128.lean
Lean4Web: https://live.lean-lang.org/#project=MathlibDemo&url=https%3A%2F%2Fraw.githubusercontent.com%2FKitaKen1%2Fmo507128-lean%2Fmain%2FMO507128_Lean4Web.lean

AI Usage Disclosure: This formalization was assisted by ChatGPT 5.6 sol and Codex GPT 5.6 sol with xhigh reasoning.

### DIFF
--- a/FormalConjectures/Mathoverflow/507128.lean
+++ b/FormalConjectures/Mathoverflow/507128.lean
@@ -24,10 +24,18 @@ asked by user [*Junyan Xu*](https://mathoverflow.net/users/3332/junyan-xu)
 
 namespace Mathoverflow507128
 
-/-- There exists a proper ideal `I` in a (commutative) total ring `R` of fractions that is an
-invertible module. If `I ⊊ R` is such an example, `I` must have infinite order in the Picard group,
-and `R` must not be Noetherian (otherwise it must be semi-local and therefore have trivial Picard
-group). -/
+/-- There exists a proper ideal `I` in a (commutative) total ring `R` of fractions that is an invertible module.
+If `I ⊊ R` is such an example, `I` must have infinite order in the Picard group.
+Moreover, `R` must not be Noetherian (otherwise it must be semi-local and therefore have trivial Picard group).
+
+The linked formal proof gives the following explicit construction.
+
+Let `D = ℂ[X, Y] / (Y² - X³)` be the coordinate ring of the cuspidal cubic.
+Let `P = (X - 1, Y - 1)`, which is an invertible ideal.
+Let `M` be the direct sum of the evaluation fibres at cusp parameters `r ≠ 1`.
+Form the idealization `R = D ⋉ M`.
+The desired ideal is the range of the multiplication map `R ⊗[D] P → R`.
+-/
 @[category research solved, AMS 13,
   formal_proof using lean4 at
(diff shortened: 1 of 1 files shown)


FormalConjectures/Mathoverflow/507128.lean
```diff
@@ -28,7 +28,9 @@ namespace Mathoverflow507128
 invertible module. If `I ⊊ R` is such an example, `I` must have infinite order in the Picard group,
 and `R` must not be Noetherian (otherwise it must be semi-local and therefore have trivial Picard
 group). -/
```
mo271: can you mention the explicit example?
From you linked repo, I found 
```
The concrete base ring is the coordinate ring
`D = ℂ[X,Y]/(Y²-X³)` of the cuspidal cubic. The ideal
`P = (X-1, Y-1)` is invertible. A detector module is assembled from evaluation
fibres away from the point corresponding to `P`, and the idealization
`R = D ⋉ M` is formed. The formal proof establishes:

1. every nonunit of `R` is a zero divisor, hence `IsFractionRing R R`;
2. the ideal induced by `P` is proper;
3. multiplication on the detector module has the injectivity needed to show
   that this ideal is an invertible `R`-module.
```

mo271 (commented):
Thanks for the contribution. 
@alreadydone can you check that the link solution is an answer to your question?

FormalConjectures/Mathoverflow/507128.lean
KitaKen1: Thank you for the review.
I have added a concrete example to the docstring.

alreadydone:
> @alreadydone can you check that the link solution is an answer to your question?

~~The solution is rather long and I can't immediately check the validity~~ (Update: the 735-line file on Lean Web editor does look valid), but coincidentally Patricia Purtill emailed me on July 24 with a solution in collaboration with GPT-5.6, which I [formalized with Aristotle](https://aristotle.harmonic.fun/dashboard/requests/a5a57b22-dd28-41f0-9a1e-cbd61740591b), so **it's fair to mark this as formally solved**. It's not an end-to-end formalization though, as it takes the existence of a Dedekind domain (which can be taken to be the coordinate ring of an elliptic curve) with an infinite order element in the class group as an assumption. The Weierstrass curve @KitaKen1 uses is singular (not elliptic), and the construction is probably a slight tweak while essentially the same.

I plan to post an answer on MO (since she declined to do so) and add the example to https://leanprover-community.github.io/mathlib4_docs/Counterexamples/InvertibleModuleNotIdeal.html.

mo271 (approved):
Thanks, LGTM!

PR #4659, opened 2026-07-29
feat(Paper/MonochromaticQuantumGraph): solve all-even integer cases

KitaKen1:
This PR marks the all-even integer cases in `MonochromaticQuantumGraph.lean` as solved, based on the following general theorem:

```lean
∀ N D : Nat, N ≥ 6 → Even N → D ≥ 3 →
  ¬ ∃ W : WeightsN N D ℤ, EqSystemN N D W
```

This settles all six integer declarations in `MonochromaticQuantumGraph.lean`; the six trinary-integer declarations follow immediately.

Note: This overlaps PR #4511 on the `N = 6, D = 3` integer case and its trinary corollary.

The Lean proof was developed and formalized by KitaKen1 (Kenta Kitamura).

Proof repository: https://github.com/KitaKen1/monochromatic-quantum-graphs-lean

FormalConjecturesWrappers.lean: https://github.com/KitaKen1/monochromatic-quantum-graphs-lean/blob/d3ed1892ef181f5f5f5d61d9b5817f05b53a6675/lean/QuantumLean/FormalConjecturesWrappers.lean

Lean4Web: https://live.lean-lang.org/#url=https%3A%2F%2Fraw.githubusercontent.com%2FKitaKen1%2Fmonochromatic-quantum-graphs-lean%2Frefs%2Fheads%2Fmain%2Flean4web%2FQuantumGraphAllEvenLean4Web.lean

Verification: the external proof contains no `sorry`, and its final axiom audit reports only `propext`, `Classical.choice`, and `Quot.sound`.

AI Usage Disclosure: This formalization is assisted by ChatGPT 5.6 sol and Codex GPT 5.6 sol with xhigh reasoning.

### DIFF
--- a/FormalConjectures/Paper/MonochromaticQuantumGraph.lean
+++ b/FormalConjectures/Paper/MonochromaticQuantumGraph.lean
@@ -655,75 +655,6 @@ theorem eqSystem4_no_solution_ge4_int :
 
 /-- For $N = 6$ and $D = 3$, does there exist no solution to the monochromatic quantum graph
 equation system over $\mathbb{Z}$? -/
-@[category research open, AMS 5 14 81]
-theorem eqSystem6_no_solution_d3_int :
-    answer(sorry) ↔
-      ¬ ∃ W : WeightsN 6 3 ℤ, EqSystemN 6 3 W := by
-  sorry
-
-/-- For $N = 6$ and $D = 5$, does there exist no solution to the monochromatic quantum graph
-equation system over $\mathbb{Z}$? -/
-@[category research open, AMS 5 14 81]
-theorem eqSystem6_no_solution_d5_int :
-    answer(sorry) ↔
-      ¬ ∃ W : WeightsN 6 5 ℤ, EqSystemN 6 5 W := by
-  sorry
-
-/-- For $N = 6$ and all $D \geq 3$, does there exist no solution to the monochromatic quantum graph
-equation system over $\mathbb{Z}$? -/
-@[category research open, AMS 5 14 81]
-theorem eqSystem6_no_solution_ge3_int :
-    answer(sorry) ↔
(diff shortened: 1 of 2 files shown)


mo271:
Thanks for the contribution!

The CI is only failing because one of the theorems is in the list  `FC100OpenSet1` and we keep track of how many of those are solved. Should be easy to fix by changing `94` to 93 in `FormalConjectures/Subsets/FC100OpenSet1.lean`


@MarioKrenn6240 can you take a look!

williamjblair:
Checked all twelve links.

Every URL is commit-pinned rather than `main`, and each `#L` range lands on the wrapper theorem for the declaration it annotates. Modulo whitespace, all twelve wrapper statements are character-for-character the statements in `MonochromaticQuantumGraph.lean`.

The proof repo depends on formal-conjectures through lake at `f7349f32`, so `WeightsN` and `EqSystemN` are this repo's own definitions rather than a restatement that could drift. That revision is an ancestor of `main` and the file is identical between the two, and it pins the same Mathlib revision as here.

No `sorry`, `axiom` or `native_decide` across the 25 files under `lean/`; four greps look like hits and all four are docstrings. `AxiomAudit.lean` covers all twelve wrappers. The `FC100OpenSet1` counts match what @mo271 asked for.

The `lean4web` export doesn't compile against the Mathlib the proof pins (36 errors at 18 sites, all `cases ... with | head | tail` where that Mathlib gives an `Or`). Those proofs aren't under `lean/` at all, so it looks like a stale parallel artifact. Nothing FC links to is affected, but anyone clicking the Lean4Web link to check will hit errors.

I verified statements, links and axioms by reading, not by building the `lean/` tree.

mo271 (approved):
Thanks again!

Issue #461, opened 2025-07-28
Erdős Problem 265


mo271:
### What is the conjecture
    
https://www.erdosproblems.com/265

  How fast can $a_n\to \infty$ grow if 
$$\sum\frac{1}{a_n}$$ and $$\sum\frac{1}{a_n-1}$$
are both rational?

Status: open

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

cosmicgenius:
Hi! I would like to take this conjecture!

What's the general approach to formalizing something of the form "how fast can X grow?" or in general anything that says "analyze/find asymptotics for this sequence" or something? I know there are concrete asymptotics conjectured on the erdosproblems site. Should we just formalize those, or is there something like `answer(sorry)` that we can do here?

papanokechi:
@cosmicgenius — you asked this in September and it looks like it never got a reply, so here is an attempt at an answer, plus an offer.

**Are you still working on this?** If so, please treat everything below as notes and take whatever is useful; I'll stand down. If not, I have a file that builds and I'm happy to open the PR.

## On "how fast can X grow?"

For this problem specifically the source page resolves the ambiguity itself. It ends with:

> It remains open whether one can achieve $\limsup a_n^{1/2^n} > 1$.

So the vague "how fast can it grow" has a precise open decision problem attached, and that is what I formalised as the primary statement, in the `answer(sorry) ↔ P` style CONTRIBUTING recommends for questions:

```lean
theorem erdos_265 : answer(sorry) ↔
    ∃ a : ℕ → ℕ, IsRationalPairSequence a ∧ GrowthExceedsOne a
```

For the "find the asymptotics" reading I added a companion declaration taking the supremum over achievable doubly-exponential rates, since the page says "the remaining question is the precise exponent possible":

```lean
theorem erdos_265.variants.sup_rate : sSup AchievableRates = answer(sorry)
```

The repo accepts multiple formalisations per problem and these answer genuinely different questions, so I don't think you have to choose.

## One trap worth flagging, because it silently inverts the statement

Do **not** write the growth condition as `1 < Filter.limsup (fun n => (a n : ℝ) ^ (1 / 2 ^ n)) atTop`.

[...]

PR #4660, opened 2026-07-29
Erdos Problem 265: sequences with two rational reciprocal sums

papanokechi:
Towards #461 (open, unassigned, no file in the repo; see my comment there — @cosmicgenius, happy to defer if you're still working on it).

Formalises [Erdős Problem 265](https://www.erdosproblems.com/265), which the site currently lists as having no formalised statement.

> Let $1 \leq a_1 < a_2 < \cdots$ be an increasing sequence of integers. How fast can $a_n \to \infty$ grow if $\sum \frac{1}{a_n}$ and $\sum \frac{1}{a_n - 1}$ are both rational?

## Two formalisations, because the page asks two questions

The page ends with a precise open decision problem — *"It remains open whether one can achieve $\limsup a_n^{1/2^n} > 1$"* — and also says *"the remaining question is the precise exponent possible"*. Those are different questions, so both are stated:

| declaration | reading |
|---|---|
| `erdos_265` | is $\limsup a_n^{1/2^n} > 1$ achievable? (`answer(sorry) ↔ …`) |
| `erdos_265.variants.sup_rate` | `sSup AchievableRates = answer(sorry)` |

Reviewers should feel free to ask for one to be dropped.

## Why the growth condition is not written with `Filter.limsup`

The natural transcription `1 < Filter.limsup (fun n => (a n : ℝ) ^ (1 / 2 ^ n)) atTop` **inverts the statement**. `Filter.limsup f atTop` unfolds to `sInf {b | ∀ᶠ n, f n ≤ b}`; for a real sequence unbounded above that set is empty, and `Real.sInf_empty : sInf ∅ = 0`. So the limsup is the junk value `0` and the inequality evaluates to `False` exactly for the fastest-growing sequences — the ones the problem is about. I verified this in Lean rather than assuming it (`limsup (fun n => (n : ℝ)) atTop = 0` is provable).

Instead:

```lean
def GrowthExceedsOne (a : ℕ → ℕ) : Prop :=
  ∃ c : ℝ, 1 < c ∧ ∃ᶠ n in atTop, c ^ (2 ^ n) ≤ (a n : ℝ)
```

`∃ᶠ` and not `∀ᶠ`: `limsup > 1` is an infinitely-often condition, whereas `∀ᶠ n, c ≤ x n` gives `liminf ≥ c` and would formalise a strictly stronger, different conjecture.

`sSup AchievableRates` has the same latent hazard, which is why the folklore upper bound is stated in the file — without `BddAbove` the supremum would silently collapse to `0` too.

## A statement ambiguity, reported upstream rather than papered over

The literal wording allows $a_1 = 1$, making $\sum \frac{1}{a_n - 1}$ ill-defined; the stated example $a_n = \binom{n}{2}$ has the same problem at $n = 2$ and must be read from $n = 3$ (which is also where the clean values $\sum 1/a_n = 1$ and $\sum 1/(a_n - 1) = 11/9$ come from — both checked numerically and by telescoping).

The file requires `2 ≤ a 0` (the minimal hypothesis — `StrictMono` propagates it), and `∀ n, 2 ≤ a n` is a proved `@[category API]` lemma with no `sorry`, so the choice is visible rather than buried. Filed on the source database as [teorth/erdosproblems#359](https://github.com/teorth/erdosproblems/issues/359); happy to adjust here if the maintainers there settle it differently.

## Contents

- `IsRationalPairSequence` — the admissible sequences
- `two_le_of_isRationalPairSequence` — `@[category API]`, **proved**
- `cantor_example` — `@[category test]`
- `erdos_265`, `erdos_265.variants.sup_rate` — `@[category research open]`
- `variants.kovac_tao`, `variants.super_exponential` — `@[category research solved]`, [KoTa24] (arXiv:2406.17593); the $a_n^{1/n} \to \infty$ statement is flagged in its docstring as a *consequence* of their $\beta > 1$ theorem, not its literal form
- `variants.folklore_irrational`, `variants.folklore_upper_bound` — `@[category research solved]`

## Build

`lake build FormalConjectures.ErdosProblems.«265»` completes successfully on toolchain v4.27.0 with the repo's style linters enabled (`copyright`, `namespace`, `ams_attribute`, `category_attribute`, `moduleDocstring`, `latex_docstring`). Only the expected `declaration uses 'sorry'` warnings.

---

*Disclosure: this file and PR description were drafted with AI assistance and reviewed by me. Per the repo's Mathlib-aligned AI conventions, the load-bearing claims were checked by execution rather than assertion — the `limsup` collapse and the `∃ᶠ`/`∀ᶠ` distinction were proved in Lean, the Cantor constants were computed independently, and the reference text was read from the source page rather than recalled.*

### DIFF
--- a/FormalConjectures/ErdosProblems/265.lean
+++ b/FormalConjectures/ErdosProblems/265.lean
@@ -1,0 +1,168 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# Erdős Problem 265
+
+*Reference:* [erdosproblems.com/265](https://www.erdosproblems.com/265)
+
+Let $1 \leq a_1 < a_2 < \cdots$ be an increasing sequence of integers. How fast can
+$a_n \to \infty$ grow if $\sum \frac{1}{a_n}$ and $\sum \frac{1}{a_n - 1}$ are both
+rational?
+
+Two points of care in this formalisation.
+
+*The literal statement admits $a_1 = 1$*, which makes $a_1 - 1 = 0$ and the second
+series ill-defined. We therefore require $2 \leq a_0$; combined with `StrictMono` this
+gives $2 \leq a_n$ for all $n$. Cantor's example correspondingly begins at $n = 3$.
+This ambiguity has been reported upstream at
+[teorth/erdosproblems#359](https://github.com/teorth/erdosproblems/issues/359).
+
+*The growth condition is deliberately not stated as* $1 < \limsup a_n^{1/2^n}$ over
+$\mathbb{R}$. `Filter.limsup` unfolds to `sInf {b | ∀ᶠ n, f n ≤ b}`, and
+`Real.sInf_empty : sInf ∅ = 0`, so an unbounded sequence yields the junk value $0$ and
+the inequality would evaluate to `False` precisely for the fastest-growing sequences —
+inverting the intended meaning. We use the equivalent existential form with `∃ᶠ`
+(frequently): $\limsup > 1$ is an infinitely-often condition, so `∀ᶠ` would encode the
+strictly stronger $\liminf > 1$.
+-/
+
+open Filter
+
+open scoped Topology
+
+namespace Erdos265
+
+/--
+A strictly increasing sequence $a$ of integers with $2 \leq a_0$, such that both
+$\sum \frac{1}{a_n}$ and $\sum \frac{1}{a_n - 1}$ converge to rational numbers.
+-/
+def IsRationalPairSequence (a : ℕ → ℕ) : Prop :=
+  StrictMono a ∧
+    2 ≤ a 0 ∧
+    (∃ q : ℚ, ∑' n, (1 : ℝ) / (a n : ℝ) = q) ∧
+    (∃ q : ℚ, ∑' n, (1 : ℝ) / ((a n : ℝ) - 1) = q)
+
+/--
+$2 \leq a_0$ together with `StrictMono` already forces $2 \leq a_n$ everywhere, so
+taking the minimal hypothesis in `IsRationalPairSequence` loses nothing.
+-/
+@[category API, AMS 11]
+theorem two_le_of_isRationalPairSequence {a : ℕ → ℕ} (ha : IsRationalPairSequence a) :
+    ∀ n, 2 ≤ a n :=
+  fun n => le_trans ha.2.1 (ha.1.monotone (Nat.zero_le n))
+
+/--
+Cantor's example. The source page states it as $a_n = \binom{n}{2}$; we index from
+$n = 3$, i.e. $a_n = \binom{n+3}{2} = 3, 6, 10, \ldots$, because $\binom{2}{2} = 1$
+would make $a_n - 1 = 0$ (the ambiguity reported at teorth/erdosproblems#359).
+Here $\sum \frac{1}{a_n} = 1$, and since $\binom{n}{2} - 1 = \frac{(n-2)(n+1)}{2}$ the
+second series telescopes to $\sum \frac{1}{a_n - 1} = \frac{11}{9}$. Both are rational.
+-/
+@[category test, AMS 11]
+theorem cantor_example : IsRationalPairSequence (fun n : ℕ => (n + 3).choose 2) := by
+  sorry
+
+/--
+The growth condition $\limsup a_n^{1/2^n} > 1$, stated existentially to avoid the
+real-valued `limsup` junk value described in the module docstring.
+-/
+def GrowthExceedsOne (a : ℕ → ℕ) : Prop :=
+  ∃ c : ℝ, 1 < c ∧ ∃ᶠ n in atTop, c ^ (2 ^ n) ≤ (a n : ℝ)
+
+/--
+**Erdős Problem 265.** The precise question that remains open is whether the growth
+exponent can exceed $1$.
+-/
+@[category research open, AMS 11]
+theorem erdos_265 : answer(sorry) ↔
+    ∃ a : ℕ → ℕ, IsRationalPairSequence a ∧ GrowthExceedsOne a := by
+  sorry
+
+/-- The set of achievable doubly-exponential rates $\beta$. -/
+def AchievableRates : Set ℝ :=
+  {β : ℝ | 1 < β ∧ ∃ a : ℕ → ℕ, IsRationalPairSequence a ∧
+    Tendsto (fun n : ℕ => (a n : ℝ) ^ (1 / (β ^ n : ℝ))) atTop atTop}
+
+/--
+The folklore upper bound is what makes `sSup AchievableRates` meaningful rather than a
+junk value: without `BddAbove` the supremum would silently collapse to $0$, exactly as
+the naive `limsup` formulation does.
+-/
+@[category research solved, AMS 11]
+theorem achievableRates_bddAbove : BddAbove AchievableRates := by
+  sorry
+
+/--
+Companion supremum formulation of "how fast can it grow", for the reading of the problem
+that asks for the extremal rate rather than for a yes/no decision.
+-/
+@[category research open, AMS 11]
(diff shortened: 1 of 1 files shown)


papanokechi:
@googlebot I signed it!

FormalConjectures/ErdosProblems/265.lean
```diff
@@ -0,0 +1,168 @@
...
+This ambiguity has been reported upstream at
+[teorth/erdosproblems#359](https://github.com/teorth/erdosproblems/issues/359).
+
+*The growth condition is deliberately not stated as* $1 < \limsup a_n^{1/2^n}$ over
+$\mathbb{R}$. `Filter.limsup` unfolds to `sInf {b | ∀ᶠ n, f n ≤ b}`, and
+`Real.sInf_empty : sInf ∅ = 0`, so an unbounded sequence yields the junk value $0$ and
+the inequality would evaluate to `False` precisely for the fastest-growing sequences —
+inverting the intended meaning. We use the equivalent existential form with `∃ᶠ`
+(frequently): $\limsup > 1$ is an infinitely-often condition, so `∀ᶠ` would encode the
+strictly stronger $\liminf > 1$.
```
danielchin, line 42:
```suggestion
```

We'll reference the problem down below in the doc strings of the individual theorems so no need to have a large comment up here.

FormalConjectures/ErdosProblems/265.lean
```diff
@@ -0,0 +1,168 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# Erdős Problem 265
+
+*Reference:* [erdosproblems.com/265](https://www.erdosproblems.com/265)
```
danielchin, line 22:
```suggestion
*References:*
- [erdosproblems.com/265](https://www.erdosproblems.com/265)
- [ErGr80] Erdős, P. and Graham, R., Old and new problems and results in combinatorial number theory. Monographies de L'Enseignement Mathematique (1980).
- [Er88c] Erdős, P., On the irrationality of certain series: problems and results. New advances in transcendence theory (Durham, 1986) (1988), 102-109.
- [KoTa24] Kova\vC, V. and Tao T., On several irrationality problems for Ahmes series. arXiv:2406.17593 (2024).
```

Let's add all the references from the website up here.

FormalConjectures/ErdosProblems/265.lean
```diff
@@ -0,0 +1,168 @@
...
+$\mathbb{R}$. `Filter.limsup` unfolds to `sInf {b | ∀ᶠ n, f n ≤ b}`, and
+`Real.sInf_empty : sInf ∅ = 0`, so an unbounded sequence yields the junk value $0$ and
+the inequality would evaluate to `False` precisely for the fastest-growing sequences —
+inverting the intended meaning. We use the equivalent existential form with `∃ᶠ`
+(frequently): $\limsup > 1$ is an infinitely-often condition, so `∀ᶠ` would encode the
+strictly stronger $\liminf > 1$.
+-/
+
+open Filter
+
```
danielchin, line 46:
```suggestion
```

nit: let's remove this extra line

FormalConjectures/ErdosProblems/265.lean
```diff
@@ -0,0 +1,168 @@
...
+namespace Erdos265
+
+/--
+A strictly increasing sequence $a$ of integers with $2 \leq a_0$, such that both
+$\sum \frac{1}{a_n}$ and $\sum \frac{1}{a_n - 1}$ converge to rational numbers.
+-/
+def IsRationalPairSequence (a : ℕ → ℕ) : Prop :=
+  StrictMono a ∧
+    2 ≤ a 0 ∧
```
danielchin, line 57: I see in the top level comment you have this be 2 to avoid the case where the second sequence has a $a_n$ of 1, but I don't think this works for the first sequence because you're also excluding $a_n$ from being 1 in the first sequence.

FormalConjectures/ErdosProblems/265.lean
```diff
@@ -0,0 +1,168 @@
...
+theorem two_le_of_isRationalPairSequence {a : ℕ → ℕ} (ha : IsRationalPairSequence a) :
+    ∀ n, 2 ≤ a n :=
+  fun n => le_trans ha.2.1 (ha.1.monotone (Nat.zero_le n))
+
+/--
+Cantor's example. The source page states it as $a_n = \binom{n}{2}$; we index from
+$n = 3$, i.e. $a_n = \binom{n+3}{2} = 3, 6, 10, \ldots$, because $\binom{2}{2} = 1$
+would make $a_n - 1 = 0$ (the ambiguity reported at teorth/erdosproblems#359).
+Here $\sum \frac{1}{a_n} = 1$, and since $\binom{n}{2} - 1 = \frac{(n-2)(n+1)}{2}$ the
+second series telescopes to $\sum \frac{1}{a_n - 1} = \frac{11}{9}$. Both are rational.
```
danielchin, line 75: Keep docstrings as verbatim to the site as possible.

`Cantor observed that $a_n=\binom{n}{2}$ is such a sequence.`

FormalConjectures/ErdosProblems/265.lean
```diff
@@ -0,0 +1,168 @@
...
+-/
+@[category test, AMS 11]
+theorem cantor_example : IsRationalPairSequence (fun n : ℕ => (n + 3).choose 2) := by
```
danielchin, line 78: This seems to be a different sequence that the sequence that Cantor observed.

FormalConjectures/ErdosProblems/265.lean
```diff
@@ -0,0 +1,168 @@
...
+/--
+The growth condition $\limsup a_n^{1/2^n} > 1$, stated existentially to avoid the
+real-valued `limsup` junk value described in the module docstring.
+-/
+def GrowthExceedsOne (a : ℕ → ℕ) : Prop :=
+  ∃ c : ℝ, 1 < c ∧ ∃ᶠ n in atTop, c ^ (2 ^ n) ≤ (a n : ℝ)
+
+/--
+**Erdős Problem 265.** The precise question that remains open is whether the growth
+exponent can exceed $1$.
```
danielchin, line 90:
```suggestion
Let $1\leq a_1<a_2<\cdots$ be an increasing sequence of integers. How fast can $a_n\to \infty$ grow if\[\sum\frac{1}{a_n}\quad\textrm{and}\quad\sum\frac{1}{a_n-1}\]are both rational?
```

FormalConjectures/ErdosProblems/265.lean
```diff
@@ -0,0 +1,168 @@
...
+@[category test, AMS 11]
+theorem cantor_example : IsRationalPairSequence (fun n : ℕ => (n + 3).choose 2) := by
+  sorry
+
+/--
+The growth condition $\limsup a_n^{1/2^n} > 1$, stated existentially to avoid the
+real-valued `limsup` junk value described in the module docstring.
+-/
+def GrowthExceedsOne (a : ℕ → ℕ) : Prop :=
+  ∃ c : ℝ, 1 < c ∧ ∃ᶠ n in atTop, c ^ (2 ^ n) ≤ (a n : ℝ)
```
danielchin, line 86: This can just be inlined into the problem if it's only being used in one place.

I'll also note that this doesn't actually look like it belongs as a property of the main Erdos 265 problem, but  it looks like a variant.

FormalConjectures/ErdosProblems/265.lean
```diff
@@ -0,0 +1,168 @@
...
+  sorry
+
+/--
+Erdős believed $a_n^{1/n} \to \infty$ to be possible. This follows from the doubly
+exponential construction of Kovač and Tao [KoTa24] recorded below, since
+$\beta^n / n \to \infty$ for any $\beta > 1$; it is a consequence of their theorem
+rather than its literal statement.
+
+[KoTa24] Kovač, V. and Tao, T., On several irrationality problems for Ahmes series.
+         arXiv:2406.17593 (2024).
```
danielchin, line 126:
```suggestion
```

FormalConjectures/ErdosProblems/265.lean
```diff
@@ -0,0 +1,168 @@
...
+    ∃ a : ℕ → ℕ, IsRationalPairSequence a ∧
+      Tendsto (fun n : ℕ => (a n : ℝ) ^ (1 / (n : ℝ))) atTop atTop := by
+  sorry
+
+/--
+Kovač and Tao [KoTa24] proved that such a sequence can grow doubly exponentially: there
+is some $\beta > 1$ with $a_n^{1/\beta^n} \to \infty$.
+
+[KoTa24] Kovač, V. and Tao, T., On several irrationality problems for Ahmes series.
+         arXiv:2406.17593 (2024).
```
danielchin, line 139:
```suggestion
```

danielchin (changes requested):
I did a partial review of this so far. It looks like `erdos_265` is combining multiple problems on the page and I request that you do a closer review of this yourself to make sure that it reflects the main Erdos Problem 265.

PR #4666, opened 2026-07-30
chore: remove maxRecDepth modifications

felixpernegger:
There are some more, which are kind of hard to avoid though


mo271 (approved):
Thanks, LGTM!

PR #4670, opened 2026-07-30
chore: remove an `erw`

felixpernegger:
`erw` is considered defeq abuse (i.e. is tracked as tech debt in mathlib).
As you can see from the diff, the only `erw` in this repository has a rather trivial solution


mo271 (approved):
Thanks, LGTM!

PR #4671, opened 2026-07-31
chore: improve classical vs constructive reasoning

felixpernegger:
Until now, formal conjectures deals with classical reasoning very liberally, in particular many files open the Classical namespace globally, which is unfortunate; in mathlib this is linted against, since by doing this a theorem might accidentally use classical results, even though it does not need to.

This PR removes all global `open Classical` and turns on the linter against it (the linter says very clearly what one ought to do instead, so i dont think that should pose a problem).
Furthermore, it also optimises `open Classical in`: Prefer `open scoped Classical` (or `classical`) and move it into the proof term whenever possible

(I have made a series on PRs about this topic in mathlib a few weeks ago, i.e. [this](https://github.com/leanprover-community/mathlib4/pull/41414))


mo271 (approved):
Thanks, LGTM!

PR #4672, opened 2026-07-31
chore: modulize all but `FormalConjectures/`

felixpernegger:
Modulizes the 4 files not in FormalConjectures/ which are not yet modulized


FormalConjecturesUtil/Linters/CategoryDocstringLinterTest.lean
```diff
@@ -34,7 +37,7 @@ Note: This linter can be disabled with `set_option linter.style.category_docstri
 #guard_msgs in
 @[category research open]
 theorem flagged_open_missing_docstring : True := by
-  sorry
+  trivial
```
felixpernegger, line 40: Here the `sorry` warning apparently got higher prio, so i had to change the proof

mo271 (approved):
Thanks, LGTM!

PR #4675, opened 2026-08-01
Fix Erdős Problem 361 formalization

### LINKED ISSUE #4440: Erdős Problem 361: shadowed parameter and incorrect subset-sum predicate

Erdős Problem 361 is misformalized in
[`FormalConjectures/ErdosProblems/361.lean`](https://github.com/google-deepmind/formal-conjectures/blob/b2e608fc52d765510915a244bb69b1a2741acc3c/FormalConjectures/ErdosProblems/361.lean#L35-L68).
There are two independent defects.

1. The outer parameter is a fixed real `c` with `0 < c`, but
   ```lean
   hA : ∀ c n, A n = ...
   ```
   introduces fresh binders that shadow it. Lean elaborates these as `(c n : ℕ)`, so the outer
   real `c` and `hc` are unused. The hypothesis requires one function `A` to describe the extremum
   for every natural cutoff multiplier simultaneously and is inconsistent.

2. For a candidate set `B`, the filter predicate
   ```lean
   n ≠ ∑ a ∈ B, a
   ```
   checks only that the sum of the entire candidate set differs from `n`. The problem requires that
   **no subset** of `B` sum to `n`, i.e. something equivalent to
   ```lean
   ∀ C ∈ B.powerset, n ≠ ∑ a ∈ C, a
   ```

Consequently the declaration does not faithfully represent the informal problem.

### Trivial proof/disproof

The shadowed binder makes every one of the three declarations provable by contradiction. At
`c = 0, n = 1`, the current extremum is `0`; at `c = 2, n = 1`, it is `2`. Thus `hA` forces both
`A 1 = 0` and `A 1 = 2`:

```lean
  have h₀ := hA 0 1
  have h₂ := hA 2 1
  norm_num at h₀ h₂
  change A 1 = 0 at h₀
  change A 1 = 2 at h₂
  norm_num [h₀] at h₂
```

The second defect also has a small counterexample independent of shadowing. For target `n = 4`
and universe `{1, 2, 3}`, the current expression returns `3`, because the whole set has total sum
`6 ≠ 4`. The intended expression returns `2`, because the subset `{1, 3}` sums to `4`.

### Suggested fix

Define the extremal function directly for the fixed outer `c`, and quantify over every subset of a
candidate:

```lean
def extremal (c : ℝ) (n : ℕ) : ℕ :=
  ((Finset.Icc 1 ⌊c * n⌋₊).powerset.filter fun A ↦
    ∀ B ∈ A.powerset, n ≠ ∑ x ∈ B, x).sup Finset.card
```

If the existing `A`/`hA` interface is retained, the binder must at least become `hA : ∀ n, ...`,
using the fixed outer real `c`.

The current `bigO`, `bigTheta`, and `smallO` declarations also do not formalize the separate question
"Does this depend on `n` in an irregular way?" A correction should either state a specific meaningful
asymptotic target or introduce an explicit mathematical definition of irregularity.

### Additional context

- Current problem page: https://www.erdosproblems.com/361
- Original tracking issue: https://github.com/google-deepmind/formal-conjectures/issues/512
- Introducing PR: https://github.com/google-deepmind/formal-conjectures/pull/590
- Primary source: Erdős–Graham, *Old and New Problems and Results in Combinatorial Number Theory*,
  printed p. 59:
  https://web.archive.org/web/20250825124722id_/https://mathweb.ucsd.edu/~ronspubs/80_11_number_theory.pdf

The primary source uses integers strictly less than `n/k`, with `k` fixed and `n` large. The current
Erdős Problems page modernizes this to fixed `c > 0` and `{1, ..., ⌊cn⌋}`. That source-level
normalization is separate from the two unequivocal Lean defects above.

### Choose either option

Robertboy18:
Fixes #4440 :)

The old statement shadowed the fixed `c` and only checked whether the whole candidate set summed to `n`. This fixes both by defining the extremal function directly and checking every subset!

I also added the small `n = 4` test from the issue, where the maximum for `{1, 2, 3}` is `2`!


FormalConjectures/ErdosProblems/361.lean
```diff
@@ -22,50 +22,47 @@ import FormalConjecturesUtil
...
-    (hA : ∀ c n, A n = ((Finset.Icc 1 ⌊c * n⌋₊).powerset.filter
-      (fun B ↦ n ≠ ∑ a ∈ B, a)).sup Finset.card) :
-    (fun n ↦ (A n : ℝ)) =Θ[atTop] (answer(sorry) : ℕ → ℝ) := by
-  sorry
+def maxSubsetSumAvoidingCard (N n : ℕ) : ℕ :=
+  ((Finset.Icc 1 N).powerset.filter (AvoidsSubsetSum n)).sup Finset.card
+
+/--
+For fixed `c` and `n`, this is the size of the largest set
+`A ⊆ {1, ..., ⌊c n⌋}` such that no subset of `A` sums to `n`.
```
mo271: please use latex markdown in docstring

FormalConjectures/ErdosProblems/361.lean
```diff
@@ -22,50 +22,47 @@ import FormalConjecturesUtil
...
 -/
 @[category research open, AMS 11]
-theorem erdos_361.smallO
-    (c : ℝ) (hc : 0 < c)
-    (A : ℕ → ℕ)
-    (hA : ∀ c n, A n = ((Finset.Icc 1 ⌊c * n⌋₊).powerset.filter
-      (fun B ↦ n ≠ ∑ a ∈ B, a)).sup Finset.card) :
-    (fun n ↦ (A n : ℝ)) =o[atTop] (answer(sorry) : ℕ → ℝ) := by
+theorem erdos_361 (c : ℝ) (hc : 0 < c) :
+    subsetSumAvoidanceNumber c = answer(sorry) := by
```
mo271, line 68: this is now asking for an exact formula, right? before it was asymptotic only? What is the intention of the question on erdosproblems?
In case of doubt we could add both

mo271 (commented):
Thanks for the contribution!
LGTM generally, left two comments

FormalConjectures/ErdosProblems/361.lean
Robertboy18: fixed, I switched the new mathematical docstrings to LaTeX!

FormalConjectures/ErdosProblems/361.lean
Robertboy18: yeah, the previous revision was asking for the exact function. I kept that and added a separate Θ statement for the asymptotic growth, so both readings are included now!

mo271 (approved):
Thanks, LGTM!

PR #4686, opened 2026-08-02
feat(WrittenOnTheWallII): prove Conjecture 322

MrBrain295:
Proven by @Aristotle-Harmonic.

### DIFF
--- a/FormalConjectures/WrittenOnTheWallII/GraphConjecture322.lean
+++ b/FormalConjectures/WrittenOnTheWallII/GraphConjecture322.lean
@@ -41,7 +41,8 @@ vertices `v` of `l(v)` — the independence number of the neighborhood `N(v)` of
 Here `l(v) = α(G[N(v)])` is the independence number of the subgraph induced by the
 open neighborhood of `v`.
 -/
-@[category research open, AMS 5]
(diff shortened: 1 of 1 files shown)


mo271:
Thanks for your contribution! 

Can you please add in the dostring an informal sketch of the proof?!

MrBrain295:
> Thanks for your contribution!
> 
> Can you please add in the dostring an informal sketch of the proof?!

@mo271 done!

Danie-I (commented):
Unfortunately this conjecture is misformalized. There should be indepNeighborsCard comp G . (complement of graph G)

mo271:
> Unfortunately this conjecture is misformalized. There should be indepNeighborsCard comp G . (complement of graph G)

Perhaps it is worth looking systematically for missing `comp`?

Danie-I:
> > Unfortunately this conjecture is misformalized. There should be indepNeighborsCard comp G . (complement of graph G)
> 
> Perhaps it is worth looking systematically for missing `comp`?

Yes, it would be worth to do it. As far as I remember we had the same issue befeore.

mo271:
closing in favor of #4430

PR #4687, opened 2026-08-02
feat(WrittenOnTheWallII): prove Conjecture 65

MrBrain295:
https://github.com/MrBrain295/formal-conjectures/blob/65/FormalConjectures/WrittenOnTheWallII/GraphConjecture65.lean


SamuelSchlesinger:
Please read the details of the proof here, it shows the deficiencies of the definition of distMin. We talked about it in #4514. Happy to be shown that I am wrong and in fact this theorem is trivial.

Danie-I (commented):
This problem does not look correctly formalized. dist_min is minimum distance among the vertices from the set. This is not at most 1 as is stated in the proof.

PR #4688, opened 2026-08-02
chore: modulize FormalConjectures/

felixpernegger:
Modulizes all remaining files.

Many `native_decide` stuffs breaks; I (better: codex) tried to restore the proofs f there is a simple replacement, which worked for most; but overall still 27 sorry's are added.

Zulip discussion:
[#Formal conjectures > Modulization](https://leanprover.zulipchat.com/#narrow/channel/524981-Formal-conjectures/topic/Modulization/with/623771711)


FormalConjectures/WrittenOnTheWallII/Test.lean
mo271, line 1: this has plenty of simple proofs deleted, if it is easy it would be nice to restore them

mo271 (commented):
Thanks!
Left a comment. I would also be interested in @eric-wieser's comments.

FormalConjectures/WrittenOnTheWallII/Test.lean
felixpernegger: The issue is the heavy use of native_decide in the repo, which breaks the module system. I'll try to replace as far as possible

felixpernegger: I tried a bit, but this is actually not so easy to fix. Especially since VSCode preview also has a bug related to this...

If it's okay, i leave it as is for now and fix those in a later PR?

mo271: sure, let's do it in a follow-up!

felixpernegger:
FYI I have posted about the `native_decide` + module system issue here
https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/native_decide.20in.20module.20system.20issue/with/614561087

mo271:
one way forward here would be to merge it as proposed, removing some of our `native_decide` and put them back when the issue has been worked out.

felixpernegger:
Yeah I agree; I also think it would be nice to somewhat reduce the `native_decide` overall. Mathematically it doesnt really matter, as it is only used for tests etc, but still...

felixpernegger:
I keep getting merge conflicts; I would be very happy to get some feedback (here or on the Zulip thread)

mo271 (approved):
Thanks, LGTM!

I suppose another merge/rebase will be needed here, I'm happy to not put anything on the merge queue to get this merged here now

PR #4693, opened 2026-08-02
feat(Erdős 176): formalize discrepancy progression numbers and exact values

### LINKED ISSUE #405: Erdős Problem 176
### What is the conjecture
    
https://www.erdosproblems.com/176

  Let $N(k,\ell)$ be the minimal $N$ such that for any $f:\\{1,\ldots,N\\}\to\\{-1,1\\}$ there must exist a $k$-term arithmetic progression $P$ such that

$$\left\lvert \sum_{n\in P}f(n)\right\rvert\geq \ell.$$

Find good upper bounds for $N(k,\ell)$. Is it true that for any $c>0$ there exists some $C>1$ such that
$$N(k,ck)\leq C^k?$$
What about
$$N(k,2)\leq C^k$$
or
$$N(k,\sqrt{k})\leq C^k?$$

Status: open

### Choose either option

arex1337:
## What changed

- added `Erdos176.ForcesDiscrepancy`;
- added the `ℕ∞`-valued `Erdos176.discrepancyAPNumber`;
- formalized the three questions as `erdos_176.parts.i`–`.iii`;
- added the exact solved variants `N(13,2) = 158` and `N(15,2) = 225`.

## Statement fidelity

The coloring domain is exactly `{1, …, N}`. Progressions have exactly `k`
distinct terms through `Set.IsAPOfLength`, sums are computed in `ℤ`, and only
the threshold comparison is cast to `ℝ`. The infimum is taken in `ℕ∞`, so
absence of a finite forcing length is represented by `⊤`.

The source’s part (i) says “for any `c > 0`.” Since a `k`-term sign sum has
absolute value at most `k`, the source is impossible for `c > 1`; the
formalization documents the meaningful range `0 < c ≤ 1` and retains the
`c = 1` boundary.

## Exact certificates

The two exact values use a witness at `N - 1` and DRAT-checked UNSAT
certificate data at `N`. The live archival record is
https://doi.org/10.5281/zenodo.21761884 (v1.2).

These declarations are marked solved without `formal_proof`: the evidence is
an exact computational certificate, not a proof-assistant formalization.

Closes #405.

## Checks

- independent witness enumeration for all 949 (`k=13`) and 1,680 (`k=15`)
  arithmetic progressions
- exact regenerated CNF hashes matched the archive
- all 4,096 `k=13` cubes have unique `VERIFIED` results
- the `k=15` checker log ends `s VERIFIED` with return code zero
- local archive size and MD5 match Zenodo
- `lake --wfail build` on this exact PR head

Disclosure: the formalization and analysis were AI-generated under human
direction; this exposition was human-prepared, and the reported exact values
were checked by exact certificate tooling.

### DIFF
--- a/FormalConjectures/ErdosProblems/176.lean
+++ b/FormalConjectures/ErdosProblems/176.lean
@@ -43,7 +43,7 @@ here.
   on Erdős's Discrepancy Threshold N(k,2).
   [Zenodo](https://doi.org/10.5281/zenodo.20763838).
 - [Ly26] Lystad, T. A. Erdős Problem 176 — exact small values of the discrepancy
-  threshold N(k,2) with machine-checkable DRAT certificates (v1.2: N(15,2) = 225).
+  threshold N(k,2) with machine-checkable DRAT certificates (v1.3: split packaging).
   [Zenodo](https://doi.org/10.5281/zenodo.21739846) (concept DOI; resolves to the
   latest version).
 -/
--- a/FormalConjectures/ErdosProblems/176.lean
+++ b/FormalConjectures/ErdosProblems/176.lean
@@ -19,11 +19,33 @@ import FormalConjecturesUtil
 /-!
 # Erdős Problem 176
 
+Write $N(k,\ell)$ for the least $N$ such that every $\{-1,1\}$-coloring of
+$\{1,\ldots,N\}$ admits a $k$-term arithmetic progression whose sum has
+absolute value at least $\ell$. Erdős asks whether $N(k,ck)$, $N(k,2)$ and
+$N(k,\sqrt{k})$ are at most exponential in $k$.
+
+Two exact values of $N(k,2)$ are recorded at the end of the file. The choice of
+$k$ there is not arbitrary: a $k$-term $\{-1,1\}$ sum has the parity of $k$, so
+for even $k$ one has $|s| \geq 1 \iff |s| \geq 2$ and hence
+$N(k,2) = N(k,1) = 2^t(k-1)+1$ for $k = 2^t m$ with $m$ odd [Sp73]. The open
+content of $N(k,2)$ therefore sits at odd $k$. There $N(3,2) = 9$ is the
+classical van der Waerden number $W(2,3)$, and $N(k,2)$ for $k = 5,7,9,11$ was
+computed in [Go26]. $k = 13$ and $k = 15$ are the first two beyond that frontier
+for which both sides carry exact machine-checkable certificates [Ly26]; for odd
+$k \geq 17$ only lower bounds are known, so no further exact value is stated
+here.
+
 *References:*
 - [erdosproblems.com/176](https://www.erdosproblems.com/176)
-- [Sp73] Spencer, J. Problems 185. Bull. Canad. Math. Soc. (1973), 185.
-- [Ly26] Lystad, T. A. Machine-checkable UNSAT certificates for exact values of
-  Erdős discrepancy thresholds. [Zenodo v1.2](https://doi.org/10.5281/zenodo.21761884).
+- [Sp73] Spencer, J. Solution to Problem P.185.
+  Bull. Canad. Math. Soc. 16 (1973), 464.
+- [Go26] Goss, M. J., Jr. The Parity Collapse and Entropy Drain: First Bounds
+  on Erdős's Discrepancy Threshold N(k,2).
+  [Zenodo](https://doi.org/10.5281/zenodo.20763838).
+- [Ly26] Lystad, T. A. Erdős Problem 176 — exact small values of the discrepancy
+  threshold N(k,2) with machine-checkable DRAT certificates (v1.2: N(15,2) = 225).
+  [Zenodo](https://doi.org/10.5281/zenodo.21739846) (concept DOI; resolves to the
+  latest version).
 -/
 
 open Set ENat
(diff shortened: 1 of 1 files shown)


mo271 (commented):
LGTM in general. 

Perhaps @KitaKen1 could also review this, since they have commented on the comment section of that problem?

KitaKen1:
Thanks for the ping, @mo271. Scope: definitions and parts (i)–(iii) only. I haven't checked [Ly26], so no opinion on the two exact values.

Those look faithful to the source — `IsAPOfLength` carries `ENat.card s = k`, so degenerate `d = 0` progressions are excluded, which was the thing I wanted to check. Notes, none blocking:

**1. `2 ≤ k` is undocumented.** At `k = 1`, `discrepancyAPNumber 1 2 = ⊤`, so (ii) would be false without it. The `0 < c ≤ 1` narrowing is explained; this one isn't.

**2. The `= 158` docstring covers `N = 157` but not `N ≤ 156`.**

**3. Why `k = 13, 15` specifically?** The docstrings don't say.

**4. [Ly26].** I couldn't download it — the archive was too large.

arex1337:
Thanks for the careful read, @KitaKen1, and for saying what you did and did not
check. Three are fixed in this push; the fourth is fixed in a new Zenodo version
that is built and waiting on upload. Docstrings and references only, no
statement, hypothesis or category tag touched.

**1. `2 ≤ k`.** You are right that it was load-bearing and unexplained. Now
stated in the `parts.ii` docstring, with your reason (`discrepancyAPNumber 1 2 =
⊤`, so (ii) would be false without it); the bound underneath it, that a `k`-term
sign sum has absolute value at most `k` so the value is `⊤` whenever `ℓ > k`, is
in the `discrepancyAPNumber` docstring just above. It binds only in (ii). At `k ≤ 1` the thresholds in (i) (`c ≤ 1`) and (iii)
(`√k`) are met by a single term, so there the hypothesis just keeps the three
uniform. That is said too.

**2. `N ≤ 156`.** Fair, the docstring stopped one step short.
`ForcesDiscrepancy` is monotone in `N` (restricting a coloring of `{1,…,N}` to
`{1,…,N'}` leaves every progression of the smaller interval and its sum
unchanged), so it cuts out an upward-closed set and `discrepancyAPNumber` is a
threshold rather than merely a smallest witness. The avoiding coloring at 157
therefore rules out every `N' ≤ 157` at once. Now in the `ForcesDiscrepancy` and
`discrepancyAPNumber` docstrings and in both variants.

**3. Why `k = 13, 15`.** Now in the module docstring. Even `k` carries no open
content: a `k`-term sum has the parity of `k`, so `N(k,2) = N(k,1) = 2^t(k−1)+1`
for `k = 2^t m` with `m` odd. Among odd `k`, `N(3,2) = 9` is the classical van
der Waerden number `W(2,3)`, and `k = 5,7,9,11` were computed by M. J. Goss, Jr.
([10.5281/zenodo.20763838](https://doi.org/10.5281/zenodo.20763838)), now cited
as `[Go26]`; ours are independent confirmations. `k = 13` and `k = 15` are the
first two past that frontier with both sides machine-certified. For odd `k ≥ 17`
we have lower bounds only, so there is no further exact value to state.

**4. The archive.** Entirely our fault, and thank you for saying so rather than
giving up quietly. It was one 5.17 GiB tarball. Two things:

The lower side never needed a download. Both witnesses are short strings, so here
they are. `+` is `+1`, `-` is `-1`, position `i` is `n = i+1`:

`k = 13`, `N = 157`:

```
+++++++------+++++++------+++++++------+++++++------+++++++------+++++++------+++++++-------++++++-------++++++-------++++++-------++++++-------++++++----+--
```

`k = 15`, `N = 224`:

```
++++++++--------++++++++--------++++++++--------++++++++--------++++++++--------++++++++--------++++++++--------++++++++--------++++++++--------++++++++--------++++++++--------++++++++--------++++++++--------++++++++--------
```

```python
def max_ap_sum(s, k):
    f = [1 if c == '+' else -1 for c in s]
    N = len(f)
    return max(abs(sum(f[a + j*d] for j in range(k)))
               for a in range(N)
               for d in range(1, N)
               if a + (k-1)*d < N)

# both return 1, so every k-AP sum is ±1 and neither coloring has |sum| >= 2
```

That settles `N(13,2) > 157` and `N(15,2) > 224` in about a second, with nothing
downloaded. The upper side is the part that genuinely needs the multi-GB proof,
because the DRAT file is the certificate.

And the packaging is fixed. v1.3 splits the record into two files:

NK2_CERTIFICATES_v1.3_lite.tar.gz, 4.3 MiB — everything except the proof objects: both witnesses, every DIMACS instance, the deterministic generators, the exact evaluator, the audit harness, every solver and drat-trim transcript, the SHA-256 manifests.
NK2_CERTIFICATES_v1.3_drat.tar.gz — the DRAT proofs.
4,108 of the 4,804 files were proof objects and over 99.7% of the bytes, so almost everything checkable had been sitting behind almost all of the download.

[10.5281/zenodo.21840279](https://doi.org/10.5281/zenodo.21840279), or [10.5281/zenodo.21739846](https://doi.org/10.5281/zenodo.21739846) for the concept DOI, which is what [Ly26] now cites so it always resolves to the latest version. The lite bundle is self-contained for the lower side: unpack it and its own exact integer evaluator, code/ap176.py, decides both witnesses (avoids(f, k, 2)) with nothing from the DRAT bundle -- the snippet above is the same check. The upper side still needs the proofs, since the DRAT file is the certificate.

KitaKen1:
@mo271 My comments here are only about whether these belong as variants of Erdős 176. I’m not convinced that (k=13,15) are natural variants simply because they are the next odd cases; I’d leave that repository-organization question to the maintainers. I haven’t checked [Ly26], so no opinion on the two exact values.

Issue #4700, opened 2026-08-03
FormalConjecturesTest never gets built, 8 of 10 modules fail


williamjblair:
`lake test` doesn't run:

```
lake test
✖ [0/0] Running job computation
error: no such file or directory (error code: 2)
  file: .../FormalConjecturesTest.lean
```

`lake build FormalConjecturesTest` fails the same way: the `lean_lib` entry in `lakefile.toml` has
no `globs`, so lake looks for a root module `FormalConjecturesTest.lean`, which doesn't exist. No
workflow builds the lib either, so it has never been compiled, and 8 of its 10 modules currently
fail. The main build is unaffected.

Most of those failures come from one omission: the test lib doesn't set the `weak.linter.style.*`
options that `FormalConjectures` sets, so the linters are off and every linter test expects a
warning that is never emitted.

Adding `globs = ["FormalConjecturesTest.+"]` and those four options makes the lib build. Six modules
still fail:

- `Util/Answer`: `Expr.const imported as meta; consider adding import Lean.Expr`, plus two IR check failures
- `Util/Linters/ModuleDocstringLinter`: `Unknown option linter.style.moduleDocstring`, though the file imports the module that registers it
- `Util/Linters/CategoryLinter`, `ExistsImplicationLinter`: `#guard_msgs` mismatches
- `ForMathlib/Computability/TuringMachine`: `Unknown constant State1` / `State2` from `turing_machine%`
- `Util/Attributes/Basic`: two separate problems, below

Most look like meta/public import issues from #4665.

In `Util/Attributes/Basic`, `@[category graduate]` (line 51) is not a valid category (`Category` is
`textbook | research | test | API`). Separately, `#AMS` produces no output when compiled: it reads
constructor docstrings via `Lean.findDocString?`, and imported docstrings are only exported at
`.server` level, so the lookup returns `none` outside the language server (62 constants, 0
docstrings). Docstrings are readable inside the module that declares them, so the list could be
built in `AMS.lean`. That one is a lib bug rather than a stale test, so possibly a separate PR.

I can take this. Separate CI job, or add to the existing build (#4393 and #4368 already touch it)?
And `#AMS` here or separately?

mo271:
cc-ing @felixpernegger

felixpernegger:
I'll look into this in more detail later; but first of all it looks like FormalConjecturesTest isn't checked in the CI? Is there any reason for this

mo271:
[...]

PR #4703, opened 2026-08-03
ci: pin actions to commit SHAs and fix a template injection

williamjblair:
The org-level Actions scan wants actions pinned by hash and rejects expanding attacker-controllable values into shell, so any PR touching a workflow currently fails. #4368 is red for this, on findings unrelated to its own diff.

- Pinned all 11 unpinned action references across five workflows, `@<sha> # <tag>` as in `welcome_pull_request.yml`. Each SHA resolves to the tag it's commented with.
- Routed `inputs.website_only` and `github.ref_name` through `env` in "Detect build mode" rather than into the shell. A branch name can break out of the `[[ ]]`.
- Moved `pages: write` and `id-token: write` off the workflow and onto the `deploy` job, which is the only one that uses them.
- Added `permissions:` blocks to `check_copyright_header.yml` and `update_erdosproblems.yml`, which had none and inherited the repo default. The copyright script only reads the working tree; the erdosproblems dispatch authenticates with `REPO_DISPATCH_PAT` rather than `GITHUB_TOKEN`, so it gets `permissions: {}`.

The last two came out of fixing the first two, since each round of findings hid the next. All eight workflows now set permissions explicitly and zizmor passes.

The scan only looks at files a PR changes, so none of this showed up until #4368 happened to touch a workflow. #4700 is the same kind of thing.

### DIFF
--- a/.github/workflows/build-and-docs.yml
+++ b/.github/workflows/build-and-docs.yml
@@ -40,7 +40,7 @@ jobs:
     steps:
 
       - name: Checkout project
-        uses: actions/checkout@v6
+        uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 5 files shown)


mo271 (approved):
Thanks, LGTM

PR #4704, opened 2026-08-03
Mark Weak Tiling Problem 4.3 as solved

KitaKen1:
This PR changes `WeakTiling.problem_4_3` from `answer(sorry)` to `answer(False)` and links a kernel-checked Lean proof.

The Lean proof was prepared by [KitaKen1 (Kenta Kitamura)](https://github.com/KitaKen1).

The external proof establishes:

```lean
WeakTiling.Counterexample.not_problem_4_3 :
  ¬ ∀ (Ω : Set ℝ) (_ : IsFiniteUnionOfIntervals Ω)
      (ν' : Measure ℝ) (_ : IsWeakTilingMeasure Ω ν'),
    ∃ (T : ℕ → Set ℝ) (c : ℕ → ℝ≥0),
      (∀ i, IsProperTiling Ω (T i)) ∧ ∑' i : ℕ, c i = 1 ∧
      ν' = Measure.sum
        (fun i ↦ (c i : ℝ≥0∞) •
          Measure.sum (fun t : T i ↦ Measure.dirac (t : ℝ)))
```

Proof: https://github.com/KitaKen1/weak-tiling-counterexample/blob/5bf93234cc51f02fd7681407d77dcebde592f3ac/formal-conjectures-v4.27.0/WeakTilingCounterexample.lean

Repository: https://github.com/KitaKen1/weak-tiling-counterexample

Lean4Web: https://live.lean-lang.org/#url=https%3A%2F%2Fraw.githubusercontent.com%2FKitaKen1%2Fweak-tiling-counterexample%2Fmain%2Flean4web-v4.33.0-rc1%2FWeakTilingCounterexampleLean4Web.lean

`#print axioms` reports only `propext`, `Classical.choice`, and `Quot.sound`; there is no `sorryAx`.

AI Usage Disclosure: This formalization is assisted by ChatGPT 5.6 sol and Codex GPT 5.6 sol with xhigh reasoning.

### DIFF
--- a/FormalConjectures/Paper/WeakTiling.lean
+++ b/FormalConjectures/Paper/WeakTiling.lean
@@ -94,9 +94,11 @@ theorem problem_4_2 :
 /-- **Problem 4.3.** Let $\Omega \subset \mathbb{R}$ be a finite union of intervals and $\nu$
     a weak tiling measure for $\Omega$. Must $\nu$ be expressible as a convex combination of
     proper tiling measures? -/
-@[category research open, AMS 42 46]
+@[category research solved, AMS 42 46,
+  formal_proof using lean4 at
+    "https://github.com/KitaKen1/weak-tiling-counterexample/blob/5bf93234cc51f02fd7681407d77dcebde592f3ac/formal-conjectures-v4.27.0/WeakTilingCounterexample.lean"]
 theorem problem_4_3 :
-    answer(sorry) ↔ ∀ (Ω : Set ℝ) (_ : IsFiniteUnionOfIntervals Ω)
+    answer(False) ↔ ∀ (Ω : Set ℝ) (_ : IsFiniteUnionOfIntervals Ω)
       (ν : Measure ℝ) (_ : IsWeakTilingMeasure Ω ν),
       ∃ (T : ℕ → Set ℝ) (c : ℕ → ℝ≥0), (∀ i, IsProperTiling Ω (T i)) ∧ ∑' i : ℕ, c i = 1 ∧
       ν = Measure.sum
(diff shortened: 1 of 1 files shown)


mo271:
Hi, since @aditya-ramabadran suggested to add those problems and @Sfgangloff
formalised them, perhaps they have some opinion on if this is indeed solving problem 4.3 from https://arxiv.org/abs/2506.23631, i.e. if the problem was formalised correctly.

FormalConjectures/Paper/WeakTiling.lean
```diff
@@ -94,9 +94,11 @@ theorem problem_4_2 :
 /-- **Problem 4.3.** Let $\Omega \subset \mathbb{R}$ be a finite union of intervals and $\nu$
     a weak tiling measure for $\Omega$. Must $\nu$ be expressible as a convex combination of
     proper tiling measures? -/
```
mo271: can you add an informal proof sketch for the counterexample? (And also mention that you found it)

mo271 (commented):
thanks!

FormalConjectures/Paper/WeakTiling.lean
KitaKen1: Thanks, I added an informal proof sketch and mentioned that I found the counterexample through an AI-assisted investigation.

williamjblair:
Built the linked file on FC's toolchain: compiles, and `#print axioms` on `not_problem_4_3` and `problem_4_3_solved` gives the standard three. The `lrat_std_proof` terms go through mathlib's `FromLRAT`, so kernel-checked rather than `native_decide`.

It restates FC's four `WeakTiling` definitions locally instead of importing them. I diffed all four against this branch: character-identical, and `not_problem_4_3` matches `problem_4_3` apart from `ν` vs `ν'`.

Whether the paper's 4.3 is the same question as the formalisation is the bit @mo271 asked the original formalisers about, not something I have a view on.

aditya-ramabadran:
I checked this against Problem 4.3 in the paper and I think it does solve it! Congrats 

The definitions seem right to me and the counterexample gives a weak tiling measure with $\nu({7})=1/2$ and $\nu({15})=0$, while every proper tiling has equal mass at 7 and 15. So no convex combination of proper tilings can equal $\nu$ :)

mo271 (approved):
Thanks, LGTM!

Issue #4706, opened 2026-08-04
Erdős Problem 146


ctjlewis:
*This issue was written by Claude Opus 5.*

### What is the Erdős problem?

https://www.erdosproblems.com/146

If $H$ is bipartite and is $r$-degenerate, that is, every induced subgraph of $H$ has minimum degree $\leq r$, then
$$\mathrm{ex}(n;H) \ll n^{2-1/r}.$$

Conjectured by Erdős and Simonovits [ErSi84]. Alon, Krivelevich, and Sudakov [AKS03] have proved $\mathrm{ex}(n;H) \ll n^{2-1/4r}$, and prove the full conjectured bound if $H$ is bipartite and the maximum degree in one side of the bipartition is $r$.

Status: solved (negatively). erdosproblems.com still records the problem as open as of its edit of 18 January 2026. It was refuted by OpenAI, *Ten advances in mathematics and theoretical computer science*, Ch. 10 (2026), which exhibits a bipartite $2$-degenerate $H$ with $\mathrm{ex}(n;H)$ eventually exceeding any multiple of $n^{3/2}$; since the conjecture is universally quantified over $r$, that case refutes it. A Lean 4 proof is available at https://github.com/openai/ten-proofs.

A strengthening to every $r \geq 2$, with degeneracy exactly $r$ and exponent $2-1/r+1/(28r^2)$, is available at https://github.com/EvolvingPrograms/erdos-simonovits-degeneracy. It generalises the OpenAI construction and builds on its Lean development. It is machine-checked, and the Lean formalisation was checked by Sai Gajjala (New York University). It has not been peer reviewed. Whether it should be recorded here, and under which category, is a question for review of the pull request.

The repository currently contains no formalised statement of this problem.

### [AMS categories](https://github.com/google-deepmind/formal-conjectures/labels?q=ams-)

* ams-05

### Choose either option
- [x] I plan on adding this Erdős problem to the repository
- [ ] This issue is up for grabs: I would like to see this Erdős problem added by somebody else

### References

- [ErSi84] Erdős, P. and Simonovits, M., *Cube-supersaturated graphs and related problems*. Progress in graph theory (Waterloo, Ont., 1982) (1984), 203-218.
- [AKS03] Alon, N., Krivelevich, M. and Sudakov, B., *Turán numbers of bipartite graphs and related Ramsey-type questions*. Combin. Probab. Comput. **12** (2003), 477-494.

mo271:
[...]

PR #4707, opened 2026-08-04
feat(ErdosProblems/146): counterexamples at every degeneracy r ≥ 2

ctjlewis:
Linked issue: #4706. Builds on #4715, now merged.

**What this PR adds**

Three additions to `FormalConjectures/ErdosProblems/146.lean`. Nothing already in the file is restated — `erdos_146` and `erdos_146.variants.two_degenerate_counterexample` are #4715's, untouched, and the new statements use its file-local `IsDegenerate`.

- `erdos_146.variants.counterexample` — for every $r \geq 2$ there is a connected bipartite graph $H_r$ of degeneracy **exactly** $r$, and a $c > 0$, with $\mathrm{ex}(n;H_r) \geq c\,n^{2-\frac1r+\frac1{28r^2}}$ for all large $n$.
- `erdos_146.variants.counterexample_three` — at $r = 3$ the exponent sharpens to $\frac53 + \frac1{160}$.
- `not_bigO_of_lower_bound` — proved here (no `sorry`): one bipartite $r$-degenerate graph with a lower bound $c\,n^{2-1/r+\varepsilon}$ contradicts the $O(n^{2-1/r})$ bound, since $n^\varepsilon$ outgrows any constant. This is what makes the two variants above refutations rather than isolated bounds.

**Relation to #4715**

#4715 ports the $r = 2$ refutation of [OpenAI26, Ch. 10], which settles the conjecture, since it is universally quantified over $r$. The construction there fixes a Gibbs weight numerically to reach $r = 2$; leaving that weight free gives a counterexample at every $r$, which is what is added here. At $r = 2$, `counterexample` recovers `two_degenerate_counterexample` with the explicit gain $\varepsilon = \frac1{112}$, and additionally pins the degeneracy to exactly $2$. The Lean development behind these statements imports OpenAI's, vendored verbatim under Apache-2.0.

**Status**

The new result is not peer reviewed. It was announced publicly in August 2026; the Lean formalisation was checked by Sai Gajjala (New York University). erdosproblems.com does not yet record the problem as solved (last edited 18 January 2026). The two new variants are tagged `research solved` on the strength of the linked machine-checked proofs; if you would prefer them omitted, or recorded differently, say so and I will adjust.

**Verification**

`lake --wfail build FormalConjectures.ErdosProblems.«146»` passes with no warnings. `#print axioms Erdos146.not_bigO_of_lower_bound` gives `propext, Classical.choice, Quot.sound`. The file's four `sorry`s are #4715's two statements plus the two new variants, each carrying a `formal_proof` link.

Both links are pinned to `6b03c46` of `EvolvingPrograms/erdos-simonovits-degeneracy` (green CI), at `proofs/Theorem12.lean` (`rDegenerateExtremalCounterexample_exact`) and `proofs/Theorem12r3.lean` (`threeDegenerateExtremalCounterexample_sharp`). @williamjblair — this re-pins from `87510f34`, which you checked; the repo was restructured since, and this SHA matches its current layout.

*Disclosure: the formalisation in this PR and the linked development were produced with AI assistance (Claude Fable 5 and Claude Opus 5), per the Mathlib AI usage conventions.*


ctjlewis:
+WIP

ctjlewis:
-WIP

mo271:
Can you compare how your pr compares to #4715, which ports from the formalisations from the OpenAI repo?

ctjlewis:
@mo271 OpenAI's proof is for $r=2$. This proof is for all $r\ge2$, it expands on OpenAI's result. Whether we include the 146 problem statement In Lean from this PR or the other one should not make a difference.

ctjlewis:
@mo271 We refactored on top of the 146 statement from #4715. That would have to merge first.

ctjlewis:
@mo271 This PR will suggest putting the degeneracy stuff at `Combinatorics/SimpleGraph.IsDegenerate`, so that it's not just defined in-place in 146, that way future problems can import it from a common directory. But if we want to leave it in-place let me know, not really relevant until after the other one merges either way.

williamjblair:
Checked both new links. `EvolvingPrograms/erdos-simonovits-degeneracy` has green CI on the pinned SHA, and across its 43 files there is no `axiom` and no `sorry` in the import chain: the real ones sit in `challenges/`, which state the problems, and `lib/` imports only other `lib/` modules and Mathlib. `rDegenerateExtremalCounterexample_exact` and `threeDegenerateExtremalCounterexample_sharp` are unconditional and match `counterexample` and `counterexample_three` here.

`lib/LedgerTuned.lean` takes §10.4(c) as an explicit hypothesis rather than proving it, which the file says. The top-level theorems carry no such hypothesis, so it is discharged or unused. That is the place to start if anyone audits deeper.

`Build project` has never run here, so nothing has type-checked in CI yet, and it still sits on top of unmerged #4715.

PR #4715, opened 2026-08-04
ErdosProblems: add 146, 180, 183 with openai/ten-proofs formal_proof links

### LINKED ISSUE #382: Erdős Problem 146
### What is the conjecture
    
https://www.erdosproblems.com/146

  If $H$ is bipartite and is $r$-degenerate, that is, every induced subgraph of $H$ has minimum degree $\leq r$, then
$$\mathrm{ex}(n;H) \ll n^{2-1/r}.$$

Status: open

### Choose either option

### LINKED ISSUE #407: Erdős Problem 180
### What is the conjecture
    
https://www.erdosproblems.com/180

  If $\mathcal{F}$ is a finite set of finite graphs then $\mathrm{ex}(n;\mathcal{F})$ is the maximum number of edges a graph on $n$ vertices can have without containing any subgraphs from $\mathcal{F}$. Note that it is trivial that $\mathrm{ex}(n;\mathcal{F})\leq \mathrm{ex}(n;G)$ for every $G\in\mathcal{F}$. Is it true that, for every $\mathcal{F}$, there exists $G\in\mathcal{F}$ such that
$$\mathrm{ex}(n;G)\ll_{\mathcal{F}}\mathrm{ex}(n;\mathcal{F})?$$

Status: open

### Choose either option

### LINKED ISSUE #409: Erdős Problem 183
### What is the conjecture
    
https://www.erdosproblems.com/183

  Let $R(3;k)$ be the minimal $n$ such that if the edges of $K_n$ are coloured with $k$ colours then there must exist a monochromatic triangle. Determine
$$\lim_{k\to \infty}R(3;k)^{1/k}.$$

Status: open

### Choose either option

williamjblair:
fixes #382, fixes #407 fixes #409 

Adds statements and `formal_proof` links for three Erdős problems resolved in [openai/ten-proofs](https://github.com/openai/ten-proofs), pinned to `94bc0fe`. Not my work.

| Problem | What the proof gives |
|--:|---|
| 146 | a connected bipartite 2-degenerate `H` with `ex(n;H) ≥ c·n^(3/2+ε)`, so the conjectured `ex(n;H) ≪ n^(2-1/r)` fails at `r = 2` |
| 180 | a family with no single member controlling its extremal number, so the Erdős–Simonovits compactness conjecture is false |
| 183 | `R(3;k)^(1/k) → ∞`, plus an explicit superexponential lower bound |

FC had no file for any of the three, so the statements are new rather than links onto existing ones.

## Why these three

`problems.yaml` upstream already records `formal_status: Lean` for 146, 180 and 183, each pointing at this exact commit, with `formalized: no`. So these are gaps erdosproblems has flagged and already accepted the formalization for.

The repo is Apache-2.0, its `formalization.yaml` reports `sorry_count: 0` with axioms exactly `[propext, Classical.choice, Quot.sound]`, and it ships with a paper.

ten-proofs contains twelve results in total; the other nine are not Erdős problems and would need FC statements in areas the repo does not currently cover (sphere packing, metric codes, sofic groups, Connes rigidity, Ehrhart, quantum parallel repetition, closest-vector hardness, permanent lower bounds). Those want domain review rather than transcription, so they are left out here.

## Statements

Each FC statement mirrors the linked theorem's own definitions, so the correspondence is an
unfolding rather than an argument.

- **183** — `TriangleFree` and `ForcesMonochromaticTriangle` are stated over Mathlib's
  `TopEdgeLabeling`, as the proof does, and `multicolourTriangleRamsey` is the same `sInf`. The
  headline is the divergence; `erdos_183.variants.explicit_lower_bound` carries the explicit bound.
- **146** — `IsDegenerate` and `neighborsWithin` match the proof's, and the headline uses the same
  `IsBigO` phrasing over `SimpleGraph (Fin q)`. Headline is `answer(False)`; the 2-degenerate
  counterexample is the linked variant.
- **180** — `FiniteGraph`, `FamilyFree`, `familyExtremal`, `IsCyclicFamily` and `IsCompactFamily`
  all mirror the proof's, and the headline is its `CompactnessConjectureStatement`. Headline is
  `answer(False)`; the counterexample is the linked variant.

An earlier version of this PR wrote the statements independently and got 180 wrong: the family was
`Finset (SimpleGraph (Fin q))`, which forces every member onto a common vertex count, where the
problem is about a finite set of finite graphs of any orders. It also dropped the `IsCyclicFamily`
hypothesis. Both are fixed.

## What I have not done

I have not rebuilt the proofs. They pin Lean 4.32 against their own Mathlib, so the `sorry_count: 0`
and axiom claims are from the repo's own `formalization.yaml` and its CI rather than something
rerun here.

### DIFF
--- a/FormalConjectures/ErdosProblems/183.lean
+++ b/FormalConjectures/ErdosProblems/183.lean
@@ -1,0 +1,73 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# Erdős Problem 183
+
+*References:*
+- [erdosproblems.com/183](https://www.erdosproblems.com/183)
+- [Er61] Erdős, P., *Graph theory and probability. II*. Canad. J. Math. (1961), 346-352.
+- [OpenAI26] OpenAI, *Ten advances in mathematics and theoretical computer science*. (2026).
+-/
+
+open Filter
+
+open scoped Topology
+
+namespace Erdos183
+
+/-- A `k`-colouring of the edges of `K_n` forces a monochromatic triangle when every colouring
+admits three distinct vertices whose three edges share a colour. -/
+def ForcesMonochromaticTriangle (n k : ℕ) : Prop :=
+  ∀ c : Sym2 (Fin n) → Fin k, ∃ x y z : Fin n, x ≠ y ∧ x ≠ z ∧ y ≠ z ∧
+    c s(x, y) = c s(x, z) ∧ c s(x, y) = c s(y, z)
+
+/-- $R(3;k)$, the minimal `n` such that every `k`-colouring of the edges of `K_n` contains a
+monochromatic triangle. -/
+noncomputable def multicolourTriangleRamsey (k : ℕ) : ℕ :=
+  sInf {n : ℕ | ForcesMonochromaticTriangle n k}
+
+/--
+Let $R(3;k)$ be the minimal $n$ such that if the edges of $K_n$ are coloured with $k$ colours
+then there must exist a monochromatic triangle. Determine
+$$\lim_{k\to \infty}R(3;k)^{1/k}.$$
+
+There is no finite limit: $R(3;k)^{1/k}\to\infty$. This was established by OpenAI [OpenAI26]
+along with the explicit superexponential lower bound in
+`erdos_183.variants.explicit_lower_bound`.
+-/
+@[category research solved, AMS 5, formal_proof using lean4 at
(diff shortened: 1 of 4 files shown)


FormalConjectures/ErdosProblems/183.lean
```diff
@@ -0,0 +1,77 @@
...
+open Filter
+
+open scoped Topology
+
+namespace Erdos183
+
+/-- A `k`-colouring of the edges of `K_n` is triangle-free when no colour class contains a
+triangle. -/
+def TriangleFree {n k : ℕ} (C : SimpleGraph.TopEdgeLabeling (Fin n) (Fin k)) : Prop :=
+  ∀ colour : Fin k, (C.labelGraph colour).CliqueFree 3
```
mo271: Something like this could be added to FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Clique.lean or Coloring.lean
```lean
/-- A complete graph edge labeling is `m`-clique-free if no monochromatic subgraph contains an `m`-clique. -/
def SimpleGraph.TopEdgeLabeling.CliqueFree {V ι : Type*} (C : TopEdgeLabeling V ι) (m : ℕ) : Prop :=
  ∀ c : ι, (C.labelGraph c).CliqueFree m
```
an then use that here (untested).

mo271 (approved):
Thanks, LGTM, left one optional suggestion

FormalConjectures/ErdosProblems/183.lean
williamjblair: Done in a2b6dda3, though in `EdgeColouring.lean` rather than `Clique.lean`: it sits next to `IsEdgeColouring`, and `Clique.lean` has seven importers that would have picked up the `EdgeLabeling` dependency for it.

`TriangleFree` is gone and `ForcesMonochromaticTriangle` reads `¬ C.CliqueFree 3`.

mo271 (approved):
Thanks, LGTM!

PR #4717, opened 2026-08-04
docs: make agent workflow focused and proportional

williamjblair:
Reduced to ~150 lines. Most of it restated [CONTRIBUTING.md](CONTRIBUTING.md), so it now links there and keeps what an agent can't infer.

A few points:

- **Commands.** The old headline requirement was a full `lake --wfail build`, ~25 min warm, so nobody runs it. Now it's the module you touched, with the guillemet quoting numbered files need. Plus the two `--wfail` traps, `open Classical` and AMS tag order.
- **Degenerate cases**, where formalisations break and the old file said nothing. `ZMod 0` is `ℤ`, `x / 0` is `0`, `sInf ∅` is `0`. Four statements of mine this week were wrong or nearly wrong on one of those.
- **`#print axioms`.** A file with `sorry` still compiles. Same check for a proof cited with `formal_proof`.

On the size, the general convention is ~150 lines, Codex truncates past 32 KiB, and shorter files measurably do better.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -73,6 +73,37 @@ The same applies to a proof you cite with `formal_proof`. Read the file for `sor
 are claiming. A repository saying it proves a conjecture may prove something weaker, and the
 statement is often the same theorem under a different name.
 
+## Checking your own work
+
+Each of these has produced a wrong claim in this repository, and each is cheap to avoid.
+
+**Read what matched, not how many.** Searching the site for `coprime` returned four results,
+which looked like the search covering statement text. All four matched the theorem *name*.
+A count tells you nothing about why something matched.
+
+**`grep sorry` hits prose.** A file whose only two `sorry` matches were inside a comment
+describing a plan was nearly written off as incomplete. Use `#print axioms` to decide, and
+grep only to find where to look.
+
+**Check the data before describing it.** "The statements are already in `conjectures.json`"
+was wrong: `extract_names` is run with `--exclude=statement`, so they are not. Open the file.
+
+**Measure at the right moment.** An environment probe run from a later command says nothing
+about what an attribute saw while elaborating, because the declaration is rewound in between.
+If a claim is about when something happens, instrument that point.
+
+**A no-op edit looks like a successful one.** A `str.replace` whose pattern no longer matches
+changes nothing and reports nothing; so does a `PATCH` that writes back identical content.
+Assert that the edit landed, and re-read the file rather than the exit code.
+
+**Shell quoting eats Lean source silently.** An unquoted heredoc expanded `` `A` `` and `$K_n`
+in docstrings to nothing. The file still compiled, so only a reader noticed. Quote the
+delimiter, and reread any docstring a script wrote.
+
+**Dry-run anything that closes or deletes.** A comparison between an `int` and a set of
+strings matched nothing, which would have closed 34 of 35 issues rather than the 6 intended.
+Print what a destructive step would do, against real data, before letting it do it.
+
 ## Statement fidelity
 
 The docstring quotes the source, and the Lean says exactly that. When they can differ, the
@@ -96,6 +127,7 @@ left `sorry` exercises nothing.
 - [ ] `#print axioms` on anything claimed to be proved
 - [ ] no `sorry` under `FormalConjecturesForMathlib/`
 - [ ] `git status` before `git add`: generated files and `__pycache__` sweep in easily
+- [ ] any file a script edited has been reread, not just rebuilt
 - [ ] `Fixes #1, fixes #2` in the description, with the keyword repeated. `Fixes #1, #2`
       closes only the first
 - [ ] formalisation choices and caveats in the pull request description, not in the Lean file
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -27,7 +27,7 @@ Quote it. The guillemets in a numbered file are part of the module name.
 ```bash
 lake --wfail build FormalConjecturesForMathlib   # if you changed a shared definition
 lake --wfail test                                # if you changed anything under FormalConjecturesUtil
-lake --wfail build                               # over an hour; leave it to CI
+lake --wfail build                               # ~25 min warm; leave it to CI
 ```
 
 `--wfail` turns warnings into failures, which is how CI runs. Two that catch people out:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,140 +1,149 @@
 # Agent guidelines
 
-Formal Conjectures states open mathematical problems in Lean 4. It is a statement
-repository, not a proof repository: the value is in the statement saying what the source
-says, and almost every problem is `sorry`.
+Formal Conjectures states open mathematical problems in Lean 4. This is a repository of
+statements, not a repository of proofs. Each statement must say what its source says. Almost
+every problem contains `sorry`.
 
-[CONTRIBUTING.md](CONTRIBUTING.md) is the reference for conventions, folders, and the
-attributes. This file is what an agent needs on top of it.
+[CONTRIBUTING.md](CONTRIBUTING.md) gives the conventions, the directories and the attributes.
+This file adds only what CONTRIBUTING does not tell you.
 
-For the shape of a file, copy one rather than a template. *File structure conventions* in
-CONTRIBUTING gives the skeleton, and
-[`FormalConjectures/ErdosProblems/13.lean`](FormalConjectures/ErdosProblems/13.lean) is 62
-lines showing the rest: a definition with its docstring, a `research solved` headline citing
-its source, and a `research open` variant using `answer(sorry)`. A file in the repository
-cannot drift away from the repository.
+Copy an existing file. Do not copy a template. *File structure conventions* in CONTRIBUTING
+gives the skeleton.
+[`FormalConjectures/ErdosProblems/13.lean`](FormalConjectures/ErdosProblems/13.lean) gives the
+remainder in 62 lines: a definition with its docstring, a `research solved` statement with its
+source, and a `research open` variant with `answer(sorry)`. A template can become incorrect. A
+file in the repository cannot, because the build compiles it.
 
 ## Commands
 
-Build the module you touched, not the project:
+Build only the module that you changed. Do not build the project.
 
 ```bash
 lake --wfail build 'FormalConjectures.ErdosProblems.«361»'
 ```
 
-Quote it. The guillemets in a numbered file are part of the module name.
+Put the module name in quotation marks. The guillemets in a numbered file are part of the
+module name.
 
 ```bash
 lake --wfail build FormalConjecturesForMathlib   # if you changed a shared definition
-lake --wfail test                                # if you changed anything under FormalConjecturesUtil
-lake --wfail build                               # ~25 min warm; leave it to CI
+lake --wfail test                                # if you changed FormalConjecturesUtil
+lake --wfail build                               # 25 minutes when warm. Let CI do this.
 ```
 
-`--wfail` turns warnings into failures, which is how CI runs. Two that catch people out:
+`--wfail` makes each warning into a failure. CI uses the same option. Two warnings cause most
+of these failures:
 
-- `open Classical` trips `linter.style.openClassical`. Use `open scoped Classical in` on the
-  declaration that needs it, or a `Decidable` instance.
-- AMS tags must be ascending: `AMS 15 51`, not `AMS 51 15`.
+- `open Classical` causes the `linter.style.openClassical` warning. Write
+  `open scoped Classical in` before the declaration that needs it. As an alternative, supply
+  a `Decidable` instance.
+- The AMS tags must be in ascending order. Write `AMS 15 51`. Do not write `AMS 51 15`.
 
-## Where a statement goes
+## Where to put a statement
 
-`FormalConjectures/<Source>/` for problems, one file per problem, named for it. Reusable
-mathematics goes in `FormalConjecturesForMathlib/`, which must be `sorry`-free. `sorry` is
-expected in `FormalConjectures/` and nowhere else.
+Put a problem in `FormalConjectures/<Source>/`. Use one file for each problem. Give the file
+the name of the problem. Put reusable mathematics in `FormalConjecturesForMathlib/`. That
+directory must not contain `sorry`. Only `FormalConjectures/` can contain `sorry`.
 
-Search before you define. Much of what a problem needs is already in Mathlib, in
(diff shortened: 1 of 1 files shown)


williamjblair:
oops didnt mean to post in official repo, found that agents.md was a bit too ceremonial so changed the one in my fork

mo271:
Perhaps there is something to be learned here, nonetheless? Wdyt?
 @Paul-Lez @williamjblair 
 The AGENTS.md was quickly written down and not yet iterated with a lot

williamjblair:
I didnt spend a lot of time on this one, and was still iterating so Ill add a proper one here we can review? I agree it would be good to update

mo271:
> I didnt spend a lot of time on this one, and was still iterating so Ill add a proper one here we can review? I agree it would be good to update

cool, no rush...

williamjblair:
Here is the AGENTS.md I've been using, it also includes some common mistakes etc and falls under the recommended length (~150 lines). Still let's iterate more if you or @Paul-Lez have ideas for improvement

AGENTS.md
```diff
@@ -1,564 +1,140 @@
...
-
-end ProblemName
-```
+- [ ] `lake --wfail build <module>` passes for what you touched
+- [ ] docstring quotes the source, with a reference in the module docstring
+- [ ] every theorem has `category` and at least one `AMS` tag, in ascending order
+- [ ] degenerate inputs tried: empty, zero, division
+- [ ] `#print axioms` on anything claimed to be proved
+- [ ] no `sorry` under `FormalConjecturesForMathlib/`
+- [ ] `git status` before `git add`: generated files and `__pycache__` sweep in easily
```
Paul-Lez: The pycache mention here seems a little strange - is this really needed?

AGENTS.md
```diff
@@ -1,564 +1,140 @@
...
-   def myFunction (x : ℕ) : ℕ := x + 1
-   ```
+- quantifier order and scope
+- `≤` against `<`, `∀ᶠ` against `∀`, asymptotic equivalence against same order
+- a hypothesis the prose states and the Lean drops
+- `∃ x, P x → Q`, which is almost always meant as `∃ x, P x ∧ Q` and is trivially true as
+  written. There is a linter for it
 
-5. **UpperCamelCase in snake_case contexts**: When something named with `UpperCamelCase` is part of something named with `snake_case`, it is referenced in `lowerCamelCase`:
+Prove the `test` and `API` statements you add. They exist to exercise a definition, and one
```
Paul-Lez: I don't think we should enforce this (often it's fine to leave these unproved if the proof is really long!)

AGENTS.md
```diff
@@ -1,564 +1,140 @@
...
-   class HasGδSingletons (X : Type*) [TopologicalSpace X] : Prop
-   structure MyStructure where
-     field1 : ℕ
-     field2 : String
-   inductive Color where
-     | Red | Green | Blue
-   ```
+**Dry-run anything that closes or deletes.** A comparison between an `int` and a set of
+strings matched nothing, which would have closed 34 of 35 issues rather than the 6 intended.
+Print what a destructive step would do, against real data, before letting it do it.
```
Paul-Lez: What's the intended meaning here?

AGENTS.md
```diff
@@ -1,564 +1,140 @@
...
+# Agent guidelines
 
-This document provides guidelines for AI agents working on the Formal Conjectures repository.
+Formal Conjectures states open mathematical problems in Lean 4. It is a statement
+repository, not a proof repository: the value is in the statement saying what the source
+says, and almost every problem is `sorry`.
 
-## Repository Structure
+[CONTRIBUTING.md](CONTRIBUTING.md) is the reference for conventions, folders, and the
+attributes. This file is what an agent needs on top of it.
```
Paul-Lez:
```suggestion
[CONTRIBUTING.md](CONTRIBUTING.md) is the reference for conventions, folders, and the
attributes.
```

AGENTS.md
```diff
@@ -1,564 +1,140 @@
-# Agent Guidelines for Formal Conjectures
+# Agent guidelines
 
-This document provides guidelines for AI agents working on the Formal Conjectures repository.
+Formal Conjectures states open mathematical problems in Lean 4. It is a statement
+repository, not a proof repository: the value is in the statement saying what the source
+says, and almost every problem is `sorry`.
```
Paul-Lez:
```suggestion
Formal Conjectures states open mathematical problems in Lean 4. It is a statement
repository, not a proof repository: almost every problem is `sorry`.
```

AGENTS.md
```diff
@@ -1,564 +1,140 @@
...
-When the problem is solved:
-
-- Replace `answer(sorry)` with `answer(True)` or `answer(False)` as appropriate
-- Update the category to `research solved`
-- If a formal proof exists, add `formal_proof using <kind> at "<link>"`
-
-**Note:** Providing a term inside `answer()` does NOT automatically mean the problem is mathematically solved - trivial or tautological answers don't count as solutions.
-
-### File Organization
+Anything else, `sorryAx` above all, means it is not proved.
```
Paul-Lez: I think we should beware here: sometimes we let users prove test statements with `native_decide`.

AGENTS.md
```diff
@@ -1,564 +1,140 @@
...
-   lemma add_comm (a b : ℕ) : a + b = b + a := by sorry
-   ```
+**A no-op edit looks like a successful one.** A `str.replace` whose pattern no longer matches
+changes nothing and reports nothing; so does a `PATCH` that writes back identical content.
+Assert that the edit landed, and re-read the file rather than the exit code.
 
-2. **Props and Types** (inductive types, structures, classes) use `UpperCamelCase`:
+**Shell quoting eats Lean source silently.** An unquoted heredoc expanded `` `A` `` and `$K_n`
+in docstrings to nothing. The file still compiled, so only a reader noticed. Quote the
+delimiter, and reread any docstring a script wrote.
```
Paul-Lez: Are we sure we want these in the AGENTS.md file? It's not clear to me that those fit in to the usual workflows that people have for this repo!

Paul-Lez (commented):
I think minimising this file is a good idea. Two things:
- we should have a think about what's the minimal amount necessary in this file. For a few things here (e.g. misformalisation checks, etc) perhaps these should go in separate files?
- It may be worth experimenting with the content of the file. For example, I think most frontier models now know that `Nat.sub` truncates, and so removing it might not change how good models are at formalising things in the repo.

Paul-Lez:
One little tip here: I've found that asking models to write in "ASD-STE100 Simplified Technical English" works super well to make the documentation they write more concise and clear. I think it might be worth trying this here too!

williamjblair:
Thanks, STE helps a lot! And great catches with the others. I agree AGENTS.md should stay lightweight while more specific instructions can be offloaded to skills. Re. your comment on Zulip, I'll open an issue related to improving AI docs/skills since I think that would also be very useful. Perhaps as @franzhusch said we can use some docs here as a starting point: https://github.com/ryantuck/formal-conjectures/pull/1

AGENTS.md
```diff
@@ -1,564 +1,141 @@
-# Agent Guidelines for Formal Conjectures
+# Agent guidelines
 
-This document provides guidelines for AI agents working on the Formal Conjectures repository.
+Formal Conjectures states open mathematical problems in Lean 4. It is a statement
+repository, not a proof repository: almost every problem is `sorry`. Each statement must say
+what its source says.
```
mo271, line 5: not really true: `test`, `API` and often `textbook` don't say and often don't even have docstrings.

williamjblair:
An update on this, since #4876 has moved.

You asked here what the minimal content of this file is, and whether some of it belongs in separate
files. #4899 is the first of those separate files. It takes the misformalisation material and the
review procedure out of this file's remit, so the cut here now has somewhere to point.

That makes the order matter. This one should merge first, because it rewrites the whole file.
#4899 adds five lines to the top of `AGENTS.md`, and I will rebase them onto the shortened version.

I also applied your other points from this review to #4899. It is in ASD-STE100, and I dropped a
table of junk values such as `Nat.sub`, for the reason you gave here.

AGENTS.md
```diff
@@ -1,564 +1,141 @@
...
 
-## Repository Structure
+[CONTRIBUTING.md](CONTRIBUTING.md) is the reference for conventions, folders, and the
+attributes. This file adds only what CONTRIBUTING does not tell you.
 
-### Main Directories
+Copy an existing file. Do not copy a template. *File structure conventions* in CONTRIBUTING
+gives the skeleton.
+[`FormalConjectures/ErdosProblems/13.lean`](FormalConjectures/ErdosProblems/13.lean) gives the
```
mo271: perhaps one could write something like "follow the style of an existing file, start by copying it"
I would not probably not refer to a specific file and definiely not hard code the number of lines here!

AGENTS.md
```diff
@@ -1,564 +1,141 @@
...
-  - `OptimizationConstants/` - Problems from [Tao's Optimization Constants](https://teorth.github.io/optimizationproblems/)
-  - Other sources as appropriate
+## Commands
 
-  **IMPORTANT**: Some subdirectories contain their own `README.md` with more specific conventions
-  (e.g. naming, docstring, and reference rules). Always check for and follow a subdirectory's
-  `README.md` before applying the general guidelines below. For example,
-  `FormalConjectures/ErdosProblems/README.md` defines naming patterns for multi-part questions,
-  estimate questions, and variants that supplement the general conventions in this file.
+Build only the module that you changed. Do not build the project.
```
mo271:
```suggestion
Build only the modules that you changed. Do not build the project.
```

AGENTS.md
```diff
@@ -1,564 +1,141 @@
...
+```bash
+lake --wfail build 'FormalConjectures.ErdosProblems.«361»'
 ```
 
-For variants of a problem, use dotted notation within the same namespace:
-
-```lean
-theorem main_conjecture : ... := by
-  sorry
+Put the module name in quotation marks. The guillemets in a numbered file are part of the
```
mo271: put the name in quotation marks is not necessary in my shell (perhaps it is overly specific? 
Is this really needed where you are running this command?

AGENTS.md
```diff
@@ -1,564 +1,141 @@
...
-@[category research solved, AMS 11]
-theorem main_conjecture.variants.weaker_version : WeakerStatement := by
-  sorry
-
-@[category test, AMS 11]
-theorem main_conjecture.variants.small_cases : MainStatement with (n < 100) := by
-  interval_cases n <;> decide
-```
+The three axioms above are the bar for a `research solved` statement. A `test` statement can
+use `native_decide`, which adds `Lean.ofReduceBool` and `Lean.trustCompiler`. These two move
```
mo271: I think native_decide might add other axioms, depending on what is going on?
also should we encourage 
`native +decide`?

AGENTS.md
```diff
@@ -1,564 +1,141 @@
...
-  sorry
+Put the module name in quotation marks. The guillemets in a numbered file are part of the
+module name.
 
-theorem main_conjecture.variants.special_case : ... := by
-  sorry
+```bash
+lake --wfail build FormalConjecturesForMathlib   # if you changed a shared definition
+lake --wfail test                                # if you changed FormalConjecturesUtil
+lake --wfail build                               # 25 minutes when warm. Let CI do this.
```
mo271: what does "warm" mean here? Mathlib cache?

25 min of course depends on the system. Since it perfectly parallelises, it can be faster on a system with many cpus/sufficient RAM: on my machine I just tried and I can build it in under 4min after `lake clean`. Of course doing `lake exe cache get` before to get the mathlib cache. 
Running `lake exe cache get` is perhaps good to add to the instructions here?!

AGENTS.md
```diff
@@ -1,564 +1,141 @@
...
-- `@[category test]` - Sanity check or unit test for definitions
-- `@[category API]` - Basic theory around a new definition
+Put a problem in `FormalConjectures/<Source>/`. Use one file for each problem. Give the file
+the name of the problem. Put reusable mathematics in `FormalConjecturesForMathlib/`. That
+directory must not contain `sorry`. Only `FormalConjectures/` can contain `sorry`.
 
-**Examples:**
+Search before you write a definition. Mathlib, `FormalConjecturesForMathlib/` and the adjacent
+problem files already contain much of what a problem needs. The names are difficult to guess.
+Examples: `trianglesContaining`, `InGeneralPosition`, `NonTrilinear`, `distinctDistances`.
```
mo271: perhaps add also notation for R², and `<<` as examples here?

mo271 (commented):
Thanks, looks good generally, added a few comments

AGENTS.md
Paul-Lez: As an aside, perhaps we should try to find a way to make the `ForMathlib` directory easier to search?

williamjblair:
Thanks, all six applied.

- **Copying a file.** Rewritten as "follow the style of an existing file, start by copying one from
  the same directory". No named file and no line count.
- **Build command.** Took your suggestion.
- **Quotation marks.** Removed. It was my shell, not a repository convention.
- **"25 minutes when warm".** Removed the number. It now says "the whole project. Let CI do this."
- **Notation.** Added `ℝ²` and `≪` to the examples.
- **`native_decide`.** You are right that the axiom set is not fixed. It now says `decide +native`,
  which the repo uses 128 times against 61 for `native_decide`, and that it adds *at least*
  `Lean.ofReduceBool` and `Lean.trustCompiler`, with the exact set depending on the proof. I also
  added that `decide +kernel` stays within the three axioms. I checked that:
  `#print axioms Erdos316.erdos_316`, which uses it, gives `[propext, Classical.choice, Quot.sound]`.

@Paul-Lez on searching `FormalConjecturesForMathlib`: agreed, esp the notation since you cannot grep for a definition whose name you have never seen. Naming two examples in this file is a patch rather than a fix.

One note on order. #4899 adds five lines to the top of this file, pointing at a new
`REVIEW_MATH.md` that holds the misformalisation material. That answers your first question here,
about what belongs in separate files. This pull request should merge first, and I will rebase those
five lines onto the shortened version.

mo271 (approved):
Thanks LGTM

We can always iterate further...

williamjblair:
Yes, something I've considered as well is having something like benchmarks at some point, this will be more applicable once we have skills and some other tooling (perhaps comparator as well). But essentially, the idea would be taking 10+ formalizations, and benchmarking on different variations of the workflow, skills, etc. My fear now is we don't know what is "good", aside from just semantically

Update: turns out this already exists - https://agentskills.io/skill-creation/evaluating-skills! Will try it out for the skills

PR #4722, opened 2026-08-04
ErdosProblems/337: use the asymptotic additive basis predicate

### LINKED ISSUE #4721: Erdős Problem 337 uses the exact `Set.IsAddBasis` where the asymptotic notion is meant

All four theorems in 337 use `Set.IsAddBasis` for "additive basis", but that is the exact notion, not the asymptotic one the problem is about.

https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/337.lean#L52-L58

`Set.IsAddBasisOfOrder A n` unfolds to `∀ a, a ∈ n • A`, so `hA` has to be all of `ℕ` rather than a tail of it. On `ℕ` that pins down two elements of `A` immediately:

```lean
example {A : Set ℕ} {n : ℕ} (hn : n ≠ 0) (h : A.IsAddBasisOfOrder n) : 0 ∈ A ∧ 1 ∈ A
```

`0` is a sum of `n` naturals only if all of them are `0`, and `1` only if one is `1` and the rest `0`. So every set the statement quantifies over contains `0` and `1`, and in particular `A ⊆ A + A`.

That is not the class Turjányi [Tu84] or Ruzsa and Turjányi [RT85] work with, and the statement itself gives it away: the counting function is `(A ∩ Icc 1 N).ncard`, counting from `1`, which only makes sense if `0` is not being forced into `A`.

The effect differs by theorem, since the class only shrinks:

- `erdos_337` and `variants.three_fold` and `variants.ruzsa_turjanyi` are `∀` over the class, so they say less than intended. `three_fold` understates what Ruzsa and Turjányi proved.
- `variants.h_fold` is `∃` over the class, so it asks for more than intended.

`erdos_337` still has the right answer either way, because Turjányi's counterexample `A` can be patched to `A ∪ {0, 1}` without changing `A(N)` or the ratio by more than a constant. But none of the four is the problem as posed.

### Suggested fix

Use `Set.IsAsymptoticAddBasis`, which is in the same ForMathlib file and already used correctly by 330, 868 and 881:

https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjecturesForMathlib/Combinatorics/Additive/Basis.lean

```lean
def IsAsymptoticMulBasisOfOrder (A : Set M) (n : ℕ) : Prop := ∀ᶠ a in cofinite, a ∈ A ^ n
def IsAsymptoticMulBasis (A : Set M) : Prop := ∃ n, A.IsAsymptoticMulBasisOfOrder n
```

That is `∃ h, hA` contains every sufficiently large integer, which is what the sources mean.

PR to follow.

### Additional context

Two other files touch the same predicate, and I am not proposing to change either:

- 326 uses `IsAddBasisOfOrder 2`. Bases of order 2 are sometimes stated exactly in the literature, so this one is a judgement call rather than a clear error.
- 741 uses `IsAddBasisOfOrder (A ∪ {0}) 2`. The `∪ {0}` works around the same forcing, which is some evidence that the exact predicate is the awkward default.

Found while checking whether the plby formalisation of 337 could be linked from #3998. It could not, because its basis hypothesis is the asymptotic one.

### Choose either option

williamjblair:
Fixes #4721.

All four theorems in 337 used `Set.IsAddBasis`, which is the exact notion:

```lean
def IsAddBasisOfOrder (A : Set M) (n : ℕ) : Prop := ∀ a, a ∈ n • A
```

On `ℕ` that forces `0` and `1` into `A` — `0` is a sum of `n` naturals only if all of them are `0`, and `1` only if one is `1` and the rest `0`. So the statement was quantifying over sets that all contain `0` and `1`, and in particular satisfy `A ⊆ A + A`. The counting function in the same statement is `(A ∩ Icc 1 N).ncard`, counting from `1`, which is a sign the exact reading was not what was wanted.

Turjányi [Tu84] and Ruzsa and Turjányi [RT85] work with asymptotic bases: some finite `h` has `hA` containing every sufficiently large integer. This switches all four theorems to `Set.IsAsymptoticAddBasis`, which lives in the same ForMathlib file and is already what 330, 868 and 881 use.

The class only shrinks under the old reading, so the effect differed by theorem: `erdos_337`, `variants.three_fold` and `variants.ruzsa_turjanyi` are `∀` over the class and said less than intended (`three_fold` understated Ruzsa–Turjányi), while `variants.h_fold` is `∃` and asked for more. `erdos_337` keeps `answer(False)` either way, since Turjányi's counterexample survives being patched to `A ∪ {0, 1}`.

I also added two sentences to the headline docstring saying which notion is meant, so the next reader does not have to unfold it.

The `formal_proof` link is now in on `erdos_337`. `Erdos337.not_erdos_337` in [plby/lean-proofs](https://github.com/plby/lean-proofs/blob/68da20b96673899166e94638f5a7fffeb7231d35/src/latest/ErdosProblems/Erdos337.lean) refutes the same proposition: its basis hypothesis is `∃ N₀, Set.Ici N₀ ⊆ iterated_sumset A k`, which is our `IsAsymptoticAddBasis` written out, and `#print axioms` gives `[propext, Classical.choice, Quot.sound]`. It indexes both counting functions by a real `x` through `⌊x⌋₊` where we index by `N : ℕ`, so the transfer needs a reindexing step in the little-o and in the `Tendsto`. I noted that in the docstring rather than leaving it implicit.

I did not link the variants. `not_erdos_337a` there covers `h = 2` only (its `ratio_fun A 3` is the two-fold sumset), so it does not give `variants.h_fold` for general `h`, and the three-fold and Ruzsa–Turjányi variants are not in that file at all.

`scripts/check_erdos_status.py` no longer reports 337.

### DIFF
--- a/FormalConjectures/ErdosProblems/337.lean
+++ b/FormalConjectures/ErdosProblems/337.lean
@@ -48,10 +48,14 @@ $$
 The answer is no, and a counterexample was provided by Turjányi [Tu84]. This was generalised (to
 the replacement of $A+A$ by the $h$-fold sumset $hA$ for any $h\geq 2$) by Ruzsa and Turjányi
 [RT85].
+
+"Additive basis" is `Set.IsAsymptoticAddBasis`: some finite $h$ has $hA$ containing every
+sufficiently large integer. The exact notion `Set.IsAddBasis`, which asks that $hA$ be all of
+$\mathbb{N}$, would force $0, 1 \in A$ and is not the class these results are about.
 -/
 @[category research solved, AMS 5 11]
 theorem erdos_337 : answer(False) ↔
-    ∀ A : Set ℕ, A.IsAddBasis →
+    ∀ A : Set ℕ, A.IsAsymptoticAddBasis →
       (fun N : ℕ ↦ ((A ∩ Icc 1 N).ncard : ℝ)) =o[atTop] (fun N : ℕ ↦ (N : ℝ)) →
       Tendsto (fun N : ℕ ↦ (((A + A) ∩ Icc 1 N).ncard : ℝ) / ((A ∩ Icc 1 N).ncard : ℝ))
         atTop atTop := by
@@ -63,7 +67,7 @@ by Ruzsa and Turjányi [RT85].
 -/
 @[category research solved, AMS 5 11]
 theorem erdos_337.variants.h_fold : ∀ h : ℕ, 2 ≤ h →
-    ∃ A : Set ℕ, A.IsAddBasis ∧
+    ∃ A : Set ℕ, A.IsAsymptoticAddBasis ∧
       (fun N : ℕ ↦ ((A ∩ Icc 1 N).ncard : ℝ)) =o[atTop] (fun N : ℕ ↦ (N : ℝ)) ∧
       ¬ Tendsto (fun N : ℕ ↦
           ((h • A ∩ Icc 1 N).ncard : ℝ) / ((A ∩ Icc 1 N).ncard : ℝ))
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

Issue #491, opened 2025-07-29
Erdős Problem 330


mo271:
### What is the conjecture
    
https://www.erdosproblems.com/330

  Suppose $A\subset\mathbb{N}$ is a minimal basis with positive density. Is it true that, for any $n\in A$, the (upper) density of integers which cannot be represented without using $n$ is positive?

Status: open

### Prerequisites needed
With Density from `ForMathlib` this should be straightforward. (I don't think we have `basis` yet, but an ad-hoc definition should do).

### Choose either option
- [ ] I plan on working on this conjecture
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

ayushdebnath012:
I want to work on this issue please assign me this issue @mo271

franzhusch:
Was added in #678

PR #4731, opened 2026-08-04
ErdosProblems/330: ask for positive upper density, not an existing density

### LINKED ISSUE #4730: Erdős Problem 330 requires a density to exist where the problem asks only for positive upper density

`erdos_330_statement` asks the unrepresentable set to *have* a density and for that density to be positive. The problem only asks for positive upper density.

https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/330.lean#L48-L52

```lean
theorem erdos_330_statement :
    answer(sorry) ↔ ∃ (A : Set ℕ), ∃ h, MinAsymptoticAddBasisOfOrder A h ∧ A.HasPosDensity ∧
    ∀ n ∈ A, Set.HasPosDensity (UnrepWithout A n h) := by
```

`Set.HasPosDensity S` is `∃ α > 0, S.HasDensity α`, so it carries the existence of the limit. `UnrepWithout A n h` is the set of integers not representable as a sum of at most `h` elements of `A \ {n}`, and there is no reason for such a set to have a natural density. Requiring one asks a strictly harder question.

The docstring in the file already says "the (upper) density", and so does #491, the issue this problem was added from:

> Suppose $A\subset\mathbb{N}$ is a minimal basis with positive density. Is it true that, for any $n\in A$, the **(upper)** density of integers which cannot be represented without using $n$ is positive?

So the intended reading is on record in two places; only the Lean disagrees.

### Suggested fix

`Set.upperDensity` is already in ForMathlib:

```lean
    ∀ n ∈ A, 0 < (UnrepWithout A n h).upperDensity
```

PR to follow.

### Additional context

I have left `A.HasPosDensity` alone, for the set `A` itself. The source text says "minimal basis with positive density" there with no parenthetical, and for a basis the natural-density reading is defensible, so that one is a judgement call rather than a clear error. A second opinion would help here, from someone who knows the additive-bases convention better than I do.

This does not on its own make the problem linkable. [Jayyhk/erdos-lean](https://github.com/Jayyhk/erdos-lean/blob/f8a51976fd2e66a52b4928c109fb9ae877a1a507/problems/330/Erdos330.lean) has a sorry-free proof of a statement in this shape, already phrased with `0 < Set.upperDensity`, but its representation set is `twoFold A \ twoFold (A \ {a})` where `UnrepWithout` is the complement of `Rep (A \ {n}) · h`, which allows *at most* `h` summands rather than exactly two. Those are different sets and neither contains the other, so a link would need that gap closed as well. Found while working through the status-sync backlog in #3998; #4047 tracks 330 as a plain status mismatch.

### Choose either option

williamjblair:
Fixes #4730.

`erdos_330_statement` asked the unrepresentable set to have a density and for that density to be positive:

```lean
∀ n ∈ A, Set.HasPosDensity (UnrepWithout A n h)
```

`Set.HasPosDensity S` is `∃ α > 0, S.HasDensity α`, so it carries the existence of the limit. `UnrepWithout A n h` is the set of integers not representable as a sum of at most `h` elements of `A \ {n}`, which has no reason to have a natural density. Asking for one is a strictly harder question than the one posed, and both the file's own docstring and #491, the issue this problem came from, say "the **(upper)** density".

Changed to:

```lean
∀ n ∈ A, 0 < (UnrepWithout A n h).upperDensity
```

and added a docstring sentence saying why, so the next reader does not switch it back.

I left `A.HasPosDensity` alone for the set `A` itself. The source says "minimal basis with positive density" there with no parenthetical, and the natural-density reading is defensible for a basis, so that is a judgement call rather than a clear error. Flagged in the issue for a second opinion.

The statement is `research open` with `answer(sorry)`, so nothing that was being claimed changes; what changes is which question is being asked.

### DIFF
--- a/FormalConjectures/ErdosProblems/330.lean
+++ b/FormalConjectures/ErdosProblems/330.lean
@@ -45,11 +45,16 @@ def MinAsymptoticAddBasisOfOrder (A : Set ℕ) (h : ℕ) : Prop :=
 Does there exist a minimal basis $A \subset \mathbb{N}$ with positive density
 such that, for any $n \in A$, the (upper) density of integers which
 cannot be represented without using $n$ is positive?
+
+The unrepresentable set is not asked to have a density, only to have positive upper density,
+so `Set.upperDensity` is used for it rather than `Set.HasPosDensity`. A set of integers that
+cannot be represented without $n$ has no reason to have a natural density, and requiring one
+would ask a strictly harder question than the one posed.
 -/
 @[category research open, AMS 5 11]
 theorem erdos_330_statement :
     answer(sorry) ↔ ∃ (A : Set ℕ), ∃ h, MinAsymptoticAddBasisOfOrder A h ∧ A.HasPosDensity ∧
-    ∀ n ∈ A, Set.HasPosDensity (UnrepWithout A n h) := by
+    ∀ n ∈ A, 0 < (UnrepWithout A n h).upperDensity := by
   sorry
 
 end Erdos330


mo271 (approved):
Thanks, LGTM!

Issue #4734, opened 2026-08-04
`turing_machine%` cannot declare its state inductive under the module system


williamjblair:
`turing_machine%` cannot declare its state type any more, so every use of it fails:

```
FormalConjecturesTest/ForMathlib/Computability/TuringMachine.lean:56:10: Unknown constant `State1`
FormalConjecturesTest/ForMathlib/Computability/TuringMachine.lean:59:10: Unknown constant `State2`
```

`mkStateType` builds the inductive from inside a term elaborator:

```lean
let stx ← `(command| inductive $(indName) $typeConstructors:ctor*)
liftCommandElabM <| elabCommand stx
```

`FormalConjecturesForMathlib/Computability/TuringMachine/Notation.lean` is a `public meta section`, and the declaration this adds is not visible to the elaborating module afterwards, so the very next step fails to find `State{n}`.

To be clear about what is and is not the problem: importing the module as `public meta import` (needed for the elaborator) plus a plain `public import` of `...TuringMachine.BusyBeavers` (needed for `Stmt.write`) clears every other error in that test file. `Unknown constant State{n}` is what remains, and it is inside the elaborator rather than at the import site.

Surfaced by #4700. I have commented out the `turing_machine%` examples in that test with a pointer here so the library can build; the halting tests above them now pass.

williamjblair:
Fixed in #4735.

`liftCommandElabM` runs the generated commands in a fresh scope that does not carry the surrounding module's visibility. Under the module system that makes them private, so the inductive was actually named `_private.<module>.0.State1` and the `checkTypeQ (.const stateName [])` on the next line could not find it. I confirmed that by listing constants matching `State1` after calling `mkStateType` directly.

Declaring the inductive and its `Inhabited` instance `public` in the quotation fixes the naming. That is the right call independently of the bug: the state type appears in the type of the term the elaborator returns, so it cannot be private without leaking into user code.

All nine `turing_machine%` tests are restored and pass.

williamjblair:
Fixed by #4735, merged. GitHub only auto-closed #4700 from that PR: `Fixes #4700, #4732, ...` needs the keyword before each number, so the rest stayed open.

Issue #4733, opened 2026-08-04
`#AMS` prints nothing outside the language server


williamjblair:
`#AMS` prints an empty list when compiled. It only works in the language server.

```lean
elab "#AMS" : command => do
  let env ← Lean.getEnv
  let lines ← (List.range 97).filterMapM fun n => do
    let nm : Name := Name.str ``AMS (ToString.toString n)
    if ← Lean.hasConst nm then
      if let some doc := ← Lean.findDocString? env nm then
        return s!"{n} {doc.trimAscii}"
```

`hasConst` finds all 62 constructors, but `findDocString?` returns `none` for every one of them, so the output is empty. Imported docstrings are only exported at `.server` level, which is why this looks fine while editing and produces nothing under `lake build`.

Docstrings are readable inside the module that declares them, so the list could be built once in `AMS.lean` and printed from a stored constant rather than recovered by lookup downstream.

Surfaced by #4700, since `FormalConjecturesTest/Util/Attributes/Basic.lean` has a `#guard_msgs` test for `#AMS` that had never been compiled. I have set that test to the empty output it currently produces, with a pointer here.

williamjblair:
Fixed in #4735.

The list is now read once in `AMS.lean`, where the constructor docstrings are still visible, and stored in a constant that `#AMS` prints. Two things that were not obvious: a declaration generated inside a quotation needs `mkIdent` rather than a quoted identifier, or it picks up a macro scope and no downstream module can find it, and it needs `@[expose]` or `value?` returns `none` for importers.

While restoring the test I also found an off-by-one unrelated to any of this: the loop ran over `List.range 97`, which stops at 96, while there is a `«97»` constructor. Mathematics education had never been in the output even in the editor. The restored test has all 63 subjects.

williamjblair:
Fixed by #4735, merged. GitHub only auto-closed #4700 from that PR: `Fixes #4700, #4732, ...` needs the keyword before each number, so the rest stayed open.

Issue #4732, opened 2026-08-04
The `category` attribute's sorry-free check never fires for theorems


williamjblair:
Once `FormalConjecturesTest` builds (#4700), this test fails:

```lean
/-- warning: If a problem has a sorry-free proof, it should not be categorised as `open`. -/
#guard_msgs in
@[category research open]
theorem test_theorem_with_docstring : 1 + 1 = 2 := by rfl
```

No warning is emitted. The check lives in the `category` attribute:

```lean
if status == .research .open then
  let env ← getEnv
  if (env.find? decl).bind (·.value?) |>.any (!·.hasSorry) then
    logWarning "If a problem has a sorry-free proof, it should not be categorised as `open`."
```

`applicationTime := .afterTypeChecking`, but theorem bodies are elaborated asynchronously, so at that point `value?` is still `none` for a theorem and the branch never runs. It does work for a `def`:

```lean
@[category research open]
theorem probe_term : 1 + 1 = 2 := rfl   -- no warning
@[category research open]
theorem probe_tac  : 1 + 1 = 2 := by rfl -- no warning
@[category research open]
def probe_def : Nat := 42                -- warns
```

Since every problem in the repo is a theorem, this check has been dead for a while. It would presumably have to move somewhere that runs after the proof is available, e.g. into `categoryLinter`, which already runs per command.

I have marked the test as recording current behaviour in the #4700 PR so the build can go green, with a pointer here. Whoever fixes this should flip it back.

williamjblair:
Fixed in #4735.

Moved out of the `category` attribute into `CategoryLinter.checkNotOpenIfSorryFree`. There was no way to make it work in the attribute: `.afterCompilation` does not help either, since theorems are not compiled. A linter runs once the command has finished, so `Environment.findAsync?` plus `toConstantInfo` waits on the elaboration task without deadlocking.

Behaviour I checked: `research open` and sorry-free warns; `research open` and sorried stays quiet; `research solved` and sorry-free stays quiet.

Nothing in the repo trips it — `lake build --wfail FormalConjectures` passes at 8950 jobs with the check live. That agrees with `scripts/extract_names.lean`, which turns out to have had the same check at whole-environment level all along, where `value?` is available. It writes to stderr behind a `|| true` in CI, so it has been correct but invisible.

williamjblair:
[...]

PR #4735, opened 2026-08-04
Make FormalConjecturesTest build, and put it in CI

### LINKED ISSUE #4700: FormalConjecturesTest never gets built, 8 of 10 modules fail

williamjblair:
Fixes #4700, #4732, #4733, #4734.

Rebased on #4777, which added the `lake --wfail test` step to CI and the `FormalConjecturesTest` library entry. The library entry still doesn't build the tests, though:

```toml
[[lean_lib]]
name = "FormalConjecturesTest"
globs = ["FormalConjecturesUtil.+"]
```

so `lake test` compiles `FormalConjecturesUtil` a second time and never touches a test file. The fix is `Util` to `Test` in that glob, and it is why the modules below stayed broken with CI green.

### Making the modules build

The test files need their util module as `public meta import`, not `public import`: a plain import registers neither the option nor the elaborator in time, so `set_option linter.style.moduleDocstring true` was an unknown option and the linters silently never ran. `Util/Answer` needs `import Lean.Expr`, which it was getting through a transitive meta import, and the Turing machine test needs `...TuringMachine.BusyBeavers` for `Stmt.write`.

The linter options the tests need are now `set_option` at the top of the files that test them, rather than lakefile entries, so a test carries its own switch.

### Three library bugs it turned up

**#4733.** `#AMS` read the subject names with `findDocString?` at the point of use, and imported docstrings are only exported at `.server` level, so it printed the right list in the editor and an empty one under `lake build`. The list is read once in `AMS.lean`, where the docstrings are still visible, and stored in a constant. Two things are commented in place because they are easy to trip over: the generated declaration needs `mkIdent` or it picks up a macro scope no downstream module can resolve, and it needs `@[expose]` or its value isn't exported. That also surfaced an off-by-one, `List.range 97` against a `«97»` constructor, so Mathematics education had never been listed. All 63 subjects are tested.

**#4732.** The "sorry-free proof categorised as `open`" check never fired for a theorem. It lived in the `category` attribute, and attributes run before a theorem's proof term exists, so `value?` was always `none`. It fired for a `def`, which is how I pinned it down and why the gap went unnoticed: every problem in the repo is a theorem, so the check was dead everywhere it mattered.

It is now `CategoryLinter.checkNotOpenIfSorryFree`, which runs after the command finishes and waits on the elaboration task with `findAsync?`. Resolving the name uses `expandDeclIdCore` plus the `_root_` and `private` handling from `mkDeclName`; `expandDeclId` itself calls `applyVisibility`, which throws `already been declared` once the declaration exists. The `private` case matters: 27 categorised declarations in the repo are private, and the check skipped all of them under the old `currNamespace ++ declId[0].getId`.

**#4734.** `turing_machine%` could not declare its `State{n}` inductive, so every use failed. `liftCommandElabM` runs the generated commands in a fresh scope that doesn't carry the module's visibility, so the inductive came out private, named `_private.<module>.0.State{n}`, and the lookup right after it failed with `Unknown constant State{n}`. Declaring it and its `Inhabited` instance `public` fixes the naming, and is right anyway since the state type appears in the type of the term the elaborator returns. All nine uses compile and the tests are restored rather than left commented out.

### Result

All 10 test modules build with every test restored, and none records broken behaviour. `lake --wfail test` and `lake --wfail build` both pass.

For a sense of what the gap allowed: #4368 carried literal `<<<<<<<` conflict markers in `FormalConjecturesTest/Util/Attributes/Basic.lean` for two weeks, through every green CI run, until Paul-Lez spotted them by eye.


williamjblair:
Pushed a fix for one of the three, #4733. `#AMS` now prints the real list under `lake build`, and its test is restored in full rather than recording the empty output.

Two things worth knowing if you hit the same pattern elsewhere: a declaration generated inside a quotation needs `mkIdent` or it gets a macro scope and becomes invisible downstream, and it needs `@[expose]` or `value?` comes back `none` for importers. Both are commented in the source.

It also shook out an off-by-one that had nothing to do with the module system: the range stopped at 96 while there is a `«97»` constructor, so Mathematics education was never in the output. The restored test has all 63.

#4732 and #4734 are still open and still recorded as current behaviour.

williamjblair:
#4732 fixed too, so two of the three are now done and only #4734 is left recording current behaviour.

The sorry-free check moved out of the `category` attribute into `CategoryLinter`. It could never have worked where it was: attributes run before a theorem's proof term exists, so `value?` was always `none`. The reason nobody noticed is that it *did* fire for a `def`, and every problem in the repo is a theorem.

A linter runs after the command has finished, so `findAsync?` can wait on the elaboration task without deadlocking. I checked all three cases behave: open and sorry-free warns, open and sorried stays quiet, solved and sorry-free stays quiet.

Worth a reviewer's attention since this turns a dead check live across the whole repo. `lake build --wfail FormalConjectures` passes at 8950 jobs, so nothing currently trips it, which matches what `extract_names` reports — it has had the same check at whole-environment level all along, printing to stderr behind a `|| true`.

williamjblair:
#4734 fixed as well, so all three are done and every test in the library is restored. Nothing is left recording broken behaviour.

The cause was the same shape as the `#AMS` one. `liftCommandElabM` runs the generated commands in a fresh scope that does not carry the module's visibility, so `inductive State1 ...` was declared private and landed as `_private.<module>.0.State1`. The lookup on the next line then failed. Declaring it `public` fixes the name, and it should have been public regardless: the state type appears in the type of the term the elaborator hands back, so a private one would leak into user code.

The nine `turing_machine%` tests are uncommented again.

`lake --wfail test` and `lake build --wfail FormalConjectures FormalConjecturesForMathlib` both pass.

williamjblair:
CI is green, and the new step is cheap: `Build test library` ran 23:29:34 to 23:29:47, so **13 seconds** on top of the existing build. It reuses everything the main build already compiled, which is why adding it to that job rather than a separate workflow matters.

That is also the first time `FormalConjecturesTest` has been compiled in CI at all.

williamjblair:
Rebased onto main now that #4777 has landed.

Two things came out of the rebase. #4777 already runs `lake --wfail test` alongside ForMathlib and Util, which is a better place for it than the separate step this had, so that step is gone.

The other is worth a look: on main the test library is declared as

```toml
[[lean_lib]]
name = "FormalConjecturesTest"
globs = ["FormalConjecturesUtil.+"]
```

so it globs Util rather than Test, and `lake --wfail test` compiles nothing under `FormalConjecturesTest/`. That is the same gap this PR is about, in a new form. Fixed here back to `FormalConjecturesTest.+`, and I checked it builds: touching a test module is a no-op because lake traces content hashes, but an actual edit rebuilds `FormalConjecturesTest.Util.Attributes.Basic`.

What is left after the rebase is the eight test modules that did not compile, plus the three library bugs that surfaced once they did: #4733, #4734 and #4732.

mo271 (approved):
Thanks, LGTM!

lakefile.toml
```diff
@@ -41,10 +41,15 @@ globs = ["FormalConjecturesUtil.*"]
...
-globs = ["FormalConjecturesUtil.+"]
+globs = ["FormalConjecturesTest.+"]
 [lean_lib.leanOptions]
 # Switch off warnings generated by `sorry`
 warn.sorry = false
+# the tests exercise our own linters, so they have to be on here too
+weak.linter.style.ams_attribute = true
+weak.linter.style.category_attribute = true
+weak.linter.style.moduleDocstring = true
+weak.linter.style.latex_docstring = true
```
eric-wieser: Another option would be to turn on the linters on the individual files that test them, which might be less annoying.

FormalConjecturesUtil/Linters/CategoryLinter.lean
```diff
@@ -53,6 +53,25 @@ def toCategory
       | _ => return false
   | _ => return #[]
 
+/-- Warns when a problem categorised as `research open` turns out to have a sorry-free proof.
+
+This check cannot live in the `category` attribute, which is where it used to be. Attributes run
+before a theorem's proof term is available, because theorem bodies elaborate asynchronously, so
+`value?` is still `none` there and the check silently never fired for a theorem. It did fire for
+a `def`, which is why the gap went unnoticed.
```
eric-wieser: Can't attributes choose to run at different execution phases, or are all of them before asynchronous proofs?

FormalConjecturesTest/Util/Linters/ExistsImplicationLinter.lean
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 -/
 module
 
-public import FormalConjecturesUtil.Linters.ExistsImplicationLinter
+public meta import FormalConjecturesUtil.Linters.ExistsImplicationLinter
```
eric-wieser: An alternative would be to edit `Linters/ExistsImplicationLinter.lean` to use `meta section`, which I think is used by the majority of our linters.

.github/workflows/build-and-docs.yml
```diff
@@ -125,6 +125,7 @@ jobs:
           lake --wfail build
           rm -f FormalConjectures/All.lean
 
+
```
eric-wieser:
```suggestion
```

FormalConjecturesUtil/Linters/CategoryLinter.lean
```diff
@@ -72,6 +91,12 @@ def categoryLinter : Linter where
         if prob_status.size == 0 then
           logLintIf linter.style.category_attribute outStx
             "Missing problem category attribute"
+          return
+        match stx with
+          | `(command| $_:declModifiers theorem $declId $_:bracketedBinder* : $_ := $_)
+          | `(command| $_:declModifiers lemma $declId $_:bracketedBinder* : $_ := $_) =>
```
eric-wieser: Here I think you can match `declSig` instead, rather than expanding it in the match.

FormalConjecturesUtil/Attributes/AMS.lean
```diff
@@ -188,13 +188,32 @@ meta unsafe def numToAMSSubjects (n : Nat) : MetaM AMS := do
...
-        return s!"{n} {doc.trimAscii}"
+        return some s!"{n} {doc.trimAscii}"
     return none
-  Lean.logInfo ("\n".intercalate lines)
+  -- `mkIdent` rather than writing the name in the quotation: a quoted identifier picks up a
+  -- macro scope, and the constant then has a hygienic name no downstream module can find.
+  let name := Lean.mkIdent (.mkSimple "amsSubjects")
+  elabCommand (← `(command| @[expose] def $name : String := $(quote ("\n".intercalate lines))))
+
+#recordAMSSubjects
```
eric-wieser:
```suggestion
run_command do
  let env ← Lean.getEnv
  let lines ← (List.range 98).filterMapM fun n => do
    let nm : Name := Name.str ``AMS (ToString.toString n)
    if ← Lean.hasConst nm then
       if let some doc := ← Lean.findDocString? env nm then
        return some s!"{n} {doc.trimAscii}"
    return none
  -- `mkIdent` rather than writing the name in the quotation: a quoted identifier picks up a
  -- macro scope, and the constant then has a hygienic name no downstream module can find.
  let name := Lean.mkIdent (.mkSimple "amsSubjects")
  elabCommand (← `(command| @[expose] def $name : String := $(quote ("\n".intercalate lines))))
```

eric-wieser: (something might be wonky with the github diff here, but the suggestion is really just `run_command`)

FormalConjecturesUtil/Attributes/AMS.lean
```diff
@@ -188,13 +188,32 @@ meta unsafe def numToAMSSubjects (n : Nat) : MetaM AMS := do
...
+
+/-- The `#AMS` outputs a list of the AMS Math Subjects and their corresponding indices -/
+elab "#AMS" : command => do
+  let some ci := (← Lean.getEnv).find? (.mkSimple "amsSubjects")
```
eric-wieser:
```suggestion
  let some ci := (← Lean.getEnv).find? ``amsSubjects
```
should do a compile-times check that the name is ok.

FormalConjecturesUtil/Linters/CategoryLinter.lean
eric-wieser: Namely, `afterTypeChecking`

williamjblair: Yes, and I tested it: `.afterCompilation` does see the proof term, so the check works there in an ordinary file.

It still does nothing inside a module though (and all ten files in `FormalConjecturesTest` are modules) so an attribute-based check has no way to be tested. That is what pushed it into the linter rather than the phase being wrong. I noted it in the docstring, and written up in #4780.

williamjblair:
All applied in c1f774d2. One change: `run_command` isn't defined here so it's `run_cmd`, and the doc comment above it had to become a plain comment since it no longer attaches to a declaration. Thanks for the thorough review!

FormalConjecturesUtil/Linters/CategoryLinter.lean
```diff
@@ -72,6 +92,12 @@ def categoryLinter : Linter where
         if prob_status.size == 0 then
           logLintIf linter.style.category_attribute outStx
             "Missing problem category attribute"
+          return
+        match stx with
+          | `(command| $_:declModifiers theorem $declId:declId $_:declSig $_:declVal)
+          | `(command| $_:declModifiers lemma $declId:declId $_:declSig $_:declVal) =>
+            checkNotOpenIfSorryFree declId
```
eric-wieser: Does this now not run on `def`s?

FormalConjecturesTest/Util/Linters/ExistsImplicationLinter.lean
eric-wieser: @williamjblair, I think this one is worth considering

FormalConjecturesUtil/Attributes/AMS.lean
```diff
@@ -188,13 +188,31 @@ meta unsafe def numToAMSSubjects (n : Nat) : MetaM AMS := do
...
     let nm : Name := Name.str ``AMS (ToString.toString n)
     if ← Lean.hasConst nm then
       if let some doc := ← Lean.findDocString? env nm then
-        return s!"{n} {doc.trimAscii}"
+        return some s!"{n} {doc.trimAscii}"
     return none
-  Lean.logInfo ("\n".intercalate lines)
+  -- `mkIdent` rather than writing the name in the quotation: a quoted identifier picks up a
+  -- macro scope, and the constant then has a hygienic name no downstream module can find.
+  let name := Lean.mkIdent (.mkSimple "amsSubjects")
```
eric-wieser:
```suggestion
  let name := Lean.mkIdent `amsSubjects
```

FormalConjecturesUtil/Attributes/AMS.lean
```diff
@@ -188,13 +188,31 @@ meta unsafe def numToAMSSubjects (n : Nat) : MetaM AMS := do
...
+  elabCommand (← `(command| @[expose] def $name : String := $(quote ("\n".intercalate lines))))
+
+/-- The `#AMS` outputs a list of the AMS Math Subjects and their corresponding indices -/
+elab "#AMS" : command => do
+  -- ``amsSubjects rather than a string: this fails to compile if the name above changes.
+  let some ci := (← Lean.getEnv).find? ``amsSubjects
+    | throwError "`amsSubjects` is missing"
+  let some (.lit (.strVal subjects)) := ci.value?
+    | throwError "`amsSubjects` is not a string literal"
+  Lean.logInfo subjects
```
eric-wieser: Does `Lean.logInfo amsSubjects` work?

FormalConjecturesUtil/Linters/CategoryLinter.lean
williamjblair: Good catch, it did not. Seven declarations have a category on a `def`, and those were the ones the attribute check actually worked for. I just added an arm for them in 78ac0dc2 and checked both a `def` and a `theorem` warn.

FormalConjecturesTest/Util/Linters/ExistsImplicationLinter.lean
williamjblair: Tried it: `public meta section` in place of the meta import gives the same `#guard_msgs` failures as a plain `public import`, so the linter is not registered by the time the file elaborates. `public meta import` seems to be what actually makes it fire, unless you meant something narrower than swapping the section?

FormalConjecturesUtil/Attributes/AMS.lean
williamjblair: It does once the generated definition is marked `meta` - as a plain `def` the elaborator can't refer to it ("`amsSubjects` not marked `meta`"). Done in 278c6894 so `#AMS` is now just that one line.

FormalConjecturesTest/Util/Linters/ExistsImplicationLinter.lean
eric-wieser: I mean `public meta section` in the linter file, not the test file.

williamjblair: Ah my bad, will do

FormalConjecturesUtil/Linters/CategoryLinter.lean
eric-wieser: I guess for consistency these should use `$declId:declId $_:declSig $_:declVal` too; that would also ensure that `theorem foo : forall l : List, ... | [] => _ | x :: xs => _` would be handled.

FormalConjecturesTest/Util/Linters/CategoryLinter.lean
```diff
@@ -74,7 +77,11 @@ Note: This linter can be disabled with `set_option linter.style.category_attribu
 example : 1 + 1 = 2 := by
   rfl
 
-/-- warning: If a problem has a sorry-free proof, it should not be categorised as `open`. -/
+/--
+warning: If a problem has a sorry-free proof, it should not be categorised as `open`.
```
eric-wieser, line 81: Please add a test for `lemma` and `example`; I think `example` doesn't work with your changes.

FormalConjecturesUtil/Linters/CategoryLinter.lean
eric-wieser: Perhaps worth raising this on Zulip, but no need to do so before wrapping up this PR. Thanks for investigating!

FormalConjecturesTest/Util/Linters/ExistsImplicationLinter.lean
williamjblair: You were right. The linter file opens `public meta section` and the test file is back to a plain `public import`.

Its imports had to move too: `Lean.Linter.Basic` and `FormalConjecturesForMathlib.Tactic.Linter.Term` are `public meta import` now, plus `Lean.Meta.AppBuilder` and `Lean.Elab.Command` which it was getting transitively.

FormalConjecturesUtil/Linters/CategoryLinter.lean
williamjblair: Done, both arms use `$declId:declId $_:declSig $_:declVal`. That does catch the match-alternative case:

```lean
theorem probe : ∀ l : List Nat, l = l
  | [] => rfl
  | _ :: _ => rfl
```

warns now and didn't before, though nothing in the repo is written that way today. `example` needed its own arm since it has no `declId`, so the attribute check is factored into `checkExactlyOneCategory`.

FormalConjecturesTest/Util/Linters/CategoryLinter.lean
williamjblair: Added, and you were right about `example`: it records no tag and leaves no constant, so there's nothing to look up. The attribute check still fires on them.

Tests: a sorry-free `lemma` marked `research open` (warns), the same as an `example` (doesn't), and the match-alternative theorem from the other thread.

FormalConjecturesTest/Util/Linters/CategoryLinter.lean
```diff
@@ -74,13 +77,47 @@ Note: This linter can be disabled with `set_option linter.style.category_attribu
...
+lemma test_lemma_sorry_free : 1 + 1 = 2 := by
+  rfl
+
+-- The `category` attribute on an `example` records no tag, and an `example` leaves no
+-- declaration to look a proof up under, so the sorry-free check cannot reach it. The
+-- attribute check above still applies.
+#guard_msgs in
+@[category research open]
+example : 1 + 1 = 2 := by
+  rfl
```
eric-wieser, line 107: I think the sorry-free check would reach it with the old version; perhaps we should do the check in the old place for `example`, or in two places?

williamjblair: I checked, and it doesn't reach it in the old place either. With one categorised theorem and one categorised example in a file:

```
1 tag(s)
  Probe.probe_thm
  env has Probe._example: false
  env has _example: false
```

The example leaves no tag and no constant, so `env.find? decl` has nothing to look up. Happy to add the second site if you can see a way to get at the declaration that I've missed.

eric-wieser: > The example leaves no tag and no constant, so env.find? decl has nothing to look up.

I believe the constant is in the environment while the attribute is elaborating, and then is rewound. I suppose it's possible that the constant remains in the environment even for the linter, but I suspect it is not.

williamjblair: You're right, my probe ran after the rewind. Logging `value?` from inside the `add` handler:

| | theorem | example | def |
|---|---|---|---|
| `.afterTypeChecking` (what it was) | none | none | present |
| `.afterCompilation`, ordinary file | present | present | present |
| `.afterCompilation`, module | none | none | present |

So the old check never fired for an `example` either, but at `.afterCompilation` in a problem file it would, and the linter can't. Downside is it can't be tested, since `FormalConjecturesTest` is modules. Want it back for that case?

eric-wieser: Oh, interesting, I didn't expect the behavior to match `theorem`. I guess we only care about the module version going forward.

FormalConjecturesForMathlib/Computability/TuringMachine/Notation.lean
```diff
@@ -149,7 +149,13 @@ section Main
 
 /-- `mkStateType n` adds to the environment an inductive type with `n` states
 names `State{n}` with constructors the symbols `A, B, ...`, together with the instance
-`Inhabited (State{n})`. -/
+`Inhabited (State{n})`.
+
+Both declarations have to be `public`. `liftCommandElabM` runs them in a fresh scope that does
+not carry the surrounding module's visibility, so without it they are declared private and end
```
eric-wieser, line 155: This surprises me, since this works:
```lean
import Mathlib

@[expose] public section

open Lean Elab Command

run_meta liftCommandElabM <| elabCommand <| ← `(def $(mkIdent `foo) := 1)

theorem foo_eq : foo = 1 := rfl
```
are you sure it's true?

eric-wieser: Ah sorry, I obviously forgot `module`!

FormalConjecturesTest/Util/Linters/CategoryLinter.lean
williamjblair: Surprised me too, I assumed `example` would be the odd one out. Will leave as is then

FormalConjecturesUtil/Attributes/AMS.lean
```diff
@@ -188,13 +188,26 @@ meta unsafe def numToAMSSubjects (n : Nat) : MetaM AMS := do
...
 
-/-- The `#AMS` outputs a list of the AMS Math Subjects and their corresponding indices -/
-elab "#AMS" : command => do
+-- Records the subject list as a constant.
+--
+-- This has to happen here, in the module that declares the constructor docstrings. Imported
+-- docstrings are only exported at `.server` level, so `Lean.findDocString?` returns `none` for
+-- every constructor once `AMS` is reached through an import: the list looks right in the editor
+-- and comes out empty under `lake build`. Reading the docstrings once, where they are still
+-- visible, and storing the result avoids that.
```
eric-wieser: Could use a `/-` comment here, since `/--` wasn't legal.

FormalConjecturesUtil/Linters/CategoryLinter.lean
```diff
@@ -53,25 +53,62 @@ def toCategory
...
+This check used to live in the `category` attribute, at `.afterTypeChecking`, where it never fired
+for a theorem: proof terms elaborate asynchronously, so `value?` was still `none`. It did fire for
+a `def`, which is why the gap went unnoticed.
+
+`.afterCompilation` is late enough to see the proof term in an ordinary file, but not inside a
+module, and every file in `FormalConjecturesTest` is a module. An attribute-based check therefore
+cannot be given a test. A linter runs once the command has finished and works in both, so the
+check lives here and `findAsync?` waits on the elaboration task. -/
+def checkNotOpenIfSorryFree (declId : Syntax) : CommandElabM Unit := do
+  let declName := (← getCurrNamespace) ++ declId[0].getId
```
eric-wieser: I think better to use [Lean.Elab.expandDeclId](https://leanprover-community.github.io/mathlib4_docs/Lean/Elab/DeclModifiers.html#Lean.Elab.expandDeclId) here, assuming its in our lean version.

eric-wieser:
Please update the description in light of the merged #4777

FormalConjecturesUtil/Linters/CategoryLinter.lean
williamjblair: This found a real bug, thanks. 27 of the categorised declarations here are private, and `currNamespace ++ declId[0].getId` gave the unmangled name for all of them, so the check was skipping them entirely. Test added.

`expandDeclId` itself I couldn't use, since it goes through `applyVisibility` and the linter runs after the declaration exists:

```
linter categoryLinter failed: a non-private declaration
`CategoryLinter.test_theorem_with_docstring` has already been declared
```

So it's `expandDeclIdCore` plus the `_root_` and `private` handling from `mkDeclName`.

eric-wieser: Perhaps worth a comment to that effect above `expandDeclIdCore`

eric-wieser (approved):
Thanks!

williamjblair:
> Thanks!

Thanks for all your help!

FormalConjecturesUtil/Linters/CategoryLinter.lean
williamjblair: Already there in 58f08a13, I think you were looking at the commit before it.

eric-wieser:
I've merged main in the hope that this restarts the CI that didn't run due to a GitHub outage.

williamjblair:
> I've merged main in the hope that this restarts the CI that didn't run due to a GitHub outage.

Thanks, I think they were running it in a slow queue, but this seems to run immediately now

eric-wieser:
> One character, but it's why the modules below stayed broken with CI green.

What do you mean by this? Which character?

williamjblair:
Four, not one, sorry. The glob says `FormalConjecturesUtil.+` under `name = "FormalConjecturesTest"`, so it is `Util` -> `Test`. I reworded it

PR #4736, opened 2026-08-04
fix(ErdosProblems): use lower density in 968

SamuelSchlesinger:
Erdős Problem 968 asks whether these sequences can have positive lower density. The Lean statement used `HasPosDensity`, which is stronger: it requires the ordinary natural density to exist and be positive.

This updates the main statement and the known decreasing-step result to say exactly that the lower density is positive. No unrelated notation is changed.


mo271 (approved):
Thanks, LGTM

PR #4737, opened 2026-08-04
fix(ErdosProblems): use upper density in 298

SamuelSchlesinger:
The source and cited result for Erdős Problem 298 concern positive upper density, but the main Lean theorem required an exact positive natural density. That stronger statement can fail when the density oscillates even though the upper density is positive.

This makes positive upper density the main statement and preserves the previous exact-density reading as an explicit variant.


mo271 (approved):
Thanks, LGTM

PR #4738, opened 2026-08-04
fix(ErdosProblems): state Romanoff's theorem with primes

SamuelSchlesinger:
The set used for Romanoff's theorem was defined by `primeFactors.card ≤ 1`. That also admits 1 and prime powers, while the theorem is specifically about integers of the form prime plus a power of two.

This states the set directly using a prime `p` and an exponent `k`, matching the cited theorem.

### DIFF
--- a/FormalConjectures/ErdosProblems/851.lean
+++ b/FormalConjectures/ErdosProblems/851.lean
@@ -32,14 +32,14 @@ prime divisors.
 -/
 def TwoPowAddSet (r : ℕ) := {(2 ^ k + n) | (k : ℕ) (n : ℕ) (_ : n.primeFactors.card ≤ r)}
 
+/-- The set of integers of the form `2^k+p`, where `k ≥ 0` and `p` is prime. -/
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/851.lean
```diff
@@ -32,14 +32,14 @@ prime divisors.
 -/
 def TwoPowAddSet (r : ℕ) := {(2 ^ k + n) | (k : ℕ) (n : ℕ) (_ : n.primeFactors.card ≤ r)}
```
mo271, line 34:
```suggestion
```

can this be delteted then?

mo271 (commented):
Thanks, one question about deleting the now unused def

FormalConjectures/ErdosProblems/851.lean
SamuelSchlesinger: It’s still used by the main erdos_851 theorem below, so it needs to stay.

mo271 (approved):
Thanks, LGTM!

PR #4739, opened 2026-08-04
fix(ErdosProblems): restore the ratio in 417

SamuelSchlesinger:
The Lean statement inverted the volume ratio from Erdős Problem 417. Taking the reciprocal changes both parts of the conjecture and is especially problematic when a volume is infinite.

This restores the numerator and denominator used by the source and states the ratio in `ENNReal`, where infinite volume is represented correctly.


mo271 (approved):
Thanks, LGTM

PR #4740, opened 2026-08-04
fix(ErdosProblems): record the answer to 1044

SamuelSchlesinger:
Erdős Problem 1044 has known infimum 2, and the file already included that value in a solved variant, but the main theorem still used `answer(sorry)`.

This records 2 directly as the answer to the headline question.


mo271 (approved):
Thanks, LGTM!

PR #4741, opened 2026-08-04
fix(ErdosProblems): use upper density in 741

SamuelSchlesinger:
The source's intended positive-density statement is about upper density. The main Lean theorem instead stated the stronger exact natural-density version, which is known to be false.

This makes positive upper density the main theorem and preserves the false exact-density reading as a named variant.


mo271:
This is just flipping the naming, right?

mo271 (approved):

PR #4742, opened 2026-08-04
fix(Wikipedia): make Frobenius traces integer-valued

SamuelSchlesinger:
For an elliptic curve, the Frobenius trace `a_p = p + 1 - #E(F_p)` can be negative. Representing it as a natural number truncates those negative values and changes the modularity statement.

This changes the trace to an integer-valued function, so the equation states the intended quantity.


mo271 (approved):
Thanks, LGTM!

PR #4743, opened 2026-08-04
fix(Books): record Borwein series convergence

SamuelSchlesinger:
The file cites and states the result that the Borwein sine series converges, but its headline theorem still left the yes-or-no answer unknown.

This replaces `answer(sorry)` with `answer(True)` so the statement records the known affirmative answer.


mo271 (approved):
Thanks, LGTM!

PR #4744, opened 2026-08-04
fix(Wikipedia): record the moving sofa answer

SamuelSchlesinger:
The theorem asks which member of the stated sofa family has the greatest area, but leaves that answer unknown even though the following result identifies Gerver's sofa as the maximizer.

This records the volume of Gerver's sofa as the answer to the question.


mo271 (approved):
Thanks, LGTM!

PR #4745, opened 2026-08-04
fix(ErdosProblems): record the answer to 1071(ii)

SamuelSchlesinger:
Part (ii) of Erdős Problem 1071 is marked solved and links a formal proof, but its theorem still used `answer(sorry)`.

This changes the answer to `True`, matching the cited affirmative solution.


mo271 (approved):
Thanks, LGTM

@Paul-Lez I thought we were already linting against that?

PR #4746, opened 2026-08-04
fix(ErdosProblems): record the maximizer for 434

SamuelSchlesinger:
The file describes the known interval that maximizes the number of coprime binomial coefficients, but the theorem leaves that interval unknown. Its range also allowed `k = 1`, whose gcd is `n`, contrary to the intended proper-divisor condition.

This records the known maximizing interval and restricts the range to `2 ≤ k ≤ n`, so the counted set matches the problem.


mo271 (approved):
Thanks, LGTM!

PR #4748, opened 2026-08-04
Fix two dead `formal_proof` links

williamjblair:
Two `formal_proof` links are dead. Both were unpinned `main` links, which is how they got that way.

| Problem | Was | Now |
| --- | --- | --- |
| 331 | `Woett/Lean-files/blob/main/ErdosProblem%23331.lean` | the file is `ErdosProblem331.lean`, no `#` |
| 392 | `.../blob/main/PrimeNumberTheoremAnd/Erdos392.lean` | moved upstream to `PrimeNumberTheoremAnd/IEANTN/Erdos392.lean` |

Both are pinned to a commit now rather than `main`, so an upstream rename cannot break them again.

I found these by checking every `formal_proof` link in the repo, 244 of them, against a `curl`. These two were the only ones that did not return 200, so the rest are healthy. Worth knowing that fork-hosted links are not the fragile ones: 61 of the 79 `formal_conjectures`-kind links point off `google-deepmind` and every one of them resolves. What rots is an unpinned path, wherever it is hosted.

That suggests a periodic link check would earn its keep, since nothing currently notices when one of these goes stale. Happy to add one as a scheduled workflow if that is wanted, though it is a separate change from this.

### DIFF
--- a/FormalConjectures/ErdosProblems/331.lean
+++ b/FormalConjectures/ErdosProblems/331.lean
@@ -41,7 +41,7 @@ any $n\geq 1$ there is exactly one solution to $n=a+b$ with $a\in A$ and $b\in B
 
 This was formalized in Lean by van Doorn using Aristotle.
 -/
-@[category research solved, AMS 11, formal_proof using lean4 at "https://github.com/Woett/Lean-files/blob/main/ErdosProblem%23331.lean"]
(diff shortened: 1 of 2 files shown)


mo271 (approved):
Thanks, LGTM!

Issue #4394, opened 2026-07-04
Formal Conjectures: Review, Verification, and Preservation Loop


williamjblair:
One `conjectures` toolkit for contribution review, exact-target proof verification and retained evidence. Contributors use their existing agents. FC owns deterministic operations; maintainers retain mathematical and merge decisions.

## Delivery map — 10 September 2026

**10 active FC infrastructure PRs: 3 foundations, 4 consolidated deliveries and 3 independent CI improvements.** Mathematical corrections and superseded prototypes are separate. No additional umbrella issue or replacement PR is needed.

| PR | Delivery | Required predecessors | Current disposition |
| --- | --- | --- | --- |
| #4899 | Review foundation and contracts | None | Open for review |
| #5375 | Shared metadata and one native `conjectures.json` | None | Open; final CI/deployment qualification |
| #5337 | Native proof-workspace exporter | Generator #7 | Draft; exact-pin qualification |
| #5386 | CLI **and Actions preparation** | #4899, #5375 | Active draft; absorbs #5356 |
| #5387 | Exact proof workspaces and verification | #5386, #5337, Comparator #87 | Active draft |
| #5388 | Evidence publication **and problem-page display** | #5386 | Active draft; absorbs #5389 |
| #4828 | Status **and proof-link metadata consumers** | #5375 | Active draft; absorbs #4749 |
| #5460 | Cancel superseded PR runs and lint workflows | None | Open for review; includes former #5469 |
| #5468 | Targeted problem-only PR validation | Maintainer acceptance of PR/queue coverage split | Draft; hosted targeted path qualified |
| #5461 | Reuse queue-built websites, retain Lean/cache refresh | #5435 | Draft; real queue-to-main qualification remains |

```mermaid
flowchart LR
  R["#4899 Review"] --> C["#5386 CLI + Actions"]
  M["#5375 Metadata"] --> C
  M --> S["#4828 Status + links"]
  C --> E["#5388 Publish + display evidence"]
  C --> P["#5387 Proof verification"]
  G["Generator #7"] --> X["#5337 Exporter"] --> P
  V["Comparator #87"] --> P
```

Review evidence can ship without proof verification. #5387 supplies proof records to #5388 when qualified. CI changes are a separate lane; their overlapping workflow changes must be reconciled as they land. #5435 is an external cache-fix dependency, not another PR created by this consolidation.

## PR and branch policy

- **Open draft:** active implementation with dependencies or qualification remaining.
[...]

PR #4749, opened 2026-08-04
Consolidated into #4828: proof-link checking

williamjblair:
**Consolidated into #4828 on 10 September 2026.**

The implementation and tests are retained in that PR's branch. This PR stays closed because its scope is now delivered together with status and proof-link metadata consumers; it is not a cancelled feature or a PR waiting to reopen.

Continue development and review in #4828. Existing commits and discussion here remain historical evidence. The authoritative delivery map is #4394.

### DIFF
--- a/.github/workflows/build-and-docs.yml
+++ b/.github/workflows/build-and-docs.yml
@@ -267,7 +267,7 @@ jobs:
         if: steps.mode.outputs.website_only != 'true'
         run: |
           mkdir -p site/data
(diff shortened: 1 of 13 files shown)


williamjblair:
Note: #4393 validates a link's **shape** at build time: present when the kind requires one, and starting `http://` or `https://`. It cannot check whether the thing at the other end exists, because the build has no network.

This checks **reachability**, which was the main issue in #4748: both of those links were well-formed and pointed at nothing. It runs out of band for the same reason, so the build stays offline.

Issue #4747, opened 2026-08-04
53 `test`/`API` statements have no sorry-free proof, and the check that says so is invisible


williamjblair:
`scripts/extract_names.lean` already checks this and reports it. Nobody sees it, because the workflow step is

```yaml
lake exe extract_names --exclude=statement,docstring,moduleDocstrings > site/data/conjectures.json || true
```

so the warnings go to stderr, behind a `|| true`, in the middle of a build log. Running it today, **59** declarations are categorised `test` or `API`, carry no `formal_proof` link, and have no sorry-free proof: 47 `test` and 12 `API`. That is out of 787 `test`/`API` statements.

*Update, 7 August:* down to **53** (41 `test`, 12 `API`) out of 822, after #4751 and #4752. The largest remaining cluster is seven in `WrittenOnTheWallII.Test`, graph invariants on named small graphs, then five in Erdős 91.

Both categories are meant to be proved. `test` is a sanity check for a definition and `API` is basic theory around one, so a `sorry` in either means the definition it exists to exercise is not being exercised.

These are real rather than an artefact of how the proof term is read. `value?` is available for all of them and `hasSorry` is `true`:

```
Erdos36.M_one: value?=true hasSorry=(some false)
Erdos36.M_two: value?=true hasSorry=(some true)
```

Most look like small decidable facts left for later. `Erdos1052.isUnitaryPerfect_87360` is explicit:

```lean
theorem isUnitaryPerfect_87360 : IsUnitaryPerfect 87360 := by
  -- TODO: Find a quicker proof. This one is too slow.
  stop
  norm_num [IsUnitaryPerfect, properUnitaryDivisors]
  decide +kernel
```

### Full list

| Module | n | Declarations |
| --- | --: | --- |
| `WrittenOnTheWallII.Test` | 7 | `C6_cvetkovic`, `C6_girth`, `K4_cvetkovic`, `Star5_cvetkovic`, `house_cvetkovic`, `petersen_cvetkovic`, `petersen_girth` |
| `ErdosProblems.91` | 5 | `circleSeven_optimal`, `dissimilar_circleSeven_wheelSeven`, `equiTriangle_unique_optimal`, `unitSquare_optimal`, `wheelSeven_optimal` |
| `OEIS.317940` | 5 | `a_1`, `a_2`, `a_3`, `a_4`, `a_5` |
[...]

PR #4750, opened 2026-08-05
OEIS/317940: prove the five term checks

williamjblair:
Part of #4747, which lists 59 `test`/`API` statements with no sorry-free proof. These five are the largest single-file group.

`a_1` through `a_5` assert the first five terms of A317940. They were all `sorry`, so nothing was checking that the recurrence in this file produces the sequence it claims to.

### How

`f` is defined by `WellFounded.fix`, so the tests cannot be discharged by `decide`. The useful move is one lemma:

```lean
private lemma f_eq (n : ℕ) :
    f n = if n = 0 then 0 else if n = 1 then 1 else
      (a046644 n -
        ∑ d ∈ n.divisors, if _h : 1 < d ∧ d < n then f d * f (n / d) else 0) / 2
```

`rw [f, WellFounded.fix_eq]` on its own unfolds the recursive occurrences too, which leaves a raw `fix` term that no later rewrite can fold back. Doing it under `conv_lhs` keeps the recursive calls as `f`, and everything else follows from that.

After that:

- `f_prime` handles `2`, `3` and `5` in one go. Every divisor of a prime is `1` or `p`, so the interior sum is empty.
- `a_4` is the only one with a nonempty interior sum. `4 = 2 ^ 2` contributes `f 2 * f 2`, and `a046644 4 = 8`, giving `f 4 = 7 / 2` and `a 4 = 7`.

The helpers are `private` and tagged `@[category API]`, since the linters require a category on every lemma.

### Result

`scripts/extract_names.lean` goes from 63 unproved `test`/`API` statements to 58. The five also had no docstrings, which `extract_names` was separately warning about, so they have one each now.

The file's one remaining `sorry` is `f_nonnegative`, the actual conjecture, which is `research solved` with a `formal_proof` link. That is the convention and I have left it alone.

### DIFF
--- a/FormalConjectures/OEIS/317940.lean
+++ b/FormalConjectures/OEIS/317940.lean
@@ -97,25 +97,84 @@ A317940: the numerator of `f n`.
 noncomputable def a (n : ℕ) : ℤ :=
   (f n).num
 
+/-- The defining recurrence of `f`, with the recursive calls folded back to `f`. -/
+@[category API, AMS 11]
+private lemma f_eq (n : ℕ) :
+    f n = if n = 0 then 0 else if n = 1 then 1 else
+      (a046644 n -
+        ∑ d ∈ n.divisors, if _h : 1 < d ∧ d < n then f d * f (n / d) else 0) / 2 := by
+  conv_lhs => rw [f, WellFounded.fix_eq]
+  rfl
+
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4751, opened 2026-08-05
ErdosProblems/36: prove M_two through M_five

williamjblair:
Part of #4747. Closes `M_two`, `M_three`, `M_four` and `M_five`.

Those four were `sorry` not because the values are unknown but because there was no way to get at them:

```lean
noncomputable def MaxOverlap (A B : Finset ℤ) : ℕ := iSup <| Overlap A B

noncomputable def M (n : ℕ) : ℕ :=
  sInf {MaxOverlap A B | (A : Finset ℤ) (B : Finset ℤ) (_ : Disjoint A B) ...}
```

An `iSup` over all of `ℤ` inside an `sInf` over an unbounded family of `Finset ℤ`. The blocker was the definition, not the statement.

### Confining both searches

| Lemma | What it gives |
| --- | --- |
| `overlap_le_card_left` | for a fixed `k` the first coordinate determines the second, so `Overlap A B k ≤ A.card` |
| `bddAbove_range_overlap` | hence the `iSup` is a genuine supremum, not the junk value `sSup` returns on an unbounded set |
| `maxOverlap_le_card_left` | the same bound for `MaxOverlap` |
| `overlap_eq_zero_of_notMem_sub` | an overlap vanishes off the differences actually realised |
| `maxOverlap_eq_sup` | `MaxOverlap A B = ((A ×ˢ B).image fun p => p.1 - p.2).sup (Overlap A B)` |
| `M_eq_image` | the pairs `M n` ranges over are exactly the `n`-element subsets of `{1, …, 2n}` with their complements |

### Making it compute

`Overlap` was already computable; only the two search operators were not. `maxOverlapC` is the `Finset.sup` form, `parts n` enumerates the subsets, and `M_eq_min'` rewrites `M n` as the minimum of a `Finset ℕ` via `Finset.Nonempty.csInf_eq_min'`. After that each value is `rw [M_eq_min']; decide`.

`M_five` searches the 252 five-element subsets of `{1, …, 10}` and needs `maxRecDepth 40000`, taking about 20 seconds. `M_four` needs `8000`. The other two are cheap.

### One judgement call

`M_two` carried a `formal_proof using formal_conjectures` link to a commit in #4153. That proof kind means the proof is at the other end of the link and the `sorry` here is expected. Since the `sorry` is now filled in place, I dropped the link rather than leaving it pointing at a commit that fills something already filled. Happy to put it back if the preference is to keep the provenance.

The three that had no docstring now have one, which `extract_names` was separately warning about.

### Effect

`scripts/extract_names.lean` goes from 63 unproved `test`/`API` statements to 59, and from 59 with no `formal_proof` link to 56.

### DIFF
--- a/FormalConjectures/ErdosProblems/36.lean
+++ b/FormalConjectures/ErdosProblems/36.lean
@@ -129,6 +129,42 @@ private lemma M_eq_image (n : ℕ) :
     · rw [Finset.card_sdiff, Finset.inter_eq_left.mpr hA, Int.card_Icc, hn]
       omega
 
+/-- A computable stand-in for `MaxOverlap`. `Overlap` is already computable; the only obstacle
+is the `iSup`, and `maxOverlap_eq_sup` says it agrees with this `Finset.sup`. -/
+private def maxOverlapC (A B : Finset ℤ) : ℕ :=
+  ((A ×ˢ B).image fun p => p.1 - p.2).sup (Overlap A B)
+
+@[category API, AMS 5 11]
+private lemma maxOverlap_eq_maxOverlapC (A B : Finset ℤ) :
+    MaxOverlap A B = maxOverlapC A B := maxOverlap_eq_sup A B
+
+/-- The `n`-element subsets of `{1, …, 2n}`. -/
+private def parts (n : ℕ) : Finset (Finset ℤ) :=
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4752, opened 2026-08-05
Wikipedia/ScholzConjecture: prove `additionChainLength_first_values`

williamjblair:
Part of #4747, and the same shape as #4751.

`additionChainLength_first_values` was `sorry` because there was no way to bound `ℓ` below:

```lean
noncomputable def additionChainLength (n : ℕ) : ℕ :=
  sInf { r | ∃ c : List ℕ, IsAdditionChain c ∧ c.getLast? = some n ∧ c.length = r + 1 }
```

An `sInf` over unbounded `List ℕ`. Writing down a chain gives an upper bound immediately; the lower bound is "no shorter chain exists", which nothing can reach without first confining the search.

### The bound

Every entry of a chain is a sum of two earlier entries, so a step at most doubles and a chain of `r` steps cannot pass `2 ^ r`:

```lean
private lemma getLast_le_two_pow {c : List ℕ} (h : IsAdditionChain c) (hne : c ≠ []) :
    c.getLast hne ≤ 2 ^ (c.length - 1)
```

Proved by reverse induction, resting on three small facts: a strictly increasing list is bounded by its last entry (`le_getLast`), every entry is at least `1` (`one_le_of_mem`), and dropping the last entry of a chain leaves a chain (`IsAdditionChain.dropLast`). That last one needs the observation that a summand can never be the final entry, since the entries are positive.

Transporting it to `ℓ` gives `n ≤ 2 ^ ℓ(n)`, and hence `lt_additionChainLength_of_two_pow_lt`, which is the lower-bound tool.

### The values

Nine of the ten then follow from that plus an explicit chain, for instance `ℓ(9) = 4` from `[1, 2, 4, 8, 9]` together with `2 ^ 3 < 9`.

`ℓ(7) = 4` is the exception, and the reason this problem is not just the doubling bound: `2 ^ 2 < 7` only gives `ℓ(7) ≥ 3`. Ruling out `3` needs the finite search the bound makes possible. A four-entry chain ending at `7` has all entries strictly between `1` and `7`, so it is `[1, x, y, 7]` with `1 < x < y < 7`, and `interval_cases` closes the twenty cases.

I also added a `Decidable` instance for `IsAdditionChain`. Every quantifier in it is bounded by the list, so a concrete chain is checkable by `decide`; without the instance Lean will not synthesise it through the `def`.

### Effect

`scripts/extract_names.lean` goes from 63 unproved `test`/`API` statements to 62, and from 59 with no `formal_proof` link to 58.

The file's two remaining `sorry`s are `scholz_conjecture` itself, which is `research open`.

### DIFF
--- a/FormalConjectures/Wikipedia/ScholzConjecture.lean
+++ b/FormalConjectures/Wikipedia/ScholzConjecture.lean
@@ -46,7 +46,130 @@ entries minus one) over all addition chains ending at $n$. -/
 noncomputable def additionChainLength (n : ℕ) : ℕ :=
   sInf { r | ∃ c : List ℕ, IsAdditionChain c ∧ c.getLast? = some n ∧ c.length = r + 1 }
 
+/-- Every quantifier in `IsAdditionChain` is bounded by the list, so membership is decidable
+and a concrete chain can be checked by `decide`. -/
+instance (c : List ℕ) : Decidable (IsAdditionChain c) := by
+  unfold IsAdditionChain; infer_instance
+
 local notation "ℓ(" n ")" => additionChainLength n
+
+@[category API, AMS 11 68]
+private lemma le_getLast {c : List ℕ} (h : c.Pairwise (· < ·)) {x : ℕ} (hx : x ∈ c) (hne : c ≠ []) :
+    x ≤ c.getLast hne := by
+  induction c using List.reverseRecOn with
+  | nil => simp at hne
+  | append_singleton ys y ih =>
+    rw [List.getLast_append_singleton]
+    rcases List.mem_append.mp hx with hy | hy
+    · exact le_of_lt ((List.pairwise_append.mp h).2.2 _ hy _ (by simp))
+    · simp only [List.mem_singleton] at hy; omega
+@[category API, AMS 11 68]
+private lemma one_le_of_mem {c : List ℕ} (h : IsAdditionChain c) {x : ℕ} (hx : x ∈ c) : 1 ≤ x := by
+  obtain ⟨hhead, hsorted, -⟩ := h
+  cases c with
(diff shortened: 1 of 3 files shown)


FormalConjectures/Wikipedia/ScholzConjecture.lean
```diff
@@ -46,8 +46,131 @@ entries minus one) over all addition chains ending at $n$. -/
 noncomputable def additionChainLength (n : ℕ) : ℕ :=
   sInf { r | ∃ c : List ℕ, IsAdditionChain c ∧ c.getLast? = some n ∧ c.length = r + 1 }
 
+/-- Every quantifier in `IsAdditionChain` is bounded by the list, so membership is decidable
+and a concrete chain can be checked by `decide`. -/
+instance (c : List ℕ) : Decidable (IsAdditionChain c) := by
+  unfold IsAdditionChain; infer_instance
+
```
mo271: might be good to move to `FormalConjecturesForMathlib`

mo271 (commented):
Thanks, LGTM, one minor suggestion

FormalConjectures/Wikipedia/ScholzConjecture.lean
williamjblair: Moved in 7c9cbb02. The instance needs `IsAdditionChain` with it, so `FormalConjecturesForMathlib/NumberTheory/AdditionChain.lean` has the definition, `additionChainLength`, and the general API: the doubling bound and the lower bound it gives.

The problem file keeps the conjecture, the values, and the `7` case, which is the only part about a specific number. `lake --wfail build` clean over the library.

mo271 (approved):
Thanks, LGTM!

PR #4758, opened 2026-08-05
ErdosProblems/298: pin the `formal_proof` link

williamjblair:
Rebased onto #4737, which landed the density change while this was queued and did it better than the version here did — upper density as `erdos_298` with the link on it, which is the statement the Lean 3 file actually proves.

What is left is the link itself: it is still `blob/master/…`. Both links fixed in #4748 rotted for exactly that reason, so this pins it to `10ef71a`.

Also adds a line naming the linked theorem, `unit_fractions_upper_density`, and noting it carries no `0 ∉ A` hypothesis — a solution containing `0` contributes `1 / 0 = 0` to the sum, so excluding `0` only narrows which sets are covered. Ours keeps it; the difference is now written down.

### DIFF
--- a/FormalConjectures/ErdosProblems/298.lean
+++ b/FormalConjectures/ErdosProblems/298.lean
@@ -36,9 +36,12 @@ upper density', which is likely what Erdős intended).
 The theorem below uses the positive-upper-density interpretation; the literal natural-density
 interpretation is recorded separately.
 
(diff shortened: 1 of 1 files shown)


williamjblair:
Superseded by #4737, which made the density change. What was left here was pinning one link, and 134 of the repo's 280 `formal_proof` links are unpinned branch refs, so singling out 298 was arbitrary.

PR #4759, opened 2026-08-05
ErdosProblems/470: record why `HasPosDensity` is the right reading

williamjblair:
Part of #4553. No change to any statement; this records why one should not be changed.

I had 470 on a list of files to switch from `Set.HasPosDensity` to positive lower density, on the general principle that Erdős' "positive density" usually means the lower one. Reading [BeEr74] settles it the other way.

Page 622, immediately before their Theorem 5:

> It is clear that the weird numbers have a density since both the abundant numbers and the pseudoperfect numbers have a density. (A weird number is abundant and not pseudoperfect.) Hence, we need only show that the density of weird numbers cannot be 0.
>
> **Theorem 5.** *The density of weird numbers is positive.*

So the density exists for an easy reason, and the paper says so: it is the density of the abundant numbers minus that of the pseudoperfect numbers, both of which exist. The work goes into showing the difference is not `0`. `HasPosDensity`, which is `∃ α > 0, HasDensity α`, is exactly that.

The same paper is careful about the distinction elsewhere, which is what made me check: on page 617 it says Erdős proved "the density of integers having property P **exists and is positive**", and on page 620 that "the density of the pseudoperfect numbers **exists**". The wording tracks a real difference rather than being loose.

I put this in the docstring rather than leaving it in a comment thread, since the next person to sweep for `HasPosDensity` will otherwise reach the same wrong conclusion I did.

### DIFF
--- a/FormalConjectures/ErdosProblems/470.lean
+++ b/FormalConjectures/ErdosProblems/470.lean
@@ -51,6 +51,13 @@ theorem erdos_470.parts.ii : answer(sorry) ↔ Set.Infinite PrimitiveWeird := by
 /--
 Benkoski and Erdős [BeEr74](https://mathscinet.ams.org/mathscinet/relay-station?mr=347726) proved
 that the set of weird numbers has positive density.
+
+`HasPosDensity` is the right reading here, rather than the positive *lower* density that Erdős'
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4761, opened 2026-08-05
Surface the extract_names category warnings

williamjblair:
Part of #4747.

`scripts/extract_names.lean` notices `research open` problems with a sorry-free proof, and `test` or `API` statements without one. It writes them with `IO.eprintln`, and the workflow step that runs it ends in `|| true`:

```yaml
lake exe extract_names --exclude=statement,docstring,moduleDocstrings > site/data/conjectures.json || true
```

So they land in the middle of a build log and nothing reads them.

### What it does

The three are not equally serious, so they are not treated the same way.

**`research open` with a sorry-free proof fails the build**, whenever there is one. It is a contradiction rather than a backlog: either the proof settles the problem and the category should be `research solved`, or it proves something weaker than the statement. There are currently none, and `CategoryLinter` in #4735 enforces the same thing from the Lean side.

**`test` and `API` without a proof are reported and do not block.** They go to `GITHUB_STEP_SUMMARY`, so the counts appear on the run page rather than needing someone to open a log:

> ### Category warnings
>
> - `test_without_proof`: 51
> - `api_without_proof`: 12
>
> These are advisory. A `test` or `API` statement without a proof is still better than no statement, and the proof often follows later.

### Checks

- Real output: prints the two counts, exits 0.
- With a `research open` violation appended: exits 1, naming it.
- With a new `test` statement without a proof appended: count goes to 52, still exits 0.

### On the earlier version

This started out as a baseline file that failed whenever any of the three counts grew. @mo271's review is right that that is wrong for two of them, and that making contributors refresh a checked-in baseline is friction for something that should not have been blocking in the first place. There is no baseline file now, which also resolves the escape-sequence nit on it.


scripts/category_warnings_baseline.json
```diff
@@ -0,0 +1,70 @@
...
+    "HartshorneConjecture.AlgebraicGeometry.Scheme.hasFiniteCoproductsVectorBundles",
+    "Kaplansky.promislow_group_is_torsionfree",
+    "LeinsterGroup.cyclic_of_perfect_is_leinster",
+    "RamanujanTau.multipliable",
+    "SimpleGraph.lemma3",
+    "SimpleGraph.lemma4"
+  ],
+  "research_open_with_proof": [],
+  "test_without_proof": [
+    "Arxiv.\u00ab1609.08688\u00bb.maximalLength_four",
```
mo271: nit: can we use `«` etc instead of the escape sequences here and everywhere?

mo271 (commented):
Thanks! 
Great idea to make these warnings more visible. 
For "research_open_with_proof" it should be completely made impossible for the CI pass without it. 
for the two other cases
"api_without_proof" and "test_without_proof", I think we should be more lenient with the following reasoning: 
 It is better to add a API/test statement without a proof than adding no such statement!
And we have seen the case often where people add test statements and shortly after somebody else adds a proof.

Perhaps we can make it a clearly visible warning instead for those cases? Right now people would have to adapt the baseline.json (for instance by running your script), perhaps we can make it more convenient somehow?

williamjblair:
Yes, great points. Reworked along those lines in 42904ca.

`research_open_with_proof` now fails whenever it is nonzero rather than when it grows, so there is no way for one to pass.

The other two no longer block and no longer need a baseline. They print, and go to `GITHUB_STEP_SUMMARY` so the counts land on the run page instead of in a log:

> ### Category warnings
>
> - `test_without_proof`: 51
> - `api_without_proof`: 12
>
> These are advisory. A `test` or `API` statement without a proof is still better than no statement, and the proof often follows later.

This also removes the baseline file entirely, so nobody has to update anything (and takes the `«` nit with it)

If you want the advisory counts more prominent than the run summary, the next step could be a sticky PR comment or something of the sort? just some ideas, open to more changes/ideas/improvements on this too

mo271 (approved):

mo271:
> If you want the advisory counts more prominent than the run summary, the next step could be a sticky PR comment or something of the sort? just some ideas, open to more changes/ideas/improvements on this to

I guess in some sense it would be nice to have an easier way of telling which of "Category warnings" are new. You approach before made that very clear, at the cost of keeping a record of offenders in a json. I suppose one option would be to tun the extract_names script twice, once at head? But doing this in each cl is costly/slow, since it requires having a complete `lake build` before, which is problematic especially for big changes (like version bumps..)

In any case what is nos in this pull request is an improvement over the status quo, thanks!

Issue #3979, opened 2026-05-10
Erdős 330: should the formal statement use positive upper density?


AllenGrahamHart:
The current formal statement of Erdős Problem 330 uses `Set.HasPosDensity` for both `A` and the unrepresented-without-`n` sets.

In the current FormalConjectures density API, `HasPosDensity` means existence of a positive natural density, not merely positive upper density. This appears stronger than the interpretation now recorded on the Erdős Problems page.

The Erdős 330 page says:

> Asked by Erdős and Nathanson. As with many of Erdős' questions, by 'positive density' here he most likely meant 'positive upper density'.

It also notes that in [Er80] Erdős was considering either lower or upper density for the set of integers not representable with a fixed `n`.

A Lean formalization of the positive-upper-density order-two version is available here:
https://github.com/AllenGrahamHart/FormalConjectures-Bench/tree/main/formalizations/erdos330

The isolated benchmark statement is:
https://github.com/AllenGrahamHart/FormalConjectures-Bench/blob/main/benchmark_sources/erdos330/UpperDensity.lean

The proved theorem states that there exists an order-two asymptotic basis `A` with positive upper density such that every element has a positive-upper-density private set.

Question for maintainers: should the FormalConjectures entry for Erdős 330 be adjusted to use upper density, should this be added as a solved variant, or should the current natural-density statement remain with an explanatory note?

Jayyhk:
> The current formal statement of Erdős Problem 330 uses `Set.HasPosDensity` for both `A` and the unrepresented-without-`n` sets.

#4762 and #4731 changed these both to use `0 < A.upperDensity` and `0 < (UnrepWithout A n h).upperDensity` respectively. This issue can now be closed.

mo271:
Thanks everybody!

PR #4762, opened 2026-08-05
ErdosProblems/330: fix representability, density, and a typo

williamjblair:
Fixes the four points @Jayyhk raised in https://github.com/google-deepmind/formal-conjectures/issues/3998#issuecomment-5187540855. I checked each against the source before changing anything; all four hold.

### `Rep` counted the wrong thing, twice

`Rep A m h` allowed *at most* `h` summands:

```lean
def Rep (A : Set ℕ) (m h : ℕ) : Prop :=
  ∃ k : ℕ, k ≤ h ∧ ∃ f : Fin k → ℕ, (∀ i, f i ∈ A) ∧ (∑ i : Fin k, f i) = m
```

Taking `k = 1` makes every element of `A` count as represented, so `UnrepWithout A n h` contained no element of `A` except possibly `n`. Those are exactly the integers the problem is about: if `a ∈ A` with `a ≠ n` is only expressible as `a = n + b`, it cannot be represented without `n` and belongs in the set, but the one-element sum `a` itself excluded it. Confirmed in Lean:

```lean
example (A : Set ℕ) (m h : ℕ) (hm : m ∈ A) (hh : 1 ≤ h) : Rep A m h :=
  ⟨1, hh, fun _ => m, fun _ => hm, by simp⟩
```

Separately, `Rep` used at most `h` while `MinAsymptoticAddBasisOfOrder` uses `IsAsymptoticAddBasisOfOrder`, which is `∀ᶠ a in cofinite, a ∈ h • A`, exactly `h`. The statement switched meaning of "representable" partway through. Erdős states the problem with `f_2(m)`, the number of solutions of `m = a_i + a_j`, which is exactly two.

Both are fixed by `Rep A m h := m ∈ h • A`, which is definitionally what the basis condition uses:

```lean
example (A : Set ℕ) (m h : ℕ) : Rep A m h ↔ m ∈ h • A := Iff.rfl
example (A : Set ℕ) (h : ℕ) :
    Set.IsAsymptoticAddBasisOfOrder A h ↔ ∀ᶠ a in Filter.cofinite, a ∈ h • A := Iff.rfl
```

One thing worth flagging for review, since it is silent if wrong: `h • A` for `h : ℕ` and `A : Set ℕ` could plausibly elaborate to scalar multiplication, `{h * a | a ∈ A}`. It does not; `pp.explicit` shows `@Set.NSMul`, the `h`-fold sumset, the same instance the basis predicate reaches through `to_additive`. I checked rather than assumed.

### Density

`A.HasPosDensity` becomes `0 < A.upperDensity`. #3979 asks for exactly this and notes the proof recorded as settling the problem establishes only positive upper density. #4731 made the same change for the unrepresentable set and deliberately left this half, which is now closed.

### Typo

`elements of `A`x`.

### DIFF
--- a/FormalConjectures/ErdosProblems/330.lean
+++ b/FormalConjectures/ErdosProblems/330.lean
@@ -25,15 +25,25 @@ import FormalConjecturesUtil
 namespace Erdos330
 
 open Set
-open scoped BigOperators
+open scoped BigOperators Pointwise
 
-/-- `Rep A m h` means `m` is a sum of at most `h` elements of `A`x. -/
-def Rep (A : Set ℕ) (m h : ℕ) : Prop :=
-  ∃ k : ℕ, k ≤ h ∧ ∃ f : Fin k → ℕ, (∀ i, f i ∈ A) ∧ (∑ i : Fin k, f i) = m
+/-- `Rep A m h` means `m` is a sum of exactly `h` elements of `A`, which is `m ∈ h • A`.
 
-/-- Integers **not** representable as a finite sum of elements with at most `h` terms of `A`
+Exactly `h` rather than at most `h`, for two reasons. It is the notion Erdős states the problem
+with, via `f_2(m)`, the number of solutions of `m = a_i + a_j` [Er80]. And it is the notion
+`IsAsymptoticAddBasisOfOrder` uses below, so the statement is not switching between two meanings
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4763, opened 2026-08-05
ErdosProblems: link Jayyhk proofs for 71, 209, 353 and 464

williamjblair:
On #3998 I listed why these four could not be linked. @Jayyhk rewrote them against the FC statements, so they can now. Pinned to `110d489`.

| Problem | What changed |
| --- | --- |
| 71 | now FC's right-hand side character for character |
| 464 | same |
| 209 | was over `ℂ`; now carries `PlaneR`, `IsLineR`, `pointMultiplicityR` and `HasGallaiTriangleR` matching FC's, and transfers the `ℂ` development across a `planeREquivComplex` |
| 353 | all five parts restated in FC's shape |

### 209

The obstruction was that the development modelled the plane as `ℂ` while FC uses `ℝ²`, and I said transferring would mean carrying an isomorphism through the whole statement. That is what the update does. `PlaneR` is `EuclideanSpace ℝ (Fin 2)`, and the four definitions are FC's verbatim, down to `pointMultiplicityR A p = {L ∈ ↑A | p ∈ L}.ncard`.

### 353

Two things I had flagged are both addressed. The areas were coordinate formulas where FC uses `volume (convexHull ℝ …)`; they are now the measure. And `IsoTrapArea1` had no convexity condition where FC's `IsIsoscelesTrapezoid` requires `IsCcwConvexPolygon`; `IsIsoscelesTrapezoidFC` now has it, and `IsCcwConvexPolygonFC` is FC's definition verbatim.

The theorem is a five-part conjunction matching FC's five statements one to one, including `MeasurableSet A` in the congruent-sides part, which the earlier version was missing. So the link goes on `erdos_353` and all four variants.

### Checks

All four files: no `sorry`, no declared `axiom`, `#print axioms` clean, and the headline theorem takes no hypotheses. `lake build --wfail FormalConjectures` passes, and `scripts/check_erdos_status.py` no longer reports any of the four; the repo total drops from 25 to 21.

90 stays unlinked. It is `axiomatic` in that catalog, assuming two published theorems, and Jayyhk says an axiom-free version is in progress.

### DIFF
--- a/FormalConjectures/ErdosProblems/209.lean
+++ b/FormalConjectures/ErdosProblems/209.lean
@@ -62,7 +62,7 @@ each of these intersection points only intersects two lines from $A$?
 Füredi and Palásti [FuPa84] showed this is false when $d\geq 4$ is not divisible by $9$.
 Escudero [Es16] showed this is false for all $d\geq 4$.
 -/
-@[category research solved, AMS 52]
+@[category research solved, AMS 52, formal_proof using lean4 at "https://github.com/Jayyhk/erdos-lean/blob/110d489ed5c07e5b216453e092e9113127c98c9a/problems/209/Erdos209.lean"]
 theorem erdos_209 : answer(False) ↔
     ∀ d : ℕ, 4 ≤ d → ∀ A : Finset (AffineSubspace ℝ ℝ²), A.card = d →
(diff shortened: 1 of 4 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4768, opened 2026-08-05
feat(Arxiv): add Atiyah–Sutcliffe Conjecture 1

### LINKED ISSUE #4756: Formalize Atiyah–Sutcliffe Conjecture 1
### What is the conjecture

Given a finite configuration of distinct points in `ℝ³`, every ordered
displacement determines a point of `ℂP¹` by stereographic projection. For each
configuration point, multiplying the associated homogeneous linear factors
produces a binary polynomial of degree `n - 1`. Atiyah–Sutcliffe Conjecture 1
says that the resulting `n` polynomials are linearly independent, equivalently
that the Atiyah determinant does not vanish.

References:
- https://doi.org/10.1098/rspa.2001.0913
- https://arxiv.org/abs/1102.4662

### Prerequisites needed

Mathlib already provides Euclidean space, complex numbers, multivariate
polynomials, finite products, and linear independence. The deterministic
projective representative is specific to this problem, so I do not propose a
new shared API.

### AMS categories

* ams-51
* ams-70

### Choose either option

derekste:
Closes #4756

## Summary

This PR models a configuration as an injective map from `Fin n` into
three-dimensional Euclidean space. It gives each displacement a deterministic
complex projective lift, turns that lift into a homogeneous linear factor,
and multiplies the factors associated with each configuration point. The
tests cover the north-pole and ordinary branches of the lift, the one-point
polynomial and linear-independence case, and a two-point polynomial expansion.
The research theorem is exactly Atiyah–Sutcliffe Conjecture 1: these point
polynomials are linearly independent.

## Formalization choices

Away from the north pole, `directionLift` chooses the unnormalized
representative `(z / w, 1)` of the paper's stereographic ratio. When the
denominator vanishes it uses `(1, 0)`, as required at the north pole.
Injectivity guarantees that displacements used by the conjecture are nonzero.
For a reversed displacement, the deterministic formula is projectively
equivalent to the paper's antipodal convention. All remaining normalization
and phase choices only rescale factors and point polynomials by nonzero
complex numbers, which does not change linear independence. The stronger
determinant inequalities in Conjectures 2 and 3 are intentionally out of scope.

## Verification

- Passed at `2bb9997e0c2b6815f3c3d0fe3f0b7ef6a8b653f0`: exact-head pipeline preflight
- The preflight included focused Lean elaboration and targeted `lake --wfail build`
- Rebased onto `origin/main` at `9118d083ffca1536f521f9a7d103201f537ea670`
- Passed in PR CI at `2bb9997e0c2b6815f3c3d0fe3f0b7ef6a8b653f0`: full-project `Build project`
- Stereographic ratio, pole convention, antipodal relation, and Conjecture 1 checked against the paper
- Current source/issue/PR collision searches checked
- No `native_decide`; only Conjecture 1's intended `sorry` remains

## AI use

AI assistance was used extensively to research this submission, draft and
refine the Lean formalization and tests, prepare the issue and PR text, and help
run local verification. I reviewed the formalization walkthrough and understand
the problem, the formal statement, its modeling choices, the completed examples,
and the role of the intentional `sorry`. I take responsibility for the submitted
content.

### DIFF
--- a/FormalConjectures/Arxiv/1102.4662/AtiyahSutcliffe.lean
+++ b/FormalConjectures/Arxiv/1102.4662/AtiyahSutcliffe.lean
@@ -69,6 +69,18 @@ theorem directionLift_xAxis :
   simp [directionLift]
 
 @[category test, AMS 51 70]
+theorem linearFactor_directionLift_ne_zero (v : Point) :
+    linearFactor (directionLift v) ≠ 0 := by
+  rw [directionLift]
+  split
+  · intro h
+    have := congrArg (MvPolynomial.coeff (Finsupp.single 0 1)) h
+    simp [linearFactor, coeff_X'] at this
+  · intro h
+    have := congrArg (MvPolynomial.coeff (Finsupp.single 1 1)) h
+    simp [linearFactor, coeff_X', Finsupp.single_eq_single_iff] at this
+
+@[category test, AMS 51 70]
 theorem onePoint_polynomial (x : Fin 1 → Point) : pointPolynomial x 0 = 1 := by
   simp [pointPolynomial]
 
--- a/FormalConjectures/Arxiv/1102.4662/AtiyahSutcliffe.lean
+++ b/FormalConjectures/Arxiv/1102.4662/AtiyahSutcliffe.lean
@@ -39,10 +39,16 @@ open MvPolynomial
 /-- A point of Euclidean three-space. -/
 abbrev Point := EuclideanSpace ℝ (Fin 3)
 
-/-- A deterministic projective lift of a direction in `ℝ³` to a pair of complex numbers.
+/-- A deterministic projective lift of a direction in $\mathbb{R}^3$ to a pair of complex
+numbers.
 
-Away from the north pole this is the representative `(z / w, 1)` from stereographic
-projection. At the north pole the denominator vanishes, so we use `(1, 0)`. -/
+Away from the north pole this is the representative $(z / w, 1)$ from stereographic
+projection. At the north pole the denominator vanishes, so we use $(1, 0)$.
+
+This is an unnormalized lift that does not impose the antisymmetric $\mathrm{SU}(2)$ convention
+of Atiyah–Sutcliffe. Since linear independence is invariant under rescaling each polynomial by
+a nonzero constant, this suffices for Conjecture 1. It would not suffice for Conjectures 2 or 3,
+which depend on the normalization $|z|^2 + |w|^2 = 1$. -/
 def directionLift (v : Point) : ℂ × ℂ :=
   if ‖v‖ - v 2 = 0 then
     (1, 0)
@@ -99,7 +105,9 @@ theorem twoPoint_xAxis_polynomial :
   rw [pointPolynomial, show (Finset.univ : Finset (Fin 2)) = {0, 1} by decide]
   simp [linearFactor, directionLift]
 
-/-- Atiyah--Sutcliffe Conjecture 1: the configuration polynomials are linearly independent. -/
+/-- [Atiyah–Sutcliffe Conjecture 1](https://doi.org/10.1098/rspa.2001.0913), stated as
+Conjecture 1.1 in [Mazur–Petrenko](https://arxiv.org/abs/1102.4662): the configuration
+polynomials are linearly independent. -/
 @[category research open, AMS 51 70]
 theorem conjecture_one {n : ℕ} (x : Fin n → Point) (hx : Function.Injective x) :
     LinearIndependent ℂ (pointPolynomial x) := by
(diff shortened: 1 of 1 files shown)


williamjblair:
References check out. `directionLift (c • v) = directionLift v` for `c > 0`, which I checked in Lean, so it is a function of the direction, and rescaling each factor by a nonzero constant leaves linear independence alone.

That rests on the lift never returning `(0, 0)`. If it could, every `pointPolynomial` would be zero and the conjecture false rather than open:

```lean
@[category test, AMS 51 70]
theorem linearFactor_directionLift_ne_zero (v : Point) :
    linearFactor (directionLift v) ≠ 0 := by
  rw [directionLift]
  split
  · intro h
    have := congrArg (MvPolynomial.coeff (Finsupp.single 0 1)) h
    simp [linearFactor, coeff_X'] at this
  · intro h
    have := congrArg (MvPolynomial.coeff (Finsupp.single 1 1)) h
    simp [linearFactor, coeff_X', Finsupp.single_eq_single_iff] at this
```

derekste:
Added `linearFactor_directionLift_ne_zero` as suggested. Thanks for checking this invariant so carefully.

FormalConjectures/Arxiv/1102.4662/AtiyahSutcliffe.lean
```diff
@@ -0,0 +1,110 @@
...
+@[category test, AMS 51 70]
+theorem twoPoint_xAxis_polynomial :
+    let x : Fin 2 → Point := fun i =>
+      if i = 0 then 0 else EuclideanSpace.single (0 : Fin 3) (1 : ℝ)
+    pointPolynomial x 0 = X 0 - X 1 := by
+  dsimp
+  rw [pointPolynomial, show (Finset.univ : Finset (Fin 2)) = {0, 1} by decide]
+  simp [linearFactor, directionLift]
+
+/-- Atiyah--Sutcliffe Conjecture 1: the configuration polynomials are linearly independent. -/
```
mo271:
```suggestion
/-- Conjecture 1 of Atiyah–Sutcliffe (Conjecture 1.1 in [Mazur–Petrenko]):
the configuration polynomials are linearly independent. -/
```

Let's mention the reference and where it is in the reference...

FormalConjectures/Arxiv/1102.4662/AtiyahSutcliffe.lean
```diff
@@ -0,0 +1,110 @@
...
+
+open MvPolynomial
+
+/-- A point of Euclidean three-space. -/
+abbrev Point := EuclideanSpace ℝ (Fin 3)
+
+/-- A deterministic projective lift of a direction in `ℝ³` to a pair of complex numbers.
+
+Away from the north pole this is the representative `(z / w, 1)` from stereographic
+projection. At the north pole the denominator vanishes, so we use `(1, 0)`. -/
```
mo271:
```suggestion
/-- A deterministic projective lift of a direction in `ℝ³` to a pair of complex numbers.

Away from the north pole this is the representative $(z / w, 1)$ from stereographic
projection. At the north pole the denominator vanishes, so we use $(1, 0)$. 

This is an unnormalized lift that does not impose the anti-symmetric SU(2) convention of
Atiyah–Sutcliffe. Since linear independence is invariant under rescaling each polynomial by a
nonzero constant, this suffices for Conjecture 1. It would not suffice for Conjectures 2 or 3,
which depend on the normalization $|z|^2 + |w|^2 = 1$.-/
```

does this make sense?

mo271 (commented):
Thanks, looks good to me mathematically just two suggestion for docstring changes

FormalConjectures/Arxiv/1102.4662/AtiyahSutcliffe.lean
derekste: Yes, this makes sense. Addressed in `2bb9997e` with the requested explanation
of the unnormalized lift, why projective rescaling preserves Conjecture 1, and
why the same choice would not suffice for Conjectures 2 or 3.

FormalConjectures/Arxiv/1102.4662/AtiyahSutcliffe.lean
derekste: Addressed in `2bb9997e`. The theorem docstring now links the original
Atiyah–Sutcliffe paper and identifies the statement as Conjecture 1.1 in
Mazur–Petrenko.

derekste:
Addressed both requested docstring changes in `2bb9997e`, rebased onto current
`main`, and reran the exact-head focused elaboration and targeted
`lake --wfail build` successfully. The full-project CI run on that commit is
green.

-awaiting-author

mo271 (approved):
Thanks, LGTM!

PR #4765, opened 2026-08-05
feat: formalise Cunningham chain conjecture (arXiv:1104.1579)

akanil18:
closes #4755 

## Summary

Adds a formal Lean 4 statement of Lenny Jones's conjecture on Cunningham chains
of exact length, as described in [arXiv:1104.1579](https://arxiv.org/abs/1104.1579).

## What was added

**New file:** `FormalConjectures/Arxiv/1104.1579/CunninghamChains.lean`

### Definitions
- `cunninghamFirst p n` — n-th term of a first-kind chain starting at `p`, iterating `q ↦ 2q + 1`
- `cunninghamSecond p n` — n-th term of a second-kind chain starting at `p`, iterating `q ↦ 2q - 1`
- `isExactFirstChain k p` — `p` starts a first-kind chain of exact length `k`
- `isExactSecondChain k p` — same for second-kind chains

### Conjectures (both `@[category research open, AMS 11]`)
- `infinitely_many_exact_first_chains` — for every `k ≥ 1`, infinitely many primes start an exact first-kind chain of length `k`
- `infinitely_many_exact_second_chains` — same for second-kind chains

### Tests (`@[category test, AMS 11]`, proved by `decide`)
- Verifies chain: 2 → 5 → 11 → 23 → 47 → 95 (first kind)
- Verifies chain: 19 → 37 → 73 → 145 (second kind)

## Notes
- No `FormalConjecturesForMathlib` additions required — uses only `Nat.Prime` and `Set.Infinite` from Mathlib
- Single new file, no existing files modified

PR #4770, opened 2026-08-05
feat(Arxiv): add Cunningham-chain conjectures

### LINKED ISSUE #4755: Formalize the Cunningham-chain infinitude conjectures
### What is the conjecture

A first-kind Cunningham chain repeatedly applies `q ↦ 2q + 1`, while a
second-kind chain repeatedly applies `q ↦ 2q - 1`. A chain has exact length
`k` when its first `k` terms are prime and the next generated term is
composite. Lenny Jones's paper records the conjecture that, for every positive
`k`, infinitely many starting primes generate exact chains of length `k`, for
each of the two kinds.

References:
- https://arxiv.org/abs/1104.1579
- https://oeis.org/A181697
- https://oeis.org/A181715

### Prerequisites needed

Problem-local recursive functions on natural numbers can represent the two
recurrences. Mathlib already supplies natural-number primality, bounded
quantification, and `Set.Infinite`, so no `FormalConjecturesForMathlib` API is
needed.

### AMS categories

* ams-11

### Choose either option

derekste:
Closes #4755

## Summary

This PR defines the natural-number iterates for first-kind and second-kind
Cunningham chains, together with predicates for exact positive length. It
proves the examples `2, 5, 11, 23, 47, 95`, an exact first-kind chain of length
5, and `7, 13, 25`, an exact second-kind chain of length 2. Two research
theorems then state the paper's conjecture that infinitely many starting
values give every positive exact length, separately for both kinds.

## Formalization choices

Terms are zero-indexed: term 0 is the starting value, terms below `k` must be
prime, and term `k` must be nonprime. That final condition distinguishes exact
length from merely having a prime prefix. The second-kind recurrence uses
natural subtraction to make a total natural-number function; underflow cannot
occur along the positive prime chains selected by the predicate. This indexing
is the zero-based translation of the source's `p₁, …, pₖ` convention.

## Verification

- `lake env lean FormalConjectures/Arxiv/1104.1579/CunninghamChain.lean`
- `lake --wfail build FormalConjectures/Arxiv/1104.1579/CunninghamChain.lean`
- Rebased onto current `origin/main`; PR CI supplies the authoritative clean full-project build
- Source definition and exact-length indexing checked against arXiv:1104.1579 and OEIS A181697/A181715
- Current source/issue/PR collision searches checked
- No `native_decide`; only the two intended conjecture `sorry`s remain

## AI use

AI was used extensively on this effort: specifically, OpenAI Codex using GPT-5.6 Sol with ultra reasoning effort. My intent at the onset of this project was to learn Lean, and thus sought a problem that might be particularly well-suited for a beginner. I used AI to comb open issues and search for a problem relating to prime numbers. I used AI in deciding how to structure the problem, and for some additional exploratory work on what a proof/disproof of this conjecture might look like (separate and not included in this statement-only PR). AI helped me stage the issue and PR, and although I have used my own words throughout as required, AI proofread everything for me just in case. I have personally reviewed and understand the submitted file, including both recurrences, zero-based indexing, natural subtraction, and the difference between exact and at-least length. I checked the two example chains and understand why the two open statements are separate.


FormalConjectures/Arxiv/1104.1579/CunninghamChain.lean
```diff
@@ -0,0 +1,74 @@
...
+are infinitely many chains of every positive exact length, of both kinds.
+
+*Reference:* Lenny Jones,
+[Polynomial Cunningham Chains](https://arxiv.org/abs/1104.1579)
+-/
+
+namespace CunninghamChain
+
+/-- The `n`th term generated from `p` by the first-kind recurrence `q ↦ 2q + 1`. -/
+def firstKindTerm (p : ℕ) : ℕ → ℕ
```
akanil18, line 35: theorem firstKindTerm_closed_form (p n : ℕ) :
  firstKindTerm p n = 2^n * p + (2^n - 1) := by
  induction n with
  | zero => rfl
  | succ n ih => by
    simp [firstKindTerm, ih, pow_succ]; ring

FormalConjectures/Arxiv/1104.1579/CunninghamChain.lean
```diff
@@ -0,0 +1,74 @@
...
+  | 0 => p
+  | n + 1 => 2 * firstKindTerm p n + 1
+
+/-- The `n`th term generated from `p` by the second-kind recurrence `q ↦ 2q - 1`. -/
+def secondKindTerm (p : ℕ) : ℕ → ℕ
```
akanil18, line 40: theorem secondKindTerm_closed_form (p n : ℕ) :
  secondKindTerm p n = 2^n * p - (2^n - 1) := by
  induction n with
  | zero => rfl
  | succ n ih => by
    simp [secondKindTerm, ih, pow_succ]; ring

theorem prime_no_underflow {p : ℕ} {i : ℕ} (h : (secondKindTerm p i).Prime) :
  1 < secondKindTerm p i := by
  exact Nat.prime_pos_iff.1 h
-- then you can use `1 < ...` to justify that subtracting 2^i - 1 does not underflow

FormalConjectures/Arxiv/1104.1579/CunninghamChain.lean
```diff
@@ -0,0 +1,74 @@
...
+  | 0 => p
+  | n + 1 => 2 * secondKindTerm p n - 1
+
+/-- `p` starts a first-kind Cunningham chain of exact positive length `k`. -/
```
akanil18, line 44: Please add a brief docstring here stating the zero‑based indexing and meaning of “exact length”: term 0 is the starting value, indices i < k must be prime, and index k must be composite -this helps future readers.

FormalConjectures/Arxiv/1104.1579/CunninghamChain.lean
derekste: Thanks. I tested this against the current project toolchain; it does not elaborate as written. The proposed lemma is also unused by the conjecture statements or tests, so I’m keeping this statement-only PR minimal.

FormalConjectures/Arxiv/1104.1579/CunninghamChain.lean
derekste: Thanks. I tested this against the current project toolchain; it does not elaborate as written. The proposed lemma is also unused by the conjecture statements or tests, so I’m keeping this statement-only PR minimal.

FormalConjectures/Arxiv/1104.1579/CunninghamChain.lean
derekste: deferring to maintainers

FormalConjectures/Arxiv/1104.1579/CunninghamChain.lean
```diff
@@ -0,0 +1,76 @@
...
+
+/-- The `n`th term generated from `p` by the first-kind recurrence `q ↦ 2q + 1`. -/
+def firstKindTerm (p : ℕ) : ℕ → ℕ
+  | 0 => p
+  | n + 1 => 2 * firstKindTerm p n + 1
+
+/-- The `n`th term generated from `p` by the second-kind recurrence `q ↦ 2q - 1`. -/
+def secondKindTerm (p : ℕ) : ℕ → ℕ
+  | 0 => p
+  | n + 1 => 2 * secondKindTerm p n - 1
```
mo271, line 42: nit: I would have followed the paper and define it with epsilon and then instantiate the first and second kind

mo271 (approved):
Thanks, LGTM!

mo271 (commented):
the examples from #4765 might be a good idea to have?
Either here or as a follow-up

akanil18:
Those examples are already in #4765 — should we go with that one?

mo271:
> Those examples are already in #4765 — should we go with that one?

sounds good to me, if it is also ok with @derekste, you could merge main when this pr is merged

derekste:
That works for me. Let's merge #4770 as the minimal statement-only formalization. After it lands, #4765 can update from `main` and be narrowed to the examples and any additional refinements, including the epsilon-based formulation if desired. This is ready to merge from my side.

-awaiting-author

derekste:
-awaiting-author

PR #4772, opened 2026-08-05
fix(Wikipedia): use n states for Busy Beaver

SamuelSchlesinger:
`BusyBeaver.Candidate n` currently assigns cardinality `n` to `Γ` and cardinality `2` to `Λ`. In the underlying `Machine Γ Λ` definition, however, `Γ` is the tape-symbol type and `Λ` is the state type. This therefore models two-state, n-symbol machines, while the file and cited Busy Beaver values are for n-state, two-symbol machines.

This swaps the two cardinality constraints and updates the canonical `Fin` sanity statement to `Machine (Fin 2) (Fin n)`. The definition now matches the standard two-symbol Busy Beaver function described by the cited source.

This was assisted by GPT 5.6 Sol via Codex.

### DIFF
--- a/FormalConjectures/Wikipedia/BusyBeaver.lean
+++ b/FormalConjectures/Wikipedia/BusyBeaver.lean
@@ -37,10 +37,10 @@ structure Candidate (n : ℕ) where
   Γ : Type
   Λ : Type
   Γ_fintype : Fintype Γ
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4773, opened 2026-08-05
fix(Paper): correct Nisan example sensitivities

SamuelSchlesinger:
The stated values do not match the Boolean function defined in this file.

`nisanExample n` accepts exactly the inputs of Hamming weight `n / 2` or `n / 2 + 1`. For a positive multiple of four, an input of weight `n / 2 + 2` has `n / 2 + 2` one-bits, and flipping any one of them moves into the accepting layer of weight `n / 2 + 1`. A direct layer count shows this is the global sensitivity, so the correct value is `n / 2 + 2`, not `n / 2`.

The block-sensitivity formula `3 * n / 4` also needs `8 ≤ n`. At `n = 4`, ordinary sensitivity is already `4`, so block sensitivity cannot equal `3`. The file's central-layer block construction gives the stated formula from `n = 8` onward.

The accompanying docstrings now record the corrected ranges and values.

This was assisted by GPT 5.6 Sol via Codex.

### DIFF
--- a/FormalConjectures/Paper/StrongSensitivityConjecture.lean
+++ b/FormalConjectures/Paper/StrongSensitivityConjecture.lean
@@ -102,12 +102,13 @@ theorem strong_sensitivity_conjecture {n : ℕ} (f : (Fin n → Bool) → Bool) 
   sorry
 
 
-/-- Simple test example,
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4774, opened 2026-08-05
Decide category warnings from the JSON, not the stderr

williamjblair:
Follow-up to #4761. The check reads `conjectures.json` now instead of parsing `extract_names` stderr:

```python
if category == "research open" and has_proof:  return "research_open_with_proof"
if category == "test" and not has_proof:       return "test_without_proof"
if category == "API" and not has_proof:        return "api_without_proof"
```

`category` and `hasSorryFreeProof` were all it ever needed, and rewording a warning can't quietly disable it any more.

That also lets `|| true` come off the extraction step. `IO.eprintln` doesn't set an exit code, so it wasn't tolerating warnings, just hiding crashes, and the old check would then see no warning text and pass.

Policy is the same as before: `research_open_with_proof` fails, `test`/`API` get counted in the run summary. Tests are stdlib-only, in a job that doesn't need the Lean build.

The "which warnings are new in this PR" part from #4761 isn't here. Notes on that are in #4394.

### DIFF
--- a/.github/workflows/build-and-docs.yml
+++ b/.github/workflows/build-and-docs.yml
@@ -31,6 +31,23 @@ permissions:
   contents: read
 
 jobs:
+  # The category check is a policy gate, so it has tests of its own. They need
+  # no Lean build, so they run alongside it rather than after it.
+  scripts:
+    runs-on: ubuntu-latest
+    name: Test scripts
+    steps:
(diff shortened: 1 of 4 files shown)


mo271 (approved):
Thanks, definitely better to read the .json instead of parsing the stderr output...

PR #4425, opened 2026-07-12
Add Erdős Problem 768 (research solved)

ericlisg:
Adds `FormalConjectures/ErdosProblems/768.lean`.

**Problem:** https://www.erdosproblems.com/768 — is there a constant c > 0 with
A(x)/x = exp(-(c + o(1)) √(log x) · log log x), where A counts integers whose
every prime factor p admits a divisor d > 1, d ≡ 1 (mod p)?

**Status:** research solved. Resolved in E. Li, *The Sylow Divisor Condition: a
Resolution of Erdős Problem 768*, arXiv:2606.24872 — the limit exists and the
constant is 1/(2√(log 2)).

**Formal proof:** https://github.com/ericlisg/erdos768-lean (tag v1.0.0,
Lean 4 / Mathlib v4.28.0). Main theorem `Erdos768.erdos_768` in
`RequestProject/Main.lean` is verbatim the `explicit_constant` variant in this
PR; the definitions here mirror the proof repository exactly so the match is
checkable by eye. No `sorry` in the dependency graph; `#print axioms` gives
`[propext, Classical.choice, Quot.sound]`, enforced via `#guard_msgs` in CI.
Formalised with Aristotle (Harmonic), directed and audited by the author.

Statement note: matches the erdosproblems.com display verbatim, including the
condition d > 1 (which is essential — with d = 1 permitted the condition holds
vacuously for every n); 1 ∈ A vacuously.

**Disclosure:** I am the author of the paper.

Co-authored-by: Aristotle (Harmonic) <aristotle-harmonic@harmonic.fun>

PR #4775, opened 2026-08-05
Build the site where it can be deployed, not on every PR

williamjblair:
The `deploy` job only runs on `main` and `-webtest` branches, so a PR can't deploy anything. It builds the whole site regardless, and the literate page render alone takes longer than compiling Lean. From the four most recent successful PR runs:

| | |
|---|---:|
| Build project (Lean) | 25, 26, 32, 26 min |
| Build literate source pages | 44, 43, 53, 43 min |
| total | 77, 78, 95, 78 min |

So this adds a `site` output next to the existing `website_only`, and gates the site half of the job on it. A PR that touches `site/`, `docbuild/`, or this workflow still builds the site. `main`, `-webtest` and `workflow_dispatch` are unchanged.

On the most recent successful PR run that's 77 minutes down to 29.

I checked the mode detection against real merge commits for the three cases (Lean-only PR, site-touching PR, push). `HEAD^1` is the base on a `pull_request` checkout, and if that diff fails for any reason it falls back to building the site.

### Merge queue

Since @mo271 asked for it above, this also adds `merge_group` to both workflows. `build-and-docs.yml` needs it so its checks report on the queue's branch, and `check_copyright_header.yml` was `pull_request` only, so a required check there would have left the queue waiting on a run that never starts.

The site and literate pages then build in the queue and on `main`, which is where breaking them matters, and a pull request skips them unless it touches the site itself.

`deploy` is unchanged, so the queue builds without publishing. One thing I can't see from here: `check-changes` isn't in `.github/workflows`, so if it is a required check it needs `merge_group` from wherever it is configured.

See the thread below on what the literate build does and doesn't currently catch, which is the open question on this one.


mo271:
One reason we have been building the literate sources pages (and not long ago even the docgen build) is to notice when a change introduced makes the litereate sources break. 
I guess one way to avoid this is start having a github merge queue, which would also be good catching break CI at HEAD like the one that happened here https://github.com/google-deepmind/formal-conjectures/pull/4697

Re the cache, I would also like to hear @eric-wieser's opinion

williamjblair:
Good point, I had this wrong. I assumed the literate build was only there for the preview.

Though I don't think it catches breakage today: both commands in that step end in `|| true` and the empty JSON files get deleted right after, so a module that stops rendering just drops out of the output and CI stays green. It hasn't failed once in the last 30 runs.

So that check needs building either way. I can put the literate build back on PRs and make it real by comparing which modules rendered against the base, or leave it out and let a merge queue cover it like you suggest. Happy either way, which would you rather?

Cache change is separable, I'll drop it from here while eric has a look.

Edit: moved cache change to #4776

williamjblair:
111 of the 262 open non-draft PRs import `FormalConjectures.Util.ProblemImports`. #4433 removed it on 17 July, so their green checks all predate the deletion. Only one has re-run since, #4425, and it failed. Merging any of them on that tick breaks main, same as #4697.

Requiring branches to be up to date before merging would catch it without a queue. Both cost a full CI run per merge.

If you do want the queue, none of our workflows trigger on `merge_group` yet, so checks wouldn't run inside it. Happy to add that.

mo271:
I think having the merge queue would be a step in the right direction: now I manually have to think to merge main manually when I think it will break things. How about we give this a try?
Let's add the correct workflow triggers here (with the benefit of running literal srouce *only* when in the merge queue and of course when being added to `main`). I will then enable the merge-queue on the github side

williamjblair:
Added in e907d822: `merge_group` on `build-and-docs.yml` and `check_copyright_header.yml`. The latter was `pull_request` only, so the queue would have waited on it forever.

Site and literate now run in the queue and on `main`, not on PRs. PRs touching `site/`, `docbuild/` or the workflow still build them; can drop that if you want.

`check-changes` isn't in `.github/workflows`, so if it's required it needs `merge_group` too.

eric-wieser:
Could you mention the merge queue change in the PR description? Otherwise LGTM.

williamjblair:
Added a merge queue section, and dropped a stray `.pyc` I had committed before `__pycache__/` was gitignored.

mo271 (approved):
Thanks, LGTM!

mo271:
merge queue enabled, let's see how it goes...

PR #4776, opened 2026-08-05
Stop pull requests writing to the shared olean cache

williamjblair:
Split out of #4775 so @eric-wieser has a diff to look at.

The repo's Actions cache sits at 9.92 GB of its 10 GB budget across 28 entries:

| scope | entries | size |
| --- | ---: | ---: |
| `refs/pull/*` | 19 | 6.63 GB |
| `refs/heads/*` | 9 | 3.28 GB |

Each olean entry is 0.36 GB and the save key includes `hashFiles('**/*.lean')`, so every PR writes its own. With the budget full, those evict each other.

A cache saved on a PR is visible only to that PR's later runs. It speeds up subsequent pushes on the same branch, and it evicts the `main` entries every PR falls back to through `restore-keys`. This drops the writes and leaves restores alone.

I'm not sure that's the right trade, which is why it's separate. If per-branch incremental builds matter more than fallback quality, say so and I'll close it.

Separately: the restore step uses `path: ~/.cache/mathlib` and the save step uses `path: ~/.cache`. If that's unintentional the archives are bigger than they need to be.

### Update, 8 August: the merge queue writes too

The queue went live today and the condition here missed it, because `merge_group` is not `pull_request`. Five entries in under three hours:

| ref | size |
| --- | ---: |
| `gh-readonly-queue/main/pr-4775-...` | 0.37 GB |
| `gh-readonly-queue/main/pr-4797-...` | 0.37 GB |
| `gh-readonly-queue/main/pr-3975-...` | 0.37 GB |
| `gh-readonly-queue/main/pr-4835-...` | 0.37 GB |
| `gh-readonly-queue/main/pr-4837-...` | 0.37 GB |

Those branches are deleted as the batch merges, so the entries are unreachable the moment they are written. That is 1.83 GB, accumulating at 0.37 GB per merge.

The condition is now `github.event_name == 'push'`. Current state:

| scope | entries | size |
| --- | ---: | ---: |
| pull requests | 18 | 6.58 GB |
| merge queue | 5 | 1.83 GB |
| `main` | 5 | 1.83 GB |
| total | 28 | 10.23 GB of 10 GB |


eric-wieser:
I'm actually no longer certain if this `lake exe cache pack` works at all; I think it only actually packs the mathlib oleans, not the formal-conjectures one.

> Separately: the restore step uses path: ~/.cache/mathlib and the save step uses path: ~/.cache. If that's unintentional the archives are bigger than they need to be.

This was probably an error, though I don't think there's much anywhere else in ~/.cache. I worry slightly that this causes the cahce to always miss, by saving `~/.cache/mathlib/foo` into `$artifact/mathlib/foo` and then restoring it to `~/.cache/mathlib/mathlib/foo`.

williamjblair:
Ordering note: this gates cache writes on `event_name != 'pull_request'`, and #4775 is what adds the `merge_group` trigger. If this lands first, merge-queue runs won't write the cache and the workflow won't report on them at all, so the queue stalls. Merging #4775 first avoids that; otherwise this needs `merge_group:` too.

williamjblair:
Had a look, and the cache never hits at all.

| | |
|---|---|
| caches stored | 27, at 374 MB each |
| total | 9.92 GB, against a 10 GB limit |
| #4796 alone | 5 of them |
| `main` | 3, saved today at 11:29, 13:04 and 18:07 |

`main` re-saving three times in a day is the giveaway. Here's a run from 17:59:

```
Cache not found for input keys:
  oleans-1c18a71c…-ca69cf33…, oleans-1c18a71c…-, oleans-
```

It even missed the bare `oleans-` prefix, with `main`'s 11:29 and 13:04 entries sitting right there. At the limit every new 374 MB entry pushes an old one out.

You were right about the paths, and it's on `main` too: restore uses `~/.cache/mathlib`, save uses `~/.cache`. The key is not helping and the second half is `hashFiles('**/*.lean')`, taken before `All.lean` gets generated and again after it's deleted (so a saved key can't match a later restore)

I couldn't settle the `cache pack` one: 7869 `.ltar` against 8612 `.lean` in the deps and 1121 in ours. This is fewer than the deps alone, which fits what you're saying, but not every `.lean` makes an `.ltar` so I don't want to call it on arithmetic.

williamjblair:
Updated now the merge queue is live. The condition here, `github.event_name != 'pull_request'`, did not cover `merge_group`, so queue builds were still writing.

Five entries since this morning, 0.37 GB each, scoped to `gh-readonly-queue/...` branches that GitHub deletes as the batch merges, so nothing can restore from them. That is 1.83 GB of the 10.23 GB the repo holds against a 10 GB budget, growing by 0.37 GB per merge.

Condition is now `github.event_name == 'push'`, which is the case where a later build can use the entry.

mo271 (approved):
Thanks, LGTM!

PR #4777, opened 2026-08-05
Fix build issues with FormalConjecturesUtil and FormalConjecturesTest

eric-wieser:
Also deletes `FormalConjecturesUtil.ForMathlib`; there is no point creating a brand new deprecated module under a different name to what it deprecates.


FormalConjecturesUtil/Linters/AnswerLinterTest.lean
eric-wieser, line 1: Presumably these test files should move to `Test`, but I'm declaring that out of scope of this PR.

FormalConjecturesUtil/Linters/LatexDocstringLinter.lean
```diff
@@ -16,13 +16,14 @@ limitations under the License.
...
 public import Lean
+public meta import Init.Data.String.Legacy
 
 /-! # The LaTeX Docstring Linter
 
 The `LatexDocstringLinter` ensures that docstrings use `$ $` or `$$ $$` for math formulas instead of `\[ \]` or `\( \)`.
 -/
 
-public section
+public meta section
```
eric-wieser, line 26: The test does not pass without this change. Note that most of the other linters use `meta section`, and the ones that don't are not tested!

Paul-Lez (approved):

mo271 (approved):

lakefile.toml
```diff
@@ -37,9 +37,11 @@ name = "FormalConjecturesForMathlib"
 
 [[lean_lib]]
 name = "FormalConjecturesUtil"
+globs = ["FormalConjecturesUtil.*"]
 
 [[lean_lib]]
 name = "FormalConjecturesTest"
+globs = ["FormalConjecturesUtil.+"]
```
YaelDillies, line 44: Surely that's wrong?

eric-wieser: Whoops, this is indeed wrong; did you already prepare a fix?

YaelDillies: No sorry, am AFK for another week

YaelDillies: This was fixed in #4735

PR #4790, opened 2026-08-06
feat(Wikipedia): add the Erdős–Moser conjecture

### LINKED ISSUE #4784: Formalize the Erdős–Moser conjecture
### What is the conjecture?

For positive integers $k$ and $m$, consider

$$1^k+2^k+\cdots+(m-1)^k=m^k.$$

The pair $(k,m)=(1,3)$ is a solution because $1+2=3$. The Erdős–Moser
conjecture says it is the only positive-integer solution.

Reference:

- B. C. Kellner,
  [“On stronger conjectures that imply the Erdős–Moser conjecture”](https://arxiv.org/abs/1003.1646)

### Prerequisites needed

Mathlib's finite intervals, finite sums, and natural-number powers are enough
to state the equation. No addition to `FormalConjecturesForMathlib` is needed.

### AMS categories

- ams-11

### Choose either option

derekste:
Closes #4784

## Summary

This PR formalizes the Erdős–Moser conjecture. It defines the sum of the first
$m-1$ positive $k$th powers, then states that any positive solution of
$1^k+\cdots+(m-1)^k=m^k$ must be $(k,m)=(1,3)$. Completed tests verify the
trivial solution and show that $(2,3)$ is not a solution. The main theorem has
the single intentional `sorry` because the conjecture remains open.

## Formalization choices

`powerSum k m` sums over `Finset.Ico 1 m`, so its endpoints match the paper's
$1$ through $m-1$ without subtraction or a special empty case. The theorem
assumes both variables are positive and concludes $k=1$ and $m=3$ directly;
this avoids introducing a separate solution predicate that would not be reused.
No `FormalConjecturesForMathlib` change is needed.

## Verification

- Passed: `env LEAN_NUM_THREADS=3 lake env lean FormalConjectures/Arxiv/1003.1646/ErdosMoser.lean`
- Passed: `git diff --check origin/main...HEAD`
- Confirmed: the diff adds only `FormalConjectures/Arxiv/1003.1646/ErdosMoser.lean`
- Full-project verification is delegated to this draft PR's CI checks
- Final source, issue, and PR collision searches completed against the current base
- No `native_decide`; exactly one intentional `sorry`

## AI use

AI assistance was used extensively to research this submission, draft and
refine the Lean formalization and tests, prepare the issue and PR text, and help
run local verification. I reviewed the formalization walkthrough and understand
the problem, the formal statement, its modeling choices, the completed examples,
and the role of the intentional `sorry`. I take responsibility for the submitted
content.

### DIFF
--- a/FormalConjectures/Arxiv/1003.1646/ErdosMoser.lean
+++ b/FormalConjectures/Arxiv/1003.1646/ErdosMoser.lean
@@ -53,4 +53,9 @@ theorem powerSum_one_three : powerSum 1 3 = 3 ^ 1 := by
 theorem powerSum_two_three : powerSum 2 3 = 5 ∧ powerSum 2 3 ≠ 3 ^ 2 := by
   decide
 
+/-- Zero is a solution for every positive exponent, so the conjecture must require $m>0$. -/
+@[category test, AMS 11]
+theorem powerSum_zero (k : ℕ) (hk : 0 < k) : powerSum k 0 = 0 ^ k := by
+  simp [powerSum, Nat.zero_pow hk]
+
 end ErdosMoser
--- a/FormalConjectures/Arxiv/1003.1646/ErdosMoser.lean
+++ b/FormalConjectures/Arxiv/1003.1646/ErdosMoser.lean
@@ -1,0 +1,56 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# The Erdős–Moser equation
+
+For positive integers $k$ and $m$, let
+
+$$S_k(m)=1^k+2^k+\cdots+(m-1)^k.$$
+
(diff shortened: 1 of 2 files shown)


mo271:
@derekste 
just checking: this pr and some other ones you recently opened are intentially still `draft`, i.e. not ready for review right?

derekste:
@mo271 yes draft was intentional: I wanted to give them a final once-over in the morning (and let my local verifications complete!). thanks for all the review and comments this week - I've had a lot of fun contributing to this project!!

williamjblair:
`0 < m` is necessary rather than cosmetic: `powerSum k 0` is an empty sum and `0 ^ k = 0` for `k > 0`, so `m = 0` satisfies the equation for every `k`. Suggest pinning it:

```lean
@[category test, AMS 11]
theorem powerSum_zero (k : ℕ) (hk : 0 < k) : powerSum k 0 = 0 ^ k := by
  simp [powerSum, Nat.zero_pow hk]
```

Compiles as written. Reference checks out (Kellner, arXiv:1003.1646), and both `decide` tests reduce fine.

derekste:
Added `powerSum_zero` as suggested to pin down why the `0 < m` hypothesis is essential. Thanks!

mo271 (commented):
Thank, LGTM mathematically, but I would move it to the `Wikipedia` dir: https://en.wikipedia.org/wiki/Erd%C5%91s%E2%80%93Moser_equation

Wikipedia also lists a bunch of results that might be good to add as `research solved`?!

I checked if this is also one of the problems on erdosproblems.com, but (to my surpirse) couldn't find it, perhaps my search was not good enough..

derekste:
Thanks for the review and the source-location suggestion. I moved the entry to `FormalConjectures/Wikipedia/ErdosMoser.lean` and added the Wikipedia page to the references, while retaining Kellner's arXiv paper as a supporting source. I'll keep the additional known solved results for a focused follow-up PR.

-awaiting-author

mo271 (approved):
Thanks, LGTM

If we want more statements from wikipedia, we can do it in a follow-up...

PR #4796, opened 2026-08-07
feat: Is factorization computable in polynomial time?

### LINKED ISSUE #4414: Integer factorization in P?

BoltonBailey:
This PR formalizes the question of whether integer factorization is computable in polynomial time.

Closes #4414

In the course of formalizing this and other questions related to polynomial time computability, I found it worked more easily to have a system for encodings that could be inferred from the type, so I have rewritten some of the content I submitted previously to support this.


FormalConjectures/Millenium/PvsNP.lean
```diff
@@ -49,17 +49,31 @@ The type of complexity classes. We define these as sets of decision problems.
...
+theorem isPolyTime_id {α : Type} [BitstringEncoding α] : IsPolyTime (id : α → α) :=
+  ⟨Turing.idComputableInPolyTime (BitstringEncoding.toFinEncoding α)⟩
+
 /--
 The class P is the set of decision problems
 decidable in polynomial time by a deterministic Turing machine.
 -/
 def P : ComplexityClass :=
-  { L | IsComputableInPolyTime finEncodingListBool finEncodingBoolBool L }
+  { L | IsPolyTime L }
```
Paul-Lez: What do you think of just moving a few of these useful preliminary definitions to their own file in the `FormalConjecturesForMathlib` dir?

BoltonBailey: I think that makes sense.

BoltonBailey:
I moved some of the definitions of poly-time computability predicates to their own `ForMathlib` file. I would even like to have the classes like `P` and `NP` themselves eventually end up in an upstream library like mathlib or cslib, so I think it makes sense to move those too.

BoltonBailey:
@Paul-Lez Now that we are on the new toolchain, the `BitstringEncoding` is a close replica of the Computability.Encoding structure in mathlib. I have therefore rewritten `BitstringEncoding` to be a thinner wrapper over `Computability.Encoding α Bool`

Paul-Lez:
Nice! I'll review this tonight!

FormalConjectures/Wikipedia/PolyTimeFunctions.lean
```diff
@@ -0,0 +1,55 @@
...
+
+-/
+
+namespace PolyTime
+
+open ComplexityTheory
+
+/-- Sanity check: the identity function is polynomial-time computable. -/
+@[category test, AMS 68]
+theorem isPolyTime_id {α : Type} [BitstringEncoding α] : IsPolyTime (id : α → α) :=
```
Paul-Lez: Nit: I think should go in ForMathlib since the definitions aren't in this file?

Paul-Lez (approved):

FormalConjectures/Wikipedia/PolyTimeFunctions.lean
BoltonBailey: I thought it would be good as a "category test" and it seemed like these tests only go in `FormalConjectures` proper.

BoltonBailey: I guess I can remove the label and put it in ForMathlib (I guess your point is that all other tests are when the definition is not in ForMathlib but FormalConjectures itself)

PR #4799, opened 2026-08-07
feat(GreensOpenProblems/21): Rado's boundedness conjecture

### LINKED ISSUE #1556: Green's Open Problems #21
### What is the conjecture
https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.21

Suppose that $a_1, ..., a_k$ are integers which do not satisfy Rado’s condition: thus if $\sum_{i \in I} a_i = 0$ then $I = \emptyset$. It then follows from Rado’s theorem that the equation $a_1 x_1 + · · · + a_k x_k$ is not partition regular. Write $c(a_1, ..., a_k)$ for the least number of colours required in order to colour $N$ so that there is no monochromatic solution to $a_1 x_1 + · · · + a_k x_k = 0$. Is $c(a_1, ..., a_k)$ bounded in terms of $k$ only?

* ams-05
* ams-11

### Choose either option

theebayuser:
Resolves #1556

## Summary
- Formalises [Ben Green's Open Problem 21](https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.21), Rado's boundedness conjecture (1933), open for all $k \geq 4$.
- Defines `RadoCondition` (some non-empty subset of the coefficients sums to zero) and `minColours`, the least number of colours needed to colour $\mathbb{N}$ with no monochromatic solution to $a_1x_1 + \cdots + a_kx_k = 0$.
- `green_21`: is `minColours a` bounded in terms of $k$ alone, over all `a : Fin k → ℤ` failing `RadoCondition`?
- Variants: `fox_kleitman` ($c(a_1,a_2,a_3) \leq 24$ for $k = 3$), `fox_kleitman_sharp` (Green's question about the sharp constant there), `fox_kleitman_modular` ([FoKl06, Conjecture 5]) and `milicevic` (the 2-adic variant, [ElJo23, Conjecture 11.1]).

Note: solutions are required to be positive, as in [FoKl06]: were $0$ admitted, $x_1 = \cdots = x_k = 0$ would be monochromatic under every colouring and no number of colours would ever suffice. The $x_i$ may repeat, matching Rado's setting. Three proved `@[category test]` lemmas check that `¬ RadoCondition` is Green's phrasing "if $\sum_{i \in I} a_i = 0$ then $I = \emptyset$", and pin down the degenerate $k = 0$ case.

## Test plan
- [x] `lake --wfail build 'FormalConjectures.GreensOpenProblems.«21»'` passes with no errors or warnings.
- [x] Statement, comments and bibliography checked against Green's "100 open problems" PDF.

### DIFF
--- a/FormalConjectures/GreensOpenProblems/21.lean
+++ b/FormalConjectures/GreensOpenProblems/21.lean
@@ -1,0 +1,202 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# Ben Green's Open Problem 21
+
+*References:*
+- [Gr24] [Green, Ben. "100 open problems." (2024).](https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.21)
+- [Ra33] Rado, Richard, *Studien zur Kombinatorik*. Math. Zeit. 36 (1933), 242-280.
+- [FoKl06] Fox, Jacob and Kleitman, Daniel, *On Rado's boundedness conjecture*. J. Combin. Theory
+  Ser. A 113 (2006), no. 1, 84-100.
+- [ElJo23] Ellis, David and Johnson, Robert (editors), *A collection of open problems in
+  celebration of Imre Leader's 60th birthday*. arXiv preprint arXiv:2310.18163 (2023).
+-/
+
+open Finset
+
+namespace Green21
+
+/--
+The coefficients $a_1, \dots, a_k$ satisfy **Rado's condition** if $\sum_{i \in I} a_i = 0$ for
+some non-empty $I \subseteq [k]$.
+
+For a single homogeneous equation this is exactly the criterion of Rado's theorem [Ra33]:
+$a_1x_1 + \cdots + a_kx_k = 0$ is partition regular if and only if the coefficients satisfy it.
+-/
+def RadoCondition {k : ℕ} (a : Fin k → ℤ) : Prop :=
+  ∃ I : Finset (Fin k), I.Nonempty ∧ ∑ i ∈ I, a i = 0
+
+/--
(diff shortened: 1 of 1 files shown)


williamjblair:
Read against Green's text; all four variants line up, and it builds clean.

The `d ≤ r` cap in `maxDepth` looks like a deviation but isn't. Green writes "the largest `d` such that `∑_{i∈I} aᵢ ≡ 0 (mod 2^d)`" with no bound, but if the subset sum is `0` in `ZMod (2^r)` every `d` qualifies and the `sSup` runs away. The cap is what makes it well defined, and the conclusion stays vacuous there since `2^(r - d) = 1`.

Nothing pins `minColours` where the `sInf` is a genuine minimum rather than `sInf ∅`:

```lean
@[category test, AMS 5 11]
theorem minColours_ones : minColours ![(1:ℤ), 1, 1] = 1 := by
  set S := {r | ∃ col : ℕ → Fin r, ∀ x : Fin 3 → ℕ, (∀ i, 0 < x i) →
    (∀ i j, col (x i) = col (x j)) → ∑ i, (![(1:ℤ), 1, 1]) i * x i ≠ 0} with hS
  have h1 : 1 ∈ S := by
    refine ⟨fun _ => 0, fun x hx _ => ?_⟩
    have := hx 0; have := hx 1; have := hx 2
    simp [Fin.sum_univ_three]
    omega
  have h0 : 0 ∉ S := by rintro ⟨col, -⟩; exact (col 0).elim0
  have : minColours ![(1:ℤ), 1, 1] = sInf S := rfl
  rw [this]
  refine le_antisymm (Nat.sInf_le h1) ?_
  rcases Nat.eq_zero_or_pos (sInf S) with h | h
  · exact absurd ((Nat.sInf_eq_zero.mp h).resolve_right (fun he => by rw [he] at h1; exact h1)) h0
  · exact h
```

`cla/google` is the only red check.

theebayuser:
thanks for catching that nothing exercised `minColours` where the infimum is a real minimum. Added `minColours_ones` fixing `minColours ![1, 1, 1] = 1`

i would add one change: `h ▸ Nat.sInf_mem ⟨1, h1⟩` rewrites right-to-left here and turns the `0 < x i` inside the set into `minColours ![1,1,1] < x i`, so it fails to typecheck. Your `Nat.sInf_eq_zero` route avoids that, and it goes through as

```lean
rw [minColours]
exact le_antisymm (Nat.sInf_le h1) (Nat.one_le_iff_ne_zero.2 fun h =>
  h0 ((Nat.sInf_eq_zero.mp h).resolve_right (Set.nonempty_iff_ne_empty.mp ⟨1, h1⟩)))
```

Also agreed on `maxDepth`: without the `d ≤ r` cap the `sSup` is unbounded once the subset sum is `0`, and the conclusion is vacuous there anyway since `2^(r - d) = 1`.

also forgot to remove claude as a co-author

mo271 (approved):
Thanks, LGTM!

PR #4820, opened 2026-08-07
ForMathlib/AdditionChain: use Mathlib's rel_getLast

williamjblair:
`List.le_getLast_of_pairwise_lt`, added in #4752, already exists in Mathlib as `List.Pairwise.rel_getLast`. It needs `(· ≤ ·)` rather than `(· < ·)`, since it asks for a reflexive relation, so the two call sites become

```lean
have hsub := (List.pairwise_append.mp hsorted).1.imp le_of_lt
have hla := hsub.rel_getLast hays
```

I noticed this while preparing the same file for Mathlib in leanprover-community/mathlib4#42510.

`lake --wfail build` is clean over the library.


mo271 (approved):
Thanks, LGTM

PR #4823, opened 2026-08-07
Arxiv/2607.05349: microscopic weightings on finite metric spaces

### LINKED ISSUE #4822: Existence of microscopic weightings for finite metric spaces
### What is the conjecture
Let $X = \{x_1, \dots, x_n\}$ be a finite metric space with distance matrix $D \in M_n(\mathbb{R})$ where $D_{ij} = d(x_i, x_j)$.

For $t > 0$, consider the similarity matrix $Z(t) \in M_n(\mathbb{R})$ with entries $Z(t)_{ij} = \exp(-t d(x_i, x_j))$. For sufficiently small $t > 0$, $Z(t)$ is invertible and defines a unique weighting vector $\vec{w}(t) = Z(t)^{-1} \mathbf{1}$, where $\mathbf{1} = (1, \dots, 1)^T$. When the limit

$$
\widehat{\vec{w}} = \lim_{t \to 0^+} \vec{w}(t)
$$

exists in $\mathbb{R}^n$, it is called the **microscopic weighting** on $X$.

For any symmetric real matrix $M$, a vector $\vec{g} \in \mathbb{R}^n$ is called a *gauging* for $M$ if $\mathbf{1}^T \vec{g} = 1$ and $M \vec{g} = c \mathbf{1}$ for some constant $c \in \mathbb{R}$, which is then called the *concentration* $\mathrm{con}(M) = c$; if no such vector exists, $\mathrm{con}(M) = \infty$.

The source paper proves that if the distance matrix $D$ is invertible, then $X$ admits a microscopic weighting if and only if $\mathrm{con}(D) \neq \infty$, and that $\mathrm{con}(D) = \infty$ always precludes the existence of a microscopic weighting.

**Conjecture (Roff–Willerton, 2026):** A finite metric space $X$ admits a microscopic weighting if and only if its distance matrix satisfies $\mathrm{con}(D) \neq \infty$.

* **Source Paper:** *The microscopic weighting on a metric space* ([arXiv:2607.05349](https://arxiv.org/abs/2607.05349), [HTML](https://arxiv.org/html/2607.05349), [PDF](https://arxiv.org/pdf/2607.05349))
* **Authors:** Emily Roff, Simon Willerton
* **In-Paper Location:** [Conjecture 3.3 (Page 10)](https://arxiv.org/pdf/2607.05349#page=10) in Section 3 "Microscopic weightings"
* **OpenConjecture ID:** 4556 ([OpenConjecture](https://github.com/davisrbr/conjectures-arxiv))

### Prerequisites needed
All foundational metric space and linear algebra concepts exist in Mathlib4:
* Metric spaces: [`Mathlib.Topology.MetricSpace.Basic`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/MetricSpace/Basic.html)
* Finite types and indices: [`Mathlib.Data.Fintype.Basic`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Fintype/Basic.html)
* Real matrices and vectors: [`Mathlib.Data.Matrix.Basic`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Matrix/Basic.html)
* Matrix invertibility and inverses: [`Mathlib.LinearAlgebra.Matrix.NonsingularInverse`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/LinearAlgebra/Matrix/NonsingularInverse.html)
* Real exponential: [`Mathlib.Analysis.SpecialFunctions.Exp`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/SpecialFunctions/Exp.html)

### Choose either option

williamjblair:
Closes #4822.

Conjecture 3.3 of [Roff-Willerton](https://arxiv.org/abs/2607.05349): a finite metric space admits a microscopic weighting iff its distance matrix has finite concentration.

```lean
answer(sorry) ↔ ∀ (X : Type) [Fintype X] [DecidableEq X] [Nonempty X] [MetricSpace X],
  HasMicroscopicWeighting X ↔ HasFiniteConcentration (distanceMatrix X)
```

Without `Nonempty` the statement is false on the empty space, and not for any mathematical reason: `∑ i, g i` is `0` there rather than `1`, so no gauging exists, while `X → ℝ` is a subsingleton so the weighting converges trivially. I proved both halves in Lean before adding the hypothesis.

`HasFiniteConcentration M` is `∃ g c, IsGauging M g c` rather than a value in `ℝ≥0∞`. The source writes `con(M) = ∞` when no gauging exists, so `con(D) ≠ ∞` is exactly this existence, and phrasing it this way avoids having to decide what `con` should be when two gaugings give different constants.

The two results the paper proves are here as `research solved`: infinite concentration rules out a weighting, and the equivalence holds when `D` is invertible. The first is already one direction of the conjecture, so what is open is the converse.


FormalConjectures/Arxiv/2607.05349/MicroscopicWeighting.lean
```diff
@@ -0,0 +1,103 @@
...
+vector `c`. -/
+def IsGauging (M : Matrix X X ℝ) (g : X → ℝ) (c : ℝ) : Prop :=
+  ∑ i, g i = 1 ∧ M *ᵥ g = Function.const X c
+
+/-- `M` has finite concentration, written $\mathrm{con}(M) \neq \infty$ in the source: some
+gauging for `M` exists.
+
+The source defines $\mathrm{con}(M)$ as the concentration of a gauging and sets it to $\infty$
+when there is none, so `con(M) ≠ ∞` is exactly this. Phrasing it as existence avoids having to
+say which value $\mathrm{con}$ takes when different gaugings give different constants. -/
```
mo271: But the paper's Theorem 2.7 proves that if two gaugings exist, they have the same concentration ($\operatorname{con}(A)$ is well-defined), if I understnad correctly? So different gaugings cannot give different constants. The docstring should say something like "avoids defining the actual concentration value" rather than implying it might be ambiguous, which can't really happen (but might be work to prove? or would that be a good `textbook` theorem?

FormalConjectures/Arxiv/2607.05349/MicroscopicWeighting.lean
```diff
@@ -0,0 +1,103 @@
...
+-/
+@[category research open, AMS 15 51]
+theorem microscopic_weighting_iff_finite_concentration :
+    answer(sorry) ↔ ∀ (X : Type) [Fintype X] [DecidableEq X] [Nonempty X] [MetricSpace X],
+      HasMicroscopicWeighting X ↔ HasFiniteConcentration (distanceMatrix X) := by
+  sorry
+
+/--
+One direction is known: infinite concentration rules a microscopic weighting out. Together with
+the conjecture above this leaves the converse, that finite concentration is enough.
```
mo271:
```suggestion
One direction is known: a microscopic weighting implies finite concentration (Theorem 3.1(3)).
Together with the conjecture above this leaves the converse, that finite concentration is enough.
```
Avoiding the contrapositive and giving concrete Theorem name

mo271 (approved):
Thanks, Looks great mathematically. 

More nits about the docstring: 
Add paper theorem numbers (Thm 3.1, Thm 3.8, Conj 3.3) to the docstrings for traceability

I think this is good general practice in the Arxiv subdir, we should add it to the Arxiv/README.md (when created)...

williamjblair:
Added Conjecture 3.3, Theorem 3.1(3) and Theorem 3.8 to the docstrings, and took your line-88 suggestion as written.

You are right about the concentration, and I had it backwards. The uniqueness is in the source immediately after Definition 2.3, as `c = c1ᵀv' = (vᵀAᵀ)v' = vᵀ(c'1) = c'`, and Definition 2.4 rests on it. Theorem 2.7 is the `mag(A) = 1/con(A)` relation rather than the uniqueness itself.

Short enough to prove, so I took the `textbook` suggestion:

```lean
@[category textbook, AMS 15]
theorem concentration_unique {M : Matrix X X ℝ} (hM : M.IsSymm) {g g' : X → ℝ} {c c' : ℝ}
    (h : IsGauging M g c) (h' : IsGauging M g' c') : c = c'
```

`#print axioms` gives the standard three. The `HasFiniteConcentration` docstring now says the existence phrasing keeps the value of `con` out of the statement and points at that theorem, rather than implying the value could be ambiguous.

`IsGauging` still does not require `M` symmetric, while Definition 2.3 does. `concentration_unique` takes the hypothesis where it is needed, so the generality costs nothing, but say the word if you would rather the definition track the paper.

mo271 (approved):

PR #4824, opened 2026-08-07
Arxiv/2607.05739: integer values of tan(arctan 1 + ... + arctan n)

### LINKED ISSUE #4812: Amdeberhan–Medina–Moll conjecture on integer values of tan(arctan 1 + ... + arctan n)
### What is the conjecture
For each integer $n \ge 1$, let

$$
x_n = \tan\big( \sum_{k=1}^n \arctan k \big).
$$

Equivalently, $x_n = B_n / A_n \in \mathbb{Q}$ where $Z_n = \prod_{k=1}^n (1 + i k) = A_n + i B_n \in \mathbb{Z}[i]$ in the Gaussian integers, with recurrence

$$
x_n = \frac{x_{n-1} + n}{1 - n x_{n-1}} \quad \text{for } n \ge 2, \quad x_1 = 1.
$$

The first four terms are integers: $x_1 = 1$, $x_2 = -3$, $x_3 = 0$, $x_4 = 4$.

**Conjecture (Amdeberhan–Medina–Moll, 2008):** For every integer $n \ge 5$, $x_n$ is not an integer ($x_n \notin \mathbb{Z}$).

* **Source Paper:** *Integer values of $\tan(\arctan 1+\arctan 2+\cdots+\arctan n)$ are rare* ([arXiv:2607.05739](https://arxiv.org/abs/2607.05739), [HTML](https://arxiv.org/html/2607.05739), [PDF](https://arxiv.org/pdf/2607.05739))
* **Authors:** Ken Ono
* **Original Reference:** T. Amdeberhan, L. A. Medina, V. H. Moll (2008), *Arithmetical properties of a sequence of integers related to the arctangent function*, The Ramanujan Journal 16(2), 225–230.
* **In-Paper Location:** [Conjecture (Page 1)](https://arxiv.org/pdf/2607.05739#page=1) in Section 1 "Introduction"
* **OpenConjecture ID:** 4539 ([OpenConjecture](https://github.com/davisrbr/conjectures-arxiv))

### Prerequisites needed
None. Standard rational arithmetic, Gaussian integers `GaussianInt`, or recurrence relations on $\mathbb{Q}$.

*Note*: https://github.com/AxiomMath/TanArctan contains a proof of a weaker version of the conjecture (which the linked paper is about), so it should be easy to extract the conjecture from there. 

### Choose either option

williamjblair:
Closes #4812.

Amdeberhan-Medina-Moll: for $n \geq 5$, $\tan(\arctan 1 + \cdots + \arctan n)$ is not an integer.

The sequence is defined through $Z_n = \prod_{k=1}^n (1 + ik)$ in `GaussianInt`, so $x_n = B_n/A_n$ is exact rather than a real computation, and the four known integer values come out by `decide`.

"$x_n$ is an integer" is `A n ∣ B n` rather than `∃ m : ℤ, x n = m`:

```lean
def IsIntegerValue (n : ℕ) : Prop := A n ∣ B n
```

If $A_n$ were ever $0$ then $x_n$ is a pole of the tangent, not an integer, but `(B n : ℚ) / 0` is `0` in Lean and the `∃ m` form would count it as one. The two agree whenever $A_n \neq 0$, which I checked holds for every $n \leq 3000$. Same computation says the only integer values in that range are $n = 1, 2, 3, 4$, matching the paper.

Three of the statements are proved rather than `sorry`: the four integer values, $x_1 = 1, x_2 = -3, x_3 = 0, x_4 = 4$, and that the conjecture holds on $5 \leq n \leq 10$. `#print axioms` on each gives only `propext, Classical.choice, Quot.sound`.

The tangent addition recurrence and the identification with $\tan(\sum \arctan k)$ are `research solved` and left for someone who wants them.


mo271:
before I do a more thorough review, a quick question: 
Is this aligned with 
https://github.com/AxiomMath/TanArctan? 
we should definitely mention this repo as reference, but also perhaps good to just take their formalisation if not already done so!

williamjblair:
Hadn't seen it, thanks. Turns out their `P`, `A`, `B` and `x` are the same definitions I'd written, down to `∏ k ∈ Finset.Icc 1 n, (⟨1, k⟩ : GaussianInt)`. They ran into the pole too, carrying `A n ≠ 0` as a hypothesis where I use `A n ∣ B n`.

They prove Ono's three results rather than the conjecture, and the proofs are done. The two `sorry` hits in `solution.lean` are both inside a comment sketching a plan.

Taken all three, pinned at `5382d3c`:

| | |
|---|---|
| `kernel_dvd_of_eq_intCast` | `x n = m` gives `K n ∣ 1 + m²`, and `|m| ≥ √(K n - 1)` |
| `exists_near_half_pi` | exceptional indices land within `2/n` of a multiple of `π/2` |
| `exceptional_ncard_le` | `#(E ∩ [1,N]) = O(log N)` |

The last is the paper's theorem, so the conjecture now sits next to what's known. Repo's in the references.

FormalConjectures/Arxiv/2607.05739/TanArctanSum.lean
```diff
@@ -0,0 +1,139 @@
...
+/-!
+# Integer values of $\tan(\arctan 1 + \arctan 2 + \cdots + \arctan n)$
+
+*References:*
+- [arxiv/2607.05739](https://arxiv.org/abs/2607.05739)
+  **Integer values of $\tan(\arctan 1+\arctan 2+\cdots+\arctan n)$ are rare** by *Ken Ono*
+- [AMM08] T. Amdeberhan, L. A. Medina, V. H. Moll, *Arithmetical properties of a sequence of
+  integers related to the arctangent function*. The Ramanujan Journal 16(2) (2008), 225-230.
+- [TanArctan](https://github.com/AxiomMath/TanArctan), a Lean formalisation of the three
+  results of [Ono26], MIT licensed. Its , ,  and  are the definitions used here.
```
mo271: "Its , , " what? perhaps you meant to put some lemma/theorem names or something?

FormalConjectures/Arxiv/2607.05739/TanArctanSum.lean
```diff
@@ -0,0 +1,139 @@
...
+import FormalConjecturesUtil
+
+/-!
+# Integer values of $\tan(\arctan 1 + \arctan 2 + \cdots + \arctan n)$
+
+*References:*
+- [arxiv/2607.05739](https://arxiv.org/abs/2607.05739)
+  **Integer values of $\tan(\arctan 1+\arctan 2+\cdots+\arctan n)$ are rare** by *Ken Ono*
+- [AMM08] T. Amdeberhan, L. A. Medina, V. H. Moll, *Arithmetical properties of a sequence of
+  integers related to the arctangent function*. The Ramanujan Journal 16(2) (2008), 225-230.
```
mo271:
```suggestion
- [AMM08] T. Amdeberhan, L. A. Medina, and V. H. Moll, Arithmetical properties of a sequence arising from an arctangent sum, J. Number Theory 128 (2008), no. 6, 1807–1846.
```

my bad: in the issue I gave a different paper! This one is from the arxiv and shoud be the one in question...

FormalConjectures/Arxiv/2607.05739/TanArctanSum.lean
```diff
@@ -0,0 +1,139 @@
...
+/-- The exceptional set $E = \{n \geq 5 : |x_n| > n/2 + 1\}$. An index with $A_n = 0$ is a
+pole of the tangent rather than a large value, and is counted in, reading $|x_n|$ as infinite. -/
+def exceptional : Set ℕ := {n | 5 ≤ n ∧ (A n = 0 ∨ ((n : ℚ) / 2 + 1 < |x n|))}
+
+/--
+An integer value is divisible in a way that forces it to be large: if $x_n = m$ then
+$K_n \mid 1 + m^2$, and $|m| \geq \sqrt{K_n - 1}$ once $K_n > 1$.
+-/
+@[category research solved, AMS 11, formal_proof using lean4 at
+  "https://github.com/AxiomMath/TanArctan/blob/5382d3c20ee3f30e2cbd84362eb07a7e93250348/output/solution.lean"]
```
mo271: can we get line numbers in those permalinks, especially since we link to multiple lemmas here and below

FormalConjectures/Arxiv/2607.05739/TanArctanSum.lean
```diff
@@ -0,0 +1,139 @@
...
+is not an integer.
+-/
+@[category research open, AMS 11]
+theorem tan_arctan_sum_not_integer :
+    answer(sorry) ↔ ∀ n : ℕ, 5 ≤ n → ¬ IsIntegerValue n := by
+  sorry
+
+/-- $x_n$ satisfies $x_1 = 1$ and $x_n = \dfrac{x_{n-1} + n}{1 - n x_{n-1}}$, which is the
+tangent addition formula. -/
+@[category research solved, AMS 11]
```
mo271:
```suggestion
@[category textbook, AMS 11]
```
or am I mistaken and these are difficult? 
same for `x_eq_tan_sum_arctan` below

mo271 (commented):
Thanks for picking up this issue so promptly!
Looks great, just a few comments

FormalConjectures/Arxiv/2607.05739/TanArctanSum.lean
williamjblair: Oops shell changed the backticks when I wrote that line. Fixed, it should read "Its `P`, `A`, `B` and `x`".

FormalConjectures/Arxiv/2607.05739/TanArctanSum.lean
williamjblair: Applied, thanks. I had copied the citation straight from the issue.

FormalConjectures/Arxiv/2607.05739/TanArctanSum.lean
williamjblair: Done: `#L176`, `#L453` and `#L975` for the three.

FormalConjectures/Arxiv/2607.05739/TanArctanSum.lean
williamjblair: You're right, both are standard: one is just the tangent addition formula, the other is what the sequence is defined to be. Changed to `textbook`.

mo271:
Thanks, LGTM

perhaps we can also get the opinion of @jjaassoonn?

@jjaassoonn great work on formalising the variants of the theorem you could prove!
can you have a look at this pull request, i.e. the formalisation of the remaining conjecture?

mo271 (approved):

mo271:
merging, when @jjaassoonn  has more comments, we can amend the file...

PR #4826, opened 2026-08-07
Close erdos-status-sync issues once the mismatch is gone

### LINKED ISSUE #4051: Erdős Problem 678: status mismatch (repo=solved, erdosproblems.com=formally solved)
The status of [Erdős problem 678](https://www.erdosproblems.com/678) appears to have changed.

- **[This repo](http://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/678.lean)**: `solved` (in `FormalConjectures/ErdosProblems/678.lean`)
- **[erdosproblems.com/678](https://www.erdosproblems.com/678)**: `formally solved`

Please verify and update the `@[category research ...]` annotation if appropriate.

williamjblair:
`check_erdos_status.py` opens an issue when this repository and erdosproblems.com disagree, but nothing ever closed one, so the label collects issues describing a state that no longer holds. 35 are open and 6 of them have nothing left to act on.

The run already computes everything needed: whatever is not in `mismatches` has been overtaken. Against live data it closes #4051, #4528, #4530, #4532, #4535 and #4699, which includes the four @Jayyhk pointed out on #4763.

The obvious version gets two things wrong, and I hit both.

`find_mismatches` keys problems by string. Comparing `int(...)` against that set matches nothing, so the first version I ran would have closed 34 of the 35. There's a test pinning that.

"Not in `mismatches`" also covers problems whose YAML state the script cannot read; `open (Lean)` is one it warns about today, and those are skipped before the comparison. Those must not count as resolved, so `classifiable()` gives the problems we could read and the closer leaves the rest alone. Nothing is held back by this today, but it is the difference between "they agree" and "we could not tell".

The selection is `issues_to_close`, kept apart from the `gh` calls so it has tests. `python3 -m unittest discover -s scripts` passes, 23 tests.


mo271 (approved):
Is this taken into account the problem with the issue where at erdosproblems.com the problem is marked as "solved (lean)" but the lean proof is conditional? 
I think currently we add manually "won't fix" or something? I guess those could also be closed once the coditional proof gets replaced with a real one
Otherwise looks good to me!

williamjblair:
Good question, they stay open. When our proof is conditional we keep the problem open, erdosproblems says solved, and that's a genuine disagreement, so it stays in the mismatch set. The PR only closes issues whose mismatch has gone away.

Which I think is the right default here, since the two really do disagree and someone should look. But you're right that it leaves the label collecting things nobody will act on.

To close them automatically the script would need to spot the conditional proof itself. Right now `FORMAL_PROOF_ATTR` matches plain and conditional the same way, so it can't. Since it already reads the `.lean` files, picking up the `assuming` clause is a small change, and then "erdosproblems solved, ours conditional" could be its own case that closes when the conditions go. Happy to add that in a follow-up.

Jayyhk:
The case where the Lean proof is conditional is not detected in this PR. It seems to me the script only checks if a `formal_proof` link is present, so a conditional proof with a `formal_proof` link looks the same as an unconditional one.

Also, at least one conditional linkage has had its issue closed. For problem 427, #3846 closed #3841 with a `formal_proof` link to a proof that relies on a custom axiom. Would recording these as `conditional ... assuming` be the way to handle them as in #4368?

williamjblair:
Good point about 427. The link there is a plain `formal_proof`, and the gist axiomatizes Shiu's theorem. No file in the repo uses `conditional` yet, so there is nothing for the script to detect today.

The fix therefore starts in the file. 427 becomes `conditional ... assuming erdos_427.variants.shiu`, and that statement is already there. Then the script stops counting `conditional` as formally solved. The two changes must land together: the annotation alone changes nothing, because the regex matches both forms, and only the pair makes 427 mismatch again.

After checking more broadly, 750 and 1141 also link to sorry-free proofs that assume a published theorem as an axiom, and 13 links do not name a file at all. I opened #4881 with the detail and will do the work there.

This PR only closes issues whose mismatch is gone, so it does not change what counts as solved.

PR #180, opened 2025-06-06
Add Erdos 299

callesonne:
This closes #50.

PR #146, opened 2025-06-03
Wolstenholme prime

seewoo5:
Definition(s) of Wolstenholme prime and statement of the conjecture that there are infinitely many of them.

PR #4827, opened 2026-08-07
check_erdos_status: see wrapped category attributes, and 'open (Lean)'

williamjblair:
Depends on #4826, which adds the test file this extends.

`check_erdos_status.py` is silently blind to two things.

**Wrapped attributes.** The scanner matches `@\[category research (open|solved).*\]` with `MULTILINE` but not `DOTALL`, so `.*` stops at the newline and the closing `]` has to be on the same line. 25 of the 604 problem files wrap theirs, usually to fit a `formal_proof` URL:

```lean
@[category research solved, AMS 5, formal_proof using lean4 at
  "https://github.com/..."]
theorem erdos_183 :
```

Those problems were never compared against anything. Using `[^\]]*` instead crosses newlines and still stops at the attribute's own `]`, so it cannot run into the next declaration. It finds 70 more declarations and loses none.

**`open (Lean)`.** Four problems carry it upstream and it was in none of the state sets, so `yaml_status_to_category` returned `None` and they were skipped. It is the open counterpart of `solved (Lean)`: the problem is open and a Lean statement of it exists.

Together these take the live mismatch count from 28 to 34. Three of the six are #146, #180 and #183, which erdosproblems.com still lists as open and which my own #4715 marked `formally solved` on the strength of the openai/ten-proofs links. I have not touched those files here; the point of this change is that the check can now see the disagreement and say so, and someone should decide which side is right.

`python3 -m unittest discover -s scripts` passes, 31 tests, including one pinning the wrapped-attribute case.

### DIFF
--- a/scripts/check_erdos_status.py
+++ b/scripts/check_erdos_status.py
@@ -31,8 +31,11 @@ ERDOS_DIR = os.path.join(
 # Matches the category annotation line immediately before any erdos theorem.
 # Captures the category and the problem number.
 # Note: 'formally solved' is no longer a valid category value.
+# `[^\]]*` rather than `.*`: a character class crosses newlines, so an attribute wrapped
+# over two lines is still matched, and it stops at the attribute's own `]` rather than
+# running into the next declaration. 25 problem files wrap theirs and were invisible here.
 CATEGORY_THEN_THEOREM = re.compile(
-    r"@\[category research (open|solved).*\]\s*\n"
+    r"@\[category research (open|solved)[^\]]*\]\s*\n"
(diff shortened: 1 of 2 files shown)


williamjblair:
Closing in favour of removing the scanner rather than patching it.

Even with this fix the regex sees 581 of the 604 Erdős problems, because it needs a theorem named exactly `erdos_<n>`. `site/data/conjectures.json`, which the site already publishes, has all 604 built from the Lean environment by `extract_names`, along with 2947 problems from the seven other families. Reading that is both correct and shorter than the pattern I was fixing.

Replacement on the way.

PR #4797, opened 2026-08-07
feat(ErdosProblems): Add lean proof links to 650, 692, 751, 785, 844, 1047

danielchin:
Closes #4728
Closes #4727
Closes #4726 
Closes #4713 
Closes #4711 
Closes #4710

PR #4829, opened 2026-08-07
Arxiv/2607.08366: minimum modulus for the unique multiset-sum problem

### LINKED ISSUE #4811: Minimum modulus for the unique multiset-sum problem
### What is the conjecture
Fix $n \ge 2$. A subset $A = \{a_0, a_1, \dots, a_{n-1}\} \subset \mathbb{Z}/N\mathbb{Z}$ of size $n$ is called *valid mod* $N$ if the all-ones multiset is the unique size-$n$ multiset drawn from $A$ whose sum modulo $N$ equals $p := \sum_{i=0}^{n-1} a_i$; that is, for any nonnegative integers $m_0, \dots, m_{n-1} \in \mathbb{N}$ satisfying $\sum_{i=0}^{n-1} m_i = n$:

$$
\sum_{i=0}^{n-1} m_i a_i \equiv \sum_{i=0}^{n-1} a_i \pmod N \implies m_0 = m_1 = \dots = m_{n-1} = 1.
$$

**Conjecture:** For every $n \ge 2$ and every $N < 2^n - 2^{\lfloor \log_2 n \rfloor}$, no set $A \subset \mathbb{Z}/N\mathbb{Z}$ of $n$ residues is valid mod $N$.

In other words, the super-increasing set $A = \{2^k - 1 : 0 \le k \le n-1\}$ attains the strictly least valid modulus $N_{\min}(n) = 2^n - 2^{\lfloor \log_2 n \rfloor}$ over all size - $n$ residue sets.

* **Source Paper:** *Minimum modulus for the unique multiset-sum problem* ([arXiv:2607.08366](https://arxiv.org/abs/2607.08366), [HTML](https://arxiv.org/html/2607.08366), [PDF](https://arxiv.org/pdf/2607.08366))
* **Authors:** José A. R. Fonollosa
* **In-Paper Location:** [Conjecture 1 (Page 2)](https://arxiv.org/pdf/2607.08366#page=2) in Section 1 "Introduction"
* **OpenConjecture ID:** 4435 ([OpenConjecture](https://github.com/davisrbr/conjectures-arxiv))

### Prerequisites needed
None. All required foundations (finite modular arithmetic $\mathbb{Z}/N\mathbb{Z} = \mathrm{ZMod\ } N$, `Finset`, sums, and base-2 integer logarithm `Nat.log 2 n`) are available in Mathlib4.

*Note*: the paper already links to a mathlib formalisation!

A Lean 4 statement sketch:
```lean
import Mathlib

/-- A subset `A : Finset (ZMod N)` of size `n` is valid mod `N` if the only multiset of size `n`
drawn from `A` whose sum mod `N` matches `∑ a ∈ A, a` is `A` itself (all multiplicities equal 1). -/
def IsValidMod (N : ℕ) (A : Finset (ZMod N)) (n : ℕ) : Prop :=
  A.card = n ∧
  ∀ (m : ZMod N → ℕ),
    (∑ a ∈ A, m a = n) →
    (∀ a ∉ A, m a = 0) →
    (∑ a ∈ A, (m a : ZMod N) * a = ∑ a ∈ A, a) →
    ∀ a ∈ A, m a = 1)

/-- Global Minimum Modulus Conjecture for the unique multiset-sum problem (Fonollosa, 2026).
No subset of `ZMod N` of size `n` is valid mod `N` for `N < 2^n - 2^(Nat.log 2 n)`. -/
theorem min_modulus_conjecture (n : ℕ) (hn : 2 ≤ n) (N : ℕ)
    (hN : N < 2^n - 2^(Nat.log 2 n))
    (A : Finset (ZMod N)) :
    ¬ IsValidMod N A n := by
  sorry
```

### Choose either option

williamjblair:
Closes #4811.

Conjecture 1 of [Fonollosa](https://arxiv.org/abs/2607.08366): for `n ≥ 2` and `N` below `2^n - 2^⌊log₂ n⌋`, no set of `n` residues mod `N` is valid, so the super-increasing set attains the least valid modulus.

```lean
answer(sorry) ↔ ∀ n N : ℕ, 2 ≤ n → 0 < N → N < minModulus n →
  ∀ A : Finset (ZMod N), #A = n → ¬ IsValidMod A
```

`minModulus 2 = 2`, so the range includes `N = 0`, where `ZMod 0` is `ℤ` rather than a finite modulus. `{1, 2} ⊆ ℤ` has two elements and the only pair drawn from it summing to `3` is one of each, so it is valid and the statement collapses at the very first case. Hence `0 < N`, which I checked in Lean rather than assumed.

`IsValidMod` follows the sketch in the issue but drops its `A.card = n` clause, which duplicated the `#A = n` already in the statement, and its `∀ a ∉ A, m a = 0`, which is unnecessary since every sum is already taken over `A`.

Two `API` statements are proved rather than `sorry`: that the all-ones multiset does have the right size and sum, so this is a uniqueness claim and not an existence one, and that a set with at most one element is valid trivially, which is why the conjecture starts at `n = 2`. `#print axioms` on both gives only `propext, Classical.choice, Quot.sound`.

The attainability half is stated as `research solved`, since the paper proves the super-increasing set works.

### Formalisation notes

Moved here from the Lean file, per AGENTS.md.

**Why `0 < N`.** `minModulus 2 = 2`, so the range `N < minModulus n` includes `N = 0`, and `ZMod 0` is `ℤ` rather than a finite modulus. `{1, 2} ⊆ ℤ` has two elements and the only pair drawn from it summing to `3` is one of each, so it is valid and the statement would be false there for a reason unrelated to the question. `N = 1` needs no guard: `ZMod 1` is trivial, so `#A = n ≥ 2` is already impossible.

**Why Conjecture 1 is `research open` while Theorem A is `research solved`.** The Main Theorem fixes the super-increasing set and pins the least modulus at which *that set* is valid; `jarfo/min-modulus` proves it, kernel-checked and sorry-free. Conjecture 1 is the statement that no other size-`n` set does better, and has CP-SAT infeasibility certificates per modulus for `n ≤ 7` rather than a proof.

**The link on `exists_isValidMod_minModulus`.** `theoremA` works in `ℕ` with `Nat.ModEq` and indexes the multiset by position in `[0, n)`, where this file indexes by residue in `ZMod N`. The translation is sound at the bound: the residues `2^k - 1` for `k < n` are distinct and below `2^n - 2^⌊log₂ n⌋`, so nothing collapses when they are collected into a `Finset`. I checked that for `n` up to 199.


mo271:
Thanks! Before I do a more in-depth review: 
Also here there is already a formalisation of the parts that could be proved: 
https://github.com/jarfo/min-modulus
beatifully done by @jarfo !

Perhaps we could compare what they do and adapt the pull request here if needed?
Of course any feedback from @jarfo is also most welcome!

williamjblair:
They prove different things.

@jarfo's development covers the **Main Theorem**: Theorems A and B fix the super-increasing set `{2^k - 1}` and pin the least modulus at which *that set* is valid, giving `nmin_eq : IsLeast {N | 2 ≤ N ∧ Valid n N} (2^n - 2^⌊log₂ n⌋)`. Repo CI is green on `master`, and all four files are sorry-free and axiom-free.

**Conjecture 1** is the universal statement: no set of `n` residues is valid below that bound. Section 7 backs it with one CP-SAT infeasibility certificate per modulus rather than a proof, so `research open` is right for what this file states.

The repo is now in the references, with the correspondence recorded for Theorem A, which matches `exists_isValidMod_minModulus` here.

Theorem B I left out. `Valid n N` there indexes the multiset by position in `[0, n)`; `IsValidMod` here indexes by residue, so it needs the `n` residues to stay distinct mod `N`. At the bound they do, which is why Theorem A translates. Below the bound they often do not:

| `n` | moduli below the bound where `{2^k - 1}` collapses |
|---|---|
| 5 | 10 of 26 |
| 6 | 16 of 58 |
| 8 | 35 of 246 |

At `n = 3, N = 2` the set is `{0, 1, 1}`, two residues rather than three. A `Finset`-based Theorem B would be a different statement from @jarfo's. A faithful version needs an indexed formulation alongside the set-based one, which is more than this PR.

I have left `exists_isValidMod_minModulus` without a `formal_proof` link. `theoremA` proves the same result, but a link there would say the proof establishes this statement when what it establishes is a translation of it, and that is the thing I complained about on #4797. The docstring carries the correspondence instead, including why the translation is sound at the bound.

jarfo:
Author of the paper here — thanks for picking this up, and for catching the `ZMod 0` degeneracy. `0 < N` is exactly the right guard: without it the statement is false at n = 2, N = 0, A = {1, 2} ⊂ ℤ (p = 3, and 1+1 = 2, 2+2 = 4, so the all-ones multiset is the only size-2 multiset summing to 3). Worth noting N = 1 needs no guard — `ZMod 1` is trivial, so `#A = n ≥ 2` is already impossible there.

The statement reads faithfully to Conjecture 1, including dropping the sketch's `∀ a ∉ A, m a = 0` clause (sound: `m` off `A` enters neither sum nor the conclusion). A few smaller points:

1. `exists_isValidMod_minModulus` is `research solved` and does have a Lean proof, so it could carry the `formal_proof` attribute:

   ```lean
   @[category research solved, AMS 5 11,
     formal_proof using lean4 at "https://github.com/jarfo/min-modulus"]
   ```

   The relevant declaration is `theoremA` in `MinModulus/UniqueSums.lean` (`Valid n (2 ^ n - 2 ^ Nat.log 2 n)` for `2 ≤ n`), sorry-free and kernel-checked.

2. Its docstring says "the bound is attained, so no smaller modulus can be claimed for `n` residues in general" — attainment only bounds the least valid modulus from *above*; ruling out smaller moduli is precisely the open half.

3. `AGENTS.md` asks for formalisation notes and caveats to live in the PR description rather than the Lean file, so the paragraphs on the CP-SAT certificates, on `0 < N` being "not decoration", and on the soundness of the index-to-residue translation may want to move there (the `0 < N` counterexample is worth keeping somewhere — it is the reason the guard is there).

4. The issue is labelled ams-05; `@[AMS 5 11]` covers both.

5. Minor, on the issue text in #4811: "the super-increasing set … is *conjectured* to achieve the minimal valid modulus" — for that set it is proved, N_min(n) = 2^n − 2^⌊log₂ n⌋ for all n ≥ 2 (Theorems A and B, kernel-checked as `nmin_eq`). Only minimality across all size-n sets is conjectural, as the PR docstring correctly says.

For calibration on the open half: exhaustive search over all size-n residue sets gives least valid modulus 6, 12, 28 at n = 3, 4, 5, matching the formula, and the CP-SAT infeasibility certificates cover n ≤ 7.

williamjblair:
Thanks so much for the paper + your feedback. All applied in 225c54f.

1. Link added, pinned to a commit and to `#L837-L838`.
2. Fixed. It said attainment rules out smaller moduli; it bounds from above, and `min_modulus` is the open half.
3. Moved to the PR description. Kept one line for why `0 < N` is there, and added your `N = 1` point.
4. Already `AMS 5 11`.
5. Can't edit #4811, it's @mo271's. Putting it here instead: least-for-the-super-increasing-set is proved (`nmin_eq`), only optimality across all size-`n` sets is conjectural.

I'd argued against the link on encoding grounds. You wrote both, so I'll take your reading.

mo271 (approved):
Thanks, LGTM!

PR #4830, opened 2026-08-07
ErdosProblems/80: books in dense graphs where every edge is in a triangle

### LINKED ISSUE #309: Erdős Problem 80

williamjblair:
Closes #309.

Erdős 80: with $f_c(n)$ the largest book forced on every $n$-vertex graph with at least $cn^2$ edges whose every edge lies in a triangle, estimate $f_c(n)$; in particular is $f_c(n) > n^\epsilon$, or $f_c(n) \gg \log n$?

A book is `#(G.trianglesContaining e)`, using the `trianglesContaining` already in `ForMathlib/Combinatorics/SimpleGraph/Clique.lean`, so nothing new was needed for it:

```lean
noncomputable def bookNumber : ℕ :=
  G.edgeFinset.sup fun e => #(G.trianglesContaining e)
```

`f c n` is `sInf` of the book numbers over admissible graphs, which is what "the maximal `m` that every such graph must contain" comes to. Both questions in the problem are stated, the $n^\epsilon$ one as the headline and $\gg \log n$ as a variant, since the source asks both.

`sInf ∅ = 0`, so `f c n = 0` both when some admissible graph has no book and when no graph on `n` vertices has $cn^2$ edges at all. The second is the honest reading, nothing is forced when nothing qualifies, and it only bites for `c` large enough that the class is empty, which the asymptotic questions do not reach.

Three statements are proved rather than `sorry`: that `bookNumber` vanishes exactly when no edge lies in a triangle, and the two `⊥` cases. `#print axioms` on each gives only `propext, Classical.choice, Quot.sound`.

### DIFF
--- a/FormalConjectures/ErdosProblems/80.lean
+++ b/FormalConjectures/ErdosProblems/80.lean
@@ -19,7 +19,16 @@ import FormalConjecturesUtil
 /-!
 # Erdős Problem 80
 
-*Reference:* [erdosproblems.com/80](https://www.erdosproblems.com/80)
+*References:*
+- [erdosproblems.com/80](https://www.erdosproblems.com/80)
+- [erdosproblems.com/600](https://www.erdosproblems.com/600), stated in
+  `FormalConjectures/ErdosProblems/600.lean`
+
+600 asks the same question from the other side. `Erdos600.eFunction n r` is the least edge
+count forcing some edge into `r` triangles; `f c n` here is the largest book forced once the
+edge count is at least $cn^2$. So `r ≤ f c n` and `Erdos600.eFunction n r ≤ c * n^2` say the
+same thing, and the two functions are inverse to each other in that sense. Both are built on
+`SimpleGraph.trianglesContaining`, which 600 introduced.
 -/
 
 open Filter Finset SimpleGraph
@@ -70,8 +79,8 @@ The weaker question from the same problem: is $f_c(n) \gg \log n$?
 -/
 @[category research open, AMS 5]
 theorem erdos_80.variants.log :
-    answer(sorry) ↔ ∀ c : ℝ, 0 < c → ∃ C > (0 : ℝ), ∀ᶠ n : ℕ in atTop,
-      C * Real.log n ≤ f c n := by
+    answer(sorry) ↔ ∀ c : ℝ, 0 < c →
+      (fun n : ℕ ↦ (f c n : ℝ)) ≫ (fun n : ℕ ↦ Real.log n) := by
   sorry
 
 omit [DecidableEq α] in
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/80.lean
```diff
@@ -0,0 +1,96 @@
...
+$f_c(n)$. In particular, is it true that $f_c(n)>n^\epsilon$ for some $\epsilon>0$?
+-/
+@[category research open, AMS 5]
+theorem erdos_80 :
+    answer(sorry) ↔ ∀ c : ℝ, 0 < c → ∃ ε > (0 : ℝ), ∀ᶠ n : ℕ in atTop,
+      (n : ℝ) ^ ε < f c n := by
+  sorry
+
+/--
+The weaker question from the same problem: is $f_c(n) \gg \log n$?
```
mo271, line 78: we have \gg notation in our repo, would be nice to use here?!

mo271 (commented):
Thanks, looks good generally!

Can you add problem 600 to the reference and have a look at the formalisation there (if they roughly agree and if they could use common concepts)

williamjblair:
Both done.

`≫` now used for the log variant. It replaces `∃ C > 0, ∀ᶠ n, C * log n ≤ f c n`, and I proved the two forms equivalent in Lean before swapping, since it changes the statement of an open problem.

On 600: same question from the other side, and the concepts already line up. `Erdos600.eFunction n r` is the least edge count forcing some edge into `r` triangles; `f c n` here is the largest book forced once the edge count reaches `cn²`. So `r ≤ f c n` and `eFunction n r ≤ c * n²` say the same thing, and the two are inverse in that sense. Both are built on `SimpleGraph.trianglesContaining`, which 600 introduced, so the shared concept is already there.

That is in the module docstring now, with the reference. I have left the inversion unstated as a theorem. `eFunction` is an `sInf` over edge counts and `f` an `sInf` over book numbers, both with junk `0` on empty sets, and the edge threshold is real-valued here against `ℕ` in 600, so the side conditions would end up longer than the cross-reference you asked for and would put new API into 600's file. Happy to do it as its own PR if you want it.

mo271 (approved):
Thanks looks good to me!

PR #4832, opened 2026-08-07
site: search the statements the browse page shows

### LINKED ISSUE #4831: Browse search matches theorem names only, not the statements it displays
The browse page renders each statement in prose, but the search only matches theorem names, so the text on screen is not findable.

On https://google-deepmind.github.io/formal-conjectures/browse:

| query | results |
|---|---:|
| `coprime` | 4, all with `coprime` in the theorem name |
| `coprime positive integers` | 0 |
| `distinct distances` | 0 |
| `chromatic` | 60, all name matches |

"For every positive real number ε, there exist only finitely many triples (a, b, c) of coprime positive integers" is the first statement on the page, and no phrase from it finds it.

That matters most for the people the site is for. Someone who knows the mathematics but not our naming reaches for "distinct distances" or "sphere packing", not `Erdos307.erdos_307.variants.coprime`. The repository has plenty on distinct distances; the search returns nothing for it.

Correcting myself on the mechanism: they are not in `conjectures.json`, which `build.js` deliberately strips of `statement` and `docstring` to keep the payload down. They come from `verso-fragments.json`, which the browse page already loads and already renders through `FC.problemDocHTML`. So the text is in the browser either way and no extra fetch is needed.

#4832 does it.

williamjblair:
Closes #4831.

The browse page shows each statement but the search only matched theorem names, so the text on screen was not findable. Measured on a local build of the site, before and after:

| query | before | after |
|---|---:|---:|
| `coprime positive integers` | 0 | 3 |
| `distinct distances` | 0 | 13 |
| `chromatic` | 60 | 146 |
| `erdos_330` | 1 | 1 |
| empty | 3,551 | 3,551 |

No new data is fetched. `versoFragments` is already loaded on the page and already renders these statements through `FC.problemDocHTML`; the filter now matches against it as well as the name. `build.js` deliberately keeps `statement` and `docstring` out of `conjectures.json`, and this does not change that.

The fragment HTML is stripped to text once per theorem and cached, so typing does not re-parse 3551 documents per keystroke. Timed in the browser: 13ms for the first keystroke that builds the cache, then 5 to 18ms.

The placeholder said "Search by theorem name", which was accurate and is now "Search names and statements".

### DIFF
--- a/site/src/js/browse.js
+++ b/site/src/js/browse.js
@@ -88,10 +88,31 @@ function writeURL() {
 // ---------------------------------------------------------------------------
 // Filter / sort
 // ---------------------------------------------------------------------------
+// The statement text lives in the Verso fragments, as HTML. Strip the tags once per
+// theorem and keep the result, so typing does not re-parse 3551 documents on every keystroke.
+const statementTextCache = new Map();
+
+function statementText(c) {
+  if (!statementTextCache.has(c.theorem)) {
+    const html = FC.problemDocHTML(c, versoFragments);
+    let text = '';
+    if (html) {
+      const el = document.createElement('div');
+      el.innerHTML = html;
+      text = (el.textContent || '').toLowerCase();
+    }
+    statementTextCache.set(c.theorem, text);
+  }
+  return statementTextCache.get(c.theorem);
+}
+
 function applyFilters() {
   const q = state.query.toLowerCase();
   filtered = allConjectures.filter(c => {
-    if (q && !c.displayTheorem.toLowerCase().includes(q)) return false;
+    // Match the statement as well as the name: someone looking for "distinct distances"
+    // knows the mathematics, not that we call it `Erdos307.erdos_307.variants.coprime`.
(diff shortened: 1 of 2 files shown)


site/src/js/browse.js
```diff
@@ -88,10 +88,31 @@ function writeURL() {
 // ---------------------------------------------------------------------------
 // Filter / sort
 // ---------------------------------------------------------------------------
+// The statement text lives in the Verso fragments, as HTML. Strip the tags once per
+// theorem and keep the result, so typing does not re-parse 3551 documents on every keystroke.
```
mo271:
```suggestion
// theorem and keep the result, so typing does not re-parse all the entries on every keystroke.
```
or something, that number will hopefully go up!

mo271 (approved):
Thanks, great to mark search better here!

site/src/js/browse.js
mo271: really just a nit, will also happily merge without that change...

williamjblair: Good point, done. The PR body still quotes 3551 but that one is a measurement (rather than claim about the future)

PR #4833, opened 2026-08-07
feat(Wikipedia/Hadamard): prove isHadamard_H12

kukushking:
# Description

> A square matrix $M$ with $\pm 1$-entries that satisfies the equality
> $|M| = n^{\frac{n}{2}}$ is called a *Hadamard matrix*. The Hadamard conjecture
> asks whether one exists for every $n = 4k$.
> [[Source]](https://en.wikipedia.org/wiki/Hadamard_matrix#Hadamard_conjecture)

- Closes the `category test` sorry on `isHadamard_H12`: Hadamard's own 12×12 construction satisfies the definition.
- Single self-contained proof:
    - `hmul` — the orthogonality identity `H12ᵀ * H12 = 12·1`, checked entrywise (`fin_cases` over the 144 entries, `norm_num` on each 12-term dot product). This exceeds Lean's default heartbeat budget, hence the `set_option maxHeartbeats` in the diff.
    - Entry condition — all entries `±1`, same entrywise sweep.
    - Determinant condition — derived from orthogonality: `det(H12ᵀ·H12) = det(H12)² = 12¹²`, so `|det H12| = 12⁶`, then rpow bookkeeping to match the `n^(n/2)` form.
- The file defines `IsHadamard` via `|det|` and `IsHadamard'` via `MᵀM = n·1`, with the equivalence (`isHadamard_equiv_isHadamard'`) still `sorry` in the `IsHadamard → IsHadamard'` direction (the equality case of Hadamard's determinant inequality). This proof therefore derives `|det|` from orthogonality inline rather than citing the equivalence. If the remaining direction lands at some point, `isHadamard_H12` could be golfed to route through `IsHadamard'` — happy to revisit then.
- :robot: Generated by a custom Lean-kernel-verified theorem-proving harness — an attempt loop with compiler-error feedback, driving **Claude Fable 5** as the LLM backend. Closed on attempt 1 with no target-specific hints; the interesting wrinkle was on the verification side: the entrywise sweep is correct but exceeds the default `maxHeartbeats`, so the harness retries verification under a raised budget. Reviewed and verified locally; the axiom check below is the soundness gate.

### References

- [Wikipedia: Hadamard matrix — Hadamard conjecture](https://en.wikipedia.org/wiki/Hadare)

# Testing

:white_check_mark: Builds cleanly on current `main`:
```shell
$ lake build FormalConjectures.Wikipedia.Hadamard
Build completed successfully (8044 jobs).

:white_check_mark: Kernel axiom check — only Lean foundation axioms, no sorryAx, no user axioms:
'Hadamard.isHadamard_H12' depends on axioms: [propext, Classical.choice, Quot.sound]

:white_check_mark: Theorem statement is byte-identical to the original (no signature dr

### CODE REVIEW

File: FormalConjectures/Wikipedia/Hadamard.lean [hunk 1]
```diff
@@ -108,12 +108,30 @@ def H12 : Matrix (Fin 12) (Fin 12) ℝ :=
      1,  1, -1,  -1,  1,  1,  -1, -1,  1,  -1,  1, -1;
      1,  1, -1,   1, -1,  1,   1, -1, -1,  -1, -1,  1;
      1,  1, -1,   1,  1, -1,  -1,  1, -1,   1, -1, -1 ]
+set_option maxHeartbeats 1600000 in
```
mo271: I'm, a bit worried about raising the maxHeartbeats here, generally we try to avoid it. 

How long does it take before/after to `lake build` this file?

mo271 (commented):
Thanks for the contribution!

FormalConjectures/Wikipedia/Hadamard.lean
kukushking: Yep makes sense. Took about ~33s on an M5 (5s for sorry), with heartbeat 800000 was reliably timing our every time

kukushking: @mo271

mo271: This will then be one of the slowest file in our repo. Perhaps we can speed it up?

If not let's at least add a TODO to speed it up in the future

kukushking: I refactored a bit, but now it's bigger than L50 though :)

PR #4837, opened 2026-08-08
ci: only welcome first-time contributors

mo271:


mo271:
to annoying for frequent contributors...

Paul-Lez (approved):

PR #4850, opened 2026-08-10
Arxiv/2607.06396: the Alon-Tarsi short cycle cover conjecture

### LINKED ISSUE #4813: Alon–Tarsi 7/5 short cycle cover conjecture for bridgeless graphs
### What is the conjecture
**Conjecture (Alon–Tarsi, 1985):** Every bridgeless finite graph $G = (V, E)$ contains a family of cycles $\mathcal{C}$ such that every edge of $G$ belongs to at least one cycle in $\mathcal{C}$, and the total length of the cycles satisfies

$$
\sum_{C \in \mathcal{C}} |E(C)| \le \frac{7}{5} |E(G)|.
$$

* **Source Paper:** *Some new results on Sylvester colorings of cubic graphs* ([arXiv:2607.06396](https://arxiv.org/abs/2607.06396), [HTML](https://arxiv.org/html/2607.06396), [PDF](https://arxiv.org/pdf/2607.06396))
* **Authors:** Luca Ferrarini, Vahan Mkrtchyan
* **Original Reference:** N. Alon, M. Tarsi (1985), *Covering multigraphs by simple circuits*, SIAM Journal on Algebraic Discrete Methods 6(3), 445–450.
* **In-Paper Location:** [Conjecture 4 (Page 3)](https://arxiv.org/pdf/2607.06396#page=3) in Section 1 "Introduction and Notations"
* **OpenConjecture ID:** 4523 ([OpenConjecture](https://github.com/davisrbr/conjectures-arxiv))

### Prerequisites needed
None. Standard graph theory: simple graphs `SimpleGraph`, bridgelessness / 2-edge-connectivity, subgraphs and cycles `Subgraph.IsCycle`, and finite edge-set cardinalities.

### Choose either option

williamjblair:
Closes #4813.

Alon and Tarsi (1985): every bridgeless graph has a list of cycles covering every edge, with total length at most `7/5 |E|`. Quoted from Conjecture 4 of [Ferrarini and Mkrtchyan](https://arxiv.org/abs/2607.06396), which is where the issue points.

```lean
answer(sorry) ↔ ∀ (V : Type) [Fintype V] [DecidableEq V] (G : SimpleGraph V)
  [DecidableRel G.Adj], Bridgeless G →
  ∃ C : Multiset (Cycle G), IsCycleCover G C ∧
    (totalLength C : ℚ) ≤ 7 / 5 * #G.edgeFinset
```

`Multiset` rather than `Finset`, since the source says "a list of cycles" and nothing stops a cover using one twice. `ℚ` on the right so `7/5` is the rational, not `1`.

Two things are proved rather than stated:

- `three_le_length`, that every cycle has at least three edges. Without it `7/5 |E|` reads as though short covers were free.
- `exists_cover_of_edgeFinset_eq_empty`, the empty cover on an edgeless graph.

`edgeFinset_eq_empty_of_bridgeless_of_isAcyclic` records why trees are not counterexamples: in a forest every edge is a bridge, so a bridgeless acyclic graph has no edges at all and the conjecture is vacuous there rather than false.

Builds clean under `--wfail`, and the two proved statements give the standard three axioms.

### DIFF
--- a/FormalConjectures/Arxiv/2607.06396/AlonTarsi.lean
+++ b/FormalConjectures/Arxiv/2607.06396/AlonTarsi.lean
@@ -30,37 +30,18 @@ Every bridgeless graph has a list of cycles covering every edge whose lengths su
 $\frac{7}{5}|E|$.
 -/
 
-open Finset
+open Finset SimpleGraph
 
 namespace Arxiv.«2607.06396»
 
 variable {V : Type*} [Fintype V] [DecidableEq V]
-
-/-- A cycle of `G`, carrying its basepoint. Cycles are collected into a `Multiset` below, so a
-cover may use the same cycle more than once. -/
-structure Cycle (G : SimpleGraph V) where
-  /-- The vertex the cycle starts and ends at. -/
-  base : V
-  /-- The closed walk tracing the cycle. -/
-  walk : G.Walk base base
-  /-- That walk is a cycle. -/
-  isCycle : walk.IsCycle
-
-/-- The edges a cycle traverses. -/
-def Cycle.edges {G : SimpleGraph V} (c : Cycle G) : List (Sym2 V) := c.walk.edges
-
-/-- The length of a cycle, its number of edges. -/
-def Cycle.length {G : SimpleGraph V} (c : Cycle G) : ℕ := c.walk.length
 
 /-- `C` covers `G`: every edge of `G` lies on at least one cycle of `C`. -/
 def IsCycleCover (G : SimpleGraph V) [DecidableRel G.Adj] (C : Multiset (Cycle G)) : Prop :=
   ∀ e ∈ G.edgeFinset, ∃ c ∈ C, e ∈ c.edges
 
 /-- The total length of a family of cycles. -/
-def totalLength {G : SimpleGraph V} (C : Multiset (Cycle G)) : ℕ := (C.map Cycle.length).sum
-
-/-- `G` is bridgeless: no edge is a bridge. -/
-def Bridgeless (G : SimpleGraph V) : Prop := ∀ e ∈ G.edgeSet, ¬ G.IsBridge e
+def totalLength {G : SimpleGraph V} (C : Multiset (Cycle G)) : ℕ := (C.map SimpleGraph.Cycle.length).sum
 
 /--
 **Conjecture 4 (Alon-Tarsi, 1985).** Every bridgeless graph has a list of cycles covering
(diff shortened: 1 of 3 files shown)


FormalConjectures/Arxiv/2607.06396/AlonTarsi.lean
```diff
@@ -0,0 +1,105 @@
...
+
+/-- A cycle of `G`, carrying its basepoint. Cycles are collected into a `Multiset` below, so a
+cover may use the same cycle more than once. -/
+structure Cycle (G : SimpleGraph V) where
+  /-- The vertex the cycle starts and ends at. -/
+  base : V
+  /-- The closed walk tracing the cycle. -/
+  walk : G.Walk base base
+  /-- That walk is a cycle. -/
+  isCycle : walk.IsCycle
```
mo271: I think I see why you would define this structure instead of just using `IsCycle`: you want to talk about the collection of all cycles, hence need to form a type. `structure` seems reasonable here, but 
Alternatively you could do 
```
/-- A cycle is a basepoint together with a closed walk on that basepoint that is a cycle. -/
abbrev Cycle (G : SimpleGraph V) := Σ (u : V), { p : G.Walk u u // p.IsCycle }
-- Constructing a term:
def myCycle (u : V) (w : G.Walk u u) (hw : w.IsCycle) : Cycle G :=
  ⟨u, ⟨w, hw⟩⟩
-- Accessing fields (uses positional indexing):
def Cycle.base (c : Cycle G) : V := c.1
def Cycle.walk (c : Cycle G) : G.Walk c.1 c.1 := c.2.1
def Cycle.length (c : Cycle G) : ℕ := c.2.1.length
```
but I think a structure is better here.

Perhaps even a good fit for `FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Cycle.lean`?

FormalConjectures/Arxiv/2607.06396/AlonTarsi.lean
```diff
@@ -0,0 +1,105 @@
...
+      ∃ C : Multiset (Cycle G), IsCycleCover G C ∧
+        (totalLength C : ℚ) ≤ 7 / 5 * #G.edgeFinset := by
+  sorry
+
+/-- An acyclic graph is bridgeless only if it has no edges, since in a forest every edge is a
+bridge. So the conjecture says nothing about trees beyond the empty cover. -/
+@[category API, AMS 5]
+theorem edgeFinset_eq_empty_of_bridgeless_of_isAcyclic (G : SimpleGraph V) [DecidableRel G.Adj]
+    (hacyc : G.IsAcyclic) (hbr : Bridgeless G) : G.edgeFinset = ∅ := by
+  sorry
```
mo271:
```suggestion
  rw [SimpleGraph.edgeFinset_eq_empty]
  ext u v
  simp only [SimpleGraph.bot_adj, iff_false]
  intro hadj
  exact hbr s(u, v) hadj (G.isAcyclic_iff_forall_edge_isBridge.mp hacyc hadj)
```

let's prove?

mo271: perhaps good to move to `FormalConjecturesForMathlib`?

FormalConjectures/Arxiv/2607.06396/AlonTarsi.lean
```diff
@@ -0,0 +1,105 @@
...
+every edge, with $\sum_{C} |E(C)| \leq \frac{7}{5}|E(G)|$.
+-/
+@[category research open, AMS 5]
+theorem alon_tarsi_short_cycle_cover :
+    answer(sorry) ↔ ∀ (V : Type) [Fintype V] [DecidableEq V] (G : SimpleGraph V)
+      [DecidableRel G.Adj], Bridgeless G →
+      ∃ C : Multiset (Cycle G), IsCycleCover G C ∧
+        (totalLength C : ℚ) ≤ 7 / 5 * #G.edgeFinset := by
+  sorry
+
```
mo271, line 57:
```suggestion

omit [DecidableEq V] in
```
should also work here, as below

mo271 (commented):
Thanks, Looks great!
Just some suggestions

williamjblair:
Thank you - `Cycle`, the bridgeless predicate and the acyclic lemma are now in `FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Cycle.lean`, with your proof. I also renamed `Bridgeless` to `IsBridgeless` to sit alongside `IsBridge` and `IsAcyclic`, and kept the structure rather than the sigma type.

The Arxiv file keeps a test that chains the two: an acyclic bridgeless graph has no edges, so the empty cover works. Only the conjecture is sorried now.

mo271 (approved):
Thanks, LGTM!

PR #4861, opened 2026-08-10
ForMathlib/Data/Finset/Card: drop a lemma Mathlib already has

williamjblair:
`Finset.card_le_card_iff_of_subset` is Mathlib's `Finset.eq_iff_card_le_of_subset` under a different name. It is there at the Mathlib revision this repo pins, so the duplicate has been sitting here since the file was written.

The only user was `card_lt_card_iff_of_subset` just below it, which now rewrites with the Mathlib lemma. `Johnson.lean` uses that one and is unchanged.

Name matching will not find this sort of thing, since the names differ. I checked by restating the lemma in a file that imports only Mathlib and running `exact?`.


mo271 (approved):
Thanks, LGTM!

PR #4863, opened 2026-08-10
fix(Wikipedia): rule out negative radii in SquarePacking

### LINKED ISSUE #4859: Wikipedia/SquarePacking: least-radius statements are false because negative radii are allowed
In the [`FormalConjectures/Wikipedia/SquarePacking.lean`](https://github.com/google-deepmind/formal-conjectures/blob/c252a41054125b5fd9c8356e2137cd9b55337657/FormalConjectures/Wikipedia/SquarePacking.lean#L60-L61), it defines `Circle (r : ℝ)` using only `r ^ 2`. This allows negative radii. As a result, the following two declarations, both currently marked `research open`, can be resolved immediately by disproving their formal statements:

- [`least_three_square_packing_in_circle`](https://github.com/google-deepmind/formal-conjectures/blob/c252a41054125b5fd9c8356e2137cd9b55337657/FormalConjectures/Wikipedia/SquarePacking.lean#L136-L139)
- [`least_fifteen_circle_packing_in_circle`](https://github.com/google-deepmind/formal-conjectures/blob/c252a41054125b5fd9c8356e2137cd9b55337657/FormalConjectures/Wikipedia/SquarePacking.lean#L179-L182)

This does not solve either intended packing problem. The immediate resolution is possible only because the formal statements are misformalized.

### Trivial proof/disproof
- If `0 < r`, then `Circle (-r) = Circle r`, so the same packing is feasible at `-r`, with `-r < r`.
- If `r ≤ 0`, then `r ^ 2 < (r - 1) ^ 2`, hence `Circle r ⊆ Circle (r - 1)`. The same packing is therefore feasible at `r - 1`, with `r - 1 < r`.

Thus every feasible radius has a smaller feasible radius, so no least element exists.

### Suggested fix
Restrict the candidate radii to nonnegative reals:

```lean
@[category research open, AMS 51]
theorem least_three_square_packing_in_circle :
    IsLeast
      {r : ℝ | 0 ≤ r ∧ Nonempty (Packing 3 UnitSquare (Circle r))}
      answer(sorry) := by
  sorry

@[category research open, AMS 51]
theorem least_fifteen_circle_packing_in_circle :
    IsLeast
      {r : ℝ | 0 ≤ r ∧ Nonempty (Packing 15 UnitCircle (Circle r))}
      answer(sorry) := by
  sorry
```

Alternatively, use `ℝ≥0` for the radius type.

### Additional context
This affects 2 of the 5 declarations marked `research open`. The other 3 minimize `Square x` and are unaffected.

### Choose either option

bocowgill:
Fixes #4859.

## Summary

`Circle` previously took `r : ℝ` and depended only on `r ^ 2`, making `Circle (-r) = Circle r`. Consequently, the feasible-radius sets in the two circle-minimising statements were unbounded below and could have no least element.

Following the discussion on #4859, this changes the radius type to `ℝ≥0`, making negative radii unrepresentable. The two minimised sets and their `answer` values now also use `ℝ≥0`.

The fixed positive-radius expressions are rewritten using `NNReal.sqrt`. After coercion to `ℝ`, `Real.coe_sqrt` shows that they denote the same values as before. The change also adds a proved sanity check that `Circle 0 = ∅`.

The `Square` definition is unchanged because nonpositive side lengths already produce the empty set and are not affected by this defect.

## Reference clarification

Also adds the [Pirl (1969)](https://doi.org/10.1002/mana.19690400110) reference and notes that he conjectured the 15-circle configuration to be optimal. The existing Graham–Lubachevsky–Nurmela–Östergård reference remains the source for the stated value.

## Verification

- `lake --wfail build`
- `lake --wfail test`
- `extract_names` + category checker: `research_open_with_proof: 0`
- Changed-file category diagnostics: all zero

## AI assistance disclosure

Drafted with AI assistance and machine-checked by the Lean kernel. I reviewed the resulting formalisation and take responsibility for the submitted change.

### DIFF
--- a/FormalConjectures/Wikipedia/SquarePacking.lean
+++ b/FormalConjectures/Wikipedia/SquarePacking.lean
@@ -29,11 +29,16 @@ In each case, we provide a known upper bound, and ask for the least such size.
 - [Wikipedia on packing of circles in a square](https://en.wikipedia.org/wiki/Circle_packing_in_a_square)
 - Friedman, Erich (2009), "Packing unit squares in squares: a survey and new results",
   Electronic Journal of Combinatorics, 1000, Dynamic Survey 7
+- Pirl, U. (1969),
+  ["Der Mindestabstand von $n$ in der Einheitskreisscheibe gelegenen Punkten"](https://doi.org/10.1002/mana.19690400110),
+  Mathematische Nachrichten, 40: 111–124
 - A website with visualizations of packings:
   [link](https://erich-friedman.github.io/packing/)
 -/
 
 open EuclideanGeometry
+
+open scoped NNReal
 
 universe u
 
@@ -54,11 +59,12 @@ def UnitSquare : Set ℝ² := Square 1
 
 /--
 A circle of a particular radius as a subset of the Euclidean plane.
+The radius is a nonnegative real, so that `Circle r` is always the disc of radius $r$.
 Not including border, so that circles that touch at the border are disjoint,
 but a circle internal to another shape is a subset of that shape.
 -/
-def Circle (r : ℝ) : Set ℝ² :=
-  {p : ℝ² | p 0 ^ 2 + p 1 ^ 2 < r ^ 2}
+def Circle (r : ℝ≥0) : Set ℝ² :=
+  {p : ℝ² | p 0 ^ 2 + p 1 ^ 2 < (r : ℝ) ^ 2}
 
 /--
 The unit circle as a subset of the Euclidean plane.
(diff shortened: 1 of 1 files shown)


bocowgill:
@BoltonBailey This implements the `ℝ≥0` radius design you suggested in #4859. The PR is ready for review; a review of the definition/API choice would be appreciated when convenient.

BoltonBailey (approved):
Bit unfortunate to lose the notation, but looks like a correct fix.

mo271 (approved):
Thanks, LGTM!

mo271:
Thanks for the fix and thanks for the review @BoltonBailey!

Issue #2388, opened 2026-02-24
also list `formally solved` on stats page


mo271:
Make sure our new `formally solved ...` categories are represented at

https://google-deepmind.github.io/formal-conjectures/

Perhaps both also aggregated as "solved", "formally solved" (and then split up by `formalProofKind`

PR #4874, opened 2026-08-11
feat(site): plot open, solved and formally proved over time

christianTNG:
## Summary

- add `site/status_counts.py`, which counts open statements, solved statements, and statements with a link to a formal proof at every commit
- generalise `plot_growth.py` so a plot can carry several series, and generate a second plot from those counts
- show it on the stats page as "Statements by status", next to the existing file-count plot
- cache counts per blob hash so the history walk stays cheap

Related to #2388, which asks for formally solved statements to be represented on the site. That issue is assigned to @Paul-Lez, so please close this if it overlaps with work already in progress.

## Why

The stats page shows how many Lean files the repository has over time, but nothing about what is in them. The landing page has the current open/solved/formally proved counts, but no history, so it is not visible whether formal proofs are keeping pace with new statements. This adds that history.

## How the counts are obtained

The counts on the site come from `lake exe extract_names`, which needs a full Lean build and so cannot be repeated for ~1600 commits. `status_counts.py` reads them from the source text instead, following the same rules `extract_names` applies:

- only `theorem`, `lemma` and `instance` declarations carry a category
- declarations Lean treats as internal are dropped, that is `private` ones and those with a `_`-prefixed name component
- a declaration counts once towards formally proved however many `formal_proof` attributes it carries
- comments and docstrings are stripped first, since they mention the attribute names in prose

Before #3645 a formal proof was recorded as the category `research formally solved` rather than a separate `formal_proof` attribute, so that spelling is recognised too. Formal proofs were only recorded from January 2026 onwards, so the third series is genuinely zero before then.

This is a second implementation of rules that live in `extract_names`, and it can drift if the attribute syntax changes. To make that visible, the script prints its last data point, which the workflow log shows next to the counts the build produces.

## Verification

- counts checked against the `data/conjectures.json` the live site serves: at the commit it was built from they agree exactly, 1202 open, 1415 solved, 328 formally proved
- `python site/plot_growth.py` with the versions CI pins (`pandas==2.2.3 numpy==2.2.3 plotly==5.20.0`), 16s for all four plots, roughly what the file-count plot alone took before
- `cd site && node build.js`, then both plots checked in a browser in light and dark mode

## AI assistance disclosure

Drafted with AI assistance. The change was built and checked locally, as described under Verification.

### DIFF
--- a/.github/workflows/build-and-docs.yml
+++ b/.github/workflows/build-and-docs.yml
@@ -255,6 +255,8 @@ jobs:
           echo "Downloading growth plots from $LIVE_URL ..."
           curl -sfL "$LIVE_URL/data/file_counts_white.html" -o site/data/file_counts_white.html || true
           curl -sfL "$LIVE_URL/data/file_counts_dark.html" -o site/data/file_counts_dark.html || true
+          curl -sfL "$LIVE_URL/data/status_counts_white.html" -o site/data/status_counts_white.html || true
+          curl -sfL "$LIVE_URL/data/status_counts_dark.html" -o site/data/status_counts_dark.html || true
           node -e "
             const fs = require('fs');
             const data = JSON.parse(fs.readFileSync('/tmp/live_conjectures.json', 'utf8'));
--- a/site/build.js
+++ b/site/build.js
@@ -690,6 +690,10 @@ async function main() {
   const darkPlotPath = path.join('data', 'file_counts_dark.html');
   if (fs.existsSync(whitePlotPath)) fs.copyFileSync(whitePlotPath, 'site/data/file_counts_white.html');
   if (fs.existsSync(darkPlotPath)) fs.copyFileSync(darkPlotPath, 'site/data/file_counts_dark.html');
+  const whiteStatusPlotPath = path.join('data', 'status_counts_white.html');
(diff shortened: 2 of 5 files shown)


christianTNG:
formal-conjectures-statements-by-status

mo271 (commented):
before taking a more closer look: 

typically it is brittle to parse the lean code via python, but I can see how running `extract_names`, also this script didn't always exists and not always in the same form...

In any case, many thanks for the contribution!

christianTNG:
Thanks for raising the brittleness of parsing Lean from Python. Your own caveat is why I went this way: `extract_names` needs a full build and hasn't always existed, so it can't be re-run over the history.

To make any drift loud rather than silent, I drafted [this branch](https://github.com/google-deepmind/formal-conjectures/compare/main...christianTNG:formal-conjectures:stats-status-over-time-with-testing) on top of this PR: unit tests for the parsing rules, and a CI step comparing the counts with `extract_names` on every build. They match exactly today.

Happy to fold it in, or leave it out if you'd rather keep this PR small. If a written walkthrough is useful, my AI agent annotated the diffs here: https://storage.googleapis.com/chstdu-tng/formal-conjectures-changes.html (static page, nothing to run).

PR #4877, opened 2026-08-11
ErdosProblems/80: restrict c to the range where the hypothesis is satisfiable

### LINKED ISSUE #4867: Erdős Problem 80: both open statements are false for c = 2
This concerns Erdős Problem 80 in `FormalConjectures/ErdosProblems/80.lean`.

The two `research open` declarations, `erdos_80` and `erdos_80.variants.log`, quantify over every `c > 0`. This makes both statements false at `c = 2` for an unintended reason.

Permalink: [`80.lean`, lines 51–83](https://github.com/google-deepmind/formal-conjectures/blob/8adf5b7d199d83d7c69d2f89d0e7327362ead98a/FormalConjectures/ErdosProblems/80.lean#L51-L83)

### Trivial proof/disproof
`Admissible c G` means that `G` has at least `cn²` edges and every edge lies in a triangle. The value `f c n` is the minimum, over all such `n`-vertex graphs, of the largest number of triangles sharing one edge. It is `0` if there are no such graphs.

A simple graph on `n` vertices has at most `n(n - 1) / 2` edges. For `c = 2`, at least `2n²` edges would be required, which is impossible for `n ≥ 1`. Hence `f 2 n = 0`. Therefore both open declarations have the answer `False`.

### Suggested fix
Restrict `c` to the feasible range in both declarations, for example `0 < c ∧ c < 1/2`.

### Additional context
The file was introduced in #4830. Original problem: https://www.erdosproblems.com/80

### Choose either option

williamjblair:
Closes #4867.

Both statements quantified over every `c > 0`. `Admissible c G` needs `c * n^2 ≤ #G.edgeFinset`, but a simple graph on `n` vertices has at most `n(n-1)/2` edges, so nothing qualifies once `c ≥ 1/2`. At `c = 2` the set inside `sInf` is empty for every `n`, `f 2 n = 0`, and both answers are `False` for reasons unrelated to the question. Adding `c < 1/2` is @KitaKen1's suggestion and it is the right bound: at `c` just under `1/2` the complete graph qualifies from about `n = 100` on.

Also corrects the `f` docstring. It said the empty case was "the honest reading: nothing is forced when nothing qualifies", which is true as far as it goes but was exactly what hid the problem. It now says when the empty case occurs and why the statements exclude it.

My error, from #4830. Statements only, no proofs affected.

### DIFF
--- a/FormalConjectures/Arxiv/2607.05349/MicroscopicWeighting.lean
+++ b/FormalConjectures/Arxiv/2607.05349/MicroscopicWeighting.lean
@@ -44,11 +44,8 @@ noncomputable def distanceMatrix (X : Type*) [Fintype X] [MetricSpace X] : Matri
 /-- The weighting $\vec{w}(t) = Z(t)^{-1}\mathbf{1}$ at scale `t`.
 
 `Matrix.inv` is `0` on singular matrices, so this is only the intended vector where `Z t` is
-invertible. That holds for all small enough `t > 0`, which is where the limit below is taken.
-Continuity alone does not give this: `Z 0` is the all-ones matrix, which is singular once `X`
-has two points, so `(Z t).det` tends to `0` as `t → 0`. The reason is instead that `(Z t).det`
-is analytic in `t` and tends to `1` as `t → ∞`, since the off-diagonal entries vanish there.
-So it is not identically zero, and its zeros are isolated. -/
+invertible. That is enough here: `Z 0` is the all-ones matrix and `Z` is continuous, so `Z t`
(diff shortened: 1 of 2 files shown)


FormalConjectures/Arxiv/2607.05349/MicroscopicWeighting.lean
mo271, line 1: this change should go elsewhere?

mo271 (commented):
Thanks, LGTM for change to ErdosProblems/80

williamjblair:
Oops, thank for catching that. Removed now

mo271 (approved):
Thanks, LGTM!

PR #4878, opened 2026-08-11
Arxiv/2605.02731: Dean's conjecture on cycles of length divisible by k

williamjblair:
Closes #4857. OpenConjecture ID 2455.

Dean says that a graph with a minimum degree of `k` or more has a cycle. The length of that cycle divides by `k`. This is true for each integer `k` that is 3 or more.

```lean
answer(sorry) ↔ ∀ (k : ℕ), 3 ≤ k → ∀ (V : Type) [Fintype V] [DecidableEq V]
  (G : SimpleGraph V) [DecidableRel G.Adj], k ≤ G.minDegree →
    ∃ m ∈ G.cycleLengths, k ∣ m
```

The file gives five statements. One is the conjecture. One is the case `k = 5`, which is open. Three are the solved cases:

| statement | case | source |
| --- | --- | --- |
| `dean_conjecture` | all `k ≥ 3` | Dean 1988 |
| `variants.five` | `k = 5` | open |
| `variants.three` | `k = 3` | Chen and Saito 1994 |
| `variants.four` | `k = 4` | Dean, Lesniak and Saito 1993 |
| `variants.six_le` | `k ≥ 6` | Liu, Ma and Zhao 2026 |

Two conditions prevent a trivial answer. A cycle has a length of 3 or more, so the divisor is never `0`. `SimpleGraph.minDegree` is `0` for a graph with no vertices and for a graph with no edges, so the hypothesis excludes both. The two `test` statements show this. Their axioms are `propext`, `Classical.choice` and `Quot.sound` only.

The file uses `SimpleGraph.cycleLengths` from `FormalConjecturesForMathlib`.

Builds clean. Five sorries, one for each research statement.

PR #4879, opened 2026-08-11
Arxiv/2606.03696: Bondy's conjecture on longest cycles

### LINKED ISSUE #4858: Bondy's conjecture on longest cycles in highly connected graphs
### What is the conjecture

Let $k \ge 1$ be an integer, and let $G$ be a $k$-connected graph on $n$ vertices. If

$$\delta(G) \ge \frac{n + k(k-1)}{k+1},$$

then for every longest cycle $C$ of $G$, every path in $G - V(C)$ has at most $k-1$ vertices.

**Context:** Proposed by Bondy in 1980 as a higher-connectivity generalization of a classical theorem of Nash-Williams. The cases $k = 1$ and $k = 2$ are precisely Dirac's theorem and Nash-Williams' theorem, respectively. The case $k = 3$ has been verified. The conjecture remains open for all $k \ge 4$. In the source paper, Ma, Ning, and Zhao confirm Bondy's conjecture for all **sufficiently large** graphs (Theorem 1.1), but the full conjecture for all graph sizes remains open.

* **Source Paper:** *Longest cycles and Dirac-type results in highly connected graphs* ([arXiv:2606.03696v1](https://arxiv.org/abs/2606.03696v1), [HTML](https://arxiv.org/html/2606.03696v1), [PDF](https://arxiv.org/pdf/2606.03696v1))
* **Authors:** Jie Ma, Bo Ning, Ziyuan Zhao
* **In-Paper Location:** [Conjecture 1 (Page 1)](https://arxiv.org/pdf/2606.03696v1#page=1) in Section "Introduction"
* **OpenConjecture ID:** 4126 ([OpenConjecture](https://github.com/davisrbr/conjectures-arxiv))

### Prerequisites needed

* **Simple graphs:** [`Mathlib.Combinatorics.SimpleGraph.Basic`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/SimpleGraph/Basic.html)
* **Connectivity:** [`Mathlib.Combinatorics.SimpleGraph.Connectivity.Subgraph`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/SimpleGraph/Connectivity/Subgraph.html)
* **Walks and paths:** [`Mathlib.Combinatorics.SimpleGraph.Walk.Basic`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/SimpleGraph/Walk/Basic.html)
* **Degree sums:** [`Mathlib.Combinatorics.SimpleGraph.DegreeSum`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/SimpleGraph/DegreeSum.html)
* **Missing:** Mathlib does not currently have a definition of $k$-connectedness for simple graphs, longest cycles, or vertex-deleted subgraphs $G - V(C)$. These would need to be defined.

### Choose either option

williamjblair:
Closes #4858. OpenConjecture ID 4126.

Take a `k`-connected graph `G` on `n` vertices. Bondy says that if the minimum degree is at least `(n + k(k-1))/(k+1)`, then no path outside a longest cycle has more than `k-1` vertices.

## Two definitions this file adds

Mathlib has `SimpleGraph.IsEdgeConnected` but no vertex connectivity, so the file states it:

```lean
def IsKConnected (G : SimpleGraph V) (k : ℕ) : Prop :=
  k < Fintype.card V ∧ ∀ S : Finset V, S.card < k → (G.induce ((S : Set V)ᶜ)).Connected
```

`offWalk C` gives the vertices that the cycle does not touch. `G.induce (offWalk C)` is then the graph `G - V(C)` of the conjecture. A longest cycle is a cycle whose length equals `SimpleGraph.circumference`, which `FormalConjecturesForMathlib` supplies.

## Check on the reading

The source says the case `k = 1` is Dirac's theorem. The file agrees. At `k = 1` the conclusion is `P.support.length + 1 ≤ 1`. Every walk has a support of length 1 or more, so no path can exist outside the cycle. The cycle must then cover every vertex, which is what Dirac gives. I proved `1 ≤ P.support.length` to confirm this.

The bound reads `+ 1 ≤ k` and not `≤ k - 1`, because subtraction on `ℕ` is truncated. The two forms agree for `k ≥ 1`.

## Statements

| statement | case | status |
| --- | --- | --- |
| `bondy_conjecture` | all `k ≥ 1` | open |
| `variants.one` | `k = 1` | Dirac |
| `variants.two` | `k = 2` | Nash-Williams |
| `variants.three` | `k = 3` | proved |
| `variants.large` | all `k`, large graphs | Ma, Ning and Zhao 2026 |

The cases `k ≥ 4` stay open for graphs of every size.

Builds clean. Five sorries, one for each research statement.

### DIFF
--- a/FormalConjectures/Arxiv/2606.03696/BondyLongestCycles.lean
+++ b/FormalConjectures/Arxiv/2606.03696/BondyLongestCycles.lean
@@ -1,0 +1,128 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# Bondy's conjecture on longest cycles in highly connected graphs
+
+*References:*
+- [arxiv/2606.03696](https://arxiv.org/abs/2606.03696)
+  **Longest cycles and Dirac-type results in highly connected graphs**
+  by *Jie Ma, Bo Ning, Ziyuan Zhao*, where this is Conjecture 1.
+- [Bo80] Bondy, J. A., Longest paths and cycles in graphs of high degree. (1980).
+
+Take a `k`-connected graph with a large minimum degree. Bondy says that the graph outside any
+longest cycle holds no long path.
+
+The case `k = 1` is Dirac's theorem and the case `k = 2` is the theorem of Nash-Williams. The
+case `k = 3` is proved. The cases `k ≥ 4` are open.
+-/
+
+open SimpleGraph
+
+namespace Arxiv.«2606.03696»
+
+variable {V : Type*} [Fintype V] [DecidableEq V]
+
+/-- `G` is `k`-connected: it has more than `k` vertices, and it stays connected after the
+removal of any set of fewer than `k` vertices.
+
+Mathlib has `SimpleGraph.IsEdgeConnected` but no vertex connectivity, so this file states it. -/
+def IsKConnected (G : SimpleGraph V) (k : ℕ) : Prop :=
+  k < Fintype.card V ∧ ∀ S : Finset V, S.card < k → (G.induce ((S : Set V)ᶜ)).Connected
+
+/-- The set of vertices that a walk does not touch. For a longest cycle `C` this set carries
+the graph `G - V(C)` of the conjecture. -/
+def offWalk {a b : V} {G : SimpleGraph V} (w : G.Walk a b) : Set V :=
+  {v | v ∉ w.support}
+
+/--
+**Conjecture 1 (Bondy, 1980).** Let $k \geq 1$ and let $G$ be a $k$-connected graph on $n$
+vertices. If $\delta(G) \geq \frac{n + k(k-1)}{k+1}$, then for every longest cycle $C$ of $G$,
+every path in $G - V(C)$ has at most $k-1$ vertices.
+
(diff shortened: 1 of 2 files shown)


FormalConjectures/Arxiv/2605.02731/DeanCycles.lean
mo271, line 1: drop this file in the merge https://github.com/google-deepmind/formal-conjectures/pull/4878

FormalConjectures/Arxiv/2606.03696/BondyLongestCycles.lean
```diff
@@ -0,0 +1,128 @@
...
+namespace Arxiv.«2606.03696»
+
+variable {V : Type*} [Fintype V] [DecidableEq V]
+
+/-- `G` is `k`-connected: it has more than `k` vertices, and it stays connected after the
+removal of any set of fewer than `k` vertices.
+
+Mathlib has `SimpleGraph.IsEdgeConnected` but no vertex connectivity, so this file states it. -/
+def IsKConnected (G : SimpleGraph V) (k : ℕ) : Prop :=
+  k < Fintype.card V ∧ ∀ S : Finset V, S.card < k → (G.induce ((S : Set V)ᶜ)).Connected
```
mo271: potentially move to `IsKConnected` and `not_isKConnected_of_card_le` into 
FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Connectivity.lean
.

mo271 (commented):
looks, good mathematically, thanks!
just two comments

williamjblair:
Thanks, both done.

Rebased on main, so `DeanCycles.lean` is gone from the diff now that #4878 has merged.

`IsKConnected` and `not_isKConnected_of_card_le` moved to
`FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Connectivity.lean`, next to
`InternallyDisjoint` and `InfinitelyConnected`. That file needed one extra import,
`Mathlib.Combinatorics.SimpleGraph.Connectivity.Connected`, because `Paths` alone does not give
`SimpleGraph.Connected`.

I left `offWalk` in the problem file. It exists to name `G - V(C)` for this conjecture, so it does
not look like general API to me. Say the word if you would rather it moved too.

mo271 (approved):
Thanks, LGTM!

PR #4523, opened 2026-07-21
fix(GreensOpenProblems/72): add N ≥ k

anagnorisis2peripeteia:
Otherwise we can disprove `no_k_in_line_big` by taking `k = 10 ^ 37 + 1` and `N = 1`.

A retained Lean witness proves
`AllowedSetSize (10 ^ 37 + 1) 1 = 1`, hence
`¬ NoKInLineFor (10 ^ 37 + 1) 1`. The existing declaration nevertheless applies to this
case because it assumes only `10 ^ 37 < k`.

[Theorem 1.1 of GK2025](https://arxiv.org/abs/2510.17743) assumes that the grid size is at
least the number of collinear points. This PR therefore adds the missing `hN : k ≤ N`
hypothesis and records `N ≥ k` in the docstring. It does not change the predicate,
classification, or proof placeholder.

Fixes #4522.

## Update (2026-07-22)

Addressed the maintainer's three prose suggestions after checking the rendered file:

- corrected `N \lt 2` to `N > 2` and removed the extra article;
- documented the actual definition `AllowedSetSize k N = (k - 1) * N` with the parameters
  in declaration order; and
- corrected `verfied` to `verified` without the malformed trailing backtick in the rendered
  suggestion.

The branch was rebased onto current `main`, and the complete gate sequence was rerun on the
amended commit.

## Validation

- `git diff --check origin/main...HEAD`
- `lake env lean FormalConjectures/GreensOpenProblems/72.lean`
- `lake env lean .pr-gates/green72-grid-bound-witness.lean`
- `.pr-gates/proof-green72-74f5a6b/verify.sh` on base `9002c58` and HEAD `74f5a6b`
- `lake exe mk_all --lib FormalConjectures`
- `lake exe cache unpack`
- `lake exe cache get`
- `lake --wfail build`

The paired check uses the same `k = 10 ^ 37 + 1`, `N = 1` scenario on both revisions.
On the base, the theorem accepts a machine-checked false conclusion; on HEAD, the new
`k ≤ N` premise excludes that case. The full Lean 4.27.0 warning-as-error build completed
all 8,892 jobs against the exact unchanged source hash. Screenshots and performance testing
are not applicable to this statement-only correction.

Review history: https://gist.github.com/anagnorisis2peripeteia/dcf401e6c9d07eb4aa8ae2d851587006

## AI assistance disclosure

OpenAI Codex was used to compare the cited source with the Lean declaration, prepare the
one-file correction and paired base/HEAD witness, and run validation.

PR #4880, opened 2026-08-11
Add proofs for four textbook statements

christianTNG:
### What

Replaces `sorry` with a full proof in four `category textbook` statements. Following review, the proofs are inlined here rather than referenced through `formal_proof` links, so all four `formal_proof` attributes added earlier in this PR have been removed again.

| Statement | File | Lines added |
|---|---|---|
| `Green72.allowedSetSize_le` | `FormalConjectures/GreensOpenProblems/72.lean` | 20 |
| `Erdos649.erdos_649.variants.no_solution_two_seven` | `FormalConjectures/ErdosProblems/649.lean` | 27 |
| `Erdos44.maxSidonSubsetCard_icc_bound` | `FormalConjectures/ErdosProblems/44.lean` | 42 |
| `Erdos786.erdos_786.parts.i.example` | `FormalConjectures/ErdosProblems/786.lean` | 68 |

### One statement change

`Green72.allowedSetSize_le` loses its `k ≤ N` hypothesis and now reads `{k N : ℕ} : AllowedSetSize k N ≤ (k - 1) * N`, as suggested in review. The bound holds unconditionally: the column fiber pigeonhole argument bounds each of the `N` columns by `k - 1` points independently of whether `k ≤ N` holds. The docstring drops the matching `for $N \geq k$`.

This does not touch `no_k_in_line_big`, which asserts the equality `AllowedSetSize k N = (k - 1) * N` and therefore does need `k ≤ N` as added in #4523. `allowedSetSize_le` has no call sites, and `AllowedSetSize` is not referenced outside this file.

### Checks

- `lake --wfail build` passes: full build, 9025 jobs, no errors and no warnings.
- All four theorems are `sorry`-free. `#print axioms` reports only `propext`, `Classical.choice`, `Quot.sound` for each.
- No `native_decide` and no `decide +native` in any of the four files.

### Follow up

Happy to look at moving `maxSidonSubsetCard_icc_bound` next to the `SidonSet` definitions in `FormalConjecturesForMathlib` as you suggested, but I would prefer to do that in a separate PR to keep this one focused.

*Disclosure: the proofs were developed with AI assistance (LLM agents) under a machine verification gate: clean `lake --wfail build`, zero `sorry`, and an axiom audit on every theorem.*

### DIFF
--- a/FormalConjectures/ErdosProblems/44.lean
+++ b/FormalConjectures/ErdosProblems/44.lean
@@ -28,11 +28,51 @@ namespace Erdos44
 
 -- Reference: https://arxiv.org/pdf/2103.15850
 /-- The maximum size of a Sidon set in `{1, ..., N}` is less than or equal to `2 * √N`. -/
-@[category textbook, AMS 5 11, formal_proof using formal_conjectures at
-  "https://github.com/christianTNG/formal-conjectures/blob/2d9d7f1380c4897f30004f6c909e42e820d3c65f/FormalConjectures/ErdosProblems/44.lean#L32"]
+@[category textbook, AMS 5 11]
 theorem maxSidonSubsetCard_icc_bound (N : ℕ) (hN : 1 ≤ N) :
     maxSidonSubsetCard (Icc 1 N) ≤ 2 * Real.sqrt N := by
-  sorry
+  have key : maxSidonSubsetCard (Finset.Icc 1 N) ^ 2 ≤ 4 * N := by
+    -- Pick a Sidon subset of `{1, ..., N}` of maximal size.
+    obtain ⟨B, hB, hBeq⟩ := Finset.exists_mem_eq_sup
+      ((Finset.Icc 1 N).powerset.filter fun B : Finset ℕ ↦ IsSidon (B : Set ℕ))
+      ⟨∅, by simp [IsSidon]⟩ Finset.card
+    rw [maxSidonSubsetCard, hBeq]
+    simp only [Finset.mem_filter, Finset.mem_powerset] at hB
+    obtain ⟨hsub, hsid⟩ := hB
+    -- Distinct ordered pairs of elements of `B` have distinct differences, and each difference
+    -- lies in `{-N, ..., N}`.
+    have hcard : B.offDiag.card ≤ (Finset.Icc (-(N : ℤ)) N).card := by
+      refine Finset.card_le_card_of_injOn (fun p ↦ (p.1 : ℤ) - p.2) ?_ ?_
+      · rintro ⟨a, b⟩ hab
+        rw [Finset.mem_coe, Finset.mem_offDiag] at hab
+        have ha' := Finset.mem_Icc.mp (hsub hab.1)
+        have hb' := Finset.mem_Icc.mp (hsub hab.2.1)
+        simp only [Finset.mem_coe, Finset.mem_Icc]
+        omega
+      · rintro ⟨a, b⟩ ha ⟨c, d⟩ hc hac
+        rw [Finset.mem_coe, Finset.mem_offDiag] at ha hc
+        simp only at hac
+        have hsum : a + d = c + b := by omega
+        rcases hsid a ha.1 c hc.1 d hc.2.1 b ha.2.1 hsum with ⟨h1, h2⟩ | ⟨h1, h2⟩
+        · simp [h1, h2]
+        · exact absurd h1 ha.2.2
+    have hIcc : (Finset.Icc (-(N : ℤ)) N).card = 2 * N + 1 := by
+      rw [Int.card_Icc]
+      omega
+    rw [Finset.offDiag_card, hIcc] at hcard
+    have hle : B.card ≤ N := by simpa using Finset.card_le_card hsub
+    rcases Nat.eq_zero_or_pos B.card with h0 | hpos
+    · simp [h0]
+    have hkk : B.card ≤ B.card * B.card := Nat.le_mul_of_pos_left _ hpos
+    calc B.card ^ 2 = B.card * B.card := sq B.card
+      _ ≤ 2 * N + 1 + B.card := by omega
+      _ ≤ 4 * N := by omega
+  have h1 : ((maxSidonSubsetCard (Finset.Icc 1 N) : ℝ)) ^ 2 ≤ 4 * N := by exact_mod_cast key
+  have h2 : (2 : ℝ) * Real.sqrt N = Real.sqrt (4 * N) := by
+    rw [show (4 : ℝ) * N = 2 ^ 2 * N by ring, Real.sqrt_mul (by positivity),
+      Real.sqrt_sq (by norm_num)]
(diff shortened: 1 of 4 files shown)


FormalConjectures/ErdosProblems/44.lean
```diff
@@ -28,7 +28,8 @@ namespace Erdos44
 
 -- Reference: https://arxiv.org/pdf/2103.15850
 /-- The maximum size of a Sidon set in `{1, ..., N}` is less than or equal to `2 * √N`. -/
-@[category textbook, AMS 5 11]
+@[category textbook, AMS 5 11, formal_proof using formal_conjectures at
+  "https://github.com/christianTNG/formal-conjectures/blob/2d9d7f1380c4897f30004f6c909e42e820d3c65f/FormalConjectures/ErdosProblems/44.lean#L32"]
```
mo271: I think this proof is short enough (and the statement interesting enough to add th eproof here instead o f using a `formal_proof` link. Potentially this could even be moved to the definitions  of SidonSet `FormalConjecturesForMathlib`  at some point, it seems like a statement that could be in `Mathlib`?

FormalConjectures/ErdosProblems/649.lean
```diff
@@ -49,7 +49,8 @@ theorem erdos_649 : answer(False) ↔
 In fact, the answer to this question as written is easily seen to be no, since there are no
 solutions to $2^k\equiv -1\pmod{7}$, and hence this fails with $p=2$ and $q=7$.
 -/
-@[category textbook, AMS 11]
+@[category textbook, AMS 11, formal_proof using formal_conjectures at
+  "https://github.com/christianTNG/formal-conjectures/blob/35efde29598fd5104e5bc91f1e53d77931e7f856/FormalConjectures/ErdosProblems/649.lean#L53"]
```
mo271: please inline the proof, looks short and clean enough.

FormalConjectures/ErdosProblems/786.lean
```diff
@@ -66,7 +66,8 @@ theorem erdos_786.parts.ii : answer(sorry) ↔
 /--
 An example of such a set with density $\frac 1 4$ is given by the integers $\equiv 2\pmod{4}$
 -/
-@[category textbook, AMS 11]
+@[category textbook, AMS 11, formal_proof using formal_conjectures at
+  "https://github.com/christianTNG/formal-conjectures/blob/6ce00ce5ec1d12b91ffacc369e03536427e556b7/FormalConjectures/ErdosProblems/786.lean#L70"]
```
mo271: even though we are here about the line limit, I also propose to add the full proof here

FormalConjectures/GreensOpenProblems/72.lean
```diff
@@ -45,7 +45,8 @@ noncomputable def AllowedSetSize (k : ℕ) (N : ℕ) : ℕ :=
 
 /-- By the pigeon hole principle, the size of a subset of an $N \times N$ grid such that no $k$
 points lie on a line is bounded by $\leq (k - 1) * N$ for $N \geq k$. -/
-@[category textbook, AMS 5 52]
+@[category textbook, AMS 5 52, formal_proof using formal_conjectures at
+  "https://github.com/christianTNG/formal-conjectures/blob/d5aaa2cb443e948637e695c5810fb70e93a9cbcd/FormalConjectures/GreensOpenProblems/72.lean#L49"]
```
mo271: please inline the proof -- it is only 20 lines..

mo271 (commented):
Thanks, I looked at all four proofs and I think it would be great to inline all of them in this case.
Generally I think for `textbook` proofs should generally be achievable within the line limit, especially because they are often potentially good material for `FormalConjecturesForMathlib`

christianTNG:
Thanks a lot for the very quick review, much appreciated.


I have followed your suggestion and inlined all four proofs: the `formal_proof` links are gone again. Full `lake --wfail build` passes with no errors and no warnings, and `#print axioms` reports only `propext`, `Classical.choice` and `Quot.sound` for each of the four theorems.

FormalConjectures/ErdosProblems/786.lean
christianTNG: Inlined, thanks for waiving the line limit here.

FormalConjectures/ErdosProblems/44.lean
christianTNG: Inlined. I agree it looks like a natural fit next to the `SidonSet` definitions in `FormalConjecturesForMathlib`. Happy to do that move in a separate PR if you want it.

FormalConjectures/GreensOpenProblems/72.lean
christianTNG: Inlined. Note this forced `h` to `_h`, since the proof does not use `k ≤ N`.

FormalConjectures/ErdosProblems/649.lean
christianTNG: Inlined.

FormalConjectures/GreensOpenProblems/72.lean
```diff
@@ -46,9 +46,27 @@ noncomputable def AllowedSetSize (k : ℕ) (N : ℕ) : ℕ :=
 /-- By the pigeon hole principle, the size of a subset of an $N \times N$ grid such that no $k$
 points lie on a line is bounded by $\leq (k - 1) * N$ for $N \geq k$. -/
 @[category textbook, AMS 5 52]
-theorem allowedSetSize_le {k : ℕ} {N : ℕ} (h : k ≤ N) :
+theorem allowedSetSize_le {k : ℕ} {N : ℕ} (_h : k ≤ N) :
```
mo271:
```suggestion
theorem allowedSetSize_le {k : ℕ} {N : ℕ} :
```
Seems like we don't need that hypothesis for the proof? If so -- let's remove it!

FormalConjectures/GreensOpenProblems/72.lean
```diff
@@ -46,9 +46,27 @@ noncomputable def AllowedSetSize (k : ℕ) (N : ℕ) : ℕ :=
 /-- By the pigeon hole principle, the size of a subset of an $N \times N$ grid such that no $k$
 points lie on a line is bounded by $\leq (k - 1) * N$ for $N \geq k$. -/
```
mo271:
```suggestion
points lie on a line is bounded by $\leq (k - 1) * N$. -/
```
Also from the doctring

mo271 (commented):
Thanks, looks great now, just one suggestion to make one theorem stronger by removing the unused hypothesis

FormalConjectures/GreensOpenProblems/72.lean
christianTNG: Good catch, applied both. The hypothesis was never needed for this bound, so removing it is a strict improvement and the `_h` binder is gone. Rationale is in the commit message.

christianTNG:
Thanks again for the quick turnaround. Hypothesis removed and docstring adjusted, `lake --wfail build` still clean.

christianTNG:
-awaiting-author

mo271 (approved):
Thanks, LGTM

Issue #4881, opened 2026-08-11
`formal_proof` does not record what the linked proof assumes


williamjblair:
## Problem

The `formal_proof` attribute makes a claim about a file in another repository. Nothing checks that file. A proof that assumes a published theorem as an `axiom` therefore looks the same as a complete one, both to a reader and to `check_erdos_status.py`.

#4368 added `conditional formal_proof ... assuming <decl>` for this case. No file uses it yet.

## What is in the repo today

I fetched every external `formal_proof` target and read it. There are 338 attributes. 87 use the `formal_conjectures` kind. 251 point outside this repository. 237 of those name a file, and all 237 fetched.

Three targets have no `sorry` but assume a published theorem as an axiom:

| Problem | Axiom | Target |
| --- | --- | --- |
| 427 | `shiu_consecutive_primes` | [gist](https://gist.githubusercontent.com/JohnEdwardJennings/e2c6ef0daab55857b7cc9d340de7af84/raw/8ff97800e38582c71246a238e7541a9d69488cbd/Erdos427.lean) |
| 750 | `stiebitz_lower_bound` | [Shashi456/erdos-formalizations](https://github.com/Shashi456/erdos-formalizations/blob/main/Erdos/P750/Proof.lean) |
| 1141 | `theorem_1_3`, `mertens_third_theorem` | [yuta0x89/ErdosProblems](https://github.com/yuta0x89/ErdosProblems/blob/a1319f732cdee5140faf47d984e2c451c1184803/Erdos1141.lean) |

Each axiom is used in the proof of the linked result. The docstring on 1141 already says this: "Oriike [Or26] formalised the deduction in Lean."

This count is a lower bound. I read the linked file only. A linked file can import other files in the same repository that hold axioms of their own.

## Targets that no check can read

18 links do not name a file: 5 erdosproblems forum threads, 5 commit or pull-request-commit pages, 3 repository roots, 2 Lean web editor URLs, 2 gist landing pages, and 1 file on another host. A reader must find the proof without help, and a checker cannot look at all.

I said 13 here at first. That count missed the commit pages, four of which point into this repository's own history.

## What to do

1. Annotate the three problems above as `conditional formal_proof ... assuming <decl>`. Each assumed statement must be a declaration in the same file. 427 already has one, `erdos_427.variants.shiu`. 750 and 1141 need one added.
[...]

PR #4883, opened 2026-08-11
ErdosProblems/316,399: use the `formal_conjectures` kind for in-repo proofs

### LINKED ISSUE #4882: 316 and 399 link to themselves instead of using the `formal_conjectures` kind
## Problem

Erdős 316 and 399 are proved in this repository. The attribute on each says the proof is elsewhere, and links to the file's own URL on GitHub.

`FormalConjectures/ErdosProblems/316.lean:40`
```
@[category research solved, AMS 5 11, formal_proof using lean4 at "https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/316.lean"]
```

`FormalConjectures/ErdosProblems/399.lean:51`
```
@[category research solved, AMS 11, formal_proof using lean4 at "https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/399.lean"]
```

`erdos_316` is proved at 316.lean:45-53 with `decide +kernel`. `erdos_399` is proved at 399.lean:54-56. Neither proof has a `sorry`. The `sorry`s in the two files are on variants, which is normal.

## Why this matters

- The link points at the declaration it is attached to, so it tells a reader nothing.
- It uses the mutable `main` ref, so it will drift as the file changes.
- It counts as an external proof. #4749 checks it as one, and any later check on target content will read the whole file, including the variants.

## Fix

Use the kind for a proof that lives here:

```
formal_proof using formal_conjectures at ""
```

The repository already writes it this way in two other places. The docstring credits stay as they are.

## Related

#4881 covers links whose target assumes an axiom. That is a different defect and needs new mechanism. This one does not.

williamjblair:
Closes #4882.

Erdős 316 and 399 are proved in this repository. The attribute on each said the proof was elsewhere, and linked to the file's own URL on GitHub. Both now use the kind for a proof that lives here.

`erdos_316` is proved at 316.lean:45-53 with `decide +kernel`. `erdos_399` is proved at 399.lean:54-56. Neither proof has a `sorry`. The `sorry`s in the two files are on variants, which is normal.

### Why the old form was wrong

- The link pointed at the declaration it was attached to, so it told a reader nothing.
- It used the mutable `main` ref, so it would drift as the file changed.
- It counted as an external proof. #4749 checks it as one.

### Checks I ran

`lake build FormalConjectures.ErdosProblems.«316» FormalConjectures.ErdosProblems.«399»` completes with no error and no warning. The empty link is valid for the `formal_conjectures` kind, where the proof is the statement itself, and the repository already writes it this way in two other places.

### Not in this change

#4881 covers links whose target assumes an axiom. That defect needs new mechanism, and 316 and 399 are not examples of it.

### DIFF
--- a/FormalConjectures/ErdosProblems/316.lean
+++ b/FormalConjectures/ErdosProblems/316.lean
@@ -37,7 +37,7 @@ The minimal counterexample is $\{2,3,4,5,6,7,10,11,13,14,15\}$, found by Tom Sto
 
 This was formalized in Lean by Mehta.
 -/
-@[category research solved, AMS 5 11, formal_proof using lean4 at "https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/316.lean"]
+@[category research solved, AMS 5 11, formal_proof using formal_conjectures at ""]
 theorem erdos_316 : answer(False) ↔ ∀ A : Finset ℕ, 0 ∉ A → 1 ∉ A →
     ∑ n ∈ A, (1 / n : ℚ) < 2 → ∃ (A₁ A₂ : Finset ℕ),
       Disjoint A₁ A₂ ∧ A = A₁ ∪ A₂ ∧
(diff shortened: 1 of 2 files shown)


mo271 (approved):
Thanks for the cleanup! 
I created #4889 to make sure we catch those before they are added to the repo.

PR #4884, opened 2026-08-11
ErdosProblems/427: mark the linked proof conditional on Shiu's theorem

williamjblair:
**Program lane:** Foundation

## What this establishes

The proof linked from `erdos_427` declares Shiu's theorem as an axiom, so the annotation must not claim an unconditional complete proof. This change uses the conditional `formal_proof` mechanism from #4368 and points its `assuming` clause at the existing `erdos_427.variants.shiu` statement. The assumption declarations move only because the name must resolve during elaboration.

The linked gist's axiom and the named assumption agree up to notation. The focused module build is clean. This case also remains the negative calibration fixture: the retained Comparator run reported the same external Shiu axiom declared by the PR.

## Dependency and sequence

This is the first real proof-specific condition for schema 2 in #4894 and the calibration datum used by the verification lane in [#4394](https://github.com/google-deepmind/formal-conjectures/issues/4394). Status policy belongs to #4828; other conditional annotations are separate changes.

## Review question

Does the conditional annotation faithfully describe the linked proof's exact external assumption, without changing the mathematical statement?

## Explicit boundary

This PR does not prove Shiu's theorem, make the proof unconditional, decide status policy, or turn the retained Comparator result into maintainer acceptance. It introduces no Vela authority or dependency.

### DIFF
--- a/FormalConjectures/ErdosProblems/427.lean
+++ b/FormalConjectures/ErdosProblems/427.lean
@@ -38,17 +38,6 @@ def erdos427 : Prop := ∀ (n d : ℕ),
     d ∣ ∑ i ∈ Finset.Ico n (n + k), i.nth Nat.Prime
 
 /--
-**Erdős Problem 427**: is it true that, for every $n$ and $d$, there exists $k$ such that
-$$
-  d \mid p_{n + 1} + \cdots + p_{n + k},
-$$
-where $p_r$ denotes the $r$th prime?
--/
-@[category research solved, AMS 11, formal_proof using lean4 at "https://gist.githubusercontent.com/JohnEdwardJennings/e2c6ef0daab55857b7cc9d340de7af84/raw/8ff97800e38582c71246a238e7541a9d69488cbd/Erdos427.lean"]
-theorem erdos_427 : answer(True) ↔ erdos427 := by
-  sorry
-
-/--
 The statement of Shiu's theorem:
 for any $k \geq 1$ and $(a, q) = 1$ there exist infinitely many $k$-tuples of consecutive primes
 $p_m, \dots, p_{m + k - 1}$ all of which are congruent to $a$ modulo $q$.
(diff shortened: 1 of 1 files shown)


williamjblair:
This PR's claim, checked mechanically. The full review loop from #4394 ran once against this exact PR: the `erdos_427` workspace was generated (#4951), the linked gist was fetched at its pinned revision and bridged into `Submission/` (`Finset.sum_Ico_eq_sum_range` and one `Nat.add_sub_cancel_left`; the gist's statement matches this file's index convention exactly), and comparator, sandboxed under landrun, replayed the result:

```
uncaught exception: Illegal axiom detected: 'External.shiu_consecutive_primes'
```

That is this PR's thesis, derived by machinery from the artifacts: the proof is complete except for Shiu's theorem, so `conditional formal_proof ... assuming` is the right annotation.

```json
{
  "schemaVersion": 1,
  "pr": 4884,
  "declaration": "Erdos427.erdos_427",
  "proof": "gist JohnEdwardJennings/e2c6ef…@8ff97800",
  "checks": {
    "generate": "pass",
    "workspaceBuild": "pass",
    "comparator": "fail: illegal axiom External.shiu_consecutive_primes"
  }
}
```

Two facts the run established beyond this PR. The gist compiles under v4.27 and its final theorem instantiates this file's statement with a two-line bridge, so the link is sound apart from the axiom. And the whole loop, PR to report, runs on stock CI: [run](https://github.com/williamjblair/formal-conjectures/actions/runs/31862100273), [workflow](https://github.com/williamjblair/formal-conjectures/blob/review-loop/.github/workflows/review-loop.yml). It lives on a fork branch because where such checks belong in this repository's CI is #4930's question.

mo271 (approved):
Thanks, LGTM!

Jayyhk:
An unconditional single-file proof can be found [here](https://github.com/Jayyhk/erdos-lean/blob/main/problems/427/Erdos427.lean). Big thanks to Boris for formalizing the original axiom!

PR #4885, opened 2026-08-11
ErdosProblems/750: mark the linked proof conditional on Stiebitz's theorem

williamjblair:
Part of #4881. Follows #4884, which did the same for 427.

The proof linked from `erdos_750` is not complete on its own. It declares Stiebitz's theorem as an axiom and derives the result from it:

```lean
axiom stiebitz_lower_bound : ∀ {V : Type u} (G : SimpleGraph V) (r : ℕ),
    IsRecursivelyBuiltMr r G → (r : ℕ∞) ≤ G.chromaticNumber
```

The attribute now says `conditional formal_proof ... assuming erdos_750.variants.stiebitz`.

### Why this is more than one line

427 already stated its assumption. This one did not, and the `assuming` clause must name a declaration. Stating Stiebitz's theorem needs the generalised Mycielskian, which Mathlib does not have in any form. The file therefore gains:

- `MycVerts`, the vertices of `Mₛ(G)`: `s` levels and one apex.
- `MycAdj`, the adjacency relation.
- `genMyc`, the graph itself, as `SimpleGraph.fromRel (MycAdj s G)`.
- `IsRecursivelyBuiltMr`, the class `Mᵣ` of graphs built from `K₂` by iterated generalised Mycielskians.
- `erdos_750.variants.stiebitz`, the statement, with a `sorry` proof.

Stiebitz's theorem is about that class and not about arbitrary graphs. `χ(Mₛ(H)) = χ(H) + 1` is false in general, and the docstring says so.

### Why `fromRel`

`MycAdj` is already symmetric and irreflexive, so `SimpleGraph.fromRel` adds nothing to it: the `v ≠ w` conjunct follows from the relation, and the `r v w ∨ r w v` disjunction collapses. Using it avoids two proof obligations that carry no mathematical content.

Two `test` statements check that this packaging keeps the edges it should:

- `genMyc_apex_adj_top`: the apex is adjacent to every top-level vertex.
- `genMyc_adj_level_zero`: two level-`0` vertices are adjacent when the underlying vertices are.

Both are proved, not `sorry`.

### Checks I ran

`lake build FormalConjectures.ErdosProblems.«750»` completes with no error and no warning, including the two `test` proofs.

### Still open in #4881

1141 has the same defect. It assumes Pollack's Theorem 1.3 and Mertens' third theorem. `check_erdos_status.py` still counts a conditional proof as formally solved; that fix belongs on #4828, which reads the published extract and can use the `proofConditions` field directly.

### DIFF
--- a/FormalConjectures/ErdosProblems/750.lean
+++ b/FormalConjectures/ErdosProblems/750.lean
@@ -31,11 +31,88 @@ import FormalConjecturesUtil
   (N.S.) (1995), 61-65.
 - [ErHa67b] Erdős, P. and Hajnal, András, On chromatic graphs. Mat. Lapok (1967), 1--4.
 - [UlamErdos750](https://www.ulam.ai/research/erdos750.pdf)
+- [St85] Stiebitz, M., _Beiträge zur Theorie der färbungskritischen Graphen_. Habilitation,
+  TH Ilmenau (1985).
+- [SaSt89] Sachs, H. and Stiebitz, M., _On constructive methods in the theory of colour-critical
+  graphs_. Discrete Math. (1989), 287-296.
+- [MuSt19] Müller, T. and Stehlík, M., _Generalised Mycielski graphs and the Borsuk-Ulam theorem_.
+  Electron. J. Combin. (2019), P4.8.
 -/
 
-open Filter Finset NNReal
+open Filter Finset NNReal SimpleGraph
+
+universe u
 
 namespace Erdos750
+
+/--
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

Jayyhk:
An unconditional single-file proof can be found [here](https://github.com/Jayyhk/erdos-lean/blob/main/problems/750/Erdos750.lean). Big thanks to Boris for formalizing the original axiom!

PR #4886, opened 2026-08-11
ErdosProblems/1141: mark the linked proof conditional on its two external inputs

williamjblair:
Part of #4881. Follows #4884 (427) and #4885 (750). This completes step 1 of that issue.

The proof linked from `erdos_1141` is the deduction and not the whole result, which the docstring already said: "Oriike [Or26] formalised the deduction in Lean." It declares two axioms:

```lean
axiom theorem_1_3 (ε A : ℝ) (hε : 0 < ε) (hA : 0 < A) :
    ∃ m0 : ℕ, ∀ m : ℕ, m > m0 → ∀ χ : DirichletCharacter ℂ m, MulChar.IsQuadratic χ →
      Real.rpow (Real.log (m : ℝ)) A ≤ ((residuePrimesUpTo m χ ε).card : ℝ)

axiom mertens_third_theorem (n : ℕ) (hn : 3 ≤ n) :
    1 / (3 * Real.log n) ≤ ∏ p ∈ (Finset.range (n + 1)).filter Prime, (1 - 1 / (p : ℝ))
```

The attribute now says `conditional formal_proof ... assuming erdos_1141.variants.pollack_1_3 erdos_1141.variants.mertens_third`.

### What the file gains

Neither assumption was stated, so both are added with `sorry` proofs, together with the two definitions Theorem 1.3 needs: `residuePrimeUpperBound` for the cutoff $m^{1/4+\varepsilon}$, and `residuePrimesUpTo` for the primes below it on which the character is $1$. `DirichletCharacter` and `MulChar.IsQuadratic` come from Mathlib, so nothing else had to be built.

### Where the wording differs from the axioms

- The axiom writes `Real.rpow x y` and this writes `x ^ y`. The exponent is real, so the power is `Real.rpow` either way.
- The axiom filters `Finset.range (n + 1)` by `_root_.Prime`, since its file opens `Finset Real` and not `Nat`. This filters by `Nat.Prime`, which is the same predicate on `ℕ`, so the two products are equal.
- `m > m0` is written `m₀ < m`.

Nothing else moved. The bound in the Mertens statement is the weakened one the proof assumes, not the theorem: the true constant is $e^{-\gamma} \approx 0.561$ and the assumed constant is $1/3$. The docstring says so, since a reader who knows Mertens would otherwise wonder why the statement is loose.

### Checks I ran

`lake build FormalConjectures.ErdosProblems.«1141»` completes with no error and no warning. It first warned about a second module docstring, which is now a plain comment.

I did not add `test` statements here. #4885 needed them because it repackaged the adjacency relation through `SimpleGraph.fromRel`; these two statements are transcriptions with no such change.

### What remains in #4881

`check_erdos_status.py` still counts a conditional proof as formally solved. That fix belongs on #4828, which reads the published extract and can use the `proofConditions` field. Separately, 13 `formal_proof` links name no file, so no check can read them.

### DIFF
--- a/FormalConjectures/ErdosProblems/1141.lean
+++ b/FormalConjectures/ErdosProblems/1141.lean
@@ -28,6 +28,8 @@ import FormalConjecturesUtil
 - [Or26] Y. Oriike, [Lean formalisation of Erdős problem 1141](https://github.com/yuta0x89/ErdosProblems/blob/a1319f732cdee5140faf47d984e2c451c1184803/Erdos1141.lean) (2026)
 - [Po17] P. Pollack, Bounds for the first several prime character nonresidues. Proc. Amer. Math. Soc.
   (2017), 2815--2826.
+- [Me1874] F. Mertens, Ein Beitrag zur analytischen Zahlentheorie. J. Reine Angew. Math. (1874),
+  46--62.
 - [Va99] Various, Some of Paul's favorite problems. Booklet produced for the conference "Paul Erdős
   and his mathematics", Budapest, July 1999 (1999).
 -/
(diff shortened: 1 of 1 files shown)


yuta0x89:
Author of the linked file here — thank you for the careful reading, and for #4881 more broadly.

I can confirm the annotation is accurate. The linked proof is the deduction only: it declares exactly the two axioms named here (`Pollack17.theorem_1_3` and `mertens_third_theorem`; see the `#print axioms` output recorded at the end of the file). The transcriptions of `residuePrimeUpperBound`, `residuePrimesUpTo`, and the two statements match my file up to the notational differences already listed in the PR description, each of which gives the same statement. The weakened Mertens constant `1/3` is also exactly the form my proof assumes (following the same axiomatised form as Pietro Monticone's `Erdos237.lean`), so stating the weak form rather than the true `e^{-γ}` asymptotic is the right call.

`conditional formal_proof ... assuming` is precisely the status I intended the file to communicate, so this change looks correct to me.

williamjblair:
Appreciate it @yuta0x89, thank you!

mo271 (approved):
Thanks, LGTM!

Jayyhk:
An unconditional single-file proof can be found [here](https://github.com/Jayyhk/erdos-lean/blob/main/problems/1141/Erdos1141.lean). Big thanks to Boris for formalizing the original axioms!

PR #4895, opened 2026-08-12
Point `formal_proof` links at files rather than commit and landing pages

williamjblair:
Make **11 proof locators across 10 files** point to an exact file and immutable revision.

| Before | After |
| --- | --- |
| Commit or PR-commit page | Pinned file and relevant line |
| Gist landing page or Lean playground | Pinned raw Lean file |
| Repository root | Pinned proof file |

For ClaudesCycles, the link selects `Solution.lean`: `Challenge.lean` contains the intentional placeholder and uses a product triple, whereas FC uses `Fin 3 → ZMod m`.

**Rebased on 16 September 2026.** `main` now pins the Boxdot Conjecture link to a file itself, so that change is dropped from this PR and `main`'s version is kept.

**Validation:** earlier focused Lean builds passed. At the rebased head `afe2e941`, all 11 replacement URLs return HTTP 200. Reachability does not establish proof correctness or a verified cross-repository bridge.

This correction is independently mergeable. It has no metadata, CLI or verification prerequisite. Related assumption/status work remains in #4881; roadmap #4394.


williamjblair:
Also @mo271 one question about the five forum links that I did not change.

Three of them hold the proof in a `live.lean-lang.org/#codez=...` link. That URL compresses the source into its fragment. No file exists, so a checker has nothing to read. These are 303, 1051, and 434, which has the link twice.

38 is different. Its docstring names `github.com/spicylemonade/erdos-38`, so a file exists there.

For the three, which do you prefer?

1. Keep the thread link.
2. Use the `#codez` URL. It holds the proof, but nobody can read it.
3. Ask the author for a gist.

I'm leaning towards option 1 but your call, since it is about what `formal_proof` promises

williamjblair:
Rebased onto `main`. `main` had already pinned the Boxdot Conjecture link to a file, so I dropped that change and kept `main`'s version. That leaves 11 links in 10 files, and all 11 return HTTP 200 at `afe2e941`. The PR is mergeable again.

@mo271 the question above about the three `#codez` forum links (303, 1051, 434) is still open. I'd still keep the thread links unless you prefer another option.

mo271 (approved):
Thanks, LGTM!

PR #4900, opened 2026-08-12
feat: formalise construction of the 668 Hadamard matrix example

Paul-Lez:
WIP.


FormalConjectures/Wikipedia/Hadamard.lean
```diff
@@ -123,11 +132,35 @@ theorem HadamardConjecture.variants.first_cases (k : ℕ) (h : k ≤ 166) :
     ∃ M, IsHadamard (n := 4 * k) M := by
   sorry
 
+/-- The order-668 integer witness, cast entrywise to the real numbers. -/
+def H668 : Matrix (Fin 668) (Fin 668) ℝ := fun i j => H668Int i j
```
eric-wieser:
```suggestion
def H668 : Matrix (Fin 668) (Fin 668) ℝ := .of fun i j => H668Int i j
```

FormalConjectures/Wikipedia/Hadamard.lean
```diff
@@ -15,13 +15,16 @@ limitations under the License.
 -/
 
 import FormalConjecturesUtil
+import FormalConjectures.Wikipedia.Hadamard668
```
eric-wieser, line 18: I think we don't want this committed on master, to avoid big slow dependencies needed to state the problems. It's fine of course to leave this in a side-branch and reference it.

Paul-Lez: yeah that's the plan! the goal was just to have a PR to reference to people (and then repurpose the PR to link the proof via a `formal_proof` attribute)

FormalConjectures/Wikipedia/Hadamard668Defs.lean
```diff
@@ -0,0 +1,150 @@
...
+
+/-- The $4\times4$ sign pattern $Z$ for the bottom-left block. -/
+def Z : Matrix Q Q ℤ :=
+  !![-1, -1, -1, 1;
+     -1, -1, 1, -1;
+     -1, 1, -1, -1;
+     1, -1, -1, -1]
+
+/-- Repeat each entry of `Y` across a block of 166 columns. -/
+def Y_tilde : Matrix Q (Q × C) ℤ := fun i j => Y i j.1
```
eric-wieser, line 134: perhaps:
```suggestion
/-- Repeat each entry of `Y` across a block of 166 columns. -/
def Y_tilde : Matrix Q (Q × C) ℤ := Y.submatrix id Prod.fst
```

mo271:
perhaps one could factor out the part marking this problem as solved from this longer pr? 
@Paul-Lez

Deicyde (commented):
I reconstructed the 668×668 sign matrix from the submitted bit sequences and checked every row pair with exact integer arithmetic: its Gram matrix is 668I. The cast-to-real argument uses the proved forward Hadamard implication, so the unrelated unfinished reverse implication does not invalidate the new result. CI passes; I did not replay Lean locally, and the agreed plan to retain this long construction externally still applies.

PR #4928, opened 2026-08-13
feat(site): move WIP banner to comments and votes section

mo271:


mo271:
https://mo271.github.io/formal-conjectures//about/
and 
https://mo271.github.io/formal-conjectures/
now has a preview of how it will look like after this change

danielchin (approved):

Paul-Lez (approved):
Yeah that's something we really needed to do!

Issue #4896, opened 2026-08-12
Tracking: possible misformalizations found in statement audits


KitaKen1:
## Purpose

This issue tracks possible mismatches between Formal Conjectures statements
and their cited sources, together with degenerate boundary cases and a few
`answer`-semantics questions.

The initial audit was discussed on Zulip:
https://leanprover.zulipchat.com/#narrow/channel/524981-Formal-conjectures/topic/Possible.20misformalizations/with/616080167

These are triage leads, not claims that every item is a confirmed bug. For
each item, the goal is to confirm the intended statement, link any dedicated
issue/PR, and check it off here once resolved.

## Source / statement mismatches

- [ ] Optimization Constant C1a — integration interval
- [ ] Erdős 996 — Fourier coefficient index and tail norm
- [ ] Erdős 697 — reversed thresholds and the `m = 0` case
- [ ] Erdős 887 — `answer` scope and the coefficient in `rosenfeld_4`
- [ ] Erdős 918 — fixed `ω` versus a universal ordinal binder
- [ ] Erdős 1167 — omitted `κ α > r` condition
- [ ] Erdős 757 — `= 11` versus `11 ≤`
- [ ] Green 72 — apparent reversal of the intended conclusion

## Boundary cases / vacuous wrappers

- [x] Erdős 940 `large_integers` — range starts at `r = 2` — fixed by [PR #4933](https://github.com/google-deepmind/formal-conjectures/pull/4933)
- [ ] Erdős 694 — empty totient fiber makes the wrapper vacuous
- [ ] Erdős 939 — positivity of summands is missing
- [ ] Green 21 `fox_kleitman_modular` — `k = 0`

## Interpretation / `answer` semantics

- [ ] Erdős 357 weak-monotone variant — treatment of repeated values
- [ ] Self-answer / classical-choice patterns — Erdős 195, 188, 1047

## Existing tracking

- Erdős 80: #4867
- Erdős 319: #4892
- Erdős 769: #4806

## Open-PR blockers

- [ ] PR #4660 — `tsum` without `Summable` / `HasSum`
- [ ] PR #4866 — missing nondegeneracy / lattice hypotheses

[...]

PR #4933, opened 2026-08-13
ErdosProblems/940: restrict large_integers to r >= 3

williamjblair:
Fixes an item on #4896, reported by @KitaKen1.

`erdos_940.variants.large_integers` quantified over `r ≥ 2`. The source opens "Let $r\geq 3$", and the whole problem sits under that hypothesis.

At `r = 2` the statement says that almost every integer is a sum of at most two `2`-powerful numbers, so that set is cofinite. `erdos_940.variants.two` is in the same file, is categorised `research solved`, and says the same set has density `0`. `Set.HasDensity` is a `Tendsto`, so the density is unique. Both cannot hold, and so the `answer` slot on a `research open` statement was already decided, for a reason the source excludes.

`erdos_940` itself has `∀ r ≥ 3` and says "Let $r \ge 3$." in its docstring. Only the variant drifted.

The docstring also omitted the bound, so it now states it.

The source treats `r = 2` separately and in the opposite direction: Erdős [Er76d] called the density-zero claim "easy" for `r = 2`, and Baker and Brüdern [BaBr94] proved it.

### Checks I ran

`lake --wfail build 'FormalConjectures.ErdosProblems.«940»'` passes.

I also checked the `r = 2` case numerically before trusting the internal contradiction. Enumerating the `2`-full numbers below `10^7` and taking all pairwise sums, 4212552 integers below `10^7` are not a sum of at most two of them, and the representable proportion falls as the range grows: 0.5964 at `10^3`, 0.5842 at `10^6`, 0.5787 at `10^7`. So the set is not cofinite, independently of `variants.two`.

### DIFF
--- a/FormalConjectures/ErdosProblems/940.lean
+++ b/FormalConjectures/ErdosProblems/940.lean
@@ -56,12 +56,13 @@ theorem erdos_940.variants.three_cubes :
 
 
 /--
-It is not known if all large integers are the sum of at most $r$-many $r$-powerful numbers.
+Let $r \ge 3$. It is not known if all large integers are the sum of at most $r$-many
+$r$-powerful numbers.
 -/
 @[category research open, AMS 11]
 theorem erdos_940.variants.large_integers :
     answer(sorry) ↔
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, good catch, also @KitaKen1

PR #4934, opened 2026-08-13
ErdosProblems/939: positivity, the answered triples question, and a misquoted identity

williamjblair:
Three corrections to `939.lean`, each against [erdosproblems.com/939](https://www.erdosproblems.com/939).

**Positivity.** `Erdos939Sums` did not require positive summands, and `0` is vacuously `r`-powerful (`primeFactors 0 = ∅`). Witness: `{0, 1} ∈ Erdos939Sums 4`, which makes `variants.examples`, there to record Cambie's `r = 5` example, provable from `{0, 1}` alone. Fix: `0 < s` on the summand condition. The main statement was not vacuous; the witness only settles `r = 4`.

**The triples question is answered.** `variants.triples` was `research open`; the source cites Nitaj [Ni95] proving infinitely many coprime 3-powerful triples with `a + b = c`. Now `research solved`, `answer(True)`. I verified the source's example (`756249048 + 19902511 = 919^3`) and the [Ni95] reference via Crossref.

**A misquoted identity.** The docstring read `$3761^5$` where the source has `3^7 · 61^5`; the equation only balances the second way. Docstring-only; the Lean never mentions that side.

The Euler counterexample also requires positive summands, so a zero summand cannot make it a claim about fewer powers. Its proof uses current Mathlib’s `push Not` syntax.

Validation on the updated head: `lake --wfail build 'FormalConjectures.ErdosProblems.«939»'`. The changed module builds without warnings. This mathematical correction is independently mergeable.

### DIFF
--- a/FormalConjectures/ErdosProblems/939.lean
+++ b/FormalConjectures/ErdosProblems/939.lean
@@ -19,7 +19,10 @@ import FormalConjecturesUtil
 /-!
 # Erdős Problem 939
 
-*Reference:* [erdosproblems.com/939](https://www.erdosproblems.com/939)
+*References:*
+- [erdosproblems.com/939](https://www.erdosproblems.com/939)
+- [Ni95] Nitaj, A., _On a conjecture of Erdős on 3-powerful numbers_. Bull. London Math. Soc.
+  (1995), 317-318.
 -/
 open Nat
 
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4935, opened 2026-08-13
Paper/ClaudesCycles: the even case is settled in the revised source

williamjblair:
`cube_hamiltonian_arc_decomposition_even` was `research open` with `answer(sorry)`, and the module docstring said the even case remains open. Knuth's note, revised 14 April 2026, says the opposite on page 6: "**Breaking news:** The problem for even values of m is no longer in doubt!" The file merged five weeks before the revision, so it was faithful when written. Now `research solved`, `answer(True)`.

Checked: Knuth's construction satisfies `HasHamiltonianArcDecomposition` as written at `m = 3, 5, 7, 9`; exhaustive search at `m = 2` finds none, so the impossibility half is not vacuous; even-case decompositions verified at `m = 4, 6`.

Also four proved API/test statements for the file's five bespoke definitions, which had none (CONTRIBUTING asks for them). `#print axioms` gives the three standard axioms on each. `not_hasHamiltonianArcDecomposition_one` shows the `1 < m` hypothesis is load-bearing.


mo271 (approved):
Thanks, LGTM!

PR #4936, opened 2026-08-13
check_erdos_status: add a --problem filter

williamjblair:
`check_erdos_status.py` prints every mismatch in the repository. A reviewer working on one problem has to read a list of 22 and satisfy themselves that their number is absent.

`--problem N` filters to that problem, and says plainly when there is nothing:

```
$ python3 scripts/check_erdos_status.py --problem 80
Erdős problem 80: the repository and erdosproblems.com agree.
[]

$ python3 scripts/check_erdos_status.py --problem 90
[
  {
    "number": "90",
    "lean_status": "solved",
    "yaml_status": "formally solved"
  }
]
```

The exit code keeps its meaning: `0` when the named problem agrees, `1` when it does not. The message goes to stderr, so the JSON on stdout stays machine-readable.

Without the flag nothing changes, including `--create-issues`.

### Why

This came out of a review pass. Two independent reviewers ran the script on a file, saw a repository-wide list, and had to work out that the output was not about the file in front of them. One of them ran it on a `Paper/` module, where it is not applicable at all.

### Checks I ran

Four new tests, and the 27 in `scripts/` pass. Also ran it against live data for a problem that agrees and one that does not.

### DIFF
--- a/scripts/check_erdos_status.py
+++ b/scripts/check_erdos_status.py
@@ -7,6 +7,7 @@ on the main theorem in the corresponding .lean file.
 
 Usage:
   python check_erdos_status.py               # Print mismatches as JSON
+  python check_erdos_status.py --problem 80  # Only that problem
   python check_erdos_status.py --create-issues  # Also create GitHub issues
 """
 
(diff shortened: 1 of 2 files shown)


williamjblair:
Folding this into #4828 rather than carrying two pull requests against the same script. They conflicted: #4828 rewrites how the statuses are read and this changed the same import block in the tests.

#4828 now carries the `--problem` filter as a second commit, rebased on current `main`. All 43 script tests pass there, and the filter behaves the same against the rewritten script: `--problem 80` reports agreement and exits 0, `--problem 90` reports the mismatch and exits 1.

Closing. No work is lost.

PR #4937, opened 2026-08-13
scripts: print the LaTeX statement of an Erdős problem

williamjblair:
Reading a problem from erdosproblems.com is ad hoc today, and two failure modes recur.

**The site refuses `Python-urllib` with a 403.** It answers a request that identifies itself. A reviewer who hits that 403 tends to give up on the source and reconstruct the statement from a neighbouring problem file, which is how a wrong bound gets copied instead of caught.

**The rendered page runs terms together.** `3^7\cdot 61^5` reads as `3761^5`, and that misreading reached this repository. The `/latex/<n>` view keeps the source, so the reader sees the `\cdot`.

`scripts/erdos_problem.py N` prints the statement, the remarks, the cross-referenced problems and the reference list:

```
$ python3 scripts/erdos_problem.py 940
Let $r\geq 3$. A number $n$ is $r$-powerful if for every prime $p$ which divides $n$ ...

Erd\H{o}s \cite{Er76d} claims that the claim that the set has density $0$ is 'easy' for $r=2$
(a potential 'easy argument' is given in the comments by Tao; this was first proved in the
literature by Baker and Br\"{u}dern \cite{BaBr94}). ...

## References
[BaBr94] Baker, R. C. and Br\"udern, J., On sums of two squarefull numbers. ...

## See also
[941], [1081], [1107]
```

The remarks matter as much as the statement. erdosproblems.com records several answered questions only there, so a formalisation whose `category` disagrees with its source usually disagrees with a remark rather than with the problem text.

`check_erdos_status.py` reads `data/problems.yaml` from `teorth/erdosproblems`, which carries status and tags and no statement text. The two scripts answer different questions and neither replaces the other.

Twelve tests, all on saved markup, no network request. The user agent names the tool and the repository rather than imitating a browser.

### DIFF
--- a/scripts/erdos_problem.py
+++ b/scripts/erdos_problem.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python3
+# Copyright 2025 The Formal Conjectures Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Print the LaTeX statement of an Erdős problem.
+
+    python3 scripts/erdos_problem.py 940
+
+Reads `erdosproblems.com/latex/<n>`, which serves the LaTeX the site renders,
+and prints the statement, the remarks, the cross-referenced problems and the
+reference list.
+
+Prefer this to the rendered page. Rendering runs terms together, so that
+`3^7\\cdot 61^5` reads as `3761^5`, and that misreading has already reached this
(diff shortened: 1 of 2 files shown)


mo271 (commented):
Some very general comment: 

I am not convinced that this script and similar scripts are really so helpful: 
A current agentic system should be able to pull a website (and also the latex source) without needing an extra script, no?
Perhaps all it needs is the hint that the url is of the form `erdosproblems.com/latex/123` or whatever, and that should be sufficient?
That again could be put in ErdosProblems/README.md. 

Otherwise I guess we would need similar script how to access the OEIS (there is a github one can download, which is probably better than accessing their website every time?) or the Arvix (with a script for not only getting the pdf, but rather the (potentially zipped) .tex source,  and unpacks it, or ocr'ed the pdf etc etc, you see where I'm going with this: 

If think this is intented to assist and AI agent, then it is probably not necessary to have it, or pretty much instant for the agent to creat this script. If it is intented for human, then the question is why not directly access the website?

mo271:
But thanks for the work nonetheless!

williamjblair:
Yep agreed on all above, this came out of the testing from the new review skill - didnt actually mean to put this up. I agree there is a better way to do this, even its just improved guidance - will close this for now

PR #4940, opened 2026-08-13
ErdosProblems/918: `ω` is the fixed ordinal, not a variable

williamjblair:
Four `.ii` statements bind `ω` as a universally quantified variable. Their docstrings say $\aleph_{\omega+1}$ and $\aleph_\omega$. The file disagrees with itself.

The cause is a missing scope. This file has `open scoped Cardinal`. [623.lean](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/623.lean#L25) has `open scoped Cardinal Ordinal` and writes `ℵ_ ω` as a constant. Without the `Ordinal` scope there is no `ω` to refer to, so a binder was added to make the statements elaborate.

This PR opens `Ordinal` and drops the four binders. Verified: `example : (ω : Ordinal) = Ordinal.omega0 := rfl`, and neither statement takes an `Ordinal` argument now.

### One statement was false

`variants.eq_aleph_0.parts.ii` is `@[category textbook]`, so it asserts a provable statement. With the binder it had to hold at `ω = 0`, where it reads $\aleph_1$ and $\aleph_0$. The complete graph on $\aleph_1$ vertices witnesses the existential it denies: a proper colouring of a complete graph is injective, and every countable induced subgraph is complete on countably many vertices.

`variants.eq_aleph_0_all_subgraphs.parts.ii` survives its binder for an unrelated reason. `SimpleGraph.Subgraph` permits arbitrary `verts` with empty `Adj`, so an edgeless subgraph breaks the `∀ H` clause for every `G`.

### Two further commits

`[ErHa69]` appears in four docstrings. The reference list defines `[ErHa68b]` and `[Er69b]`, and the source cites `\cite{Er69b}`. Changed.

The same docstrings carry the source's justification, which the source states for general subgraphs only: "(assuming subgraph and not induced subgraph was intended)". The induced pair asserts it anyway. A note records the gap.

The note on `all_subgraphs.parts.i` said it is unclear whether the general-subgraph question is open. For `≤ ℵ₀` it is not a separate question: every subgraph on `W` sits inside the induced subgraph on `W`, and the top subgraph induced on `W` has `coe = G.induce W`.

### Open question

Both induced statements keep `@[category textbook]`. For `eq_aleph_0.parts.i` I could neither prove nor refute it, and the source does not back it. A maintainer's call.

### DIFF
--- a/FormalConjectures/ErdosProblems/918.lean
+++ b/FormalConjectures/ErdosProblems/918.lean
@@ -81,17 +81,22 @@ theorem erdos_918.variants.erdos_hajnal (k : ℕ) (hk : 0 < k) : ∃ (V : Type u
       ∀ (W : Set V) (_ : #W < ℵ_ k), (G.induce W).chromaticCardinal ≤ ℵ₀ := by
   sorry
 
-/-- In [ErHa69] the questions are stated with $= \aleph_0$ rather than $\leq\aleph_0$. This is
+/-- In [Er69b] the questions are stated with $= \aleph_0$ rather than $\leq\aleph_0$. This is
 a likely typo since it can be shown that no such graph exists in this case.
 
 This is the first question with induced subgraphs. -/
+-- Formalisation note: the source states the impossibility for general subgraphs, in a
+-- parenthetical, "assuming subgraph and not induced subgraph was intended". It does not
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4941, opened 2026-08-13
GreensOpenProblems/72: ask Green's question rather than assert its opposite

williamjblair:
`green_72` asserted `NoKInLineFor 3 N` for every `N ≥ 3`, that is `AllowedSetSize 3 N = 2 * N`. Green asks the opposite:

> for $N$ sufficiently large is it impossible to have a set of size $2N$ with this property?

His comments add that $2N$ configurations exist only "for $N$ up to around 50", that $(3/2 + o(1))N$ is achievable for arbitrary $N$, and that "my personal suspicion is that this is optimal". The file's other cited reference agrees: Guy and Kelly conjectured $c \approx 1.874$, and Guy later corrected it to $c \approx 1.814$. Both are below 2.

So the file asserted, as a `research open` theorem, what both its cited sources expect to fail. It now asks the source's question:

```lean
theorem green_72 : answer(sorry) ↔ ∀ᶠ N in Filter.atTop, ¬ NoKInLineFor 3 N
```

**This also fixes a decided answer slot.** As a bare theorem, `green_72` proved the right-hand side of `green_72.variants.eventually`, so a `research open` slot was settled by a neighbour. As a slot it settles nothing. The two are not negations, since negating an `∀ᶠ` gives an `∃ᶠ`, so both stay open, and the variant's docstring now says how they differ.

**The module docstring** asked whether more than $2N$ points force three on a line. That is the pigeonhole half, which `allowedSetSize_le` proves twenty lines later. It now states Green's question.

**Left alone.** `NoKInLine` contains the same claim at `k = 3`. GK2025 proves it for `k > 10^37`, so the range is a maintainer's call. Its docstring now records that nobody conjectures it across `k > 2`.

`lake --wfail build` passes.

Validation on the updated head: `lake --wfail build 'FormalConjectures.GreensOpenProblems.«72»'`. The changed module builds without warnings. This mathematical correction is independently mergeable.

### DIFF
--- a/FormalConjectures/GreensOpenProblems/72.lean
+++ b/FormalConjectures/GreensOpenProblems/72.lean
@@ -26,8 +26,10 @@ for $N$ sufficiently large, is it impossible to have a set of size $2N$ with thi
 The upper bound $2N$ is the easy half and is `allowedSetSize_le` below, by pigeonhole on the
 columns. The open content is whether $2N$ is attained. Green records that it is for $N$ up to
 around 50, that $(3/2 + o(1))N$ points are achievable for arbitrary $N$, and that his "personal
-suspicion is that this is optimal"; the Guy-Kelly heuristic in the Wikipedia reference points the
-same way, at $\pi/\sqrt3 \approx 1.814$. So the expected answer to the question above is yes.
+suspicion is that this is optimal". The Wikipedia reference points the same way: Guy and Kelly
+conjectured $c = \sqrt[3]{2\pi^2/3} \approx 1.874$, and after an error in the heuristic was found
+Guy corrected it to $c = \pi/\sqrt3 \approx 1.814$. Both are below $2$, so the expected answer to
+the question above is yes.
 
 *References:*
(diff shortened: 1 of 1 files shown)


williamjblair:
**Advisory FC review — CLEAN**

Reviewed `fadba095d4ad43b234d5ddb3c89a11da25eebbb9` against `5a13d480c38611f1381ba345f5933e9c29544c08`. Coverage: complete. Produced by a local operator. Maintainers decide acceptance.


[Archived report and evidence provenance](https://github.com/williamjblair/formal-conjectures/tree/1d6fb351052effe45f22d3128ffb0fdf0d312416/runs/20260909T151243Z-ed09352feacf). Raw source documents and private artifacts are omitted.

mo271:
Thanks!


> **Left alone.** `NoKInLine` contains the same claim at `k = 3`. GK2025 proves it for `k > 10^37`, so the range is a maintainer's call. Its docstring now records that nobody conjectures it across `k > 2`.

I will make a follow-up on this!

mo271 (approved):

PR #4945, opened 2026-08-14
ErdosProblems/92: record both questions as disproved

williamjblair:
The source marks problem 92 as disproved. Its remark gives the route:

> This is a stronger form of the unit distance conjecture (see [90]). As such the recent disproof of [90] also disproves this.

This repository already records that disproof: `Erdos90.erdos_90` is `research solved` with `answer(False)`. `92.lean` still said `research open`.

Both variants become `research solved` with `answer(False)`. The strong form follows from the weak one, since `c / log log n` tends to 0, so `n ^ (c / log log n)` is of the form `n ^ o(1)`.

`lake --wfail build` passes, 8054 jobs.

`scripts/check_erdos_status.py` does not catch this. It skips any `.variants.` suffix, and `92.lean` has no bare `erdos_92`, so the file never enters the comparison. 34 Erdős files are invisible for the same reason. I can send that separately if it is wanted.

### DIFF
--- a/FormalConjectures/ErdosProblems/92.lean
+++ b/FormalConjectures/ErdosProblems/92.lean
@@ -19,6 +19,10 @@ import FormalConjecturesUtil
 /-!
 # Erdős Problem 92
 
+Both questions here are disproved, by way of Erdős Problem 90. The source says so directly: this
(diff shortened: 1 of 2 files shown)


mo271 (approved):
Thanks, LGTM

PR #4946, opened 2026-08-14
feat(OEIS) 43 conjectures from `auto_oeis`

mo271:
This adds all the conjectures in the `auto_oeis` branch which oeis sequence number at least 357506.


KitaKen1:
Thanks for putting together this large OEIS batch. I checked a few files and may have found five points worth double-checking. I have not audited the whole PR, and I used AI assistance to condense my much longer notes, so I may have introduced transcription or summarization errors.

1. **A357565:** with the current definition, `(m,p,r) = (2,3,2)` gives
   `u(9,2) - u(3,2) ≡ 2 * 3^8 (mod 3^9)`.
   Was `p = 3` intended to be included in `conjecture_4`?

2. **A383327:** both `147` and `2055` appear to give the value `67`. My enumeration gives `a(67)=5` and `a(69)=1`, so the final `67` in the set may perhaps be `69`.

3. **A379240:** the helper named `A376418` here seems different from OEIS A376418. Under the current helper, `56` and `60` have the same triple `(92,4,2)` although `f` distinguishes them. Could the helper name or definition have been mixed up?

4. **A374265:** the affirmative boundedness statement seems to conflict with this public Lean proof of unboundedness for what appears to be the same process:
   https://github.com/epoch-research/LeanOpenProblems-results/blob/main/runs/oeis-full-50usd-ant-j0j0g4uzligm1k41/oeis_a374265_conjecture_1_boundedness/Submission/Spec.lean

5. **A363414:** `n = 0` already witnesses the existential because `a 0 = 0`. If positive indices were intended, perhaps the witness should require `0 < n`.

I may be missing some context, so I would appreciate an independent check. I can share the small scripts and longer notes if useful.

mo271:
Thanks for the reivew @KitaKen1 ! Those were quite useful -- great to also go the extra mile and catch typos upstream at the oeis!
I fixed all of them. 
Perhaps you could also take a look at
https://github.com/google-deepmind/formal-conjectures/pull/5016 ?

FormalConjectures/OEIS/357506.lean
```diff
@@ -0,0 +1,81 @@
...
+/--
+The stronger congruence $a(p) \equiv 27 \pmod{p^5}$ holds for all primes $p \ge 3$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_1 : ∀ (p : ℕ), Nat.Prime p → p ≥ 3 → a p ≡ 27 [MOD (p ^ 5)] := by
+  sorry
+
+/--
+for $r \ge 2$, $a(p^r) \equiv a(p^{r-1}) \pmod{p^{3r+3}}$ for all primes $p \ge 5$.
+-/
```
codrut3, line 72: Attribute to Peter Bala, Oct 13 2022

FormalConjectures/OEIS/357506.lean
```diff
@@ -0,0 +1,81 @@
...
+theorem a_2 : a 2 = 20577 := by
+  rfl
+
+@[category test, AMS 11]
+theorem a_3 : a 3 = 60353937 := by rfl
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 287798988897 := by rfl
+
+
```
kghristova, line 61: extra new line here for some reason

FormalConjectures/OEIS/357506.lean
```diff
@@ -0,0 +1,81 @@
...
+@[category test, AMS 11]
+theorem a_5 : a 5 = 1782634331587527 := by rfl
+
+/--
+The stronger congruence $a(p) \equiv 27 \pmod{p^5}$ holds for all primes $p \ge 3$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_1 : ∀ (p : ℕ), Nat.Prime p → p ≥ 3 → a p ≡ 27 [MOD (p ^ 5)] := by
```
kghristova, line 69: maybe change to `3 ≤ p` to make the style of the statement consistent with the next one

FormalConjectures/OEIS/357565.lean
```diff
@@ -0,0 +1,99 @@
...
+    (m + 2) * (choose (n + k - 1) k) ^ 2 + (2 * m) * (choose (n + k - 1) k) ^ 3
+
+@[category test, AMS 11]
+theorem a_0 : a 0 = 5 := by rfl
+
+@[category test, AMS 11]
+theorem a_1 : a 1 = 10 := by rfl
+
+@[category test, AMS 11]
+theorem a_2 : a 2 = 114 := by rfl
```
kghristova, line 58: new line here

FormalConjectures/OEIS/357565.lean
```diff
@@ -0,0 +1,99 @@
...
+@[category test, AMS 11]
+theorem a_4 : a 4 = 109106 := by rfl
+
+/--
+$a(p) \equiv a(1) \pmod{p^5}$ for all odd primes $p$ except $p = 5$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_1 (p : ℕ) (hp : Nat.Prime p) (h_ne2 : p ≠ 2) (h_ne5 : p ≠ 5) :
+    (a p) ≡ (a 1) [MOD (p ^ 5)] := by
+  sorry
```
kghristova, line 71: new line here

FormalConjectures/OEIS/357565.lean
```diff
@@ -0,0 +1,99 @@
...
+/--
+$a(p^r) \equiv a(p^{r-1}) \pmod{p^{3r+3}}$ for $r \ge 2$ and all primes $p \ge 3$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_2 (p r : ℕ) (hp : Nat.Prime p) (h_pge3 : p ≥ 3) (hr : r ≥ 2) :
+    (a (p ^ r)) ≡ (a (p ^ (r - 1))) [MOD (p ^ (3 * r + 3))] := by
+  sorry
```
kghristova, line 78: new line here

FormalConjectures/OEIS/357565.lean
```diff
@@ -0,0 +1,99 @@
...
+/--
+$u(p) \equiv u(1) \pmod{p^5}$ holds for all primes $p \ge 7$ and positive integer $m$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_3 (m p : ℕ) (hm : m > 0) (hp : Nat.Prime p) (h_pge7 : p ≥ 7) :
+    (u p m) ≡ (u 1 m) [MOD (p ^ 5)] := by
+  sorry
```
kghristova, line 85: new line here

FormalConjectures/OEIS/357565.lean
```diff
@@ -0,0 +1,99 @@
...
+namespace OeisA357565
+
+
+/--
+The sequence $a(n) = 3 \sum_{k=0}^n \binom{n+k-1}{k}^2 + 2 \sum_{k=0}^n \binom{n+k-1}{k}^3$.
+-/
+def a (n : ℕ) : ℕ :=
+  (range (n + 1)).sum fun k =>
+    let b := choose (n + k - 1) k
+    3 * b ^ 2 + 2 * b ^ 3
```
kghristova, line 39: not sure it makes sense to have the `let b := choose (n + k - 1) k` -  it's not done in the def of `u`.  Also not sure if it's worth opening Finset just for the Finset.range and Finset.sum

FormalConjectures/OEIS/357565.lean
```diff
@@ -0,0 +1,99 @@
...
+
+/--
+The generalized sequence $u(n, m)$ from the conjecture section:
+$u(n, m) = (m + 2) \sum_{k = 0}^{m \cdot n} \binom{n+k-1}{k}^2 + 2m \sum_{k = 0}^{m \cdot n}
+\binom{n+k-1}{k}^3$.
+Note that $a(n) = a\_u(n, 1)$.
```
kghristova, line 45: what is `a\_u(n, 1)`?

FormalConjectures/OEIS/357569.lean
```diff
@@ -0,0 +1,80 @@
...
+
+/--
+The generalized sequence $u(k, n) = 2 \binom{3n}{n}^k - k \cdot 3^{k+1} \binom{2n}{n}$ for $k
+\ge 1$.
+Note that $u(2, n) = 2 \cdot a(n)$.
+-/
+def u (k n : ℕ) : ℤ :=
+  2 * (Int.ofNat ((3 * n).choose n)) ^ k -
+  (k : ℤ) * ((3 : ℤ) ^ (k + 1)) * Int.ofNat ((2 * n).choose n)
+
```
kghristova, line 64: you could do 
` 2 * ((3 * n).choose n : ℤ) ^ k -
  (k : ℤ) * ((3 : ℤ) ^ (k + 1)) * (((2 * n).choose n) : ℤ)`

for consistency (i.e. to avoid having Int.ofNat and ( : ℤ) in the same statement).

FormalConjectures/OEIS/357569.lean
```diff
@@ -0,0 +1,80 @@
...
+theorem conjecture_1 (p r : ℕ) (hp : Nat.Prime p) (hp3 : p ≥ 3) (hr : r ≥ 2) :
+  a (p ^ r) ≡ a (p ^ (r - 1)) [ZMOD ((p : ℤ) ^ (3 * r + 3))] :=
+by sorry
+
+/--
+For $k \ge 1$, the sequence $u(k, n)$ satisfies the same supercongruences
+$u(k, p^r) \equiv u(k, p^{r-1}) \pmod{p^{3r+3}}$ for $r \ge 2$ and all primes $p \ge 3$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_2 (k p r : ℕ) (hk : 1 ≤ k) (hp : Nat.Prime p) (hp3 : p ≥ 3) (hr : r ≥ 2) :
```
kghristova, line 76: maybe flip `(hk : k ≥ 1)` for consistency

FormalConjectures/OEIS/357569.lean
```diff
@@ -0,0 +1,80 @@
...
+@[category test, AMS 11]
+theorem a_1 : a 1 = -45 := by rfl
+
+@[category test, AMS 11]
+theorem a_2 : a 2 = 63 := by rfl
+
+@[category test, AMS 11]
+theorem a_3 : a 3 = 6516 := by rfl
+
+
```
kghristova, line 52: double new line

FormalConjectures/OEIS/357565.lean
```diff
@@ -0,0 +1,99 @@
...
+theorem a_3 : a 3 = 2926 := by rfl
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 109106 := by rfl
+
+/--
+$a(p) \equiv a(1) \pmod{p^5}$ for all odd primes $p$ except $p = 5$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_1 (p : ℕ) (hp : Nat.Prime p) (h_ne2 : p ≠ 2) (h_ne5 : p ≠ 5) :
```
kghristova, line 69: could have `(hp : p.Prime)` throughout since `Nat` is open

FormalConjectures/OEIS/357674.lean
```diff
@@ -0,0 +1,102 @@
...
+
+namespace OeisA357674
+
+
+/--
+The sequence $a(n) = \left( \sum_{k=0}^{2n} \binom{n+k-1}{k} \right)^4 \left( \sum_{k=0}^{2n}
+\binom{n+k-1}{k}^2 \right)^3$.
+-/
+def a (n : ℕ) : ℕ :=
+  let S1 : ℕ := Finset.sum (range (2 * n + 1)) (fun k => (n + k - 1).choose k)
```
kghristova, line 39: either `Finset.range` and remove the open Finset module, or `(range (2 * n + 1)).sum`

FormalConjectures/OEIS/357674.lean
```diff
@@ -0,0 +1,102 @@
...
+The general sequence $u(n, m)$ from conjecture 3.
+$u(n, m) = \left( \sum_{k = 0}^{m*n} \binom{n+k-1}{k} \right)^{2m} \cdot \left( \sum_{k =
+0}^{m*n} \binom{n+k-1}{k}^2 \right)^{m+1}$.
+Note that `a n = u n 2`.
+-/
+def u (n m : ℕ) : ℕ :=
+  let S1 : ℕ := Finset.sum (range (m * n + 1)) (fun k => (n + k - 1).choose k)
+  let S2 : ℕ := Finset.sum (range (m * n + 1)) (fun k => ((n + k - 1).choose k) ^ 2)
+  S1 ^ (2 * m) * S2 ^ (m + 1)
+
```
kghristova, line 53: same comments re: Finset as above

FormalConjectures/OEIS/357958.lean
```diff
@@ -0,0 +1,107 @@
...
+
+@[category test, AMS 11]
+theorem a_5 : a 5 = 4112539 := by rfl
+
+/--
+$a(p) \equiv a(1) \pmod{p^5}$ for all primes $p \ge 5$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_1 :
+  ∀ (p : ℕ), Nat.Prime p → 5 ≤ p → (a p) ≡ (a 1) [MOD p^5] :=
```
kghristova, line 79: could change to`p.Prime`. Applies throughout the conjecture statements

FormalConjectures/OEIS/358340.lean
```diff
@@ -0,0 +1,95 @@
...
+  have h_ge := hx.1.1
+  interval_cases x <;> (revert hx; norm_num [is_zeroless, is_n_digit])
+
+/--
+It has been proved that there exist infinitely many zeroless squares and cubes but there is
+apparently no proof for 4th powers, 5th powers, etc.
+
+Formalized as the conjecture that the set of natural numbers whose fourth power is zeroless is
+infinite.
+This is equivalent to the statement that the set $\{ m : ℕ \mid \text{is\_n\_digit}(m, n) \land
```
codrut3, line 86: The statements aren't really equivalent: there can be infinitely many zeroless m^4 numbers but a sparse set, so that the n-th digit numbers are missed for some n.

FormalConjectures/OEIS/361713.lean
```diff
@@ -0,0 +1,80 @@
...
+open Finset Nat
+
+namespace OeisA361713
+
+
+/--
+The sequence $a(n) = \sum_{k=0}^{n-1} \binom{n}{k}^2 \binom{n+k-1}{k}^2$.
+-/
+def a (n : ℕ) : ℕ :=
+  Finset.sum (Finset.range n) fun k => (n.choose k) ^ 2 * ((n + k - 1).choose k) ^ 2
```
codrut3: Nit: use sum operator

FormalConjectures/OEIS/361713.lean
```diff
@@ -0,0 +1,80 @@
...
+/--
+The supercongruence $a(p) \equiv a(1) \pmod{p^5}$ holds for all primes $p \ge 7$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_1 (p : ℕ) (hp : Nat.Prime p) (hp7 : 7 ≤ p) :
+    a p ≡ a 1 [MOD p ^ 5] := by
+  sorry
+
+/--
+for $r \ge 2$, the supercongruence $a(p^r) \equiv a(p^{r-1}) \pmod{p^{4r+1}}$ holds for all
```
codrut3: Nit: For

FormalConjectures/OEIS/361714.lean
```diff
@@ -0,0 +1,78 @@
...
+open Nat Int Finset
+
+namespace OeisA361714
+
+
+/--
+The sequence $a(n) = \sum_{k=0}^{n-1} (-1)^{n+k+1} \binom{n}{k} \binom{n+k-1}{k}^2$.
+-/
+noncomputable def a (n : ℕ) : ℕ :=
+  (Finset.sum (Finset.range n) fun k : ℕ =>
```
codrut3: Nit: use sum operator

FormalConjectures/OEIS/361883.lean
```diff
@@ -0,0 +1,79 @@
...
+def a (n : ℕ) : ℕ :=
+  if n = 0 then 0
+  else
+    -- Calculate the numerator sum S in ℕ
+    -- We use binomial(n+k-1, n-1) which is equal to binomial(n+k-1, k)
+    -- This makes the dependency on 'n - 1' explicit for the lower index.
+    let S : ℕ := Finset.sum (range (n + 1)) fun k =>
+      (n + 2 * k) * (Nat.choose (n + k - 1) (n - 1)) ^ 3
+
+    -- Division is exact since a(n) is an integer sequence.
```
codrut3: Is this obvious?

FormalConjectures/OEIS/357506.lean
mo271: yeah, looks a bit off, fixing it, but I rather hope in auto-formatting in the futures: I think there are also lots of preexisting finds of that sort , for instance grepping for `\n\n\n\n` in the repo gives more than 10 results.

FormalConjectures/OEIS/361883.lean
```diff
@@ -0,0 +1,79 @@
...
+theorem a_4 : a 4 = 150722 := by rfl
+
+
+/--
+The central binomial coefficients $u(n) := \binom{2n}{n}$ satisfy the supercongruences
+$u(n \cdot p^r) \equiv u(n \cdot p^{r-1}) \pmod{p^{3r}}$ for positive integers $n$ and $r$
+and all primes $p \ge 5$. We conjecture that the present sequence $a(n)$ satisfies the same
+congruences.
+-/
+@[category research open, AMS 11]
```
codrut3, line 71: This is reported as solved in https://arxiv.org/pdf/2608.11941

FormalConjectures/OEIS/363983.lean
```diff
@@ -0,0 +1,74 @@
...
+
+namespace OeisA363983
+
+
+/--
+The sequence $a(n) = \sum_{k=\lfloor(n+1)/2\rfloor}^n (-1)^{n+k} \binom{n}{k} \binom{n+k-1}{k}
+\binom{2k}{n}$.
+-/
+def a (n : ℕ) : ℕ :=
+  (Finset.sum (Finset.range (n + 1)) fun k : ℕ =>
```
codrut3: Nit: use sum operator

FormalConjectures/OEIS/363983.lean
```diff
@@ -0,0 +1,74 @@
...
+    -- The expression must result in ℤ due to the alternating sign.
+    let sign_factor : ℤ := (-1) ^ (n + k)
+    -- Binomial coefficients (Nat.choose) are implicitly coerced to ℤ for multiplication.
+    -- (n + k - 1).choose k is written as ((n + k).pred.choose k) in Mathlib's Nat.choose syntax.
+    let term_val : ℤ := (n.choose k) * ((n + k).pred.choose k) * ((2 * k).choose n)
+    sign_factor * term_val
+  ).toNat
```
codrut3, line 44: Is it clear this is nonnegative?

FormalConjectures/OEIS/357569.lean
mo271: fixed many of them, not just the ones mentioned here..

FormalConjectures/OEIS/357506.lean
mo271: done

FormalConjectures/OEIS/364175.lean
```diff
@@ -0,0 +1,210 @@
...
+
+/--
+The sequence $a(n) = \frac{(6n)! (2n/3)!}{(3n)! (2n)! (5n/3)!}$ for $n$ divisible by 3.
+-/
+noncomputable def a (n : ℕ) : ℕ :=
+  let n_r : ℝ := n.cast
+  let val_R : ℝ :=
+    (Real.Gamma (6 * n_r + 1) * Real.Gamma (2 / 3 * n_r + 1)) /
+    (Real.Gamma (3 * n_r + 1) * Real.Gamma (2 * n_r + 1) * Real.Gamma (5 / 3 * n_r + 1))
+  (round val_R).toNat
```
codrut3, line 40: From the OEIS page:
This sequence is only conjecturally an integer sequence.

So you can add this as a conjecture too.

FormalConjectures/OEIS/357565.lean
mo271: inlined, kept Finset open

FormalConjectures/OEIS/364178.lean
```diff
@@ -0,0 +1,168 @@
...
+
+/--
+The sequence $a(n) = \frac{(10n)! (3n)! (n/2)!}{(6n)! (5n)! (3n/2)! n!}$ for even $n$.
+-/
+noncomputable def a (n : ℕ) : ℕ :=
+  (round
+    ((Gamma (10 * (↑n : ℝ) + 1) * Gamma (3 * (↑n : ℝ) + 1) * Gamma ((↑n : ℝ) / 2 + 1)) /
+     (Gamma (6 * (↑n : ℝ) + 1) * Gamma (5 * (↑n : ℝ) + 1) *
+      Gamma (3 * (↑n : ℝ) / 2 + 1) * Gamma (↑n + 1)))
+  ).toNat
```
codrut3, line 41: This sequence is only conjecturally an integer sequence.
(from the OEIS page)

FormalConjectures/OEIS/357565.lean
mo271: good cathc, it is just `a_(n, 1)`

FormalConjectures/OEIS/357569.lean
mo271: done also below

FormalConjectures/OEIS/357569.lean
mo271: done

FormalConjectures/OEIS/357565.lean
mo271: fixed here and in 22 other files

FormalConjectures/OEIS/357674.lean
mo271: done here and everywhere

FormalConjectures/OEIS/357674.lean
mo271: done here and elsewhere

mo271: actually I now use sum notation throughout

FormalConjectures/OEIS/357565.lean
```diff
@@ -0,0 +1,97 @@
...
+    ((m + 2) * (choose (n + k - 1) k) ^ 2 + (2 * m) * (choose (n + k - 1) k) ^ 3)
+
+@[category test, AMS 11]
+theorem a_0 : a 0 = 5 := by rfl
+
+@[category test, AMS 11]
+theorem a_1 : a 1 = 10 := by rfl
+
+@[category test, AMS 11]
+theorem a_2 : a 2 = 114 := by rfl
```
mo271, line 56:
```suggestion
theorem a_2 : a 2 = 114 := by rfl

```

FormalConjectures/OEIS/363983.lean
mo271: already done thoughout

FormalConjectures/OEIS/370092.lean
```diff
@@ -0,0 +1,94 @@
...
+-/
+open Finset Nat
+
+namespace OeisA370092
+
+/--
+$a(0) = 1$, and $a(n) = (-1)^n + \frac{1}{2} \sum_{j=1}^n (1 - (-1)^j - (-2)^j) \binom{n}{j}
+a(n-j)$ for $n > 0$.
+-/
+noncomputable def a (n : ℕ) : ℚ :=
```
codrut3, line 36: I can see a(n) is always an integer, so I think it's safe to convert.

FormalConjectures/OEIS/370092.lean
```diff
@@ -0,0 +1,94 @@
...
+The reduction of `a n` modulo `k`. Since all terms of `a n` are integers, we take the
+numerator of the rational number representation, which is the integer value, and reduce it
+modulo `k`.
+This requires `k > 0`, which is guaranteed by `2 < k`.
+-/
+noncomputable def a_mod_k (k : ℕ) (n : ℕ) : ZMod k :=
+  Int.cast (a n).num
+
+/--
+%C a Conjecture: Let k > 2 be a positive integer. The sequence obtained by reducing a(n) modulo
```
codrut3: Nit: delete %C a

FormalConjectures/OEIS/376930.lean
```diff
@@ -0,0 +1,78 @@
...
+theorem a_3 : a 3 = 2 := by rfl
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 3 := by rfl
+
+@[category test, AMS 11]
+theorem a_5 : a 5 = 1 := by rfl
+
+/--
+It is not known if the sequence contains any negative terms (which may happen if two primes are
```
codrut3, line 66: Shouldn't the sequence map to Z then? What does Lean do in these cases?

FormalConjectures/OEIS/379240.lean
```diff
@@ -0,0 +1,170 @@
...
+It is conjectured that this is also the lexicographically earliest infinite sequence such
+that a(i) = a(j) => A003415(i) = A003415(j), A085731(i) = A085731(j) and A376418(i) = A376418(j),
+for all i, j >= 1, i.e., the restricted growth sequence transform of the triple
+[A003415(n), A085731(n), A376418(n)]. This is true if for every pair of $i$ and $j$ for which
+$i \ne j$, and $\mathtt{A376418}(i) = \mathtt{A376418}(j) > 0$, the ordered pairs
+$[\mathtt{A003415}(i), \mathtt{A085731}(i)]$ and $[\mathtt{A003415}(j), \mathtt{A085731}(j)]$
+differ from each other.
+-/
+@[category research open, AMS 11]
+theorem a_eq_rgs_transform (n : ℕ) : a n = A379240_conjecture n := by
```
codrut3, line 167: I honestly have a very hard time understanding this sequence.

It was also disproved by Tom's recent paper.

FormalConjectures/OEIS/385391.lean
```diff
@@ -0,0 +1,102 @@
...
+    · exact h⟩
+
+/--
+a(1) = A002110(0), a(2) = A002110(1), a(3) = A002110(2), a(6) = A002110(3), a(7) = A002110(4),
+a(10) = A002110(5), ...?
+This conjecture is formalized as a conjunction of the listed equalities, implying a general
+pattern related to A065295.
+-/
+@[category research open, AMS 11]
+theorem a_eq_primorial :
```
codrut3, line 93: This seems like the wrong formalization. I think the purpose of the conjecture is to characterize n such a(n) = primorial(k). Or:
is it true that for every k, there exists a unique n_k such that a(n_k) = primorial(k), and n_0 < n_1 < ...

Tom's paper claims a proof, but the proof only validates the six numbers (which were already known).

FormalConjectures/OEIS/385958.lean
```diff
@@ -0,0 +1,242 @@
...
+- [A385958](https://oeis.org/A385958)
+-/
+open Nat Finset
+
+namespace OeisA385958
+
+/--
+A helper function to find the largest prime $p$ such that $p-1$ divides $2 \cdot k$.
+This is the definition of $a(n)$ given $b(n-1)=k$.
+Since $k \ge 1$, $2k \ge 2$, and the set of such primes is non-empty (it always contains $p=2$).
```
codrut3, line 35: And p = 3 too, no?

FormalConjectures/OEIS/359634.lean
```diff
@@ -0,0 +1,91 @@
...
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 3 := by rfl
+
+/--
+Conjecture (OEIS a, C-line): A zero has not appeared in the sequence $a(n)$.
+This is equivalent to $\forall n : \mathbb{N}, a(n) \neq 0$.
+-/
+@[category research open, AMS 11]
+theorem never_zero (n : ℕ) : a n ≠ 0 := by
```
kghristova, line 88: call this conjecture? Also, this seems to be a conjectured in the loose sense. They say up to certain point, there has been no zero but haven't explicitly conjectured this as global behaviour. I guess we could conjecture this and see if it's true. 

Also, there's this question, but it might be difficult to formalise: `How does the lower envelope of this sequence behave`

FormalConjectures/OEIS/359634.lean
```diff
@@ -0,0 +1,91 @@
...
+      max_contiguous_length
+  )
+
+@[category test, AMS 11]
+theorem a_0 : a 0 = 1 := by
+  rfl
+
+@[category test, AMS 11]
+theorem a_1 : a 1 = 1 := by
+  trivial
```
kghristova, line 71: weird that this is by trivial and all others by rfl

FormalConjectures/OEIS/361711.lean
```diff
@@ -0,0 +1,74 @@
...
+@[category test, AMS 11]
+theorem a_5 : a 5 = 126 := by rfl
+
+/--
+The supercongruence $a(p^k) \equiv a(p^{k-1}) \pmod{p^{3k}}$ holds for all primes $p \ge 5$ and
+positive integers $k$.
+-/
+@[category research open, AMS 11]
+theorem conjecture (p : ℕ) (hp : p.Prime) (h_geq_5 : 5 ≤ p) (k : ℕ) (hk : k > 0) :
+    a (p ^ k) ≡ a (p ^ (k - 1)) [ZMOD (p ^ (3 * k) : ℕ)] := by
```
kghristova, line 71: do we need the explicit cast in `(3 * k) : ℕ`?

FormalConjectures/OEIS/361711.lean
```diff
@@ -0,0 +1,74 @@
...
+
+@[category test, AMS 11]
+theorem a_5 : a 5 = 126 := by rfl
+
+/--
+The supercongruence $a(p^k) \equiv a(p^{k-1}) \pmod{p^{3k}}$ holds for all primes $p \ge 5$ and
+positive integers $k$.
+-/
+@[category research open, AMS 11]
+theorem conjecture (p : ℕ) (hp : p.Prime) (h_geq_5 : 5 ≤ p) (k : ℕ) (hk : k > 0) :
```
kghristova, line 70: could have `(hk : 0 < k)` for consistency with hp

FormalConjectures/OEIS/361714.lean
```diff
@@ -0,0 +1,76 @@
...
+
+/--
+The sequence $a(n) = \sum_{k=0}^{n-1} (-1)^{n+k+1} \binom{n}{k} \binom{n+k-1}{k}^2$.
+-/
+noncomputable def a (n : ℕ) : ℕ :=
+  (∑ k ∈ range n,
+    (
+      (-1 : ℤ) ^ (n + k + 1) *
+      (n.choose k : ℤ) *
+      ((Nat.choose (n + k - 1) k : ℤ) ^ 2)
```
kghristova, line 40: `(n + k - 1).choose k` instead of `(Nat.choose (n + k - 1) k`

FormalConjectures/OEIS/361714.lean
```diff
@@ -0,0 +1,76 @@
...
+- [A361714](https://oeis.org/A361714)
+-/
+open Nat Int Finset
+
+namespace OeisA361714
+
+/--
+The sequence $a(n) = \sum_{k=0}^{n-1} (-1)^{n+k+1} \binom{n}{k} \binom{n+k-1}{k}^2$.
+-/
+noncomputable def a (n : ℕ) : ℕ :=
```
kghristova, line 35: I would suggest we change the type of the sequence here to `a (n : ℕ) : \Z`. This would remove the awkward natAbs at the end of the def, and also avoid casting explicitly to \Z in the LHS of the conjecture :)

FormalConjectures/OEIS/361714.lean
```diff
@@ -0,0 +1,76 @@
...
+theorem a_3 : a 3 = 82 := by rfl
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 1063 := by rfl
+
+/--
+The supercongruence $a(p) \equiv a(1) \pmod{p^5}$ holds for all primes $p \ge 7$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_1 {p : ℕ} (hp : p.Prime) (hp_ge_7 : 7 ≤ p) :
```
kghristova, line 63: do we want the p to be implicit?

FormalConjectures/OEIS/361714.lean
```diff
@@ -0,0 +1,76 @@
...
+  (a p : ℤ) ≡ a 1 [ZMOD (p ^ 5 : ℤ)] :=
+  by sorry
+
+/--
+For $r \ge 2$, the supercongruence
+$a(p^r) \equiv a(p^{r-1}) \pmod{p^{3r+3}}$ holds for all primes $p \ge 7$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_2 {p r : ℕ} (hp : p.Prime) (hp_ge_7 : 7 ≤ p) (hr_ge_2 : 2 ≤ r) :
```
kghristova, line 72: do we want p and r to be explicit as it is in the other conjectures?

FormalConjectures/OEIS/361715.lean
```diff
@@ -0,0 +1,66 @@
...
+@[category test, AMS 11]
+theorem a_3 : a 3 = 82 := by rfl
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 745 := by rfl
+
+/-- The supercongruence $a(p) \equiv a(1) \pmod{p^5}$ holds for all primes $p \ge 7$. -/
+@[category research open, AMS 11]
+theorem conjecture_1 (p : ℕ) (hp : p.Prime) (hp7 : 7 ≤ p) :
+  (a p : ℤ) ≡ a 1 [ZMOD (p ^ 5 : ℕ)] := by sorry
```
kghristova, line 56: why are we casting to \Z here and doing ZMod? surely everything is in \N?

FormalConjectures/OEIS/361715.lean
```diff
@@ -0,0 +1,66 @@
...
+
+/--
+for $r \ge 2$, the supercongruence $a(p^r) \equiv a(p^{r-1}) \pmod{p^{3r+3}}$ holds for
+all primes $p \ge 5$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_2 (p r : ℕ) (hp : p.Prime) (hp5 : 5 ≤ p) (hr : 2 ≤ r) :
+  (a (p ^ r) : ℤ) ≡ a (p ^ (r - 1)) [ZMOD (p ^ (3 * r + 3) : ℕ)] := by sorry
```
kghristova, line 64: same comment with the \Z cast. Also we have a weird \N cast.. I think we just want ` a (p ^ r) ≡ a (p ^ (r - 1)) [ZMOD p ^ (3 * r + 3) ]`

FormalConjectures/OEIS/361715.lean
```diff
@@ -0,0 +1,66 @@
...
+@[category test, AMS 11]
+theorem a_4 : a 4 = 745 := by rfl
+
+/-- The supercongruence $a(p) \equiv a(1) \pmod{p^5}$ holds for all primes $p \ge 7$. -/
+@[category research open, AMS 11]
+theorem conjecture_1 (p : ℕ) (hp : p.Prime) (hp7 : 7 ≤ p) :
+  (a p : ℤ) ≡ a 1 [ZMOD (p ^ 5 : ℕ)] := by sorry
+
+/--
+for $r \ge 2$, the supercongruence $a(p^r) \equiv a(p^{r-1}) \pmod{p^{3r+3}}$ holds for
```
kghristova: For

codrut3 (approved):
Forgot to approve :)

FormalConjectures/OEIS/357506.lean
mo271: done

FormalConjectures/OEIS/361713.lean
mo271: done, here and everywhere

FormalConjectures/OEIS/361883.lean
mo271: Yes, in fact every term of the sum is divisible by $n$ individually: using the absorption identity $k \binom{n+k-1}{k} = n \binom{n+k-1}{k-1}$, we have $$(n+2k)\binom{n+k-1}{k} = n\left(\binom{n+k-1}{k} + 2\binom{n+k-1}{k-1}\right) = n\left(\binom{n+k}{k} + \binom{n+k-1}{k-1}\right)$$ which gives $n \mid (n+2k)\binom{n+k-1}{k}^3$ for every $k$.

added a comment

FormalConjectures/OEIS/361883.lean
```diff
@@ -0,0 +1,76 @@
...
+The sequence $a(n) = \frac{1}{n} \sum_{k=0}^n (n+2k) \binom{n+k-1}{k}^3$.
+-/
+def a (n : ℕ) : ℕ :=
+  if n = 0 then 0
+  else
+    -- Calculate the numerator sum S in ℕ
+    -- We use binomial(n+k-1, n-1) which is equal to binomial(n+k-1, k)
+    -- This makes the dependency on 'n - 1' explicit for the lower index.
+    let S : ℕ := ∑ k ∈ range (n + 1),
+      (n + 2 * k) * (Nat.choose (n + k - 1) (n - 1)) ^ 3
```
kghristova, line 42: we have open Nat so can do (n-1).choose (n + k - 1)

FormalConjectures/OEIS/361883.lean
kghristova: I'm not sure it is either since it seems the simplification "forgets" the ^ 3 at the end. I think we want to do integer division just in case

FormalConjectures/OEIS/361883.lean
kghristova: Is the first part of the second part of the statement solved? I don't see why we care to state the first part in the docstring tbh, we should only say 

For all positive integers $n, r$ and primes $p \ge 5$, $a(n \cdot p^r) \equiv a(n \cdot p^{r-1}) \pmod{p^{3r}}$

FormalConjectures/OEIS/361883.lean
```diff
@@ -0,0 +1,76 @@
...
+theorem a_4 : a 4 = 150722 := by rfl
+
+/--
+The central binomial coefficients $u(n) := \binom{2n}{n}$ satisfy the supercongruences
+$u(n \cdot p^r) \equiv u(n \cdot p^{r-1}) \pmod{p^{3r}}$ for positive integers $n$ and $r$
+and all primes $p \ge 5$. We conjecture that the present sequence $a(n)$ satisfies the same
+congruences.
+-/
+@[category research open, AMS 11]
+theorem supercongruence {p n r : ℕ} (hp : p.Prime) (hp5 : 5 ≤ p) (hn : 0 < n) (hr : 0 < r) :
```
kghristova, line 72: maybe change name to conjecture?

FormalConjectures/OEIS/363414.lean
```diff
@@ -0,0 +1,83 @@
...
+-/
+open Nat Finset Complex Real Filter Asymptotics ZMod Int
+
+namespace OeisA363414
+
+/--
+$a(n) = \frac{1}{2} \operatorname{Im}\left( \prod_{k=0}^n (1 + 2k i) \right)$.
+-/
+noncomputable def a (n : ℕ) : ℤ :=
+  let P_n : Complex :=
```
kghristova, line 36: P_n : \C

FormalConjectures/OEIS/363414.lean
```diff
@@ -0,0 +1,83 @@
...
+    Finset.prod (range (n + 1))
+    (fun k : ℕ ↦ (1 : Complex) + ((2 * k : ℕ) : ℝ) * Complex.I)
```
kghristova, line 38: (fun k : ℕ ↦ (1 : ℂ) + ((2 * k) : ℂ) * Complex.I) - we don't need all these coarcions

FormalConjectures/OEIS/363983.lean
```diff
@@ -0,0 +1,73 @@
...
+open Nat Finset Int
+
+namespace OeisA363983
+
+/--
+The sequence $a(n) = \sum_{k=\lfloor(n+1)/2\rfloor}^n (-1)^{n+k} \binom{n}{k} \binom{n+k-1}{k}
+\binom{2k}{n}$.
+-/
+def a (n : ℕ) : ℕ :=
+  (∑ k ∈ range (n + 1),
```
kghristova, line 37: The range of k is $\lfloor((n+1)/2) \rfloor ...n$

FormalConjectures/OEIS/363983.lean
kghristova: It looks like it from the sequence but it's not immediately obvious I think. So maybe better to leave it in \Z?

kghristova: Also, we seem to be explicitly coarcign the sequence terms to \Z in the formalisation of the conjecture. So why not just define in \Z?

FormalConjectures/OEIS/363983.lean
```diff
@@ -0,0 +1,73 @@
...
+theorem a_4 : a 4 = 1310 := by rfl
+
+/--
+The Franel numbers satisfy the supercongruences
+$A000172(n p^r) \equiv A000172(n p^{r-1}) \pmod{p^{3r}}$
+for all primes $p \ge 5$ and positive integers $n$ and $r$.
+The present sequence satisfies the same supercongruences.
+-/
+@[category research open, AMS 11]
+theorem supercongruence (p n r : ℕ) (hp : p.Prime) (h_p_ge_5 : p ≥ 5) (hn : n > 0) (hr : r
```
kghristova, line 68: change theorem name to conjecture_1? or conjecture?

FormalConjectures/OEIS/364173.lean
```diff
@@ -0,0 +1,167 @@
...
+@[category test, AMS 11]
+theorem a_4 : a 4 = 7012604550 := by
+  unfold a
+  norm_num
+
+/--
+Conjecture 1: This sequence is an integer sequence, i.e., $a(n) \in \mathbb{Z}$ for all $n$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_1_integrality : ∀ (n : ℕ), a n ∈ Set.range (Int.cast : ℤ → ℝ) := by
```
kghristova, line 150: I don't like this phrasing too much, it's a bit messy.. How about 

``` 
theorem conjecture_1_integrality:  ∀ (n : ℕ)  ∃ (k : ℤ), a n = k := by
  sorry

File: FormalConjectures/OEIS/364173.lean [hunk 52]
```diff
@@ -0,0 +1,167 @@
...
+primes $p \ge 5$ and all positive integers $n$ and $r$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_2_supercongruence
+    (h_int : ∀ m : ℕ, a m ∈ (Set.range (fun (x : ℤ) => (x : ℝ)))) :
+  ∀ (p : ℕ) (hp : Nat.Prime p) (h_p_ge_5 : 5 ≤ p)
+    (n r : ℕ) (hn : n > 0) (hr : r > 0),
+  (Classical.choose (h_int (n * p ^ r)) : ℤ)
+  ≡ (Classical.choose (h_int (n * p ^ (r - 1))) : ℤ)
+  [ZMOD ((p : ℤ) ^ (3 * r))] :=
```
kghristova, line 164: oof this statement is super clunky too :( I wonder if we can either integrate it into the previous conjecture or define an integer sequence version of `a n`. Or maybe do something like this

```
theorem conjecture_2_supercongruence
  (p n r : ℕ) (hp : Nat.Prime p) (hp5 : 5 ≤ p) (hn : 0 < n) (hr : 0 < r) : 
  ∀ (n : ℕ), ∃ (k₁ k₂ : ℤ),  
  a (n * p ^ r) = k₁ ∧ a (n * p ^ (r - 1)) = k₂ ∧ k₁ ≡ k₂ [ZMOD (p : ℤ) ^ (3 * r)] := by
  sorry

File: FormalConjectures/OEIS/364173.lean [hunk 53]
```diff
@@ -0,0 +1,167 @@
...
+
+*References:*
+- [A364173](https://oeis.org/A364173)
+-/
+open scoped Real
+
+namespace OeisA364173
+
+/--
+The sequence $a(n) = \frac{(9n)! (2n)! (3n/2)!}{(9n/2)! (4n)! (3n)! n!}$ for even $n$.
```
kghristova, line 33: why do we need n even here?

FormalConjectures/OEIS/364175.lean
```diff
@@ -0,0 +1,202 @@
...
+-/
+open Real Nat Int
+
+namespace OeisA364175
+
+/--
+The sequence $a(n) = \frac{(6n)! (2n/3)!}{(3n)! (2n)! (5n/3)!}$ for $n$ divisible by 3.
+-/
+noncomputable def a (n : ℕ) : ℕ :=
+  let n_r : ℝ := n.cast
```
kghristova, line 36: the .cast is a bit weird

FormalConjectures/OEIS/364175.lean
kghristova: I agree with Codrut. it feels weird just coercing it to \N

kghristova: Also, do we really want all these Gamma functions?

FormalConjectures/OEIS/364175.lean
```diff
@@ -0,0 +1,202 @@
...
+
+*References:*
+- [A364175](https://oeis.org/A364175)
+-/
+open Real Nat Int
+
+namespace OeisA364175
+
+/--
+The sequence $a(n) = \frac{(6n)! (2n/3)!}{(3n)! (2n)! (5n/3)!}$ for $n$ divisible by 3.
```
kghristova, line 33: are we really restricting ourselves to n divisible by 3 by using the Gamma functions?

FormalConjectures/OEIS/364176.lean
```diff
@@ -0,0 +1,211 @@
...
+open Real Nat
+
+namespace OeisA364176
+
+/--
+The sequence $a(n) = \frac{(15n)! (5n/2)! (2n)!}{(15n/2)! (6n)! (5n)! n!}$ for even $n$.
+-/
+noncomputable def a (n : ℕ) : ℝ :=
+  let n_r : ℝ := n.cast
+  let num_int_15 : ℝ := (15 * n).factorial.cast
```
kghristova, line 37: why do we need the num_int_15 and num_int_2? We've already cast n to a Real

FormalConjectures/OEIS/364176.lean
```diff
@@ -0,0 +1,211 @@
...
+
+*References:*
+- [A364176](https://oeis.org/A364176)
+-/
+open Real Nat
+
+namespace OeisA364176
+
+/--
+The sequence $a(n) = \frac{(15n)! (5n/2)! (2n)!}{(15n/2)! (6n)! (5n)! n!}$ for even $n$.
```
kghristova, line 33: Again, are we really restricting to n even? I don't know where this is being used

FormalConjectures/OEIS/364176.lean
```diff
@@ -0,0 +1,211 @@
...
+-/
+noncomputable def a (n : ℕ) : ℝ :=
+  let n_r : ℝ := n.cast
+  let num_int_15 : ℝ := (15 * n).factorial.cast
+  let num_int_2 : ℝ := (2 * n).factorial.cast
+  let num_frac_5_halves : ℝ := Real.Gamma (5 * n_r / 2 + 1)
+
+  let den_frac_15_halves : ℝ := Real.Gamma (15 * n_r / 2 + 1)
+  let den_int_6 : ℝ := (6 * n).factorial.cast
```
kghristova, line 42: same comment fot den_int_6 and den_int_5

FormalConjectures/OEIS/364176.lean
```diff
@@ -0,0 +1,211 @@
...
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 126694219977836700 := by
+  norm_num [a]
+
+/--
+Conjecture 1: This sequence is an integer sequence, i.e., $a(n) \in \mathbb{Z}$ for all $n$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_1_integrality : ∀ (n : ℕ), a n ∈ Set.range (Int.cast : ℤ → ℝ) := by
```
kghristova, line 187: I would suggest once again swapping to something more idiomatic like 

``` 
theorem conjecture_1_integrality:  ∀ (n : ℕ)  ∃ (k : ℤ), a n = k := by
  sorry

File: FormalConjectures/OEIS/364176.lean [hunk 60]
```diff
@@ -0,0 +1,211 @@
...
+  let k_r := n * p ^ r
+  let k_r_minus_1 := n * p ^ (r - 1)
+  -- Define the modulus as a real number
+  let modulus : ℝ := (p ^ (3 * r)).cast
+  -- The premise is the conjectural integrality of the two relevant terms,
+  -- i.e., they are in the image of Int.cast
+  (a k_r ∈ Set.range (Int.cast : ℤ → ℝ)) ∧ (a k_r_minus_1 ∈ Set.range (Int.cast : ℤ → ℝ)) →
+  -- The conclusion is the divisibility condition: modulus divides the difference.
+  -- This is formalized as the quotient being an integer.
+  (a k_r - a k_r_minus_1) / modulus ∈ Set.range (Int.cast : ℤ → ℝ)
```
kghristova, line 208: same comment as in 364173. Again, maybe something like

```
theorem conjecture_2_supercongruence
  (p n r : ℕ) (hp : Nat.Prime p) (hp5 : 5 ≤ p) (hn : 0 < n) (hr : 0 < r) : 
  ∀ (n : ℕ), ∃ (k₁ k₂ : ℤ),  
  a (n * p ^ r) = k₁ ∧ a (n * p ^ (r - 1)) = k₂ ∧ k₁ ≡ k₂ [ZMOD (p : ℤ) ^ (3 * r)] := by
  sorry

File: FormalConjectures/OEIS/364178.lean [hunk 61]
```diff
@@ -0,0 +1,167 @@
...
+*References:*
+- [A364178](https://oeis.org/A364178)
+-/
+open Real
+open Nat
+
+namespace OeisA364178
+
+/--
+The sequence $a(n) = \frac{(10n)! (3n)! (n/2)!}{(6n)! (5n)! (3n/2)! n!}$ for even $n$.
```
kghristova, line 34: do we really only define this for n even?

FormalConjectures/OEIS/365416.lean
```diff
@@ -0,0 +1,221 @@
...
+def condition (k : ℕ) : Prop :=
+  IsPrimePow (2 * k - 1) ∧ IsPrimePow (2 * k + 1)
+
+/--
+$a(n)$ is the $n$-th integer $k$ such that $2k-1$ and $2k+1$ are both prime powers.
+-/
+noncomputable def a (n : ℕ) : ℕ :=
+  (n - 1).nth condition
+
+@[category API, AMS 11]
```
kghristova, line 43: Are these lemmas helpful to the conjecture?

FormalConjectures/OEIS/365416.lean
```diff
@@ -0,0 +1,221 @@
...
+-/
+def IsCompositePrimePow (m : ℕ) : Prop :=
+  ∃ (p e : ℕ), p.Prime ∧ 1 < e ∧ p ^ e = m
+
+/--
+According to Pillai's conjecture, k = 13 is the only term such that 2*k-1 and 2*k+1 both have
+exponent greater than 1.
+-/
+@[category research open, AMS 11]
+theorem only_thirteen :
```
kghristova, line 216: rename to conjecture

FormalConjectures/OEIS/365416.lean
```diff
@@ -0,0 +1,221 @@
...
+/--
+Predicate for a number to be a prime power with exponent strictly greater than 1.
+This is equivalent to being a composite prime power (a perfect power whose base is prime).
+-/
+def IsCompositePrimePow (m : ℕ) : Prop :=
+  ∃ (p e : ℕ), p.Prime ∧ 1 < e ∧ p ^ e = m
+
+/--
+According to Pillai's conjecture, k = 13 is the only term such that 2*k-1 and 2*k+1 both have
+exponent greater than 1.
```
kghristova, line 213: this conjecture is currently unrelated to the sequence.. Not sure if we want to discard it or if we keep it, it seems that $a(8) = 13$, so maybe we could rephrase this in terms of $a(8)$ to seem somewhat related?

FormalConjectures/OEIS/366833.lean
```diff
@@ -0,0 +1,250 @@
...
+namespace OeisA366833
+
+/--
+Number of times $n$ appears in A362965.
+-/
+noncomputable def a (n : ℕ) : ℕ :=
+  if n = 0 then 0
+  else
+    -- p_n (1-indexed) is Nat.nth Nat.Prime (n-1) (0-indexed). Since n > 0, n-1 is safe.
+    let p_n   : ℕ := Nat.nth Nat.Prime (n - 1)
```
kghristova, line 39: double space after p_n

FormalConjectures/OEIS/366833.lean
```diff
@@ -0,0 +1,250 @@
...
+    -- p_{n+1} is Nat.nth Nat.Prime n
+    let p_np1 : ℕ := Nat.nth Nat.Prime n
+
+    -- Count the number of prime powers in the inclusive interval [p_n, p_{n+1}]
+    let count_prime_powers : ℕ :=
+      Finset.card ((Finset.Icc p_n p_np1).filter IsPrimePow)
+
+    -- Subtracting 1 is safe since both p_n and p_{n+1} are prime powers, giving a count >= 2.
+    count_prime_powers - 1
```
kghristova, line 48: We can make this cleaner by combining the last two in

` Finset.card ((Finset.Ico p_n p_np1).filter IsPrimePow)` 

but then we might need to make some small changed to whatever comes after

FormalConjectures/OEIS/366833.lean
```diff
@@ -0,0 +1,250 @@
...
+  rw [prime_3, prime_4]
+  rw [filter_Icc_7_11]
+  rfl
+
+/--
+Conjecture: a(n) can be only 1, 2, or 3 (with the first occurrences of 3 appearing at n = 4, 9,
+30, 327 and 3512).
+-/
+@[category research open, AMS 11]
+theorem values_in_one_two_three : ∀ (n : ℕ), 1 ≤ n → a n ∈ ({1, 2, 3} : Finset ℕ) := by
```
kghristova, line 247: rename to "conjecture"

FormalConjectures/OEIS/369462.lean
```diff
@@ -0,0 +1,157 @@
...
+theorem a_3 : a 3 = 0 := a_eq_0_of_N_le_50 (by norm_num) (by norm_num)
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 0 := a_eq_0_of_N_le_50 (by norm_num) (by norm_num)
+
+/--
+Conjecture a: Is there only a finite number of 0's in this sequence?
+-/
+@[category research open, AMS 11]
+theorem finitely_many_zeros : answer(sorry) ↔ {n : ℕ | a n = 0}.Finite := by
```
kghristova, line 154: rename to "conjecture"

Also.. why not just {`n : ℕ | a n = 0}.Finite` ?

FormalConjectures/OEIS/369462.lean
```diff
@@ -0,0 +1,157 @@
...
+namespace OeisA369462
+
+/--
+Number of representations of $12n-1$ as $pq + pr + qr$ with odd primes $p \le q \le r$.
+-/
+noncomputable def a (n : ℕ) : ℕ :=
+  if 1 ≤ n then
+    let N : ℕ := 12 * n - 1
+    -- N is the target number. Since p*q < N, p, q, r are all bounded by N.
+    let B := N
```
kghristova, line 40: do we need both `N` and `B`?

FormalConjectures/OEIS/369462.lean
```diff
@@ -0,0 +1,157 @@
...
+    let search_range := range (B + 1)
+    let search_space := search_range.product (search_range.product search_range)
```
kghristova, line 42: Gemini is suggesting this optimisation of the def

```
noncomputable def a (n : ℕ) : ℕ :=
  if 1 ≤ n then
    let N := 12 * n - 1
    let range3N := Finset.Icc 3 N
    let search_space := range3N ×ˢ (range3N ×ˢ range3N)

    Finset.card (search_space.filter (fun ⟨p, ⟨q, r⟩⟩ =>
      p.Prime ∧ q.Prime ∧ r.Prime ∧
      p ≤ q ∧ q ≤ r ∧
      p * q + p * r + q * r = N
    ))
  else
    0

File: FormalConjectures/OEIS/370092.lean [hunk 71]
```diff
@@ -0,0 +1,94 @@
...
+  | m_plus_one@(m + 1) =>
+    -- The sum is over j from 1 to m_plus_one. k runs from 0 to m.
+    let sum_val : ℚ := ∑ k ∈ range m_plus_one,
+      let j : ℕ := k + 1 -- j is the index for the sum
+
+      let a_term : ℚ := a (m_plus_one - j)
+      let coeff_factor : ℚ := 1 - (-1 : ℚ)^j - (-2 : ℚ)^j
+
+      (m_plus_one.choose j : ℚ) * coeff_factor * a_term
+    (-1 : ℚ)^m_plus_one + (1 / 2) * sum_val
```
kghristova, line 48: the def is fine but a bit clunky. Gemini is suggesting this optimisation if we want to make it a bit more idiomatic 

```
noncomputable def a (n : ℕ) : ℚ :=
  match n with
  | 0 => 1
  | m + 1 =>
    let sum_val : ℚ := ∑ j ∈ Finset.Ico 1 (m + 2),
      ((m + 1).choose j : ℚ) * (1 - (-1 : ℚ)^j - (-2 : ℚ)^j) * a (m + 1 - j)
      
    (-1 : ℚ)^(m + 1) + (1 / 2) * sum_val

File: FormalConjectures/OEIS/370092.lean [hunk 72]
```diff
@@ -0,0 +1,94 @@
...
+-/
+noncomputable def a_mod_k (k : ℕ) (n : ℕ) : ZMod k :=
+  Int.cast (a n).num
+
+/--
+Conjecture: Let k > 2 be a positive integer. The sequence obtained by reducing a(n) modulo
+k is eventually periodic with the period dividing phi(k) = A000010(k).
+-/
+@[category research open, AMS 11]
+theorem bounds (k : ℕ) (hk : 2 < k) :
```
kghristova, line 90: rename to conjecture?

FormalConjectures/OEIS/374265.lean
```diff
@@ -0,0 +1,173 @@
...
+@[category test, AMS 11]
+theorem a_3 : a 3 = 6 := singleton_min' 6 _ reachable_3 _
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 24 := singleton_min' 24 _ reachable_4 _
+
+/--
+Is the sequence $a(n)$ bounded?
+
+The sequence is unbounded because deleting zero digits preserves the base-10 digit sum,
```
kghristova, line 164: so this is solved?

FormalConjectures/OEIS/374605.lean
```diff
@@ -0,0 +1,65 @@
...
+theorem a_4 : a 4 = 3116125 := by rfl
+
+/--
+Conjecture: for prime $p \ge 5$, $a(n)$ is divisible by $p^3$ for integer $n$ in the interval
+$[\lceil\frac{2p + 1}{3}\rceil, p - 1]$.
+The lower bound $\lceil\frac{2p + 1}{3}\rceil$ for $p \in \mathbb{N}$ is expressed using
+natural number division as $(2 * p + 1 + 2) / 3 = (2 * p + 3) / 3$.
+-/
+@[category research open, AMS 11]
+theorem supercongruence (p : ℕ) (hp : Nat.Prime p) (hp5 : 5 ≤ p) :
```
kghristova, line 59: rename to conjecture?

FormalConjectures/OEIS/374605.lean
```diff
@@ -0,0 +1,65 @@
...
+  ∀ n : ℕ,
+    (2 * p + 3) / 3 ≤ n →
+    n ≤ p - 1 →
+    (p ^ 3 : ℕ) ∣ a n := by sorry
```
kghristova, line 63: i don't think you need the `: ℕ `

FormalConjectures/OEIS/375178.lean
```diff
@@ -0,0 +1,82 @@
...
+
+open Nat Finset
+
+namespace OeisA375178
+
+/--
+The sequence $a(n) = \sum_{k=0}^{n-1} \binom{n+k-1}{k}^3$.
+-/
+def a (n : ℕ) : ℕ :=
+  ∑ k ∈ range n, (multichoose n k) ^ 3
```
kghristova, line 37: why not `(n + k - 1).choose k`? I'm not convinced `multichoose n k` is what we want here

FormalConjectures/OEIS/375178.lean
```diff
@@ -0,0 +1,82 @@
...
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 9065 := by simp [a, multichoose_eq, sum_range_succ]; decide
+
+/--
+The generalized sequence $b_m(n) = \sum_{k = 0}^{n-1} \binom{n+k-1}{k}^{2m+1}$.
+-/
+noncomputable
+def b (m : ℕ) (n : ℕ) : ℕ :=
+  ∑ k ∈ range n, (multichoose n k) ^ (2 * m + 1)
```
kghristova, line 59: same comment about multichoose

FormalConjectures/OEIS/375178.lean
```diff
@@ -0,0 +1,82 @@
...
+
+/--
+Conjecture: for a positive integer m, define a sequence $b_m(n)$ as
+$b_m(n) = \sum_{k = 0}^{n-1} \binom{n+k-1}{k}^{2m+1}$.
+We conjecture that the supercongruence $b_m(p) \equiv 1 \pmod{p^{2m+3}}$ holds
+for all primes $p \ge 2m + 5$, and for $r \ge 2$, the supercongruence
+$b_m(p^r) \equiv b_m(p^{r-1}) \pmod{p^{3r+2m+1}}$ also holds for all primes $p \ge 2m + 5$.
+-/
+@[category research open, AMS 11]
+theorem conjecture_2a (m : ℕ) (hm : 0 < m) (p : ℕ) (hp : Nat.Prime p) :
```
kghristova, line 69: p.Prime since we have Nat open

FormalConjectures/OEIS/375178.lean
```diff
@@ -0,0 +1,82 @@
...
+theorem a_4 : a 4 = 9065 := by simp [a, multichoose_eq, sum_range_succ]; decide
+
+/--
+The generalized sequence $b_m(n) = \sum_{k = 0}^{n-1} \binom{n+k-1}{k}^{2m+1}$.
+-/
+noncomputable
+def b (m : ℕ) (n : ℕ) : ℕ :=
+  ∑ k ∈ range n, (multichoose n k) ^ (2 * m + 1)
+
+/--
```
kghristova, line 61: we're missing the conjecture about a(n)! `We conjecture that, in fact, the stronger congruence a(p) == 1 (mod p^5) holds for all primes p >= 7.`

```
theorem conjecture_1 (p : ℕ) (hp : p.Prime) (hp_ge_7 : 7 ≤ p) : 
a p ≡ 1 [MOD (p ^ 5)] := by sorry

File: FormalConjectures/OEIS/376462.lean [hunk 80]
```diff
@@ -0,0 +1,86 @@
...
+@[category test, AMS 11]
+theorem a_2 : a 2 = 109 := by rfl
+
+@[category test, AMS 11]
+theorem a_3 : a 3 = 3317 := by rfl
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 121501 := by rfl
+
+/--
```
kghristova, line 63: I think split this into two conjectures as we have been doing and write the docstring in terms of a (n) rather than the Apery numbers. i.e.  (also remove the weird casts to \Z)

```
/--
For all primes $p \ge 5$ and all
positive integers $n$ and $r$, $a (n * p^r) \equiv a(n * p^{r-1}) \pmod{p^{3r}}$ 
-/
theorem conjecture _1 :
  ∀ (p n r : ℕ),
    p.Prime →
    5 ≤ p →
    0 < n →
    0 < r →
      a (n * p ^ r)  ≡ a (n * p ^ (r - 1))  [MOD (p ^ (3 * r)] := by sorry 

  
/--
For all primes $p \ge 5$ and all
positive integers $n$ and $r$, $a(n * p^r - 1) \equiv a(n * p^{r-1} - 1) \pmod{p^{3r}}$
-/
theorem conjecture _2 :
  ∀ (p n r : ℕ),
    p.Prime →
    5 ≤ p →
    0 < n →
    0 < r →
      a (n * p ^ r - 1) ≡ a (n * p ^ (r - 1) - 1) [MOD p ^ (3 * r)]  := by sorry

File: FormalConjectures/OEIS/376930.lean [hunk 81]
```diff
@@ -0,0 +1,78 @@
...
+
+/--
+$a(0)=0, a(1)=1$; for $n>1$, $a(n) = a(n-1)+a(n-2)$, except when $a(n-1)$ is an odd prime, in
+which case $a(n) = a(n-1)-a(n-2)$.
+-/
+def a : ℕ → ℕ
+| 0 => 0
+| 1 => 1
+| n + 2 =>
+  let an_1 := a (n + 1)
```
kghristova, line 40: not sure if these lets are adding much to the formalization

FormalConjectures/OEIS/376930.lean
kghristova: Usually you get a type mismatch. It won't automatically change the int to a nat. I think we should definitely change the type in the def.

FormalConjectures/OEIS/376930.lean
```diff
@@ -0,0 +1,78 @@
...
+It is not known if the sequence contains any negative terms (which may happen if two primes are
+adjacent or separated by one other term).
+
+Formalization: Since the sequence is defined in $\mathbb{N}$, all terms are non-negative by
+definition. The conjecture's mathematical content is that whenever the subtraction rule $a(n+2)
+= a(n+1) - a(n)$ is applied, the result in $\mathbb{Z}$ is non-negative. In the context of
+$\mathbb{N}$ arithmetic, this is equivalent to asserting that $a(n+1) \ge a(n)$.
+-/
+@[category research open, AMS 11]
+theorem monotonically_increasing_at_primes :
```
kghristova, line 75: change name to "conjecture_1"?

FormalConjectures/OEIS/376930.lean
```diff
@@ -0,0 +1,78 @@
...
+  ∀ n : ℕ, (Nat.Prime (a (n + 1)) ∧ a (n + 1) > 2) → a (n + 1) ≥ a n := by sorry
+
```
kghristova, line 77: There seem to be 2 more conjectures we haven't formalised:

- If a(n) continues to grow, it is not clear whether this sequence will contain an infinite number of prime numbers.
- Beyond the trivial case of 1, it is not clear if any number will appear more than three times in the sequence.

FormalConjectures/OEIS/377224.lean
```diff
@@ -0,0 +1,93 @@
...
+/-- The quadratic form $t(5t+1)/2$, which is an integer for all $t \in \mathbb{Z}$. -/
+def Q2 (t : ℤ) : ℤ := (t * (5 * t + 1)) / 2
+
+/--
+Number of ways to write $n$ as $x(5x+1) + y(5y+1)/2 + z(5z+1)/2$ with $y(5y+1) \le z(5z+1)$.
+-/
+def a (n : ℕ) : ℕ :=
+  let N : ℤ := n
+  -- A conservative absolute bound for the variables is $n+1$.
+  let B : ℤ := n + 1
```
kghristova, line 46: Not sure what these extra variables are adding

FormalConjectures/OEIS/377224.lean
```diff
@@ -0,0 +1,93 @@
...
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 2 := by decide
+
+/--
+a(n) = 0 only for n = 1.
+Also, a(n) = 1 only for n = 0, 2, 3, 5, 7, 14, 16, 19, 37, 43, 58, 61, 79.
+-/
+@[category research open, AMS 11]
+theorem a_eq_one_iff :
```
kghristova, line 88: change name to "conjecture_1"

FormalConjectures/OEIS/377224.lean
```diff
@@ -0,0 +1,93 @@
...
+  (∀ (n : ℕ), a n = 0 ↔ n = 1) ∧
+  (∀ (n : ℕ), a n = 1 ↔ n ∈ ({0, 2, 3, 5, 7, 14, 16, 19, 37, 43, 58, 61, 79} : Finset ℕ)) :=
+by sorry
+
```
kghristova, line 92: there is also conjecture 2 which is not formalised

FormalConjectures/OEIS/379643.lean
```diff
@@ -0,0 +1,174 @@
...
+  let count_primes_mod_b (b : ℕ) : ℕ :=
+    ((range (p_n + 1)).filter (fun p => p.Prime ∧ p % 8 = b)).card
+
+  (count_primes_mod_b 5 : ℤ) - (count_primes_mod_b 1 : ℤ)
+
+/-- Conjecture: no prime appears on the negative y-axis.
+That is, for every $n \ge 1$, if the $x$-coordinate $a(n)$ is $0$, then the $y$-coordinate $b(n)$
+must be non-negative. -/
+@[category research open, AMS 11]
+theorem not_zero_and_negative : ∀ (n : ℕ), 0 < n → ¬ (a n = 0 ∧ b n < 0) := by sorry
```
kghristova, line 172: rename to conjecture?

FormalConjectures/OEIS/381159.lean
```diff
@@ -0,0 +1,196 @@
...
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 4 := by unfold a; rw [h4]
+
+/--
+51st All-Russian Mathematical Olympiad for Schoolchildren. Problem.
+Let us call a natural number "lopsided" if it is greater than 1 and all its prime divisors end
+with the same digit.
+Is there an increasing arithmetic progression with a difference not exceeding 2025,
+consisting of 150 natural numbers, each of which is "lopsided"? (A. Chironov)
```
kghristova, line 183: apparently this has been disproved:
https://arxiv.org/pdf/2608.11941

FormalConjectures/OEIS/381358.lean
```diff
@@ -0,0 +1,93 @@
...
+@[category test, AMS 11]
+theorem a_4 : a 4 = 3 := by
+  symm
+  norm_num [a]
+  norm_num [t]
+  simp_all [run_lengths_nat]
+
+/--
+If it exists, the limit of $\mathrm{a}(n)^{1/n}$ as $n \to \infty$.
+The conjecture is that this limit exists.
```
kghristova, line 86: Also find a value for it, right?

It's apparently resolved here too : https://arxiv.org/abs/2608.11941

FormalConjectures/OEIS/381358.lean
```diff
@@ -0,0 +1,93 @@
...
+-/
+@[category research open, AMS 11]
+theorem limit_exists :
```
kghristova, line 89: change name to "conjecture_1" or something?

FormalConjectures/OEIS/383327.lean
```diff
@@ -0,0 +1,73 @@
...
+theorem a_4 : a 4 = 4 := by rfl
+
+/--
+Conjecture based on OEIS comment:
+For $n \le 100$, when $n \in \{1, 3, 5, 9, 15, 23, 35, 63, 65, 69\}$, there is only one $m$
+such that $A049802(m) = n$, so $a(n) = 1$.
+
+Note: The OEIS comment lists $67$ instead of $69$, but this is a typo in the OEIS comment:
+the sequence terms in OEIS have $a(67) = 5$ (e.g. $A049802(147) = A049802(2055) = 67$) and
+$a(69) = 1$.
```
kghristova, line 65: is this a conjecture though or is it known?

FormalConjectures/OEIS/385391.lean
```diff
@@ -0,0 +1,102 @@
...
+
+/-- A384237: Number of divisors $d$ of $n$ such that $d^d \equiv d \pmod n$. -/
+def A384237 (n : ℕ) : ℕ :=
+  (n.divisors.filter fun d : ℕ => (d ^ d) % n = d % n).card
+
+/--
+The smallest integer $k$ such that $A384237(k) = n$.
+-/
+noncomputable def a (n : ℕ) : ℕ :=
+  sInf {k : ℕ | A384237 k = n}
```
kghristova, line 40: Don't need the explicit type annotation `k : ℕ` here

FormalConjectures/OEIS/385391.lean
kghristova: Agree with Codrut

FormalConjectures/OEIS/386660.lean
```diff
@@ -0,0 +1,65 @@
...
+
+@[category test, AMS 11]
+theorem a_3 : a 3 = 5 := by rfl
+
+@[category test, AMS 11]
+theorem a_4 : a 4 = 7 := by rfl
+
+@[category test, AMS 11]
+theorem a_5 : a 5 = 11 := by rfl
+/--
```
kghristova, line 55: new line before the docstring missing

FormalConjectures/OEIS/386660.lean
```diff
@@ -0,0 +1,65 @@
...
+The limit of $a(n)^{1/n}$ exists.
+The numerical evidence suggests a limit of approximately $1.7086...$
```
kghristova, line 57: do we need this final sentence?

FormalConjectures/OEIS/386888.lean
```diff
@@ -0,0 +1,121 @@
...
+@[category test, AMS 11]
+theorem a_5 : a 5 = 0 := a_zero_for_small 5 (by decide)
+
+/--
+If n is an odd number greater than 905, or an even number greater than 1466, then we have a(n) > 0.
+Also, a(n) > 1 for all n > 2258.
+(In the case k = m = 3 for the general conjecture by Sun.)
+-/
+@[category research open, AMS 11]
+theorem eventually_positive :
```
kghristova, line 114: this is conjecture 2. What about conjecture 1? i.e.


For m > 0, let P(m) denote the set of all sums of m consecutive primes. We make the following general conjecture motivated by Goldbach's conjecture.
Conjecture 1: Let k and m be positive integers.
(i) Each sufficiently large integer n == k + m (mod 2) can be written as p + q, where p and q belong to P(k) and P(m) respectively.
(ii) If k is odd, then every sufficiently large integer n == m (mod 2) can be written as 2*p + q, where p and q belong to P(k) and P(m) respectively.

PR #4948, opened 2026-08-14
Balance the math delimiters in 28 docstrings

williamjblair:
Rendering every problem file as LaTeX found 28 docstrings whose `$` do not pair. A missing
closing delimiter sets the rest of the sentence in maths on the generated site, and the build
does not see it.

Nine are prize amounts: `($500)` and `($250)` in Erdős 593 and 595, and five OEIS entries
writing "offered a $2,400 prize". Each opens maths mid sentence, and each becomes `\$`.

Three close with a backtick instead of a dollar. Erdős 697 writes
`$\delta < \frac{m ^ \alpha + 1}{m}` `, and Erdős 884 writes `for $n \to \infty` ` in each of
its two copies of the same paragraph. That sentence in 884 also closes `\left(` with both a
bare `)` and a `\right)`, so the rendered formula gains a paren. The `\ll` form of the same
expression two lines above has no paren there, which is what settles it.

Two are misspellings. ErdosProblems/30 writes `O_{\varespilon}(N^\varespilon)`, wrong twice on
one line, and never closes the group.

One needed the source rather than a guess. OEIS/115366 read `$a(n)$A006880(n) \rightarrow
1.77...$`, where the stray delimiter could have been a slash, a comma or a product. A115366's
own comment reads `Conjecture: a(n)/A006880(n) -> 1.77...`, so it is a ratio.

The rest are missing closers, a doubled `$$`, a stray `$` in `$f(z) ∈ $ℂ[z]$`, and a missing
opener in GreensOpenProblems/4.

Docstrings only: 28 changed lines, none of them Lean. All 26 modules build with `--wfail`.

### DIFF
--- a/FormalConjectures/Arxiv/1609.08688/sIncreasingrTuples.lean
+++ b/FormalConjectures/Arxiv/1609.08688/sIncreasingrTuples.lean
@@ -76,7 +76,7 @@ theorem lt₂_example_2 : ![5, 6, 1] <₂ ![7, 7, 7] := ⟨0, 2, by simp, by sim
 @[category test, AMS 5]
 theorem lt₂_example_3 : ![7, 7, 7] <₂ ![7, 8, 9] := ⟨1, 2, by simp, by simp⟩
 
-/-- but $(1, 2, 3)$ is not $2$-less than $(1, 2, 4). -/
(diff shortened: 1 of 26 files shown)


mo271 (approved):
Thanks, LGTM!

Issue #2143, opened 2026-02-03
Implement attribute for Optimization Constants


Paul-Lez:
See discussion here: https://github.com/google-deepmind/formal-conjectures/pull/1899#discussion_r2734051775

I intend to have a go at this.

mo271:
Cool, @Paul-Lez! I think a first version could be very basic, we can always add stuff later

BoltonBailey:
Hey @Paul-Lez did you make any progress on this? If not, would you mind if I made a PR? I'd be interested in trying to realize Felix's proposal of making ties between the optimization constants repo and formal-conjectures.

Paul-Lez:
Hey Bolton that sounds great - I haven't had time to do this yet!

mo271:
@BoltonBailey go for it!

PR #4950, opened 2026-08-14
feat: add "optimization_constant" attribute

BoltonBailey:
This PR adds an attribute for "optimization constants" tracked by https://teorth.github.io/optimizationproblems/ 

As discussed in #2143 and other linked places from there, it seems like this would be nice way to track connections between that repo and formal-conjectures.


BoltonBailey:
There's maybe an interesting question of what we want the attribute to apply to. Maybe we just want to apply it to definitions of types with a linear order, and then make another attribute or a variant of this attribute to denote theorems on upper bounds and lower bounds / open questions about `= answer(sorry)`?

The drawback I think is that some of these optimization constant theorems are expressed not as inequalities to a separately-defined constants, so we might have to introduce new definitions. On the other hand, maybe making a habit of doing that would be a good thing because it would make it possible to do more analysis of these constants as a group?

Issue #4930, opened 2026-08-13
Formal Conjectures and LeanEval: integration and proof packages


williamjblair:
## Goal

Add Formal Conjectures to LeanEval as a separate problem set, reusing LeanEval and Comparator rather than building another evaluation system. FC maintainers keep control of statements, metadata, and acceptance; build and Comparator results are evidence for review.

**Superseded in part by [lean-eval#536](https://github.com/leanprover/lean-eval/pull/536) (19 Aug).** That plan settles several questions this issue originally asked, and its §10 differs from what was proposed here. Updated below to record what is settled and narrow this issue to what is still open.

## Settled by lean-eval#536 §10

- **Group scope.** Only open conjectures. FC's research-solved statements do not get their own policy or tab; they are a future source of candidates for the draft group. *This issue originally proposed two frozen lists, one solved and one open. That is superseded.*
- **Submission policy.** Public submissions required from the start, no embargo, no opt-out; submission URL and exact commit public, with the accepted snapshot archived. *This issue originally proposed keeping solved-list submissions private until a fixed release time. That is superseded.*
- **Generator ownership.** One shared generator, extracted from `EvalTools` into `leanprover/lean-eval-generator` and consumed as a pinned dependency by both lean-eval and the FC importer, which does not fork generation logic. *This answers the "which workspace code lives where" question.*
- **Hosting and importer.** FC problems vendored into lean-eval; LeanEval remains the trusted statement repository and supplies the pin regime and CI; the FC side owns an importer (#4951) that maps declarations and metadata to LeanEval modules and manifests and emits PRs that lean-eval CI validates. Each manifest records the FC source commit and declaration id.
- **Merge rights.** #536 proposes giving FC contributors merge rights for problem PRs in the group.
- **First set.** The frozen FC100 open list.

## Still open

Measurements below come from the 415 existing `formal_proof` links at FC `9f5ee77`, resolved to 143 (repository, revision) pairs and cloned; all 143 still resolve. 16 of the 415 point at gists or forum threads rather than repositories, so 399 are assessable. These are properties of links, not judgments about anyone's mathematics.

**What fields belong in the v1 submission format?**

142 of 415 links (34%) name a branch or tag rather than a commit, so the evidence can move after the link is written. Requiring an immutable revision removes that class.

[...]

PR #4428, opened 2026-07-14
chore: bump to v4.32.2

YaelDillies:
Depends on #2356, #3691, #4025 and #4356

PR #4951, opened 2026-08-14
Superseded: generate a LeanEval workspace from an FC declaration

williamjblair:
**Superseded by #5337.** New development uses native Lean extraction and the shared generator; CLI orchestration is #5387. Keep this PR open as historical context until #5337 lands upstream, then close it as superseded. Do not extend the source-reconstruction implementation. Roadmap: #4394.

The earlier proposal is retained below as history.

---

**Tracking:** #4394 · **Integration:** #4930 · **Architecture RFC:** #5158
**Upstream plan:** [lean-eval#536](https://github.com/leanprover/lean-eval/pull/536) workstream 9 · coordination [lean-eval#533](https://github.com/leanprover/lean-eval/issues/533)

This PR adds the FC-side importer from §10 of the LeanEval plan. It turns one FC declaration into one evaluation workspace, through a versioned contract with a pinned external generator.

# 1. The system

```mermaid
flowchart TB
    subgraph FC["Formal Conjectures — this PR"]
        A["Declaration<br/>erdos_730, @[category ...]"]
        B["Elaborated facts<br/>ranges · binders · slot types"]
        C["Marked-up module<br/>statement + copied closure"]
        A --> B --> C
    end
    C --> R["Request JSON<br/>schema version 1"]
    R --> G
    subgraph G["lean-eval-generator — pinned 77373a53"]
        GG["Render workspace"]
    end
    GG --> WS["Workspace file map<br/>+ FC provenance sidecar"]
    WS --> LE
    subgraph LE["LeanEval"]
        LL["Catalog · target pins · scoring"]
    end
    LL --> SOL["Solver writes Submission.lean"]
    SOL --> CMP
    subgraph CMP["Comparator"]
        CC["Sandbox · export · identity<br/>+ axiom check · kernel replay"]
    end
    CC --> V(["Verdict"])
```

Each stage owns one thing, and the artifact between stages is checkable.

| Stage | Owns | Hands over |
|---|---|---|
| **Formal Conjectures** | The trusted statement and where it came from | A request: one Mathlib-only Lean module per problem, hole spans, and the target pins to build at |
| **lean-eval-generator** | Turning a module into a workspace, deterministically | A file map, each file with its SHA-256 |
| **LeanEval** | The catalog, the target pins, lifecycle and scoring | A workspace a solver can clone |
| **Comparator** | Judging a submission adversarially | A verdict: accept or reject |

**This PR is the first box.** It does not generate workspaces, score submissions, or judge proofs. The placeholder generator it once carried is deleted, not forked.

# 2. Why the module is Mathlib-only

LeanEval vendors its problems. A Challenge cannot fetch this repository when a solver builds it. So the statement cannot import its own FC module, and every FC declaration it depends on must travel with it.

The importer therefore copies the declaration's FC-local closure into the module, each copied declaration carrying the `open`, `variable`, `universe`, `set_option` and `local notation` in force where it was written. Copying is a construction, so it can be wrong in ways only Lean sees. `--verify` elaborates the assembled module before anything is written.

# 3. The generated workspace

```mermaid
flowchart LR
    CD["ChallengeDeps.lean<br/>copied FC closure"] --> CH["Challenge.lean<br/>trusted statement + hole"]
    CH --> SO["Solution.lean<br/>fixed adapter"]
    SU["Submission.lean<br/>the solver works here"] --> SO
    SO --> WT["WorkspaceTest.lean"]
    CF["config.json<br/>targets + permitted axioms"] --> WT
    WT --> CMP(["Comparator"])
```

`Solution.lean` is fixed. It fails to build if a submission changes the statement. `config.json` names the permitted axioms, so a submission that leaves `sorry` reports `sorryAx` and is rejected. The FC provenance sidecar rides beside these files and records where the statement came from.

# 4. Two toolchains, on purpose

The importer reads facts from an environment elaborated at this repository's **Lean 4.33.1**. The workspace builds at LeanEval's **Lean 4.33.0**. The gap is one Mathlib patch release, and CI builds across it, so the gap is observed rather than assumed.

An FC-side copying defect fails at `--verify`, on this side. A genuine target incompatibility fails at the target build. The two gates tell those apart.

The `import Mathlib` header is faithful, not convenient: every problem file imports `FormalConjecturesUtil`, which `public import`s all of Mathlib. Each statement already elaborates under the full library. The header drops only the FC-local layer, which travels as the copied closure.

# 5. Result at this head

**100/100 declarations import, verify, classify (92 open + 8 solved), generate, and compile at target pins. `comparator/known_failures.toml` is empty, and the audit asserts exactly that.**

# 6. Guarantees at the boundary

Each row is a check, not a convention.

| Guarantee | Mechanism |
|---|---|
| One source snapshot per workspace | `pins()` holds every source file read to one revision: the statement, its copied dependencies, its copied notation. An untracked file refuses. The toolchain and manifest are recorded as the environment observed, not held to the pin, so a toolchain bump can still run these jobs |
| Emitted bytes are generated bytes | One serialisation is piped, written by `--emit-import`, and recorded as `digests.request`. The seam test replays it byte for byte |
| The response cannot escape its workspace | Unsafe paths, unknown fields, missing fields, and missing or extra workspace ids all refuse. A `--set` batch validates fully before the first write |
| Recorded axioms are enforced axioms | The generated `config.json` is compared against the manifest at generation |
| Catalog policy is stated, not assumed | Group, status, visibility, revision, submitter and tags arrive as an explicit `ImportPolicy` |
| Provenance names its producers | Each sidecar records copied slices with digests, the statement as range plus digest, and the importer, generator and target pins |

# 7. CI

| Job | Runs on | Asserts |
|---|---|---|
| Adapter unit tests | Every PR | Python only, no Lean build, 14 seconds |
| Adapter boundaries | `comparator/**` | A smoke set covering each defect class the audit found; the emitted request regenerates identical files; three Comparator verdicts |
| FC100 source audit | Changes to what it measures: adapter, templates, problems, pins, ledger, actions, subset file, toolchain files. Also weekly and on dispatch | All 100 import, verify at FC pins, classify and generate. Failures match the ledger exactly |
| FC100 target compile | Weekly and on dispatch, against the set the audit generated | All 100 Challenges compile at LeanEval's pins |

The adapter's checks live with the adapter, not beside the corpus build. A pull request that only edits a statement should not pay for them, and could not pass them: `pins()` refuses a source file that differs from the merge base, which is exactly what such a pull request changes.

The three verdicts exist because a gate that only ever accepts proves nothing:

1. The workspace as generated must be **rejected**. Its Submission is `sorry`, which adds `sorryAx`, which the config does not permit.
2. The same workspace, actually proved, must be **accepted**.
3. An `answer(sorry)` hole filled with the proposition from the other side of the iff, closed by `Iff.rfl`, is **accepted** and resolves nothing. A definition hole is where a machine check cannot stand in for a human reading the answer. The generated README says so; this asserts it.

The audit is split because its halves cost differently. The source half is one Mathlib, this repository's, and it is where an FC-side defect shows, so it gates every pull request that touches the adapter. The target half is a second Mathlib at LeanEval's toolchain, and it catches drift in the other repository rather than a defect here, so it runs weekly against the set the source half generated.

The bundle is the evidence, not a summary: the exact request bytes, every workspace with its sidecar, the report, and a manifest digesting each file. `make_comparator_workspace.py --set` is the same run locally.

# 8. Audit response

All findings from the [whole-set audit](https://github.com/google-deepmind/formal-conjectures/pull/4951#issuecomment-5357768564) are closed.

| Finding | Disposition |
|---|---|
| Guillemet modules split on their dots | Fixed. A round-trip property test over ~1,100 real modules holds the codec |
| Two `conjecture` declarations shared one id | Fixed. Ids are qualified names; batch import asserts global uniqueness; the per-PR job generates the real collision pair |
| 92 open + 8 solved in an "open" set | By design, per #5075. The frozen list keeps every member in the open-conjectures group. The category rides along as a tag |
| 11 source-extraction failures | All fixed: file-scoped notation, macros and `noncomputable section` carried with the preamble; FC notation copied only where it was in force; namespaces created before any copied `open`; statement-rooted `match` auxiliaries accepted; ascribed slots read at their own position |
| `irrational_e_to_e` failed at target pins | Not Mathlib drift. A dropped `local notation "e" => exp 1` was masked at FC pins by auto-bound implicits. The preamble fix closed it |
| `Erdos1092` failed at target pins | Was genuine instance drift. Retired by the 4.33.1 bump (#4428). The ledger is now empty |

Dotted names surfaced during migration. The generator anchors on a declaration's last component, so `erdos_1038.parts.i` would generate `theorem i`. Dotted names are restated under their slug, and the provenance records the FC name.

# 9. Files to review

| Module | Role |
|---|---|
| `fc_source.py` | Reads this repository's Lean; strict boundary for the extractor payload |
| `fc_leaneval_importer.py` | Declaration → module + provenance |
| `comparator_facts.lean` | Ranges, binders, slot types, category tags, from the elaborated environment |
| `leaneval_interface.py` | Builds the request, checks the response |
| `leaneval_generator_cli.py` | Runs the pinned binary |
| `known_failures.py` | The ledger format, shared by both gates |
| `templates/WorkspaceTest.lean` | FC-supplied; the contract requires the consumer to provide it |

# 10. Open questions

1. **Provenance in the contract.** Schema version 1's `source` is one free-text line. The FC commit, blob and declaration id that §10 requires have no home, so they travel as a sidecar. Should a future contract carry a passthrough provenance object? The `formalization.yaml` overlap belongs here: the question is which object owns which fields.
2. `definition_names` is checkable only against the comparator commit pinned in `tools.toml`.
3. Which side watches FC commits for changes to imported declarations is unassigned. The sidecar records what is needed to answer it.

[lean-eval-generator#4](https://github.com/leanprover/lean-eval-generator/issues/4) proposes three upstream hardening items. None changes contract v1. Nothing here waits on them.

# 11. Not in scope

Disproof support (Wave 4). Opening problem PRs to lean-eval, which waits on the launch gates and on consolidation with @Paul-Lez's prototype. Editing `FC100OpenSet1.lean`, resolved by #5075.

The source-reconstruction half of the adapter is a compatibility boundary, not a design. It recovers facts the elaborated environment already knows, because this repository has no export boundary yet. #5158 proposes the manifest that would let it be deleted. Nothing in that RFC blocks this PR.


kim-em:
Upstream coordination update: the shared package is now public at https://github.com/leanprover/lean-eval-generator, with the frozen v1 contract at `a726789593eeac5c32ad82760061cd5bf6cae662`; LeanEval PR leanprover/lean-eval#540 pins that exact commit. The boundary is a versioned JSON CLI/schema rather than a Python import: the FC-owned adapter supplies exact marked module bytes, resolved hole ranges/dependencies, target pins, template bytes, and the read-only context root, and receives the complete generated file map with SHA-256 digests. This matches the ownership split described here while avoiding a second generator implementation. Whole-FC100 answer-slot validation under LeanEval target pins and the dependency audit remain launch gates; no FC import is being enabled by this update.

kim-em:
Whole-FC100 audit against this PR completed at FC head `9f5ee773841921f460b4a26a3552f5eca4accaa0` and the configured LeanEval target (`leanprover/lean4:v4.33.0`, Mathlib `6f1ef4e5dd604a435bddba4747b13970cd65d2a1`). No FC files were edited.

Source inventory evidence:

- `lake build FormalConjecturesAnswerPostpone comparator_facts`: 9,170 jobs, success.
- `FC100OpenSet1`: 100 declarations = 92 `research open` + 8 `research solved`.
- answer slots: 48 `Prop` + 13 non-`Prop` = 61.

The exhaustive importer run found three set-wide contract defects before per-declaration results:

1. `module_source_path()` splits on every dot, including dots inside guillemet-quoted module components. For example `FormalConjectures.Arxiv.«0912.2382».CurlingNumberConjecture` is decoded as the nonexistent `FormalConjectures/Arxiv/091/2382»/CurlingNumberConjecture.lean`. I continued the audit with an in-memory environment-derived module/path map.
2. Default IDs are not unique: `OeisA303656.conjecture` and `OeisA308734.conjecture` both become workspace ID `conjecture`.
3. The advertised open set itself still contains the eight solved entries, as already noted in this PR.

With only the module decoder shim, source-side `--verify` equivalent checks passed for 89/100. The 11 failures were:

- `Erdos125.erdos_125.variants.positive_unequal_density`: unresolved `.lowerDensity` / `.upperDensity` field notation;
- `Erdos324.erdos_324`: generated `match_1` has no copied ancestor;
- `Erdos332.erdos_332`: copied-module type mismatch;
- `Erdos890.erdos_890.parts.a`: function expected;
- `Erdos92.erdos_92.variants.weak`: copied superscript notation does not parse;
- `Green9.green_9_iii`: expected token / application mismatch;
- `PoincareConjecture.poincare_conjecture.variants.smooth_dimension_four`: copied superscript notation does not parse;
- `OpenQuantumProblem23.hasSICPOVM_60`: copied dependency requires `noncomputable`;
- `EllipticCurveRank.RatEllipticCurve.twentyone_le_rank_height_count_asymptotic`: unknown copied namespace;
- `Koethe.KotherConjecture.variants.matrixOver_KotherRadical`: ambiguous copied term;
- `RamseyNumbers.ramsey_number_five_five`: malformed copied declaration around a parenthesized binder.

I then compiled every source-valid generated `ChallengeDeps` + `Challenge` pair in one shared target project at LeanEval pins. Result: 87/89 passed (intentional `sorry` warnings only). Two are genuine source/target compatibility failures:

- `Erdos1092.f_asymptotic_general`: its copied `f` definition no longer synthesizes `Fintype ↑H.verts` / `Fintype ↑H.coe.edgeSet` at the target pin.
- `Irrational.irrational_e_to_e`: the generated target statement has unknown identifier `e` at both occurrences.

So the whole-set answer-slot/target audit is now reproduced, and it currently fails closed for 13 declarations total (11 source extraction, 2 target compatibility), in addition to the module decoder, ID collision, and 92/8 set correction. The two pilot declarations remain useful, but do not cover these cases. I recommend making the corrected frozen-set whole-run a CI artifact before enabling LeanEval import.

kim-em:
FC100 dependency audit reproduced at current PR head `a101a9c72d2abec5f62f4fcd93b2c0b65c6cec5e` after a clean `lake build FormalConjecturesAnswerPostpone comparator_facts` (9,178 jobs, success). All 100 set declarations resolved through the elaborated environment. The statement closures contain 280 dependency edges, 158 unique FC constants across 74 FC modules, and 56 elaborator-generated auxiliary edges. No FC100 target depends on another FC100 target.\n\nOne closure contract violation remains: `Erdos324.erdos_324` needs generated constant `Erdos324.erdos_324.match_1`, but the generated constant has no copied source ancestor. The importer correctly fails closed on this; it is the same Erdos324 failure seen in the whole target audit. All other 99 elaborated dependency closures passed the structural checks (unique names, FC-local module ownership, source ranges, and generated-ancestor coverage). Canonical audit report SHA-256: `849dd1b89be95bec6562ea2b6a51b44e714570723745474d7c35264b279305db`.\n\nThis completes reproduction of the dependency audit, not import authorization. The corrected frozen-set CI job should retain this whole-set closure check, and import remains blocked on the previously reported 11 extraction failures, two target-pin incompatibilities, dotted-module decoder, ID collision, and 92-open/8-solved set correction.

williamjblair:
Thank you for both audits. Every finding is addressed at the new HEAD, and I’ve restructured the PR around the published v1 contract.

**Current whole-set result: 100/100 declarations import, classify, and generate successfully through the pinned v1 generator.** The only expected target-side failure is the known Lean/Mathlib drift for `Erdos1092.f_asymptotic_general`.

### 1. v1 contract migration

The integration seam is now the pinned `lean-eval-generator` at `a7267895`, configured under `[generator]` in `comparator/tools.toml`.

The adapter now:

- builds the v1 request with module content, resolved hole spans, category-derived `group`, and LeanEval pins;
- materialises the context root expected by the contract, including the module file and a synthesised `.ilean` containing exactly the computed spans;
- verifies every response digest.

The placeholder generator has been removed.

`--emit-import` now emits the actual v1 request. CI replays that request through the pinned generator and verifies that its file map matches direct generation.

### 2. Set-wide audit findings

**Module decoding:** fixed. `module_name` and `module_source_path` now use the same guillemet-aware codec, with a round-trip property test across every real FC module.

**Workspace ID collisions:** fixed. Default IDs are now qualified declaration names, e.g. `OeisA303656_conjecture` and `OeisA308734_conjecture`. Batch import asserts global uniqueness, and PR CI explicitly exercises the real collision pair plus a declaration from a guillemet-containing module.

**92/8 classification:** fixed at import time. The importer reads each declaration’s elaborated `@[category ...]` annotation and maps it to the v1 group:

- research/open → `open-conjectures`
- solved/textbook/test → `formalization-evaluation`
- anything else → rejected

The whole-set report records the resulting 92/8 split. The underlying FC set-file question is maintainer-owned and is raised separately in #5075.

### 3. All 11 source-generation failures are fixed

They fell into five classes:

- **Missing file-scoped context:** `local notation`, `local macro`, and `noncomputable section` declarations now travel with the preamble. This fixes Irrational, Poincaré, Ramsey, Erdos125, OpenQuantumProblem23, and Erdos890.
- **FC-defined notation:** notation is copied only when its token is used and the relevant scope was actually active. Scoped commands are retained only where the module opens their namespace, which prevents the modal-logic `⊆` overload from leaking into set statements. This fixes Erdos92, Poincaré, and Green9.
- **Namespace ordering:** required namespaces are created before copied `open` commands. This fixes EllipticCurveRank and Koethe.
- **Elaboration-generated structure:** statement-rooted `match_1` auxiliaries are accepted because re-elaborating the restated statement regenerates them. This fixes Erdos324 and the corresponding dependency-audit violation.
- **Answer ascriptions:** `(answer(sorry) : T)` is now interpreted at its own source position, where elaboration otherwise loses the annotation. This fixes Erdos332.

### 4. The two target failures

**`irrational_e_to_e`: fixed.** This was not Mathlib drift. The importer had dropped:

`local notation "e" => exp 1`

FC’s current pins happened to mask the issue through auto-bound implicits. Preserving the preamble fixes it.

**`Erdos1092.f_asymptotic_general`: genuine dependency drift.** This is a `Fintype` instance difference between the FC and LeanEval environments. It is recorded in `comparator/known_failures.toml`.

FC’s planned Lean 4.33 bump should retire the exception, and the audit will fail once the entry becomes stale so it cannot silently remain.

### 5. Whole-set CI

`.github/workflows/fc100-audit.yml` now reproduces the full audit on dispatch, weekly, and whenever its own configuration changes.

It performs:

1. import + `--verify` of the complete FC set at FC pins;
2. one v1 request through the pinned generator;
3. compilation of every generated `Challenge` under the shared-project LeanEval environment;
4. exact comparison of failures in both stages against `known_failures.toml`.

Because dispatch/scheduled workflows cannot run before the workflow exists on the default branch, the path trigger ensures the audit also runs on this PR.

Its artifact contains both the per-declaration report and the emitted v1 request.

**Local result at current HEAD: 100/100 imported, classified as 92 open + 8 solved, and generated through the pinned binary in a single request.**

### Two upstream points

**1. Guillemets in generator module paths**

The generator’s own `moduleSourcePath` / `ileanPath` handling currently splits module names on every `.`, including dots inside guillemet identifiers. A component such as `«0912.2382»` therefore mis-resolves.

This is the same class of bug fixed in the adapter. For now, the adapter avoids it by generating single-identifier module names.

**2. Provenance in the v1 contract**

Section 10 requires each imported problem to retain its FC source commit and declaration ID. v1 has no provenance field, so these currently travel in an `fc-provenance.json` sidecar.

Would you prefer a passthrough/provenance object in v2, or is the sidecar the intended long-term representation?

One related detail: dotted FC declaration names are restated under their slug because the generator anchors on the final name component. Otherwise `erdos_1038.parts.i`, for example, would become `theorem i`. The provenance record retains the original qualified FC declaration name.

williamjblair:
The whole-set audit now runs green on this PR: **100/100 declarations import and verify at FC pins (92 open + 8 solved), and 97/100 generated Challenges compile at LeanEval target pins, with the three exceptions exactly matching `comparator/known_failures.toml`** — per-declaration reports in the run artifact: https://github.com/google-deepmind/formal-conjectures/actions/runs/32412072356

Of the three recorded exceptions, two are Mathlib drift between the pin sets (`Erdos1092`'s `Fintype` instances, `EllipticCurveRank`'s `⟮ℚ⟯` notation) and retire with FC's 4.33 bump — the audit will fail on the stale entries and force their removal. The third is a small generator-side gap, adding to the two upstream points from earlier:

**3. Challenge reconstruction drops `set_option` context.** `Erdos125.erdos_125.variants.positive_unequal_density` needs `set_option quotPrecheck false` for its set-builder `local notation`; the emitted module carries it attached to the notation command, but the reconstructed `Challenge.lean` re-extracts the notation without it and fails at `quot_precheck`. Since `Challenge.lean` is generator-owned, this side can't amend it. Happy to PR the fix to `lean-eval-generator` with this case as a regression fixture if that's easier than context-switching mid-overhaul — otherwise the repro is one `--set` run away.

Also for reviewers: the adapter now lives whole under `comparator/` (code and tests in `comparator/adapter/`, config and docs beside them). If you would prefer to keep everything under scripts/, lmk

kim-em:
The remaining generator-owned Erdos 125 failure is fixed in https://github.com/leanprover/lean-eval-generator/pull/1.

I verified the patch against this PR's exact head (`54abfe518d4484b74bba5470589841bb37e927c6`): the adapter generated `Erdos125.erdos_125.variants.positive_unequal_density`, the active `set_option quotPrecheck false` preceded both local notations in `Challenge.lean`, `Submission.lean`, and `Solution.lean`, and the complete workspace built successfully at the pinned Lean 4.33 / Mathlib revision (with only the two expected submission-hole `sorry` warnings).

Once the generator PR receives its required independent review and merges, I can supply the final immutable generator SHA for `comparator/tools.toml`.

kim-em:
The quoted-module-path issue is now fixed upstream in leanprover/lean-eval-generator#2. It preserves `«0912.2382»` as one semantic name component and resolves the expected `Arxiv/0912.2382/...` source and `.ilean` paths; the PR includes the exact FC-style regression fixture and byte-golden parity checks.

On provenance: the generator v1 request/response schema is intentionally frozen, so `fc-provenance.json` is the intended v1 boundary rather than adding an arbitrary passthrough field. Please keep that sidecar strict, deterministic/digested, and owned by the FC importer. A future generator v2 should use a typed provenance object, but that does not need to block the current FC100 import.

williamjblair:
Thank you — both upstream fixes verified against our exact head is exactly what we needed. Once leanprover/lean-eval-generator#1 and leanprover/lean-eval-generator#2 merge I'll bump `[generator]` to the merged SHA and drop the `Erdos125` entry from `known_failures.toml` (the audit fails on a stale entry by design, so it will tell me).

On provenance, understood: the sidecar is the v1 boundary. It is now what you describe, in cef02049: unknown keys are refused on load, serialisation is key-sorted, and a `digests` block binds the exact `moduleContent` that crossed the seam to the SHA-256 of every generated file from the response — so a workspace carries its own chain from FC commit → module bytes → generated bytes, and each link can be checked without this repository.

kim-em:
One more cross-check: I built `FormalConjectures.Arxiv.«0912.2382».CurlingNumberConjecture` at this PR's exact head and generated `curling_number_conjecture` with `--verify` against the binary from generator #2. The complete workspace was produced successfully, including the expected FC provenance path `FormalConjectures/Arxiv/0912.2382/CurlingNumberConjecture.lean`.

No FC adapter change is needed for this fix: the adapter already decodes the real FC source path correctly and deliberately gives the synthetic v1 context module a collision-free single-component name. Generator #2 removes the upstream limitation without changing that valid v1 arrangement.

kim-em:
I completed the full FC100 audit against this exact head (`54abfe518d4484b74bba5470589841bb37e927c6`) using a generator built from #1 followed by #2 on `a726789`:

- source/import: 100/100 imported, `--verify` passed, and generated; 92 open + 8 solved
- target: 98/100 Challenges compile at the pinned LeanEval Lean/Mathlib revisions
- Erdos125 now compiles, so its generator-side known-failure entry can be removed after the pin advances
- the only target failures are the two documented pin-drift cases, Erdos1092 and EllipticCurveRank

Canonical report SHA-256 values are `cb85aaa1c8bed7932a449d12bcdfc7ab57a8d6d3261e17121348c0adc6517fac` (source) and `cd2d41e6ad72685f2d88c70f9fc39ecd10014bad3140c4b13876488f51aa0086` (target). This validates the exact post-merge outcome before the immutable generator pin is advanced.

kim-em:
Targeted cross-branch evidence for the two remaining target-pin failures:

- merged this PR head `cef02049664622e48b62a247ebaa8a4afe5c04bb` locally with the clean head of #4428 (`905ec88ddb2ebf39cc4bea413660457fc9657a00`); the merge was conflict-free;
- built the affected FC source modules at #4428's Lean 4.32/Mathlib pins;
- generated `Erdos1092.f_asymptotic_general` and `EllipticCurveRank.RatEllipticCurve.twentyone_le_rank_height_count_asymptotic` using the locally composed lean-eval-generator #1 + #2;
- compiled both generated Challenges at LeanEval's exact target pins: Lean 4.33.0 and Mathlib `6f1ef4e5dd604a435bddba4747b13970cd65d2a1`.

Result: **2/2 target Challenges compile, zero failures**. This confirms the two recorded cross-pin failures are retired by #4428's source adaptations; no LeanEval-side compatibility shim is needed. The importer normally requires the source commit to be reachable from upstream `main`, so the local integration used a test-only source-revision override and did not publish provenance or generated workspaces. A full 100-member rerun should follow after #4428 and the two generator fixes merge.

kim-em:
Lean 4.32 exposed a correctness bug in the importer’s parameter extraction: a theorem proved by bare `sorry` can be stored as a full-type `sorryAx`, so proof-value lambda arity loses declaration-header parameters. I prepared a follow-up commit on top of this PR’s current head:

https://github.com/kim-em/formal-conjectures/commit/6487e90485f13bc41c52258d4ca1ae89838a28bb

Please cherry-pick it:

`git cherry-pick 6487e90485f13bc41c52258d4ca1ae89838a28bb`

The fix recovers the header boundary from source syntax while retaining binder names/explicitness from the elaborated telescope; handles outer variables, conclusion `∀`/`→` binders, scoped notation, and copied noncomputable/source-only dependencies; and adds a CI self-test.

Validation:
- this PR’s Lean 4.27 pins: full FC100 source import 100/100, 0 failures (92 open, 8 solved); extractor build/self-test green; 115 adapter tests green; 11 focused real imports verify
- combined Lean 4.32 bump: FC100 source import 100/100, 0 failures
- LeanEval target pins: 100/100 generated Challenges compile
- latest refactored PR head `66dcc617070a72e234e8e00f81a72fd40ed2eb6f`: clean extractor build, binder-boundary self-test, and all 115 adapter tests green with the composed generator (no skips)
- actionlint, py_compile, and diff check green

I attempted the allowed-maintainer fast-forward directly, but GitHub returned 403 for the fork branch, hence the cherry-pick.

williamjblair:
Cherry-picked as 0bfb9f7a (your authorship intact) — thank you for finding this and for the fix; the explicit `copy_dependencies` escape hatch for opaque-value erasure is the right shape, fail-closed and audible. Checked here before pushing at the 4.27 pins: `comparator_facts --self-test` passes, 115 adapter tests pass, and four real imports verify — `Erdos340.erdos_340.variants.co_density_zero_sub` (the new case), `isSumOfThreeCubes_iff_mod_9`, `Erdos888.erdos_888`, `Erdos92.erdos_92.variants.weak`.

7c24bef7 updates `known_failures.toml` to record the retirement path you verified for each entry: the two drift cases with the toolchain bump (your 2/2 at the target pins on the merge with #4428), and Erdos125 when `[generator]` advances past leanprover/lean-eval-generator#1. Once leanprover/lean-eval-generator#1 and leanprover/lean-eval-generator#2 merge I will bump the pin and delete all three entries in one commit, and the audit will confirm it.

kim-em:
Terminology follow-up from lean-eval: we now reserve unqualified **v1/v2** exclusively for problem sets. Please spell the current wire format **generator schema version 1** (or **schema-version-1 request**) and call the hypothetical successor a **future generator contract revision**, rather than “v1” and “v2.” Frozen filenames/identifiers such as `request-v1.schema.json` remain unchanged. I could not edit this upstream PR description directly.

williamjblair:
Done in 2654e42d and in the description: the contract is now "generator schema version 1" / "schema-version-1 request" throughout, and the successor "a future generator contract revision"; `request-v1.schema.json` and others keep their names.

kim-em:
Coordination update: the standalone generator fixes and synchronization are now merged at `leanprover/lean-eval-generator@77373a539b31f8f304c852f288d7d8469cceebff`. The LeanEval consumer has also merged as `leanprover/lean-eval@b91d4757aa0d7776c02540c9089df54fa0d0658a` in leanprover/lean-eval#553 after its full hosted catalog matrix passed. This PR currently references the older generator pin `a7267895`; its owner can now advance to `77373a53` and rerun the FC workspace/parity checks when ready. No review assignment is being made by this comment.

williamjblair:
Advanced to `77373a53` in 37fd9ab3. The Erdos125 ledger entry is retired: its generated Challenge now carries the notation's `set_option quotPrecheck false`, and the whole-set audit on the new pin is green with the remaining two toolchain-drift entries asserted exactly (100/100 imported, 92 open / 8 solved). The per-PR workspace job is green too. The two drift entries stay until the FC bump.

kim-em:
The FC toolchain bump that was meant to retire the final two target-drift entries has now landed on current `main` (`fd5c0ab5`, Lean 4.33.1 / Mathlib `0df444a`). This PR is currently 44 commits behind/diverged from that base.

I checked the integration shape read-only: merging current `main` into this head is conflict-free. The next owner update should:

- sync `comparator-workspaces` with exact current `main`;
- remove the now-stale `Erdos1092.f_asymptotic_general` and `EllipticCurveRank.RatEllipticCurve.twentyone_le_rank_height_count_asymptotic` blocks from `comparator/known_failures.toml`;
- reconcile the remaining Lean 4.27 prose and `lean4export` source pin with the 4.33.1 source tree;
- rerun the extractor self-test, adapter tests, focused real imports, and the full FC100 source/generator/target audit with an empty expected-failure ledger.

Any new failure should remain fail-closed and be investigated rather than re-ledgered. The authenticated LeanEval maintainer account still has read-only access to the PR author's fork, so I cannot safely push this synchronization myself.

williamjblair:
Synced to `main` (`fd5c0ab5`) and retired the ledger in `29d04d64`: the two target-drift entries are gone, the 4.27 prose now says 4.33.1, the interface-test fixtures carry the real toolchain, and `[tools].lean4export` points at the `v4.33.0` tag with a note, since lean4export has no `v4.33.1` tag upstream yet.

Evidence on the new head:

- extractor self-test, 113 adapter tests, and the focused imports pass locally;
- the local whole-set source/generator sweep imports 100/100 (92 open / 8 solved) with the empty ledger asserted exactly;
- the [FC100 whole-set audit](https://github.com/google-deepmind/formal-conjectures/actions/runs/32696992696) is green, so the target stage compiled every generated Challenge with zero recorded failures;
- the [workspace job at LeanEval pins](https://github.com/google-deepmind/formal-conjectures/actions/runs/32696992652) is green; the Build job is still running.

Nothing was re-ledgered.

williamjblair:
Since the ledger retirement, two passes landed on this branch: a code-review pass (commits `e99ba049..bd282283` — the binder-position answer slots, the plural `*References:*` citations, the batched extractor, and the module/test reorganisation), and a hardening series (`b68dc53e..20e3051d`) responding to an external deep review of the PR. Each of that review's claims was verified against the code before acting; the confirmed ones are landed, one was refuted, and the refutation is recorded in the tree.

What the series changes:

- **One source snapshot.** `pins()` previously checked only the statement's own file while the closure copier read dependency and notation files from the working tree. It now holds every file whose text reaches the workspace to the one pinned revision, and refuses untracked files. The sidecar records each copied dependency as the emitted slice (declaration, module, path, range, digest), replaces the full original text — which shipped the source's proof body — with range plus digest, and adds a `producer` section: importer commit, generator revision with contract version, target pins.
- **The seam bytes are literal.** Generation, `--emit-import` and the sidecar's `digests.request` now share one serialisation; the binary runs against the relative `context` root from the staging directory, and the seam test replays the emitted file byte-for-byte.
- **The response fails closed.** Unknown or missing fields, unsafe paths, a workspace missing or extra against the request, and a response naming `fc-provenance.json` all refuse; a `--set` batch validates everything before the first workspace lands.
- **Policy is explicit.** Group, status, visibility, revision, submitter and tags arrive as an `ImportPolicy` the command constructs; `permitted_axioms` is checked against the generated `config.json` instead of being recorded unverified. The batch protocol is JSON lines both ways, the ledger refuses unknown keys and duplicates, five `tools.toml` keys nothing read are deleted, and the whole-set audit now triggers on everything its outcome depends on and uploads the request, workspaces and sidecars with a digest manifest.

The refuted claim was that `import Mathlib` widens the source environment. It cannot, here: every problem file imports exactly `FormalConjecturesUtil`, which `public import`s all of Mathlib, so each statement already elaborates under the full library and the header narrows only the FC-local layer, which travels as the copied closure. That corpus fact now sits on `MODULE_PREAMBLE` and in OWNERSHIP; the real residual gap — the Mathlib revision — is recorded on both sides of the sidecar and built by the cross-pin job.

Evidence of no behaviour change: the FC100 sweep after the series is byte-identical to the one before it in every workspace file; the only diffs are the new sidecar schema and the newly emitted `request.json`. 100/100 import, 92 open / 8 solved, empty ledger matched, audit green.

Two follow-ups are filed rather than folded in: leanprover/lean-eval-generator#4 (in-process SHA-256, response path validation, `--build-info`; contract v1 unchanged, nothing here waits on it) and #5152 (the `ProblemSpec` precursor `Metadata.lean` anticipates, now that this PR is its consumer — a follow-up, not a gate). From this side the branch is complete; happy to mark it ready for review whenever the launch gates say so.

williamjblair:
I think this LeanEval integration is now in a good place with all upstream changes now merged (and no blockers I can see) so I will set to ready for review. Thanks @kim-em for all the support! @mo271 @Paul-Lez it would be great for you both to review when you have time. Paul, specifically, I know you also have a comparator integration so perhaps we can compare both and converge on the best design/engineering direction. In the process of working on this PR, I did a deeper dive into the surrounding architecture and ecosystem around FC and potential directions/future work (perhaps for the upcoming workshop too + related to recent discussions in Zulip) in #5158

Paul-Lez:
Hey @williamjblair thanks for working on this!
One small question I have to get started (which is perhaps more for @kim-em): is there a world where the FC100 open problems added to Lean eval are allowed to depend on `FormalConjecturesUtil` (or some subset needed to state those)? 
The reason I ask is that it would make the design here a lot simpler: currently, the infrastructure here is making changes to the syntax of the file, and (from what I've understood so far) parsing source files from Python, which can be brittle.

@williamjblair Re the comparator integration I've been working on, this is intended to be somewhat minimal (i.e. generate the bare minimum needed to run comparator on a given declaration, e.g. this would be useful as a check for people who want to submit solutions to problems that are too large to fit in the repo). Since Lean-Eval may have different constraints, so it's not clear that my implementation would be the right fit there.

williamjblair:
I agree on your complexity/brittleness points (the closure copying and the notation copying are the brittle part which exist only because a workspace cannot depend on this repository)

There are two main design decisions to fix this however

**1. Let the workspace depend on this repository.** This removes the copying:

- **Contract.** Generator contract v1 has six top-level fields and allows no others. The generated lakefile carries one `[[require]]`, filled from the request's `mathlib` field. Both need a change before a workspace can require anything else.
- **Mathlib.** One Lake project resolves one Mathlib. We pin `0df444a` at 4.33.1; a workspace pins `6f1ef4e` at 4.33.0. Both repositories would have to track the same one, so LeanEval would bump when we bump.
- **Trust.** Comparator trusts a Challenge's imports. Today that is the few declarations we copy in. Importing `FormalConjecturesUtil` would trust every FC definition for every problem.

These would be calls for @kim-em.

**2. Make this repository export its facts.** Nothing here emits ranges, notation or namespaces, so the adapter rebuilds them from source text. #5158 proposes the `ProblemSpec` manifest for it.

Neither decision changes the rest: the proof still becomes `sorry`, and each `answer(sorry)` still becomes a definition the solver fills. Both of those already come from the elaborator, not from parsing.

Sounds good on the comparator work too!

PR #4965, opened 2026-08-15
fix: replace answer(sorry) with answer(True) in solved statements

theebayuser:
Per CONTRIBUTING ("If the problem has been solved, `answer(sorry)` should be replaced by `answer(True)` or `answer(False)` accordingly"), this replaces `answer(sorry)` with `answer(True)` in 7 statements already tagged `research solved` whose docstrings state the cited paper proved exactly the formalized claim:

- `FormalConjectures/GreensOpenProblems/14.lean`: `green_14_lower_bound_green` [Gr21], `green_14_lower_bound_hunter` [Hu22], `green_14_lower_bound_brown_landman_robertson` [BLR08], `green_14_lower_bound_li_shu` [LiSh10], `green_14_upper_bound_schoen` [Sc20], `green_14_upper_bound_kelley_meka` [KeMe23]. Each cited result was checked against the paper; the neighbouring `green_14_quadratic` already uses `answer(False)`.
- `FormalConjectures/Paper/LatinSquare.lean`: `oddOrderLeq9LatinSquareTransversal` (transversals of odd-order Latin squares verified for $n \le 9$).

Deliberate `answer(sorry)` placeholders in solved statements were left untouched: ZFC-independence placeholders (`ErdosProblems/1119.lean`, `1175.lean`, documented in their docstrings), value-form constant placeholders (`ErdosProblems/416.lean`, `975.lean`), and question-form headlines whose concrete answer is recorded in a variant (`ErdosProblems/1007.lean`, `1026.lean`).

- [x] `lake --wfail build 'FormalConjectures.GreensOpenProblems.«14»' FormalConjectures.Paper.LatinSquare` passes with no warnings
- [x] Docstrings unchanged; only the `answer( )` terms edited

### DIFF
--- a/FormalConjectures/GreensOpenProblems/14.lean
+++ b/FormalConjectures/GreensOpenProblems/14.lean
@@ -87,42 +87,42 @@ theorem green_14_quadratic :
 /-- [Gr21] proved a lower bound of shape $W(3, r) \gg \exp(c(\log r)^{4/3-o(1)})$. -/
 @[category research solved, AMS 5 11]
 theorem green_14_lower_bound_green :
-    answer(sorry) ↔ ∃ c : ℝ, ∃ (o : ℕ → ℝ) (_ : Tendsto o atTop (𝓝 0)),
+    answer(True) ↔ ∃ c : ℝ, ∃ (o : ℕ → ℝ) (_ : Tendsto o atTop (𝓝 0)),
     (fun (r : ℕ) => Real.exp (c * (Real.log r)^(4/3 - o r))) =O[atTop] fun r => (W 3 r : ℝ) := by
   sorry
 
 /-- [Hu22] improved this to $W(3, r) \gg \exp(c(\log r)^{2-o(1)})$. -/
 @[category research solved, AMS 5 11]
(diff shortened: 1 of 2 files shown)


mo271:
nice! Can we have a linter for that (I though at some point we had already written it, but apparently I'm mistaken...)?

mo271 (approved):
Thanks!
Happy to merge as is and then the linter could be done in a follow up?

theebayuser:
> nice! Can we have a linter for that (I though at some point we had already written it, but apparently I'm mistaken...)?

sure, I think you're referring to #[1958](github.com/google-deepmind/formal-conjectures/pull/1958), which catches the quantifier scoping but doesn't read the category. I'll open a follow up pr

mo271 (approved):

mo271:
> > nice! Can we have a linter for that (I though at some point we had already written it, but apparently I'm mistaken...)?
> 
> sure, I think you're referring to #[1958](github.com/google-deepmind/formal-conjectures/pull/1958), which catches the quantifier scoping but doesn't read the category. I'll open a follow up pr

Thanks, looking forward to the follow up pr!

PR #4966, opened 2026-08-15
update documentation on placeholder/stubs

mo271:


mo271:
@williamjblair we had that in AGENTS.md, but it was removed in the update, it seems like it is important to have!

williamjblair:
Agreed, good to have this back

If it's useful, we could also make CI catch two of these automatically: a check that fails if any file in FormalConjectures/ contains an axiom or opaque declaration. 

Edit: Actually saw #4968 nice!

mo271:
Already prepared the linter for opaque in #4968, good to have the same for axiom as wellAm 15.08.2026 12:30 schrieb Will Blair ***@***.***>:williamjblair left a comment (google-deepmind/formal-conjectures#4966)
Agreed, good to have this back
If it's useful, we could also make CI catch two of these automatically: a check that fails if any file in FormalConjectures/ contains an axiom or opaque declaration.

Paul-Lez (approved):

PR #4968, opened 2026-08-15
feat(FormalConjecturesUtil/Linters): StubLinter

mo271:


FormalConjecturesUtil/Linters/OpaqueLinter.lean
```diff
@@ -0,0 +1,47 @@
...
+
+register_option linter.style.opaque_decl : Bool := {
+  defValue := false
+  descr := "enable the linter to forbid opaque declarations"
+}
+
+namespace OpaqueLinter
+
+/-- The opaque linter checks that no `opaque` definitions are present. -/
+def opaqueLinter : Linter where
```
Paul-Lez: I guess this can very easily be made into a linter that also flags axioms at the same time - maybe that's worth doing?

Paul-Lez (approved):

FormalConjecturesUtil/Linters/OpaqueLinter.lean
mo271: @Paul-Lez good point, added that and also checking for sorried-out definitions. Renamed the linter "StubLinter" instead of "OpaqueLinter"

PR #4450, opened 2026-07-17
Add the first 64 files from AutoOeis to formal-conjectures

codrut3:
Processes several sequences from AutoOeis (auto-formalisations of conjectures from the [OEIS](https://oeis.org/)) to match the formal-conjectures style.

I discarded sequences that had no conjectures, or the conjecture was impossible to formalise (e.g. https://oeis.org/A109224).
In some cases, I added known facts (with proofs) as lemmas.

For a few sequences it was very hard to prove the first few terms. In this case, I left them out, rather than using `sorry`.

Files that referred to the same sequence were combined into a single entry, with multiple conjectures, numbered as conjecture1, conjecture2, and so on.

PR #4985, opened 2026-08-15
OEIS/108864: encode the perfect deficiency A109883, not |sigma(n) - 2n|

### LINKED ISSUE #4984: OEIS A108864: `A` encodes |σ(n) − 2n| ≤ 10, not the perfect deficiency A109883; the odd-terms question is untouched
## Summary

`FormalConjectures/OEIS/108864.lean` defines the sequence membership predicate as

```lean
def A (n : ℕ) : Prop :=
  let sigmaOneN : ℕ := (Nat.divisors n).sum id
  0 < n ∧ ((sigmaOneN : ℤ) - 2 * (n : ℤ)).natAbs ≤ 10
```

while its own docstring says *"its perfect deficiency is $\le 10$"*.

[A108864](https://oeis.org/A108864) is **"Numbers n such that the perfect deficiency of n
(A109883) is <= 10"**, and the perfect deficiency [A109883](https://oeis.org/A109883) is a
*greedy divisor-subtraction* quantity — *"start subtracting from n its divisors beginning from
1 until one reaches a number smaller than the last divisor subtracted or reaches the last
nontrivial divisor < n"*. It is not the deviation from perfection `|σ₁(n) − 2n|`. The docstring
and the Lean disagree, so by AGENTS.md (*"If they disagree, the Lean is incorrect"*) the Lean
is the side that drifted.

**One published value decides this, with no conjecture involved.** `24` is the twentieth term
of A108864's DATA. Its perfect deficiency is `0`:

```
24 − 1 − 2 − 3 − 4 − 6 − 8 = 0
```

But `σ₁(24) = 60` and `|60 − 48| = 12 > 10`, so the current Lean predicate `A 24` is **false**.
A number the source lists is excluded by the formalization, so the two sets are provably
different. The divergence runs both ways: `56` is *not* in A108864 (its perfect deficiency is
`20`), yet `σ₁(56) = 112 = 2 · 56`, so `A 56` holds.

Permalink:
https://github.com/google-deepmind/formal-conjectures/blob/638da20efd8eeeed2993fc2550fc596dc90c1ce8/FormalConjectures/OEIS/108864.lean

## Counterexample

The mis-encoding is not cosmetic — it decides the open statement. `conjecture` is tagged
`@[category research open]` and reads

```lean
theorem conjecture : answer(sorry) ↔ ∀ n > 58, Even (a n)
```

with `a n = Nat.nth A n`. Let `L = {n > 0 : |σ₁(n) − 2n| ≤ 10}` be the set the current `A`
actually defines. `L` is infinite (`|σ₁(2^k) − 2^(k+1)| = 1` for every `k`), so `Nat.nth A` is
total and the statement is not vacuous. Enumerating `L` (0-indexed):

```
1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 14, 15, 16, 18, 20, 21, 22, 26, 28, 32, 40, 44, 50,
52, 56, 64, 68, 70, 88, 104, 110, 128, 130, 136, 152, 184, 196, 256, 315, 368, 464, 496,
512, 592, 650, 656, 836, 884, 1012, 1024, 1155, 1696, 1888, 1952, 2048, 2144, 2272, 2336,
4030, 4096, 5830, 8128, 8192, 8384, 8768, 8925, 11096, 16384, …
```

**`n = 67`**: `a 67 = 8925`, which is odd, and `67 > 58`. So `Even (a 67)` is false and the
right-hand side fails.

The premise holds: `8925 = 3 · 5² · 7 · 17`, so
`σ₁(8925) = 4 · 31 · 8 · 18 = 17856`, `2 · 8925 = 17850`, and `|17856 − 17850| = 6 ≤ 10`.

`67` is the *least* violating index: indices `59…66` are `2336, 4030, 4096, 5830, 8128, 8192,
8384, 8768`, all even. Further odd terms follow at `a 74 = 32445` and `a 94 = 442365`. The odd
elements of `L` with their indices are

```
(0,1) (2,3) (4,5) (6,7) (8,9) (10,11) (13,15) (17,21) (40,315) (52,1155) (67,8925)
(74,32445) (94,442365)
```

so `answer` is currently forced to `False` — a `research open` slot that is already decided.

**The `> 58` bound is the giveaway.** In the *real* A108864, `1155` sits at 0-indexed position
**58**, which is exactly the bound the declaration uses. In the set the current predicate
defines, `1155` sits at index **52**. The index arithmetic was calibrated against the published
sequence while the predicate defines a different one.

**The OEIS question itself is untouched and remains open.** `8925` is not a term of A108864 at
all — its perfect deficiency is `2969`. In the real sequence the odd terms up to `3 · 10⁵` are
exactly `1, 3, 5, 7, 9, 11, 15, 21, 315, 1155` (103 terms in that range), so nothing
contradicts *"Is 1155 the last odd number in this sequence?"*. What `n = 67` refutes is the
formalization, not the conjecture.

## Suggested fix

Encode the perfect deficiency and leave `conjecture` `research open`:

```lean
/-- The perfect deficiency of `n`, A109883. -/
def perfectDeficiency (n : ℕ) : ℕ :=
  ((List.range (n + 1)).filter (· ∣ n)).foldl (fun r d => if r < d then r else r - d) n

def A (n : ℕ) : Prop :=
  0 < n ∧ perfectDeficiency n ≤ 10
```

`(List.range (n + 1)).filter (· ∣ n)` is the list of divisors of `n` in increasing order, and
the single guard `if r < d` captures both stopping rules of the A109883 description: once the
remainder is smaller than one divisor it is smaller than every later one, so the fold is
constant from that point on; and for `n > 1` the divisor `n` itself is never subtracted,
because the remainder is already `< n` once `1` has been subtracted. (Filtering a range rather
than sorting `Nat.divisors` keeps the definition reducible in the kernel, which the `test`
theorems need; `Finset.sort` does not reduce under `decide`.)

With this definition the file's docstring, its A108864 reference and its index arithmetic all
become correct simultaneously: `a 58 = 1155`, and `∀ n > 58, Even (a n)` is once again the
question A108864 asks.

**The trade-off, stated explicitly.** The other defensible repair is to keep
`|σ₁(n) − 2n| ≤ 10` and close the declaration with `answer(False)` plus the `n = 67`
certificate. That is sound as Lean, but it would record inside the benchmark a refutation of a
conjecture that has *not* been refuted, and it would leave the module docstring, the A108864
reference and the `58` in the statement all describing a sequence the file no longer defines. I
have gone with the encoding fix on the strength of AGENTS.md's *"Each statement must say what
its source says"*, and because #4933 (Erdős 940) took the same shape — repair the drifted
statement in place, keep it `research open`, and put the contradiction in the PR description
rather than in the Lean. Happy to switch to the `answer(False)` form if maintainers prefer it;
the `n = 67` witness above is exactly what that version would certify.

I have the fix prepared and will open a PR referencing this issue.

This came out of the campaign's statement-audit lane, which compares `formal-conjectures`
declarations against their cited sources and then searches for witnesses where the two diverge.
It is a formalization counterexample of the same kind as the source/statement mismatches
collected in #4896 and #4923, not a mathematical advance on A108864.

The source audit, the witness derivation and the duplicate search are recorded immutably at:

https://github.com/Kuberwastaken/c5-k4/blob/125be1d2ad3da8ade60615d1f6d7d3b1722a098a/results/expansion/live-search-2026-08-15/oeis-hunt-part1.md#L314-L397

The full submission record, including the build and axiom audit for the fix, is at:

https://github.com/Kuberwastaken/c5-k4/blob/125be1d2ad3da8ade60615d1f6d7d3b1722a098a/results/expansion/publication/oeis_108864_submission.md

## Independent verifier

The root cause was pinned by re-implementing A109883 literally from its NAME and checking it
against published data, in three independently written code paths (two Python, one Lean):

https://github.com/Kuberwastaken/c5-k4/blob/125be1d2ad3da8ade60615d1f6d7d3b1722a098a/results/expansion/live-search-2026-08-15/scripts/part1/c109883b.py#L1-L27

- A109883's first **79** published terms are reproduced exactly.
- Filtering that quantity by `≤ 10` reproduces A108864's **61** published DATA terms exactly,
  placing `1155` at 0-indexed position `58`.
- `perfectDeficiency(24) = 0`, `perfectDeficiency(56) = 20`, `perfectDeficiency(8925) = 2969`.

The proposed Lean definition was checked against the same two published lists by `#eval` and
agrees on both. The enumeration of `L` and the index of `8925` were recomputed independently by
a numpy divisor-sum sieve to `10⁷` and by plain trial division with no sieve:

https://github.com/Kuberwastaken/c5-k4/blob/125be1d2ad3da8ade60615d1f6d7d3b1722a098a/results/expansion/live-search-2026-08-15/scripts/part1/c108864.py

https://github.com/Kuberwastaken/c5-k4/blob/125be1d2ad3da8ade60615d1f6d7d3b1722a098a/results/expansion/live-search-2026-08-15/scripts/part1/c108864b.py

`lake --wfail build 'FormalConjectures.OEIS.«108864»'` passes warning-clean on the prepared
fix (Lean `v4.27.0`, repo-pinned Mathlib). `#print axioms` gives `[propext]` for the A109883
value test and `[propext, Classical.choice, Quot.sound]` for the term theorems and the API
lemma — no `native_decide`, no `Lean.ofReduceBool`. `conjecture` keeps its `sorryAx`, as it
should: this submission does not close it.

## Source/status note

- A108864 (OFFSET `1,2`), NAME *"Numbers n such that the perfect deficiency of n (A109883) is
  <= 10"*. Its only comment is *"Is 1155 the last odd number in this sequence?"* — the
  statement the file formalizes. DATA:
  `1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 14, 15, 16, 18, 20, 21, 22, 24, 26, 28, 30, 32, 40,
  42, 44, 50, 52, 60, 64, 68, 72, 110, 120, 126, 128, 130, 136, 144, 150, 152, 180, 184, 204,
  228, 256, 315, 462, 496, 512, 528, 592, 656, 750, 884, 1012, 1024, 1155, 1188, 1248`.
- A109883 gives the perfect deficiency. Its first terms are
  `0, 1, 2, 1, 4, 0, 6, 1, 5, 2, 10, 2, 12, 4, 6, 1, 16, 6, 18, 8, …`; note `A109883(24) = 0`,
  which is what puts `24` in A108864.
- The current file entered the repository in #4450 (`Add the first 64 files from AutoOeis`),
  which is the only commit that has ever touched it. The mismatch looks like an automated
  reading of "perfect deficiency" as the abundance/deficiency `|σ₁(n) − 2n|`, which is a
  different and more common notion.
- Status of the mathematics: **open**, both before and after this report.

## AI assistance disclosure

Claude Code and delegated coding agents assisted with the source audit, the search for the
witness, the Lean repair, the independent verification and the preparation of this report. The
submitter reviewed the mathematical statement, the counterexample and the Lean artifact, and
takes responsibility for the submission.

## Choose either option

Kuberwastaken:
## Summary

`FormalConjectures/OEIS/108864.lean` defined its membership predicate as
`0 < n ∧ |σ₁(n) − 2n| ≤ 10`, while the file's own docstring — and
[A108864](https://oeis.org/A108864)'s NAME — say *"the perfect deficiency of `n` is `≤ 10`"*.
The perfect deficiency is [A109883](https://oeis.org/A109883), a greedy divisor-subtraction
quantity, not the deviation from perfection.

The two sets are provably different from published data alone. `24` is the twentieth term of
A108864's DATA and its perfect deficiency is `0` (`24 − 1 − 2 − 3 − 4 − 6 − 8 = 0`), but
`σ₁(24) = 60` and `|60 − 48| = 12 > 10`, so the old predicate rejected a term the source lists.
Conversely `56 ∉ A108864` (perfect deficiency `20`) yet `σ₁(56) = 2 · 56`, so the old predicate
accepted it.

The drift also decided the open statement. Under `|σ₁(n) − 2n| ≤ 10` the sequence has
`a 67 = 8925`, which is odd with `67 > 58`, so `∀ n > 58, Even (a n)` was false and the
`answer` hole on a `research open` declaration was already forced to `False`. Further odd terms
appear at `a 74 = 32445` and `a 94 = 442365`.

This PR replaces the predicate with the perfect deficiency itself and leaves the conjecture
open. It does not resolve A108864's question, which was never in danger: `8925` has perfect
deficiency `2969` and is not a term of A108864 at all, and the real sequence has no odd term
after `1155` up to `3 · 10⁵`.

Fixes #4984

## Follow-up discovery pattern

The bound in the statement is what exposed the drift. `∀ n > 58, Even (a n)` encodes *"1155 is
the last odd term"*, and `1155` sits at 0-indexed position **58** in the published A108864 — but
at position **52** in the set the old predicate defined. The index arithmetic had been
calibrated against the real sequence while the predicate defined a different one, so comparing
a declaration's numeric constants against the sequence its predicate actually generates is a
cheap and general detector for this class of mis-encoding.

## Formal content

No proof is added and none is claimed. `conjecture` keeps `answer(sorry)` and
`@[category research open]`; its statement text is unchanged. What changed is the set it is
about.

- `perfectDeficiency` is new:
  ```lean
  def perfectDeficiency (n : ℕ) : ℕ :=
    ((List.range (n + 1)).filter (· ∣ n)).foldl (fun r d => if r < d then r else r - d) n
  ```
  `(List.range (n + 1)).filter (· ∣ n)` lists the divisors of `n` in increasing order. The
  single guard `if r < d` captures both stopping rules of A109883's description: once the
  remainder is below one divisor it is below every later one, so the fold is constant
  thereafter; and for `n > 1` the divisor `n` is never subtracted, because the remainder is
  already `< n` after `1` has been. For `n = 1` the only divisor `1` is subtracted, giving `0`,
  the value A109883 records. `perfectDeficiency 0 = 0`, and `A`'s `0 < n` guard keeps that junk
  value out of the sequence.

  Divisors are listed by filtering a range rather than by sorting `Nat.divisors` because the
  fold needs them in increasing order and the filtered range reduces in the kernel;
  `Finset.sort` does not reduce under `decide`, which made the `test` theorems fail.
- `mem_divisorList_iff` (new, `category API`) proves the filtered range is exactly
  `Nat.divisors n` for `0 < n`, so the definition is not silently folding over the wrong list.
- `perfectDeficiency_values` (new, `category test`) pins A109883's first ten published terms.
- `a_19 : a 19 = 24` (new, `category test`) pins the term that separates the two readings — it
  is exactly the value the old predicate got wrong.
- `a_0`…`a_4` are unchanged and still pass; `1, 2, 3, 4, 5` lie in both sets.
- The module docstring now names A109883, states the distinction, and cites both sequences.

## Source/status note

- A108864 (OFFSET `1,2`): *"Numbers n such that the perfect deficiency of n (A109883) is
  <= 10"*. Sole comment: *"Is 1155 the last odd number in this sequence?"* — the statement
  `conjecture` formalizes. Status: **open**, before and after this PR.
- A109883: *"Start subtracting from n its divisors beginning from 1 until one reaches a number
  smaller than the last divisor subtracted or reaches the last nontrivial divisor < n. Define
  this to be the perfect deficiency of n."*
- The file entered the repository in #4450 (`Add the first 64 files from AutoOeis`), the only
  commit that has ever touched it. The mismatch looks like an automated reading of "perfect
  deficiency" as the abundance/deficiency `|σ₁(n) − 2n|`.
- Alternative repair considered and rejected: keep `|σ₁(n) − 2n| ≤ 10` and close the
  declaration with `answer(False)` and the `n = 67` certificate. That is sound as Lean, but it
  would record a refutation of a conjecture that has not been refuted and would leave the
  docstring, the A108864 reference and the `58` describing a sequence the file no longer
  defines. AGENTS.md (*"Each statement must say what its source says"*) and the shape of #4933
  point the other way. Issue #4984 states the trade-off; if you prefer the `answer(False)`
  form, the `n = 67` witness there is exactly what it would certify and the switch is
  mechanical.

## Verification

`lake --wfail build 'FormalConjectures.OEIS.«108864»'` — passes, warning-clean, exit 0
(Lean `v4.27.0`, repo-pinned Mathlib).

`#print axioms`:

```
'OeisA108864.perfectDeficiency_values' depends on axioms: [propext]
'OeisA108864.mem_divisorList_iff'      depends on axioms: [propext, Classical.choice, Quot.sound]
'OeisA108864.a_0'                      depends on axioms: [propext, Classical.choice, Quot.sound]
'OeisA108864.a_19'                     depends on axioms: [propext, Classical.choice, Quot.sound]
'OeisA108864.conjecture'               depends on axioms: [propext, sorryAx, Classical.choice, Quot.sound]
```

No `decide +native`, no `native_decide`, no `Lean.ofReduceBool` / `Lean.trustCompiler`.
`conjecture` carries `sorryAx` by design.

Evaluated from the new definition and matching the published data:

| Check | Result |
|---|---|
| A109883's first **79** published terms | reproduced exactly |
| A108864's **61** published DATA terms, from `perfectDeficiency n ≤ 10` | reproduced exactly |
| 0-indexed position of `1155` | **58**, matching the `> 58` in `conjecture` |
| `perfectDeficiency` at `8925` / `24` / `56` | `2969` / `0` / `20` |

Reproduced independently in Python from A109883's NAME (two separately written stopping-rule
formulations), and the counterexample enumeration was recomputed both by a numpy divisor-sum
sieve to `10⁷` and by plain trial division with no sieve. Record:

https://github.com/Kuberwastaken/c5-k4/blob/125be1d2ad3da8ade60615d1f6d7d3b1722a098a/results/expansion/publication/oeis_108864_submission.md

**Limitations.** Only the changed module was built; the full project was not. `a 58 = 1155` is
not asserted as a Lean `test` theorem — `decide` on `Nat.count A 1155 = 58` would need the
kernel to reduce `perfectDeficiency` on every `n < 1155`. It is verified by `#eval` and by the
independent recomputations above, not by the kernel.

## AI assistance disclosure

Claude Code and delegated coding agents assisted with the source audit, the search for the
witness, the Lean repair, the independent verification and the preparation of this pull
request. The submitter reviewed the mathematical statement, the counterexample and the Lean
artifact, and takes responsibility for the submission.

### DIFF
--- a/FormalConjectures/OEIS/108864.lean
+++ b/FormalConjectures/OEIS/108864.lean
@@ -19,25 +19,58 @@ import FormalConjecturesUtil
 /-!
 # Numbers $n$ such that the perfect deficiency of $n$ is $\le 10$.
 
-We formally define the property satisfied by elements of the sequence,
-using the sum of divisors function $\sigma_1(n)$.
+A108864 lists the numbers $n$ whose *perfect deficiency* is at most $10$. The perfect
+deficiency is [A109883](https://oeis.org/A109883), a greedy divisor-subtraction quantity. It is
+not the deviation $|\sigma_1(n) - 2n|$ from perfection: for example $24$ is a term of A108864
+and has perfect deficiency $0$, while $|\sigma_1(24) - 2 \cdot 24| = 12$.
 
 *References:*
 - [A108864](https://oeis.org/A108864)
+- [A109883](https://oeis.org/A109883)
 -/
 
 namespace OeisA108864
 
-open Nat Finset Int
+open Nat Finset
+
+/--
+The *perfect deficiency* of $n$, [A109883](https://oeis.org/A109883): "start subtracting from
+$n$ its divisors beginning from $1$ until one reaches a number smaller than the last divisor
+subtracted or reaches the last nontrivial divisor $< n$".
+
+`(List.range (n + 1)).filter (· ∣ n)` is the list of divisors of $n$ in increasing order (see
+`mem_divisorList_iff`). The fold subtracts each of them from the running remainder unless the
+remainder has already fallen below it. That single guard captures both stopping rules of the
+quoted description. Once the remainder is smaller than one divisor it is smaller than every
+later divisor, so the fold is constant from that point on; and for $n > 1$ the divisor $n$
+itself is never subtracted, because the remainder is already $< n$ once $1$ has been
+subtracted. For $n = 1$ the only divisor $1$ is subtracted, giving `perfectDeficiency 1 = 0`,
+which is the value A109883 records.
+
+`perfectDeficiency 0 = 0`. The sequence predicate below excludes $0$ explicitly, so this junk
+value is never used.
+
+The divisors are listed by filtering a range rather than by sorting `Nat.divisors`, because
+the fold has to see them in increasing order and the filtered range reduces in the kernel,
+which the `test` theorems below rely on.
+-/
+def perfectDeficiency (n : ℕ) : ℕ :=
+  ((List.range (n + 1)).filter (· ∣ n)).foldl (fun r d => if r < d then r else r - d) n
+
+/-- The list folded over by `perfectDeficiency` is exactly the set of divisors of `n`. -/
+@[category API, AMS 11]
+theorem mem_divisorList_iff {n d : ℕ} (hn : 0 < n) :
+    d ∈ (List.range (n + 1)).filter (· ∣ n) ↔ d ∈ n.divisors := by
+  simp only [List.mem_filter, List.mem_range, Nat.lt_succ_iff, decide_eq_true_eq,
+    Nat.mem_divisors]
+  exact ⟨fun h => ⟨h.2, hn.ne'⟩, fun h => ⟨Nat.le_of_dvd hn h.1, h.1⟩⟩
 
 /--
 The condition for a number $n$ to be in the sequence.
-It satisfies $0 < n$ and its perfect deficiency is $\le 10$, using the sum of divisors
-function $\sigma_1(n)$.
+It satisfies $0 < n$ and its perfect deficiency is $\le 10$.
 -/
 def A (n : ℕ) : Prop :=
-  let sigmaOneN : ℕ := (Nat.divisors n).sum id
-  0 < n ∧ ((sigmaOneN : ℤ) - 2 * (n : ℤ)).natAbs ≤ 10
+  0 < n ∧ perfectDeficiency n ≤ 10
 
 instance : DecidablePred A := by
   unfold A
@@ -49,6 +82,12 @@ The primary defining sequence `a`.
 -/
 noncomputable def a (n : ℕ) : ℕ :=
   n.nth A
+
+/-- The first ten terms of [A109883](https://oeis.org/A109883), the perfect deficiency. -/
+@[category test, AMS 11]
+theorem perfectDeficiency_values :
+    (List.range' 1 10).map perfectDeficiency = [0, 1, 2, 1, 4, 0, 6, 1, 5, 2] := by
+  decide
 
 /-- Term theorems verifying the first few values of the sequence against the official OEIS b-file -/
 @[category test, AMS 11]
@@ -84,6 +123,20 @@ theorem a_4 : a 4 = 5 := by
   have h5 : A 5 := by decide
   have hcnt : Nat.count A 5 = 4 := by decide
   have := Nat.nth_count (p := A) h5
+  rwa [hcnt] at this
+
+/--
+$24$ is the twentieth term of A108864. Subtracting $1, 2, 3, 4, 6, 8$ from $24$ leaves $0$, so
+the perfect deficiency of $24$ is $0$.
+
+This term separates the perfect deficiency from the deviation from perfection:
+$\sigma_1(24) = 60$, so $|\sigma_1(24) - 2 \cdot 24| = 12 > 10$.
+-/
+@[category test, AMS 11]
+theorem a_19 : a 19 = 24 := by
+  have h24 : A 24 := by decide
+  have hcnt : Nat.count A 24 = 19 := by decide
+  have := Nat.nth_count (p := A) h24
   rwa [hcnt] at this
 
 /--


codrut3:
Hi @Kuberwastaken , thank you a lot for your contribution and for finding this issue!

Instead of defining A109883 in the same file, can you create a new entry for it that follows [the style guide](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/OEIS/README.md)? So it defines the sequence `a`, has term theorems (instead of block validation like `perfectDeficiency_values`) etc. Then import this sequence in A108864. Additionally, can you add more term theorems to 108864.lean? For the first 5 terms $a_0, a_1, \ldots, a_4$, instead of $a_{19}$.

PR #1489, opened 2026-01-05
chore: prefer `answer(sorry) ↔ ...`

mo271:
This changes occurrences of  `... ↔ .nswer(sorry)` with `answer(sorry) ↔ ...`, which is hopefully less error prone. (In fact we fix some bugs this way here)

Drive-by: adding some line break here and there in order to satisfy the 100 character limit.

PR #1328, opened 2025-12-03
feat(ErdosProblems): 1093

zer-art:
This PR formalizes Erdős Problem 1093 regarding the deficiency of binomial coefficients.

Closes #1093

PR #4986, opened 2026-08-15
fix(ErdosProblems/1093): `k`-smooth means prime factors `≤ k`, not `< k`

### LINKED ISSUE #4975: Erdős 1093: `deficiency` uses Mathlib `smoothNumbers k` (primes < k) where `k`-smooth means primes ≤ k
## Summary

`Erdos1093.deficiency` counts `i` with `n - i ∈ smoothNumbers k`. Mathlib's
`Nat.smoothNumbers k` means "all prime factors **strictly less than** `k`", while
"`k`-smooth" in the problem statement — and in standard usage — means "all prime
factors **at most** `k`". The two predicates differ exactly when `k` is prime, so
`deficiency` computes a different function from the intended one, and both open
declarations `erdos_1093.parts.i` and `.parts.ii` ask their question about that
different function.

https://github.com/google-deepmind/formal-conjectures/blob/638da20efd8eeeed2993fc2550fc596dc90c1ce8/FormalConjectures/ErdosProblems/1093.lean#L29-L33

```lean
/--
If defined, the deficiency is the count of $0 \le i < k$ such that $n - i$ is $k$-smooth.
-/
noncomputable def deficiency (n k : ℕ) : ℕ :=
  #{i ∈ range k | n - i ∈ smoothNumbers k}
```

Mathlib (`Mathlib/NumberTheory/SmoothNumbers.lean`):

```lean
/-- `smoothNumbers n` is the set of *`n`-smooth positive natural numbers*, i.e., the
positive natural numbers all of whose prime factors are less than `n`. -/
def smoothNumbers (n : ℕ) : Set ℕ := {m | m ≠ 0 ∧ ∀ p ∈ primeFactorsList m, p < n}
```

So the docstring says "`n - i` is `k`-smooth" (prime factors `≤ k`) but the code
says prime factors `< k`.

## Proof

The two readings differ on `m` exactly when `k` is prime and `k ∣ m`; for composite
`k` no `m` separates them, which is why the discrepancy is easy to miss.

Three eligible pairs where they differ (all values recomputed independently with an
exact factoriser):

| `C(n,k)` | `k` | `k` prime | prime factors `≤ k` | current Lean, `< k` | separating `n - i` |
|---|---|---|---|---|---|
| `C(7,3)` | 3 | yes | **1** | **0** | `6 = 2·3` |
| `C(23,5)` | 5 | yes | **1** | **0** | `20 = 2²·5` |
| `C(47,11)` | 11 | yes | **4** | **3** | `44 = 2²·11` |

All three satisfy the side conditions the two declarations impose, so they are
genuine members of the sets being quantified over: `2k ≤ n` holds, and no prime
`p ≤ k` divides the binomial coefficient (`C(7,3) = 35`, `C(23,5) = 33649`,
`C(47,11) = 17417133617`). The side condition itself is encoded faithfully as
`∀ p, p.Prime → p ∣ choose n k → k < p`; only the threshold inside `deficiency`
diverges.

Consequence: two of the deficiency-1 examples on the problem page leave the
deficiency-1 set of `parts.i`, and `C(47,11)` — the only deficiency-4 example
recorded there — becomes a deficiency-3 example. Under the `≤ k` reading every
example in the page's catalogue reproduces the deficiency stated there (23/23);
under the current `< k` reading exactly the three above change. That computation
is recorded in the linked audit.

**The underlying mathematics is unaffected.** Neither `parts.i` nor `parts.ii`
becomes false — they become open questions about a different function. The
Erdős–Lacampagne–Selfridge question itself is untouched.

## Suggested fix

One token, since `smoothNumbers (k+1) = {m ≠ 0 | ∀ p ∈ primeFactorsList m, p < k+1}`
is exactly "all prime factors `≤ k`":

```lean
noncomputable def deficiency (n k : ℕ) : ℕ :=
  #{i ∈ range k | n - i ∈ smoothNumbers (k + 1)}
```

Equivalently, restore the local predicate that the introducing PR originally used.

Found while auditing `research open` declarations for faithfulness to their cited
sources. Full derivation, the 23-entry recomputation under both thresholds, and the
duplicate search:

https://github.com/Kuberwastaken/c5-k4/blob/214325f5962530512b32f31e9e3f63648496f886/results/expansion/live-search-2026-08-15/candidate-verification.md

## Source/status note

erdosproblems.com/1093 is `open`. It defines the quantity as: "the deficiency is
the number of `0 ≤ i < k` such that `n−i` is `k`-smooth, that is, divisible only by
primes `≤ k`" (read 2026-08-15; recorded in the linked audit).

Provenance, for context rather than blame: the first commit of the introducing PR
#1328 defined the predicate by hand and correctly —

```lean
/-- A number $n$ is $k$-smooth if all its prime factors are $\le k$. -/
def IsKSmooth (k n : ℕ) : Prop := ∀ p, p.Prime → p ∣ n → p ≤ k
```

— and it was replaced by `smoothNumbers k` in response to a review suggestion to
reuse Mathlib. The `≤`/`<` difference between the two does not appear to have been
raised in the thread, and the two later fixes to this file (#1489, #1532) did not
touch it.

## AI assistance disclosure

audit, the independent recomputation, and drafting. The submitter reviewed the
declaration, the argument, and the suggested repair, and takes responsibility for
the submission.

### Choose either option

Kuberwastaken:
Fixes #4975.

```lean
noncomputable def deficiency (n k : ℕ) : ℕ :=
  #{i ∈ range k | n - i ∈ smoothNumbers k}
```

erdosproblems.com/1093 (read 2026-08-15):

> For $n \geq 2k$ we define the deficiency of $\binom{n}{k}$ as follows. If
> $\binom{n}{k}$ is divisible by a prime $p \leq k$ then the deficiency is
> undefined. Otherwise, **the deficiency is the number of $0 \leq i < k$ such that
> $n-i$ is $k$-smooth, that is, divisible only by primes $\leq k$.**

Mathlib (`Mathlib/NumberTheory/SmoothNumbers.lean`):

```lean
/-- `smoothNumbers n` is the set of *`n`-smooth positive natural numbers*, i.e., the
positive natural numbers all of whose prime factors are less than `n`. -/
def smoothNumbers (n : ℕ) : Set ℕ := {m | m ≠ 0 ∧ ∀ p ∈ primeFactorsList m, p < n}
```

So the docstring said "$k$-smooth" (primes $\le k$) and the code said primes $< k$.
The two readings differ on `m` exactly when `k` is prime and `k \mid m`, which is why
the discrepancy is easy to miss: for composite `k` no `m` separates them.

The problem page lists 23 worked examples. Recomputing all 23 with exact integer
arithmetic (own trial-division factoriser, no CAS):

| | primes `≤ k` (source) | primes `< k` (current Lean) |
|---|---|---|
| catalogue entries reproduced | **23 / 23** | **20 / 23** |

The three that change are exactly the ones with `k` prime:

| $\binom{n}{k}$ | `k` | `k` prime | source | current Lean | separating `n - i` |
|---|---|---|---|---|---|
| $\binom{7}{3}$ | 3 | yes | **1** | **0** | $6 = 2\cdot 3$ |
| $\binom{23}{5}$ | 5 | yes | **1** | **0** | $20 = 2^2\cdot 5$ |
| $\binom{47}{11}$ | 11 | yes | **4** | **3** | $44 = 2^2\cdot 11$ |

All three satisfy the side conditions the two declarations impose, so they are
genuine members of the sets being quantified over: `2 * k ≤ n` holds, and no prime
`p ≤ k` divides the binomial coefficient ($\binom{7}{3} = 35$,
$\binom{23}{5} = 33649$, $\binom{47}{11} = 17417133617$). That side condition is
encoded faithfully as `∀ p, p.Prime → p ∣ choose n k → k < p`; only the threshold
inside `deficiency` diverged.

Consequently two of the page's deficiency-1 examples left the deficiency-1 set of
`erdos_1093.parts.i`, and $\binom{47}{11}$ — the page's only deficiency-4 example —
became a deficiency-3 example, moving it into the set of `.parts.ii`.

Neither open declaration was *false* before this change; they were open questions
about a different function.

## The repair

One token: `smoothNumbers (k + 1)` unfolds to
`{m | m ≠ 0 ∧ ∀ p ∈ primeFactorsList m, p < k + 1}`, i.e. exactly "all prime factors
`≤ k`". Under it all 23 catalogue entries reproduce the deficiency the page states.

The docstring now also carries the source's clarifying clause ("divisible only by
primes $\le k$") plus one sentence naming the `< m` convention of
`Nat.smoothNumbers`, so the next reader does not have to rediscover the off-by-one.

For context rather than blame: the first commit of the introducing PR #1328 defined
the predicate by hand and correctly —

```lean
/-- A number $n$ is $k$-smooth if all its prime factors are $\le k$. -/
def IsKSmooth (k n : ℕ) : Prop := ∀ p, p.Prime → p ∣ n → p ≤ k
```

— and it was replaced by `smoothNumbers k` in response to a review suggestion to
reuse Mathlib. The `≤`/`<` difference does not appear to have been raised in that
thread, and the two later fixes to this file (#1489, #1532) did not touch it.

Verified:

- The file elaborates with no diagnostics under
  `lean -DwarningAsError=true` with the option set the `FormalConjectures` library
  uses in `lakefile.toml` (`autoImplicit=false`, `relaxedAutoImplicit=false`,
  `warn.sorry=false`, and the `ams_attribute`, `category_attribute`,
  `conditional_formal_proof`, `moduleDocstring`, `latex_docstring`, `namespace`,
  `openClassical`, `copyright` linters). I confirmed those linters are actually
  live in that configuration with two negative controls (reordering the `AMS` tags
  and deleting the `category` attribute both fail as expected).
- Both declarations still end in `sorry` and remain open; nothing became provable
  or trivial.
- The 23-entry recomputation above, under both thresholds.

Not verified: I did **not** run `lake build` on this machine, so I have not
exercised the full build (imports of downstream modules, the `test` driver, or
`decide`-based elaboration elsewhere). CI is the authoritative check.

## Deliberately not in this diff

A `test` lemma pinning e.g. `deficiency 7 3 = 1` would catch this shape of error
directly, but `deficiency` is `noncomputable` (membership in a `Set ℕ`), so such a
lemma needs `deficiency` restructured around a decidable predicate. That is a
larger change than this fix and belongs in its own PR.

## AI assistance disclosure

check, the recomputation of the catalogue, and drafting. The submitter reviewed the
declaration, the source text, the computation and the diff, and takes

### DIFF
--- a/FormalConjectures/ErdosProblems/1093.lean
+++ b/FormalConjectures/ErdosProblems/1093.lean
@@ -27,10 +27,14 @@ namespace Erdos1093
 open Finset Nat
 
 /--
-If defined, the deficiency is the count of $0 \le i < k$ such that $n - i$ is $k$-smooth.
+If defined, the deficiency is the count of $0 \le i < k$ such that $n - i$ is $k$-smooth,
+that is, divisible only by primes $\le k$.
+
+`Nat.smoothNumbers m` is the set of positive naturals all of whose prime factors are
+$< m$, so "$k$-smooth" in the sense above is `Nat.smoothNumbers (k + 1)`.
 -/
 noncomputable def deficiency (n k : ℕ) : ℕ :=
-  #{i ∈ range k | n - i ∈ smoothNumbers k}
+  #{i ∈ range k | n - i ∈ smoothNumbers (k + 1)}
 
 /--
 Are there infinitely many binomial coefficients with deficiency 1?


mo271 (commented):
Thanks!

We should check if the same mistake has been mistake in other places where `smoothNumbers` are used!

Kuberwastaken:
Checked, `smoothNumbers` appears in only one other place, Erdős 961, and it is correct there.

961 writes `Nat.smoothNumbers (k + 1)`, i.e. prime factors `< k + 1`, which is `≤ k`. So `∉ Nat.smoothNumbers (k + 1)` says "has a prime factor `> k`", matching the source exactly ("every set of $n$ consecutive integers $> k$ contains an integer divisible by a prime $> k$"). It already compensates for the Mathlib convention, so no fix is needed.

There is one small wording issue in the `sylvester_schur` docstring it says “not (k+1)-smooth”. In the usual convention, that should be “not k-smooth”. The formal statement is right; only the prose is slightly off.

I also checked the other places where smoothness is expressed differently:

369 bounds prime factors directly, so there’s no convention issue.
648 uses the largest prime factor, and 851 counts prime factors — different notions.
Millenium/Poincare’s SmoothConjectureFor is unrelated.

So 1093 seems to be the only actual instance.

mo271 (approved):
Thanks, LGTM!

PR #5018, opened 2026-08-17
fix(Books/BugeaudDistributionModuloOne): Problem10_6 reassessment

rwst:
The current version already shows a naive formalization can be closed trivially. This PR gives more solutions from the literature and corrects which variant is still open. The appended (unpublished AI) review "note" gives all the details.

[paper-BB6.pdf](https://github.com/user-attachments/files/31138944/paper-BB6.pdf)


mo271 (commented):
the proofs are a little long, perhaps this is better via the `formal_proof using formal_conjectures at [permalink]` mechanism?

mo271 (approved):
Thanks, LGTM!

PR #4899, opened 2026-08-12
Add a review skill, reports and evaluation harness

williamjblair:
Adds an advisory review procedure that compares Lean statements with their cited sources and retains findings, check evidence, and exact inputs for later inspection. Maintainers decide acceptance.

```mermaid
flowchart LR
    I["Pinned review inputs"] --> R["Semantic review<br/>Findings + gaps"]
    I --> B["Independent build<br/>Fresh container"]
    R --> A["Assemble report"]
    B --> A
    A --> O["JSON + Markdown<br/>Evidence + freshness"]
```

Scratch checks support findings; they cannot replace the independent build receipt. Missing evidence stays incomplete.

| Component | Review here |
| --- | --- |
| **Skill** — source fidelity, statement soundness, proof/status claims; prepared or manual inputs | [Procedure](https://github.com/williamjblair/formal-conjectures/blob/c06bb66ad9a240221272f73977a70ee49693df5e/.agents/skills/formal-conjectures-review/SKILL.md) |
| **Reports** — exact input bindings, retained evidence, freshness and rereviews | [Contracts](https://github.com/williamjblair/formal-conjectures/blob/c06bb66ad9a240221272f73977a70ee49693df5e/scripts/review-report/README.md) |
| **Evaluations** — paired skill/baseline runs with preserved failures and provisional grades | [Method and limitations](https://github.com/williamjblair/formal-conjectures/blob/c06bb66ad9a240221272f73977a70ee49693df5e/scripts/review-eval/README.md) |

**Validation:** 75 offline tests and skill validation pass. The [evidence packet](https://github.com/williamjblair/formal-conjectures/tree/b5710e346fc9e17317bfa614216a1f6def9a3971) retains Docker/Lean isolation controls, fixture builds and historical model runs. A [fresh isolation replay](https://github.com/williamjblair/formal-conjectures/blob/toolkit-v0.2.0rc3/toolkit/qualification/rc3-isolation.json) also passed with an explicit emulation timeout; ordinary review bounds remain unchanged. No problem modules change.

**Development evaluation — GPT-5.6 Sol high, one attempt per case and condition**

| Case | Without skill | With skill |
| --- | --- | --- |
| Known domain defect | Found; repair checked | Found; repair checked |
| Corrected counterpart | No findings | No findings |
| Jacobson candidate | Timeout; no report | Universe-scope finding; reference disputed |
| Source unavailable | Unresolved semantic finding + nit | No findings; source coverage incomplete |
[...]

PR #5020, opened 2026-08-17
docs: make AGENTS.md more concise, and split into other files

Paul-Lez:
This PR make the AGENTS.md file more concise, addresses a few comments from a previous PR that had been missed, and splits the AGENTS.md file into several files, some of which have workflow specific instructions (these could be useful later for making skills).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,147 +1,100 @@
 # Agent guidelines
 
 Formal Conjectures states open mathematical problems in Lean 4. It is a statement
-repository, not a proof repository: almost every problem is `sorry`. Each statement must say
-what its source says.
+repository, not a proof repository: almost every problem is `sorry`.
 
-[CONTRIBUTING.md](CONTRIBUTING.md) is the reference for conventions, folders, and the
-attributes. This file adds only what CONTRIBUTING does not tell you.
+[CONTRIBUTING.md](CONTRIBUTING.md) is the reference for conventions, folders, and attributes.
+Also follow any `README.md` in the directory that you change.
 
-Follow the style of an existing file. Start by copying one from the same directory. *File
-structure conventions* in CONTRIBUTING gives the skeleton, and a neighbouring file gives the
-rest: a definition with its docstring, a statement with its source, and a variant. A template
-can become incorrect. A file in the repository cannot, because the build compiles it.
+Use these detailed guides when the task needs them:
 
-## Commands
+- [PROOFS.md](PROOFS.md) for proofs and `formal_proof` claims
+- [STATEMENTS.md](STATEMENTS.md) for adding and reviewing statements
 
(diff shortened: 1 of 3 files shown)


mo271:
Thanks!

also cc-ing @williamjblair who recently re-worked the file

williamjblair:
Nice, this looks good to me. I think the ideal workflow is going to be something closer to AGENTS.md acting as general guidelines + pointing to the correct skills or references. I added an example review skill with evals here #4899

Paul-Lez:
Yeah I'm sure the extra files can be reworked a little as we go. I guess currently they also function as documentation for humans looking at the repo, while the skills that will be added can be more focused on agents.

AGENTS.md
```diff
@@ -1,147 +1,100 @@
...
-[CONTRIBUTING.md](CONTRIBUTING.md) is the reference for conventions, folders, and the
-attributes. This file adds only what CONTRIBUTING does not tell you.
+[CONTRIBUTING.md](CONTRIBUTING.md) is the reference for conventions, folders, and attributes.
+Also follow any `README.md` in the directory that you change.
 
-Follow the style of an existing file. Start by copying one from the same directory. *File
-structure conventions* in CONTRIBUTING gives the skeleton, and a neighbouring file gives the
-rest: a definition with its docstring, a statement with its source, and a variant. A template
-can become incorrect. A file in the repository cannot, because the build compiles it.
+Use these detailed guides when the task needs them:
```
mo271, line 9: what is "task" here and below?

STATEMENTS.md
```diff
@@ -0,0 +1,55 @@
...
+- the order and scope of all quantifiers
+- strict and non-strict bounds
+- all hypotheses and domain restrictions
+- equality, asymptotic equivalence, and order relations
+- the direction of implications
+- every variant and special case
+- the category and answer recorded by the source
+
+Be careful with `∃ x, P x → Q`. The intended statement is usually `∃ x, P x ∧ Q`; the first
+form can be trivially true.
```
mo271, line 32: this and some other hints here (like `opaque`) are now also linted for and the agent presumably runs`lake build --wfail` at some point, so perhaps we could drop those? perhaps still good to keep them... just some thought

mo271 (approved):
Thanks, LGTM!

AGENTS.md
Paul-Lez: Here task would be whatever the agent has been asked to do - happy to add some clarification if you think it's a good idea

STATEMENTS.md
Paul-Lez: I figured that since this is also user documentation (in a follow up PR I'll move them to set up a proper user documentation directory), it would be nice to record this sort of rule.

PR #5041, opened 2026-08-18
Erdős 269: add the three-prime smooth-prefix LCM identity

wcook04:
Closes #5051.

This PR adds a finite structural identity associated with Erdős Problem 269.

For three pairwise distinct primes `p`, `q`, `r` and **any** cutoff `x`, the
least common multiple of all values `p^i q^j r^k ≤ x` is

    p^(Nat.log p x) * q^(Nat.log q x) * r^(Nat.log r x).

Unlike my other contributions, **the proof is entirely in this repository** —
there is no external `formal_proof` link and no trust dependency outside
Mathlib. It is short: every smooth-prefix value divides the coordinatewise
maximal product, each maximal pure power occurs in the prefix, and pairwise
coprimality of powers of distinct primes gives the reverse divisibility.

The pairwise-distinctness hypotheses are load-bearing rather than decorative.
At `p = q = 2`, `r = 5`, `x = 100` the two sides are `1600` and `102400`.
A worked positive instance: at `p,q,r = 2,3,5` and `x = 100`, both sides are
`129600 = 2⁶·3⁴·5²`.

The identity holds at `x = 0` as well, where the prefix is empty and both
sides are `1`, so the statement carries no non-vacuity hypothesis.

## Category

`textbook`, following CONTRIBUTING: this is elementary valuation bookkeeping
included as a building block of a research problem, not a research-level
result. The running least common multiple is the object Erdős 269 is about,
so the identity is directly related to the problem.

## Files

- `FormalConjecturesForMathlib/NumberTheory/ThreePrimeSmoothLcm.lean` (new)
- `FormalConjecturesForMathlib.lean` (index import)
- `FormalConjectures/ErdosProblems/269.lean` (one restatement)

## Validation

- `lake --wfail build FormalConjecturesForMathlib.NumberTheory.ThreePrimeSmoothLcm`
- `lake --wfail build FormalConjecturesForMathlib`
- `lake --wfail build 'FormalConjectures.ErdosProblems.«269»'`
- `#print axioms` on both new theorems: `[propext, Classical.choice, Quot.sound]`,
  no `sorryAx`
- No `sorry`, `axiom` or `opaque` in the new file

## Scope

This does not prove rationality, irrationality or transcendence of `series`.
The finite-prime case of Erdős 269, including the three-prime case, remains
open. The two existing open declarations are untouched. The identity is
elementary and no originality or priority claim is made.

[...]

PR #5038, opened 2026-08-18
Erdős 249: add dyadic-kernel, visible-lattice, and 10^34 denominator-exclusion variants

wcook04:
Closes #5037.

`249.lean` asks whether $\sum_n \phi(n)/2^n$ is irrational. This PR does not answer that.

The strongest new statement is a machine-checked finite exclusion for that constant:

```lean
theorem erdos_249.variants.denominator_gt_79639646646701375323355774875831053 :
    ∀ p : ℚ, p.den ≤ 79639646646701375323355774875831053 →
      (∑' n : ℕ, ((Nat.totient n : ℝ)) / (2 : ℝ) ^ n) ≠ (p : ℝ)
```

If the series is rational, its reduced denominator exceeds $7.96 \times 10^{34}$. A rational with a still-larger denominator is not ruled out.

The other three statements are structural facts about the same constant (dyadic-kernel rank $2^e+1$, an explicit odd-core basis, and the half-open visible-lattice generating-function identity). None of them decides irrationality.

Proofs:

- denominator exclusion: https://github.com/wcook04/plectis-lean-erdos249-257/blob/f88e8b686908010a43e9078dda49abbabcfc4079/Erdos249257/CertificateKernel.lean#L18384-L18389
- rank: https://github.com/wcook04/plectis-lean-erdos249-257/blob/92d73cf7a84a1993817020d615b9b046c6ac4b19/adapters/FormalConjecturesVariants.lean#L276-L280
- basis: https://github.com/wcook04/plectis-lean-erdos249-257/blob/92d73cf7a84a1993817020d615b9b046c6ac4b19/adapters/FormalConjecturesVariants.lean#L284-L288
- visible lattice: https://github.com/wcook04/plectis-lean-erdos249-257/blob/c04870f7f7f2166f38fc4077033792ef6486f209/Erdos249257/GeometricCoprimality.lean#L120-L151

Repository: https://github.com/wcook04/plectis-lean-erdos249-257

The first two coefficient theorems need six definitions in `FormalConjecturesForMathlib/NumberTheory/TotientKernel.lean`. The visible-lattice identity and the denominator exclusion use only Mathlib vocabulary, so they link directly.

No originality or priority claim is made for the Farey/gap method. The contribution is the Lean-checked bound on this specific series.

AI Usage Disclosure: the linked Lean development and this contribution were produced with AI assistance. I have reviewed the statement identity of the four variants, the linked proofs, and this diff, and I take responsibility for the submission.

> [!NOTE]
[...]

PR #5059, opened 2026-08-19
Erdős 243: add Koizumi's Sylvester and Millin rigidity corollaries

### LINKED ISSUE #5058: Erdős 243: record Koizumi's Sylvester and Millin rigidity corollaries
`FormalConjectures/ErdosProblems/243.lean` currently contains only the open
statement: if $a_n$ is increasing with $a_n/a_{n-1}^2 \to 1$ and
$\sum 1/a_n \in \mathbb{Q}$, must $a_n = a_{n-1}^2 - a_{n-1} + 1$ eventually?

Koizumi, *Irrationality of the reciprocal sum of doubly exponential sequences*,
[arXiv:2504.05933](https://arxiv.org/abs/2504.05933) (2025), studies precisely
this. His Question 5 (p. 3) is the problem verbatim, and his Theorem 1 yields
two rigidity corollaries that would sit naturally in the file:

- **Corollary 2.** If $2/3 \le a_n^2/a_{n+1} \le 4/3$ and $\sum 1/a_n = 1$,
  then $a_n$ is Sylvester's sequence $2, 3, 7, 43, 1807, \dots$.
- **Corollary 3.** If $a_n^2/a_{n+1} \le 2/3$ and
  $\sum 1/a_n = (5-\sqrt5)/2$, then $a_n = F_{2^n}$.

These are a useful contrast with `erdos_243`: the ratio hypothesis is a bound
rather than a limit, the sum is pinned to an exact value rather than merely
rational, and the conclusion is exact equality at every index rather than an
eventual recurrence.

Sylvester's sequence does not appear to be in Mathlib, so it would need a small
`FormalConjecturesForMathlib` file.

(The same paper's Theorem 4 is already cited in `263.lean`.)

wcook04:
Closes #5058.

`243.lean` currently holds only the open statement. Koizumi,
[arXiv:2504.05933](https://arxiv.org/abs/2504.05933) (2025), studies exactly
this problem — his Question 5 (p. 3) is Erdős 243 verbatim — and proves two
rigidity corollaries this adds.

**`erdos_243.variants.sylvester_of_tsum_eq_one`** (his Corollary 2, p. 2): if
$2/3 \le a_n^2/a_{n+1} \le 4/3$ and $\sum 1/a_n = 1$, then $a_n$ is Sylvester's
sequence $2, 3, 7, 43, 1807, \dots$. The contrast with `erdos_243` is the point:
there the ratio *tends to* $1$ and the sum is merely rational, and the
conclusion is an *eventual* recurrence; here a two-sided bound plus an exact sum
pins the sequence down at every index.

**`erdos_243.variants.fib_of_tsum_eq_millin`** (his Corollary 3): if
$a_n^2/a_{n+1} \le 2/3$ and $\sum 1/a_n = (5-\sqrt5)/2$, then $a_n = F_{2^n}$.

**`erdos_243.sylvester_satisfies_recurrence`** — a one-line `test` recording
that Sylvester's sequence satisfies the recurrence in the conclusion of
`erdos_243`, so the two statements are visibly about the same recurrence.

Sylvester's sequence is not in Mathlib, so
`FormalConjecturesForMathlib/NumberTheory/Sylvester.lean` defines it with basic
API (`two_le_sylvester`, `strictMono_sylvester`, and the first values). That
file is fully proved — no `sorry`, and every declaration audits to
`[propext, Classical.choice, Quot.sound]` or less.

Checks:

- Builds clean against current `main` (`33c6a2d`): 0 errors, 0 warnings.
- Both corollaries were checked numerically against their claimed witnesses
  before formalising. Sylvester gives ratios
  $4/3, 9/7, 49/43, 1849/1807$ — all in $[2/3, 4/3]$, with $s_1^2/s_2 = 4/3$
  meeting the bound exactly — and $\sum 1/s_n = 1$. The Fibonacci witness gives
  ratios converging to $1/\sqrt5 \approx 0.447 \le 2/3$ and
  $\sum 1/F_{2^n} = 1.381966011250105 = (5-\sqrt5)/2$.
- The same paper's Theorem 4 is already cited in `263.lean`; these corollaries
  are not used there.

No originality is claimed — the theorems are Koizumi's. The contribution is
their formal statement plus the Sylvester definition needed to express them.

AI Usage Disclosure: this contribution was produced with AI assistance. I have
read the corollaries in the source paper, checked the index conventions against
his $n \ge 1$ numbering, verified the witnesses numerically, and reviewed this
diff, and I take responsibility for the submission.

> [!NOTE]
> These are adjacent solved results, not special cases of `erdos_243`, which
> remains open and is untouched.

### DIFF
--- a/FormalConjectures/ErdosProblems/243.lean
+++ b/FormalConjectures/ErdosProblems/243.lean
@@ -19,7 +19,10 @@ import FormalConjecturesUtil
 /-!
 # Erdős Problem 243
 
-*Reference:* [erdosproblems.com/243](https://www.erdosproblems.com/243)
+*References:*
+- [erdosproblems.com/243](https://www.erdosproblems.com/243)
+- [Koi25] Koizumi, J., Irrationality of the reciprocal sum of doubly exponential
+  sequences. arXiv:2504.05933 (2025).
 -/
 
 open Filter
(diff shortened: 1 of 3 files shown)


wcook04:
Note for whoever merges this: #5038, #5041 and #5059 each add one `public import` line to `FormalConjecturesForMathlib.lean` in the same alphabetical slot (after `SmoothScale`, before `WallSunSunPrimes`). They are independently mergeable against `main`. Merging one will need a one-line rebase on the other two. The new ForMathlib files themselves do not overlap.

PR #5060, opened 2026-08-19
fix(Paper): scope the LatinSquare answer slots inside their quantifier

### LINKED ISSUE #5009: Paper/LatinSquare: two `answer( )` slots sit outside a `variable {n}` binder — `answer(False)` is unreachable, and the AnswerLinter cannot see it
## Summary

`Paper/LatinSquare.lean` opens with `variable {n : ℕ}` at line 33. Two `research open`
declarations use `n` in their statements but not in their own binder lists, so each elaborates
as `∀ {n : ℕ}, answer(sorry) ↔ P n` with a **single** answer slot outside the `∀ n`. The
consequence is that `answer(False)` is unreachable in one case and, in the other, the
declaration is provable only if `P n` has the same truth value for every `n` — neither of which
is the intended reading of the `answer(sorry) ↔ …` idiom.

The repository's `AnswerLinter` exists to catch exactly this ("Move the quantifiers outward…"),
but its `contains_early_args` helper reads `decl[2][0]`, the binder list in the theorem's own
`declSig`, and `variable`-introduced binders are not there. I include a machine-checked
demonstration of the blind spot below, since it is probably the more useful half of this report.

https://github.com/google-deepmind/formal-conjectures/blob/638da20efd8eeeed2993fc2550fc596dc90c1ce8/FormalConjectures/Paper/LatinSquare.lean#L33

```lean
variable {n : ℕ}
```

https://github.com/google-deepmind/formal-conjectures/blob/638da20efd8eeeed2993fc2550fc596dc90c1ce8/FormalConjectures/Paper/LatinSquare.lean#L53-L60

```lean
/-- Conjecture 3.2 in [Wa2011]: Each Latin square of odd order has at least one transversal. -/
theorem oddOrderLatinSquareTransversal : answer(sorry) ↔
    Odd n → ∀ (L : LatinSquare n), ∃ σ, IsTransversal L σ := by
  sorry
```

https://github.com/google-deepmind/formal-conjectures/blob/638da20efd8eeeed2993fc2550fc596dc90c1ce8/FormalConjectures/Paper/LatinSquare.lean#L84-L91

```lean
/-- Conjecture 5.1 in [Wa2011]: Every latin square has a near-transversal -/
theorem latinSquareNearTransversal : answer(sorry) ↔
    ∀ (L : LatinSquare n), ∃ ρ σ, IsNearTransversal L ρ σ := by
  sorry
```

## Consequences

**`oddOrderLatinSquareTransversal`: `answer(False)` is unreachable.** The right-hand side is
`Odd n → …`, which is *vacuously true* whenever `n` is even. So the answer is forced to `True`
whatever the status of Conjecture 3.2, and the declaration then asserts the conjecture outright
rather than asking it:

```lean
theorem answer_forced_true {Q : ℕ → Prop} (A : Prop) (h : ∀ n : ℕ, A ↔ (Odd n → Q n)) : A :=
  (h 2).mpr (fun h2 => absurd h2 (by decide))
```

The `answer(sorry) ↔ …` idiom exists so that a solver can record either verdict; here one
verdict cannot be recorded.

**`latinSquareNearTransversal`: provable only if the conjecture is `n`-uniform.** Same scoping,
without the vacuity: `∀ {n}, A ↔ P n` says every `P n` has the truth value `A`. If the
near-transversal conjecture held for some orders and failed for others, **no** answer would make
the declaration provable.

The sibling `latinSquareOrder11Transversal` is immune — it fixes `LatinSquare 11`, so `n` never
enters — and is not part of this report.

## The linter blind spot

Machine-checked against Mathlib v4.27.0 (`lean -DautoImplicit=false
-DrelaxedAutoImplicit=false`, exit 0):

```lean
set_option linter.style.answer_attribute true

/-- Explicit binder in the `declSig`: the linter fires. -/
theorem explicitBinder (m : ℕ) : answer(sorry) ↔ (0 < m) := by sorry

section
variable {n : ℕ}

/-- Same statement, binder introduced by `variable`: the linter is silent. -/
theorem sectionBinder : answer(sorry) ↔ (0 < n) := by sorry

end
```

Output — one warning, on `explicitBinder` only:

```
P7.lean:6:0: warning: Move the quantifiers outward:
instead of
`theorem foo (bar : ℕ) (baz : ℕ) : answer(sorry) ↔ ...`
likely the intention was
`theorem foo : answer(sorry) ↔ ∀ᵉ (bar : ℕ) (baz : ℕ), ...`
```

Nothing is reported for `sectionBinder`, which is the shape both `LatinSquare` declarations
have. Extending `contains_early_args` to consult the elaborated declaration's telescope (or the
section variables in scope) rather than the surface `declSig` binders would close this; that
would fall under #1407.

## Suggested fix

Move the quantifier inside, as the linter's own message says:

```lean
theorem oddOrderLatinSquareTransversal : answer(sorry) ↔
    ∀ n : ℕ, Odd n → ∀ (L : LatinSquare n), ∃ σ, IsTransversal L σ := by
  sorry

theorem latinSquareNearTransversal : answer(sorry) ↔
    ∀ n : ℕ, ∀ (L : LatinSquare n), ∃ ρ σ, IsNearTransversal L ρ σ := by
  sorry
```

I have not opened a PR: #4965 is open against this file and would conflict, and whoever lands
first can reasonably carry this change.

## What is *not* claimed

Conjectures 3.2 and 5.1 of [Wa2011] are untouched; both remain open. `molsExistenceProblem`
(line 212) is closable by `rfl` with the set on its right-hand side, but that is the case
CONTRIBUTING explicitly places out of scope ("one can provide trivial answers that aren't
mathematically interesting…"), so I do not report it.

## Duplicate check

Offline sweep of all 1747 issues and 3237 PRs (with per-PR file lists) on `LatinSquare`,
`molsExistenceProblem`, `MOLS`, `Transversal`, and the path
`FormalConjectures/Paper/LatinSquare.lean`. The open PR **#4965** touches this file but changes
only `oddOrderLeq9LatinSquareTransversal` (`answer(sorry)` → `answer(True)`, a `research solved`
declaration), not either declaration reported here — I re-checked its diff immediately before
filing. Issues #2270/#2271/#2272 are the original formalization requests. Not in #4896, #4923 or
#4927.

## Related

Same defect family as #5005 (five instances in `GreensOpenProblems`) and #5008
(`Paper/VoronovskajaTypeFormula`); this one is distinguished by being invisible to the linter
that was written for it.

Found while auditing `research open` declarations in the non-Wikipedia collections:

## AI assistance disclosure

Prepared with AI assistance (Claude, Anthropic) for the audit, the Lean witnesses and drafting.
The submitter reviewed the declarations, the probe output, and the suggested repair, and takes

### Choose either option

theebayuser:
`oddOrderLatinSquareTransversal` and `latinSquareNearTransversal` pick up the section
`variable {n : ℕ}`, so each elaborated as `∀ {n}, answer(sorry) ↔ P n` with the answer slot
outside the quantifier. As analyzed in #5009, `answer(False)` was then unreachable for the
odd-order statement (`Odd n → …` is vacuous at even `n`, forcing the answer to `True`), and the
near-transversal statement was provable only if the conjecture held uniformly in `n`. This moves
`∀ n` inside the iff for both.

- Docstrings unchanged: both remain the verbatim Conjectures 3.2 and 5.1 of [Wa2011], which
  quantify over all (odd) orders, so the `∀ n` form is the faithful reading. Both stay
  `research open`. (The `Odd n → …` shape dates to #4186; the scoping issue is orthogonal to
  that fix.)
- Not touched: `latinSquareOrder11Transversal` (fixed order 11) and
  `oddOrderLeq9LatinSquareTransversal` (binds its own `∀ n ≤ 9`; edited in #4965, which this PR
  does not conflict with — different declarations). `molsExistenceProblem` is out of scope per
  CONTRIBUTING, as #5009 itself notes.
- The AnswerLinter blind spot #5009 reports (section variables are invisible to
  `contains_early_args`) will be addressed separately under #1407.

closes #5009

- [x] `lake --wfail build FormalConjectures.Paper.LatinSquare` passes with no warnings
- [x] Elaborated statements read back via `extract_names`: the answer slot is now outermost,
      `∀ n` inside the iff; docstrings byte-identical

### DIFF
--- a/FormalConjectures/Paper/LatinSquare.lean
+++ b/FormalConjectures/Paper/LatinSquare.lean
@@ -56,7 +56,7 @@ Each Latin square of odd order has at least one transversal.
 -/
 @[category research open, AMS 5]
 theorem oddOrderLatinSquareTransversal : answer(sorry) ↔
-    Odd n → ∀ (L : LatinSquare n), ∃ σ, IsTransversal L σ := by
+    ∀ (n : ℕ), Odd n → ∀ (L : LatinSquare n), ∃ σ, IsTransversal L σ := by
   sorry
 
 /--
@@ -87,7 +87,7 @@ Every latin square has a near-transversal
 -/
 @[category research open, AMS 5]
 theorem latinSquareNearTransversal : answer(sorry) ↔
-    ∀ (L : LatinSquare n), ∃ ρ σ, IsNearTransversal L ρ σ := by
+    ∀ (n : ℕ) (L : LatinSquare n), ∃ ρ σ, IsNearTransversal L ρ σ := by
   sorry
 
 /-- The number of transversals of the Cayley table of the cyclic group $\mathbb{Z}_n$ -/
--- a/FormalConjectures/Paper/LatinSquare.lean
+++ b/FormalConjectures/Paper/LatinSquare.lean
@@ -30,13 +30,11 @@ This file formalizes some conjectures and theorems around latin squares.
 
 namespace LatinSquare
 
-variable {n : ℕ}
-
 /--
 Two latin squares of the same order are orthogonal if superimposing them gives each ordered pair of
 symbols at most once.
 -/
-def Orthogonal (L M : LatinSquare n) : Prop :=
+def Orthogonal {n : ℕ} (L M : LatinSquare n) : Prop :=
   Function.Injective fun p : Fin n × Fin n => (L.mat p.1 p.2, M.mat p.1 p.2)
 
 /-- A family of latin squares is mutually orthogonal if any two distinct members are orthogonal. -/


FormalConjectures/Paper/LatinSquare.lean
mo271: can we move this to the definition of `Orthogonal`, which seems to be the only place that uses it now?

mo271 (commented):
Thanks, good catch!

FormalConjectures/Paper/LatinSquare.lean
theebayuser: sure!

mo271 (approved):

PR #5061, opened 2026-08-19
ci: wrap lake build like in mathlib

mo271:
also enables `gh-problem-matcher-wrap`


scripts/lake-build-wrapper.py
```diff
@@ -0,0 +1,389 @@
+#!/usr/bin/env python3
+# Copyright 2026 The Formal Conjectures Authors.
```
eric-wieser, line 2: I would guess you need the mathlib copyright here too.

scripts/lake-build-wrapper.py
```diff
@@ -0,0 +1,389 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Adapted from leanprover-community/mathlib4 (scripts/lake-build-wrapper.py):
+https://github.com/leanprover-community/mathlib4/blob/master/scripts/lake-build-wrapper.py
```
eric-wieser: Can you include the actual commit you copied from?

eric-wieser (approved):

scripts/lake-build-wrapper.py
mo271: done

scripts/lake-build-wrapper.py
mo271: added

PR #4397, opened 2026-07-05
feat(ErdosProblems/660): add convex-polyhedron distinct-distances statement

theebayuser:
Resolves #2123

## Summary
- Formalises [Erdős Problem 660](https://www.erdosproblems.com/660) (open): whether the $n$ vertices of a convex polyhedron in $\mathbb{R}^3$ determine at least $(1-o(1))\frac{n}{2}$ distinct distances.
- Adds `altman_planar`, the planar analogue proved by Altman: $n$ vertices of a convex polygon determine at least $\lfloor n/2 \rfloor$ distinct distances.
- Adds `Er75f`, the variant from the website: Erdős's claim in [Er75f] that Altman proved the vertices determine $\gg n$ many distinct distances (no reference given, so stated as `research open` with `answer(sorry)`).
- `IsPolyhedronVertices` reuses Mathlib's `ConvexIndependent` for convex position rather than introducing a new definition.
- Generalises `EuclideanGeometry.distinctDistances` in `FormalConjecturesForMathlib/Geometry/2d.lean` from `Finset ℝ²` to `Finset E` over any `MetricSpace E`, so the same definition is shared between the $\mathbb{R}^2$ and $\mathbb{R}^3$ statements (backward compatible; existing `ℝ²` call sites are unaffected).

## Test plan
- [x] `lake --wfail build 'FormalConjectures.ErdosProblems.«660»'` passes with no errors or warnings.
- [x] Rebuilt all existing users of `distinctDistances` (`91`, `98`, `655`, `659`, `1082`) after the generalisation — no regressions.
- [x] Problem statement checked against the live erdosproblems.com page.

PR #5063, opened 2026-08-19
refactor(ForMathlib): generalize distance API from ℝ² to MetricSpace

theebayuser:
Follow-up deferred from the #4397 review:

- `distanceMultiplicity` and `distinctDistancesFrom` move from `Geometry/2d.lean` to `Geometry/Metric.lean`, generalized from `ℝ²` to any `MetricSpace`, joining `distanceSet`/`distinctDistances` which already live there.
- `unitDistancePairsCount` was a duplicate encoding of the pre-existing `unitDistNum`, so it is deleted; its two call sites (ErdosProblems/90, 96) now use `unitDistNum`.
- New lemma `unitDistNum_le_choose_two` replaces the two hand-rolled `BddAbove` proofs in those files.
- No statement semantics change in this PR.

- [x] `lake --wfail build FormalConjecturesForMathlib` + all caller modules (90, 96, 756, 94, 958, 1082, 1084, 1085)
- [x] `lake --wfail test`; `#print axioms unitDistNum_le_choose_two` = `[propext, Classical.choice, Quot.sound]`

### DIFF
--- a/FormalConjectures/ErdosProblems/90.lean
+++ b/FormalConjectures/ErdosProblems/90.lean
@@ -47,7 +47,7 @@ open Finset
 The set of all possible numbers of unit distances for a configuration of $n$ points.
 -/
 noncomputable def unitDistanceCounts (n : ℕ) : Set ℕ :=
-  {unitDistancePairsCount points | (points : Finset ℝ²) (_ : points.card = n)}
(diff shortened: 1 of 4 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5065, opened 2026-08-19
fix(ForMathlib): exclude the base point in distinctDistancesFrom

### LINKED ISSUE #5064: distinctDistancesFrom counts the self-distance 0
**File:** `FormalConjecturesForMathlib/Geometry/2d.lean` (moving to `Geometry/Metric.lean` in #5063)

`distinctDistancesFrom points pt` is documented as "the number of distinct distances between a given point and all other points", and both consumers state $R(x_i)=\#\{\lvert x_j-x_i\rvert : j\neq i\}$ (see the docstrings of `maximalDistinctDistancesFrom` and `erdos_653`). But the definition takes the image over all of `points`, so whenever `pt ∈ points` the self-distance $0$ is included and the count is off by one from the documented meaning.

The off-by-one is currently compensated downstream: `erdos_1082.parts.ii` writes `distinctDistancesFrom A a - 1`. `maximalDistinctDistancesFrom` happens to be value-invariant (every $R(x_i)$ shifts by exactly 1), so no statement is wrong today, but the definition contradicts its own docstring and the compensating `- 1` is a trap for future users.

Fix: erase `pt` before taking the image, and drop the `- 1` in 1082 (propositionally equal, since `dist x pt = 0 ↔ x = pt`). PR to follow.

- [x] I plan on working on this
- [ ] This issue is up for grabs

theebayuser:
Closes #5064. Stacked on #5063 (the diff shows both commits until that one merges; only the last commit is this PR).

- `distinctDistancesFrom` now erases `pt` before taking the image, matching its docstring and the $j\neq i$ in the docstrings of both consumers.
- `erdos_1082.parts.ii` drops its compensating `- 1`: since `dist x pt = 0 ↔ x = pt`, the old count is exactly the new count plus one, so the statement is propositionally equal and the linked `formal_proof` permalink (0aca4d71) still establishes it. The docstring's "each point only determines 3 distances" now also matches the definition's value.
- `maximalDistinctDistancesFrom` / `erdos_653` are value-invariant: every $R(x_i)$ shifts by exactly 1, which preserves the number of distinct values.

- [x] `lake --wfail build FormalConjecturesForMathlib FormalConjectures.ErdosProblems.«1082» FormalConjectures.ErdosProblems.«653»`
- [x] `python3 scripts/check_category_warnings.py` on the elaborated 1082: clean

### DIFF
--- a/FormalConjectures/ErdosProblems/1082.lean
+++ b/FormalConjectures/ErdosProblems/1082.lean
@@ -51,6 +51,6 @@ This counterexample has originally been found by Heiko Harborth.
 @[category research solved, AMS 51, formal_proof using formal_conjectures at "https://github.com/google-deepmind/formal-conjectures/blob/0aca4d71095301c0fd2dca32611b7addb2ea735c/FormalConjectures/ErdosProblems/1082.lean"]
 theorem erdos_1082.parts.ii : answer(False) ↔
     ∀ (A : Finset ℝ²) (hA : A.Nonempty) (hA_n3c : NonTrilinear (A : Set ℝ²)),
-    ∃ (a : ℝ²) (ha : a ∈ A), A.card / 2 ≤ distinctDistancesFrom A a - 1 := by
+    ∃ (a : ℝ²) (ha : a ∈ A), A.card / 2 ≤ distinctDistancesFrom A a := by
   sorry
 end Erdos1082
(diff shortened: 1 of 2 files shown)


mo271 (approved):
Thanks, LGTM

PR #5071, opened 2026-08-20
chore(Util/StubLinter): correctly handle Prop-valued instances

YaelDillies:
These should not be considered stub definitions since they are Prop-valued.

### DIFF
--- a/FormalConjectures/Wikipedia/EllipticCurveRank.lean
+++ b/FormalConjectures/Wikipedia/EllipticCurveRank.lean
@@ -49,7 +49,7 @@ open Module (finrank)
 Consequently, the rank is always finite, so `finrank ℤ E⟮K⟯ = 0` really means that the group of
 rational points is torsion, not that it is of infinite rank. -/
 @[category research solved, AMS 11 14]
-theorem mordell_weil {K} [Field K] [NumberField K] [DecidableEq K] (E : WeierstrassCurve K)
+instance mordell_weil {K} [Field K] [NumberField K] [DecidableEq K] (E : WeierstrassCurve K)
     [E.IsElliptic] : Module.Finite ℤ E⟮K⟯ := by
   sorry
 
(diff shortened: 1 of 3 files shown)


FormalConjectures/Wikipedia/EllipticCurveRank.lean
```diff
@@ -49,7 +49,7 @@ open Module (finrank)
 Consequently, the rank is always finite, so `finrank ℤ E⟮K⟯ = 0` really means that the group of
 rational points is torsion, not that it is of infinite rank. -/
 @[category research solved, AMS 11 14]
-theorem mordell_weil {K} [Field K] [NumberField K] [DecidableEq K] (E : WeierstrassCurve K)
+instance mordell_weil {K} [Field K] [NumberField K] [DecidableEq K] (E : WeierstrassCurve K)
```
Paul-Lez, line 52: Hm I think at one point we decided that `instance` shouldn't have a `category` tag (i.e. problems can't be stated as instances). Should we go back on this?

YaelDillies: Where did this discussion happen?

mo271: Don't know where this discussion happen, but there might be downstream stuff that expects it (which could also be fixed) For instance (no pun intended) I don't even know if the `extract_names` script handles it correctly. 
In any case it *can* be easily avoided in all `instance`s, right? By preluding it with a named theorem

```
@[category research solved]
theorem foo : bar := by 
sorry

instance bar := foo
```

So we could lint for and forbid category tags for instances without loosing anything (except possibly convenience)

Paul-Lez: I think this happened in gChat (possibly DMs between me and Moritz) about a year ago or even longer!

YaelDillies: I don't really mind, but what's your rationale for disallowing this? that statements with a category tag should be called explicitly?

mo271: One idea to have names to statements is:  When people start solving them (and keep track in the future lean-lang/eval for example, but really whereever) is easier to refer to them if they have a name

YaelDillies: Sure, but that's an argument against unnamed instances, not against instances? Paul told me privately that a reason was that our statement extraction used to perform badly on instances. Has that changed?

PR #5072, opened 2026-08-20
feat(Wikipedia/EllipticCurveRank): new curve rank 30

mo271:


mo271:
@alreadydone Can you have a look? You added the original formalisation (thanks again for that!)

FormalConjectures/Wikipedia/EllipticCurveRank.lean
```diff
@@ -152,6 +153,13 @@ theorem twentyone_le_rank_height_count_asymptotic :
       ∀ H : ℕ, 1 < H → {E ∈ heightLE H | 21 ≤ E.rank}.ncard ≤ (H : ℝ) ^ f H := by
   sorry
 
+/-- Is there an elliptic curve over ℚ of rank at least 31?
+The largest known rank of an elliptic curve over ℚ as of 2026 is at least 30.
+See https://elliptic-rank.icarm.cloud/curve/273. -/
+@[category research open, AMS 11 14]
+theorem exists_rank_ge_thirtyone : ∃ E : RatEllipticCurve, 31 ≤ E.rank := by
+  sorry
```
alreadydone, line 161: This is a consequence of `unbounded_rank_conjecture` above; do we want it? You'd have to update it if a new rank record is found ...

FormalConjectures/Wikipedia/EllipticCurveRank.lean
```diff
@@ -160,6 +168,38 @@ open _root_.WeierstrassCurve
...
+
+/-- The rank of the ranksunbounded curve is exactly 30. -/
+@[category research open, AMS 11 14]
+theorem rank_ranksunbounded30 : finrank ℤ ranksunbounded30⟮ℚ⟯ = 30 := by
+  sorry
+
 /-- The elliptic curve over ℚ of rank at least 29 found by Elkies and Klagsbrun in 2024.
 It has rank exactly 29 assuming the generalized Riemann hypothesis. -/
 def elkiesKlagsbrun29 : WeierstrassCurve ℚ where
```
alreadydone, line 205: https://leanprover.zulipchat.com/#narrow/channel/600132-Bridging-Lean-and-the-LMFDB-workshop/topic/Certifying.20elliptic.20curves/near/607989160 has a sorry-free proof that this curve has rank at least 29 that might be worth linking to, and https://leanprover.zulipchat.com/#narrow/channel/600132-Bridging-Lean-and-the-LMFDB-workshop/topic/Certifying.20elliptic.20curves/near/612148082 has some discussion how to certify a rank 1 curve. (@Paul-Lez had access to the channel and I've added you too.)

alreadydone (approved):
Thanks, LGTM!

FormalConjectures/Wikipedia/EllipticCurveRank.lean
mo271: That's fantastic to hear, and yes we should add that (either completely or as using the `formal_proof using formal_conjectures at [permalink]` mechanism. 

But not here, but rather in a follow-up. 

CC-ing @b-mehta, perhaps they are up for it?!

FormalConjectures/Wikipedia/EllipticCurveRank.lean
mo271: Yes! Let's have both variants!

It is not entirely clear which one implies which (if you think there might be a disproof of either statement)... 
Of course everyone expect unbounded, but I think finding a curve of 31 would be notable and we should then update it here....

FormalConjectures/Wikipedia/EllipticCurveRank.lean
alreadydone: I just read on ICARM Zulip that the rank 30 lower bound has also been Lean verified: https://icarm.zulipchat.com/#narrow/channel/519875-general/topic/Elliptic.20Curve.20Rank.20Leaderboard/near/617688514

FormalConjectures/Wikipedia/EllipticCurveRank.lean
alreadydone: This file also contains a conjecture that contradicts unboundedness conjecture, so I wouldn't say "everyone" ...

FormalConjectures/Wikipedia/EllipticCurveRank.lean
b-mehta: Both of these proofs were using my machinery, the rank 29 and rank 30 ones. I'm trying to clean up my repo to make it public today.

b-mehta: To be precise, I proved a slightly different statement, involving
```
def HasRankGE (W : WeierstrassCurve ℚ) (n : ℕ) : Prop :=
  ∃ H : Submodule ℤ W.toAffine.Point, Module.Finite ℤ H ∧ n ≤ Module.finrank ℤ H
```
instead, which is meaningful even without Mordell-Weil. But of course Mordell-Weil is true, so it shouldn't make a difference!

Paul-Lez (approved):
Very cool! I didn't expect this file to be updated so soon!

FormalConjectures/Wikipedia/EllipticCurveRank.lean
mo271: fantastic to hear, @b-mehta!
and Mordell-Weil, we have at least the statement of a version of it

https://github.com/google-deepmind/formal-conjectures/blob/9c4d5821819656af53c5473ded2116ea14a7ff1c/FormalConjectures/Wikipedia/EllipticCurveRank.lean#L52

 in the same file...

alreadydone: Mordell-Weil was done by Michael Stoll [last month](https://leanprover.zulipchat.com/#narrow/channel/116395-maths/topic/thoughts.20on.20elliptic.20curves/near/612117156), and shortly after the Mordell-Weil group of a rank 1 curve [was determined](https://leanprover.zulipchat.com/#narrow/channel/116395-maths/topic/thoughts.20on.20elliptic.20curves/near/612863469).

PR #5073, opened 2026-08-20
State Carathéodory umbilics via fundamental forms

Paul-Lez:
this should be merged into the previous PR! (didn't mean to open it!)


Paul-Lez:
Superseded: the fundamental-form statement changes will be folded into #5066, and the corresponding proof adaptation into #5070.

PR #5077, opened 2026-08-20
feat(FormalConjecturesUtil): linter for answer(sorry) in solved statements

theebayuser:
follow-up pr to #4965: a linter that catches the `answer(sorry)` placeholders that
PR fixed by hand. Part of #1407.

`linter.style.category_answer` warns when a declaration tagged `@[category research solved]`
still writes `answer(sorry)`. The check has to be syntactic: under the default `google.answer`
setting, `answer(sorry)` at `Prop` elaborates to a bare `True`, so by the time the term exists
the placeholder is gone. It scans the whole `declSig`, so it also catches value-position
answers such as `IsGreatest S answer(sorry)`, not just `answer(sorry) ↔ …`.

This is the category-aware complement to #1958's `AnswerLinter`, which checks quantifier
scoping but does not read the category.

- The shared `toCategorySyntax`/`toCategories` helpers move from `CategoryDocstringLinter` into
  `Attributes/Basic` so both linters use them.
- Six statements deliberately keep the placeholder and opt out with
  `set_option linter.style.category_answer false in`, each with a comment giving the reason:
  independence of ZFC (1119, 1175), constants not known explicitly (416, twice, and 975), and a
  TODO in `GraphConjecture34`.
- In 1007 and 1026 the question-form headline and the answered variant say the same thing twice,
  so this PR merges each pair into one theorem carrying the answer, rather than opting out.
- Off by default; enabled repo-wide through `weak.linter.style.category_answer` in
  `lakefile.toml`, matching how the other style linters are switched on.
- A repo-wide scan finds no `research solved` declaration with a surviving `answer(sorry)`
  beyond those six. For reference, `main` currently has 1575 such declarations with 8 carrying
  the placeholder; this branch has 1573 with none, the difference being the two merged pairs.

Re-verified on current `main` at v4.33.1 after the 08-23 toolchain bumps:

- [x] `lake --wfail build FormalConjecturesForMathlib FormalConjecturesUtil` — clean (8896 jobs)
- [x] `lake --wfail test` — clean (8903 jobs); the new
      `FormalConjecturesTest/Util/Linters/CategoryAnswerLinter.lean` pins both the firing and the
      silent cases with `#guard_msgs`
- [x] Negative control: temporarily removing one opt-out makes the linter fire at the expected
      line, and restoring it silences the warning again

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -335,6 +335,13 @@ is outside of the scope of this repository.
   problem has been solved, `answer(sorry)` should be replaced by
   `answer(True)` or `answer(False)` accordingly.
 
+  The `linter.style.category_answer` linter enforces this: it warns when a
+  declaration tagged `@[category research solved]` still uses `answer(sorry)`.
+  A statement that deliberately keeps the placeholder, for instance because the
+  problem is independent of ZFC or because the answer is a constant that is not
+  known explicitly, can opt out with
+  `set_option linter.style.category_answer false in`.
+
   If the problem is not stated as a question, the following style is preferred:
 
   ```lean
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -335,13 +335,6 @@ is outside of the scope of this repository.
   problem has been solved, `answer(sorry)` should be replaced by
   `answer(True)` or `answer(False)` accordingly.
 
-  The `linter.style.category_answer` linter enforces this: it warns when a
-  declaration tagged `@[category research solved]` still uses `answer(sorry)`.
-  A statement that deliberately keeps the placeholder, for instance because the
-  problem is independent of ZFC or because the answer is a constant that is not
-  known explicitly, can opt out with
-  `set_option linter.style.category_answer false in`.
-
   If the problem is not stated as a question, the following style is preferred:
 
   ```lean
--- a/FormalConjectures/ErdosProblems/1007.lean
+++ b/FormalConjectures/ErdosProblems/1007.lean
@@ -36,6 +36,9 @@ variable {V : Type*}
 /-- The complete tripartite graph $K_{1,3,3}$. -/
 abbrev K133 := SimpleGraph.completeMultipartiteGraph fun i : Fin 3 => Fin (![1, 3, 3] i)
 
+-- The headline keeps the question form; the answer is recorded in
+-- `erdos_1007.variants.dimension_four` below.
+set_option linter.style.category_answer false in
 /--
 The dimension of a graph $G$ is the minimal $n$ such that $G$ can be embedded in $\mathbb{R}^n$
 such that every edge of $G$ is a unit line segment.
--- a/FormalConjectures/ErdosProblems/1007.lean
+++ b/FormalConjectures/ErdosProblems/1007.lean
@@ -36,30 +36,18 @@ variable {V : Type*}
 /-- The complete tripartite graph $K_{1,3,3}$. -/
 abbrev K133 := SimpleGraph.completeMultipartiteGraph fun i : Fin 3 => Fin (![1, 3, 3] i)
 
--- The headline keeps the question form; the answer is recorded in
--- `erdos_1007.variants.dimension_four` below.
-set_option linter.style.category_answer false in
 /--
 The dimension of a graph $G$ is the minimal $n$ such that $G$ can be embedded in $\mathbb{R}^n$
 such that every edge of $G$ is a unit line segment.
 
 What is the smallest number of edges in a graph with dimension $4$?
 
-The smallest number of edges is $9$, achieved solely by $K_{3,3}$, proved by House [Ho13]. An
+Answer: The smallest number of edges is $9$, achieved solely by $K_{3,3}$, proved by House [Ho13]. An
 alternative proof was given by Chaffee and Noble [ChNo16], who also prove that the smallest
 number of edges in a graph of dimension $5$ is $15$ (achieved by $K_6$ and $K_{1,3,3}$).
 -/
-@[category research solved, AMS 5 52]
-theorem erdos_1007 :
-    IsLeast {m | ∃ (n : ℕ) (G : SimpleGraph (Fin n)), G.HasDimension 4 ∧ G.edgeSet.ncard = m}
-      answer(sorry) := by
-  sorry
-
-/--
-The smallest number of edges in a graph of dimension $4$ is $9$.
--/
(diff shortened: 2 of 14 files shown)


FormalConjectures/ErdosProblems/1007.lean
mo271:
```suggestion
/--
The dimension of a graph $G$ is the minimal $n$ such that $G$ can be embedded in $\mathbb{R}^n$
such that every edge of $G$ is a unit line segment.

What is the smallest number of edges in a graph with dimension $4$?

Answer: The smallest number of edges is $9$, achieved solely by $K_{3,3}$, proved by House [Ho13]. An
alternative proof was given by Chaffee and Noble [ChNo16], who also prove that the smallest
number of edges in a graph of dimension $5$ is $15$ (achieved by $K_6$ and $K_{1,3,3}$).
-/
@[category research solved, AMS 5 52, formal_proof using lean4 at "https://github.com/plby/lean-proofs/blob/main/src/v4.29.1/ErdosProblems/Erdos1007.lean"]
theorem erdos_1007 :
IsLeast {m | ∃ (n : ℕ) (G : SimpleGraph (Fin n)), G.HasDimension 4 ∧ G.edgeSet.ncard = m}
9 := by
sorry
```

Let's only have this question once

FormalConjectures/ErdosProblems/1026.lean
mo271:
```suggestion
/--
Let $x_1,\ldots,x_n$ be a sequence of distinct real numbers. Determine
$$
\max\left(\sum x_{i_r}\right),
$$
where the maximum is taken over all monotonic subsequences.

This is as Erdős posed the problem in [Er71], which is rather ambiguous. Discussion between
several users in the comments section has led to the following precise possible question, as
posed by van Doorn:

What is the largest constant $c$ such that, for all sequences of $n$ real numbers
$x_1,\ldots,x_n$,
$$
\max\left(\sum x_{i_r}\right) > (c-o(1))\frac{1}{\sqrt{n}}\sum x_i
$$
(where again the maximum is taken over all monotonic subsequences)?

Cambie makes the stronger conjecture that if $x_1,\ldots,x_{k^2}$ are distinct positive real
numbers with $\sum x_i=1$ then there is always a monotonic subsequence with sum at least $1/k$.

This stronger conjecture appears to have been first proved by Tidor, Wang, and Yang [TWY16],
and is also implicit in work of Wagner [Wa17]. A proof was given and formalised by Aristotle
(see the comments), with an alternative proof provided by Chan. In particular, this shows that
$c=1$.
-/
@[category research solved, AMS 5]
theorem erdos_1026.variants : IsGreatest admissibleConstants 1 := by
sorry
```

Let's only record this question once

FormalConjectures/ErdosProblems/416.lean
```diff
@@ -69,6 +69,8 @@ theorem erdos_416.variants.Erdos : ∃ f : ℝ → ℝ, f =o[atTop] (1 : ℝ →
     ∀ᶠ x in Filter.atTop, V x = x * x.log ^ (-1 + f x) := by
   sorry
 
+-- The constant `C` is not known explicitly, so the answer stays a placeholder.
```
mo271:
```suggestion
-- TODO: define the constant from the paper, 0.8178146... here and add it in the statement
```

FormalConjectures/ErdosProblems/416.lean
```diff
@@ -81,6 +83,8 @@ theorem erdos_416.variants.Maier_Pomerance :
       ∀ᶠ x in Filter.atTop, (V x : ℝ) = x / x.log * (rexp <| (C + f x) * x.log.log.log ^ 2) := by
   sorry
 
+-- The constants `C₁, C₂, C₃` are not known explicitly, so the answer stays a placeholder.
+set_option linter.style.category_answer false in
```
mo271:
```suggestion
-- TODO: add the constants `C₁, C₂, C₃` from the paper here
set_option linter.style.category_answer false in
```

FormalConjecturesTest/Util/Linters/CategoryAnswerLinter.lean
```diff
@@ -0,0 +1,82 @@
...
+@[category research open]
+theorem not_flagged_open_answer_sorry : answer(sorry) ↔ 1 + 1 = 2 := by
+  sorry
+
+set_option linter.style.category_answer false in
+#guard_msgs in
+/-- A deliberate placeholder can opt out of the linter. -/
+@[category research solved]
+theorem not_flagged_opt_out : answer(sorry) ↔ 1 + 1 = 2 := by
+  sorry
```
mo271:
```suggestion
```

I think this is a little too much testing...

FormalConjecturesTest/Util/Linters/CategoryAnswerLinter.lean
```diff
@@ -0,0 +1,82 @@
...
+
+/--
+warning: Declarations tagged `@[category research solved]` should not use `answer(sorry)`: the answer should be filled in explicitly, e.g. `answer(True)`.
+
+Note: This linter can be disabled with `set_option linter.style.category_answer false`
+-/
+#guard_msgs in
+/-- The `sorry` is still a placeholder when it carries a type ascription. -/
+@[category research solved]
+theorem flagged_ascribed_answer_sorry : answer((sorry : Nat)) = 2 := by
```
mo271, line 47:
```suggestion
theorem flagged_ascribed_answer_sorry : answer(sorry : Nat) = 2 := by
```

CONTRIBUTING.md
```diff
@@ -335,6 +335,13 @@ is outside of the scope of this repository.
   problem has been solved, `answer(sorry)` should be replaced by
   `answer(True)` or `answer(False)` accordingly.
 
+  The `linter.style.category_answer` linter enforces this: it warns when a
+  declaration tagged `@[category research solved]` still uses `answer(sorry)`.
+  A statement that deliberately keeps the placeholder, for instance because the
+  problem is independent of ZFC or because the answer is a constant that is not
+  known explicitly, can opt out with
+  `set_option linter.style.category_answer false in`.
+
```
mo271:
```suggestion
```

Let's not mention this here, this file is long enough and I don't think we (should) mention any other linters

mo271 (commented):
LGTM, thanks! 

left some comments

FormalConjectures/ErdosProblems/1007.lean
theebayuser: sure

FormalConjecturesTest/Util/Linters/CategoryAnswerLinter.lean
theebayuser: removed, thanks

mo271 (approved):

PR #5078, opened 2026-08-20
Mark OEIS A078590 as solved

KitaKen1:
This PR changes `OeisA78590.conjecture` from a positive `research open` proposition to an `answer(False)` statement and links a kernel-checked Lean counterexample.

The Lean formalization was prepared by [KitaKen1 (Kenta Kitamura)](https://github.com/KitaKen1).

The external proof establishes the exact corrected target theorem:

```lean
theorem OeisA78590Counterexample.oeis_A078590_solved :
    answer(False) ↔ ∀ n : ℕ, 3 ≤ n →
      OeisA78590.a (n - 2) ∣ 2 ^ OeisA78590.a (n - 1) + 1
```

Proof: https://github.com/KitaKen1/oeis-a078590-counterexample/blob/3bdf1cee8e7e4cbc4fe83354e4be8ca1eee46de3/lean/OeisA78590CounterexampleFC.lean#L55-L61

Repository: https://github.com/KitaKen1/oeis-a078590-counterexample

Lean4Web: https://live.lean-lang.org/#url=https%3A%2F%2Fraw.githubusercontent.com%2FKitaKen1%2Foeis-a078590-counterexample%2Frefs%2Fheads%2Fmain%2Flean4web%2FOeisA78590CounterexampleLean4Web.lean

`#print axioms` reports only `propext`, `Classical.choice`, and `Quot.sound`; there is no `sorryAx`.

AI Usage Disclosure: This formalization was developed with assistance from OpenAI Codex.

### DIFF
--- a/FormalConjectures/OEIS/78590.lean
+++ b/FormalConjectures/OEIS/78590.lean
@@ -59,9 +59,17 @@ theorem a_4 : a 4 = 9 := by
 theorem a_5 : a 5 = 171 := by
   rfl
 
-/-- Are all terms integers? -/
-@[category research open, AMS 11]
-theorem conjecture (n : ℕ) (hn : 3 ≤ n) : a (n - 2) ∣ 2 ^ a (n - 1) + 1 := by
+/--
+Are all terms integers?
+
+No. Exact divisibility first fails at $n = 7$; the corresponding rational term has reduced
+denominator $19$.
+-/
+@[category research solved, AMS 11,
+  formal_proof using lean4 at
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

Since the proof is so short, can we potentially inline it?!

KitaKen1:
@mo271 
Thank you for the suggestion.
Would it be acceptable to retain the current commit-pinned `formal_proof` link? 

Although the final theorem in the linked file is only seven lines, it depends on about 40 lines of supporting modular-arithmetic lemmas. 
I tried inlining the complete proof into the `by` block, but Lean reported `(kernel) deep recursion detected` even after increasing `maxRecDepth` to 20,000. 

Given this, I thought retaining the proof link would keep the change focused. However, I would be happy to revise it if you would prefer another approach.

mo271 (approved):
Thanks, LGTM

PR #5084, opened 2026-08-21
fix: no need to `update verso`

mo271:


YaelDillies (commented):
Why do we update anything in this action? Isn't `lake update formal_conjectures` basically guaranteed to break the build?

mo271:
right, should already be in the lake-manifest.json

YaelDillies (approved):

site/README.md
```diff
@@ -107,6 +107,7 @@ lake exe extract_names --exclude=statement,docstring,moduleDocstrings,fileFirstA
 # Warning: the literate build step can take a long time (30+ minutes).
 cd docbuild
 export MATHLIB_NO_CACHE_ON_UPDATE=1
+lake update verso
 lake update formal_conjectures
```
mo271:
```suggestion
```

PR #5086, opened 2026-08-21
feat: enable cache in ci

mo271:


YaelDillies (approved):
Looks reasonable to me!

.github/workflows/build-and-docs.yml
```diff
@@ -130,6 +130,22 @@ jobs:
...
           lake exe cache get
 
+      - name: Restore local lake build
+        if: steps.mode.outputs.website_only != 'true'
+        uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: .lake/build
+          key: lake-build-${{ runner.os }}-${{ hashFiles('lean-toolchain', 'lake-manifest.json') }}-${{ github.sha }}
+          restore-keys: |
+            lake-build-${{ runner.os }}-${{ hashFiles('lean-toolchain', 'lake-manifest.json') }}-
```
YaelDillies, line 140: Are you sure this `restore-keys` matches keys of the form `lake-build-${{ runner.os }}-${{ hashFiles('lean-toolchain', 'lake-manifest.json') }}-${{ github.sha }}`? Is the match prefix-basic?

.github/workflows/build-and-docs.yml
```diff
@@ -130,6 +130,22 @@ jobs:
...
+
+      - name: Save local lake build
```
YaelDillies, line 142: Why saving just after restoring? Surely this does not achieve anything?

.github/workflows/build-and-docs.yml
```diff
@@ -130,6 +130,22 @@ jobs:
...
+        if: steps.mode.outputs.website_only != 'true' && github.event_name == 'push'
+        uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: .lake/build
```
YaelDillies, line 146: This path doesn't exist yet, so saving currently fails in CI: https://github.com/google-deepmind/formal-conjectures/actions/runs/32578602162/job/97044509497

You probably should add a step to create the folder if it doesn't exist yet, or maybe you don't because the folder will exist if you move this step down as suggested in my above comment

.github/workflows/build-and-docs.yml
```diff
@@ -192,7 +208,7 @@ jobs:
         if: steps.mode.outputs.website_only != 'true' && github.event_name == 'push'
         uses: actions/cache/save@6f8efc29b200d32929f49075959781ed54ec270c # v3
         with:
-          path: ~/.cache
+          path: ~/.cache/mathlib
           key: oleans-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.lean') }}
```
YaelDillies, line 212: Isn't this key too specific? Surely we want to download the cache if eg one Lean file changed, because `lake build` will figure it by itself that it needs to rebuild that one file. Right now, changing any Lean file changes the key

YaelDillies:
Sorry, I wanted to get this merged quickly to see if it helped with my long string of PRs, but of course it didn't because I didn't properly review it...

PR #5106, opened 2026-08-23
fix(Wikipedia): correct cuboidOne docstring — the cited artifact is a proof, not a disproof

dawsonamf:
The docstring on `cuboidOne` reads "The DeepMind prover agent has found a formal **disproof** of this statement."

The `formal_proof` attribute on the same declaration points at [EulerBrick.lean@34c93bba#L1804](https://github.com/google-deepmind/formal-conjectures/blob/34c93bbad127a9a5354b9d53478d338eb65edb88/FormalConjectures/Wikipedia/EulerBrick.lean#L1804), which proves `CuboidOne` affirmatively (`theorem cuboidOne : CuboidOne`, discharged via `cuboid_poly_irreducible_Z`), and the independent reference cited in the same docstring (Asiryan, [arxiv/2510.11768](https://arxiv.org/abs/2510.11768)) likewise **proves** irreducibility. The declaration itself asserts `CuboidOne` and is tagged `research solved`.

So the docstring's "disproof" contradicts both artifacts it describes. This PR changes the one word: "disproof" → "proof".

(Left as is: the surrounding "informal solution" phrasing, which is direction-neutral.)


dawsonamf:
CLA signed.

mo271 (commented):
Thanks, good catch

mo271 (approved):

PR #5108, opened 2026-08-24
fix(ErdosProblems/394): correct lower-bound direction

### LINKED ISSUE #5062: Erdős Problem 394: `lower_bound` has the sides of `≫` swapped (states an upper bound instead of the lower bound)
In `erdos_394.variants.lower_bound` the two sides of `≫` are swapped relative to the docstring.

Because of this, the file's open `hall_conjecture` would follow from this `research solved` statement (see below).

https://github.com/google-deepmind/formal-conjectures/blob/f40ecd9b3b79a7aa1a64b7d2f3b2623d4a6c3081/FormalConjectures/ErdosProblems/394.lean#L106-L113

The docstring says the sum dominates (a trivial **lower** bound):

> Since $t_2(p)=p-1$ for prime $p$ it is trivial that $\sum_{n\leq x}t_2(n)\gg \frac{x^2}{\log x}$.

but the code puts the sum on the other side:

```lean
theorem erdos_394.variants.lower_bound :
    (fun x ↦ x ^ 2 / Real.log x) ≫
    (fun x ↦ ∑ n ∈ Icc 1 ⌊x⌋₊, (t 2 n : ℝ)) := by
  sorry
```

[`f ≫ g` is notation for `g =O[atTop] f`](https://github.com/google-deepmind/formal-conjectures/blob/33c6a2dcafc5e3cdebfc083203b0a051309ea5b4/FormalConjecturesForMathlib/Analysis/Asymptotics/Basic.lean#L24-L25).

So as written this states the **upper** bound $\sum_{n\le x} t_2(n) = O(x^2/\log x)$ — the reverse of the docstring — while being marked `research solved`.

### Trivial proof/disproof
The swap is visible by comparing the docstring with the code above.

And since $\log 2 < 1$, an $O(x^2/\log x)$ upper bound implies $o(x^2/(\log x)^c)$ for every $c < \log 2$ — which is exactly the statement of the open `hall_conjecture`.

### Suggested fix
Swap the arguments back:

```lean
theorem erdos_394.variants.lower_bound :
    (fun x ↦ ∑ n ∈ Icc 1 ⌊x⌋₊, (t 2 n : ℝ)) ≫
    (fun x ↦ x ^ 2 / Real.log x) := by
  sorry
```

No `@[category ...]` annotation needs to change:

- `lower_bound` stays `research solved` — after the swap it is the trivial known lower bound described in the docstring.
- `hall_conjecture` stays `research open` — the fix is what makes that label meaningful again (currently it could be "closed" as a corollary of a `research solved` statement, without any mathematics).
- No other declaration in the file is affected.

### Additional context
This is only a statement correction — the Erdős–Hall conjecture (`hall_conjecture`) itself remains open. Related status-sync issue: #4803.

### Choose either option

linrock:
Fix the direction of the lower bound to match its docstring.

Resolves https://github.com/google-deepmind/formal-conjectures/issues/5062

### DIFF
--- a/FormalConjectures/ErdosProblems/394.lean
+++ b/FormalConjectures/ErdosProblems/394.lean
@@ -107,7 +107,7 @@ Since $t_2(p)=p-1$ for prime $p$ it is trivial that $\sum_{n\leq x}t_2(n)\gg \fr
 -/
 @[category research solved, AMS 11]
 theorem erdos_394.variants.lower_bound :
-    (fun x ↦ x ^ 2 / Real.log x) ≫
+    (fun x ↦ x ^ 2 / Real.log x) ≪
     (fun x ↦ ∑ n ∈ Icc 1 ⌊x⌋₊, (t 2 n : ℝ)) := by
   sorry
 


mo271 (approved):
Thanks!

PR #5112, opened 2026-08-24
 ci: remove error suppression from verso build steps

mo271:


mo271:
This currently only passes with the cherry-pick fix for this issue https://github.com/leanprover/verso/issues/968

Contrary to before we now want this to fail loudly when verso can't build a file in the future

YaelDillies (approved):

PR #5016, opened 2026-08-16
Add more conjectures from AutoOeis to formal-conjectures

codrut3:
Please review in particular:
- 145062.lean and 181830.lean, which define an axiom for the formalisation;
- 145355.lean: the conjecture is not fully explained on the OEIS page.

I included 5153.lean and 3625.lean since these sequences are used in the definition of other sequences.

PR #5114, opened 2026-08-24
Mention recently solved conjectures.

codrut3:
This is a follow-up to PR #5016


FormalConjectures/OEIS/129365.lean
```diff
@@ -59,38 +59,53 @@ theorem a_4 : a 4 = 1 := by decide +native
 
 /--
 Conjecture (1): $a(n)$ is always an integer (the denominator divides the numerator).
+
+Answer: true, see linked proof.
```
mo271, line 63: potentially replace with proof sketch (same for the other 3 here)

mo271 (approved):
Thanks!

Issue #5113, opened 2026-08-24
investigate caching


mo271:
- Check if we can get `lake cache` working in our repo

- Investigate caching to `verso`

tadamcz:
Should this still be open or closed @mo271 ?

Paul-Lez:
I think this is now solved, but let's wait for @mo271 to confirm

mo271:
I think the caching is working fine for now. 
One thing that we don't have yet, and I thought would be useful at some point is making `lake exe cache get` work, but @eric-wieser said there might be alternatives

eric-wieser:
(the alternative is `lake cache`)

mo271:
also caching for our `verso` build shoul be investigated

PR #5155, opened 2026-08-25
fix(cache): Fix CI build caching and remove redundant All.lean step

danielchin:
#5113 

1) Move `.lake/build` cache after save to run after build
2) Remove `hashFiles('**/*.lean')` from mathlib cache. Mathlib oleans only depend on `lean-toolchain` and `lake-manifest.json`
3) Remove `All.lean` generation. `globs = ["FormalConjectures.+"]` in `lakefile.toml`.

Did some testing locally, editing one file takes less than a minute to build now.

### DIFF
--- a/.github/workflows/build-and-docs.yml
+++ b/.github/workflows/build-and-docs.yml
@@ -146,6 +146,16 @@ jobs:
         run: |
           ./scripts/mk_all_formathlib.sh
 
+      # Generating All.lean aggregates all problem files into a single module to detect
+      # clashing declaration names across problem modules during compilation.
+      - name: Generate All.lean
+        if: steps.mode.outputs.website_only != 'true'
+        run: |
+          rm -f FormalConjectures/All.lean
+          lake exe mk_all --lib FormalConjectures || true
+          mv FormalConjectures.lean FormalConjectures/All.lean
+          echo "set_option linter.style.moduleDocstring false" >> FormalConjectures/All.lean
+
       - name: Begin Lean problem matcher
         if: steps.mode.outputs.website_only != 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-docs.yml
```diff
@@ -142,26 +141,11 @@ jobs:
...
         run: |
           ./scripts/mk_all_formathlib.sh
 
-      - name: Generate All.lean
-        if: steps.mode.outputs.website_only != 'true'
-        run: |
-          rm -f FormalConjectures/All.lean
-          lake exe mk_all --lib FormalConjectures || true
-          mv FormalConjectures.lean FormalConjectures/All.lean
-          echo "set_option linter.style.moduleDocstring false" >> FormalConjectures/All.lean
```
mo271, line 163: I think the reason I kept this is because it will detect clashing declaration names. 
It should not take too long, perhaps we can keep it for now

.github/workflows/build-and-docs.yml
```diff
@@ -179,7 +163,6 @@ jobs:
         if: steps.mode.outputs.website_only != 'true'
         run: |
           python3 scripts/lake-build-wrapper.py /tmp/build_summary_problems.json lake --wfail build
-          rm -f FormalConjectures/All.lean
```
mo271, line 182: ... then keep this as well

mo271 (approved):
the caching changes LGTM, thanks!
Not so sure about removing All.lean, perhaps good insteead add a comment explaining why the step is there?

.github/workflows/build-and-docs.yml
danielchin: Added back and left a comment.

.github/workflows/build-and-docs.yml
danielchin: Added back.

PR #5170, opened 2026-08-25
feat(Wikipedia): Square root sum problem

BoltonBailey:
This PR adds a file for the open problem regarding the polynomial time computability of the square root sum problem. 

This follows the work in #4796 adding a predicate for polynomial time computation and @Paul-Lez 's comment [here](https://github.com/google-deepmind/formal-conjectures/issues/4414#issuecomment-5207994025) that we should add new complexity conjectures as their own PRs.

Co-authored-by: [Paramveersingh-S](https://github.com/Paramveersingh-S)


BoltonBailey:
I guess this series of PRs will also subsume #4416, so because my formalization here is similar to that one I will put the author of that PR as a coauthor here.

PR #5180, opened 2026-08-27
proposal: add Tau Ceti as a dependency

Paul-Lez:
This PR proposes adding Tau-Ceti as a dependency to the formal-conjectures repo.
I'd be curious for feedback from other contributors and maintainers! The goal here is to be able to formalise many more conjectures, using the additional formalised material contained in Tau-Ceti. 
At this point in time, it may be useful for us to be able to distinguish between the conjectures that depend only on Mathlib + FormalConjecturesForMathlib and those that only depend on Tau-Ceti, so it may be worth thinking about what's the best way to do so!


mo271:
One goal of Formal Conjectures has been to have formalisations of the informal statements of open problems that are fairly easy to check, i.e. it should be easy to convince ourselves that a formalisation is faithful to the informal statement. This is already quite hard, given how easy it is do to misformalisation. Depending on TauCeti might make this more difficult, because one would need to check also potentially the definitions in TauCeti, right?
(I just assume that definitions in Mathlib can be trusted to be a faithful translation of informal to formal statements)

However, I completely agree that it would unlock many statements, which would be great. 

If we decide to add it as a dependency, I would like to make sure it is always clear to distinguish those problems that use TauCeti from those which don't. (Perhaps easy by not making it a default import). 
Of course the next question might then also be if we allow a `FormalConjecturesForTauCeti` directory.

In any case, I would appreciate @eric-wieser's thoughts here. (We can also have the discussion over at Zulip..)

Paul-Lez:
Re trusting Tau-Ceti, I agree that this expands a lot the trust boundary! On the other hand, there is precedent: the Palomar registry treats Tau-Ceti as one of the trusted import libraries that are allowed by comparator.

If we were to go forward with this, I think one reasonable thing could be to have a `FormalConjecturesWithTauCeti` to distinguish conjectures that depend on Tau-Ceti, and enforce that all conjectures importing Tau Ceti go in that directory.

PR #5181, opened 2026-08-27
Add the bunkbed conjecture (disproved, Gladkov–Pak–Zimin 2024)

henrykmichalewski:
Adds `FormalConjectures/Wikipedia/BunkbedConjecture.lean` — Kasteleyn's bunkbed conjecture in Bernoulli bond percolation, recorded as **disproved** (`answer(False)`) following Gladkov, Pak and Zimin, *The bunkbed conjecture is false* ([arXiv:2410.02545](https://arxiv.org/abs/2410.02545)), whose 7222-vertex counterexample at $p = \tfrac12$ builds on Hollom's hypergraph counterexample ([arXiv:2401.07301](https://arxiv.org/abs/2401.07301)).

The formalisation is fully combinatorial (no measure theory): the connection probability is the finite sum $\sum_{F \subseteq E} p^{|F|}(1-p)^{|E|-|F|}\,[u \leftrightarrow_F v]$ over edge subsets, with reachability in `fromEdgeSet F`.

| Declaration | Content | Status |
|---|---|---|
| `bunkbed G T`, `connectionProbability H p u v` | bunkbed graph with posts `T`; percolation connection probability | definitions |
| `bunkbed_conjecture` | $\mathbb P((u,0)\leftrightarrow(v,1)) \le \mathbb P((u,0)\leftrightarrow(v,0))$ for all $G, T, p, u, v$ | `answer(False)`, `research solved` |
| `.variants.complete_graph` | holds for complete graphs (Richthammer 2022; de Buyer 2018 at $p=\tfrac12$) | `sorry` + references |
| `.variants.p_near_one` | holds for $p$ close to 1 (Hutchcroft–Kent–Nizić-Nikolac 2023) | `sorry` + reference |
| `sum_weights_eq_one`, `connectionProbability_le_one`, `connectionProbability_self` | basic API | **proved** |
| `.variants.diagonal` | the $u = v$ case of the inequality | **proved** |

### Verification
`lake build FormalConjectures.Wikipedia.BunkbedConjecture` on v4.33.1: clean, all linters pass.

### DIFF
--- a/FormalConjectures/Wikipedia/BunkbedConjecture.lean
+++ b/FormalConjectures/Wikipedia/BunkbedConjecture.lean
@@ -160,7 +160,7 @@ lemma connectionProbability_self {W : Type*} [Fintype W] (H : SimpleGraph W) (p 
 When `u = v` the right-hand side is the probability that `(u, 0)` is connected to itself, which
 is `1`, so the inequality is just `connectionProbability_le_one`.
 -/
-@[category research solved, AMS 5 60]
+@[category test, AMS 5 60]
 theorem bunkbed_conjecture.variants.diagonal {V : Type} [Fintype V] [DecidableEq V]
     (G : SimpleGraph V) [DecidableRel G.Adj] (T : Finset V) (p : ℝ)
     (hp₀ : 0 ≤ p) (hp₁ : p ≤ 1) (u : V) :
--- a/FormalConjectures/Wikipedia/BunkbedConjecture.lean
+++ b/FormalConjectures/Wikipedia/BunkbedConjecture.lean
@@ -1,0 +1,172 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# The bunkbed conjecture
+
+*References:*
+* [Wikipedia](https://en.wikipedia.org/wiki/Bunkbed_conjecture)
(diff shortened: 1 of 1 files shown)


j2d9w5xtjn-png:
The statement here percolates the posts, but in the bunkbed model the posts are always retained.

`connectionProbability (bunkbed G T) p u v` runs Bernoulli percolation over all of `(bunkbed G T).edgeFinset`, and by the definition of `bunkbed` that edge set includes the post edges. Gladkov–Pak–Zimin set it up the other way round (§1, just before Conjecture 1.1):

> In the bunkbed percolation, the usual bond percolation is performed only on edges in $G$ and $G'$, while all edges in $\overline{T}$ are retained (i.e., not deleted).

So `bunkbed_conjecture` is a different random model from Conjecture 1.1, and Theorem 1.2 is not directly a counterexample to it. Conditioning on the set $S \subseteq T$ of open posts writes the randomized-post probability as an average
$$\mathbb{P}^{\text{rand}}_{p,T} = \sum_{S \subseteq T} p^{|S|}(1-p)^{|T|-|S|}\,\mathbb{P}^{\text{bb}}_{p,S},$$
and for the GPZ graph ($|T| = 3$) every $S \subsetneq T$ term is $\ge 0$ — BBC is known for at most two transversal vertices, and the $S = \emptyset$ term is strictly positive — so the sign of the average does not follow from Theorem 1.2.

I think `answer(False)` is still correct, but only through an extra step rather than the cited theorem: attach to each $w \in T$ new vertices $w_1,\dots,w_k$ adjacent only to $w$, and take $\{w_i\}$ as the transversal set. Each $i$ gives an independent open crossing $w - w_i - w_i' - w'$ of probability $p^3$, so the post at $w$ is effectively open with probability $1 - (1-p^3)^k \to 1$, while the pendants add no other connectivity between vertices of $G$. Letting $k \to \infty$ recovers the retained-post model, and GPZ applies. Worth stating that reduction explicitly if you want to keep the current definition.

The cleanest fix is probably to let `connectionProbability` take per-edge probabilities and set them to `1` on posts, or to percolate only $E \cup E'$ and keep $\overline{T}$ intact.

The two variants inherit the same mismatch. Both happen to remain true as stated (each is an average over $S \subseteq T$ of cases where the retained-post result is known), but they are then no longer literally [Ri22] / [HKN23].

One smaller thing: `variants.complete_graph` is about complete graphs, but [Ri22] (arXiv:2204.12931) is the complete *bipartite* case — GPZ cite [dB16, dB18, HL19] for complete graphs. Might be worth double-checking that attribution.

PR #5183, opened 2026-08-27
feat: P vs. BPP

### LINKED ISSUE #5175: The P vs BPP question
### What is the conjecture

The P vs. BPP problem/conjecture asks/states that the complexity class [BPP](https://en.wikipedia.org/wiki/BPP_(complexity)) (the set of decision problems decidable in polynomial time with bounded error) is equal to P.

### Prerequisites needed

We've already formalized some complexity theoretic concepts in service of P vs. NP, so there is not much more that needs to be done to add this conjecture besides defining BPP itself. I have a branch [here](https://en.wikipedia.org/wiki/BPP_(complexity)) which indicates that this can be done in only a few lines.

* ams-68

### Choose either option

BoltonBailey:
Adds the P vs. BPP problem from computational complexity theory.

Closes #5175

### DIFF
--- a/FormalConjecturesForMathlib/Computability/Complexity.lean
+++ b/FormalConjecturesForMathlib/Computability/Complexity.lean
@@ -103,16 +103,16 @@ such that there exists a polynomial `p` over ℕ and a poly-time Turing machine 
 where for all `x`, a uniformly random string `r` of length `p (|x|)` satisfies
 `R (x, r) = L x` with probability at least `2/3`.
 
-Here the probability is expressed by counting: at least `2/3` of the `2 ^ p (|x|)` random
-strings `r : Fin (p (|x|)) → Bool` must make `R` answer correctly.
+Here the probability is the fraction (as a rational number) of the `2 ^ p (|x|)` random
+strings `r : Fin (p (|x|)) → Bool` that make `R` answer correctly.
 
 See Definition 7.2 in Arora-Barak (2009).
 -/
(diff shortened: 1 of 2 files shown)


FormalConjecturesForMathlib/Computability/Complexity.lean
```diff
@@ -97,4 +97,22 @@ whose complements are in NP.
...
   { L | Lᶜ ∈ NP }
 
+/--
+The class BPP is the set of decision problems
+such that there exists a polynomial `p` over ℕ and a poly-time Turing machine
+where for all `x`, a uniformly random string `r` of length `p (|x|)` satisfies
+`R (x, r) = L x` with probability at least `2/3`.
+
+Here the probability is the fraction (as a rational number) of the `2 ^ p (|x|)` random
+strings `r : Fin (p (|x|)) → Bool` that make `R` answer correctly.
```
Paul-Lez: small nit: these days we prefer tex notation since that renders nicer on the website!

FormalConjecturesForMathlib/Computability/Complexity.lean
```diff
@@ -97,4 +97,22 @@ whose complements are in NP.
 def coNP : DecisionComplexityClass :=
   { L | Lᶜ ∈ NP }
 
+/--
+The class BPP is the set of decision problems
```
Paul-Lez: I think the docstring doesn't say what `L` is?

PR #5205, opened 2026-08-28
feat(Paper): add the Tu-Deng conjecture

yurekami:
Adds the Tu-Deng conjecture (Tu & Deng, Des. Codes Cryptogr. 60 (2011), 1-14; IACR ePrint 2009/272) as `FormalConjectures/Paper/TuDengConjecture.lean`.

**What:** For `k ≥ 2` and a nonzero residue `t` modulo `2^k - 1`, the number of pairs of residues `(a, b)` with `a + b = t` and `w(a) + w(b) ≤ k - 1` (binary weight of the representatives in `0, …, 2^k - 2`) is at most `2^(k-1)`. Stated over `ZMod (2^k - 1)` with `Set.ncard`; binary weight is `(Nat.digits 2 n).sum`.

**Why:** The conjecture underlies Tu and Deng's constructions of Boolean functions with optimal algebraic immunity and is open. It was not yet in the repository (no hits for Tu-Deng in code, issues, or PRs).

**Faithfulness:** Statement checked against the source (ePrint §3, "Conjecture"): `k > 1`, `a, b ∈ Z_{2^k−1}`, `0 < t < 2^k − 1`, `w(a) + w(b) ⩽ k − 1`, bound `2^{k−1}`. The formalized set semantics were additionally brute-force checked in Python for all `t` at `k = 2, …, 12` (no violations of the bound, and the equality cases match the known min-cyclic-gap characterization).

**How checked:** The statement was verified against the source text and brute-forced numerically as above. I could not complete a local `lake --wfail build` (mathlib cache download is impractically slow on my connection), so CI is the build check here; I will fix promptly if it flags anything.

### DIFF
--- a/FormalConjectures/Paper/TuDengConjecture.lean
+++ b/FormalConjectures/Paper/TuDengConjecture.lean
@@ -1,0 +1,56 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+import FormalConjecturesUtil
+
+/-!
+# Tu-Deng Conjecture
+
+*References:*
+* [Tu, Z., Deng, Y., *A conjecture about binary strings and its applications on constructing
+  Boolean functions with optimal algebraic immunity*, Des. Codes Cryptogr. **60** (2011),
+  1–14](https://doi.org/10.1007/s10623-010-9413-9)
+* [IACR ePrint 2009/272](https://eprint.iacr.org/2009/272)
+
+For an integer `k ≥ 2`, identify the residues modulo `2 ^ k - 1` with the integers
+`0, 1, …, 2 ^ k - 2`, and let `w a` denote the binary weight of `a` (the number of ones in
+its binary expansion). The Tu-Deng conjecture states that for every residue `t ≠ 0`,
+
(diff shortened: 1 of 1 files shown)


yurekami:
Note on the red `zizmor-output` check: it scanned this branch's checked-out workflow files, which come from my fork's stale January main (my token lacks the `workflow` scope, so I cannot push a branch containing the newer upstream workflow commits or sync the fork). The findings are all against those old workflow files, none against this change, and `Build project` / `Test scripts` / `check-copyright` pass. Once I restore the scope I will rebase onto current `main`, which should return zizmor to its usual `skipping` state — or I'm happy to do whatever is easier for you.

FormalConjectures/Paper/TuDengConjecture.lean
```diff
@@ -0,0 +1,56 @@
...
+
+namespace TuDengConjecture
+
+/-- The binary weight of a natural number: the number of ones in its binary expansion. -/
+def binaryWeight (n : ℕ) : ℕ := (Nat.digits 2 n).sum
+
+/--
+**The Tu-Deng conjecture.** For `k ≥ 2` and a nonzero residue `t` modulo `2 ^ k - 1`, there
+are at most `2 ^ (k - 1)` pairs of residues `(a, b)` with `a + b = t` whose binary weights
+(of their representatives in `0, …, 2 ^ k - 2`) sum to at most `k - 1`.
```
mo271: can you use latex markdown in the docstring here (and above)

mo271 (approved):
Thanks, LGTM mathematically, just one comment  about docstrings

FormalConjectures/Paper/TuDengConjecture.lean
yurekami: Done in 277601b — both docstrings (module doc and theorem statement) now use inline LaTeX, matching the `$k \le 29$` style used elsewhere in the repo. Thanks for the review!

yurekami:
Rebased onto current main (the fork is synced now), so the branch no longer carries the stale January workflow files — zizmor should report normally on this run. Diff content is unchanged: the single TuDengConjecture.lean file with the LaTeX docstrings.

PR #5206, opened 2026-08-28
feat(paper): Least squares matching

### LINKED ISSUE #5179: The number of optimal least-squares matchings between two sets of n points in the plane under rotation
### What is the conjecture
Consider two lists of points $A=(a_1,\ldots,a_n)$ and $B=(b_1,\ldots,b_n)$, $a_i,b_j\in \mathbb{R}^2$ of length $n$. For any congruent copy $B'$ of $B$. We can look for the least-squares matching between $A$ and $B'$: The permutation $\pi\in S_n$ that minimizes $\sum_{i=1}^n \lVert a_i-b'_{\pi_i}\rVert^2.$
We consider the set $S_{\mathrm{opt}} = \{ \pi\in S_n \mid \exists B'\colon \pi \text{the unique optimal matching between } A and B'\}$.
The conjecture is that the number of permutations in $S_{opt}$ is polynomially bounded in $n$.

We need to count only unique optimal permutations, because there can be degenerate situations with an exponential number of optimal permutations. If the conjecture is true, it implies a polynomial-time algorithm for least-squares matching under rotations.

The optimum permutation is unaffected by positive scaling of $A$ or of $B'$, as well as translation of one of the sets.

References with some weaker results on a related problem (which I might also propose):
-  Günter Rote: Partial least-squares point matching under translations. In: 26th European Workshop on Computational Geometry (EuroCG'10), Dortmund, March 2010, pp. 249–251, Editor: Jan Vahrenhold. https://page.mi.fu-berlin.de/rote/Papers/abstract/Partial+least-squares+point+matching+under+translations.html
- Rinat Ben-Avraham, Matthias Henze, Rafel Jaume, Balázs Keszegh, Orit E. Raz, Micha Sharir, Igor Tubis. Partial-Matching RMS Distance Under Translation: Combinatorics and Algorithms. Algorithmica 80, 2400–2421 (2018). https://doi.org/10.1007/s00453-017-0326-0

### Prerequisites needed
Euclidean geometry, finite point sets in the plane, distances

### [AMS categories]

* ams-52

### Choose either option
- [X] I plan on adding this conjecture to the repository

guenterrote:
fixes #5179

2 conjectures stated

### DIFF
--- a/FormalConjectures/Paper/LeastSquaresPointMatching.lean
+++ b/FormalConjectures/Paper/LeastSquaresPointMatching.lean
@@ -26,7 +26,7 @@ Editor: Jan Vahrenhold.
 
 Rinat Ben-Avraham, Matthias Henze, Rafel Jaume, Balázs Keszegh, Orit E. Raz, Micha Sharir, Igor Tubis.
 Partial-Matching RMS Distance Under Translation: Combinatorics and Algorithms. Algorithmica 80, 2400–2421 (2018).
-`https://doi.org/10.1007/s00453-017-0326-0
+`https://doi.org/10.1007/s00453-017-0326-0`
 
 -/
 namespace LeastSquaresPointMatching
--- a/FormalConjectures/Paper/LeastSquaresPointMatching.lean
+++ b/FormalConjectures/Paper/LeastSquaresPointMatching.lean
@@ -1,0 +1,70 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# The number of least-squares matchings under rotations is polynomially bounded
+
+*References:* Günter Rote: Partial least-squares point matching under translations.
+In: 26th European Workshop on Computational Geometry (EuroCG'10), Dortmund, March 2010, pp. 249–251,
+Editor: Jan Vahrenhold.
+`https://page.mi.fu-berlin.de/rote/Papers/abstract/Partial+least-squares+point+matching+under+translations.html`
+
+Rinat Ben-Avraham, Matthias Henze, Rafel Jaume, Balázs Keszegh, Orit E. Raz, Micha Sharir, Igor Tubis.
+Partial-Matching RMS Distance Under Translation: Combinatorics and Algorithms. Algorithmica 80, 2400–2421 (2018).
+`https://doi.org/10.1007/s00453-017-0326-0
+
+-/
+namespace LeastSquaresPointMatching
+
+/--
+Consider two lists of points $A=(a_1,\ldots,a_n)$ and $B=(b_1,\ldots,b_n)$, $a_i,b_j\in \mathbb{R}^2$ of length $n$.
+For any congruent copy $B'$ of $B$, we can look for the least-squares matching between $A$ and $B'$:
+The permutation $\pi\in S_n$ that minimizes $\sum_{i=1}^n \lVert a_i-b'_{\pi_i}\rVert^2.$
+We consider the set
+$S_{\mathrm{opt}} = \{ \pi\in S_n \mid \exists B'\colon \pi \text{the unique optimal matching between $A$ and $B'$}\}$.
+The conjecture is that the number of permutations in $S_{opt}$ is bounded by $O(n^d)$ for some constant $d$.
+
+We need to count only unique optimal permutations, because there can be degenerate situations with an exponential number of optimal permutations. If the conjecture is true, it implies a polynomial-time algorithm for least-squares matching under rotations.
+
+Remark: The optimum permutation is unaffected by positive scaling of $A$ or of $B'$, as well as translation of one of the sets.
+
+There is a variation of the problem, where the set $A$ is smaller than $B$, and we look for the least-squares
+injective mapping from $A$ to $B$ (partial matching), but now we only consider tranlations.
+This is the problem considered in the two papers cited above. The conjectured bound is O(m²n²).
+-/
+
+def SumOfSquaredDistances {m n : ℕ} (A : Fin m → ℝ × ℝ) (B : Fin n → ℝ × ℝ) (π : Fin m ↪ Fin n) : ℝ :=
+  ∑ i, dist (A i) (B (π i)) ^ 2
+
+@[category research open, AMS 52]
+theorem least_squares_point_matching_under_rotations_is_polynomially_bounded (d : ℕ) :
+  ∃ c : ℕ, ∀ n : ℕ, ∀ A B : Fin n → ℝ × ℝ,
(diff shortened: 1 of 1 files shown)


FormalConjectures/Paper/LeastSquaresPointMatching.lean
```diff
@@ -0,0 +1,101 @@
...
+or of $B'$, as well as translation of one of the sets.
+Therefore, it is sufficient to consider only rotations around the origin in the conjecture.
+(One could also omit the `unit` condition in the definition of `Rotation_2`.)
+-/
+@[category research open, AMS 52]
+theorem least_squares_point_matching_under_rotations_is_polynomially_bounded (d : ℕ) :
+  ∃ c : ℕ, ∀ n : ℕ, ∀ A B : Fin n → ℝ × ℝ,
+  {π : Fin n ↪ Fin n | ∃ R : Rotation_2,
+    let B' := fun i => applyRotation_2 R (B i)
+    (∀ π' : Fin n ↪ Fin n, π' ≠ π → SumOfSquaredDistances A B' π' < SumOfSquaredDistances A B' π)}.ncard
```
mo271:
```suggestion
    (∀ π' : Fin n ↪ Fin n, π' ≠ π → SumOfSquaredDistances A B' π' > SumOfSquaredDistances A B' π)}.ncard
```

This should be the same here as in the Translation version, right?
Perhaps in both cases we could flip the inequalities to go for `<`

FormalConjectures/Paper/LeastSquaresPointMatching.lean
```diff
@@ -0,0 +1,101 @@
...
+Partial-Matching RMS Distance Under Translation: Combinatorics and Algorithms. Algorithmica 80, 2400–2421 (2018).
+`https://doi.org/10.1007/s00453-017-0326-0
+
+-/
+namespace LeastSquaresPointMatching
+
+variable {m n : ℕ}
+
+/-- The sum of squared distances between corresponding points under a mapping $\pi$. -/
+def SumOfSquaredDistances (A : Fin m → ℝ × ℝ) (B : Fin n → ℝ × ℝ) (π : Fin m ↪ Fin n) : ℝ :=
```
mo271:
```suggestion
def SumOfSquaredDistances (A : Fin m → ℝ²) (B : Fin n → ℝ²) (π : Fin m ↪ Fin n) : ℝ :=
```

Use this instead of ℝ × ℝ throughout

FormalConjectures/Paper/LeastSquaresPointMatching.lean
```diff
@@ -0,0 +1,101 @@
...
+  sorry
+-- The specific conjecture is that this is true for d1 = d2 = 2.
+
+
+/-- A rotation in Euclidean 2-space around the origin.
+It is specified by the pair (cos(θ), sin(θ)) for some angle θ. -/
+structure Rotation_2 where
+  cos_θ : ℝ
+  sin_θ : ℝ
+  unit : cos_θ ^ 2 + sin_θ ^ 2 = 1
```
mo271: use Orientation.rotation (θ : Real.Angle) : V ≃ₗᵢ[ℝ] V

FormalConjectures/Paper/LeastSquaresPointMatching.lean
```diff
@@ -0,0 +1,101 @@
...
+is unaffected by positive scaling of $A$
+or of $B'$, as well as translation of one of the sets.
+Therefore, it is sufficient to consider only rotations around the origin in the conjecture.
+(One could also omit the `unit` condition in the definition of `Rotation_2`.)
+-/
+@[category research open, AMS 52]
+theorem least_squares_point_matching_under_rotations_is_polynomially_bounded (d : ℕ) :
+  ∃ c : ℕ, ∀ n : ℕ, ∀ A B : Fin n → ℝ × ℝ,
+  {π : Fin n ↪ Fin n | ∃ R : Rotation_2,
+    let B' := fun i => applyRotation_2 R (B i)
```
mo271:
```suggestion
    let B' := fun i => EuclideanGeometry.o.rotation θ (B i)
```
(untested) should work when using  ℝ²?

FormalConjectures/Paper/LeastSquaresPointMatching.lean
```diff
@@ -0,0 +1,101 @@
...
+This would imply a polynomial-time algorithm for partial least-squares matching under translations.
+More specifically, the conjecture is that bound is O(m²n²).
+
+We must count only unique optimal permutations, because there can be degenerate situations with
+an exponential number of optimal permutations, for example, if $A$ lies on the $x$-axis and $B$ on the $y$-axis.
+If the conjecture is true, it implies a polynomial-time algorithm for least-squares matching under rotations.
+-/
+@[category research open, AMS 52]
+theorem least_squares_partial_point_matching_under_translations_is_polynomially_bounded (d1 d2 : ℕ) :
+  ∃ c : ℕ, ∀ m n : ℕ, (m < n) → ∀ A : Fin m → ℝ × ℝ, ∀ B : Fin n → ℝ × ℝ,
```
mo271:
```suggestion
  ∃ c : ℕ, ∀ m n : ℕ, (m ≤ n) → ∀ A : Fin m → ℝ × ℝ, ∀ B : Fin n → ℝ × ℝ,
```
more a question: which is it? docstring and issue say less or equal.

FormalConjectures/Paper/LeastSquaresPointMatching.lean
```diff
@@ -0,0 +1,101 @@
...
+  {π : Fin m ↪ Fin n |
+    ∃ T : ℝ × ℝ, -- translation
+    let B' := fun i => B i + T
+    (∀ π' : Fin m ↪ Fin n, π' ≠ π → SumOfSquaredDistances A B' π' > SumOfSquaredDistances A B' π)}.ncard
+    ≤ c * m ^ d1 * n ^ d2 + 1 -- add +1 to account for the case m=0
+    := by
+  sorry
+-- The specific conjecture is that this is true for d1 = d2 = 2.
```
mo271: this should be in a docstring.
Also: 
Both `d1` `d2` and d appear as theorem parameters (implicit ∀), making the conjecture say: "for every choice of exponents $d_1, d_2$, there exists a constant $c$ such that..." This is correct (if it holds for $d_1 = d_2 = 2$, it holds for any larger exponents), but nonstandard. A more natural formulation would fix $d_1 = d_2 = 2$ (or state a specific conjecture alongside the general one).

FormalConjectures/Paper/LeastSquaresPointMatching.lean
```diff
@@ -0,0 +1,101 @@
...
+# The number of least-squares matchings under rotations is polynomially bounded
+
+*References:* Günter Rote: Partial least-squares point matching under translations.
+In: 26th European Workshop on Computational Geometry (EuroCG'10), Dortmund, March 2010, pp. 249–251,
+Editor: Jan Vahrenhold.
+`https://page.mi.fu-berlin.de/rote/Papers/abstract/Partial+least-squares+point+matching+under+translations.html`
+
+Rinat Ben-Avraham, Matthias Henze, Rafel Jaume, Balázs Keszegh, Orit E. Raz, Micha Sharir, Igor Tubis.
+Partial-Matching RMS Distance Under Translation: Combinatorics and Algorithms. Algorithmica 80, 2400–2421 (2018).
+`https://doi.org/10.1007/s00453-017-0326-0
```
mo271:
```suggestion
`https://doi.org/10.1007/s00453-017-0326-0`
```

mo271 (commented):
Thanks for the PR, sounds like a fun conjecture, left a few comments

FormalConjectures/Paper/LeastSquaresPointMatching.lean
guenterrote: The m=n case is trivial; so logically, it makes not difference

PR #5209, opened 2026-08-29
GreensOpenProblems/27: values of m at small primes

pawelkwaczynski:
The file states the asymptotics of `m` but no concrete value, so nothing currently ties the
definition to a number. This adds values at nine primes, all proved by finite computation.

For `p = 3, 5, 7` the subset lattice of `ZMod p` is small enough to decide outright, so these are
exact: `m 3 = 3`, `m 5 = 4`, `m 7 = 5`. Both directions go through, which is the part that would
catch a misreading of the definition; an upper bound on its own would not.

For `p = 53, 59, 61, 67, 71, 73` only the upper bound is within reach:

    m 53 ≤ 14, m 59 ≤ 15, m 61 ≤ 15, m 67 ≤ 16, m 71 ≤ 16, m 73 ≤ 16

Those are the values recorded in [A398173](https://oeis.org/A398173) and the witnesses are the sets
recorded with that sequence. The matching lower bounds come from an exhaustive search over subsets
of `ZMod p`, far out of reach of a kernel computation; the section says so rather than leaving it
implicit.

All nine are kernel-checked, with no `native_decide`:

    'Green27.m_3_eq'            depends on axioms: [propext, Classical.choice, Quot.sound]
    'Green27.m_5_eq'            depends on axioms: [propext, Classical.choice, Quot.sound]
    'Green27.m_7_eq'            depends on axioms: [propext, Classical.choice, Quot.sound]
    'Green27.m_53_le_fourteen'  depends on axioms: [propext, Classical.choice, Quot.sound]
    'Green27.m_59_le_fifteen'   depends on axioms: [propext, Classical.choice, Quot.sound]
    'Green27.m_61_le_fifteen'   depends on axioms: [propext, Classical.choice, Quot.sound]
    'Green27.m_67_le_sixteen'   depends on axioms: [propext, Classical.choice, Quot.sound]
    'Green27.m_71_le_sixteen'   depends on axioms: [propext, Classical.choice, Quot.sound]
    'Green27.m_73_le_sixteen'   depends on axioms: [propext, Classical.choice, Quot.sound]

Cost: the file elaborates in about nine minutes on my laptop, most of it in the six larger `decide`
calls (roughly 50 s at `p = 53`, growing with `p`). If that is too much for CI I am happy to drop
the larger primes, or to move them to `native_decide` under `category test`.

The `set_option` lines are needed: instance search gives up on the default `synthInstance.maxSize`
even though the instance exists, and the kernel needs the higher recursion and heartbeat limits.

Note: #4549 also touches this file. The two changes are in different declarations.

### DIFF
--- a/FormalConjectures/GreensOpenProblems/27.lean
+++ b/FormalConjectures/GreensOpenProblems/27.lean
@@ -23,6 +23,7 @@ References:
 - [Gr24] [Green, Ben. "100 open problems." (2024).](https://people.maths.ox.ac.uk/greenbj/papers/open-problems.pdf#problem.27)
 - [Be23] Bedert, Benjamin. "On unique sums in Abelian groups." Combinatorica 44.2 (2024): 269-298.
 - [St76] Straus, E. G. "Differences of residues (mod p)." Journal of Number Theory 8.1 (1976): 40-42.
+- [A398173] [OEIS A398173](https://oeis.org/A398173), the values of $m(p)$ for the first primes.
 -/
 
 open Asymptotics Filter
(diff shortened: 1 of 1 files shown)


pawelkwaczynski:
Checking in on this one: CLA and checks are green and the PR has been quiet since 29 Aug. It adds exact values of m at p = 3, 5, 7 (both directions) and upper bounds at six larger primes matching OEIS A398173, no native_decide. Happy to split it or trim the CI cost if that helps review. Who is the right person to look at it?

pawelkwaczynski:
Correction to my comment above: I wrote that the checks were green. That was wrong, and I am sorry for the noise. `Build project` has been failing since 29 Aug (run 33244051532).

For what it is worth, the failure does not look like it comes from this PR. In that run `lake build FormalMathlib` finished with "Build completed successfully (8893 jobs)", and eleven minutes into the next build step the job ended with exit code 143 and "The runner has received a shutdown signal. This can happen when the runner service is stopped". That is the runner going away mid-build, not a Lean error.

Could someone re-run that job? If the build cost is the issue, I am happy to split the PR or trim it.

PR #5210, opened 2026-08-29
feat(ErdosProblems): formalise Erdős problem 449

hbakkaoui-hash:
Formalises the statement of [Erdős Problem 449](https://www.erdosproblems.com/449), on the function `r(n)` = the number of pairs of divisors `d₁ < d₂ < 2 d₁` of `n`.

### Content (namespace `Erdos449`)
- Definitions `tauPlus`, `closePairs`, `r`, `block`, `blocks`, matching `ErdosProblems/448`.
- Sanity checks `r_six`, `r_twelve`, `tauPlus_twelve` (by `decide`).
- `card_sq_eq`, `sum_blockLt_card_le`, and the dyadic-block Cauchy–Schwarz bound
  `sq_card_divisors_le : τ(n)^2 ≤ τ⁺(n) * (2 r(n) + τ(n))` — all fully proved (no `sorry`).
- `erdos_449` (`research solved`, answer `False`) and the quantitative variant
  `positive_density` are `sorry`: they require the analytic disproof of Problem 448.

### Note on the inequality
The commentary on erdosproblems.com/449 (as of 2026-08-28) states the intermediate bound as
`r(n) + τ(n) ≥ τ(n)^2 / τ⁺(n)`, which is off by a factor of 2 on `r(n)` — it already fails at
`n = 6` (`r = 1`, `τ = 4`, `τ⁺ = 3`, so `r + τ = 5 < 16/3`). The correct form produced by the
Cauchy–Schwarz argument is `τ⁺(n) · (2 r(n) + τ(n)) ≥ τ(n)^2`, which is what `sq_card_divisors_le`
proves. The conclusion of the problem is unaffected.

The sequence `r(n)` has been submitted to the OEIS (A399440); `τ⁺(n)` is A397433.
I used Claude Code, powered by the Claude Fable 5.1 model, as an AI agent to assist with the work. The agent drafted the proofs of the two lemmas (`card_sq_eq` and `sum_blockLt_card_le`) and corrected the Cauchy–Schwarz lemma, while I compiled and reviewed the final result.

### DIFF
--- a/FormalConjectures/ErdosProblems/449.lean
+++ b/FormalConjectures/ErdosProblems/449.lean
@@ -1,0 +1,257 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# Erdős Problem 449
+
+*References:*
+- [erdosproblems.com/449](https://www.erdosproblems.com/449)
+- [erdosproblems.com/448](https://www.erdosproblems.com/448) (the negative solution used here)
+- [OEIS A397433](https://oeis.org/A397433): the integer sequence $\tau^+(n)$.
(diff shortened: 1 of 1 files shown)


hbakkaoui-hash:
I've added a `formal_proof` attribute to `erdos_449`, pointing to plby/lean-proofs, where the negative answer is proved without `sorry` (`not_erdos_449`, L361). The link is pinned to the same commit as the one already used for `erdos_448`.

The two statements match: same definition of `r(n)`, `τ(n) = n.divisors.card`, and the same `HasDensity 1` formulation, so `answer(False) ↔ P` here is `¬ P` there. I left the `positive_density` variant unlinked, since plby proves a differently shaped statement.

Disclosure: I used an AI assistant (Claude) to find this proof, check that the statements match and prepare the change; I reviewed it before pushing.

PR #5212, opened 2026-08-29
fix(ErdosProblems/700): scope the characterization answer

### LINKED ISSUE #5211: Erdős 700(a): answer is scoped per n rather than over the characterization
## Problem

The source asks to characterize the composite integers `n` satisfying `f(n) = n / P(n)`. The current declaration instead places a proposition-valued `answer(sorry)` after `n` and its hypotheses have been bound:

```lean
theorem erdos_700.parts.i (n : ℕ) (hn : ¬ n.Prime) (hn1 : 1 < n) :
    f n = n / P n ↔ answer(sorry)
```

This permits replacing the answer with `f n = n / P n`, making the theorem `P ↔ P` independently for each admissible `n`. It therefore does not require one characterization of all such composite integers. This is a defect in the formal question, not a solution of Erdős Problem 700.

Source: https://www.erdosproblems.com/700

## Proposed fix

State part (a) as one set-level answer:

```lean
theorem erdos_700.parts.i :
    {n : ℕ | ¬ n.Prime ∧ 1 < n ∧ f n = n / P n} = answer(sorry)
```

and add a signature regression test.

I have a local patch with a kernel-checked self-answer reproduction, changed-scope build, test-driver validation, and no new proof/axiom bypass. I plan to submit it in a focused PR.

Adkid-Zephyr:
Fixes #5211

The source asks for one characterization of the composite integers satisfying `f(n) = n / P(n)`. State part (a) as a set equality, preserving the composite-number restrictions, instead of a proposition-valued answer after binding each `n`.

This changes the formal question; it does not solve Erdős Problem 700. The diff is limited to `FormalConjectures/ErdosProblems/700.lean`.

Validation on the updated branch:

- `lake --wfail build 'FormalConjectures.ErdosProblems.«700»'`
- `lake --wfail test`
- `git diff --check`


Adkid-Zephyr:
@googlebot rescan

Adkid-Zephyr:
@googlebot I signed it!

FormalConjecturesTest/ErdosProblems/Erdos700.lean
mo271, line 1: we don't want this test

mo271:
@Adkid-Zephyr thanks for the contribution!

Can you check again if you can get the CLA check green: https://github.com/google-deepmind/formal-conjectures/pull/5212/checks?check_run_id=99150866668

Adkid-Zephyr:
Thanks for fixing #5211 in #5830. I compared the landed change in `141936eb` with this PR: the theorem restatement is identical; only the docstring wording differs. This earlier PR is now superseded.

For the review record, the requested test was removed in `dbd7bd81`, and the refreshed CLA check passed. Closing this PR because the fix is already on main. The issue and this PR retain the original report and proposed repair.

Issue #5154, opened 2026-08-25
Formalize the perfect-play value of chess


bocowgill:
### What is the conjecture

Under perfect play from the standard initial position, chess has the game-theoretic value of a draw. Equivalently, White has a strategy that avoids losing and Black has a strategy that avoids losing.

The earliest published statement I can find appears in the anonymously authored *Traité théorique et pratique du jeu des échecs* (1775), pp. 162–163. Later bibliographies attribute the work to Léger, Bernard, Carlier, and Verdoni, without identifying the author of this passage.

References:

* [Société d'Amateurs, *Traité théorique et pratique du jeu des échecs* (1775)](https://archive.org/details/traitthoriqueet00damgoog)
* [FIDE Laws of Chess](https://rcc.fide.com/fide-laws-of-chess_fulltexthtml/)
* [Wikipedia, *Solving chess*](https://en.wikipedia.org/wiki/Solving_chess)
* [Wikipedia, *First-move advantage in chess*](https://en.wikipedia.org/wiki/First-move_advantage_in_chess)

I propose to formalize current FIDE over-the-board chess. The model would include legal moves, checkmate, stalemate, dead positions, claim-based threefold repetition and 50-move draws, and automatic fivefold repetition and 75-move draws. It would exclude clocks, resignation, and draws by agreement because those are not properties of the abstract game tree.

### Prerequisites needed
Mathlib and this repository do not currently contain a chess model or a general perfect-information game framework suitable for this statement. The proposed file therefore needs problem-specific definitions for:

* the board, pieces, castling rights, en passant state, and move legality;
* histories and the FIDE draw rules;
* terminal outcomes, pure strategies, and finite plays from the initial position.

The basic finite types are already supported by [`Fin`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Fin/Basic.html) and [`Fintype`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Fintype/Basic.html). No general-purpose addition to `FormalConjecturesForMathlib` appears necessary.

I have a self-contained prototype of approximately 636 lines importing only `FormalConjecturesUtil`. Its main statement is `WhiteCanAvoidLoss ∧ BlackCanAvoidLoss`, and `lake --wfail build FormalConjectures.Wikipedia.Chess` succeeds.

Before opening a PR, I welcome guidance on the following:

1. Is this conjecture appropriate for the repository?
[...]

PR #5216, opened 2026-08-30
Formalize the perfect-play value of chess

bocowgill:
## Summary

This adds a self-contained model of standard chess under the current FIDE rules and states that its perfect-play value is a draw: each player has a strategy that avoids losing from the initial position.

The model covers legal moves, checkmate, stalemate, dead positions, repetition, and the 50- and 75-move rules. It leaves out clocks, resignation, and agreed draws, since those are not part of the abstract game tree.

The file is under `Books/` because the earliest source cited is the 1775 *Traité des Amateurs*. For now, the chess-specific definitions stay in the problem file.

Addresses #5154.

## Review notes

I would especially appreciate feedback on:

- whether a self-contained model of this size is appropriate here;
- whether `WhiteCanAvoidLoss ∧ BlackCanAvoidLoss` is the right statement;
- whether any of the game definitions should move to `FormalConjecturesForMathlib/`.

## Testing

`lake --wfail build FormalConjectures.Books.Chess`

## AI assistance

I used Codex to help develop the Lean code. I reviewed the final definitions, statement, sources, and diff.

### DIFF
--- a/FormalConjectures/Books/Chess.lean
+++ b/FormalConjectures/Books/Chess.lean
@@ -1,0 +1,640 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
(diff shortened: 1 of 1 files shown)


bocowgill:
This is now ready for review, and all CI checks pass.

bocowgill:
@parabamoghv @dwrensha @antonbobkov -- since each of you has worked on formalizing chess in Lean, I welcome any high-level feedback on this PR, especially on the treatment of legal play and history-dependent draw rules, and on whether the current level of abstraction is appropriate for this conjecture. No need for a line-by-line review.

For context, [Formal Conjectures](https://google-deepmind.github.io/formal-conjectures/) collects formalized statements of conjectures in Lean, both as a benchmark and as a way to clarify the conjectures themselves. The perfect-play value of chess seems like a natural fit. There may be other chess conjectures worth adding as well, and this PR could provide some of the groundwork for them in later PRs.

PR #5234, opened 2026-09-01
fix(Website): Fix scrolling on website

danielchin:
Tested locally, this should fix the issue with scrolling mentioned here: [#Formal conjectures > New Website @ 💬](https://leanprover.zulipchat.com/#narrow/channel/524981-Formal-conjectures/topic/New.20Website/near/619499345)


mo271 (approved):
Nice! 
Instead of testing locally (or in addition to it) one can name the the branch for the pull request foo-webtest and it will build the website in your fork (one also has to enable the serving the website in your fork..)

PR #5235, opened 2026-09-01
feat(site): add Lean syntax highlighting

Paul-Lez:


Paul-Lez:
This is now superseeded.

PR #5241, opened 2026-09-01
fix(ErdosProblems/41): allow repeated summands

Robertboy18:
The source asks for uniqueness of sums `a + b + c` with `a, b, c ∈ A`, apart from reordering. It does not require the three summands to be distinct.

The current helper represents a representation by a `Finset`. That removes multiplicity, so terms such as `a + a + b` cannot be represented at all. In particular, a collision such as `1 + 1 + 3 = 1 + 2 + 2` lies outside the current predicate.

This changes `NtupleCondition` to use `Multiset`:

- membership still requires every summand to lie in `A`;
- `card = n` still records the number of summands;
- multiplicities are preserved;
- multiset equality identifies exactly the trivial coincidences obtained by permutation.

The two theorem statements continue to use the same helper and remain open; this only repairs the hypothesis they express.

Source: https://www.erdosproblems.com/41

Verification: `lake --wfail build 'FormalConjectures.ErdosProblems.«41»'` passes, and repeated-summand/permutation examples were checked in Lean.

AI assistance disclosure: GPT-5.6 Pro helped with the source audit, Lean examples, and drafting. I checked the source, reviewed the change, and compiled it locally.

### DIFF
--- a/FormalConjectures/ErdosProblems/41.lean
+++ b/FormalConjectures/ErdosProblems/41.lean
@@ -28,17 +28,25 @@ namespace Erdos41
 variable {α : Type} [AddCommMonoid α]
 
 /--
-For a given set `A`, the n-tuple sums `a₁ + ... + aₙ` are all distinct for `a₁, ..., aₙ` in `A`
-(aside from the trivial coincidences).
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM

PR #5243, opened 2026-09-01
fix(ErdosProblems/354): use gamma in part ii

Robertboy18:
This fixes two direct transcription errors in Erdős 354.

1. Part (i) uses base `2` in the Lean statement, but its docstring displayed powers of an unbound `γ`. The docstring now displays `α, 2α, 4α, …` and `β, 2β, 4β, …`, matching the source and the theorem.
2. Part (ii) binds `γ ∈ (1, 2)` but passed the literal `2` to `FloorMultiples.interleave`. The quantified `γ` therefore had no effect and part (ii) repeated part (i). It now calls `FloorMultiples.interleave α β γ`.

Source: https://www.erdosproblems.com/354

Verification: `lake --wfail build 'FormalConjectures.ErdosProblems.«354»'` passes.

AI assistance disclosure: GPT-5.6 Pro helped with the source audit, Lean checks, and drafting. I checked the source, reviewed the change, and compiled it locally.

### DIFF
--- a/FormalConjectures/ErdosProblems/354.lean
+++ b/FormalConjectures/ErdosProblems/354.lean
@@ -33,9 +33,9 @@ noncomputable def FloorMultiples.interleave (a b γ : ℝ) (n : ℕ) : ℤ :=
   else
     FloorMultiples b γ (n / 2)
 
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM

PR #5245, opened 2026-09-01
fix(GreensOpenProblems/24): render one-third correctly

Robertboy18:
The two Green 24 docstrings write the conjectured constant as

```tex
{1}{3}
```

which renders as `13`. They now use

```tex
\frac{1}{3}
```

matching the Lean statement `gamma = 1 / 3` and the cited source. This also corrects the prose rendered from this file on downstream problem pages.

Verification: `lake --wfail build 'FormalConjectures.GreensOpenProblems.«24»'` passes, and no malformed `{1}{3}` occurrence remains in the file.

AI assistance disclosure: GPT-5.6 Pro helped identify and draft this correction. I reviewed the source file and compiled the change locally.


mo271 (approved):
Thanks, LGTM

PR #5246, opened 2026-09-01
fix(ErdosProblems/996): correct Fourier partial sums and tails

Robertboy18:
Fixes the Erdős 996 item tracked in #4896.

There are two independent transcription errors in the current file.

First, `fourierPartial` sums over `i ∈ [-k, k]` but every summand uses `fourierCoeff f k`. The result is one fixed coefficient multiplied by a sum of basis functions, rather than the `k`th Fourier partial sum. The coefficient now uses the summation index: `fourierCoeff f i`.

Second, the source hypothesis controls the Fourier tail

```text
‖f - f_k‖₂,
```

but both Lean statements take the norm of `fourierPartial f k` itself. Both now use

```lean
eLpNorm (⇑f - fourierPartial f k) 2 (haarAddCircle (T := 1)).
```

This corrects the helper and the hypotheses in both the main problem and the neighboring `log2` result.

Source: https://www.erdosproblems.com/996

Verification: the focused build and the full `lake --wfail build` both pass (10,097 jobs for the full build), with a clean `git diff --check`.

AI assistance disclosure: GPT-5.6 Pro helped with the source audit, Lean checks, and drafting. I checked the source, reviewed the change, and compiled the full repository locally.

### DIFF
--- a/FormalConjectures/ErdosProblems/996.lean
+++ b/FormalConjectures/ErdosProblems/996.lean
@@ -33,7 +33,7 @@ namespace Erdos996
 
 noncomputable def fourierPartial {T : ℝ} [hT : Fact (0 < T)] (f : Lp ℂ 2 (@haarAddCircle T hT))
     (k : ℕ) : AddCircle T → ℂ :=
-  fun x => ∑ i ∈ Icc (-k : ℤ) k, fourierCoeff f k • fourier i x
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5256, opened 2026-09-03
Use IsSemiprime method

codrut3:
This addresses a remaining task from https://github.com/google-deepmind/formal-conjectures/pull/4450.

I searched through the repo for ad-hoc semiprime definitions and this was the only instance I found.


mo271 (approved):
Thanks, LGTM!

PR #5268, opened 2026-09-04
feat(Wikipedia/Koethe): relations between different formulations

mo271:
in light of https://github.com/tadamcz/koethe

### DIFF
--- a/FormalConjectures/Wikipedia/Koethe.lean
+++ b/FormalConjectures/Wikipedia/Koethe.lean
@@ -26,6 +26,8 @@ open Ideal TwoSidedIdeal Polynomial
 
 open Matrix
 
+universe u v w
+
 variable {R : Type*}
 
 variable [Ring R]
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/Koethe.lean
```diff
@@ -26,6 +26,8 @@ open Ideal TwoSidedIdeal Polynomial
 
 open Matrix
 
+universe u v w
```
smmercuri: Are these needed? Can you get away with inferred universes?

FormalConjectures/Wikipedia/Koethe.lean
```diff
@@ -45,6 +47,66 @@ def KotheRadical : TwoSidedIdeal R := sSup {I : TwoSidedIdeal R | IsNil I}
 -- This is often denoted `Nil*(R)`
 local notation "Nil* " R => KotheRadical R
 
+/- ### Basic API for nil ideals -/
+
+/-- A subset of a nil subset is nil. -/
+@[category API, AMS 16]
+theorem IsNil.mono {S : Type*} [SetLike S R] {I J : S} (h : (I : Set R) ⊆ J) (hJ : IsNil J) :
```
smmercuri, line 54: Is there some way of getting rid of `SetLike S R`? Maybe we could just state everything for `TwoSidedIdeal` and then use `TwoSidedIdeal.asIdeal` when needed? Not really a blocker for this PR, but worth thinking about to avoid these set coercions.

FormalConjectures/Wikipedia/Koethe.lean
smmercuri: `I` and `hI` are not needed here right?

FormalConjectures/Wikipedia/Koethe.lean
```diff
@@ -79,6 +141,277 @@ theorem KotherConjecture.variants.matrixOver_KotherRadical
...
+
+section ColumnIdeal
+
+variable {n : Type*} [Fintype n] [DecidableEq n]
+
+/-- The left ideal of `Matrix n n R` consisting of the matrices with entries in `I` which are
+supported on the `j`-th column. -/
+def columnIdeal (I : TwoSidedIdeal R) (j : n) : Ideal (Matrix n n R) where
+  carrier := {M | ∀ i k, M i k ∈ I ∧ (k ≠ j → M i k = 0)}
+  zero_mem' := fun _ _ => ⟨I.zero_mem, fun _ => rfl⟩
```
smmercuri:
```suggestion
  zero_mem' _ _ := ⟨I.zero_mem, fun _ => rfl⟩
```
and similarly for the other fields, will golf a couple of lines

FormalConjectures/Wikipedia/Koethe.lean
smmercuri:
```suggestion
/-- The **Köthe conjecture**: for any positive integer `n`, the Köthe radical of `R` is the matrix ideal `M_n(Nil*(R))`. -/
```

FormalConjectures/Wikipedia/Koethe.lean
mo271: ack
not really sure how to best do this, if we improve it let's do it sin a follow up

smmercuri: Happy to leave this for a future PR

FormalConjectures/Wikipedia/Koethe.lean
mo271: done

smmercuri (approved):

PR #5308, opened 2026-09-07
feat: Lang-Trotter conjecture 

bwangpj:
Co-authored-by: @zachary-feng 
Co-authored-by: Claude Opus 5

### DIFF
--- a/FormalConjectures/Arxiv/2109.14256/LangTrotter.lean
+++ b/FormalConjectures/Arxiv/2109.14256/LangTrotter.lean
@@ -19,57 +19,69 @@ import FormalConjecturesUtil
 /-!
 # The Lang–Trotter conjecture
 
-*Reference:* [arxiv/2109.14256](https://arxiv.org/abs/2109.14256)
-**Lang–Trotter Conjecture for CM Elliptic Curves**
-by *Daqing Wan* and *Ping Xi*, Conjecture 1.2.
-
-The original conjecture is due to S. Lang and H. Trotter, *Frobenius distributions in
-`GL₂`-extensions*, Lecture Notes in Mathematics 504, Springer, 1976.
+*References:*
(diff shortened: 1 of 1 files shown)


j2d9w5xtjn-png:
Reviewed at `edc27ba1`. Nice file — the reference notes on the constant are unusually good. One issue with the statement.

**The `c = 0` case is lost.** The docstring on L79–80 says that `c_{E,r} = 0` is to be read as "only finitely many primes `p` with `a_p(E) = r`", and that "the limit formulation below covers both cases". It does not: convergence of `primeCount E r x * log x / √x` to `0` is compatible with `π_{E,r}(x) → ∞`. Anything growing more slowly than `√x / log x` lands in that branch — e.g. `π_{E,r}(x) ~ √x / (log x)²` gives `1 / log x → 0` while producing infinitely many qualifying primes.

So on the `c = 0` branch `lang_trotter` (L86–91) asserts strictly less than Conjecture 1.2 of [WX21], where the vanishing of the constant means finiteness (and, as the note at L123–125 says, comes from a congruence obstruction — see [Ka09]). Since `c` is existentially quantified, the weak branch is always available, which makes the whole statement weaker than intended rather than only in the degenerate case.

Adding the clause is enough, and elaborates as-is against the branch:

```lean
theorem lang_trotter (E : WeierstrassCurve ℚ) [E.IsElliptic] (r : ℤ)
    (hr : HasCM E → r ≠ 0) :
    ∃ c : ℝ, 0 ≤ c ∧
      Tendsto (fun x : ℝ ↦ primeCount E r x * Real.log x / Real.sqrt x) atTop (𝓝 c) ∧
      (c = 0 → {p : Nat.Primes | letI := Fact.mk p.2
        GoodPrime E p ∧ trace E p = r}.Finite) := by
  sorry
```

PR #5309, opened 2026-09-07
feat(Wikipedia): certify rank lower bounds for high-rank elliptic curves

mo271:
Add ECCompute (https://github.com/b-mehta/EllipticCurveRank) as a lake dependency and use its certified 2-descent certificates to prove, for the Elkies (28), Elkies-Klagsbrun (29), ranksunbounded (30) and Claude-Alpöge-Howell (31) curves, that the Mordell-Weil group contains a finitely generated subgroup of the stated rank. The `finrank` bounds follow under a `Module.Finite ℤ E.Point` hypothesis (Mordell-Weil), recorded as `variants.of_finite`.

The points and descent labels live next to the problem file; the wrapper tactic `certify_curve_here` resolves them relative to the source file, since upstream `certify_curve` reads paths relative to the working directory.


mo271:
The idea here is to remove the computations and the added dependency on @b-mehta's repo and add the `formal_proof using formal_conjectures` label

PR #5312, opened 2026-09-07
feat: independence from ZFC

felixpernegger:
A weak definition to state some `Prop` is independent from ZFC (in the sense that it lies between some predefined notions, like CH, that are known to be independent from ZFC)

This code is from my pibase project, in particular

https://github.com/felixpernegger/pibase-lean/pull/1310
https://github.com/felixpernegger/pibase-lean/pull/1315

some of the code about CH i copied from the stale mathlib PR #34075 

This was discussed on Zulip here (and is on the list...): https://leanprover.zulipchat.com/#narrow/channel/524981-Formal-conjectures/topic/Independence.20of.20ZFC/with/621894403

I have thought about this topic quite a bit and think this is the best approach


FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
```diff
@@ -0,0 +1,286 @@
...
+
+See `ContinuumHypothesis.iff_aleph0_covby_continuum` and
+`ContinuumHypothesis.iff_continuum_eq_aleph_one` for typical characterizations.
+-/
+@[mk_iff ContinuumHypothesis.iff_continuum_eq_aleph_one']
+class ContinuumHypothesis where
+  /-- See `of_continuum_eq_aleph_one'` for the universe-generic version. -/
+  private of_continuum_eq_aleph_one' ::
+  /-- See `continuum_eq_aleph_one` for the universe-generic version. -/
+  private continuum_eq_aleph_one' : (𝔠 : Cardinal.{0}) = ℵ₁
```
eric-wieser, line 94: I guess this is taken from https://github.com/leanprover-community/mathlib4/pull/34075

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
```diff
@@ -0,0 +1,286 @@
...
+end ContinuumHypothesis
+
+section GeneralizedContinuumHypothesis
+
+/-- Staement of the generalized continuum hypothesis.
+Note this unfortunately is universe dependent, which is unavoidable. -/
+class GeneralizedContinuumHypothesis.{u} where
+  succ_cardinal_eq_pow {o : Ordinal.{u}} : ℵ₀ ≤ ℵ_ o → ℵ_ (o + 1) = 2 ^ (ℵ_ o)
+
+instance [h : GeneralizedContinuumHypothesis.{v}] : GeneralizedContinuumHypothesis.{v} where
```
eric-wieser: What am I missing here; were these supposed to be different universes?

FormalConjecturesTest/Util/Independence.lean
```diff
@@ -0,0 +1,98 @@
...
+
+-/
+
+@[expose] public section
+
+universe u
+
+/--
+A small set of axioms that are independent of ZFC (and Lean's slightly stronger type theory).
+This list is critical. Some of them are redudant, but worth including for clarity.
```
eric-wieser: What do you mean by `critical`?

eric-wieser:
```suggestion
This list is critical. Some of them are redundant, but worth including for clarity.
```

FormalConjecturesTest/Util/Independence.lean
```diff
@@ -0,0 +1,98 @@
...
+theorem independencePairs_implies {P : Prop × Prop} (hP : P ∈ IndependencePairs) :
+    P.1 → P.2 := by sorry
+-/
+
+/-- `P` is independent of ZFC and Lean's type theory.
+Note: More precisely, the implication can be proved under some set theory axiom known to be
+independent and so does its negation.
+
+A proper, exhaustive and `Prop`-valued notion of independence is likely
+not possible to define in Lean. -/
```
eric-wieser: I think this needs to remark on the relevence of `u`

FormalConjecturesTest/Util/Independence.lean
```diff
@@ -0,0 +1,98 @@
...
+  left
+  exact ⟨P, by simpa⟩
+
+theorem ContinuumHypothesis.independent : Independent.{u} ContinuumHypothesis := by
+  apply independent_of_mem_independenceSet
+  simp [IndependenceSet]
+
+theorem NotContinuumHypothesis.independent : Independent.{u} ContinuumHypothesis := by
+  apply independent_of_mem_independenceSet
+  simp [IndependenceSet]
```
eric-wieser, line 105: These statements are surely not supposed to be identical!

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
felixpernegger: yes (i wrote it in PR description)

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
felixpernegger: If I use `u` for everything either the defs or theorems fail (whether I not add `Type u` at the top). Interestingly enough this does not happen in my repo

FormalConjecturesTest/Util/Independence.lean
felixpernegger: ok

FormalConjecturesTest/Util/Independence.lean
felixpernegger: in the sense that changing it slightly (i.e. one member misformalised) the predicate really breaks. I'll make it clearer

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
eric-wieser: What I mean is that this looks to be the identity implication?

eric-wieser: I assume this is supposed to say something like "if v is larger than u then this is true"?

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
eric-wieser: Let's add a comment in the module docstring too.

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
felixpernegger: oh yeah

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
felixpernegger: its mentioned on line 38 already actually I just noticed

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
```diff
@@ -0,0 +1,286 @@
...
+  obtain ⟨i, iS, ri⟩ := this
+  have iN : i.Nonempty := by
+    contrapose! Sn
+    rwa [← Sn]
+  exact Set.not_disjoint_iff_nonempty_inter.mpr (dense_iff_inter_open.mp dr i (So i iS) iN) ri
+
+
+end Topology
+
+/-
```
eric-wieser: This should be a `/--` comment. Can you add docstrings to the other `Def`s too?

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
```diff
@@ -0,0 +1,286 @@
...
+Martin's axiom for a cardinal. See https://en.wikipedia.org/wiki/Martin%27s_axiom. We choose a topological definition. -/
+def MartinsAxiomFor (k : Cardinal.{v}) : Prop :=
+  ∀ᵉ (X : Type v) (_ : TopologicalSpace X) (s : Set (Set X)),
+    Nonempty X → T2Space X → CompactSpace X → CountableChainCondition X →
+      (∀ a ∈ s, IsNowhereDense a) → Set.sUnion s = Set.univ → k < #s
+
+theorem martinsAxiomFor_of_le {k c : Cardinal.{v}} (h : k ≤ c) (hc : MartinsAxiomFor c) :
```
eric-wieser: This should be called `MartinsAxiomFor.anti`

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
```diff
@@ -0,0 +1,286 @@
...
+  exact Set.not_disjoint_iff_nonempty_inter.mpr (dense_iff_inter_open.mp dr i (So i iS) iN) ri
+
+
+end Topology
+
+/-
+Martin's axiom for a cardinal. See https://en.wikipedia.org/wiki/Martin%27s_axiom. We choose a topological definition. -/
+def MartinsAxiomFor (k : Cardinal.{v}) : Prop :=
+  ∀ᵉ (X : Type v) (_ : TopologicalSpace X) (s : Set (Set X)),
+    Nonempty X → T2Space X → CompactSpace X → CountableChainCondition X →
```
eric-wieser: I think more convenient if you use `[]` binders for the typeclasses here.

felixpernegger: how to do this here?

eric-wieser:
```suggestion
  ∀ (X : Type v) [TopologicalSpace X]
    [Nonempty X] [T2Space X] [CompactSpace X] [CountableChainCondition X] (s : Set (Set X)),
```

felixpernegger: thx

FormalConjecturesTest/Util/Independence.lean
```diff
@@ -0,0 +1,103 @@
...
+
+https://math.stackexchange.com/questions/3764313/why-is-ma-not-provable-from-zfc and
+https://en.wikipedia.org/wiki/Continuum_hypothesis
+-/
+def IndependenceSet : Set Prop :=
+  {ContinuumHypothesis, NotContinuumHypothesis,
+    GeneralizedContinuumHypothesis.{u}, ¬ GeneralizedContinuumHypothesis.{u}, MartinsAxiom,
+      ¬ MartinsAxiom, MartinsAxiom ∧ ¬ ContinuumHypothesis}
+
+/-- (Distinct) Members of `IndependenceSet` `(A, B)` satisfying `A → B` (mathematicallly). -/
```
eric-wieser:
```suggestion
/-- (Distinct) Members of `IndependenceSet` `(A, B)` satisfying `A → B` (mathematically). -/
```

FormalConjecturesTest/Util/Independence.lean
```diff
@@ -0,0 +1,103 @@
...
+    (GeneralizedContinuumHypothesis.{u}, MartinsAxiom),
+    (ContinuumHypothesis, MartinsAxiom),
+    (NotContinuumHypothesis, ¬ GeneralizedContinuumHypothesis.{u}),
+    (¬ MartinsAxiom, NotContinuumHypothesis),
+    (¬ MartinsAxiom, ¬ GeneralizedContinuumHypothesis.{u}),
+    (MartinsAxiom ∧ ¬ ContinuumHypothesis, MartinsAxiom),
+    (MartinsAxiom ∧ ¬ ContinuumHypothesis, NotContinuumHypothesis),
+    (MartinsAxiom ∧ ¬ ContinuumHypothesis, ¬ GeneralizedContinuumHypothesis.{u}) }
+
+/- TODO: prove this (bottleneck is Martin's axiom)
```
eric-wieser: Which implication specifically is the bottleneck?

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
```diff
@@ -0,0 +1,289 @@
...
+
+section Topology
+
+/-- A topological space satisfying the countable chain condition:
+Every collection of pariwise disjoint, nonempty open sets is countable. -/
+class CountableChainCondition (X : Type*) [TopologicalSpace X] : Prop where
+  countable_chain_condition : ∀ ⦃S : Set (Set X)⦄,
+    S.PairwiseDisjoint id → (∀ s ∈ S, IsOpen s) → S.Countable
+
+lemma Set.countable_of_setminus_singleton {α : Type*} {s : Set α} {a : α}
```
eric-wieser:
```suggestion
lemma Set.countable_of_sdiff_singleton {α : Type*} {s : Set α} {a : α}
```

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
```diff
@@ -0,0 +1,289 @@
...
+    (h : (s \ {a}).Countable) : s.Countable :=
+  Countable.of_sdiff h <| countable_singleton a
+
+theorem countableChainCondition_iff_ex_nonempty_chain (X : Type*) [TopologicalSpace X] :
+    CountableChainCondition X ↔ ∀ ⦃S : Set (Set X)⦄,
+      S.PairwiseDisjoint id → (∀ s ∈ S, IsOpen s) → ∅ ∉ S → S.Countable := by
```
eric-wieser: I don't quite understand the name here, where is the `Exists` in the statement?

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
```diff
@@ -0,0 +1,289 @@
...
+  refine ⟨fun h S Sd So _ ↦ h.countable_chain_condition Sd So, fun h ↦ ?_⟩
+  refine { countable_chain_condition := fun S Sd So ↦ ?_ }
+  apply Set.countable_of_setminus_singleton (a := ∅)
+  apply h (PairwiseDisjoint.subset Sd sdiff_subset) fun _ h ↦ So _ h.1
+  simp
+
+--TODO: proof can probably be golfed a lot
+/-- Theorem T21: P26 (SeparableSpace) => P29 (CountableChainCondition) -/
```
eric-wieser: Could you include a remark that these are from pi-base?

j2d9w5xtjn-png:
Reviewed at `5d2e22f6`. `→` binds more loosely than `∧`, so the second disjunct on L75 does not say what it looks like. `#print Independent` on this branch:

```
def Independent.{v} : Prop → Prop :=
fun P => ((∃ A ∈ IndependenceSet, (A ↔ P)) ∨ (∃ Q ∈ IndependencePairs, (Q.1 → ((P ∧ P) → Q.2))))
```

so it is `Q.1 → ((P ∧ P) → Q.2)`, not `(Q.1 → P) ∧ (P → Q.2)`.

That branch is then satisfied by *any* refutable `P`: `P ∧ P` is uninhabited, so `(P ∧ P) → Q.2` holds vacuously and the outer `Q.1 → …` follows. Both of these compile sorry-free against the branch:

```lean
theorem independent_False : Independent.{0} False := by
  right
  refine ⟨(GeneralizedContinuumHypothesis.{0}, ContinuumHypothesis), ?_, ?_⟩
  · left; rfl
  · intro _ h; exact h.1.elim

theorem independent_two_eq_three : Independent.{0} (2 = 3) := by
  right
  refine ⟨(GeneralizedContinuumHypothesis.{0}, ContinuumHypothesis), ?_, ?_⟩
  · left; rfl
  · intro _ h; exact absurd h.1 (by decide)
```

which is exactly the outcome the file's closing comment says should not be reachable.

Parenthesising restores the intent:

```lean
def Independent.{v} (P : Prop) : Prop :=
  (∃ A ∈ IndependenceSet.{v}, A ↔ P) ∨
  (∃ Q ∈ IndependencePairs.{v}, (Q.1 → P) ∧ (P → Q.2))
```

I checked this prints as intended. With it the exploit closes: proving `Independent False` through the second disjunct would need `¬ Q.1` for one of the five left components, none of which is refutable. The first disjunct, `independent_of_mem_independenceSet` and the two `.independent` theorems are unaffected either way.

FormalConjecturesTest/Util/Independence.lean
felixpernegger: CH => Martins axiom, which is essentially https://en.wikipedia.org/wiki/Rasiowa%E2%80%93Sikorski_lemma. I'll write this in the doc

FormalConjecturesTest/ForMathlib/SetTheory/Axioms.lean
felixpernegger: i changed the name, i think this was from an earlier version or something

FormalConjecturesTest/Util/Independence.lean
felixpernegger: nevermind, we can just prove it now

felixpernegger:
> Reviewed at `5d2e22f6`. `→` binds more loosely than `∧`, so the second disjunct on L75 does not say what it looks like. `#print Independent` on this branch:
> 
> ```
> def Independent.{v} : Prop → Prop :=
> fun P => ((∃ A ∈ IndependenceSet, (A ↔ P)) ∨ (∃ Q ∈ IndependencePairs, (Q.1 → ((P ∧ P) → Q.2))))
> ```
> 
> so it is `Q.1 → ((P ∧ P) → Q.2)`, not `(Q.1 → P) ∧ (P → Q.2)`.
> 
> That branch is then satisfied by _any_ refutable `P`: `P ∧ P` is uninhabited, so `(P ∧ P) → Q.2` holds vacuously and the outer `Q.1 → …` follows. Both of these compile sorry-free against the branch:
> 
> ```lean
> theorem independent_False : Independent.{0} False := by
>   right
>   refine ⟨(GeneralizedContinuumHypothesis.{0}, ContinuumHypothesis), ?_, ?_⟩
>   · left; rfl
>   · intro _ h; exact h.1.elim
> 
> theorem independent_two_eq_three : Independent.{0} (2 = 3) := by
>   right
>   refine ⟨(GeneralizedContinuumHypothesis.{0}, ContinuumHypothesis), ?_, ?_⟩
>   · left; rfl
>   · intro _ h; exact absurd h.1 (by decide)
> ```
> 
> which is exactly the outcome the file's closing comment says should not be reachable.
> 
> Parenthesising restores the intent:
> 
> ```lean
> def Independent.{v} (P : Prop) : Prop :=
>   (∃ A ∈ IndependenceSet.{v}, A ↔ P) ∨
>   (∃ Q ∈ IndependencePairs.{v}, (Q.1 → P) ∧ (P → Q.2))
> ```
> 
> I checked this prints as intended. With it the exploit closes: proving `Independent False` through the second disjunct would need `¬ Q.1` for one of the five left components, none of which is refutable. The first disjunct, `independent_of_mem_independenceSet` and the two `.independent` theorems are unaffected either way.
> 
> _Reviewed with [Claude Code](https://claude.com/claude-code)._

thanks

PR #5321, opened 2026-09-07
feat(Wikipedia): shannon capacity of C7

felixpernegger:
there is a lot more you can say about it, but i think this is a good start. The upper bound is the Lovasz number

### DIFF
--- a/FormalConjectures/Wikipedia/Shannon.lean
+++ b/FormalConjectures/Wikipedia/Shannon.lean
@@ -119,11 +119,9 @@ theorem root_367_le_shannonCapacityC7 : (367 : ℝ) ^ (5 : ℝ)⁻¹ ≤ shannon
   apply le_trans ?_ (indepNum_strongPow_le_shannonCapacity (cycleGraph 7) (n := 5) (by norm_num))
   simp only [Nat.cast_ofNat]
   gcongr
-  convert Nat.cast_le.mpr threesixseven_le_indepNum_pow_five
+  have (a b : ℕ) : a ≤ b ↔ (a : ℝ) ≤ b := by exact Iff.symm Nat.cast_le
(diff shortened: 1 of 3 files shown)


FormalConjectures/Wikipedia/Shannon.lean
```diff
@@ -0,0 +1,127 @@
...
+
+variable {α β : Type u}
+
+namespace SimpleGraph
+
+/-- Tensor product of simple graphs. `(a₁, b₁)` is adjacent to `(a₂, b₂)` if `a₁` is adjacent to
+`a₂` and `b₁` is adjacent to `b₂`.
+
+Replace once https://github.com/leanprover-community/mathlib4/pull/43170 lands.-/
+def tensorProd (G : SimpleGraph α) (H : SimpleGraph β) : SimpleGraph (α × β) where
```
Paul-Lez: I guess this is maybe better to include in the ForMathlib dir ?

FormalConjectures/Wikipedia/Shannon.lean
```diff
@@ -0,0 +1,127 @@
...
+
+infixl:70 " ⊠ " => strongProd
+
+/-- The power of a simple graph w. r. t. the strong product. -/
+def strongPow (G : SimpleGraph α) : (n : ℕ) → SimpleGraph (Fin n → α)
+  | 0 => completeGraph _
+  | n + 1 => (Fin.succFunEquiv α n).symm.simpleGraph <| (strongPow G n) ⊠ G
+
+/-- The Shannon capacity $\Theta(G) = \sup_n \sqrt[n]{\alpha(G^{\boxtimes n})}$ of a graph. -/
+noncomputable def shannonCapacity (G : SimpleGraph α) : ℝ :=
```
Paul-Lez: This should probably also be in the ForMathlib I think!

PR #5323, opened 2026-09-07
The Buchsbaum-Eisenbud-Horrocks conjecture

Thmoas-Guan:
The statement of Buchsbaum Eisenbud Horrocks conjecture and Total Rank conjecture.
Add solved cases of Total Rank conjecture following arXiv/1702.02560

### DIFF
--- a/FormalConjectures/Arxiv/1702.02560/BuchsbaumEisenbudHorrocks.lean
+++ b/FormalConjectures/Arxiv/1702.02560/BuchsbaumEisenbudHorrocks.lean
@@ -1,5 +1,5 @@
 /-
-Copyright 2025 The Formal Conjectures Authors.
+Copyright 2026 The Formal Conjectures Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/FormalConjectures/Arxiv/1702.02560/BuchsbaumEisenbudHorrocks.lean
+++ b/FormalConjectures/Arxiv/1702.02560/BuchsbaumEisenbudHorrocks.lean
@@ -31,7 +31,7 @@ open CategoryTheory RingTheory.Sequence
 variable (R : Type u) [CommRing R] [ConnectedSpace (PrimeSpectrum R)]
     (M : ModuleCat.{u} R) [Module.Finite R M] [Nontrivial M]
     (P : ProjectiveResolution M) (fin : ∃ n, ∀ i > n, Limits.IsZero (P.complex.X i))
-    (c : ℕ) (ceq : c = Ideal.height (Module.annihilator R M))
+    (c : ℕ) (ceq : c = Ideal.height (Module.annihilator R M)) [∀ i, Module.Finite R (P.complex.X i)]
 
 /-- The `Buchsbaum-Eisenbud-Horrocks Conjecture` about lower bound of Betti number. -/
 @[category research open, AMS 13]
--- a/FormalConjectures/Arxiv/1702.02560/BuchsbaumEisenbudHorrocks.lean
+++ b/FormalConjectures/Arxiv/1702.02560/BuchsbaumEisenbudHorrocks.lean
@@ -1,0 +1,36 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+
+# Buchsbaum Eisenbud Horrocks Conjecture
+
+-/
+
+universe u
+
+open CategoryTheory
+
+/-- The `Buchsbaum-Eisenbud-Horrocks Conjecture` about lower bound of Betti number. -/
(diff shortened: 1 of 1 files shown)


FormalConjectures/Arxiv/1702.02560/BuchsbaumEisenbudHorrocks.lean
```diff
@@ -0,0 +1,72 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
```
mo271:
```suggestion
Copyright 2026 The Formal Conjectures Authors.
```

FormalConjectures/Arxiv/1702.02560/BuchsbaumEisenbudHorrocks.lean
```diff
@@ -0,0 +1,72 @@
...
+namespace Arxiv.«1702.02560»
+
+universe u
+
+open CategoryTheory RingTheory.Sequence
+
+variable (R : Type u) [CommRing R] [ConnectedSpace (PrimeSpectrum R)]
+    (M : ModuleCat.{u} R) [Module.Finite R M] [Nontrivial M]
+    (P : ProjectiveResolution M) (fin : ∃ n, ∀ i > n, Limits.IsZero (P.complex.X i))
+    (c : ℕ) (ceq : c = Ideal.height (Module.annihilator R M))
```
mo271:
```suggestion
    (c : ℕ) (ceq : c = Ideal.height (Module.annihilator R M)) [∀ i, Module.Finite R (P.complex.X i)]
```

Is something like this needed?

Thmoas-Guan: Sorry, I though a non-fg one would imply the ones after all non-fg, but it is not the case? Adding this would be correct choice.

Thmoas-Guan:
BTW, is providing these enough for now? Or is it better to provide more on Betti number and minimal resolution?

j2d9w5xtjn-png:
Reviewed at `40cc0112`. The `fin` and `ceq` hypotheses declared on L33–34 never appear in any of the four theorem types, so Lean drops them — Lean 4 only auto-includes `variable`s that a declaration actually mentions, plus instance binders whose types mention already-included variables. `[∀ i, Module.Finite R (P.complex.X i)]` survives for that second reason; `fin` and `ceq` are explicit binders and do not.

`#check @BuchsbaumEisenbudHorrocksConjecture` on this branch:

```
BuchsbaumEisenbudHorrocksConjecture : ∀ (R : Type u_1) [inst : CommRing R] [ConnectedSpace (PrimeSpectrum R)]
  (M : ModuleCat R) [Module.Finite R ↑M] [Nontrivial ↑M] (P : CategoryTheory.ProjectiveResolution M) (c : ℕ)
  [∀ (i : ℕ), Module.Finite R ↑(P.complex.X i)] [IsNoetherianRing R] (i : ℕ),
  ↑(c.choose i) ≤ Module.rank R ↑(P.complex.X i)
```

No `fin`, no `ceq`, and all four declarations are like this. So `c` is an arbitrary natural number with no link to `Ideal.height (Module.annihilator R M)`, and the resolution is not required to be finite. That makes both open statements false rather than merely too strong. Taking `R = ℚ`, `M = ℚ` and `P = ProjectiveResolution.self` — so `X 0 = ℚ` and `X (i + 1) = 0` — both of these compile against the branch:

```lean
noncomputable abbrev Mq : ModuleCat.{0} ℚ := ModuleCat.of ℚ ℚ
noncomputable abbrev Pq : ProjectiveResolution Mq := ProjectiveResolution.self Mq

theorem Pq_isZero_succ (n : ℕ) : IsZero (Pq.complex.X (n + 1)) := by
  rw [ProjectiveResolution.self_complex]
  exact HomologicalComplex.isZero_single_obj_X _ 0 Mq (n + 1) (by simp)

theorem Pq_rank_succ (n : ℕ) : Module.rank ℚ (Pq.complex.X (n + 1)) = 0 := by
  have := ModuleCat.subsingleton_of_isZero (Pq_isZero_succ n)
  exact rank_subsingleton' ℚ _

instance Pq_finite : ∀ i, Module.Finite ℚ (Pq.complex.X i)
  | 0 => by rw [ProjectiveResolution.self_complex]; exact inferInstanceAs (Module.Finite ℚ ℚ)
  | (n + 1) => by
      have := ModuleCat.subsingleton_of_isZero (Pq_isZero_succ n)
      exact Module.Finite.of_finite

-- at `c = 2`, `i = 1` the conjecture claims `2 ≤ 0`
theorem BEH_is_false : False := by
  have h := BuchsbaumEisenbudHorrocksConjecture ℚ Mq Pq 2 1
  rw [Pq_rank_succ 0] at h
  simp at h
```

and the same for `TotalRankConjecture ℚ Mq Pq 2`, which claims `4 ≤ 1`. (Here the resolution is finite, so it is specifically the missing `ceq` that bites; a non-finite `P` breaks it further.)

`include fin ceq in` before each theorem is enough — I checked that both hypotheses then reappear in the elaborated type:

```lean
include fin ceq in
@[category research open, AMS 13]
theorem BuchsbaumEisenbudHorrocksConjecture [IsNoetherianRing R] :
    ∀ i, c.choose i ≤ Module.rank R (P.complex.X i) := by
  sorry
```

**Separately, on the two solved variants.** L62 and L69 sum over `Finset.range (c + 1)`. Walker's theorem in the cited [arXiv:1702.02560](https://arxiv.org/abs/1702.02560) bounds the *total* Betti number `∑ᵢ βᵢ` over the whole (finite) resolution; truncating the sum at degree `c` is strictly stronger, because `∑_{i=0}^{c} C(c, i) = 2 ^ c` is exactly what the Buchsbaum–Eisenbud–Horrocks conjecture itself supplies in that range. So as written these two claim the open statement rather than the proved one, and should sum over the whole resolution — which `fin` is presumably there to make sense of.

PR #5327, opened 2026-09-07
feat: files for periods of elliptic curves

WilliamCoram:
@multramate asked for the definition of periods of elliptic curves. This PR defines two versions via integrals and lattices and show they are equivalent. 

When David decides what version he wants I will remove the unnecessary code (probably everything except the integral definition), and the full work will exist in @CBirkbeck's AINTLIB repo alongside a near proof of complex uniformisation.


FormalConjecturesTest/ComplexUniformisation.lean
```diff
@@ -0,0 +1,276 @@
+/-
+Copyright 2025 The Formal Conjectures Authors.
```
mo271:
```suggestion
Copyright 2026 The Formal Conjectures Authors.
```

Multramate (commented):
From conversation, seems like we *don't* need complex uniformisation? Ideally we don't want any existential theorems with `sorry`.

WilliamCoram:
> From conversation, seems like we _don't_ need complex uniformisation? Ideally we don't want any existential theorems with `sorry`.

This has been updated to be free of uniformisation. For now all proves are included - but more than likely we will formalise the other definition of period (integral version) and use that here. In AINTLIB will likely sit this definition and the connection between them using uniformisation

FormalConjecturesTest/RealComponents.lean
```diff
@@ -0,0 +1,60 @@
...
+
+/-- The number of connected components of $E(\mathbb{R})$: $2$ if $\Delta > 0$ and $1$ if
+$\Delta < 0$, taken as the definition. A singular curve ($\Delta = 0$) is assigned the value $1$. -/
+def nrRealComponents : ℕ := if 0 < W.Δ then 2 else 1
+
+lemma nrRealComponents_of_pos (h : 0 < W.Δ) : W.nrRealComponents = 2 := if_pos h
+
+lemma nrRealComponents_of_neg (h : W.Δ < 0) : W.nrRealComponents = 1 := if_neg h.not_gt
+
+lemma nrRealComponents_pos : 0 < W.nrRealComponents := by grind [nrRealComponents]
```
Multramate, line 56: This file seems useless. I think we can just move these definitions to `RealPeriod` or `PeriodIntegral`, or just inline it completely, given that it's pretty short anyway.

FormalConjecturesTest/PeriodIntegral.lean
```diff
@@ -0,0 +1,309 @@
...
+  refine congrArg (2 * ·) (setIntegral_congr_fun measurableSet_Ioi fun x _ ↦ ?_)
+  rw [realPeriodIntegrand, ← W.eval_toPoly_twoTorsionPolynomial_sub (x + W.b₂ / 12),
+    add_sub_cancel_right]
+
+/- ## The real period as an integral -/
+
+/-- **The real period as an integral**: the integral of $|\omega|$ over the identity component of
+$E(\mathbb{R})$, multiplied by the number of connected components of $E(\mathbb{R})$. This is the
+real period of the Birch and Swinnerton-Dyer conjecture. -/
+def realPeriodIntegral : ℝ := (W.nrRealComponents : ℝ) * W.leastRealPeriodIntegral
```
Multramate, line 291: Am I understanding correctly that we want this to be the definition? I can't quite see where you proved it's the same as `realPeriod` but I guess it's not done yet?

WilliamCoram: I’m guessing this is the one you will want, the equivalence is being worked on now and will be pushed when done

FormalConjecturesTest/PeriodIntegral.lean
```diff
@@ -0,0 +1,309 @@
...
```
Multramate, line 291: Yeah, I'm convinced that this is now the definition we want, because this just needs `twoTorsionPolynomial` which already exists in mathlib. What's left to do is just to add a wrapper saying the real period of a Weierstrass curve over Q is the real period of a global minimal model (see #5345) base changed to Q. Can you merge that branch, and then extract out just this definition (with no API lemmas for now) and the period over Q in a separate file?

BarinderBanwait: @Multramate perhaps we shouldn't merge one PR into another? just get each one merged into main, and then we'll have both?

WilliamCoram: @Multramate see #5370 for the new PR, with the definition improved - the nrRealComponents was stupid as it was saying the number of connected components was 1 or 2, but never actually showed anything about connected components

WilliamCoram:
Closed as superseded by linked PR

PR #5331, opened 2026-09-07
The Bass-Quillen conjecture

Thmoas-Guan:


FormalConjectures/Arxiv/1810.00617/BassQuillen.lean
```diff
@@ -0,0 +1,90 @@
...
+universe u
+
+variable (R : Type u) [CommRing R] (ι : Type*) [Finite ι]
+
+open TensorProduct IsLocalRing
+
+/-- The `Bass Quillen Conjecture`, stating that any fg projective module over polynomial ring
+over regular ring `R` comes from a projective module over `R`. -/
+@[category research open, AMS 13]
+theorem BassQuillenConjecture [IsRegularRing R] (P : Type*) [AddCommGroup P] [Module R P]
```
felixpernegger:
```suggestion
theorem BassQuillenConjecture [IsRegularRing R] (P : Type u) [AddCommGroup P] [Module R P]
```

felixpernegger (commented):
maybe move this to the wikipedia folder as well

FormalConjectures/Arxiv/1810.00617/BassQuillen.lean
felixpernegger: same for the other statements. (else this could go very wrong due to universe issues)

FormalConjectures/Arxiv/1810.00617/BassQuillen.lean
```diff
@@ -0,0 +1,90 @@
...
+    (P : Type*) [AddCommGroup P] [Module R P]
+    [Module (MvPolynomial ι R) P] [IsScalarTower R (MvPolynomial ι R) P]
+    [Module.Projective (MvPolynomial ι R) P] :
+    ∃ (Q : Type u) (_ : AddCommGroup Q) (_ : Module R Q),
+      Nonempty (((MvPolynomial ι R) ⊗[R] Q) ≃ₗ[MvPolynomial ι R] P) := by
+  sorry
+
+end lowdimesnion
+
+/-- The `Bass Quillen Conjecture` is true for local ring of char zero. -/
```
felixpernegger, line 105: Can you add sources? (for all three statements)

FormalConjectures/Arxiv/1810.00617/BassQuillen.lean
Thmoas-Guan: We should have `Module.Finite` (I forgot, sorry), so it should be correct, but let's just focus on `P : Type u`

FormalConjectures/Arxiv/1810.00617/BassQuillen.lean
Thmoas-Guan: Are the current references good enough?

PR #5335, opened 2026-09-08
feat: add Serre's uniformity conjecture over the rationals

zachary-feng:
Co-authored-by: GPT-6 Astra

Adds `FormalConjectures/Paper/SerreUniformity.lean`, stating Serre’s uniformity question over the rationals: does there exist a bound, independent of the non-CM elliptic curve, beyond which every mod-p Galois representation is surjective?

The formalization:
- Expresses surjectivity without choosing a basis, by requiring every additive automorphism of the p-torsion subgroup to arise from a Galois automorphism.
- Uses Mathlib’s torsion subgroup and point-map definitions.
- Excludes CM curves using the thirteen rational CM j-invariants.
- Uses `answer(sorry) ↔ ...`, with the uniform bound and all curve and prime quantifiers inside the proposition.

References Serre’s original 1972 paper and the introduction of Lemos’s [paper](https://arxiv.org/abs/1702.01985).

The statement is tagged `research open`; its proof is left as `sorry`.


mo271 (approved):

BarinderBanwait (approved):
LGTM!

if anyone were inclined to add to this: many "cases" of serre uniformity are known. by this, i mean that it is known (after Mazur and Bilu--Parent and Bilu--Parent--Rebolledo) that if rho is nonsurjective for p > 37, then the image must be in normaliser of a non-split Cartan; but putting that in is definitely a separate PR)

mo271 (approved):

PR #5347, opened 2026-09-08
Arxiv/1609.08688: fix five misformalisations in sIncreasingrTuples

mo271:
- `IsComparableSet₂` required only one comparable pair to exist; the source (Definition 1.5) requires any two distinct members to be 2-comparable. Now universally quantified.
- `maximalLength_le_isBigO` allowed the exponent Ω to be negative, since `=O` only bounds the norm, making the statement follow from the trivial F(n) ≤ n². Now stated as: there is c > 0 with F(n) ≤ n² / exp(c · log*(n)) for all sufficiently large n.
- `tripleProduct`/`sequenceProduct` put the lex tag on the whole triple (and on the list) while the coordinate pairs kept the product order, so the product of two 2-increasing sequences was not 2-increasing. The source orders the pairs themselves lexicographically; the type is now `Fin 3 → α ×ₗ α`. A `decide` test checks the motivating example.
- `maximalLength_pow` concluded `∀ᶠ m` (all sufficiently large m) where Lemma 1.7 only gives arbitrarily large m; now `∃ᶠ m`. The eventual statement, which holds for every exponent β < α, is added separately as `maximalLength_pow_eventually`.

identified by @tadamcz

### DIFF
--- a/FormalConjectures/Arxiv/1609.08688/sIncreasingrTuples.lean
+++ b/FormalConjectures/Arxiv/1609.08688/sIncreasingrTuples.lean
@@ -220,51 +220,69 @@ than the other. -/
 def IsComparable₂ {α : Type*} [LT α] (t₁ t₂ : Fin 3 → α) : Prop :=
   t₁ <₂ t₂ ∨ t₂ <₂ t₁
 
-/-- A set of triples is $2$-comparable if any two of them are $2$-comparable. -/
+/-- A set of triples is $2$-comparable if any two distinct members of it are $2$-comparable. -/
 def IsComparableSet₂ {α : Type*} [LT α] (s : List (Fin 3 → α)) : Prop :=
-  ∃ t₁ t₂, t₁ ≠ t₂ ∧ t₁ ∈ s ∧ t₂ ∈ s ∧ IsComparable₂ t₁ t₂
+  ∀ t₁ ∈ s, ∀ t₂ ∈ s, t₁ ≠ t₂ → IsComparable₂ t₁ t₂
 
 open Filter in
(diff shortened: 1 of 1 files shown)


tadamcz:
Identified by automated AI pipeline in: https://tadamcz.com/fc-review-results/#/f/Arxiv/1609.08688/sIncreasingrTuples

FormalConjectures/Arxiv/1609.08688/sIncreasingrTuples.lean
```diff
@@ -220,51 +220,69 @@ than the other. -/
...
-def tripleProduct {α : Type*} (a b : Fin 3 → α) : Πₗ (_ : Fin 3), α × α := toLex (Function.prod a b)
+$((a, d), (b, e), (c, f))$, where the pairs are ordered lexicographically. -/
+def tripleProduct {α : Type*} (a b : Fin 3 → α) : Fin 3 → α ×ₗ α :=
+  fun i => toLex (a i, b i)
 
 @[simp, category API, AMS 5]
 theorem tripleProduct_const {α : Type*} (a : α) :
-    tripleProduct (fun _ => a) (fun _ => a) = toLex (fun _ => (a, a)) := by
-  simpa [tripleProduct] using funext fun i => by simp
+    tripleProduct (fun _ => a) (fun _ => a) = fun _ => toLex (a, a) :=
```
smmercuri, line 242: may as well fix these too 
```suggestion
    tripleProduct (fun _ ↦ a) (fun _ ↦ a) = fun _ ↦ toLex (a, a) :=
```

smmercuri (approved):

PR #5351, opened 2026-09-08
feat: Leopoldts conjecture

WilliamCoram:
This PR adds in five variations of Leopoldts conjecture:
- `leopoldt_conjecture` and `LeopoldtConjecture`, two low level versions that can be stated purely from existing Mathlib but are not verbatim that as stated on [Wikipedia](https://en.wikipedia.org/wiki/Leopoldt%27s_conjecture)
- `leopoldt_conjecture.variants.padicRegulator` and `leopoldt_conjecture.variants.zpRank`, two versions that match the statements on Wikipedia 
- `Leopoldt.Mihailescu.LeopoldtConjecture` (Kevins prove4me statement)

We also add a sorry free equivalence folder that says that all statements are equivalent justifying the low level versions really are Leopoldts conjecture.

Within the ForMatlib folder we have added a Leopoldt folder that contains all the required prerequisite material for both the statements and the equivalences. These are for now in a state of slop - but I plan on cleaning most of this up.

I do not expect this to be merged as is. Please can people comment on the following:
- What conjecture statements are wanted
- If the equivalences are wanted / need to be sorry free
- How much of the ForMathlib belongs in this repo, i.e. do you want only the main statements sorried with pointers

Given this information I can cut down the content to the minimum required.

Either way I plan on pushing this (as is) to @CBirkbeck's AINTLIB repo so that we can store a full copy of the work, and can be used as a pointer if we want to leave sorries in files.

Liberal use of Claude was used


WilliamCoram:
Converting to draft, will work on a shorter PR based on a discussion with @Paul-Lez

WilliamCoram:
Closed, with preference given to new PR. This branch stays as the location of the proofs of equivalence

PR #5352, opened 2026-09-08
fix: Ramsey size linear

mo271:
Ramsey size linear: use the Ramsey number, not the size Ramsey number
`IsRamseySizeLinear G` bounded the size Ramsey number r̂(G, H) (minimum
number of edges of a host graph) linearly in e(H). The notion of Erdős,
Faudree, Rousseau and Schelp, and the statements of Erdős problems 566
and 567 on erdosproblems.com, bound the ordinary Ramsey number R(G, H)
(minimum number of vertices of the host). The two are not equivalent:
K₃ is Ramsey size linear, but r̂(K₃, K_{1,t}) grows quadratically in t.

identified by @tadamcz running an AI pipeline: https://tadamcz.com/fc-review-results/#/f/ForMathlib/Combinatorics/SimpleGraph/SizeRamsey

### DIFF
--- a/FormalConjectures/ErdosProblems/566.lean
+++ b/FormalConjectures/ErdosProblems/566.lean
@@ -31,7 +31,7 @@ open SimpleGraph
 
 /--
 Let $G$ be such that any subgraph on $k$ vertices has at most $2k-3$ edges.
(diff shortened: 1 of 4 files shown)


Paul-Lez (approved):
Nice!

PR #5355, opened 2026-09-08
fix(OEIS): disprove the A067857 sign conjecture

SamuelSchlesinger:
Disprove the A067857 sign conjecture: the product of all primes at most 1000 has 168 distinct prime factors, yet its sequence value is negative. Record `answer(False)` and link the Lean proof.

The counterexample and formal proof were developed by Codex (GPT-6), prompted by Samuel Schlesinger. The proof uses only standard Lean axioms, and the changed module passes `lake --wfail build`.


mo271 (approved):
Thanks, LGTM!

PR #5361, opened 2026-09-09
 chore: link to formal proof of navier stokes

felixpernegger:
cant fall asleep

### DIFF
--- a/FormalConjectures/Millenium/NavierStokes.lean
+++ b/FormalConjectures/Millenium/NavierStokes.lean
@@ -27,6 +27,7 @@ the millennium problem specifically concerns the case n = 3.
 ## References
 - [Wikipedia](https://en.wikipedia.org/wiki/Navier%E2%80%93Stokes_existence_and_smoothness)
 - [Clay Mathematics Institute](https://www.claymath.org/wp-content/uploads/2022/06/navierstokes.pdf)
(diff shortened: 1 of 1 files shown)


Paul-Lez (commented):
Looks good to me - thanks for adding! Since this is quite recent (and big), I suggest waiting a little bit just so experts can audit both the Lean statements and the proof (though I don't expect any problem will be found)

Zhiyilang074811:
Nice chore! Ready for review.

felixpernegger:
> Looks good to me - thanks for adding! Since this is quite recent (and big), I suggest waiting a little bit just so experts can audit both the Lean statements and the proof (though I don't expect any problem will be found)

Maybe one should check the declaration one more time; but the work https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Independent.20verification.20of.20OpenAI.20NavierStokesAndEuler/with/622745273 at the very least confirms that these exact Lean declaration have a valid, formal proof in Lean.

mo271 (approved):
Thanks, I think we have sufficient consensus

Paul-Lez:
And [and independent expert had looked at the formalisation](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Independent.20verification.20of.20OpenAI.20NavierStokesAndEuler/near/623566808) so I think we're good!

PR #5363, opened 2026-09-09
fix: preserve repeated terms in Erdős 348

### LINKED ISSUE #5362: Erdős 348 loses repeated terms in its completeness checks
While preparing Erdős #348 as an input for my theorem prover, AI identified a mismatch in the Lean statement.

The problem allows a nondecreasing sequence, including repeated values. Both completeness checks currently apply `IsAddComplete` to `Set.range (Function.updateFinset a s 0)`. Converting the sequence to a set loses those repetitions.

For example, a sequence of ones can represent every natural number using distinct indexed terms. Its range is just `{1}`, which is not complete under the set-based definition. Deleting one sequence position should also leave the other occurrences available.

The existing `IsAddCompleteNatSeq'` predicate preserves indexed occurrences. I propose using it in both clauses, keeping the current "all sufficiently large integers" meaning, the open status, and the unknown answer unchanged. This is separate from the answer-placeholder concern in #4923 and retains the eventual-completeness correction from #685.

Source: https://github.com/google-deepmind/formal-conjectures/blob/990d209a1ffbe75e59e11267242675931f524824/FormalConjectures/ErdosProblems/348.lean

Problem: https://www.erdosproblems.com/348

I plan to submit the correction with Lean regression checks. AI identified the issue and helped prepare this report and the checks.

graviterra:
Fixes #5362.

AI spotted this while I was preparing #348 as an input for Ensemble Prover. The problem allows repeated terms, but `Set.range` removes them. This changes both completeness checks to use the existing indexed-sequence definition, `IsAddCompleteNatSeq'`.

For example, distinct terms of a sequence of ones can sum to any natural number. Converting it to the set `{1}` loses that property. The regression tests check this distinction, deletion of indexed terms, eventual completeness, and both clauses of the conjecture's type.

This corrects the statement; it does not claim to solve the problem. The "all sufficiently large integers" meaning, open status, and unknown answer are unchanged.

Checked with the repository's pinned Lean 4.33.1:

- `lake --wfail build 'FormalConjectures.ErdosProblems.«348»' 'FormalConjecturesTest.ErdosProblems.«348»'`
- `lake --wfail test`
- The statement regression fails on the original code and passes after the correction. The four semantic lemmas use only `propext`, `Classical.choice`, and `Quot.sound`; none depends on `sorryAx`.

AI identified the issue and helped prepare this description, the patch, and the tests.

### DIFF
--- a/FormalConjectures/ErdosProblems/348.lean
+++ b/FormalConjectures/ErdosProblems/348.lean
@@ -35,8 +35,8 @@ $A = \{a_1 \leq a_2 \leq \cdots\}$ of integers such that
 @[category research open, AMS 11]
 theorem erdos_348 :
     { (m, n) | (m) (n) (_ : m < n) (a : ℕ → ℕ) (_ : Monotone a)
-      (_ : ∀ s, s.card = m → IsAddComplete (Set.range (Function.updateFinset a s 0)))
-        (_ : ∀ t, t.card = n → ¬IsAddComplete (Set.range (Function.updateFinset a t 0))) } =
+      (_ : ∀ s, s.card = m → IsAddCompleteNatSeq' (Function.updateFinset a s 0))
+        (_ : ∀ t, t.card = n → ¬IsAddCompleteNatSeq' (Function.updateFinset a t 0)) } =
     answer(sorry) := by
   sorry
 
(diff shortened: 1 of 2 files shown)


graviterra:
Thanks for addressing this in #5824. Closing this PR since the statement correction has landed. The regression tests here remain available if useful.

PR #5367, opened 2026-09-09
Erdős 785: count 0 in A(x), and bound by the thinner set

mo271:
`counting A x` was `|A ∩ [1, x]|` while the sets range over `Set ℕ` and `IsAdditiveComplement` lets `0` contribute to sums. So `0` could be in `A` without being counted, which changes `A(x)B(x) - x` by `B(x)`, not by a constant: translating Ruzsa's construction gives admissible exact complements whose excess is unbounded below, refuting the main statement and the Ruzsa, Chen–Fang and Chen conjecture variants as formalised.

Fix: count `[0, x]`, i.e. adopt the convention of Chen–Fang (who state their results, including the sharp `A(x)B(x) - x = 1` example, for sets of nonnegative integers containing 0). For sets of positive integers this is the Erdős page's `|A ∩ [1, x]|`, and the positive-integer results (Sárközy–Szemerédi, Ruzsa, the linked Lean proof, which assumes positivity) transfer by translating both sets by 1. A module docstring records the convention.

Separately, the Sárközy–Szemerédi and Chen–Fang bounds `o(A(x))` and `O(A(x)^c)` refer to the thinner set (Ruzsa, §1: "A is small and B is large"); with A, B unlabelled the `O(A(x)^c)` version is false for the thicker set at c = 2. Both are now stated with `min (A(x), B(x))`.

identified by @tadamcz running an AI pipeline: https://tadamcz.com/fc-review-results/#/f/ForMathlib/NumberTheory/AdditiveComplement

### DIFF
--- a/FormalConjectures/ErdosProblems/785.lean
+++ b/FormalConjectures/ErdosProblems/785.lean
@@ -39,6 +39,16 @@ import FormalConjecturesUtil
 - [Ru17] Ruzsa, Imre Z., *Exact additive complements*. Q. J. Math. (2017), 227-235.
 - [SaSz94] Sárközy, A. and Szemerédi, E., *On a problem in additive number theory*.
   Acta Math. Hungar. (1994), 237-245.
+
+## Conventions
+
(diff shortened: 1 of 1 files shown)


mo271:
fixed by #5773

PR #5368, opened 2026-09-09
feat(ForMathlib): automorphic forms in the sense of Borel-Jacquet for GL(n)/ℚ

zachary-feng:
Co-authored-by: @bwangpj
Co-authored-by: Claude Code (Fable 5)

This PR formalises Borel and Jacquet's (Corvallis) definition of an automorphic form for `G = GL n / ℚ` with maximal compact `K = O n ℝ`, in the shape Buzzard states it for `GL₂`, in a new file `FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean`.

Writing `G(𝔸) = G(𝔸_f) × G(ℝ)` with `G(𝔸_f) = GL n 𝔸ᶠ[ℤ, ℚ]`, an automorphic form (`IsAutomorphicForm`) is a smooth `f : G(𝔸) → ℂ` satisfying:

- **(a)** left invariance under the diagonally embedded `G(ℚ)` (`ratDiagonal`);
- **(b1)** right invariance under some compact open subgroup of `G(𝔸_f)`;
- **(b2)** `K`-finiteness: the span of the right `K`-translates is finite-dimensional (`AutomorphicForm.IsKFinite`);
- **(c)** `Z(𝔤)`-finiteness: annihilation by an ideal of finite codimension of the centre of the universal enveloping algebra, acting by left invariant differential operators (`AutomorphicForm.IsZFinite`);
- **(d)** slow increase in the archimedean variable (`IsSlowlyIncreasing`).

The cuspidality condition (e) is out of scope (it needs Haar integration over `N(ℚ)\N(𝔸)`).

Notable points:

- The action in condition (c) is **constructed, not assumed**: `X ∈ 𝔤𝔩 n ℝ` acts by differentiation along one-parameter subgroups (`lieDeriv`), the bracket identity comes from the product rule plus symmetry of the second derivative, and complexifying and applying `UniversalEnvelopingAlgebra.lift` yields the action of the centre (`centerAction`).
- Smoothness on `GL n ℝ` is phrased as existence of a `C^∞` extension to the open set of invertible matrices (`IsSmoothOnGL`), avoiding a manifold structure (see the implementation notes for why the scoped matrix norm instances force this).
- Condition (b1) is witnessed concretely: `integralSubgroup n = GL n Ẑ` is proved compact and open in `G(𝔸_f)`.
- Sanity checks exercising every clause of the definition: the constant functions are automorphic forms (`isAutomorphicForm_const`, `isAutomorphicForm_one`), with condition (c) through the character `constantsCharacter` by which the centre acts on constants.
- The automorphic forms form a `ℂ`-submodule (`automorphicForms`), with the right translation representation of `G(𝔸_f)` on it (`rightTranslation`).

The relation to Buzzard's and Getz–Hahn's formulations of (b2) and (d) is documented in the module docstring; only the stated implications are claimed, not equivalences.

### References

- A. Borel and H. Jacquet, *Automorphic forms and automorphic representations*, in Automorphic forms, representations and L-functions (Corvallis), Proc. Sympos. Pure Math. 33 (1979), §4.
- [K. Buzzard, *Automorphic forms for GL2 over Q*](https://www.ma.imperial.ac.uk/~buzzard/maths/research/notes/automorphic_forms_for_gl2_over_Q.pdf), §1.
- [J. R. Getz and H. Hahn, *An Introduction to Automorphic Representations*, GTM 300 (2024)](https://sites.duke.edu/jgetz/files/2022/04/Graduate_Text.pdf), §6.2 and §6.3 — in that text's numbering: Definitions 6.1 (moderate growth), 6.2 (`Z(𝔤)`-finiteness), 6.4 (adelic moderate growth) and 6.5 (adelic automorphic form).


zachary-feng:
@kbuzzard Hi Kevin, would you mind taking a look?

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet/CenterAction.lean
```diff
@@ -0,0 +1,187 @@
...
+public import Mathlib.Algebra.Lie.OfAssociative
+public import Mathlib.Algebra.Lie.UniversalEnveloping
+
+@[expose] public section
+
+/-!
+# The enveloping algebra of `𝔤𝔩 n ℂ`, its centre, and their action
+
+The universal enveloping algebra `U(𝔤𝔩 n ℂ)` of the complexified Lie algebra of `GL n ℝ` and
+the action of its centre on the `C^∞` functions on `GL n ℝ` by left invariant differential
```
kbuzzard:
```suggestion
the action of its centre on the complex-valued `C^∞` functions on `GL n ℝ` by left invariant differential
```

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet/CenterAction.lean
```diff
@@ -0,0 +1,187 @@
...
+
+-- `Matrix n n ℂ` is a Lie ring under the commutator; Mathlib keeps this instance local, since
+-- it competes with the bracket of a Lie algebra given abstractly.
+attribute [local instance 100] LieRing.ofAssociativeRing
+
+/-- The commutator bracket making `𝔤𝔩 n ℂ = Matrix n n ℂ` a Lie ring, available as
+`open scoped Matrix.GeneralLinearGroup`. It agrees with the bracket used by
+`Matrix.GeneralLinearGroup.universalEnveloping` below. -/
+scoped instance instLieRingMatrixComplex : LieRing (Matrix n n ℂ) :=
+  LieRing.ofAssociativeRing
```
kbuzzard: Surely this is not needed after line 52?

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet/CenterAction.lean
```diff
@@ -0,0 +1,187 @@
...
+
+/-- The universal enveloping algebra `U(𝔤𝔩 n ℂ)` of the complexified Lie algebra of `GL n ℝ`.
+The complexification of `𝔤𝔩 n ℝ` is `𝔤𝔩 n ℂ = Matrix n n ℂ` with its commutator bracket.
+
+This abbreviation records the choice of `LieRing.ofAssociativeRing` as the bracket, so that
+downstream files can name the algebra without re-enabling that local instance. -/
+noncomputable abbrev universalEnveloping (n : Type*) [Fintype n] [DecidableEq n] : Type _ :=
```
kbuzzard:
```suggestion
abbrev universalEnveloping (n : Type*) [Fintype n] [DecidableEq n] : Type _ :=
```
Doesn't seem to be needed? And also below.

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet/CenterAction.lean
zachary-feng: Fixed, thanks!

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet/CenterAction.lean
zachary-feng: Fixed, thanks!

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet/CenterAction.lean
zachary-feng: Fixed, thanks!

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet/Subgroups.lean
```diff
@@ -0,0 +1,216 @@
...
+public import Mathlib.NumberTheory.Padics.ProperSpace
+public import Mathlib.RingTheory.DedekindDomain.FiniteAdeleRing
+public import Mathlib.Topology.Algebra.Group.Matrix
+public import Mathlib.Topology.Algebra.RestrictedProduct.TopologicalSpace
+public import Mathlib.Topology.Instances.Matrix
+
+@[expose] public section
+
+/-!
+# The subgroups in the Borel-Jacquet definition of an automorphic form
```
kbuzzard, line 30: Maybe somewhere you should cite what "Borel-Jacquet" is?

bwangpj: done!

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
```diff
@@ -0,0 +1,505 @@
...
+  finite-dimensional;
+* (c) `f` is annihilated by an ideal of finite codimension of the centre of the universal
+  enveloping algebra of the complexified Lie algebra of `G(ℝ)`, acting by left invariant
+  differential operators;
+* (d) for each `x ∈ G(𝔸_f)`, the function `y ↦ f (x, y)` on `G(ℝ)` is slowly increasing.
+
+Smooth means continuous, locally constant in the finite variable and `C^∞` in the archimedean
+one. The further condition (e) cutting out cusp forms, that the constant term along every
+unipotent radical vanishes, is not formalised here: it needs Haar integration over
+`N(ℚ) \ N(𝔸)`.
```
kbuzzard, line 52: A mathematical remark: I think that mathlib has everything needed to actually write down the definition of cuspidal. We have Haar measure, and I *think* that if N(X) just means "upper triangular matrices with 1's on the diagonal and other entries in X" then N(Z-hat x [0,1)) will be a fundamental domain for N(Q) in N(A) so you can just use that to do the integral.

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
```diff
@@ -0,0 +1,505 @@
...
+/-- The `k`-span of the right translates of `f` by the subgroup `K`. -/
+def rightTranslateSpan : Submodule k (G → k) :=
+  Submodule.span k (Set.range fun u : K => fun x => f (x * (u : G)))
+
+lemma self_mem_rightTranslateSpan : f ∈ rightTranslateSpan K k f :=
+  Submodule.subset_span ⟨1, funext fun x => by simp⟩
+
+/-- `f` is `K`-finite: the span of its right `K`-translates is finite-dimensional. This is
+condition (b2) in the definition of an automorphic form. -/
+def IsKFinite : Prop := FiniteDimensional k (rightTranslateSpan K k f)
```
kbuzzard: This could be an `abbrev` and a `class`.

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
```diff
@@ -0,0 +1,505 @@
...
+/-- If `K` is finite then every function is `K`-finite. -/
+lemma isKFinite_of_finite [Finite K] : IsKFinite K k f :=
+  FiniteDimensional.span_of_finite k (Set.finite_range _)
+
+/-- `K`-finiteness is closed under addition: the translate span of `f + g` sits inside the sum
+of the translate spans. -/
+protected lemma IsKFinite.add {f g : G → k} (hf : IsKFinite K k f) (hg : IsKFinite K k g) :
+    IsKFinite K k (f + g) :=
+  haveI : FiniteDimensional k (rightTranslateSpan K k f) := hf
+  haveI : FiniteDimensional k (rightTranslateSpan K k g) := hg
```
kbuzzard: You wouldn't need this if `IsKFinite` is a class and an abbrev

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
```diff
@@ -0,0 +1,505 @@
...
+  Submodule.finiteDimensional_of_le
+    (S₂ := rightTranslateSpan K k f ⊔ rightTranslateSpan K k g) <|
+    Submodule.span_le.mpr <| Set.range_subset_iff.mpr fun u =>
+      add_mem (Submodule.mem_sup_left (Submodule.subset_span ⟨u, rfl⟩))
+        (Submodule.mem_sup_right (Submodule.subset_span ⟨u, rfl⟩))
+
+protected lemma IsKFinite.const_smul {f : G → k} (hf : IsKFinite K k f) (c : k) :
+    IsKFinite K k (c • f) :=
+  haveI : FiniteDimensional k (rightTranslateSpan K k f) := hf
```
kbuzzard: Same here etc etc

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
```diff
@@ -0,0 +1,505 @@
...
+
+end AutomorphicForm
+
+namespace Matrix.GeneralLinearGroup
+
+open AutomorphicForm
+
+open scoped IsDedekindDomain.FiniteAdeleRing
+
+variable {n : Type*} [Fintype n] [Nonempty n]
```
kbuzzard: Here is a spurious nonempty hypothesis.

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
```diff
@@ -0,0 +1,505 @@
...
+
+/-! ### Slow increase: condition (d) -/
+
+/-- The sup norm on the entries of a matrix: for `n = 2` this is
+`|(a b; c d)| = max {|a|, |b|, |c|, |d|}`. -/
+noncomputable def entrySup (M : Matrix n n ℝ) : ℝ :=
+  Finset.univ.sup' Finset.univ_nonempty fun i =>
+    Finset.univ.sup' Finset.univ_nonempty fun j => |M i j|
```
kbuzzard: And here it is used! The "sup norm" on matrices (which I am pretty sure which mathlib has already, so probably we don't have to rewrite it) does not work when n is empty, because you can't take the sup of no real numbers. One trick which would work here would be taking norms which take values in non-negative reals, because in R_{>=0} the sup of the empty set is zero (which is the correct answer, if n is empty then Matrix n n is the zero ring (which has one element), and the norm of the unique element is supposed to be zero.

Another trick would be to use the L^2 norm instead of the L^infty norm: define the norm on matrices to be sqrt(sum_{i,j}|g_{i,j}|^2). This works fine and gives 0 for the empty matrix.

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
```diff
@@ -0,0 +1,505 @@
...
+
+@[simp]
+lemma gnorm_inv (y : GL n ℝ) : gnorm y⁻¹ = gnorm y := by
+  rw [gnorm, gnorm, inv_inv, max_comm]
+
+/-- `gnorm` is uniformly bounded below: since `y * y⁻¹ = 1`, the entries of `y` and `y⁻¹`
+cannot all be small. This makes the exponent in a slow-increase bound enlargeable, hence
+`IsSlowlyIncreasing` closed under addition. -/
+lemma inv_card_le_gnorm (y : GL n ℝ) : (Fintype.card n : ℝ)⁻¹ ≤ gnorm y := by
+  obtain ⟨i⟩ := ‹Nonempty n›
```
kbuzzard: My instinct is to be concerned about lemmas which need nonempty. This lemma looks fine for n empty, because 0^{-1}=0. Do a case split?

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
```diff
@@ -0,0 +1,505 @@
...
+  rw [inv_le_iff_one_le_mul₀ (zero_lt_one.trans_le hcard)]; nlinarith [gnorm_nonneg y]
+
+lemma gnorm_pos (y : GL n ℝ) : 0 < gnorm y :=
+  ((inv_pos.mpr (by exact_mod_cast Fintype.card_pos)).trans_le (inv_card_le_gnorm y))
+
+/-- A function `φ : GL n ℝ → ℂ` is *slowly increasing*, or of *moderate growth*, if
+`‖φ y‖ ≤ C * ‖y‖ ^ r` for some `C` and `r`. This is condition (d) in the definition of an
+automorphic form. -/
+def IsSlowlyIncreasing (φ : GL n ℝ → ℂ) : Prop :=
+  ∃ C r : ℝ, ∀ y : GL n ℝ, ‖φ y‖ ≤ C * gnorm y ^ r
```
kbuzzard: One should check that (0 : Real) ^ (0 : Real) = 1 or else this is wrong for GL_0 (I do expect constant functions to be slowly increasing in the GL_0 case).

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
```diff
@@ -0,0 +1,505 @@
...
+
+lemma IsSlowlyIncreasing.of_bounded {φ : GL n ℝ → ℂ} {C : ℝ} (h : ∀ y, ‖φ y‖ ≤ C) :
```
kbuzzard, line 299: in other words, check that this proof doesn't use [Nonempty n]!

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
```diff
@@ -0,0 +1,505 @@
...
+  enveloping algebra, acting in the archimedean variable. One ideal annihilates every
+  finite-adelic slice at once. -/
+  zFinite : IsZFinite ℂ ↥(centerUniversalEnveloping n)
+    (fun x : GL n 𝔸ᶠ[ℤ, ℚ] => (⟨fun y => f (x, y), smooth.smoothOnGL x⟩ : smoothGL n))
+  /-- (d) `y ↦ f (x, y)` is slowly increasing for each `x ∈ G(𝔸_f)`. -/
+  slowlyIncreasing : ∀ x : GL n 𝔸ᶠ[ℤ, ℚ], IsSlowlyIncreasing fun y => f (x, y)
+
+/-! ### The submodule of automorphic forms and the right translation action -/
+
+omit [Nonempty n] in
```
kbuzzard: It should always be omitted!

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet/Subgroups.lean
```diff
@@ -0,0 +1,216 @@
...
+/-- The subgroup `Γ = G(ℚ)` of `G(𝔸_f) × G(ℝ)`: the range of the diagonal embedding of
+`GL n ℚ`. Condition (a) for an automorphic form is left invariance under this subgroup. -/
+noncomputable def ratDiagonal : Subgroup (GL n 𝔸ᶠ[ℤ, ℚ] × GL n ℝ) :=
+  (diagonalEmbedding n).range
+
+/-- The orthogonal group `O n ℝ` as a subgroup of `GL n ℝ`: the matrices whose transpose is
+their inverse. This is the maximal compact subgroup of `GL n ℝ` — up to conjugacy the only one,
+by the Cartan-Iwasawa-Malcev theorem — and it is the `K` of the pair `(G, K)` in the definition
+of an automorphic form for `GL n`. -/
+def orthogonalSubgroup (n : Type*) [Fintype n] [DecidableEq n] : Subgroup (GL n ℝ) where
```
kbuzzard, line 93: Do we definitely not have this in mathlib already?

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet/Subgroups.lean
```diff
@@ -0,0 +1,216 @@
...
+  exact abs_le_one_iff_mul_self_le_one.mpr (hd ▸ Finset.single_le_sum
+    (f := fun k => M i k * M i k) (fun k _ => mul_self_nonneg _) (Finset.mem_univ j))
+
+/-! ### The compact open subgroup `GL n Ẑ`: condition (b1) -/
+
+section IntegralSubgroup
+
+open IsDedekindDomain RestrictedProduct
+
+set_option backward.isDefEq.respectTransparency false in
```
kbuzzard: Why do we need this? It's technical debt. Oh -- it seems to work fine if you just delete it?

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
kbuzzard: Oh maybe better is to demand that r is a natural? Mathematically this is no different and then for sure 0^0 will be 1. (C should still be a real)

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
bwangpj: done!

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
bwangpj: ok, I took sup in `ℝ≥0` for the norm

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
bwangpj: done!

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
bwangpj: > Oh maybe better is to demand that r is a natural? Mathematically this is no different and then for sure 0^0 will be 1. (C should still be a real)

yes, done!

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet/Subgroups.lean
bwangpj: I think we were concerned that the mathlib one is actually a sub(monoid) of n x n matrices and not GL_n, but I guess we should still use it

FormalConjecturesForMathlib/NumberTheory/AutomorphicForm/BorelJacquet.lean
kbuzzard: We don't need the definition of the cuspidal to state the hard conjectures so let's leave this for now; we should probably just focus on stating the hard conjectures, although we need to be careful that no conjecture assumes cuspidality. Certainly the one I have in mind (existence of Galois representations) does not. Later on we might need to check that functoriality works in the non-cuspidal case but I am confident that this is the case.

kbuzzard:
@Paul-Lez this looks ready to me now. It's the definition of an automorphic form for GL_n/Q. Claude agrees with me that it's mathematically correct.

The general design philosophy here is: we want to build the simplest possible nontrivial examples of automorphic forms so that we can state some really hard conjectures in the Langlands program. The proposal is to stick to GL_n/Q, the simplest family of connected reductive groups over the simplest global field. 

My motivation for this "short cut" was that it would already enable us to state the basic example of Langlands reciprocity relating algebraic automorphic forms to Galois representations. Originally I was worried that after automorphic forms we would have to define automorphic representations, but when I realised that we could short cut this and write a version of the conjecture which applies directly to automorphic forms I started to push the idea that this was a viable project which would lead to a very deep conjecture. The n=1 case of the conjecture is global class field theory, Deligne, Langlands and Wiles proved part of (but not all of) the n=2 case, and the general case is still mostly very open: partial results for general n are in one of the Annals papers which we're formalizing in the Annals project.

But even more is true: I learnt from Zachary and Bryan this week that in fact this definition unlocks much more: they have pointed out that we can also conjecture lots of instances of Langlands reciprocity. One could even define L-functions for these gadgets and state some hugely generalised version of the Riemann hypothesis. So this definition really has a lot of potential when it comes to unlocking interesting conjectures.

It is not clear to me that mathlib wants the definitions in this form, because the "correct" version of Langlands functoriality/reciprocity is for connected reductive groups, which immediately runs into two problems: (1) we don't have the definition of a connected reductive group in mathlib (it's in Tau Ceti) and (2) even when we do, it's extremely unlikely that `GeneralLinearGroup` will be equal to a connected reductive group (which will probably end up being some kind of functor, I suspect; GeneralLinearGroup "is" a functor but it's not literally a functor).

Paul-Lez (approved):
Thanks!

PR #5370, opened 2026-09-09
feat: add definition of periods of elliptic curves using the integral definition

WilliamCoram:
We add the definition of periods of elliptic curves. This will be used to state the BSD conjecture.

This supersedes #5327


FormalConjecturesForMathlib/AlgebraicGeometry/EllipticCurve/Weierstrass.lean
```diff
@@ -0,0 +1,54 @@
...
+lemma toPoly_twoTorsionPolynomial_ne_zero : W.twoTorsionPolynomial.toPoly ≠ 0 :=
+  Cubic.ne_zero_of_a_ne_zero four_ne_zero
+
+@[simp]
+lemma natDegree_toPoly_twoTorsionPolynomial : W.twoTorsionPolynomial.toPoly.natDegree = 3 :=
+  Cubic.natDegree_of_a_ne_zero four_ne_zero
+
+@[simp]
+lemma leadingCoeff_toPoly_twoTorsionPolynomial : W.twoTorsionPolynomial.toPoly.leadingCoeff = 4 :=
+  Cubic.leadingCoeff_of_a_ne_zero four_ne_zero
```
Multramate: These can be inlined to where they're used. I'm saying we don't want these lemmas just because `twoTorsionPolynomial` will be deprecated, so I don't recommend adding more stuff.

FormalConjecturesForMathlib/AlgebraicGeometry/EllipticCurve/PeriodIntegral.lean
```diff
@@ -0,0 +1,579 @@
...
+    rootMultiplicity W.e₁ W.twoTorsionPolynomial.toPoly = 1 :=
+  le_antisymm
+    (Cubic.rootMultiplicity_le_one_of_discr_ne_zero four_ne_zero
+      (W.twoTorsionPolynomial_discr_ne_zero_of_isElliptic (isUnit_iff_ne_zero.mpr two_ne_zero)) _)
+    ((rootMultiplicity_pos W.toPoly_twoTorsionPolynomial_ne_zero).mpr W.isRoot_e₁)
+
+/-- The two-torsion cubic of an elliptic curve factors as $(x - e_1) Q(x)$ with $Q$ positive on
+$[e_1, \infty)$. -/
+lemma exists_toPoly_twoTorsionPolynomial_eq_X_sub_C_mul [W.IsElliptic] : ∃ Q : ℝ[X],
+    W.twoTorsionPolynomial.toPoly = (X - C W.e₁) * Q ∧ ∀ x, W.e₁ ≤ x → 0 < Q.eval x := by
```
Multramate: Can you phrase this and lemmas below using 2-division polynomials instead? Once that's done you should move it to a separate file since it has nothing to do with real periods. Maybe we want to develop a theory of real Weierstrass curves eventually?

Multramate (commented):
Am I understanding correctly that most of the code is just API to explain the connected components?

FormalConjecturesForMathlib/AlgebraicGeometry/EllipticCurve/PeriodIntegral.lean
WilliamCoram: I think I have done what you wanted

WilliamCoram:
> Am I understanding correctly that most of the code is just API to explain the connected components?

Yes most of the API is 1. to show the integrand is integrable, and 2. to resolve the cases of when the discriminant is positive or negative so that we have `realPeriodIntegral_of_pos` and `realPeriodIntegral_of_neg` (splitting when its 2x or not)

WilliamCoram:
Davids suggestion is that this could be reduced to just:

```
variable (W : WeierstrassCurve ℝ)

def realPeriodIntegrand (x : ℝ) : ℝ := (√(W.Ψ₂Sq.eval x))⁻¹

def realPeriodIntegral : ℝ := 2 * ∫ x, W.realPeriodIntegrand x

```

and point to alternate repo's for API

Multramate:
> > Am I understanding correctly that most of the code is just API to explain the connected components?
> 
> Yes most of the API is 1. to show the integrand is integrable, and 2. to resolve the cases of when the discriminant is positive or negative so that we have `realPeriodIntegral_of_pos` and `realPeriodIntegral_of_neg` (splitting when its 2x or not)

Yeah they're both nice. I'm slightly more convinced that we want (1) more than (2) here, but I still think we can hide them.

WilliamCoram:
Closing since the basic definition file has been merged

PR #5371, opened 2026-09-09
weak bsd statement

BarinderBanwait:
we have enough to state weak BSD -- actually @tadamcz 's code was ok on this point.

so while other PRs are going on for stating strong form, i think we may as well get this in

@kbuzzard can you review?


FormalConjectures/Millenium/BSD.lean
```diff
@@ -64,10 +66,34 @@ to the whole plane. -/
...
-    ∃ L, IsLFunction E L := by
+theorem exists_isLFunction_rat : ∃ L, IsLFunction E L := by
   sorry
 
+/-- The **weak Birch and Swinnerton-Dyer conjecture** over $\mathbb{Q}$, a Clay Millennium Prize
+Problem: the $L$-series has a meromorphic continuation whose order at $s = 1$ is
+$\operatorname{rank}_{\mathbb{Z}} E(\mathbb{Q})$. -/
+@[category research open, AMS 11 14]
+theorem birch_and_swinnerton_dyer_conjecture.variants.rat :
+    ∃ L, IsLFunction E L ∧ meromorphicOrderAt L 1 = Module.finrank ℤ E.toAffine.Point := by
```
Multramate: I'd imagine you want the first part to be an assumption (at least according to Kevin and your version above)? I think you just want `_rat` following the naming convention above.

BarinderBanwait: oops, doing now

BarinderBanwait: see now @Multramate

FormalConjectures/Millenium/BSD.lean
```diff
@@ -64,10 +66,34 @@ to the whole plane. -/
 theorem exists_isLFunction (E : WeierstrassCurve K) [E.IsElliptic] : ∃ L, IsLFunction E L := by
   sorry
 
+/-- **Weak Birch and Swinnerton-Dyer conjecture** ([Tate1966], Conjecture (A)): for an elliptic
+curve over a number field $K$, a meromorphic continuation of its $L$-series has order
+$\operatorname{rank}_{\mathbb{Z}} E(K)$ at $s = 1$. -/
+@[category research open, AMS 11 14]
+theorem birch_and_swinnerton_dyer_conjecture [DecidableEq K] [E.IsElliptic] (L : ℂ → ℂ)
```
Multramate: You can also extract out the `[DecidableEq K]` and `[E.IsElliptic]` like you did for `Rat` below.

FormalConjectures/Millenium/BSD.lean
```diff
@@ -64,10 +66,34 @@ to the whole plane. -/
...
-theorem exists_isLFunction_rat (E : WeierstrassCurve ℚ) [E.IsElliptic] :
-    ∃ L, IsLFunction E L := by
+theorem exists_isLFunction_rat : ∃ L, IsLFunction E L := by
   sorry
 
+/-- The **weak Birch and Swinnerton-Dyer conjecture** over $\mathbb{Q}$, a Clay Millennium Prize
+Problem: a meromorphic continuation of the $L$-series of an elliptic curve over $\mathbb{Q}$ has
+order $\operatorname{rank}_{\mathbb{Z}} E(\mathbb{Q})$ at $s = 1$. -/
+@[category research open, AMS 11 14]
+theorem birch_and_swinnerton_dyer_conjecture_rat (L : ℂ → ℂ) (hL : IsLFunction E L) :
```
Multramate: I don't know what the policy is for problems with multiple parts, but do we want to include all of them in one statement or can we separate them to multiple in the same file? I'm saying this because the strong BSD statement will probably land soon.

Multramate (approved):
One thing you could do is `def` or `abbrev` a `weak birch_swinnerton_dyer` proposition that says: given a `IsDecidableEq K` with `NumberField K` and `E : WeierstrassCurve K` with `E.IsElliptic` and `IsLFunction E L`, we have `meromorphicOrderAt L 1 = Module.finrank \Z E.toAffine.Point`. Then the two conjectures would just be `weak_birch_swinnerton_dyer K` for a `NumberField K` and `weak_birch_swinnerton_dyer \Q` instead of stating them twice.

FormalConjectures/Millenium/BSD.lean
```diff
@@ -64,10 +66,34 @@ to the whole plane. -/
...
+    meromorphicOrderAt L 1 = Module.finrank ℤ E.toAffine.Point := by
```
tb65536: I think you want freeRank, not finrank.

tb65536: (actually, maybe it's fine, since the torsion can't contribute to a linearly independent subset over Z)

FormalConjectures/Millenium/BSD.lean
BarinderBanwait: this is probably a question for the maintainers of the repo -- @mo271 what do you think?

my 2c is that 'weak' and 'strong' BSD are separate statements, with 'strong' assuming 'weak'; so it is reasonable to state them separately. and especially so because the 'weak' form is known in analytic rank 0 and 1 cases, so it's easy to state a `research solved` theorem. And strong BSD is also known for individual curves in some cases.

FormalConjectures/Millenium/BSD.lean
Multramate: I think they're the same but `AddCommGroup.freeRank` certainly sounds clearer, so I still think it's a good idea.

FormalConjectures/Millenium/BSD.lean
Multramate: Yeah I agree with you, so we probably want to call this `weak_birch_swinnerton_dyer`. I think we can remove `and`.

FormalConjectures/Millenium/BSD.lean
BarinderBanwait: one reason to keep it as `finrank` is that that's how it is currently stated in `EllipticCurveRank.lean` file e.g. [here](https://github.com/google-deepmind/formal-conjectures/blob/8fb9b56de0ea8d2ec2a1989224012f725ca6d64c/FormalConjectures/Wikipedia/EllipticCurveRank.lean#L69); this file states rank records and Goldfeld's conjecture. so having it different in two places for different elliptic curve conjectures is not ideal.

of course, `freeRank` is the better of the two, then i can change the definition in that file too

tb65536: `freeRank` is relatively new, so it was probably introduced after `EllipticCurveRank.lean` was written?

FormalConjectures/Millenium/BSD.lean
kbuzzard:
```suggestion
namespace BSD
```
Or change the file name!

kbuzzard, line 57: Why meromorphic and not holomorphic?

FormalConjectures/Millenium/BSD.lean
BarinderBanwait: ok the latest version is using freeRank, but it now introduces a hypothesis into the new `Weak` definition, and this hypothesis is true by mordell--weil, so that might be objectionable; please check latest version @tb65536 and @Multramate

FormalConjectures/Millenium/BSD.lean
```diff
@@ -64,10 +66,40 @@ to the whole plane. -/
...
 @[category research solved, AMS 11 14]
-theorem exists_isLFunction_rat (E : WeierstrassCurve ℚ) [E.IsElliptic] :
-    ∃ L, IsLFunction E L := by
+theorem exists_isLFunction_rat : ∃ L, IsLFunction E L := by
+  sorry
+
+end Rat
+
+/-- The **weak Birch and Swinnerton-Dyer conjecture** for a number field $K$: for every elliptic
+curve $E$ over $K$, a meromorphic continuation of its $L$-series has order
```
kbuzzard, line 110:
```suggestion
curve $E$ over $K$, the analytic continuation of its $L$-series has order
```

FormalConjectures/Millenium/BSD.lean
```diff
@@ -64,10 +66,40 @@ to the whole plane. -/
...
+$\operatorname{rank}_{\mathbb{Z}} E(K)$ at $s = 1$.
+
+The rank is `AddCommGroup.freeRank`, which requires $E(K)$ to be finitely generated. That is the
+Mordell--Weil theorem, which Mathlib does not have and which this repository states as a `sorry`
+in `EllipticCurveRank.mordell_weil`, so it appears here as a hypothesis. -/
+def Weak (K : Type*) [Field K] [NumberField K] [DecidableEq K] : Prop :=
```
kbuzzard, line 116: Why does K go before the colon but E after?

FormalConjectures/Millenium/BSD.lean
kbuzzard: The advantage of `Module.finrank` is that you can drop `[AddGroup.FG E.toAffine.Point]`

FormalConjectures/Millenium/BSD.lean
BarinderBanwait: K is a parameter so the two conjectures come out as Weak K and Weak ℚ, which is the whole reason for the def. E stays quantified inside because the conjecture is about all curves over K — if E were a parameter too, both theorems would read ∀ E, Weak K E, which puts duplication back in

FormalConjectures/Millenium/BSD.lean
tb65536: But then is it really the right statement? :)

FormalConjectures/Millenium/BSD.lean
```diff
@@ -39,35 +39,64 @@ import FormalConjecturesUtil
...
+/-- The **weak Birch and Swinnerton-Dyer conjecture** for a number field $K$: for every elliptic
+curve $E$ over $K$, the analytic continuation of its $L$-series has order
+$\operatorname{rank}_{\mathbb{Z}} E(K)$ at $s = 1$.
+
+The rank is `AddCommGroup.freeRank`, which requires $E(K)$ to be finitely generated. That is the
+Mordell--Weil theorem, which Mathlib does not have and which this repository states as a `sorry`
+in `EllipticCurveRank.mordell_weil`, so it appears here as a hypothesis. -/
+def Weak (K : Type*) [Field K] [NumberField K] [DecidableEq K] : Prop :=
+  ∀ (E : WeierstrassCurve K) [E.IsElliptic] [AddGroup.FG E.toAffine.Point] (L : ℂ → ℂ),
+    IsLFunction E L → meromorphicOrderAt L 1 = AddCommGroup.freeRank E.toAffine.Point
```
tb65536, line 118:
```suggestion
    IsLFunction E L → analyticOrderAt L 1 = AddCommGroup.freeRank E.toAffine.Point
```

FormalConjectures/Millenium/BSD.lean
```diff
@@ -39,35 +39,64 @@ import FormalConjecturesUtil
...
-    (x : ℂ) : L =ᶠ[𝓝[≠] x] L' := by
-  have h2 : meromorphicOrderAt (L - L') 2 = ⊤ := meromorphicOrderAt_eq_top_iff.2 <|
-    Filter.Eventually.mono (nhdsWithin_le_nhds <| (Complex.isOpen_re_gt (3 / 2)).mem_nhds
-      (by norm_num)) fun s hs => sub_eq_zero.2 ((hL.2 s hs).trans (hL'.2 s hs).symm)
-  have key : meromorphicOrderAt (L - L') x = ⊤ := not_not.1 fun hx =>
-    (hL.1.sub hL'.1).exists_meromorphicOrderAt_ne_top_iff_forall.1 ⟨x, hx⟩ 2 h2
-  exact (meromorphicOrderAt_eq_top_iff.1 key).mono fun s hs => sub_eq_zero.1 hs
+theorem IsLFunction.unique {L L' : ℂ → ℂ} (hL : IsLFunction E L) (hL' : IsLFunction E L') :
+    L = L' :=
+  AnalyticOnNhd.eq_of_eventuallyEq (fun z _ => hL.1.analyticAt z) (fun z _ => hL'.1.analyticAt z)
```
tb65536: Maybe use `\mapsto`?

BarinderBanwait: done in latest commit

FormalConjectures/Millenium/BSD.lean
BarinderBanwait: done in latest

FormalConjectures/Millenium/BSD.lean
```diff
@@ -39,35 +39,64 @@ import FormalConjecturesUtil
...
-  have h2 : meromorphicOrderAt (L - L') 2 = ⊤ := meromorphicOrderAt_eq_top_iff.2 <|
-    Filter.Eventually.mono (nhdsWithin_le_nhds <| (Complex.isOpen_re_gt (3 / 2)).mem_nhds
-      (by norm_num)) fun s hs => sub_eq_zero.2 ((hL.2 s hs).trans (hL'.2 s hs).symm)
-  have key : meromorphicOrderAt (L - L') x = ⊤ := not_not.1 fun hx =>
-    (hL.1.sub hL'.1).exists_meromorphicOrderAt_ne_top_iff_forall.1 ⟨x, hx⟩ 2 h2
-  exact (meromorphicOrderAt_eq_top_iff.1 key).mono fun s hs => sub_eq_zero.1 hs
+theorem IsLFunction.unique {L L' : ℂ → ℂ} (hL : IsLFunction E L) (hL' : IsLFunction E L') :
+    L = L' :=
+  AnalyticOnNhd.eq_of_eventuallyEq (fun z _ ↦ hL.1.analyticAt z) (fun z _ ↦ hL'.1.analyticAt z)
+    (z₀ := 2) <| Filter.Eventually.mono ((Complex.isOpen_re_gt (3 / 2)).mem_nhds (by norm_num))
```
tb65536: Maybe `Filter.eventually_of_mem` might be more idiomatic?

FormalConjectures/Millenium/BSD.lean
```diff
@@ -39,35 +39,64 @@ import FormalConjecturesUtil
...
+namespace BSD
 
-open scoped Topology
+section NumberField
 
 variable {K : Type*} [Field K] [NumberField K] {E : WeierstrassCurve K}
 
 def IsLFunction (E : WeierstrassCurve K) (L : ℂ → ℂ) : Prop :=
-  Meromorphic L ∧ ∀ s : ℂ, 3 / 2 < s.re → L s = E.LSeries s
+  Differentiable ℂ L ∧ ∀ s : ℂ, 3 / 2 < s.re → L s = E.LSeries s
```
tb65536, line 62: You only need meromorphic to state BSD...

tb65536 (approved):

FormalConjectures/Millenium/BSD.lean
BarinderBanwait: done in latest

FormalConjectures/Millenium/BSD.lean
BarinderBanwait: done in latest

FormalConjectures/Millenium/BSD.lean
BarinderBanwait: @tb65536 approved this PR after this comment and after discussion with the other reviewers; i interpret that as an answer of 'yes' to this question, and am duly resolving this thread. feel free to open again if there is an issue.

BarinderBanwait:
@mo271 this PR is now ready for the review of the reviewers with write access, there has been much back-and-forth among the listed reviewers, and the version of Weak BSD above is the one that everyone is content with

FormalConjectures/Millenium/BSD.lean
```diff
@@ -39,35 +39,75 @@ import FormalConjecturesUtil
   [Birch and Swinnerton-Dyer](https://tadamcz.com/autoformalization-results/#/p/wp-birch-and-swinnerton-dyer-conjecture)
```
kbuzzard, line 39: I didn't see the first PR which added the original version of this file (I wish I had!) and I know that this PR did not add these references, but this is a pretty crappy format. Having `[PDF]`, for me, means that there is a reference called `[PDF]` and that this refers to one specific thing, but there are two distinct things labelled `[PDF]` and I think that both of them probably refer to the line above.

BarinderBanwait: hmmm ok yeah fair point let me fix this now in this pr

FormalConjectures/Millenium/BSD.lean
```diff
@@ -39,35 +39,75 @@ import FormalConjecturesUtil
...
+      (by norm_num)) fun s hs ↦ sub_eq_zero.2 ((hL.2 s hs).trans (hL'.2 s hs).symm)
+  have key : meromorphicOrderAt (L - L') x = ⊤ := not_not.1 fun hx ↦
     (hL.1.sub hL'.1).exists_meromorphicOrderAt_ne_top_iff_forall.1 ⟨x, hx⟩ 2 h2
-  exact (meromorphicOrderAt_eq_top_iff.1 key).mono fun s hs => sub_eq_zero.1 hs
+  exact (meromorphicOrderAt_eq_top_iff.1 key).mono fun s hs ↦ sub_eq_zero.1 hs
 
 /-- **Hasse--Weil conjecture**: the $L$-function of an elliptic curve over a number field extends
 to the whole plane. -/
 @[category research open, AMS 11 14]
 theorem exists_isLFunction (E : WeierstrassCurve K) [E.IsElliptic] : ∃ L, IsLFunction E L := by
```
kbuzzard, line 85: This is definitively *not* the Hasse--Weil conjecture. The Hasse--Weil conjecture is that there is a holomorphic extension. I am happy with the idea of conjecturing meromorphic continuation for the purposes of stating BSD (because BSD assumes *something* and I have been convinced that assuming the minimal thing is fine) but both Hasse and Weil knew well that it should be holomorphic.

FormalConjectures/Millenium/BSD.lean
```diff
@@ -39,35 +39,75 @@ import FormalConjecturesUtil
...
+section Rat
+
+variable (E : WeierstrassCurve ℚ) [E.IsElliptic]
+
 /-- The **Hasse--Weil conjecture** over $\mathbb{Q}$, a consequence of the modularity theorem. -/
 @[category research solved, AMS 11 14]
-theorem exists_isLFunction_rat (E : WeierstrassCurve ℚ) [E.IsElliptic] :
-    ∃ L, IsLFunction E L := by
+theorem exists_isLFunction_rat : ∃ L, IsLFunction E L := by
+  sorry
```
kbuzzard, line 105: Note that the stronger statement that it's holomorphic is also known.

FormalConjectures/Millenium/BSD.lean
BarinderBanwait: done in latest

FormalConjectures/Millenium/BSD.lean
BarinderBanwait: addressed in latest

FormalConjectures/Millenium/BSD.lean
BarinderBanwait: addressed in latest

mo271 (approved):
Thanks, LGTM!

FormalConjectures/Millenium/BSD.lean
```diff
@@ -31,43 +31,102 @@ import FormalConjecturesUtil
...
+
 variable {K : Type*} [Field K] [NumberField K] {E : WeierstrassCurve K}
 
+/-- `L` is an $L$-function of `E`: it is meromorphic on $\mathbb{C}$ and agrees with the
+$L$-series of `E` on $\operatorname{Re} s > 3/2$, where that series converges.
+
+The $L$-function is expected to be holomorphic, and `IsHolomorphicLFunction` is that stronger
+notion, but meromorphy is all that is needed to state the Birch and Swinnerton-Dyer conjecture.
+[Gross2011] states the conjecture under this hypothesis, and we take that form as authoritative. -/
 def IsLFunction (E : WeierstrassCurve K) (L : ℂ → ℂ) : Prop :=
```
Multramate, line 56: Honestly we could have just called this `HasMeromorphicContinuation` and the other one `HasAnalytic/HolomorphicContinuation` :)

BarinderBanwait: one for another PR 😉

Multramate: Thankfully we don't need to add deprecations in this repository :)

PR #5372, opened 2026-09-09
antiRamseyNum: colour only edges of Kₙ, and forbid rainbow copies

mo271:
`antiRamseyNum H n` had two defects, both in the shared definition:

- The colouring was `c : Sym2 (Fin n) → Fin k`, so the n diagonal pairs `s(v, v)`, which are not edges of Kₙ, carried colours that counted towards surjectivity but could never appear on an edge of a copy of H. This inflates the value, e.g. AR(3, K₃) became ≥ 5 instead of 2, and makes both Erdős 1105 statements false (already at k = 3 for the cycles, and at n = k = 5 for the paths).
- It forbade rainbow homomorphisms `H →g ⊤` rather than rainbow copies. A non-injective homomorphism can be rainbow (P₄ → K₃ folding an end vertex onto the first), so the definition rejected extremal colourings such as the star colouring for P₅.

Now the colouring is a `TopEdgeLabeling (Fin n) (Fin k)` (Mathlib's labelling of the edges of Kₙ), `IsRainbow f c` is injectivity of the pullback `c.pullback f`, and only copies `H.Copy ⊤` (injective homomorphisms) are required not to be rainbow. Erdős 1105 is unchanged and builds; a scratch proof checks that every surjective 3-colouring of K₃ now has a rainbow copy of K₃.

identified by @tadamcz running an AI pipeline: https://tadamcz.com/fc-review-results/#/f/ForMathlib/Combinatorics/SimpleGraph/Coloring/Vertex

### DIFF
--- a/FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Coloring/Vertex.lean
+++ b/FormalConjecturesForMathlib/Combinatorics/SimpleGraph/Coloring/Vertex.lean
@@ -18,7 +18,9 @@ module
 public import FormalConjecturesForMathlib.Combinatorics.SimpleGraph.Clique
 public import Mathlib.Data.NNRat.Floor
 public import Mathlib.Combinatorics.Enumerative.DoubleCounting
+public import Mathlib.Combinatorics.SimpleGraph.Coloring.EdgeLabeling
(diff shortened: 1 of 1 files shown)


mo271:
fixed by #5829

PR #5373, opened 2026-09-09
Erdős 918: GCH hypothesis for the Erdős–Hajnal graphs; record indepen…

mo271:
…dence of part (i)

On top of #4940 (ω is the fixed ordinal, citation fix, all-subgraph note):

- `variants.erdos_hajnal` asserted, unconditionally, graphs with exactly ℵ_k vertices. That is [ErHa68b, Corollary 1], which assumes GCH (the unconditional Theorem 2 gives graphs of size exp_{k-1}(ℵ₀)⁺); erdosproblems.com states the result without the size. Since the file deliberately includes the size, add the GCH hypothesis `∀ c, ℵ₀ ≤ c → 2 ^ c = Order.succ c` and cite the corollary.
- Part (i) (an (ℵ₀, ℵ₂)-chromatic graph of size ℵ₂) is independent of ZFC + GCH: existence is consistent by Baumgartner (1984), non-existence is consistent relative to a large cardinal by Foreman–Laver (1988); see Lambie-Hanson–Rinot, §2.1. Mark it and its equivalent all-subgraph form `research solved` with `answer(sorry)`, following the convention of Erdős 1119, and add the references. Part (ii) stays open (the n = 0 case in Lambie-Hanson–Rinot's account of Shelah's result).
- The notes from #4940 on the induced `= ℵ₀` variants said the impossibility argument does not carry over; it does, via an uncountable colour class. Replace the notes with that argument, including why the fixed ω matters.

identified by @tadamcz running an AI pipeline: https://tadamcz.com/fc-review-results/#/f/ErdosProblems/918

### DIFF
--- a/FormalConjectures/ErdosProblems/918.lean
+++ b/FormalConjectures/ErdosProblems/918.lean
@@ -23,6 +23,11 @@ import FormalConjecturesUtil
 - [erdosproblems.com/918](https://www.erdosproblems.com/918)
 - [ErHa68b] Erdős, P. and Hajnal, A., On chromatic number of infinite graphs. (1968), 83--98.
 - [Er69b] Erdős, P., Problems and results in chromatic graph theory. Proof Techniques in Graph Theory (Proc. Second Ann Arbor Graph Theory Conf., Ann Arbor, Mich., 1968) (1969), 27-35.
+- [Bau84] Baumgartner, J. E., Generic graph construction. J. Symbolic Logic 49 (1984), 234--240.
+- [FL88] Foreman, M. and Laver, R., Some downwards transfer properties for $\aleph_2$.
+  Adv. in Math. 67 (1988), 230--238.
+- [LHR19] Lambie-Hanson, C. and Rinot, A., Reflection on the coloring and chromatic numbers.
+  Combinatorica 39 (2019), 165--214. [arXiv:1708.06929](https://arxiv.org/abs/1708.06929)
 -/
 
 universe u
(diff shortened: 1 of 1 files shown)


mo271:
@felixpernegger can you have a look how the "indepence of ZFC" should be handled here?

FormalConjectures/ErdosProblems/918.lean
```diff
@@ -31,10 +36,17 @@ open scoped Cardinal Ordinal
...
-@[category research open, AMS 5]
+set_option linter.style.category_answer false in
+/-- Is there a graph with $\aleph_2$ vertices and chromatic number $\aleph_2$ such that every
+subgraph on $\aleph_1$ vertices has chromatic number $\leq\aleph_0$?
+
+This question is **independent of ZFC** (and of ZFC + GCH), so the statement carries
+`answer(sorry)`. In the terminology of [LHR19, §2.1] it asks for an
+$(\aleph_0, \aleph_2)$-chromatic graph of size $\aleph_2$: Baumgartner [Bau84] showed that the
+existence of such a graph is consistent with GCH, while Foreman and Laver [FL88] showed that,
+relative to a large cardinal hypothesis, its non-existence is consistent with GCH as well. -/
```
felixpernegger, line 48: I really have to preface this that I am not anything close to a logician, so it is very much possible some of the following is not accurate or helpful at all (so feel free to ignore...):

Both directions are not (easily) stateable with my current (simple) framework. We could just cheat it by adding this statement as an axiom, but that would really violate the spirit and not "prove" anything really.

1)  The positive consistency (by Baumgartner in https://www.jstor.org/stable/2274106) uses forcing. I do believe that we can (and definitely should) model some variant of that in Lean eventually, but this will require a very nontrivial amount of work.

2)
The other direction by Foreman and Laver (available here https://www.sciencedirect.com/science/article/pii/0001870888900412) uses a large cardinal axiom (its not immediate clear to me which from skimming the paper).
This would be easy to add to the current framework however: unlike CH, we cannot prove in ZFC that (assuming consistency of ZFC), that a large cardinal axiom is consistent with ZFC since they imply consistency of ZFC (indeed if you go too far out you can run into troubles https://en.wikipedia.org/wiki/Kunen's_inconsistency_theorem).
So "proving under assuming large cardinal" is arguably a weaker notion of independence we care about most, though it would be nice to formalise various large cardinal axioms in a separate def ofc.

In that sense, I would disagree for this problem to actually be completely "research solved", since we still cant prove in ZFC that this statement cant be proved in ZFC (and "merely" have very strong evidence).

felixpernegger: @mo271

mo271:
better fix in #5737

FormalConjectures/ErdosProblems/918.lean
mo271: thanks for the review!

PR #5378, opened 2026-09-09
fix(Erdos/40): state the reduction to Erdős 28 for a single g, and prove it

mo271:
Follow-up to #4993. `erdos_40.variants.implies_erdos_28` assumed `Erdos40ForSet .univ`, i.e. the property for every function g → ∞. That hypothesis is false (g = √N with A = {1}, or A = {2^k}), so the theorem was vacuous and did not record the reduction stated on erdosproblems.com, which needs the property for one g only.

The variant now takes a single `g` with `Tendsto g atTop atTop` and `Erdos40For g`, and the proof is complete: an asymptotic basis of order 2 has |A ∩ [1, N]| ≥ (√N − O(1)) / 2 ≫ √N ≥ √N / g(N) eventually.

`Erdos40ForSet` is removed, since nothing uses it any more after #4993.

identified by @tadamcz running an AI pipeline: https://tadamcz.com/fc-review-results/#/f/ErdosProblems/40

### DIFF
--- a/FormalConjectures/ErdosProblems/40.lean
+++ b/FormalConjectures/ErdosProblems/40.lean
@@ -39,14 +39,6 @@ def Erdos40For (g : ℕ → ℝ) : Prop :=
     limsup (fun N ↦ (sumRep A N : ℕ∞)) atTop = ⊤
 
 /--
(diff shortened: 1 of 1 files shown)


Paul-Lez (approved):
LGTM!

PR #5366, opened 2026-09-09
fix(HilbertProblems/5): Hilbert–Smith statements were false for uncountable discrete groups

j2d9w5xtjn-png:
Fixes #5365.

**What changed**

- `LieGroupPresentation` no longer requires the Lie group to be second countable, and `admitsLieGroupStructure_of_lieGroup` drops its `SecondCountableTopology` hypothesis. New: `admitsLieGroupStructure_of_discreteTopology` (every discrete group is a 0-dimensional Lie group) and `AdmitsLieGroupStructure.locallyCompactSpace`.
- These definitions and their proved API move to `FormalConjecturesForMathlib/Geometry/Manifold/LieGroupPresentation.lean` (CONTRIBUTING step 5); the index is regenerated. The problem file keeps the statements and gets two `test`s: `Circle` and `Multiplicative ℤ` admit a Lie group structure.
- `hilbert_smith_conjecture.variants.riemannian` now takes a Riemannian manifold in Mathlib's sense (`[EMetricSpace X]` with `IsRiemannianManifold (𝓡 n) X`), so the isometries are isometries of the Riemannian distance. Docstring gives the actual argument (Myers–Steenrod, then no small subgroups and Gleason–Montgomery–Zippin).
- `hilbert_smith_padic_formulation` is stated as `¬ FaithfulVAdd ℤ_[p] X` (same proposition as before).
- Module docstring: list of statements, implementation notes (Lie groups are not assumed second countable; the acting group need not be assumed Hausdorff), journal references. Removed the reference to arXiv:math/0103145, an unrefereed preprint claiming a proof.

**Why**

With a second countable carrier, `ℝ` with the discrete topology acting on `ℝ` by translations refutes `hilbert_smith_conjecture` and the `research solved` `dimension_three` variant (Lean disproof in the issue). The sources state the conjecture for arbitrary locally compact groups and count discrete groups as Lie groups. In the Riemannian variant the metric was independent of the manifold structure, so with the discrete metric every map was an isometry.

**How I checked**

- `lake --wfail build FormalConjecturesForMathlib 'FormalConjectures.HilbertProblems.«5»' FormalConjectures.Wikipedia.HilbertFifthProblem` (the only other importer) passes.
- The library theorems and both tests depend only on `propext`, `Classical.choice`, `Quot.sound`.
- The new definitions are satisfied by discrete groups (proved), and the hypotheses of the new Riemannian variant are satisfied by `EuclideanSpace ℝ (Fin n)` (checked with `inferInstance`).

[...]

Issue #5365, opened 2026-09-09
HilbertProblems/5.lean: the Hilbert–Smith statements are false (Lie groups required to be second countable)


j2d9w5xtjn-png:
`Hilbert5.hilbert_smith_conjecture`, `hilbert_smith_conjecture.variants.dimension_three` and `hilbert_smith_conjecture.variants.riemannian` in `FormalConjectures/HilbertProblems/5.lean` conclude `AdmitsLieGroupStructure G`, which asks for a continuous isomorphism from `G` to a Lie group whose carrier is `SecondCountableTopology`.

This makes the statements false. Take `G = ℝ` with the discrete topology acting on `ℝ` by translations. `G` is a locally compact group, the action is continuous and faithful, and `ℝ` is a connected 1-manifold, but `G` is not second countable (a discrete second countable space is countable). The Lean file below derives `False` from `hilbert_smith_conjecture` and from the `research solved` variant `dimension_three`.

The second countable carrier was added in #4595, which fixed `hilbert_fifth_problem`. The conjecture in the literature (Wikipedia, Tao's notes, Pardon) does not assume the group is second countable, and its conclusion "`G` is a Lie group" includes 0-dimensional Lie groups, i.e. arbitrary discrete groups. So the fix is to drop the `SecondCountableTopology` field of `LieGroupPresentation`. `hilbert_fifth_problem` is unaffected: it assumes `SecondCountableTopology G`, which transfers along the homeomorphism.

Separately, `variants.riemannian` takes `[MetricSpace X]` next to the file's `[TopologicalSpace X]`. Only `Isometry` uses the metric, which is otherwise unconstrained. With the discrete metric every bijection is an isometry, so the variant is the open conjecture with a smoothness hypothesis, not a solved case. It should be stated with `IsRiemannianManifold`, so that the isometries are isometries of the Riemannian distance.

<details>
<summary>Lean disproof (checked against current <code>main</code> with <code>lake env lean</code>, no errors)</summary>

```lean
import FormalConjectures.HilbertProblems.«5»

/-! # The statement `Hilbert5.hilbert_smith_conjecture` at formal-conjectures d33e35a5 is false.

Check with `lake env lean HilbertSmithDisproof.lean` from a checkout of google-deepmind/formal-conjectures
at commit d33e35a5 (or any commit before the fix), after building `FormalConjectures.HilbertProblems.«5»`.
Output: no errors. -/

open scoped Manifold ContDiff

[...]

PR #5402, opened 2026-09-09
fix(HilbertProblems/5): drop second countability from hilbert_fifth_problem

Deicyde:
Follow-up to #5366 (see [this review](https://github.com/google-deepmind/formal-conjectures/pull/5366#pullrequestreview-5157381636)); related to #5365.

**What changed**

- `hilbert_fifth_problem` no longer assumes `SecondCountableTopology G`, and its docstring says why none is needed.

**Why**

The Gleason–Montgomery–Zippin theorem holds for every Hausdorff locally Euclidean group without a countability assumption.


mo271 (approved):
Thanks, LGTM!

PR #5404, opened 2026-09-09
feat(Wikipedia): add resolution of singularities

### LINKED ISSUE #5403: Resolution of singularities in positive characteristic
### What is the conjecture

Let $k$ be a field and let $X$ be an integral scheme, separated and of finite type over $k$. A
*resolution of singularities* of $X$ is a proper birational morphism $f : Y \to X$ with $Y$
integral and smooth over $k$. Birational here means that $f$ restricts to an isomorphism between
a dense open subscheme of $Y$ and a dense open subscheme of $X$.

**Conjecture.** If $k$ is perfect of characteristic $p > 0$, then every such $X$ admits a
resolution of singularities.

Hironaka proved this in characteristic zero. In positive characteristic it is known in dimension
at most three: dimensions one and two are classical, and dimension three is due to Cossart and
Piltant. It is open from dimension four on.

Perfectness of $k$ matters for the statement in this form. If $k$ is imperfect and
$a \in k \setminus k^p$, then $\operatorname{Spec} k(a^{1/p})$ is integral, separated and of
finite type over $k$, but every scheme birational to it has function field $k(a^{1/p})$, which is
inseparable over $k$, so none of them is smooth over $k$. Over a general field one asks instead
for $Y$ regular.

**Sources:**
- https://en.wikipedia.org/wiki/Resolution_of_singularities
- H. Hironaka, *Resolution of singularities of an algebraic variety over a field of characteristic
  zero: I, II*, Ann. of Math. 79 (1964), 109-203 and 205-326,
  https://doi.org/10.2307/1970486 and https://doi.org/10.2307/1970547
- J. Kollár, *Resolution of singularities - Seattle lecture*, https://arxiv.org/abs/math/0508332,
  Theorem 36, for Hironaka's theorem stated for all schemes of finite type over a characteristic
  zero field
- H. Hauser, *On the problem of resolution of singularities in positive characteristic (or: a proof
  we are still waiting for)*, Bull. Amer. Math. Soc. 47 (2010), 1-30,
  https://doi.org/10.1090/S0273-0979-09-01274-9
- J. Lipman, *Desingularization of two-dimensional schemes*, Ann. of Math. 107 (1978), 151-207,
  https://doi.org/10.2307/1971141
- V. Cossart, O. Piltant, *Resolution of singularities of threefolds in positive characteristic I,
  II*, J. Algebra 320 (2008), 1051-1082 and 321 (2009), 1836-1976,
  https://doi.org/10.1016/j.jalgebra.2008.03.032 and
  https://doi.org/10.1016/j.jalgebra.2008.11.030. These assume the variety is quasi-projective and
  the ground field differentially finite over a perfect subfield.
- V. Cossart, O. Piltant, *Resolution of singularities of arithmetical threefolds*, J. Algebra 529
  (2019), 268-535, https://doi.org/10.1016/j.jalgebra.2019.02.017. Theorem 1.1 removes both
  restrictions: every reduced separated Noetherian quasi-excellent scheme of dimension at most
  three admits a proper birational morphism from an everywhere regular scheme, in any
  characteristic.

### Prerequisites needed

None. Mathlib has `AlgebraicGeometry.Scheme.BirationalOver`, `IsProper`, `Smooth`,
`LocallyOfFiniteType`, `QuasiCompact`, `IsSeparated` and `AlgebraicGeometry.IsIntegral`, which is
everything the statement needs.

* ams-14

### Choose either option

j2d9w5xtjn-png:
Fixes #5403

Adds resolution of singularities: an integral scheme, separated and of finite type over a field,
admits a proper birational morphism from an integral scheme that is smooth over the field. This is
Hironaka's theorem in characteristic zero and is open in positive characteristic from dimension
four on.

`FormalConjectures/Wikipedia/ResolutionOfSingularities.lean` states:

- `AlgebraicGeometry.Scheme.HasResolution` (new definition, specific to this file): the shared
  conclusion, `∃ Y f, IsIntegral Y ∧ IsProper f ∧ Smooth (f ≫ sX) ∧ Scheme.BirationalOver f (𝟙 X)`.
  Mathlib's `Scheme.BirationalOver f (𝟙 X)` over the base `X` says exactly that `f` restricts to an
  isomorphism between dense open subschemes, which is the birationality of `f` itself rather than
  of `Y` and `X` separately.
- `resolution_of_singularities` (open): the conjecture over a perfect field of characteristic $p$.
- `resolution_of_singularities_of_charZero` (solved): Hironaka's theorem. [Hir1964] is the original;
  [Kol2007] Theorem 36 states it for all schemes of finite type over a characteristic zero field,
  which is the generality used here.
- `resolution_of_singularities_of_topologicalKrullDim_le_three` (solved): the case of dimension at
  most three. In positive characteristic this is due to Cossart and Piltant. Note that their
  threefold papers [CP2008] and [CP2009] assume the variety is quasi-projective and the ground
  field differentially finite over a perfect subfield, so they do not by themselves cover the
  statement here; [CP2019] Theorem 1.1 does, resolving every reduced separated Noetherian
  quasi-excellent scheme of dimension at most three in any characteristic. It produces an
  everywhere regular source, which over a perfect field is the same as a smooth one.
- `exists_not_hasResolution_of_not_perfectField` (textbook): over an imperfect field some
  integral zero-dimensional $X$, separated and of finite type, has no resolution, so
  `PerfectField k` cannot be dropped.
- `Scheme.surjective_of_birationalOver_id` (API, proved): a proper birational morphism is
  surjective. This is what rules out the inclusion of the smooth locus as a trivial solution, and
  it pins down the birational half of the definition.
- `Scheme.hasResolution_of_smooth` (test, proved): a scheme already smooth over the base is its own
  resolution.

Two formalisation choices worth flagging:

- The open statement assumes $k$ perfect, which the source statements do. It is needed in this
  form: if $k$ is imperfect and $a \in k \setminus k^p$, then $\operatorname{Spec} k(a^{1/p})$
  satisfies every hypothesis, but every scheme birational to it has function field $k(a^{1/p})$,
  inseparable over $k$, so none of them is smooth over $k$. Over a general field one asks for $Y$
  regular instead; Mathlib has no scheme-level regularity class, so that variant is not stated.
  `exists_not_hasResolution_of_not_perfectField` states this failure.
- `IsIntegral Y` is implied by the rest in these theorems (`BirationalOver` with `X` integral forces
  `Y` irreducible, and smooth over a field forces reduced), but is kept because it is not implied
  for the arbitrary base of the definition, and so that the statement reads as the informal one.

Finite type over $k$ is `LocallyOfFiniteType sX` together with `QuasiCompact sX`, since Mathlib has
no single finite type class for scheme morphisms.

Built with `lake --wfail build 'FormalConjectures.Wikipedia.ResolutionOfSingularities'`. The two
proved lemmas are sorry-free (`#print axioms` gives only `propext`, `Classical.choice`,
`Quot.sound`).

Disclosure: this formalization was drafted with Claude Code, and reviewed by Codex and by me.

https://claude.ai/code/session_01CJ5UNFtkk7UpQr7xxv2aUH

### DIFF
--- a/FormalConjectures/Wikipedia/ResolutionOfSingularities.lean
+++ b/FormalConjectures/Wikipedia/ResolutionOfSingularities.lean
@@ -20,8 +20,11 @@ import FormalConjecturesUtil
 
 A variety $X$ over a field $k$ admits a resolution of singularities if there is a smooth
 $k$-variety $Y$ and a proper birational morphism $Y \to X$. Hironaka proved that every variety
-over a field of characteristic zero admits one. In positive characteristic this is known only in
-dimension at most three, and is open in general.
+over a field of characteristic zero admits one. In positive characteristic this is known over a
+perfect field in dimension at most three, and is open from dimension four on. Perfectness of $k$
+cannot be dropped from the statement in this form: over an imperfect field it fails already in
+dimension zero, for the reason recorded in the docstring of `resolution_of_singularities` below.
+Over an arbitrary field one asks instead that $Y$ be regular.
 
 *References:*
 * [Wikipedia](https://en.wikipedia.org/wiki/Resolution_of_singularities)
--- a/FormalConjectures/Wikipedia/ResolutionOfSingularities.lean
+++ b/FormalConjectures/Wikipedia/ResolutionOfSingularities.lean
@@ -28,15 +28,21 @@ dimension at most three, and is open in general.
 * [Hir1964] H. Hironaka, Resolution of singularities of an algebraic variety over a field of
   characteristic zero, [I](https://doi.org/10.2307/1970486) and
   [II](https://doi.org/10.2307/1970547), Ann. of Math. 79 (1964), 109--203 and 205--326.
+* [Kol2007] J. Kollár, [Resolution of singularities -- Seattle
+  lecture](https://arxiv.org/abs/math/0508332), Theorem 36.
 * [Hau2010] H. Hauser, [On the problem of resolution of singularities in positive characteristic
   (or: a proof we are still waiting for)](https://doi.org/10.1090/S0273-0979-09-01274-9),
   Bull. Amer. Math. Soc. 47 (2010), 1--30.
+* [Lip1978] J. Lipman, [Desingularization of two-dimensional
+  schemes](https://doi.org/10.2307/1971141), Ann. of Math. 107 (1978), 151--207.
 * [CP2008] V. Cossart and O. Piltant, [Resolution of singularities of threefolds in positive
   characteristic I](https://doi.org/10.1016/j.jalgebra.2008.03.032), J. Algebra 320 (2008),
   1051--1082.
 * [CP2009] V. Cossart and O. Piltant, [Resolution of singularities of threefolds in positive
   characteristic II](https://doi.org/10.1016/j.jalgebra.2008.11.030), J. Algebra 321 (2009),
   1836--1976.
+* [CP2019] V. Cossart and O. Piltant, [Resolution of singularities of arithmetical
+  threefolds](https://doi.org/10.1016/j.jalgebra.2019.02.017), J. Algebra 529 (2019), 268--535.
 -/
 
 open CategoryTheory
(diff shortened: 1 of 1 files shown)


FormalConjectures/Wikipedia/ResolutionOfSingularities.lean
```diff
@@ -0,0 +1,140 @@
...
+problem is open in positive characteristic from dimension four on.
+-/
+@[category research solved, AMS 14]
+theorem resolution_of_singularities_of_topologicalKrullDim_le_three (k : Type u) [Field k]
+    [PerfectField k] {X : Scheme.{u}} (sX : X ⟶ Spec (.of k)) [IsIntegral X]
+    [LocallyOfFiniteType sX] [QuasiCompact sX] [IsSeparated sX]
+    (hX : topologicalKrullDim X ≤ 3) :
+    Scheme.HasResolution sX := by
+  sorry
+
```
Paul-Lez, line 139: Could you add the statement that resolution of singularities fails for non perfect fields? I think this could be a nice sanity check if models can prove that!

Paul-Lez (commented):
Cool work, thanks!

Paul-Lez (approved):
LGTM modulo the comment about adding a statement for the failure of resolution of singularities over non perfect fields!

FormalConjectures/Wikipedia/ResolutionOfSingularities.lean
j2d9w5xtjn-png: Thanks, added in 2d9e2439: `exists_not_hasResolution_of_not_perfectField` says that over any imperfect field there is an integral, zero-dimensional $X$, separated and of finite type, with no resolution (the witness is $\operatorname{Spec} k(a^{1/p})$ for $a \in k \setminus k^p$). It is tagged `textbook` and left as `sorry`. I also merged main and moved the file to the module system (a4a0aded).

PR #5425, opened 2026-09-10
feat(ErdosProblems/399): prove erdos_399.variants.cambie

herakles-dev:
Fills the `sorry` in `erdos_399.variants.cambie`. The docstring already spells out
Cambie's argument ("considerations modulo 8 rule out any solutions"), so this just
formalises it directly.

Sketch: `z^4 % 8 = z % 2`, so coprimality (⇒ `x, y` not both even) gives
`x^4 + y^4 ≡ 1 or 2 (mod 8)`, never `0`; but `8 ∣ n!` for `n ≥ 4` (`4! ∣ n!`, `8 ∣ 24`).
The `n ≤ 3` cases are just `n! ≤ 6 < 16 ≤ x^4 + y^4`.

Only core Mathlib, no new imports; the statement and `@[category]` tag are unchanged.
`lake build` is clean — no errors or warnings, no `sorry` — and `#print axioms` gives the
usual `{propext, Classical.choice, Quot.sound}`.

It's ~40 lines, a bit longer than the shortest fills here but elementary throughout. Happy
to move it to a linked `@[formal_proof]` instead if you'd rather keep the file lean.

AI usage (per CONTRIBUTING → Mathlib's policy): the proof was developed with Claude (Anthropic); I've been through it, and as with any Lean proof the trust is in the kernel — `#print axioms` gives the standard trio with no `sorryAx`.

### DIFF
--- a/FormalConjectures/ErdosProblems/399.lean
+++ b/FormalConjectures/ErdosProblems/399.lean
@@ -74,7 +74,50 @@ $(x,y)=1$ and $xy>1$.
 @[category research solved, AMS 11]
 theorem erdos_399.variants.cambie {n x y : ℕ} :
     x.Coprime y → 1 < x * y → n ! ≠ x ^ 4 + y ^ 4 := by
-  sorry
+  intro hxy h1 heq
+  rcases Nat.lt_or_ge n 4 with hn | hn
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5432, opened 2026-09-10
ErdosProblems/450: sublinear-scale statement is false as formalised (proof inlined), add far-window variant

Remyq:
## What

`erdos_450` asks whether a sufficient window scale `Y` with `Y ε n = o(n)` exists. As formalised it is false for a trivial reason, unrelated to the arithmetic of divisors:

`UniformlySparse ε n y` requires *every* window `(x, x + y)` to be `ε`-sparse. The window starting at `x = n` with `y ≤ n` is full: each integer `m` with `n < m < n + y ≤ 2n` lies in `(n, 2n)` and divides itself, so `localCount n n y = y - 1`. If `Y (1/2) n / n → 0` then for large `n` we can take `y = max (Y (1/2) n) 3 ≤ n`, and sparsity at `x = n` gives `y - 1 ≤ y / 2`, contradicting `y ≥ 3`.

This PR:
- fills `answer(False)` in `erdos_450`, marks it `category research solved`, and inlines the proof (about 35 lines, within the short-proof guideline);
- adds `erdos_450.variants.far_windows` as the new `research open` statement: the same question with windows restricted to `2n ≤ x`, which removes the trivial obstruction. Whether this is the right reading is for reviewers; erdosproblems.com/450 records that the intended quantifier on `x` is unclear;
- leaves `erdos_450.linear_scale_suffices` (externally proved, #4588) untouched. The two results are consistent: under the all-`x` reading the full window forces `y ≥ (n - 1) / ε`, a linear lower bound.

## Verification

Compiled inside this repository at `b82b08f` (Lean 4.33.1, pinned mathlib). `#print axioms Erdos450.erdos_450` gives `[propext, Classical.choice, Quot.sound]`. Supporting definitions, imports and namespace are byte-identical to the current file. The only `sorry`s in the file are the two intended stubs (`far_windows`, `linear_scale_suffices`).

This settles the formal statement only. It does not resolve Erdős's question.

### DIFF
--- a/FormalConjectures/ErdosProblems/450.lean
+++ b/FormalConjectures/ErdosProblems/450.lean
@@ -49,12 +49,67 @@ How large must $y=y(\epsilon,n)$ be such that the number of integers in
 $(x,x+y)$ with a divisor in $(n,2n)$ is at most $\epsilon y$?
 
 A **linear** scale is known to suffice (see `erdos_450.linear_scale_suffices`).
-Whether the optimal scale is *sublinear* — a sufficient `Y` with `Y ε n = o(n)` —
-is open.
+
+With `UniformlySparse` quantifying over *every* window start `x`, a **sublinear** scale
+cannot suffice, for a trivial reason: the window `(n, n + y)` with `y ≤ n` is full, since each of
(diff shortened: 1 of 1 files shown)


mo271:
fixed in #5786, please re-open if something remains to be improved here

PR #5435, opened 2026-09-10
ci: save Lean build cache before documentation rebuilds

danielchin:
Documentation rebuilds some problem modules with different build traces in `.lake/build`. Saving that directory after documentation makes later Lean-only runs rebuild unchanged modules. In the benchmark, documentation changed 69 module traces, and a one-file edit rebuilt those 69 modules as well as the edited module and `All`.

Save the Lean cache immediately after Lean validation, before documentation runs. Cache documentation tools and literate data separately, and initialize documentation dependencies before restoring their build files. Use new cache keys to avoid restoring the old mixed cache.

Validation: actionlint and `git diff --check` passed. A full seed workflow passed, followed by two matched warm-cache comparisons using the same docstring edit to Erdős problem 566. All four runs passed with exact cache hits.

| Measurement | Original, run 1 | Fixed, run 1 | Original, run 2 | Fixed, run 2 |
| --- | ---: | ---: | ---: | ---: |
| Build job, excluding queue | 7m 21s | 4m 51s | 6m 54s | 7m 05s |
| Lean problem build | 3m 40s | 43s | 3m 14s | 2m 41s |
| Problem modules plus `All` rebuilt | 71 | 2 | 71 | 2 |

[Seed](https://github.com/danielchin/formal-conjectures/actions/runs/34378191276), [original 1](https://github.com/danielchin/formal-conjectures/actions/runs/34381150794), [fixed 1](https://github.com/danielchin/formal-conjectures/actions/runs/34381154096), [original 2](https://github.com/danielchin/formal-conjectures/actions/runs/34382131357), [fixed 2](https://github.com/danielchin/formal-conjectures/actions/runs/34382138735).

Both comparisons eliminated all 69 unrelated rebuilds. Overall time averaged 16% lower, but the second fixed run was 11 seconds slower overall: its edited module took 150 seconds versus 34 seconds in the first fixed run. Two samples do not establish a reliable wall-clock speedup. These were manual same-branch simulations; cross-branch PR cache restoration was not tested. The new cache keys require a main-branch run to populate them after merge.

TL;DR: 
Saves the Lean cache before documentation runs.
Caches documentation tools and output separately.
Uses new cache keys so old, mixed caches aren’t restored.

### DIFF
--- a/.github/workflows/build-and-docs.yml
+++ b/.github/workflows/build-and-docs.yml
@@ -135,17 +135,15 @@ jobs:
           lake exe cache unpack
           lake exe cache get
 
-      - name: Restore local lake build
+      - name: Restore Lean build
         if: steps.mode.outputs.website_only != 'true'
         uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
           path: |
             .lake/build
-            docbuild/.lake/build
-            _literate_html
-          key: lake-build-${{ runner.os }}-${{ hashFiles('lean-toolchain', 'lake-manifest.json', 'docbuild/lake-manifest.json', 'docbuild/lakefile.toml') }}-${{ github.sha }}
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #4519, opened 2026-07-21
fix(Wikipedia/BoundedBurnsideProblem): record negative answer

anagnorisis2peripeteia:
The current declaration treats the universal bounded-exponent question as open:

```lean
@[category research open, AMS 20]
theorem bounded_burnside_problem :
    answer(sorry) ↔ ∀ (G : Type) [Group G] (fin_gen : Group.FG G)
      (n : ℕ) (hn : n > 0) (bounded : ∀ g : G, g^n = 1), Finite G := by
  sorry
```

This exact universal proposition has a known negative answer, so this PR records it:

```diff
-@[category research open, AMS 20]
+@[category research solved, AMS 20]
 theorem bounded_burnside_problem :
-    answer(sorry) ↔ ∀ (G : Type) [Group G] ...
+    answer(False) ↔ ∀ (G : Type) [Group G] ...
```

## Why the answer is negative

Novikov and Adian proved that for every odd `n > 4381` there exist infinite, finitely
generated groups of exponent `n` [NA68]. Adian later reduced this to odd `n > 665` [Ad79].
The even case is harder: Ivanov proved `B(m, n)` infinite for `m > 1` and even `n >= 2^48`
divisible by `2^9` [Iv94], and Lysenok improved this to `m > 1` and `n >= 8000` [Ly96].
Any one of those groups, for
example the free Burnside group `B(2, 667)`, is finitely generated, has bounded exponent,
and is infinite. A single such group refutes the proposition stated here, which is
universally quantified over all finitely generated groups and all `n > 0`.

## Scope: this is not the classification problem

The Wikipedia sentence "The full solution to Burnside problem in this form is not known"
refers to the *classification*, that is, which free Burnside groups `B(m, n)` are finite.
That classification is genuinely open, with `B(2, 5)` the best known open case. This file
does not formalise the classification. It formalises the universally quantified question:

```
∀ G, Group.FG G → ∀ n > 0, (∀ g : G, g ^ n = 1) → Finite G
```

which one counterexample closes. So the two statuses are consistent: this proposition is
settled, the classification is not. Arguably the original `research open` tag sat on a
proposition that was never open.

This is a status and answer correction only. The group-theoretic predicate is unchanged and
no claim is made about the classification.

I will follow up with a docstring update carrying the references above and stating
explicitly that the classification, including `B(2, 5)`, remains open, so the same ambiguity
does not catch the next reader.

*References:*
[...]

PR #5443, opened 2026-09-10
feat(Wikipedia/BoundedBurnsideProblem): free groups

mo271:
Follow-up to https://github.com/google-deepmind/formal-conjectures/pull/4519, which recorded the negative answer to the universal
question. The genuinely open content is the classification of finite
free Burnside groups B(m, n), which the file did not formalise.

Define `freeBurnsideGroup m n` as the presented group on `Fin m` with
all n-th powers as relators, prove that its elements satisfy g ^ n = 1
and that B(m, 1) is trivial, and add:
- `burnside_problem_II`: for which (m, n) is B(m, n) finite (open);
- `finite_freeBurnsideGroup_two_five`: is B(2, 5) finite (open);
- the known finite cases n = 1 (m = 1), 2, 3 (Burnside), 4 (Sanov),
  6 (M. Hall), and the infinite cases m ≥ 2 with odd n ≥ 665
  (Novikov–Adian, Adian) or n ≥ 8000 (Lysënok), all `research solved`.

### DIFF
--- a/FormalConjectures/Wikipedia/BoundedBurnsideProblem.lean
+++ b/FormalConjectures/Wikipedia/BoundedBurnsideProblem.lean
@@ -19,18 +19,37 @@ import FormalConjecturesUtil
 /-!
 # Bounded Burnside problem
 
-*Reference:* [Wikipedia](https://en.wikipedia.org/wiki/Burnside_problem#Bounded_Burnside_problem)
+*References:*
+- [Wikipedia](https://en.wikipedia.org/wiki/Burnside_problem#Bounded_Burnside_problem)
+- [NA68] P. S. Novikov, S. I. Adian. "Infinite periodic groups I-III."
+  _Izv. Akad. Nauk SSSR Ser. Mat._ 32 (1968).
+- [Ad79] S. I. Adian. _The Burnside Problem and Identities in Groups._ Springer, 1979.
+- [Iv94] S. V. Ivanov. "The free Burnside groups of sufficiently large exponents."
+  _Internat. J. Algebra Comput._ 4 (1994), 1-308.
+- [Ly96] I. G. Lysënok. "Infinite Burnside groups of even exponent."
+  _Izv. Ross. Akad. Nauk Ser. Mat._ 60 (1996), 3-224.
 -/
 
 namespace BoundedBurnsideProblem
 
 /--
 Let $G$ be a finitely generated group, and assume there exists $n$ such that for every $g$ in $G$,
-$g^n = 1$. Is $G$ necessarily finite?
+$g^n = 1$. Must $G$ be finite?
+
+The answer is negative. Novikov and Adian proved that for every odd $n > 4381$ there exist
+infinite, finitely generated groups of exponent $n$ [NA68]; Adian later reduced this to odd
+$n > 665$ [Ad79]. The even case is harder: Ivanov proved $B(m, n)$ infinite for $m > 1$ and
+even $n \ge 2^{48}$ divisible by $2^9$ [Iv94], and Lysënok improved this to $m > 1$ and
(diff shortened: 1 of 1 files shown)


j2d9w5xtjn-png:
Reviewed at `80291a3e`; `lake --wfail build FormalConjectures.Wikipedia.BoundedBurnsideProblem` is clean here too. Two suggestions about the statements and three smaller notes — nothing in the PR is false, so I'd treat all of it as non-blocking.

**1. `odd n > 665` should be `odd n ≥ 665`** (docstring of `bounded_burnside_problem`).

Adian's theorem includes `n = 665`, and this file's own `infinite_freeBurnsideGroup_of_odd` assumes `665 ≤ n`, so the prose and the theorem currently disagree. That line comes from #4519's commit rather than yours.

In the same sentence, `odd n > 4381`: the secondary literature states Novikov–Adian inclusively (`n ≥ 4381`, e.g. [Encyclopedia of Mathematics](https://encyclopediaofmath.org/wiki/Burnside_group)), but the [Wikipedia page](https://en.wikipedia.org/wiki/Burnside_problem) cited in the module header does write `> 4381` — so that one is a choice between the primary sources and the cited one.

**2. `burnside_problem_II` classifies all of `ℕ × ℕ`,** whereas Burnside problem II asks "for which *positive* integers m, n". The extra columns are not vacuous: at `n = 0` the relator set `Set.range fun w => w ^ 0` is `{1}` and its normal closure is `⊥` (`simp` closes that), so `freeBurnsideGroup m 0 ≃* FreeGroup (Fin m)` and `B(1, 0)` is infinite, while `B(0, n)` is a subsingleton for every `n`. Both columns are decided, so no open content is added, but the answer slot now has to describe them and stops being literally the answer to the cited problem. A guard keeps the type unchanged:

```lean
theorem burnside_problem_II :
    {p : ℕ × ℕ |
      0 < p.1 ∧ 0 < p.2 ∧ Finite (freeBurnsideGroup p.1 p.2)} = answer(sorry) := by
  sorry
```

the same shape as `green_22.variants.moreira_infinite` in `GreensOpenProblems/22.lean`. (A `{n : ℕ // 0 < n} × {n : ℕ // 0 < n}` version elaborates too, but it puts coercions in the statement and makes the answer a set of subtype pairs.)

**3.** The `burnside_problem_II` docstring lists `B(m, n)` as finite "for `m = 1`", which fails at `n = 0` for the same reason. `finite_freeBurnsideGroup_one` already carries `0 < n`, so only the docstring needs the `n ≥ 1`.

**4. Optional API.** `Group.FG (freeBurnsideGroup m n)` does not synthesise — `QuotientGroup.fg` never fires through the non-reducible `PresentedGroup` — even though finite generation is what connects `freeBurnsideGroup` to `bounded_burnside_problem`. One line gives it:

```lean
instance (m n : ℕ) : Group.FG (freeBurnsideGroup m n) :=
  Group.fg_of_surjective (PresentedGroup.mk_surjective _)
```

**5. Optional, on status.** `665` is the right bound to carry as `research solved`, but it is no longer advertised as the best known: Adian announced odd `n ≥ 101` in [Proc. Steklov Inst. Math. 289 (2015)](https://doi.org/10.1134/S0081543815040045) with the full proof deferred to *Russian Math. Surveys* and not published since, and Atkarskaya–Rips–Tent claim odd `n ≥ 557` in [arXiv:2303.15997](https://arxiv.org/abs/2303.15997) (v5, 2024, still a preprint). A clause recording those would make the `665` choice read as deliberate.

Everything else checked out for me: the presented-group definition, both proofs, the `answer(False)` polarity, the finite cases `n ∈ {1, 2, 3, 4, 6}` and `m = 1`, the odd-`665` and `8000` variants (Adian for odd, Lysënok for even), `B(2, 5)` as the smallest open case, and the bibliography.

PR #5436, opened 2026-09-10
Green 14: the AKS14 lower bounds are proved, the exact values are conjectured

mo271:
The twenty statements `W_3_t_lower` for 20 ≤ t ≤ 39 were tagged `research open` and described as conjectured. In Ahmed–Kullmann–Snevily (AKS14) every one of them is a proved lower bound, certified by an explicit good partition in Appendix A (Sections "Certificates"). What AKS14 conjectures, for 20 ≤ t ≤ 30 only, is that the Table 2 bounds are the exact values; for 31 ≤ t ≤ 39 they say only that the Table 3 bounds can probably be improved.

Mark all twenty lower bounds `research solved` with `answer(True)`, and add the eleven conjectured exact values W(3, t) = v for 20 ≤ t ≤ 30 as `W_3_t_eq` with `answer(sorry)`, `research open`. Values checked against Tables 2 and 3 of the arXiv source of AKS14.

Fixes #4854 (reported by @KitaKen1).

PR #5445, opened 2026-09-10
fix(GreensOpenProblems/14): the exponential lower bounds were contentless, and the polynomial question has a known answer

j2d9w5xtjn-png:

`green_14_lower_bound_green` and `green_14_lower_bound_hunter` quantify `∃ c : ℝ` with no sign
condition. At `c = 0` the left-hand side is `Real.exp 0 = 1`, and `1 =O[atTop] fun r => (W 3 r : ℝ)`
holds, so neither declaration said anything about superpolynomial growth. Both now require `0 < c`.
`green_14_upper_bound_kelley_meka` is deliberately left alone: for an *upper* bound a small
constant makes the claim harder, not trivially true.

`green_14_polynomial` asked whether `W(k, r)` grows polynomially for fixed `k ≥ 4`, and was
`research open`. Excluding `k = 3` does not avoid the known answer. `Set.IsAPOfLength k` forces
`d ≠ 0` for `k ≥ 2`, so a red `k`-term progression contains a red 3-term one and
`mixedMonoAPGuaranteeSet k r ⊆ mixedMonoAPGuaranteeSet 3 r`; both sets are non-empty by van der
Waerden's theorem, which is what rules out the `sInf ∅ = 0` degeneracy, so `W(3, r) ≤ W(k, r)`. A
polynomial bound at `k = 4` would give one at `k = 3`, contradicting `green_14_polynomial_k_eq_3`
in the same file. The declaration becomes `answer(False)`, `research solved`, with that argument
in its docstring.

### How this was checked

on the other eight.

---

### DIFF
--- a/FormalConjectures/GreensOpenProblems/14.lean
+++ b/FormalConjectures/GreensOpenProblems/14.lean
@@ -61,11 +61,15 @@ noncomputable def W (k r : ℕ) : ℕ := sInf (mixedMonoAPGuaranteeSet k r)
 Is $W(k, r)$ a polynomial in $r$, for fixed $k$?
 
 We formulate this as asking if $W(k, r)$ has polynomial growth in $r$.
-We know it is not the case for $k = 3$ [Gr21, p.3].
+The answer is no, already for $k = 4$. Every red $k$-term progression with $k \ge 3$ contains a
+red $3$-term progression, so `mixedMonoAPGuaranteeSet k r ⊆ mixedMonoAPGuaranteeSet 3 r`, and
+both sets are nonempty by van der Waerden's theorem, so $W(3, r) \le W(k, r)$. Polynomial
+growth for one $k \ge 4$ would therefore force polynomial growth for $k = 3$, contradicting
+`green_14_polynomial_k_eq_3` [Gr21, p.3].
 -/
-@[category research open, AMS 5 11]
+@[category research solved, AMS 5 11]
 theorem green_14_polynomial :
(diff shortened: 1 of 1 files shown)


mo271:
fix for Green14 is already in #5436

more generally: I think it would make review easier if there was one pull request per fixed misformalisation

j2d9w5xtjn-png:
Split done — one pull request per misformalisation:

- this one, now Green 14 only
- #5451 rank-two Poisson conjecture
- #5452 Falconer at `d = 1`
- #5453 OEIS A069004 upper bound
- #5454 `graphSquare` joining different components

Each was already a separate commit, so every split fell on a commit boundary and I checked the
five together still reproduce this PR's original diff byte for byte. All five are rebased onto
`8faf8bcc`.

On #5436 — I think that one is a different part of the file, so this fix may still be needed.
#5436 changes the `W_3_t_lower` / `W_3_t_eq` table (its only hunk in `14.lean` is
`@@ -212,86 +212,132 @@`). This PR changes `green_14_lower_bound_green`,
`green_14_lower_bound_hunter` and `green_14_polynomial`, at lines 64–101. The two sets of
declarations are disjoint, and neither `0 < c` nor `green_14_polynomial` appears in #5436's diff.

The overlap is only the file, so whichever lands second will need a trivial rebase. Happy to close
this if you'd rather fold it into #5436, or to drop the `green_14_polynomial` half if only the
`0 < c` part is wanted.

Two more that are grouped rather than one-per-fix, in case you want them split the same way:
#5447 covers seven problem files (status corrections, not misformalisations), and #5446 covers ten
files but is a single systematic defect — every `answer(...)` slot whose type cannot mention the
binder it must depend on. Say the word and I will split either.

mo271 (approved):
Thanks, LGTM!

PR #5446, opened 2026-09-10
fix: answer slots that cannot depend on the inputs they must depend on

j2d9w5xtjn-png:

In the `with_auxiliary` answer-extraction mode the answer term is lifted to a closed top-level
definition, so an answer that mentions a theorem binder is a kernel error, and a slot typed
without that binder can never be filled in correctly.

Sweeping every problem file with `lake env lean -Dgoogle.answer=with_auxiliary` finds four
declarations that fail outright today: `erdos_506.variants.large_n` and `erdos_434.parts.i`, whose
answers are filled in and mention a theorem binder, and `erdos_596` and `green_5.variants.sl_two`,
whose answer *type* mentions one.

Each affected slot now applies a closed function to the binders, the idiom already used at
`OpenQuantumProblems/13.lean:562`. Besides the four above this covers `VoronovskajaTypeFormula`
(three slots, all missing the shape parameter `α`), Erdős 409 (`parts.i`, `variants.sigma`), 361,
321, 503, 506 (`erdos_506`, `variants.small_n`) and 1047. After the change a `with_auxiliary`
sweep over all 1268 problem files reports no free-variable error anywhere.

`erdos_361.asymptotic` is deliberately left alone: its answer is a `Θ`-class in `n` and may
legitimately not depend on `c`.

**Scope.** A slot still holding `answer(sorry)` does not fail to compile, because `sorry` is
closed — so the sweep sees only the subset of the defect where an answer has been filled in. A
subsequent exhaustive read of every `answer(` site found **twelve more slots with the same
defect**, which are *not* fixed here: `ErdosProblems/142.lean:37,56`, `319.lean:46`,
`692.lean:71`, `700.lean:72`, `975.lean:88`, `GreensOpenProblems/16.lean:47`, `24.lean:53`,
`37.lean:49,58,64,70,76`, `41.lean:69`. One of them, `erdos_700.parts.i`, looks **unsatisfiable as
written**: with a closed `Prop` answer, `f n = n / P n ↔ answer(sorry)` asserts that the left side
has the same truth value for every composite `n > 1`, and it does not — it holds at `n = 4`
(`f(4) = 2 = 4/P(4)`) and fails at `n = 8` (`f(8) = 2 ≠ 4 = 8/P(8)`). They are held back because
they were found after this change set was verified and deserve their own review. Tell me which you
prefer: I can fold them in here, or open a separate issue and PR for them.

### How this was checked

Every changed file elaborates under `lake env lean` and under
`lake env lean -Dgoogle.answer=with_auxiliary`. Sweeping all 1268 problem files in
`with_auxiliary` mode reports no free-variable error. That sweep does surface four *unrelated*
failures — `ErdosProblems/501`, `593`, `686` and `Wikipedia/JacobianConjecture` fail on `simp` or
`rw` in that mode because the answer term becomes an opaque top-level constant. None of those
files is touched here and all four fail identically on `main`, which I checked by running the
base versions; I mention them only so the sweep output is not mistaken for a regression.

`lake --wfail build` over the whole repository and `lake --wfail test` pass on the combined tree
holding all nine of these pull requests; they change disjoint files, so nothing here depends on
the other eight.

---

### DIFF
--- a/FormalConjectures/ErdosProblems/1047.lean
+++ b/FormalConjectures/ErdosProblems/1047.lean
@@ -124,7 +124,8 @@ a function of the degree of $f$.
 theorem erdos_1047.variants.max_non_convex_components (n : ℕ) :
     IsGreatest {k : ℕ | ∃ (f : ℂ[X]) (c : ℝ), f.Monic ∧ f.natDegree = n ∧ 0 < c ∧
       (componentsIn (sublevelSet f c)).ncard = (f.rootSet ℂ).ncard ∧
-      {t ∈ componentsIn (sublevelSet f c) | ¬ Convex ℝ t}.ncard = k} answer(sorry) := by
+      {t ∈ componentsIn (sublevelSet f c) | ¬ Convex ℝ t}.ncard = k}
+      ((answer(sorry) : ℕ → ℕ) n) := by
   sorry
 
 end Erdos1047
--- a/FormalConjectures/ErdosProblems/321.lean
+++ b/FormalConjectures/ErdosProblems/321.lean
@@ -42,7 +42,7 @@ $\sum_{n\in S} \frac{1}{n}$ are distinct for $S\subseteq A$. What is $R(N)$?
 -/
 @[category research open, AMS 11]
 theorem erdos_321 (N : ℕ) :
-    R N = answer(sorry) := by
+    R N = (answer(sorry) : ℕ → ℕ) N := by
   sorry
 
 /-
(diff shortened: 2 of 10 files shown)


Paul-Lez:
Oh I think that I know how to fix some of this - there's an issue with the answer elaborator that I've been meaning to fix for a while

j2d9w5xtjn-png:
Thanks. The failures this PR works around only occur with `google.answer = "with_auxiliary"`; in the default and `postpone` modes these answers can already use the binders. If you fix the elaborator, should an answer that depends on a binder be written directly, as `answer(... n ...)`? Then I will close this PR. Otherwise I will rebase it (it now conflicts with main, and the Voronovskaja part landed in #5833) and split it into one PR per file.

Paul-Lez:
Let me have a quick think about it and then get back to you!

PR #5448, opened 2026-09-10
fix(util): decide proof status from transitive dependencies, and keep the answer marker on a Prop-valued answer(sorry)

j2d9w5xtjn-png:

Two independent corrections to repository utilities. This pull request originally also carried a
set of `FormalConjectures/` documentation fixes; on review those were split out, and nothing here
touches `FormalConjectures/`.

### Proof-status checks followed only the immediate proof term

`scripts/extract_names.lean`, `CategoryLinter` and `FormalProofLinter` all decided whether a
declaration is proved with `info.value? … |>.any (!·.hasSorry)`. That looks only at the proof
body, so `theorem wrapper : P := admitted_helper` counts as proved. `OeisA237271.conjecture_3` is
a live example: its proof is `conjecture_2`, which is `sorry`, and the direct test reports it as
sorry-free. For a `research open` declaration of that shape, `check_category_warnings.py` would
turn the resulting metadata into a blocking `research_open_with_proof` CI failure, and the
conditional-proof linter would report an assumed hypothesis as discharged.

A new `ProblemAttributes.hasSorryFreeProof` uses `Lean.collectAxioms`, which reads pre-computed
axiom sets for imported declarations, so the walk stays inside the current module, and takes its
constants from the kernel environment, so it waits on a proof that is still elaborating. It
reports a declaration as established only when `sorryAx` is absent from the transitive closure of
its statement and its proof. All three call sites use it, and regression tests are added to both
linter test files and to `FormalConjecturesTest/Util/Attributes/Basic.lean`.

Note that `collectAxioms` walks the *statement* as well as the proof, so two different things make
it return `false`: a proof that depends on an admitted declaration, which is the defect above, and
a statement that mentions an admitted definition even when the proof is complete. Both are
defensible answers to "is this established", but they are different phenomena.

Repository-wide, 30 of the 5265 categorised declarations that `extract_names` exports flip to
"not proved", and none flips the other way: 27 because the proof depends on an admitted
declaration, and 3 (`OeisA87719.a_1` to `a_3`) because the statement mentions a definition built
from an admitted existence proof. By category the 30 are 12 `test`, 11 `research solved`, 4 `API`
and 3 `textbook` — no `research open` declaration changes. Two advisory CI counters move,
`test_without_proof` 53 → 65 and `api_without_proof` 13 → 17, while the blocking
`research_open_with_proof` stays at 0.

**Not done.** The metadata still does not *distinguish* a syntactically complete wrapper from a
proof with no admitted dependencies: `extract_names` exports a single `hasSorryFreeProof` boolean,
so a consumer cannot tell those two states apart. A second field would close that. Left as a
follow-up rather than widened here.

### The answer marker was dropped on a `Prop`-valued `answer(sorry)`

`defaultTargets` omits `FormalConjecturesAnswerPostpone`, and CI extracts metadata after the
ordinary build, so `extract_names` reads artifacts elaborated in `always_true` mode. In that mode
a `Prop`-valued `answer(sorry)` returned a bare `True` carrying no answer annotation, while a
non-`Prop` `answer(sorry)` and an explicit `answer(True)` both kept theirs. So `findAnswerExprs`
saw nothing and `answerKinds` could not distinguish a `Prop`-valued answer from a declaration with
no answer at all.

The `Prop` branch now annotates too, matching the other two. `CategoryAnswerLinter`'s docstring is
corrected: its check does have to stay syntactic, but because `answer(sorry)` and `answer(True)`
elaborate to the same term, not because the annotation is missing.

Checked by elaborating four declarations against the old and the new `Answer.lean`, rebuilding
only `FormalConjecturesUtil` in between: a `Prop`-valued `answer(sorry)` goes from `[]` —
indistinguishable from having no answer — to `["Prop"]`, while `answer(True)`, a non-`Prop`
`answer(sorry)` and a declaration with no answer are all unchanged. The two cases are pinned by
new `#guard_msgs` tests in `FormalConjecturesTest/Util/Answer.lean`.

The alternative would be to build `FormalConjecturesAnswerPostpone` before the extraction step in
`.github/workflows/build-and-docs.yml`. That is behaviour-preserving but roughly doubles the Lean
time in CI, since the two libraries share globs and differ only in a Lean option. Happy to switch
if you would rather not touch `Answer.lean`.

### How this was checked

`lake --wfail build FormalConjecturesForMathlib FormalConjecturesUtil`, `lake --wfail test`, and
`lake --wfail build` over the whole repository, followed by `lake exe extract_names` and
`scripts/check_category_warnings.py`. The repository-wide figures above were measured on a build
with the `FormalConjecturesAnswerPostpone` library disabled: it shares its globs with
`FormalConjectures` and writes to the same olean paths, so an ordinary local build leaves a
mixed-mode cache that `extract_names` then reads. CI is unaffected — it builds only the default
targets — but a local reproduction needs that step.

---

### DIFF
--- a/FormalConjecturesTest/Util/Answer.lean
+++ b/FormalConjecturesTest/Util/Answer.lean
@@ -73,6 +73,29 @@ set_option google.answer "always_true"
 theorem this_works : (answer(sorry) : Prop) := by
   trivial
 
+/-- A `Prop`-valued `answer(sorry)` elaborates to `True` in this mode but keeps its answer
+annotation, so metadata extracted here can still tell it apart from a statement that has no
+answer at all. -/
+theorem prop_sorry_keeps_its_marker : answer(sorry) ↔ 1 + 1 = 2 := by
+  sorry
+
+theorem no_answer_has_no_marker : 1 + 1 = 2 := by
+  rfl
+
+/-- info: 1 -/
+#guard_msgs in
+#eval show Lean.CoreM Nat from do
+  let env ← Lean.getEnv
+  let some ci := env.find? ``prop_sorry_keeps_its_marker | return 0
+  return (Google.findAnswerExprs ci.type).size
+
+/-- info: 0 -/
+#guard_msgs in
+#eval show Lean.CoreM Nat from do
+  let env ← Lean.getEnv
+  let some ci := env.find? ``no_answer_has_no_marker | return 0
+  return (Google.findAnswerExprs ci.type).size
+
 end AlwaysTrue
 
 section FindAnswerExprTests
--- a/FormalConjecturesTest/Util/Attributes/Basic.lean
+++ b/FormalConjecturesTest/Util/Attributes/Basic.lean
@@ -202,3 +202,38 @@ info: 0 General and overarching topics
 -/
 #guard_msgs in
 #AMS
+
+section HasSorryFreeProof
+
+/-! `ProblemAttributes.hasSorryFreeProof` follows the declarations that a statement and a proof
+use, and reports a declaration with no proof term at all as not proved. The declarations used
+here are Lean's own axioms; the repository does not add any of its own. -/
+
+private theorem an_admitted_thm : 1 + 1 = 3 := by sorry
+private theorem a_wrapper_over_an_admitted_thm : 1 + 1 = 3 := an_admitted_thm
+private theorem a_real_proof : 1 + 1 = 2 := rfl
+
+/-- info: true -/
+#guard_msgs in
+#eval ProblemAttributes.hasSorryFreeProof (m := Lean.CoreM) ``a_real_proof
+
+-- An `axiom` has no proof term, so it is not proved even though `sorryAx` is absent.
+/-- info: false -/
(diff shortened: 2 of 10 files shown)


mo271:
Thanks. I think it would be good to split up this pull reqeust into two: one for the files in FormalConjectures/ and everything else

j2d9w5xtjn-png:
Done — split at #5449.

The `FormalConjectures/` files were already a single commit, so the split fell exactly on a
commit boundary and no file crosses it:

- this PR keeps the two utility commits: `FormalConjecturesUtil/`, `FormalConjecturesTest/` and
  `scripts/extract_names.lean` (10 files);
- #5449 has the documentation commit: 8 files under `FormalConjectures/` (5 OEIS, 3 Wikipedia),
  docstrings only apart from removing one of two identical `not_isRegularPrime_37` declarations.

I confirmed the two halves reproduce this PR's original diff byte for byte. The two commits here
are unchanged, so the existing review still applies to them; the force-push only dropped the
documentation commit off the top.

Title and description updated to match the narrower scope.

PR #5449, opened 2026-09-10
docs(FormalConjectures): cross-references, a duplicate declaration, and copy fixes

j2d9w5xtjn-png:

Documentation-only changes to eight problem files. No statement, hypothesis or `category`
attribute changes, except that `Wikipedia/RegularPrimes` loses one of two identical declarations.

- `OEIS/51903`: `conjecture1` is equivalent to Lehmer's totient problem; the docstring now says so
  and gives the argument, cross-referencing `LehmerTotient.lehmer_totient` and
  `Erdos828.erdos_828.variants.lehmer_conjecture`.
- `OEIS/3162`: the 1974 Monthly item is Gould's **proposal** of Problem E2384, not a solution to
  it — A003162 itself says "Gould (1974) proposed the problem". Relabelled. The audit that raised
  this suggested citing Gould & Clary 1976 as the solution instead; I could not reach that item in
  any accessible source, so I did not add it on faith.
- `OEIS/323557`, `OEIS/325046`: the two sequences differ only by the sign `(−1)^j` on each
  summand, so they agree modulo 2 and the two `odd_a_implies_pronic` theorems are equivalent.
  Cross-referenced in both. A shared lemma would need a common import and problem files import
  only `FormalConjecturesUtil`, so this is a cross-reference only.
- `OEIS/2326`: `conjecture2` is equivalent to `conjecture1`, not strictly stronger; the
  lifting-the-exponent argument is recorded.
- `Wikipedia/RegularPrimes`: `not_isRegularPrime_37_first` and `_second` were the same declaration
  with the same docstring. One is kept, renamed `not_isRegularPrime_37`.
- `Wikipedia/EulerBrick`: "Shapirov" → "Sharipov" (arXiv:1108.5348 is by Ruslan Sharipov).
- `Wikipedia/ModularityConjecture`: "also know as" → "also known as"; "of a modular forms" → "of a
  modular form".

### How this was checked

holding all nine of these pull requests; they change disjoint files, so nothing here depends on
the other eight.

---

### DIFF
--- a/FormalConjectures/OEIS/2326.lean
+++ b/FormalConjectures/OEIS/2326.lean
@@ -82,6 +82,12 @@ then $a((p^{k+1}-1)/2) = p \cdot a((p^k-1)/2)$.
 Computer testing of this generalized conjecture shows that there is no counterexample for $k$
 and $p$ both up to 1000.
 - [Ahmad J. Masad](https://oeis.org/wiki/User:Ahmad_J._Masad), Oct 17 2020
+
+The two are in fact equivalent. Its $k = 2$ instance is `conjecture1`. Conversely, write
+$d = \operatorname{ord}_{p^2}(2)$. `conjecture1` for $p$ says
+$v_p(2^d - 1) = 2$, and lifting the exponent for the odd prime $p$ then gives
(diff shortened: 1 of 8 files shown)


mo271 (approved):
Thanks, LGTM!

Paul-Lez (approved):
Thanks!

PR #5450, opened 2026-09-10
feat(Paper): No loop conjectures in homological algebra

ymonbru:
We add different statements in homological algebra called "no loop" conjecture.

Let $A$ be an Artin algebra over a commutative Artinian ring $R$. The No Loop Conjectures are a family of statements relating $\text{Ext}^1_A(S,S)$ for simple $A$-modules $S$ to homological properties of $S$ and $A$.

Co-authored-by: Wassilij Gnedin (@wgnedin)


FormalConjectures/Paper/NoLoop.lean
```diff
@@ -0,0 +1,163 @@
...
+The *Strong No Loop Conjecture*:
+For any simple $A$-module $S$
+$$
+\operatorname{Ext}^1_A(S,S) \neq 0
+\quad \Longrightarrow \quad
+\operatorname{pd} S = \infty
+$$
+-/
+@[category research open, AMS 16 18]
+theorem strong_no_loop_conjecture : HasFirstSelfExt S → projectiveDimension S = ⊤ := by
```
Paul-Lez: Small nit: since you're using `type_of% strong_no_loop_conjecture` a bunch below, what do you think of defining a `StrongNoLoopConjecture` predicate, and then stating `strong_no_loop_conjecture` in terms of that and then using the same predicate below?

ymonbru: resolved in discussion

FormalConjectures/Paper/NoLoop.lean
```diff
@@ -0,0 +1,163 @@
...
+* The *Extreme No Loop Conjecture* was formulated by S. Liu and J.-P. Morin in
+[The strong no loop conjecture for special biserial algebras,
+Proc. Amer. Math. Soc. 132 (2004), 3513-3523](https://doi.org/10.1090/S0002-9939-04-07512-4).
+It was also called Extension Conjecture later on.
+* The *Strong No Loop Conjecture* was solved for finite-dimensional algebras
+over algebraically closed fields by K. Igusa, S. Liu and C. Paquette in
+[A proof of the strong no loop conjecture](https://arxiv.org/abs/1103.5361),
+published as [Adv. Math. 228 (2011), 2731-2742](https://doi.org/10.1016/j.aim.2011.06.042).
+* The *No Loop Conjecture* was solved assuming the same setup as above by K. Igusa
+[Notes on the no loop conjecture,
```
mo271:
```suggestion
[Notes on the no loops conjecture,
```

FormalConjectures/Paper/NoLoop.lean
```diff
@@ -0,0 +1,163 @@
...
+  type_of% (strong_no_loop_conjecture (R := R) (A := A)) → type_of% (no_loop_conjecture (R := R) (A := A)):= fun h S sS hE1 n => ⟨S, by
+  rw [h S hE1 ]
+  exact WithBot.LT.coe_lt_coe <| ENat.natCast_lt_top n⟩
+
+/--
+For finite-dimensional algebras over algebraically closed fields,
+the verified Strong No Loop Conjecture establishes the No Loop Conjecture.
+-/
+@[category test, AMS 16 18]
+lemma no_loop_conjecture.variant.alg_closed
```
mo271:
```suggestion
lemma no_loop_conjecture.variants.alg_closed
```

FormalConjectures/Paper/NoLoop.lean
```diff
@@ -0,0 +1,163 @@
...
+Throughout, `A` is an Artin `R`-algebra, that is, `R` is a commutative Artinian ring,
+and `A` is an `R`-algebra such that `A` is finitely generated as `R`-module.
+Let `S` be a simple `A`-module. -/
+variable {R : Type u} {A : Type v} [CommRing R] [IsArtinianRing R] [Ring A]
+variable [Algebra R A] [Module.Finite R A] (S : ModuleCat.{v} A) [Simple S]
+
+abbrev HasFirstSelfExt := ¬ Subsingleton (Ext S S 1)
+
+variable (A) in
+abbrev HasInfiniteGlobalDimension := ∀ n : ℕ, ∃ M : ModuleCat.{v} A, projectiveDimension M > n
```
mo271: can these have docstring (or potentially be inlined?)

FormalConjectures/Paper/NoLoop.lean
```diff
@@ -0,0 +1,163 @@
...
+namespace NoLoopConjectures
+
+/-
+Throughout, `A` is an Artin `R`-algebra, that is, `R` is a commutative Artinian ring,
+and `A` is an `R`-algebra such that `A` is finitely generated as `R`-module.
+Let `S` be a simple `A`-module. -/
+variable {R : Type u} {A : Type v} [CommRing R] [IsArtinianRing R] [Ring A]
+variable [Algebra R A] [Module.Finite R A] (S : ModuleCat.{v} A) [Simple S]
+
+abbrev HasFirstSelfExt := ¬ Subsingleton (Ext S S 1)
```
mo271: potentially also just using `Nontrivial (Ext S S n)` might be good?

mo271 (commented):
Thanks, looks great, jus a few minor comments

j2d9w5xtjn-png (commented):
Reviewed at `3f76000e`. I built the file in a worktree against Mathlib v4.33.1 and checked the statements against the sources. **The three conjecture statements are faithful and the file builds clean.** Everything below is documentation and coverage; none of it is a correction to the Lean.

_Disclosure: this review was produced with Claude Code._

## What I checked

- `lake --wfail build FormalConjectures.Paper.NoLoop` passes with no linter warnings. `strong_no_loop_of_extreme_no_loop` and `no_loop_of_strong_no_loop` are sorry-free: `#print axioms` gives only `propext`, `Classical.choice`, `Quot.sound`.
- All four cited DOIs resolve to the stated title, authors, volume, pages and year.
- `extreme_no_loop_conjecture` matches Extension Conjecture 2.6 of Igusa–Liu–Paquette exactly: *"If Ext¹(S,S) is non-zero, then Extⁱ(S,S) is non-zero for infinitely many integers i."* The "infinitely many" form is easy to mis-render as "for all i ≥ 1"; this file gets it right.
- Semantics all line up: `Simple S` in `ModuleCat A` is `IsSimpleModule A S`; `Projective M` is `Module.Projective A M`, so `projectiveDimension` is the module-theoretic one; `Ext S S 1 : Type v` carries an `AddCommGroup`, so `Nontrivial` really is `Ext¹(S,S) ≠ 0`; and `S` simple gives `S ≇ 0`, so `projectiveDimension S = ⊤` is exactly infinite projective dimension (`⊥` is reachable only for the zero object).
- `Algebra R A` forces `algebraMap R A` into the centre of `A`, which is what "Artin `R`-algebra" requires.
- `HasInfiniteGlobalDimension` is neither vacuous nor trivial — I checked `¬ HasInfiniteGlobalDimension k` for a division ring. Quantifying over `ModuleCat.{v} A` loses nothing: global dimension is already the supremum of the projective dimensions of the cyclic modules, and those live in `Type v`.
- ILP use right modules where this file uses left modules. The conjectures quantify over all artin algebras and `A ↦ Aᵒᵖ` preserves that class, so the two forms are equivalent.

## A. The Igusa/Lenzing sentence is not what the source says

The module docstring says the No Loop Conjecture "was solved assuming the same setup as above" (finite-dimensional over an algebraically closed field) by Igusa, and "follows also from earlier work of H. Lenzing". Igusa–Liu–Paquette, introduction:

> The no loop conjecture was first explicitly established for artin algebras of global dimension two; see [6]. For finite dimensional **elementary** algebras, as shown in [11], this can be easily derived from an earlier result of Lenzing on Hochschild homology in [14].

Three differences:

1. The scope is **elementary** algebras, not "over an algebraically closed field". The two agree up to Morita equivalence, so the current claim is not false, but "elementary" is the source's scope and it is also what makes this file's own `kQ/I` Remark apply.
2. ILP present Igusa's contribution as *deriving* the result from Lenzing. "Follows also from" reads as an independent second proof.
3. The first explicit case is not cited: E. L. Green, W. H. Gustafson and D. Zacharia, [Artin rings of global dimension two, J. Algebra 92 (1985), 375-379](https://doi.org/10.1016/0021-8693(85)90127-9).

## B. Only half of the cited theorem is stated

Igusa–Liu–Paquette, Theorem 2.2:

> Let Λ be a finite dimensional algebra over a field k, and let S be a simple Λ-module which is one dimensional over k. If S has finite projective **or injective** dimension, then Ext¹_Λ(S,S) = 0.

The injective half is a separate statement and is missing. It states cleanly (this compiles):

```lean
@[category research solved, AMS 16 18]
theorem strong_no_loop_conjecture.variants.alg_closed_inj {k : Type u}
    [Field k] [IsAlgClosed k] [Algebra k A] [Module.Finite k A] :
    Nontrivial (Ext S S 1) → injectiveDimension S = ⊤ := by
  sorry
```

The Extreme conjecture implies it too, by the dual of the existing argument (also checked):

```lean
  intro h S sS neZS
  by_contra!
  rcases (injectiveDimension_ne_top_iff _).1 this with ⟨m, hm⟩
  rcases h S neZS m with ⟨n, hn⟩
  exact hn.2 <| HasInjectiveDimensionLT.subsingleton (X := S) (m + 1) n (Nat.succ_le_of_lt hn.1) S
```

The sharper form of Theorem 2.2 — arbitrary field `k`, `S` one-dimensional over `k` — needs a `Module k ↑S` instance, which does not synthesize for a bundled `ModuleCat A` object. Probably not worth the scaffolding.

One related note: ILP give the elementary scope here too, in the Remark after Theorem 2.2 — *"The preceding theorem establishes the strong no loop conjecture for finite dimensional elementary algebras, and hence for finite dimensional algebras over an algebraically closed field."* The current `strong_no_loop_conjecture.variants.alg_closed` matches the paper's abstract, so it is faithful as it stands.

## C. The known cases of the Extreme conjecture are not recorded

Immediately after Extension Conjecture 2.6: *"It remains open except for monomial algebras and special biserial algebras; see [7, 15]."* Reference [15] is Liu–Morin, already cited; [7] is E. L. Green, Ø. Solberg and D. Zacharia, [Minimal projective resolutions, Trans. Amer. Math. Soc. 353 (2001), 2915-2939](https://doi.org/10.1090/S0002-9947-01-02687-3). STATEMENTS.md asks a reviewer to check "every variant and special case", so one sentence in the docstring would close this.

## D. The Strong conjecture is due to Zacharia

> the strong no loop conjecture, which is due to Zacharia, strengthens this to state that a vertex in the extension quiver admits no loop if it has finite projective dimension

The reference list says who posed the Extreme version but not the Strong one.

## Smaller points

1. **No Lean build ran on this PR.** The only check-runs on `3f76000e` are `check-changes`, `labeler` and `cla/google`; the "Build project" job never fired, and this branch's `.github/workflows/build-and-docs.yml` is 83 lines behind `main`. Merging `main` should fix it. It does build locally.
2. **`R` silently disappears** from `strong_no_loop_of_extreme_no_loop` and `no_loop_of_strong_no_loop`. Because `type_of%` yields a statement that does not mention `R` and there is no `include R`, both elaborate to `∀ {A : Type u} [Ring A], …` — arbitrary rings, no Artin hypothesis. That is correct and stronger, but `(R := R)` makes a reader expect otherwise. Worth a docstring line.
3. **`¬ Subsingleton` vs `Nontrivial`** are still mixed inside one statement. Using `Nontrivial` throughout works if line 143 becomes `exact not_subsingleton_iff_nontrivial.2 hn.2 <| …`.
4. **`HasInfiniteGlobalDimension` has no API test**, which CONTRIBUTING asks for on a bespoke definition. Both of these compile:

```lean
@[category test, AMS 16 18]
lemma hasInfiniteGlobalDimension_of_eq_top (M : ModuleCat.{v} A)
    (hM : projectiveDimension M = ⊤) : HasInfiniteGlobalDimension A :=
  fun n => ⟨M, hM ▸ WithBot.LT.coe_lt_coe (ENat.natCast_lt_top n)⟩

@[category test, AMS 16 18]
lemma not_hasInfiniteGlobalDimension_divisionRing {k : Type v} [DivisionRing k] :
    ¬ HasInfiniteGlobalDimension k := by
  rintro h
  obtain ⟨M, hM⟩ := h 0
  exact absurd ((projectiveDimension_le_iff M 0).2 inferInstance) (not_le.2 hM)
```

5. Its docstring describes global dimension rather than the predicate, and is missing a space in ``​`A`is``. Something like "`A` has infinite global dimension: the projective dimensions of `A`-modules are unbounded."
6. Lines 77, 116, 130, 138 and 150 exceed 100 columns.
7. `no_loop_conjecture.variants.alg_closed` is tagged `@[category test]`, but its docstring presents it as a real theorem; `research solved` may fit better. Maintainer's call.
8. The PR body's `Co-authored-by: Wassilij Gnedin (@wgnedin)` is not a parseable trailer — GitHub wants `Co-authored-by: Name <email>` — so the co-authorship will not register on a squash merge.

PR #5451, opened 2026-09-10
fix(Arxiv/math.0608009): the rank-two Poisson conjecture is false

j2d9w5xtjn-png:

`poisson_conjecture.variants.dimension_two` positively asserted `PoissonConjectureFor K 2` for
every characteristic-zero field. It is false. Long,
[arXiv:2608.23777](https://arxiv.org/abs/2608.23777) (July 2026), gives explicit
`R, T, D, S ∈ ℚ[x, q, p, z]` with `{R, D} = {T, S} = 1` and the other four generator brackets
zero, so `(x, q, p, z) ↦ (R, T, D, S)` is a Poisson endomorphism of `P₂(K)`. It is not
surjective: `(0, 0, 1/24, −1/8)` and `(1, 2/3, 247/96, −89/64)` have the same image, so every
polynomial in the image agrees at those two points, while `X₁` does not. Coefficients and points
are rational, so this descends to every field of characteristic zero.

I re-derived the six bracket identities and the point collision in exact rational arithmetic
against this repository's own `poissonBracket` convention, and separately checked a Lean proof of
`¬ PoissonConjectureFor ℚ 2` built from those polynomials: it typechecks against this module and
`#print axioms` gives `[propext, Classical.choice, Quot.sound]`.

One convention note, since it inverts the reading of the paper. Long's bracket is the negative of
this repository's — his `H_F = F_p ∂_x + F_z ∂_q − F_x ∂_p − F_q ∂_z`, against
`poissonBracket f g = ∑ᵢ (∂_{inl i} f · ∂_{inr i} g − ∂_{inr i} f · ∂_{inl i} g)` here — so his
Theorem 5.2 `{D, R} = {S, T} = 1` is `{R, D} = {T, S} = 1` in this file, which is how the
docstring states it. His Φ sends `x ↦ R, q ↦ T, p ↦ D, z ↦ S`, matching `X₁ … X₄` under the
file's own `Sum.inl i = Xᵢ`, `Sum.inr i = X_{i+n}` indexing note.

The declaration is negated and marked `research solved`. `variants.dimension_ge_three` becomes
`variants.dimension_ge_two` with hypothesis `2 ≤ n`, and the module docstring and reference list
are updated. The docstring states explicitly that this settles nothing lower in the `[AvdE07]`
chain, which propagates falsity only upwards: `¬PC₂` gives `¬JC₄`, already known, and leaves
`DC₂`, `JC₂`, `PC₁` and `DC₁` open.

### How this was checked

on the other eight.

---

### DIFF
--- a/FormalConjectures/Arxiv/math.0608009/PoissonConjecture.lean
+++ b/FormalConjectures/Arxiv/math.0608009/PoissonConjecture.lean
@@ -29,6 +29,8 @@ import FormalConjecturesUtil
   by *Hyman Bass, Edwin H. Connell, David Wright*, Bull. Amer. Math. Soc. 7 (1982), 287–330.
 - [Alp26] [Counterexample to the Jacobian conjecture](https://x.com/__alpoge__/status/2079028340955197566)
   by *Levent Alpöge* (2026), disproving the Jacobian conjecture in dimension $3$
+- [Lon26] [An Explicit Counterexample to the Rank-Two Poisson Conjecture](https://arxiv.org/abs/2608.23777)
+  by *Christopher D. Long* (2026), disproving $PC_2$
 
 The Poisson Conjecture $PC_n$ ([AvdE07], Notations 5) asserts that, over a field of
 characteristic zero, every endomorphism of the `n`-th canonical Poisson algebra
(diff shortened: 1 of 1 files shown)


FormalConjectures/Arxiv/math.0608009/PoissonConjecture.lean
mo271, line 126:
```suggestion
degree-$7$ counterexample map $F$ of [Alp26] (whose Jacobian determinant is the constant
```
 the Jacobian counterexample map of Alpöge a degree-6 map; its first
    coordinate contains the monomial X³Y³W, so the map has degree 7 (the
    other coordinates have degrees 6 and 4).

mo271 (approved):
Thanks, LGTM!

PR #5453, opened 2026-09-10
fix(OEIS/69004): the upper bound in conjecture2 fails at n = 512720

j2d9w5xtjn-png:

`conjecture2` conjoins the upper bound `π(n) ≥ a(n)` with a lower bound and three equalities. The
upper bound fails at `n = 512720 = 2⁴·5·13·17·29`, where `a(n) = 42666` and `π(n) = 42493`;
A069004 tabulates `a` only to `n = 10⁴`. I computed `a(512720)` two independent ways — a
quadratic-residue sieve that uses no primality test, and a deterministic Miller–Rabin scan over
all `s < n` — and they agree.

`conjecture2` is negated and marked `research solved`, with the counterexample in its docstring.
The sharp fact is recorded as `conjecture2.variants.upper_bound_false`, and the lower bound and
the three equalities are kept as the still-open `conjecture2.variants.without_upper_bound`.
Minimality of `512720` is not claimed.

### How this was checked

on the other eight.

---


mo271 (approved):
Thanks, LGTM!

PR #5454, opened 2026-09-10
fix(ForMathlib/SimpleGraph): graphSquare joined vertices in different components

j2d9w5xtjn-png:

`graphSquare` used `G.dist u v ≤ 2`, and Mathlib's natural-valued `SimpleGraph.dist` is `0` for
unreachable pairs, so distinct vertices in different components became adjacent: the square of the
edgeless graph on two vertices was the complete graph. Adjacency now uses `G.edist`, which is `⊤`
there. The one downstream sanity check in `WrittenOnTheWallII/GraphConjecture146.lean` moves from
`dist_le` to `edist_le`, and a regression test proving
`graphSquare (⊥ : SimpleGraph (Fin 2)) = ⊥` is added. `conjecture146` assumes connectedness and is
unaffected.

**Not in this PR.** The other distance definitions in the same file — `distToSet`, `ecc`,
`distMin`, `eccSet`, `distMaxSet`, `distavg`, `averageDistance`, `distEven` and the even-distance
family — share the `dist`-is-0-when-unreachable pattern. Every current consumer hypothesises `G.Connected`, so no
conjecture in the repository is presently wrong because of them, and only `graphSquare` is changed
here. Happy to convert the rest in a follow-up if you would like them consistent.

### How this was checked

on the other eight.

---

### DIFF
--- a/FormalConjectures/WrittenOnTheWallII/GraphConjecture146.lean
+++ b/FormalConjectures/WrittenOnTheWallII/GraphConjecture146.lean
@@ -81,13 +81,13 @@ theorem conjecture146 (G : SimpleGraph α) [DecidableRel G.Adj] (h : G.Connected
 -- Sanity checks
 
 /-- In `graphSquare G`, adjacent vertices in $G$ are also adjacent
-($\mathrm{dist} \le 1 \le 2$). -/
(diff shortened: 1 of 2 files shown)


mo271 (approved):
Thanks, LGTM!

Issue #1792, opened 2026-01-20
Farrel Jones Conjecture


franzhusch:
### What is the conjecture

For a group $G$, the assembly maps
$$H_n^G(EG; K) \to K_n(\mathbb{Z}[G])$$
and
$$H_n^G(EG; L) \to L_n(\mathbb{Z}[G])$$
are isomorphisms in the $K$-theory and $L$-theory of group rings.

**Sources:**
- https://en.wikipedia.org/wiki/Farrell%E2%80%93Jones_conjecture, https://arxiv.org/abs/math/0703548, https://www.mpim-bonn.mpg.de/node/10826, https://link.springer.com/chapter/10.1007/978-3-319-43674-6_1

### Prerequisites needed

**Formalizability Rating:** 4/5 (as of 2026-01-20)

The Farrell-Jones conjecture requires substantial infrastructure beyond current Mathlib. While basic definitions for K-theory and group theory exist, the conjecture involves: (1) equivariant K-theory and L-theory with respect to families of subgroups, (2) equivariant homology groups and classifying spaces for families, (3) assembly maps that require cohomological machinery. These are advanced topics in algebraic topology and algebraic K-theory that would require significant new theory development beyond what is currently in Mathlib.

### [AMS categories](https://github.com/google-deepmind/formal-conjectures/labels?q=ams-)

* ams-19
* ams-20
* ams-55

### Choose either option

- [ ] I plan on adding this conjecture to the repository
- [x] This issue is up for grabs: I would like to see this conjecture added by somebody else

Created by AI, edited and reviewed by me.

PR #5353, opened 2026-09-08
Kaplansky: use ordinary torsion-freeness, not IsMulTorsionFree

mo271:
Kaplansky's conjectures concern torsion-free groups: the identity is the only element of finite order. The file used Mathlib's `IsMulTorsionFree`, which requires `a ↦ a ^ n` to be injective for all `n ≠ 0` (uniqueness of roots). The two agree for commutative groups but `IsMulTorsionFree` is strictly stronger in general, and the Promislow group P fails it: `(a²b)² = b²` with `a²b ≠ b`. So `promislow_group_is_torsionfree` was false, the two `counter_unit_conjecture` statements asserted more than the cited results, and the two open conjectures were stated only for a subclass of the intended groups.

The file originally used `Monoid.IsTorsionFree`, which Mathlib later deprecated in favour of `IsMulTorsionFree` (#24311) and removed; the toolchain bump silently strengthened the hypothesis. Mathlib currently has no predicate for ordinary torsion-freeness of noncommutative groups, so this reintroduces `Monoid.IsTorsionFree` (with `to_additive`) in ForMathlib, following the former Mathlib definition, together with the implication from `IsMulTorsionFree` and the equivalence for commutative groups.

identified by @tadamcz running an AI pipeline: https://tadamcz.com/fc-review-results/#/f/Wikipedia/Kaplansky

PR #5456, opened 2026-09-10
feat(Arxiv/math.0703548): add two algebraic consequences of the Farrell–Jones conjecture

### LINKED ISSUE #5455: Algebraic consequences of the Farrell–Jones conjecture: vanishing of $\widetilde K_0(\mathbb{Z}[G])$ and the Bass conjecture
### What is the conjecture

Two algebraic consequences of the Farrell–Jones conjecture, both stated in the introduction of
Bartels–Lück–Reich. Unlike the conjecture itself (#1792), each is expressible with the algebra
already in Mathlib.

**1. Vanishing of the reduced projective class group.** If $G$ is torsionfree then
$\widetilde K_0(\mathbb{Z}[G]) = 0$, equivalently every finitely generated projective
$\mathbb{Z}[G]$-module is stably free. This is the conclusion of Theorem 0.2 (ii) for the regular
ring $\mathbb{Z}$; that theorem assumes the K-theoretic Farrell–Jones conjecture for $G$ with
coefficients in $\mathbb{Z}$.

**2. Bass conjecture for commutative integral domains** (Conjecture 0.6). Let $R$ be a commutative
integral domain, $G$ a group and $g \in G$. Suppose the order $|g|$ is infinite, or that $|g|$ is
finite and not invertible in $R$. Then for every finitely generated projective $RG$-module $P$ the
value of the Hattori–Stallings rank $\mathrm{HS}_{RG}(P)$ at $(g)$ is zero. By Theorems 0.5 (ii)
and 0.7 this follows from the Farrell–Jones conjecture with coefficients in every field of prime
characteristic.

Known cases of Conjecture 0.6: Bass for $R = \mathbb{C}$ and $G$ linear, Linnell for
$|g| < \infty$ and $R = \mathbb{Z}$, Eckmann for $R = \mathbb{Q}$ and
$\mathrm{cd}_{\mathbb{Q}} G \le 2$. The general case is open.

**Sources:**
- [arxiv/math.0703548](https://arxiv.org/abs/math/0703548) **On the Farrell–Jones Conjecture and
  its applications** by *Arthur Bartels, Wolfgang Lück, Holger Reich*, J. Topol. 1 (2008), 57–86.
  Theorem 0.2, Conjecture 0.6, and definitions (4.1), (4.2).

### Prerequisites needed

Mathlib has no algebraic K-theory of rings, so the assembly map is out of reach and #1792 stays
blocked. These two consequences are not:

- Statement 1 needs only `Module.IsStablyFree`, `Module.Finite` and `Module.Projective` over
  `MonoidAlgebra ℤ G`, all of which exist.
- Statement 2 needs the Hattori–Stallings rank, which Mathlib does not have. It can be stated on
  idempotent matrices instead, following (4.1) and (4.2) of the reference: the universal trace of
  the sum of the diagonal entries, evaluated at one conjugacy class. That is a short local
  definition.

One thing to watch: "torsionfree" here must not be rendered as Mathlib's `IsMulTorsionFree`, which
asks that every power map be injective. That is equivalent to torsion-freeness only for commutative
groups — `isMulTorsionFree_iff_not_isOfFinOrder` lives in `section CommGroup`. The Klein bottle
group is torsionfree but has $(ab)^2 = b^2$ with $ab \ne b$.

* ams-16
* ams-19
* ams-20

### Choose either option

j2d9w5xtjn-png:
Fixes #5455.

Mathlib has no algebraic K-theory of rings, so the assembly map — and with it the Farrell–Jones
conjecture itself, #1792 — is out of reach. This adds two of its algebraic consequences from the
introduction of Bartels–Lück–Reich, both expressible with projective modules and matrices over
group rings.

- `projective_module_isStablyFree`: for a torsionfree `G`, every finitely generated projective
  `ℤ[G]`-module is stably free. This is the stable-freeness form of the vanishing of
  `K̃₀(ℤ[G])`, the conclusion of Theorem 0.2 (ii) for the regular ring `ℤ`.
- `bass_conjecture_integral_domain`: Conjecture 0.6, stated on idempotent matrices.

### Formalisation choices

**The Hattori–Stallings rank.** Mathlib has none, so the local definition
`matrixHattoriStallingsTraceAt` follows (4.1) and (4.2) of the reference: the universal trace
`Σ rg · g ↦ Σ rg · (g)`, extended to matrices by summing the traces of the diagonal entries, read
off at one conjugacy class. Quantifying over idempotent matrices rather than over modules is
equivalent, since every finitely generated projective module is the image of an idempotent matrix
and the rank depends only on the isomorphism class.

**The order hypothesis.** `¬IsUnit (orderOf g : R)` covers both cases of Conjecture 0.6 at once:
`orderOf g` is `0` in the infinite-order case, and `IsDomain` supplies `Nontrivial`, so `(0 : R)`
is not a unit. It also excludes `g = 1` automatically, as the source does, because `IsUnit (1 : R)`
always holds.

**Torsion-freeness** is written out as `∀ g, IsOfFinOrder g → g = 1` rather than as
`IsMulTorsionFree`. That class asks that every power map be injective, which is equivalent to
torsion-freeness only for commutative groups — `isMulTorsionFree_iff_not_isOfFinOrder` is stated in
`section CommGroup`. The Klein bottle group is torsionfree but has `(ab)² = b²` with `ab ≠ b`, so
the class would have excluded groups the conjecture covers.

### How this was checked

`lake --wfail build 'FormalConjectures.Arxiv.«math.0703548».FarrellJonesConsequences'` on `8faf8bcc`.

The trace definition was tested against the source on examples: the identity matrix has trace `1`
at the trivial class; two conjugate transpositions in `S₃` contribute `2`, so the sum really is
over a whole conjugacy class; a class not containing `g` contributes `0`; and over a field of
characteristic zero the hypothesis is unsatisfiable for every element of a finite group, as
Conjecture 0.6 requires — `ℚ[ℤ/2]` has the idempotent `(1 + g)/2`, whose rank at `(g)` is `1/2`.

Disclosure: this formalisation was first drafted with Codex, then audited and edited with Claude
Code, and reviewed by me.

https://claude.ai/code/session_01Sq5intN7UGhriwvZzt4axC

### DIFF
--- a/FormalConjectures/Arxiv/math.0703548/FarrellJonesConsequences.lean
+++ b/FormalConjectures/Arxiv/math.0703548/FarrellJonesConsequences.lean
@@ -1,0 +1,79 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+/-!
+# Algebraic consequences of the Farrell–Jones conjecture
+
+The assembly map in the Farrell–Jones conjecture is not yet available in Mathlib. This file
+states two of its algebraic consequences using projective modules and matrices over group rings:
+the vanishing of the reduced projective class group and the Bass conjecture for commutative
+integral domains.
+
+*Reference:*
+- [BLR08] [arxiv/math.0703548](https://arxiv.org/abs/math/0703548)
+  **On the Farrell–Jones Conjecture and its applications**
+  by *Arthur Bartels, Wolfgang Lück, Holger Reich*, J. Topol. 1 (2008), 57–86.
+-/
+
+namespace Arxiv.«math.0703548»
+
+open scoped Classical in
+/--
+The value at the conjugacy class of `g` of the universal trace of a matrix over `R[G]`.
+
+For each diagonal entry, this sums the coefficients of the elements conjugate to `g`. If the
+matrix is idempotent, these values give the Hattori–Stallings rank of the projective module
+represented by the matrix.
+-/
+noncomputable def matrixHattoriStallingsTraceAt {R : Type*} [CommRing R] {G : Type*} [Group G]
+    {n : Type*} [Fintype n] (A : Matrix n n (MonoidAlgebra R G)) (g : G) : R :=
+  ∑ i, (A i i).coeff.sum fun h r ↦ if IsConj h g then r else 0
+
+/--
+**Vanishing of the reduced projective class group for integral group rings.**
+
+If `G` is torsion-free, that is, if its only element of finite order is `1`, then every finitely
+generated projective module over `ℤ[G]` is stably free. This is the stable-freeness form of the
+vanishing of $\widetilde K_0(\mathbb{Z}[G])$, which is the conclusion of Theorem 0.2 (ii) of
+[BLR08] for the regular ring $\mathbb{Z}$. That theorem assumes the K-theoretic Farrell–Jones
+conjecture for `G` with coefficients in `ℤ`.
+-/
+@[category research open, AMS 16 19 20]
+theorem projective_module_isStablyFree {G : Type*} [Group G] (hG : ∀ g : G, IsOfFinOrder g → g = 1)
+    (M : Type*) [AddCommGroup M] [Module (MonoidAlgebra ℤ G) M]
+    [Module.Finite (MonoidAlgebra ℤ G) M] [Module.Projective (MonoidAlgebra ℤ G) M] :
+    Module.IsStablyFree (MonoidAlgebra ℤ G) M := by
+  sorry
+
+/--
+**Bass conjecture for commutative integral domains, in idempotent-matrix form.**
+
+Let `A` be an idempotent matrix over `R[G]`. If the order of `g` is not invertible in `R`, then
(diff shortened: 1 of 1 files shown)


j2d9w5xtjn-png:
Note for review: #5353 reintroduces `Monoid.IsTorsionFree` in
`FormalConjecturesForMathlib/GroupTheory/Torsion.lean` for exactly the reason described under
"Torsion-freeness" above. This PR writes the hypothesis out inline so that it builds against `main`
and stays independent of that one. Once #5353 merges, `projective_module_isStablyFree` should use
`Monoid.IsTorsionFree G`; I am happy to make that change here, or in a follow-up, whichever you
prefer.

FormalConjectures/Arxiv/math.0703548/FarrellJonesConsequences.lean
```diff
@@ -0,0 +1,79 @@
...
+**Vanishing of the reduced projective class group for integral group rings.**
+
+If `G` is torsion-free, that is, if its only element of finite order is `1`, then every finitely
+generated projective module over `ℤ[G]` is stably free. This is the stable-freeness form of the
+vanishing of $\widetilde K_0(\mathbb{Z}[G])$, which is the conclusion of Theorem 0.2 (ii) of
+[BLR08] for the regular ring $\mathbb{Z}$. That theorem assumes the K-theoretic Farrell–Jones
+conjecture for `G` with coefficients in `ℤ`.
+-/
+@[category research open, AMS 16 19 20]
+theorem projective_module_isStablyFree {G : Type*} [Group G] (hG : ∀ g : G, IsOfFinOrder g → g = 1)
```
mo271, line 60: I think we can now use Monoid.IsTorsionFree, but perhaps good to keep it as inline until the fixed-in-mathlib version reaches our repo

mo271 (approved):
Thanks, very nice!

FormalConjectures/Arxiv/math.0703548/FarrellJonesConsequences.lean
j2d9w5xtjn-png: Sounds good, I'll switch once the Mathlib fix reaches the repo. (I merged main and moved the file to the module system in e62a9fec.)

PR #5460, opened 2026-09-10
ci: cancel superseded PR builds and lint workflows

williamjblair:
Cancel an older PR build when a newer run starts. Also add workflow linting to the fast test job.

**Why: obsolete builds kept running**

Audit: 100 recent PRs, September 8–10, 2026 (#5344–#5467). Of these, 88 had executed builds.

| What we observed | Amount |
| --- | ---: |
| Total build time across 162 jobs | 77.7 hours |
| Older builds still running after a newer revision started building | 40 builds across 20 PRs |
| Time those older builds consumed **after that point** | **~16 hours (21% of total build time)** |

The 16 hours measures past overlap. **We have not yet measured how much this PR saves.**

**What changes**

| Situation | Before | With this PR |
| --- | --- | --- |
| New build starts for the same PR | Older build keeps running | Older build is cancelled |
| Merge queue, `main`, manual and preview runs | Run normally | Run normally; never cancelled by this rule |
| Fast test job | No `actionlint` check | Lint all workflows with checksum-verified `actionlint` 1.7.7 |

**Checks**

| Check | Result |
| --- | --- |
| Local workflow lint + 32 script/website tests | Passed |
| Hosted Test scripts | Passed |
| Full hosted build | Running |
| Live cancellation and time saved | Not yet measured |

No new settings or credentials; existing permissions and required check names stay unchanged. Consolidates #5469; independent of #5435.

PR #5469, opened 2026-09-10
ci: lint GitHub Actions workflows in the fast test job

williamjblair:
Run `actionlint` in the existing **Test scripts** job to catch workflow syntax, expression, action-input, and dependency errors without waiting for Lean compilation.

- Pin `actionlint` 1.7.7 and verify the release archive's SHA-256 before execution.
- Check all repository workflows. Disable optional ShellCheck/Pyflakes integrations so results do not depend on which extra tools the runner has installed.
- Keep the existing job, permissions, and required check names.

**Validation:** all current workflows pass; a deliberately invalid workflow expression is rejected. Downloaded and verified the Linux archive checksum. The 26 script tests pass, and the combined workflow with #5435, #5461, and the changed-problem preflight passes lint.

Based directly on `main` and independently mergeable. No new credential or repository setting. This adds early diagnostics; it does not claim to shorten successful Lean builds.

### DIFF
--- a/.github/workflows/build-and-docs.yml
+++ b/.github/workflows/build-and-docs.yml
@@ -45,6 +45,17 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Check GitHub Actions workflows
+        # Pin the checker and archive digest; optional shell/Python linters are
+        # excluded so validation does not depend on the runner's installed tools.
+        run: |
+          curl --fail --silent --show-error --location --retry 3 \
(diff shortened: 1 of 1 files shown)


williamjblair:
Consolidated into #5460 at 52b840d24. The combined diff adds 17 lines to one workflow; actionlint, 26 script tests, six website tests, and diff checks pass. Please review #5460. Closing this duplicate PR without deleting its branch.

PR #5464, opened 2026-09-10
fix(ErdosProblems/951): add the separation hypothesis to the Beurling variant

Deicyde:
**Related.** #5479 (merged today) only changed the docstring of `erdos_951.variants.isBeurlingPrimes`; there is no overlap with this change.

**What changed**

- `erdos_951.variants.beurling` now assumes `Erdos951Prop a` in addition to `IsBeurlingPrimes a`. The docstring quotes the source and explains why the set count is then the count of generalised integers.

**Why**

`IsBeurlingPrimes a` only asks for `1 < a 0`, strict monotonicity and divergence, and `BeurlingIntegers a` is the *set* of values `∏ i, a i ^ k i`, so the counting function ignored multiplicity. For `a n = n + 2` the set of values is exactly the positive integers, the count is `⌊x⌋₊ = x + O(1)`, and `a 2 = 4 ≠ 5`, so the statement was false.

The source states Beurling's conjecture for the sequences of the problem, which satisfy the separation property `Erdos951Prop`: "Beurling conjectured that if the number of reals in $[1,x]$ of the form $\prod a_i^{k_i}$ is $x+o(\log x)$ then the $a_i$ must be the sequence of primes" ([erdosproblems.com/951](https://www.erdosproblems.com/951)). Under `Erdos951Prop a` distinct exponent vectors give distinct values, so counting the set agrees with counting generalised integers with multiplicity, and `a n = n + 2` is excluded (`4 = a 0 ^ 2 = a 2`; this is also the AlphaProof sequence behind #3643).

<details>
<summary>Lean disproof (checked against current <code>main</code> (8faf8bcc) with <code>lake env lean</code>, no errors)</summary>

```lean
import FormalConjectures.ErdosProblems.«951»
open Filter
open scoped Finsupp Nat.Prime Topology

namespace BeurlingDisproof

def allIntegers (n : ℕ) : ℝ := (n : ℝ) + 2

theorem allIntegers_isBeurling : IsBeurlingPrimes allIntegers := by
  refine ⟨by norm_num [allIntegers], ?_, ?_⟩
  · intro m n h
    simpa [allIntegers] using (Nat.cast_lt (α := ℝ)).2 h
  · exact tendsto_atTop_add_const_right _ 2 tendsto_natCast_atTop_atTop

theorem allIntegers_ne_primes : allIntegers ≠ Nat.cast ∘ Nat.nth Nat.Prime := by
  intro h
  have h2 := congrFun h 2
[...]

PR #5479, opened 2026-09-10
Fix description in erdos_951.variants.isBeurlingPrimes

mzhorvath1:
The formal Lean statement (and hence the corresponding proof) also assumes `(hm : StrictMono a)`, which was missing from the informal description, so it is added now.

### DIFF
--- a/FormalConjectures/ErdosProblems/951.lean
+++ b/FormalConjectures/ErdosProblems/951.lean
@@ -37,7 +37,7 @@ def Erdos951Prop (a : ℕ → ℝ) : Prop :=
   ∀ (k ℓ : ℕ →₀ ℕ), k ≠ ℓ → |beurlingInteger a k - beurlingInteger a ℓ| ≥ 1
 
 /-- If a strictly increasing sequence `a : ℕ → ℝ` has property `Erdos951Prop`
-and `1 < a 0`, then `a` is a set of Beurling prime numbers. -/
+and `1 < a 0`, then `a` is a sequence of Beurling prime numbers. -/
 @[category API, AMS 11]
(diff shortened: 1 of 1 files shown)


FormalConjectures/ErdosProblems/951.lean
```diff
@@ -36,8 +36,8 @@ at least one apart. -/
 def Erdos951Prop (a : ℕ → ℝ) : Prop :=
   ∀ (k ℓ : ℕ →₀ ℕ), k ≠ ℓ → |beurlingInteger a k - beurlingInteger a ℓ| ≥ 1
 
-/-- If `a` has property `Erdos951Prop` and `1 < a 0`, then `a` is a set of Beurling
-prime numbers. -/
+/-- If a strictly increasing sequence `a : ℕ → ℝ` has property `Erdos951Prop`
+and `1 < a 0`, then `a` is a set of Beurling prime numbers. -/
```
felixpernegger:
```suggestion
and `1 < a 0`, then `a` is a sequence of Beurling prime numbers. -/
```

felixpernegger (approved):
LGTM, modulo my comment (since the order of the numbers is important I guess)

Paul-Lez (approved):
Thanks!

mo271:
is this the same or a different fix than #5464, @mzhorvath1 @Deicyde

mzhorvath1:
> is this the same or a different fix than #5464, @mzhorvath1 @Deicyde

As far as I understand, this change fixes the docstring in `erdos_951.variants.isBeurlingPrimes` regarding monotonicity , while the other CL fixes a different aspect of `erdos_951.variants.beurling` (separation hypothesis).

PR #5482, opened 2026-09-10
Fix erdos_264.variants.ko_tao_neg

mzhorvath1:
Fix faithfulness of `erdos_264.variants.ko_tao_neg` by replacing `liminf > 0` in ℝ with an eventual positive lower bound.

Identified by @tadamcz using an AI pipeline: https://tadamcz.com/fc-review-results/#/f/ErdosProblems/264


mo271 (approved):
Thanks, LGTM!

PR #5483, opened 2026-09-10
fix(ErdosProblems/830.lean)

archiebrowne:
Previously the type of Erdos380 was: `sorry ↔ {(a, b) | IsAmicable a b}.Infinite`. This doesn't impose that `a \neq b`, as the wikipedia definition does. This affected `infinitely_many_amicable` in `Wikipedia/AmicableNumbers` which had type `type_of% Erdos830.erdos_830.parts.i`. 

`IsAmicable n n ↔ n = 0 ∨ n.Perfect` was true in the old formulation, so a proof of infinitely many perfect numbers would have sufficed to prove the claim. Currently `IsAmicable` looks like:

```
structure IsAmicable (a b : ℕ) : Prop where
  left : σ 1 a = a + b
  right : σ 1 b = a + b
```

Perhaps it needs another field specifying `a \neq b`?

### DIFF
--- a/FormalConjectures/ErdosProblems/830.lean
+++ b/FormalConjectures/ErdosProblems/830.lean
@@ -32,8 +32,8 @@ open scoped Classical in
 Let $A(x)$ counts the number of amicable $1\leq a\leq b\leq x$.
 -/
 noncomputable abbrev A (x : ℝ) : ℝ :=
-  Finset.card <| (Finset.Icc 1 ⌊x⌋₊ ×ˢ Finset.Icc 1 ⌊x⌋₊).filter fun (a, b) ↦
-    a ≤ b ∧ IsAmicable a b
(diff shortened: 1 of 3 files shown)


archiebrowne:
+misformalization

mo271:
@archiebrowne can you fix the definition of `IsAmicable` instead adding a `ne` field to it?

FormalConjectures/ErdosProblems/830.lean
mo271: when you fix the definition than you don't need to require a<b, I suppose

archiebrowne: Good point, I've multiplied by 1 /2 to account for this. Alternatively "a \neq b" could be replaced with "a < b" in IsAmicable

archiebrowne:
> @archiebrowne can you fix the definition of `IsAmicable` instead adding a `ne` field to it?

done

mo271 (approved):
Thanks, LGTM!

PR #5497, opened 2026-09-10
feat: Leopoldts conjecture

WilliamCoram:
This is the shortened version of #5351. We add 5 equivalent versions of Leopold's conjecture - the proofs they are equivalent will exist elsewhere (to be noted in the files when done).


WilliamCoram:
I have cleaned up the code; but I still need to look at the docstrings of files and add the pointer to where the equivalence proofs lie.

PR #5521, opened 2026-09-10
fix(ErdosProblems/394): the logarithmic saving exponent is a real number

Deicyde:
**What changed**

- In `erdos_394.parts.i`, `∃ c > 0` becomes `∃ c > (0 : ℝ)`.

**Why**

Without a type ascription the exponent `c` was inferred as a natural number, so the statement required a positive integer power of `log x`, which is `S₂(x) = O(x² / log x)`. The question asks for some positive real exponent, and the linked formal proof establishes the real-exponent statement with `c = 1/2048`; it does not certify the integer-exponent statement.

**How I checked**

- `lake --wfail build 'FormalConjectures.ErdosProblems.«394»'` passes.

Identified by the fc-review audit: https://tadamcz.com/fc-review-results/#/f/ErdosProblems/394


mo271 (approved):
Thanks!

PR #5522, opened 2026-09-10
fix(OEIS/113010): ask whether another fixed point exists

Deicyde:
**What changed**

- The right-hand side of `conjecture` is `∃ n, a n = n ∧ n ≠ 1 ∧ n ≠ 32`.

**Why**

The docstring asks "Are there any others?", but the statement said that `1` and `32` are the only positive fixed points. So `answer(True)` would have recorded a negative answer to the source's question. The `n > 0` guard is unnecessary: `a 0 = 1`, so `0` is not a fixed point.

**How I checked**

- `lake --wfail build 'FormalConjectures.OEIS.«113010»'` passes.

Identified by the fc-review audit: https://tadamcz.com/fc-review-results/#/f/OEIS/113010


mo271 (approved):
Thanks!

Issue #4927, opened 2026-08-13
Open statements with known solutions


KitaKen1:
The following declarations are marked `research open`, but known results
already settle them.

## Published results

- [ ] [OEIS A287616 — `conjecture`](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/OEIS/287616.lean)
  — [Theorem 1](https://arxiv.org/abs/2606.26035) proves the exact statement → mark solved.

- [ ] [Green 3 — `green_3`](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/GreensOpenProblems/3.lean)
  — [Theorem 1.1](https://arxiv.org/abs/2607.06073) gives answer `True` → mark solved.

- [ ] [Green 31 — two upper bounds](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/GreensOpenProblems/31.lean)
  — [Theorem 1.1](https://arxiv.org/abs/2607.01169) gives the required improved bound → mark both solved.

- [ ] [AME(7,6) and AME(7,10)](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/OpenQuantumProblems/35.lean)
  — [Theorem 1](https://arxiv.org/abs/2608.01011) proves `AME(7,d)` for every `d ≥ 3` → answer `True`.

- [ ] [AME(12,5)](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/OpenQuantumProblems/35.lean)
  — [Theorem 2](https://arxiv.org/abs/2608.05781) proves existence → answer `True`.

- [ ] [Independent Domination — even and odd](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/Arxiv/2107.00295/IndependentDomination.lean)
  — [Corollary 1.3](https://arxiv.org/abs/2202.09594) proves both displayed bounds → mark solved.

- [ ] [MathOverflow 31809](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/Mathoverflow/31809.lean)
  — [Theorem 1.2](https://arxiv.org/abs/2608.09777) gives a counterexample → answer `False`.

## Already implied in the repository

- [ ] [Green 19 — `lower` and `upper`](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/GreensOpenProblems/19.lean)
[...]

PR #5523, opened 2026-09-10
fix(Mathoverflow/31809): ask for a pretriangulated category that is not triangulated

Deicyde:
**What changed**

- The right-hand side of `mathoverflow_31809` is negated: it now says that not every pretriangulated category is triangulated.

**Why**

The question is whether a pretriangulated but not triangulated category exists. The previous right-hand side asserted that every pretriangulated category is triangulated, so `answer(True)` would have meant that no such example exists.

**How I checked**

- `lake --wfail build 'FormalConjectures.Mathoverflow.«31809»'` passes.

Identified by the fc-review audit: https://tadamcz.com/fc-review-results/#/f/Mathoverflow/31809

### DIFF
--- a/FormalConjectures/Mathoverflow/31809.lean
+++ b/FormalConjectures/Mathoverflow/31809.lean
@@ -30,7 +30,7 @@ open CategoryTheory Limits Category Preadditive Pretriangulated
 
 /-- Does there exist a category that is pretriangulated but not triangulated? -/
 @[category research open, AMS 18]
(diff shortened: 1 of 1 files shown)


mo271:
can we incorporate #4927, seems like this has a counterexample now. Which also makes it seem like we should use `answer(False)` here.

mo271 (approved):
Merging as is will make a follow up flipping it to `research solved`

PR #5526, opened 2026-09-10
fix(ErdosProblems/386): let k vary with n

Deicyde:
**What changed**

- In `erdos_386`, `∃ k ≥ 2` is moved inside `∃ᶠ n in atTop`. The docstring follows the source's wording.

**Why**

The source asks whether `binom(n, k)` with `2 ≤ k ≤ n - 2` is a product of consecutive primes infinitely often, with `k` allowed to depend on `n`. The previous quantifier order asked for one fixed `k` with infinitely many `n`, which excludes an infinite family of solutions with unbounded `k` but finitely many solutions per `k`.

**How I checked**

- `lake --wfail build 'FormalConjectures.ErdosProblems.«386»'` passes.

Identified by the fc-review audit: https://tadamcz.com/fc-review-results/#/f/ErdosProblems/386

### DIFF
--- a/FormalConjectures/ErdosProblems/386.lean
+++ b/FormalConjectures/ErdosProblems/386.lean
@@ -27,12 +27,13 @@ namespace Erdos386
 open Nat
 
 /--
-There is a $k$, such that $2 \le k \le n - 2$ and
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #5527, opened 2026-09-10
fix(ErdosProblems/592): countability guards the whole classification

Deicyde:
**What changed**

- `erdos_592` reads `β.card ≤ ℵ₀ → (OrdinalCardinalRamsey (ω ^ β) (ω ^ β) 3 ↔ answer β)`.

**Why**

Without the parentheses the statement elaborated as `(β.card ≤ ℵ₀ → R β) ↔ A β`, so every uncountable ordinal was required to satisfy the answer predicate regardless of its Ramsey property. The problem only asks to classify the countable ordinals.

**How I checked**

- `lake --wfail build 'FormalConjectures.ErdosProblems.«592»'` passes.

Identified by the fc-review audit: https://tadamcz.com/fc-review-results/#/f/ErdosProblems/592


mo271 (approved):
Thanks, LGTM!

PR #5529, opened 2026-09-10
fix(OEIS/167918): drop the existence hypothesis from conjecture1

Deicyde:
**What changed**

- `conjecture1` no longer assumes `∀ n > 0, a n ≠ 0`. The docstring explains why no hypothesis is needed.

**Why**

The hypothesis is equivalent to `∀ n > 0, ∃ k > n, S n ∣ S k`, a substantive assumption that OEIS does not make for the ratio-two conjecture. It is also unnecessary: if no suitable `k` exists then `a n = 0` and `S 0 = P 0 + P 1 = 4`, while `2 * S n ≥ 10` for positive `n`, so such `n` can never be a witness.

**How I checked**

- `lake --wfail build 'FormalConjectures.OEIS.«167918»'` passes.

Identified by the fc-review audit: https://tadamcz.com/fc-review-results/#/f/OEIS/167918

### DIFF
--- a/FormalConjectures/OEIS/167918.lean
+++ b/FormalConjectures/OEIS/167918.lean
@@ -195,11 +195,11 @@ theorem a_4 : a 4 = 7 := by
 /--
 Conjecture: $f(n, k) = 2$ for infinitely many cases, where $k = a(n)$.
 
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

I made a follow up https://github.com/google-deepmind/formal-conjectures/pull/5543, can you review @Deicyde ?

PR #5530, opened 2026-09-10
fix(ErdosProblems/1095): log g(k) ≍ k / log k is a Θ statement

Deicyde:
**What changed**

- `erdos_1095.variants.log_equivalent` uses `=Θ[atTop]` instead of `~[atTop]`.

**Why**

The source's `≍` means bounded above and below by positive constant multiples. `~[atTop]` is asymptotic equivalence, which would impose the leading constant `1` on `log g(k) / (k / log k)`, a claim the heuristic does not make.

**How I checked**

- `lake --wfail build 'FormalConjectures.ErdosProblems.«1095»'` passes.

Identified by the fc-review audit: https://tadamcz.com/fc-review-results/#/f/ErdosProblems/1095


mo271 (approved):
Thanks!

PR #5538, opened 2026-09-10
feat(Millenium): state the Hodge conjecture

Paul-Lez:
State the Hodge conjecture in `FormalConjectures/Millenium/Hodge.lean`, and add the background material it needs as a new `HodgeConjecture` library, ported from https://github.com/Paul-Lez/HodgeConjecture.

Note: this port is currently in progress, so expect other commits that simplify the structure and so on. 

Contributions are very welcome, and should be directed at the `HodgeConjecture` repo for now: this has the latest snapshot of the code, while this PR will be updated periodically!


felixpernegger:
LGTM

PR #5534, opened 2026-09-10
feat: real and complex periods (short file)

WilliamCoram:
This is a short file that only gives the definition of real and complex periods of an elliptic curve. This has no API - for the real one see FC#5370, for the complex one I have not made a PR - let me know if I should.

PR #5345, opened 2026-09-08
add basics about minimal models to BSD.lean

BarinderBanwait:
this PR adds some basic things about minimal models of elliptic curves. 

this will be required for when we define the real period (one of the things in the BSD formula), which btw @WilliamCoram is doing in #5327 

Fable has helped with this, so i request @multramate David Ang to review it critically

if it goes in, then with William's work we should get the real period

PR #5542, opened 2026-09-11
Add the strong Birch and Swinnerton-Dyer conjecture

Multramate:
States the strong Birch and Swinnerton-Dyer conjecture over a number field, alongside the weak form, and defines the period that its leading coefficient is built from.

## Depends on

This branch is stacked on the pull requests below and contains their commits, so most of the diff shown here belongs to them and will disappear as they merge. What is left is `FormalConjectures/Millenium/BSD.lean` and two definitions added to `GlobalPeriod.lean`.

- #5548 — the minimal discriminant ideal
- #5576 — the regulator
- #5577 — the Tamagawa number
- #5578 — the Tate--Shafarevich group
- #5580 — the Hasse--Weil conjecture, which supplies `HasMeromorphicContinuation`

#5534, the real and complex periods, has merged, and this branch builds on the merged file. It replaces the earlier dependency on #5345, which #5548 supersedes.

## What this adds

In `FormalConjecturesForMathlib/AlgebraicGeometry/EllipticCurve/GlobalPeriod.lean`, next to the real and complex periods from #5534:

- `WeierstrassCurve.archimedeanPeriod`: the product of the local periods over the infinite places, real at a real place and complex at a complex one.
- `WeierstrassCurve.period`: that product corrected by the minimal discriminant ideal, $\Omega(E) = \Omega_\infty(E)(|N_{K/\mathbb{Q}}(\Delta_E)| / N(\mathfrak{D}_{E/K}))^{1/12}$. The correction accounts for the local changes of minimal invariant differential, so the period is defined even when $E$ has no globally minimal equation. See [DD2010], §2.1.

In `FormalConjectures/Millenium/BSD.lean`:

- `WeierstrassCurve.Affine.Point.fg`: the Mordell--Weil theorem, as the instance that makes the rank finite.
- `BSD.Weak` and the two weak conjecture statements, over a general number field and over the rationals.
- `BSD.arithmeticLeadingCoefficient`: the predicted leading coefficient, $\Omega(E)\operatorname{Reg}(E/K)|Ш(E/K)|\prod_v c_v / (\sqrt{|\operatorname{disc}K|}\,|E(K)_{\mathrm{tors}}|^2)$.
- `BSD.Strong` and the two strong conjecture statements, asserting finiteness of the Tate--Shafarevich group, the order of vanishing at $s = 1$, and the leading coefficient.
- `BSD.Strong.weak`: the strong conjecture implies the weak one, since two meromorphic continuations agree near $1$ and so have the same order there.

## Formalisation notes

- The conjecture is stated under a meromorphic continuation rather than an analytic one, matching [Gross2011], Conjecture 2.10. `HasMeromorphicContinuation` asks for one on all of $\mathbb{C}$, which is slightly stronger than Gross, who assumes one near $s = 1$ only.
- `Weak` is conditional on a given continuation, while `Strong` asserts one exists, following [DD2010], Conjecture 2.1 (1), which asserts the continuation rather than assuming it.
- The leading coefficient is the coefficient of $(s-1)^r$, that is `meromorphicTrailingCoeffAt`, equivalently $L^{(r)}(E/K,1)/r!$.
- The real periods include every real component, so the Tamagawa product runs over the finite places only. This is the [DD2010] normalisation.
- The rank is `Module.finrank ℤ E.toAffine.Point`, finite by the Mordell--Weil instance.

## Checks

`lake --wfail build FormalConjecturesForMathlib.AlgebraicGeometry.EllipticCurve.GlobalPeriod FormalConjectures.Millenium.BSD` passes, which covers every file on the branch. The only `sorry`s are the conjecture statements themselves and the Mordell--Weil theorem; nothing in `FormalConjecturesForMathlib` contains one.

### DIFF
--- a/FormalConjectures/Millenium/BSD.lean
+++ b/FormalConjectures/Millenium/BSD.lean
@@ -39,19 +39,6 @@ import FormalConjecturesUtil
   [Birch and Swinnerton-Dyer](https://tadamcz.com/autoformalization-results/#/p/wp-birch-and-swinnerton-dyer-conjecture)
 -/
 
-namespace WeierstrassCurve
-
-open NumberField IsDedekindDomain in
-/-- The minimal discriminant ideal of an elliptic curve over a number field is the product of
-the local minimal discriminant ideals `v.asIdeal ^ W.minimalDiscriminantExponent v` over
-all nonzero prime ideals `v` of its ring of integers. Only finitely many exponents are nonzero.
-See [LMFDB](https://www.lmfdb.org/knowledge/show/ec.minimal_discriminant). -/
-noncomputable def minimalDiscriminantIdeal {K : Type*} [Field K] [NumberField K]
-    (W : WeierstrassCurve K) [W.IsElliptic] : Ideal (𝓞 K) :=
-  ∏ᶠ v : HeightOneSpectrum (𝓞 K), v.asIdeal ^ W.minimalDiscriminantExponent v
-
-end WeierstrassCurve
-
 namespace BSD
 
 open scoped Topology
--- a/FormalConjectures/Millenium/BSD.lean
+++ b/FormalConjectures/Millenium/BSD.lean
@@ -39,6 +39,19 @@ import FormalConjecturesUtil
   [Birch and Swinnerton-Dyer](https://tadamcz.com/autoformalization-results/#/p/wp-birch-and-swinnerton-dyer-conjecture)
 -/
 
+namespace WeierstrassCurve
+
+open NumberField IsDedekindDomain in
+/-- The minimal discriminant ideal of an elliptic curve over a number field is the product of
+the local minimal discriminant ideals `v.asIdeal ^ W.minimalDiscriminantExponent v` over
+all nonzero prime ideals `v` of its ring of integers. Only finitely many exponents are nonzero.
+See [LMFDB](https://www.lmfdb.org/knowledge/show/ec.minimal_discriminant). -/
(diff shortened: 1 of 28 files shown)


FormalConjectures/Millenium/BSD.lean
```diff
@@ -31,91 +31,84 @@ import FormalConjecturesUtil
...
-The $L$-function is expected to be holomorphic, and `IsHolomorphicLFunction` is that stronger
-notion, but meromorphy is all that is needed to state the Birch and Swinnerton-Dyer conjecture.
-[Gross2011] states the conjecture under this hypothesis, and we take that form as authoritative. -/
 def IsLFunction (E : WeierstrassCurve K) (L : ℂ → ℂ) : Prop :=
-  Meromorphic L ∧ ∀ s : ℂ, 3 / 2 < s.re → L s = E.LSeries s
-
-/-- `L` is a holomorphic $L$-function of `E`: it is entire and agrees with the $L$-series of `E`
-on $\operatorname{Re} s > 3/2$. This is the continuation Hasse and Weil conjectured. -/
-def IsHolomorphicLFunction (E : WeierstrassCurve K) (L : ℂ → ℂ) : Prop :=
   Differentiable ℂ L ∧ ∀ s : ℂ, 3 / 2 < s.re → L s = E.LSeries s
```
BarinderBanwait: we went over this in the weak PR, this should be Meromorphic, and we have another `IsHomomorphicLFunction` for the holomorphic version

BarinderBanwait (changes requested):

PR #5547, opened 2026-09-11
feat(Paper): Nakayama conjectures

ymonbru:
This file collects the Nakayama Conjecture together with the family of homological conjectures
surrounding it.

Let $R$ be a commutative Artinian ring and let $A$ be an Artin $R$-algebra, that is, an
$R$-algebra which is finitely generated as an $R$-module.
All modules are assumed to be finitely generated left $A$-modules.
The algebra $A$ is called *self-injective* if it is injective as a left module over itself.

if $A$ has an injective resolution all of whose terms are projective, then $A$ is self-injective

Co-authored-by: Wassilij Gnedin (@wgnedin)

### DIFF
--- a/FormalConjectures/Paper/NakayamaConjectures.lean
+++ b/FormalConjectures/Paper/NakayamaConjectures.lean
@@ -197,9 +197,14 @@ The goal of this part is to define a version of duality in artinian algebra that
 variable {C: Type u} [Category.{v} C] [Abelian C]
 variable {M E : C} (u : M ⟶ E)
 
+/--
+ `u`is an essential extension of `M` if for every subobject `s` of `E` that is non zero the pullback wuth `u`is also non zero.
+-/
 def IsEssentialExtension : Prop :=
   ∀ s : Subobject E, IsZero (Subobject.underlying.obj s) ∨ ¬ IsZero (pullback u ((MonoOver.forget E).obj (Subobject.representative.obj s)).hom)
 
+/--
+An injective hull is an essential extension that is also injective.-/
 def IsInjectiveHull : Prop := Injective E ∧ IsEssentialExtension u
 
 variable (M)
@@ -215,7 +220,11 @@ noncomputable def injectiveHullHom [HasInjectiveHull M] : M ⟶ (injectiveHull M
 lemma IsInjectiveHullInjectiveHull [HasInjectiveHull M] : IsInjectiveHull (injectiveHullHom M) :=
   (HasInjectiveHull.existsInjHull (M := M)).choose_spec.choose_spec
 
+/--
+ From wikipedia: https://en.wikipedia.org/wiki/Injective_hull
 
+ In an AB5 actegory with enough injectives (in particular categories of modules over a ring), every object has an injective hull
+-/
 @[category API, AMS 16 18]
 instance [HasFilteredColimits C] [AB5 C] [EnoughInjectives C] : HasInjectiveHull M := by
   sorry
--- a/FormalConjectures/Paper/NakayamaConjectures.lean
+++ b/FormalConjectures/Paper/NakayamaConjectures.lean
@@ -1,0 +1,100 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-/
+
+import FormalConjecturesUtil
+
+open CategoryTheory Abelian Limits
+
+/-!
+# Strong Nakayama Conjecture
+*Reference:* [R.R. Colby and K.R. Fuller, A note on the Nakayama conjectures, Tsukuba J. Math. \textbf{14} (1990), no.~2, 343--352.](https://doi.org/10.21099/tkbjm/1496161457)
+
+# Generalized Nakayama Conjecture
+# Auslander-Reiten Conjecture
+*Reference:*
+[M. Auslander and I. Reiten, On a generalized version of the Nakayama conjecture, Proc. Amer. Math. Soc. 52 (1975), 69--74.](https://doi.org/10.1090/S0002-9939-1975-0389977-6)
+
+# Tachikawa Conjectures
+*Reference:* [H. Tachikawa, Quasi-Frobenius Rings and Generalizations: QF-3 and QF-1 Rings, Lecture Notes in Mathematics, vol. 351, Springer, Berlin–Heidelberg, 1973](https://doi.org/10.1007/BFb0060005)
+
+# Nakayama Conjecture
+*Reference:* [T. Nakayama, On algebras with complete homology. Abh. Math. Sem. Univ. Hamburg 22 (1958), 300–307.](https://doi.org/10.1007/BF02941960)
+-/
+
+universe u v w
+
+variable {R : Type u} {A : Type v} [CommRing R] [IsArtinianRing R] [Ring A] [Algebra R A] [Module.Finite R A] {M : ModuleCat.{v} A} [Module.Finite A M.carrier]
+
+
+namespace NakayamaConjectures
+
+/--
+The Strong Nakayama Conjecture:
+Let A be a finite module over an aritinain ring. For any finitely generated A-module M with Ext^i(M,A) = 0 for any integer i ≥ 0 it follows that M = 0.
+-/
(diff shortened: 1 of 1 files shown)


j2d9w5xtjn-png (commented):
Thanks for this — the collection of conjectures and the justifications in the `tc1` and `nc` docstrings are the right shape. I built the branch (`0f4b365`) locally against the repo toolchain. Everything below marked *(checked)* is machine-checked against your file.

## Blocking

### 1. `dual R A` dualises the wrong module: `(⊤ : Ideal R).jacobson = ⊤`

`Ideal.jacobson I` is `sInf {J | I ≤ J ∧ J.IsMaximal}`, so for `I = ⊤` the infimum is over the empty set. Mathlib has

```lean
theorem Ideal.jacobson_top : jacobson (⊤ : Ideal R) = ⊤
```

so `R ⧸ ((⊤ : Ideal R).jacobson)` is the zero ring *(checked)*, and the intended cogenerator $E(R/J(R))$ is $E(0) = 0$. The Jacobson radical is `(⊥ : Ideal R).jacobson`, i.e. `Ring.jacobson R` (`Ideal.jacobson_bot`).

As written the intended `dual R A` is the zero module, which empties all four `*_with_dual` statements; `snc_with_dual` degenerates to "every finitely generated module is zero".

### 2. `IsEssentialExtension` does not require `Mono u`

Without it the definition is vacuous. The zero object is an "injective hull" of every object *(checked against your file)*:

```lean
example (M : C) : Abelian.IsInjectiveHull (0 : M ⟶ (0 : C)) := by
  constructor
  · exact Injective.of_iso (isZero_zero C).isoZero.symm inferInstance
  · intro s; left; exact (isZero_zero C).of_mono (Subobject.arrow s)
```

Consequently `HasInjectiveHull M` is provable for every object of every abelian category with no hypotheses *(checked)*, the sorried instance establishes nothing, and `injectiveHull M`, a `Classical.choose`, is not the injective hull.

Suggested statement:

```lean
def IsEssentialExtension : Prop :=
  Mono u ∧ ∀ s : Subobject E, s ≠ ⊥ → ¬ IsZero (pullback u s.arrow)
```

`s.arrow` also replaces `(MonoOver.forget E).obj (Subobject.representative.obj s)`, and `s ≠ ⊥` replaces `IsZero (Subobject.underlying.obj s)`.

### 3. The sorried instance fails the build

```
$ lake --wfail build FormalConjectures.Paper.NakayamaConjectures
warning: FormalConjectures/Paper/NakayamaConjectures.lean:223:0: Placeholder definitions
  (e.g., `def foo : Type := sorry`) are not allowed.
error: build failed
```

This is `weak.linter.style.stubs` on `instance [HasFilteredColimits C] [AB5 C] [EnoughInjectives C] : HasInjectiveHull M := by sorry`. The Lean build job has not run on the PR yet, so it is not visible in the checks. Note that `lake env lean` on the file alone does not show it, since the linter is enabled in the lakefile.

There is a more basic problem behind the linter: the instance is used to *define* `dual`, so the conjecture statements depend on data extracted from `sorryAx`. That is what the "never add an `axiom`, `opaque`, or `constant`" rule in [STATEMENTS.md](https://github.com/google-deepmind/formal-conjectures/blob/main/STATEMENTS.md) is aimed at. Once `Mono` is added the instance becomes a real theorem that needs a real proof. A cheaper route is to avoid the choice: make `E` and a hypothesis `IsInjectiveHull (u : ModuleCat.of R (R ⧸ Ring.jacobson R) ⟶ E)` explicit arguments of the dual-form conjectures.

## Statements

### 4. `snc_iff_snc_with_dual` claims left-right symmetry of the Strong Nakayama Conjecture

$D = \operatorname{Hom}_R(-, E)$ is an exact duality $\operatorname{mod} A \simeq (\operatorname{mod} A^{\mathrm{op}})^{\mathrm{op}}$, so

$$\operatorname{Ext}^i_A(DA, M) \cong \operatorname{Ext}^i_{A^{\mathrm{op}}}(DM, A),$$

and $M \mapsto DM$ is a bijection onto the nonzero finitely generated right modules. So `snc_with_dual` for $A$ is the Strong Nakayama Conjecture for $A^{\mathrm{op}}$, not for $A$. The `↔` as stated is a left-right symmetry claim. I could not find it in Colby-Fuller; it needs a citation, or the statement should use `Aᵐᵒᵖ`, or both sides should be quantified over all algebras.

### 5. The two `iff` lemmas quantify differently

`tc1_and_tc2_iff_nc` uses `type_of% @tc1`, quantifying over all `R` and `A`, while `nc_iff_nc_with_dual` fixes `R` and `A`. But `nc_with_dual` is already equivalent to `tc1 ∧ tc2` at a fixed `A`: for `tc1` take `M = DA`, where `Ext^i(DA, DA) = 0` for `i > 0` because `DA` is injective, and `DA` projective gives `A` self-injective; for `tc2` use that injectives are projective over a self-injective algebra. So the two lemmas together assert `tc1 A ∧ tc2 A ↔ nc A` for a fixed `A`, which is stronger than the usual statement about the class of all Artin algebras. Please use one convention and record it in the docstrings.

### 6. Universes in `tc1_and_tc2_iff_nc`

The elaborated statement has a stray unused outer `∀ {R : Type u_1} [CommRing R] [IsArtinianRing R]`, coming from `include R`, and the three `@`-forms auto-bind six independent universes `u_2 … u_7`, so the two sides range over algebras in unrelated universes. The declared `universe u v` are not used here.

### 7. `nc_with_dual` has no justification

Its docstring reads like the Auslander-Reiten conjecture. `nc` and `tc1` explain why their form agrees with the standard one; this one should too. By item 5 it is `tc1 ∧ tc2`.

### 8. Smaller points

- `snc_with_dual` says "finite-dimensional algebras", but the setting is an Artin $R$-algebra.
- `snc_with_dual` is the only conjecture phrased as a disjunction rather than an implication. The implication matches its docstring and the other statements.

## Conventions

- `namespace Abelian` after `open CategoryTheory` puts the declarations in the root namespace, `_root_.Abelian.IsEssentialExtension` *(checked with `#check`)*, not `CategoryTheory.Abelian`. Use `namespace CategoryTheory.Abelian`, or move this generic material to `FormalConjecturesForMathlib/`, which forbids `sorry` and so would need the proof of item 3.
- Naming: `snc`, `gnc`, `tc1`, `tc2`, `nc` are terse; `IsInjectiveHullInjectiveHull` should be `isInjectiveHull_injectiveHullHom`; the field `existsInjHull` should be `exists_injectiveHull`. `HasInjectiveHull`, `injectiveHull`, `injectiveHullHom` and `dual` have no docstrings.
- The commented-out `gnc_iff_arc` block is a review note in a Lean file, which [AGENTS.md](https://github.com/google-deepmind/formal-conjectures/blob/main/AGENTS.md) asks you to avoid. The name in it is also stale: #5440 uses `AuslanderReiten.auslander_reiten`, not `ArtinAlgebra.auslander_reiten`.
- 17 lines exceed 100 columns. This is not enforced by CI, but the rest of the repository follows it.
- Lines 194 and 236 write "artinian algebra" in lower case. The name is capitalised, and the file says "Artin algebra" everywhere else.
- Other typos: "actegory" → "category", "enveloppe" → "envelope", "wich" → "which", "uselfull" → "useful", "excpet" → "except", "equiped" → "equipped", "wuth" → "with", `{C: Type u}` → `{C : Type u}`, and missing spaces around Lean quotations such as ``` `A`and`A^*` ```.
- CONTRIBUTING steps 2 and 9 ask for a companion issue linked from the pull request. I did not find one for the Nakayama conjectures.
- All four DOIs resolve and match their bibliographic records. One refinement: `10.1007/BFb0060005` is the chapter "Conjectures arising from Nakayama's conjecture", pages 105-119 of LNM 351, rather than the whole book.

## What is already correct

The file compiles. The implication chain in the module docstring is accurate, and `snc`, `gnc`, `tc1`, `tc2` and `nc` match their sources, including the `i ≥ 0` versus `i > 0` ranges. The two justifications you give, that injectivity in the finitely generated category agrees with module injectivity over an Artin algebra, and that any injective resolution has the minimal one as a direct summand, are the right ones. The transport of the module structure works: `(a • f) x = f (x * a)` holds by `rfl` *(checked)*. Small note there: `IsScalarTower R A (dual R A)` does not synthesise, so the `A`-action and the `R`-action are unrelated for the elaborator. That is harmless for these statements, since they only use the `A`-action, but it is worth a comment.

---

*This review was generated by Claude Code (Opus 5) and posted by the account holder.*

j2d9w5xtjn-png:
Addendum to point 6 of my review above, since I stopped short of naming the fix: for the unused outer parameter in `tc1_and_tc2_iff_nc`, delete the `include R in` line before the lemma. `@tc1`, `@tc2` and `@nc` each quantify over their own `R` and `A`, so `include R` only prefixes the statement with `{R : Type u} [CommRing R] [IsArtinianRing R]` that nothing inside uses, and a caller has to supply an unused Artinian ring.

The universe point in that same item is separate and survives the fix: the three `@`-forms auto-bind independent universes, so the two sides of the `↔` still range over algebras in unrelated universes.

*This comment was generated by Claude Code (Opus 5) and posted by the account holder.*

Issue #5005, opened 2026-08-15
GreensOpenProblems: `answer( )` slots scoped outside the quantifier their value depends on — green_24, green_16, green_37, green_37_asymptotic, green_41 (and two `green_35` notes)


Kuberwastaken:
## Summary

Six `research open` declarations in `GreensOpenProblems` put an `answer( )` slot in a position
where the value it must take **depends on a quantifier that encloses the slot**. Each one is
therefore either

* **false**, if the answer is required to be a closed term — which is what `answer( )` means in
  `with_auxiliary` mode, where it becomes the top-level definition `<decl>._answer`; or
* **trivial**, if the answer may capture the enclosing binder, since then it can be the target
  expression itself.

In neither reading does the declaration ask Green's question. Four of the five in Part A come
with a machine-checked disproof of the closed-answer reading, reproduced in full below.

This is the defect the repository's own `AnswerLinter` describes — *"Move the quantifiers
outward … likely the intention was `theorem foo : answer(sorry) ↔ ∀ᵉ (bar : ℕ) …"* — but its
`stars_with_answer_sorry_iff` check only inspects statements of the form
`answer(sorry) ↔ …`, and none of these has that shape, so it never fires. Related: #33, #1407,
#4923.

**Not reported here.** `green_25`, `green_51` and `green_27.equivalent` *are* closable by `rfl` /
`IsEquivalent.refl` with genuinely closed answers, but that is the case CONTRIBUTING explicitly
places out of scope ("one can provide trivial answers that aren't mathematically interesting,
e.g. the set `{n : ℕ | P n}` itself … outside of the scope of this repository"), and #33 already
tracks it as a design question. The items below are different: no answer, trivial or otherwise,
makes them true.

The underlying mathematics is untouched throughout. Every one of Green's questions is exactly as
open as before; these are statements about what the Lean declarations assert.

**Machine-checking method.** No `lake build` was available on this machine (insufficient disk for
the full build; only the `FormalConjecturesUtil` / `FormalConjecturesForMathlib` oleans were
present). Each proof below was checked by copying the relevant definitions **verbatim** from
`main` @ `638da20e` into a standalone file importing `FormalConjecturesUtil`, and running

```
lean -DautoImplicit=false -DrelaxedAutoImplicit=false -DwarningAsError=true probe.lean
```

against Mathlib v4.27.0 — exit code 0, no `sorry`, no errors, no warnings. That checks the
[...]

PR #5553, opened 2026-09-11
fix(GreensOpenProblems/41): exists_better_bound follows from the Kravitz-Leng bound

j2d9w5xtjn-png:
Related to #5005 — and directly relevant to the fix suggested there (see the note below).

**What changed**

- `green_41.variants.kravitz_leng` is moved above the variant that now depends on it.
- `green_41.variants.exists_better_bound` becomes `research solved` with `answer(True)` and a
  complete proof.
- Its docstring records that the formulation does not express an improvement over [KrLe25].

**Why**

The variant asks for some `ans` with `minCopies ε ≤ ans < exp³(ε^(-C))`, where `C` is
existentially quantified in the *same* statement. One may therefore take the Kravitz–Leng bound
itself as `ans` and simply enlarge the exponent. With `C' = max C 0 + 1 > C` and `0 < ε < 1`,

```
ε^(-C) < ε^(-C')
```

since `t ↦ ε^t` is strictly decreasing for `0 < ε < 1`; three applications of strict monotonicity
of `exp` give the required strict inequality. No asymptotic improvement is involved.

**Note for #5005.** That issue flags the headline `green_41` and suggests folding it into
`green_41.variants.exists_better_bound`, "which puts `∃ ans : ℝ` **inside** the `∀ ε`". That
alone is not sufficient: as shown here, the variant is already satisfied by the bound the
repository records. Capturing genuine improvement needs the comparison constant fixed *before*
`ans` is chosen — for instance quantifying over every `C` admissible for [KrLe25] and asking for a
bound below `exp³(ε^(-C))` for that same `C`. I have not included such a reformulation here, to
keep this PR to the one defect; happy to add it if you would prefer them together.

This is not the trivial-answer case CONTRIBUTING excludes: the answer is not the statement copied
into its own slot, but a consequence of the specific bound already recorded in this file, and the
proof is the content.

### DIFF
--- a/FormalConjectures/GreensOpenProblems/41.lean
+++ b/FormalConjectures/GreensOpenProblems/41.lean
@@ -71,28 +71,42 @@ theorem green_41 :
   sorry
 
 /--
-Is there a better bound than the best-known bound from [KrLe25]?
-This is an existential version of the main problem that does not require providing the bound explicitly.
--/
-@[category research open, AMS 51 52]
-theorem green_41.variants.exists_better_bound : answer(sorry) ↔
-    ∃ C : ℝ, C > 0 ∧ ∃ ε₀ > 0, ∀ ε ∈ Ioc 0 ε₀,
-      ∃ ans : ℝ, (minCopies ε : ℝ) ≤ ans ∧ ans < Real.exp (Real.exp (Real.exp (ε ^ (-C)))) := by
-  sorry
-
-/-- Is $\varepsilon^{-C}$ rotations enough? -/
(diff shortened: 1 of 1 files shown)


mo271:
@j2d9w5xtjn-png can you resolved the merge conflict and see what remains to be fixed here?

j2d9w5xtjn-png:
Thanks. #5757 has since restated `green_41` and replaced `exists_better_bound` by `double_exponential_bound`, which the Kravitz–Leng bound does not imply, so nothing is left to fix here. Closing as superseded.

PR #5555, opened 2026-09-11
fix(OEIS/24356): the only-zero claim also requires a 4 = 0

j2d9w5xtjn-png:
**What changed**

- `conjecture` becomes `∀ n : ℕ, a n = 0 ↔ n = 4` (previously `a n = 0 → n = 4`).
- New `category test` theorem `a_4 : a 4 = 0` certifies the existence direction.

**Why**

The OEIS comment is *"I conjecture that a(4) is the only zero"*, which asserts both that `a 4 = 0`
and that nothing else vanishes. The implication alone does not say a zero exists — a sequence that
never vanishes satisfies it vacuously.

`a_4` is a kernel-reduced computation, not `native_decide`. The `4 × 4` Hankel matrix is

```
 2  3  5  7
 3  5  7 11
 5  7 11 13
 7 11 13 17
```

(using `Nat.nth_count` for the entries 13 and 17), and singularity is witnessed by the explicit
null vector `(6, -3, -2, 1)`.

The uniqueness direction remains `research open`.

### DIFF
--- a/FormalConjectures/OEIS/24356.lean
+++ b/FormalConjectures/OEIS/24356.lean
@@ -49,10 +49,37 @@ theorem a_2 : a 2 = 1 := by
   simp [a, Matrix.det_fin_two, Nat.nth_prime_zero_eq_two,
     Nat.nth_prime_one_eq_three, Nat.nth_prime_two_eq_five]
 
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5556, opened 2026-09-11
fix(ErdosProblems/272): the main asymptotic follows from the Szabo bound

j2d9w5xtjn-png:
Addresses the "Erdős 272 — main asymptotic" item of #4927.

**What changed**

- `erdos_272` becomes `research solved` with `answer(fun N : ℕ ↦ (N : ℝ) ^ 2 / 2)` and a complete
  proof.
- It is moved below `erdos_272.variants.szabo`, which the proof uses.

**Why**

The already-`research solved` Szabó variant in this file states

```
maxArithInterCard N - N² / 2  =O[atTop]  N^(5/3) (log N)³
```

Since `log(x)³ = o(x^(1/3))`, multiplying by `x^(5/3)` gives `N^(5/3)(log N)³ = o(N²)`. Scaling by
`1/2` and applying `IsBigO.trans_isLittleO` yields an `o(N²/2)` error, i.e.
`maxArithInterCard N ~ N²/2`.

**Judgement call.** This settles the headline *as formalised*, which asks only for the asymptotic
equivalence. The sharper question Szabó actually poses — whether the error term is `O(N)` — is a
separate declaration in this file and stays `research open`. If you would rather the headline
capture that sharper form, the alternative is to strengthen the statement instead of marking it
solved; happy to switch.

This is not the trivial-answer case CONTRIBUTING excludes: the answer `N ↦ N²/2` is derived from
the recorded Szabó error term, not copied from the statement.

### DIFF
--- a/FormalConjectures/ErdosProblems/272.lean
+++ b/FormalConjectures/ErdosProblems/272.lean
@@ -40,16 +40,6 @@ noncomputable def maxArithInterCard (N : ℕ) : ℕ :=
   sSup {#A | (A : _) (_ : IsArithInterSet N A)}
 
 /--
-Let $N\geq 1$. What is the largest $t$ such that there are
-$A_1,\ldots,A_t\subseteq \{1,\ldots,N\}$ with $A_i\cap A_j$ a non-empty
-arithmetic progression for all $i\neq j$?
--/
-@[category research open, AMS 5]
-theorem erdos_272 :
-    (fun N ↦ (maxArithInterCard N : ℝ)) ~[atTop] (answer(sorry) : ℕ → ℝ) := by
-  sorry
-
-/--
 Simonovits and Sós have shown that $t\ll N^2$.
 -/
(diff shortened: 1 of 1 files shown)


mo271:
Didn't look closely but it seems this is still marked open on erdosproblems.com ? Why?

j2d9w5xtjn-png:
Good catch. The site lists 272 as open because its question is the exact maximum for each $N$. Its notes describe Szabó's bound $N^2/2 + O(N^{5/3}(\log N)^3)$ as "resolving the asymptotic question", not the exact value. When I opened this PR, `erdos_272` asked only for an asymptotic equivalent, so it followed from `erdos_272.variants.szabo` in the same file. That made the headline a weaker statement than the site's question rather than a solved one. #5807 has since restated `erdos_272` as the exact value and kept it open, which is the right fix. The asymptotic is already recorded by the Szabó variant, so I am closing this.

PR #5557, opened 2026-09-11
fix(GreensOpenProblems/19): the bounds follow from the solved value C = 4

j2d9w5xtjn-png:
Addresses the "Green 19 — `lower` and `upper`" item of #4927.

**What changed**

- `green_19.lower` and `green_19.upper` become `research solved`, proved from the
  already-`research solved` `green_19 : C = 4` via `rw [green_19]` (plus `norm_num` for the
  numeric comparison).

**Why**

[Ma21]'s bounds `3.13 ≤ C` and `C ≤ 4` were tagged `research open`, but the same file already
records the exact value `C = 4` as solved, which implies both.

Both proofs are two lines and derive from the file's own solved value; no new mathematics is
claimed.

### DIFF
--- a/FormalConjectures/GreensOpenProblems/19.lean
+++ b/FormalConjectures/GreensOpenProblems/19.lean
@@ -80,15 +80,16 @@ This question has been resolved by [FSS20], showing that $C = 4$.
 theorem green_19 : C = 4 := by
   sorry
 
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #5561, opened 2026-09-11
fix(OpenQuantumProblems/35): AME(7,6), AME(7,10) and AME(12,5) all exist

j2d9w5xtjn-png:
Addresses the "AME(7,6) and AME(7,10)" and "AME(12,5)" items of #4927.

**What changed**

- New source-backed theorem `ame_7_exists_iff (d : ℕ) : ExistsAME 7 d ↔ 3 ≤ d`, recording the
  complete seven-party classification.
- `ame_7_6_open` and `ame_7_10_open` are `research solved` with `answer(True)`, each derived from
  `ame_7_exists_iff` by a complete Lean proof (no new `sorry`).
- `ame_12_5_open` is `research solved` with `answer(True)`, literature-backed like the neighbouring
  existence results in this file.
- Both papers added to the reference list.

**Sources**

Shi, Zhang, Zhao and Li, *Complete Existence Classification of Seven-Partite Absolutely Maximally
Entangled States*, [arXiv:2608.01011](https://arxiv.org/abs/2608.01011) — abstract: "We prove that
an absolutely maximally entangled state of seven qudits exists if and only if the local dimension
satisfies `d ≥ 3`." They construct cyclic quadratic-phase states for every odd dimension and a
coupled binary–odd-dimensional construction for every `d ≡ 2 (mod 4)`; with the known power-of-two
cases and the product property this covers all `d ≥ 3`. Since `6 ≥ 3` and `10 ≥ 3`, both benchmark
cases follow. This is consistent with the file's existing `ame_7_2_not_exists`.

Bevins and Bidav, *Symmetry-guided constructions of absolutely maximally entangled states in five
open cases*, [arXiv:2608.05781](https://arxiv.org/abs/2608.05781) — abstract: "We give explicit
Hermitian self-dual maximum distance separable codes with parameters `[12,6,7]_25` … The stabilizer
construction proves the existence of `AME(12,5)` …". The paper states the printed matrices are
certified by exact Hermitian products and complete square-minor enumeration.

That second paper also gives `AME(18,11)`, `AME(18,13)`, `AME(17,11)` and `AME(17,13)`, but this
file has no benchmark declarations for those, so nothing else changes here. The remaining open
benchmarks (8, 9, 10, 11 parties, and `AME(12,6)`, `AME(12,10)`) are untouched.

**Note.** The two seven-party declarations keep their `_open` suffix, which now reads oddly. I did
not rename them, since that would churn identifiers beyond the status change — happy to rename if
you would prefer.

### DIFF
--- a/FormalConjectures/OpenQuantumProblems/35.lean
+++ b/FormalConjectures/OpenQuantumProblems/35.lean
@@ -57,6 +57,12 @@ of the parties, the corresponding reduced density matrix is maximally mixed.
   and K. Życzkowski,
   *Absolutely maximally entangled pure states of multipartite quantum systems*,
   arXiv:2508.04777 (2025).
+- S. Bevins and Y. Bidav,
+  *Symmetry-guided constructions of absolutely maximally entangled states in five open cases*,
+  [arXiv:2608.05781](https://arxiv.org/abs/2608.05781) (2026).
+- F. Shi, X. Zhang, Q. Zhao, and L. Li,
+  *Complete Existence Classification of Seven-Partite Absolutely Maximally Entangled States*,
+  [arXiv:2608.01011](https://arxiv.org/abs/2608.01011) (2026).
 
 This file formalizes the problem of determining for which pairs $(n,d)$ there exists an
 absolutely maximally entangled pure state $\mathrm{AME}(n,d)$.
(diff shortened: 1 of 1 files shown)


mo271:
@MarioKrenn6240 perhaps we can first review and merge the other pending results in this file?

mo271 (approved):
Thanks, LGTM!

PR #5565, opened 2026-09-11
fix(OEIS/113271): a 4 is prime and a 5 = 135457 is not

j2d9w5xtjn-png:
**What changed**

- The `conjecture1` search starts after index `4` instead of after index `5`.
- New `category test` theorems: `a_4_prime`, `a_5 : a 5 = 135457`, and `a_5_not_prime`.
- The docstring records the discrepancy with the OEIS comment.

**Why**

The docstring repeated the OEIS comment, "The smallest primes in this (always odd) sequence are
`a(1) = 3`, `a(3) = 41` and `a(5) = 543`", and the search was therefore restricted to `5 < n`.

That comment is inconsistent with the rest of the same OEIS entry. A113271's DATA section reads
`1, 3, 9, 41, 593, 135457, ...` and its EXAMPLE section reads `a(5) = 135457 = 1^32 + 2^16 + 4^8 +
8^4 + 16^2 + 32^1`. So `543` is neither the index-five term nor prime (`543 = 3 · 181`).

Recomputing from the file's own definition confirms the DATA: `a(4) = 593`, which **is** prime, and
`a(5) = 135457 = 7 · 37 · 523`, which is not. The genuine third prime is therefore `a(4)`, and the
old bound `5 < n` excluded it from the record while starting the search one index too late.

The question itself is unchanged and still open: no further prime occurs for `n ≤ 12`.

### DIFF
--- a/FormalConjectures/OEIS/113271.lean
+++ b/FormalConjectures/OEIS/113271.lean
@@ -48,13 +48,32 @@ theorem a_3 : a 3 = 41 := by rfl
 @[category test, AMS 11]
 theorem a_4 : a 4 = 593 := by rfl
 
+@[category test, AMS 11]
+theorem a_4_prime : (a 4).Prime := by
+  rw [a_4]
+  norm_num
+
+/-- The next term from the defining formula and the official OEIS b-file. -/
+@[category test, AMS 11]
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5566, opened 2026-09-11
fix(OEIS/7468): the only-square claim also requires a 38 to be a square

j2d9w5xtjn-png:
**What changed**

- `conjecture` reads `IsSquare (a n) ↔ n = 38` instead of `IsSquare (a n) → n = 38`.

**Why**

The docstring states "The only positive integer `n` such that `a(n)` is a perfect square is
`n = 38`". That asserts two things: that `a(38)` is a square, and that no other positive index
gives one. The implication only encodes the second, and is satisfied vacuously by a sequence with
no square at all.

`a(38)` is indeed a square: row 38 runs over prime indices 703 to 740 and sums to
`207936 = 456²`. A direct scan of `n = 1 … 658` finds no other square, so the biconditional is
consistent with the data and remains open.

This PR does not add a Lean certificate for `a(38) = 207936`, which would require certifying the
38 relevant `Nat.nth Nat.Prime` values; both directions stay `sorry`, as before.

### DIFF
--- a/FormalConjectures/OEIS/7468.lean
+++ b/FormalConjectures/OEIS/7468.lean
@@ -65,7 +65,7 @@ The only positive integer $n$ such that $a(n)$ is a perfect square is $n=38$.
 - Carlos Eduardo Olivieri, Mar 09 2015
 -/
 @[category research open, AMS 11]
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5567, opened 2026-09-11
fix(OEIS/100475): ask the existential question for positive starts

j2d9w5xtjn-png:
**What changed**

- `conjecture` becomes an existential question with the answer outside the binder:
  `answer(sorry) ↔ ∃ x, 0 < x ∧ x ≠ 1 ∧ IsUltimatelyPeriodic (aStartAt x)`.
- New `category API` lemma `aStartAt_zero : aStartAt 0 n = 0`.
- New `research solved` theorem `conjecture_with_zero` recording the degenerate `x = 0` answer.

**Why**

Two problems with the old form `(x : ℕ) (h : x ≠ 1) : answer(sorry) = IsUltimatelyPeriodic
(aStartAt x)`:

1. The `answer( )` slot sits underneath the binder `x`, so a single closed proposition would have
   to equal `IsUltimatelyPeriodic (aStartAt x)` for every `x ≠ 1` simultaneously. (Same shape as
   the cases collected in #5005.)
2. `aStartAt` is totalised on `ℕ` with `if k = 0 then 0`, so `aStartAt 0` is identically zero and
   is ultimately periodic — and `0 ≠ 1`. The source recurrence uses the one-based prime index
   `p₁ = 2`, so `x = 0` is an artifact of totalisation, not a real starting value.

`conjecture_with_zero` records that degenerate answer explicitly rather than leaving it as a trap,
and the intended question now requires `0 < x`.

**Judgement call.** The OEIS comment is "Starting at other than `a(n)=1`, does this sequence ever
go into a loop?" I read "ever" as existential. A universal reading ("every positive start
eventually loops") is also defensible and would be a one-line change — happy to switch if you
prefer it.

### DIFF
--- a/FormalConjectures/OEIS/100475.lean
+++ b/FormalConjectures/OEIS/100475.lean
@@ -92,12 +92,41 @@ $N, P \in \mathbb{N}$, with $P>0$, such that for all $n \ge N$, $f(n+P) = f(n)$.
 def IsUltimatelyPeriodic (f : ℕ → ℕ) : Prop :=
   ∃ N P, P > 0 ∧ ∀ n, n ≥ N → f (n + P) = f n
 
+/-- The totalized recurrence stays at zero when initialized at zero. The OEIS recurrence itself
+uses one-based prime indices, so this is a boundary behavior of the formalization rather than a
+term of the original sequence. -/
+@[category API, AMS 11]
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #5573, opened 2026-09-11
fix(ErdosProblems/477): such a polynomial and set do exist

### LINKED ISSUE #5281: Erdős Problem 477: status mismatch (repo=open, erdosproblems.com=solved)
The status of [Erdős problem 477](https://www.erdosproblems.com/477) appears to have changed.

- **[This repo](http://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/477.lean)**: `open` (in `FormalConjectures/ErdosProblems/477.lean`)
- **[erdosproblems.com/477](https://www.erdosproblems.com/477)**: `solved`

Please verify and update the `@[category research ...]` annotation if appropriate.

j2d9w5xtjn-png:
fixes #5281

**What changed**

- `erdos_477` is `research solved` with `answer(True)`.
- `erdos_477.variants.monomial` is repaired from a per-`k` assertion into the global question
  `answer(False) ↔ ∀ k ≥ 2, …`, and marked solved.

**Why**

[erdosproblems.com/477](https://www.erdosproblems.com/477) records the problem as solved: contrary
to the expectation of Erdős and Graham, such an `A` does exist, for `f(n) = n^d` with even
`d ≥ 6`. That settles the headline existential affirmatively.

It also refutes the `monomial` variant, which asserted that no such `A` exists for `X^k` for
*every* `k ≥ 2` — the old declaration was parameterised by `k` and therefore asserted the claim
rather than asking it, and as stated it is false at any even `k ≥ 6`. Restating it as
`answer(False) ↔ ∀ k ≥ 2, …` makes one counterexample sufficient.

`erdos_477.variants.X_pow_three` (the `k = 3` case) is untouched and stays `research open`: the
even-`d` construction does not cover it.

### DIFF
--- a/FormalConjectures/ErdosProblems/477.lean
+++ b/FormalConjectures/ErdosProblems/477.lean
@@ -22,6 +22,8 @@ import FormalConjecturesUtil
 *References:*
 - [erdosproblems.com/477](https://www.erdosproblems.com/477)
 - [Sek59](http://dml.cz/dmlcz/100376) Milan Sekanina, Замечания к фактoризации беcкoнечнoй цикличеcкoй группы, Czechoslovak Mathematical Journal, Vol. 9 (1959), No. 4, 485–495
+- The resolution is recorded at [erdosproblems.com/477](https://www.erdosproblems.com/477), with a
+  proof exposition by T. F. Bloom of a construction found independently by several provers.
 -/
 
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5574, opened 2026-09-11
fix(ErdosProblems/633): the classification of triangles is known

j2d9w5xtjn-png:
**What changed**

- `erdos_633` is `research solved`, with the answer slot left as `sorry` and a localised
  `linter.style.category_answer` suppression.
- Beeson–Laczkovich–Zhang added to the references.

**Why**

[arXiv:2604.03609](https://arxiv.org/abs/2604.03609), *Solution of Erdős Problem 633* by Michael
Beeson, Miklós Laczkovich and Yan X. Zhang: "We classify triangles that can be tiled only into a
square number of congruent triangles, settling Erdős Problem 633."
[erdosproblems.com/633](https://www.erdosproblems.com/633) likewise records the problem as solved.

The answer here is a *set* of similarity classes of triangles, and the classification is too
involved to encode in the answer slot, so it is left as `sorry`. This follows the existing pattern
in the repository for solved problems whose answer resists a compact closed form — see
`ErdosProblems/416.lean`, `975.lean`, `1119.lean`, `1175.lean` and
`WrittenOnTheWallII/GraphConjecture34.lean`, which use the same localised suppression.

If you would rather keep this `research open` until someone encodes the classification, that is a
reasonable alternative and I am happy to close this.

### DIFF
--- a/FormalConjectures/ErdosProblems/633.lean
+++ b/FormalConjectures/ErdosProblems/633.lean
@@ -22,6 +22,8 @@ import FormalConjecturesUtil
 * [erdosproblems.com/633](https://www.erdosproblems.com/633)
 * [So09] Soifer, Alexander, How Does One Cut a Triangle? I
 * [So09c] Soifer, Alexander, Is there anything beyond the solution?
+* [BLZ26] Beeson, Michael, Laczkovich, Miklós, and Zhang, Yan,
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5576, opened 2026-09-11
Add the regulator of an elliptic curve

Multramate:
Adds `FormalConjecturesForMathlib/AlgebraicGeometry/EllipticCurve/Regulator.lean`, which supplies the regulator needed to state the strong Birch and Swinnerton-Dyer conjecture.

## What changed

- `naiveHeight`: the logarithmic height of the `x`-coordinate, with value zero at infinity.
- `canonicalHeight`: the limit of `4⁻ⁿhₓ(2ⁿP)`.
- `neronTatePairing`: the bilinear pairing obtained from the canonical height by polarisation.
- `neronTateMatrix` and `regulator`: the Gram matrix of that pairing on a family of points, and the absolute determinant of the Gram matrix on a basis of `E(K)` modulo torsion.

## Formalisation notes

- The definitions are adapted from the Apache-2.0-licensed Tau Ceti files `MordellWeil/NaiveHeight.lean`, `CanonicalHeight.lean`, `MordellWeil/PointModTorsion.lean` and `MordellWeil/Regulator.lean`. The module docstring records this.
- Over `ℚ` the canonical height here is twice Tau Ceti's, so a rank-`r` regulator is `2ʳ` times theirs. This gives the non-normalised regulator that appears in the Birch and Swinnerton-Dyer formula.
- `regulator` assumes `Module.Finite ℤ (W.Point ⧸ Submodule.torsion ℤ W.Point)`. That follows from the Mordell--Weil theorem, which Mathlib does not have yet, so it is an instance argument rather than a hypothesis derived here.

## Checks

`lake --wfail build FormalConjecturesForMathlib.AlgebraicGeometry.EllipticCurve.Regulator` passes. The file contains no `sorry`.


FormalConjecturesForMathlib/AlgebraicGeometry/EllipticCurve/Regulator.lean
```diff
@@ -0,0 +1,82 @@
...
+variable {K : Type*} [DecidableEq K] [Field K] [Height.AdmissibleAbsValues K] {W : Affine K}
+  (P Q : W.Point)
+
+/-- The logarithmic height `hₓ` of the `x`-coordinate with value zero at infinity. -/
+def naiveHeight : W.Point → ℝ
+  | .zero => 0
+  | .some x .. => Height.logHeight₁ x
+
+/-- The canonical height `lim_n 4⁻ⁿhₓ(2ⁿP)` in terms of the logarithmic height `hₓ`. -/
+def canonicalHeight : ℝ := Filter.atTop.limUnder fun n : ℕ ↦ (2 ^ n • P).naiveHeight / 4 ^ n
```
BarinderBanwait, line 67: this looks like the correct definition if it converges, but as P seems to be only a point on an affine curve and not an elliptic curve, i'm concerned that this gives the junk value when the limit doesn't exist. TauCeti has `tendsto_naiveHeight_two_pow_nsmul_div_four_pow` to prove convergence for elliptic curves

FormalConjecturesForMathlib/AlgebraicGeometry/EllipticCurve/Regulator.lean
```diff
@@ -0,0 +1,82 @@
...
+
+/-- The Gram matrix of the Néron--Tate height pairing on any family of `K`-rational points. -/
+def neronTateMatrix {ι : Type*} (P : ι → W.Point) : Matrix ι ι ℝ :=
+  @Matrix.gram _ _ ℝ ⟨neronTatePairing⟩ P
+
+variable (W) in
+/-- The absolute determinant of the height pairing on a finite basis of `E(K)` modulo torsion.
+For number fields with their standard height instance, this is the non-normalised regulator. -/
+def regulator [Module.Finite ℤ (W.Point ⧸ Submodule.torsion ℤ W.Point)] : ℝ :=
+  |(neronTateMatrix (Quotient.out ∘ Module.finBasis ℤ (W.Point ⧸ Submodule.torsion ℤ W.Point))).det|
```
BarinderBanwait, line 80: `Module.finBasis` uses `Module.Free.chooseBasis` which takes an arbitrary choice of basis that we can't say anything about, and i'm a bit worried about that. Again Tau Ceti provides `regulator_eq_abs_det_neronTateGramMatrix` to get the regulator from a given basis. Maybe that's ok?

FormalConjecturesForMathlib/AlgebraicGeometry/EllipticCurve/Regulator.lean
Multramate: Yeah, this should give the junk value when the limit doesn't exist, but that's the point --- I don't want a typeclass encoded in the definition, and our BSD statements do accept [W.IsElliptic], and we don't want to assume Tau Ceti.

FormalConjecturesForMathlib/AlgebraicGeometry/EllipticCurve/Regulator.lean
Multramate: I don't remember my linear algebra too well, but doesn't the choice of basis not matter? Like if the rank is 2 and we have some basis P and Q, then surely P + Q and Q is also a basis, and the regulator will be the same?

BarinderBanwait: yes, for sure, that's true; the definition is mathematically fine; i'm just concerned that someone wants to prove that the regulator of E is equal to the determinant of the Neron--Tate height matrix of their favourite MW basis, and not being able to. in other words, what is missing here is a lemma or theorem allowing one to compute the regulator from a given basis. anyway i don't think it matters since we know the regulator is independent of the MW basis

Multramate: Yeah that's fair, but we're in FC and not mathlib, so maybe this is OK :)

BarinderBanwait (approved):
LGTM!

mo271 (approved):

PR #5578, opened 2026-09-11
Add the Tate--Shafarevich group of an elliptic curve

Multramate:
Adds `FormalConjecturesForMathlib/AlgebraicGeometry/EllipticCurve/TateShafarevich.lean`, which supplies the Tate--Shafarevich group needed to state the strong Birch and Swinnerton-Dyer conjecture.

## What changed

- `galoisRepresentation`: the natural action of the absolute Galois group on the geometric non-singular points, given the discrete topology.
- `tateShafarevich`: the Tate--Shafarevich group over a number field, as the intersection of the kernels of restriction to every infinite and finite completion, inside the first continuous Galois cohomology group.

Two supporting files are needed for this to compile and are included in the same pull request:

- `FormalConjecturesForMathlib/RepresentationTheory/Homological/ContCohomology/Sha.lean` defines the Tate--Shafarevich group of a general continuous representation, as the intersection of the kernels of `Hⁿ(K, A) → Hⁿ(Kᵥ, Aᵥ)`. It is adapted from [mathlib4#43529](https://github.com/leanprover-community/mathlib4/pull/43529).
- `FormalConjecturesForMathlib/FieldTheory/AbsoluteGaloisGroup.lean` supplies the restriction maps of absolute Galois groups that `Sha.lean` uses. They are copied from [mathlib4#41415](https://github.com/leanprover-community/mathlib4/pull/41415), which is not in the `v4.33.1` Mathlib pinned here.

Both carry their upstream attribution in the file header. They can be split into separate pull requests if you prefer to review them independently.

## Formalisation notes

- The coefficient module uses `AlgebraicClosure K`, to match `Field.absoluteGaloisGroup`. Number fields are perfect, so an algebraic closure is also a separable closure. A `TODO` in the file records that `SeparableClosure K` is the right choice over an imperfect field, once the absolute Galois group API supports it directly.
- Finite places are indexed by the non-zero prime ideals of the ring of integers, and infinite places by `InfinitePlace`; the group is the intersection of the two conditions.

## Checks

`lake --wfail build FormalConjecturesForMathlib.AlgebraicGeometry.EllipticCurve.TateShafarevich` passes, which covers all three files. None of them contains a `sorry`.


FormalConjecturesForMathlib/AlgebraicGeometry/EllipticCurve/TateShafarevich.lean
```diff
@@ -0,0 +1,83 @@
...
+
+TODO: use `SeparableClosure K` for the coefficient field when the API for absolute Galois groups
+supports this directly. The automorphism group of the algebraic closure itself gives the usual
+absolute Galois group by restriction to the separable closure. -/
+noncomputable def galoisRepresentation : TopRep ℤ <| Field.absoluteGaloisGroup K := .of <|
+  .ofMonoidHom { toFun σ := { __ := (Affine.Point.map (W' := W) σ.toAlgHom).toIntLinearMap
+                              cont := continuous_of_discreteTopology }
+                 map_one' := by ext P; cases P <;> rfl
+                 map_mul' _ _ := by ext P; cases P <;> rfl }
+
```
Multramate, line 72:
```suggestion
```

PR #4382, opened 2026-07-03
feat(ErdosProblems/1): link a complete formal proof of lb_strong

MyTH-zyxeon:
Adds a formal_proof attribute to erdos_1.variants.lb_strong pointing to a complete, sorry-free Lean 4 proof hosted at a pinned commit of this fork (branch erdos1-lb-strong, commit 362eed1a).

## details

This PR adds a `formal_proof` attribute to `erdos_1.variants.lb_strong` — the √(2/π) asymptotic lower bound for the distinct subset sums problem (currently a `sorry` stub).

A complete, sorry-free Lean 4 proof of a statement verbatim-identical to `lb_strong` is hosted at a pinned commit of my fork, [`lb_strong_proved`](https://github.com/MyTH-zyxeon/formal-conjectures/blob/362eed1a8864d142ae65f51af7981a7c7530956e/FormalConjectures/Scratch/HarperCompression.lean#L2174). Following CONTRIBUTING's guidance that proofs longer than 25–50 lines should be hosted externally and linked via the `formal_proof` mechanism, this PR only adds the attribute.

About the proof (~2,700 lines across three files; `lake --wfail build` green on this repository's toolchain/Mathlib pin; no `sorry`/`admit`/`axiom`):
- **Stirling analysis** - central binomial asymptotics `C(2m,m) ~ 4^m/√(πm)` via a `stirlingSeq`-ratio argument, giving `C(n,⌊n/2⌋)·√n/2ⁿ → √(2/π)`.
- **The Dubroff–Fox–Xu reduction** (arXiv:2006.12988, second proof) - the exact bound `N ≥ C(n,⌊n/2⌋)` from the half-cube case of Harper's theorem.
- **A self-contained compression proof of Harper's vertex-isoperimetric theorem** (simplicial order, i-compressions, fixed-point dichotomy including the exceptional half-size family) — to my knowledge not currently available in Mathlib.

- **AI disclosure** - this proof was developed by multiple AI agents in an implement/review loop (every increment independently reviewed and rebuilt by two reviewer agents), under my supervision, following the guidelines in this repository's AGENTS.md. I reviewed this submission.

PR #5572, opened 2026-09-11
fix(ErdosProblems/421): a density-one sequence with distinct products exists

j2d9w5xtjn-png:
fixes [#5379](https://github.com/google-deepmind/formal-conjectures/issues/5379)

**What changed**

- `erdos_421` is `research solved` with `answer(True)`.
- Pratt's write-up added to the references.

**Why**

[erdosproblems.com/421](https://www.erdosproblems.com/421) records the problem as solved and states
plainly: "The answer is yes." A density-one sequence all of whose consecutive products
`∏_{u ≤ i ≤ v} d_i` are distinct does exist.

The page notes the provenance is "a little messy" — an approach sketched by Tao, fleshed out by
Chojecki, with full proofs claimed independently by several people — and that a simplified proof
exists. A write-up by Kyle Pratt is hosted on the problem page and is cited here.

*AI assistance: this was found by an OpenAI Codex agent during a repository-wide sweep of open declarations. The statement and the cited sources were then independently re-checked and verified with Claude Opus 5 before submission.*

PR #5571, opened 2026-09-11
fix(ErdosProblems/346): the golden-ratio limit fails for a lacunary counterexample

j2d9w5xtjn-png:
fixes [#5238](https://github.com/google-deepmind/formal-conjectures/issues/5238)

**What changed**

- `erdos_346` is `research solved` with `answer(False)`.
- The docstring records the counterexample.

**Why**

[erdosproblems.com/346](https://www.erdosproblems.com/346) records the problem as solved and states
the answer is negative: a sequence `A` with the required completeness properties was constructed
with `a_{n+1}/a_n ≥ 6/5` for all `n`, and an increasing sequence `N_j` with

```
a_{N_j}/a_{N_j-1} → φ        and        a_{N_j+1}/a_{N_j} → φ + 1/4
```

Two distinct subsequential limits means `a_{n+1}/a_n` does not converge, so the displayed limit
`(1+√5)/2` fails. The declaration is an `answer(sorry) ↔ ∀ …` question, so the counterexample gives
`answer(False)` directly.

*AI assistance: this was found by an OpenAI Codex agent during a repository-wide sweep of open declarations. The statement and the cited sources were then independently re-checked and verified with Claude Opus 5 before submission.*

PR #5568, opened 2026-09-11
fix(OEIS/1818): both Sun permanent conjectures are proved

j2d9w5xtjn-png:
**What changed**

- `conjecture1` and `conjecture2` are `research solved`.
- Both papers added to the reference list.

**conjecture1** — She, Sun and Xia, *A novel permanent identity with applications*,
[arXiv:2208.12167](https://arxiv.org/abs/2208.12167), Theorem 1.3(ii):

> Let `n > 1` be an integer, and let `ζ` be a primitive `n`th root of unity. Set
> `c_{j,k} = (1+ζ^{j-k})/(1-ζ^{j-k})` if `j ≠ k`, and `1` if `j = k`.
> If `n` is even, then `per[c_{j,k}]_{1≤j,k≤n} = ((n-1)!!)²`.

Taking `n = 2m` gives exactly this file's statement: the `2m × 2m` permanent equals
`((2m-1)!!)² = a m`. The paper records that this part "was first conjectured by the second
author", i.e. Zhi-Wei Sun, matching the attribution in the docstring.

**conjecture2** — Yang and Zhang, *Sun-type determinant and permanent congruences*,
[arXiv:2605.19502](https://arxiv.org/abs/2605.19502), Proposition 17:

> For every odd prime `p`, `per(I_{p-1} + R_p) ≡ ((p-2)!!)² (mod p²)`.

`I_{p-1} + R_p` is the `(p-1) × (p-1)` matrix with `1` on the diagonal and `(j+k)/(j-k)` off it —
this file's `fEntry` — and `((p-2)!!)² = a ((p-1)/2)`. Their proof combines She–Sun–Xia's permanent
identity with their own Cauchy permanent congruence and Morley's congruence.

Both declarations keep their literature-backed `sorry`, as is usual in this repository for results
whose proofs are not formalised.

*AI assistance: this was found by an OpenAI Codex agent during a repository-wide sweep of open declarations. The statement and the cited sources were then independently re-checked and verified with Claude Opus 5 before submission.*

PR #5581, opened 2026-09-11
fix(ErdosProblems/1): the distinct subset sums conjecture is false

j2d9w5xtjn-png:
**What changed**

- `erdos_1` is `research solved`, restated as the negation of the original claim, with a
  `formal_proof using lean4` attribute.
- `erdos_1.variants.real` is `research solved`, likewise restated, with a complete Lean reduction
  to `erdos_1` (no new `sorry`).

**Status and attribution**

[erdosproblems.com/1](https://www.erdosproblems.com/1) records this as **DISPROVED (LEAN)** — its
tag for "solved in the negative and the proof verified in Lean". The page credits the disproof to
**GPT-6 Astra**, which showed that for any `ε > 0` there are arbitrarily large `n` with
`N ≤ ε·2^n`. Since Erdős asked for `N ≫ 2^n`, no positive constant survives, so the existential
form in this file is false.

The Lean development is hosted at a pinned commit of
[`tadamcz/erdos1`](https://github.com/tadamcz/erdos1/blob/0e395153306f34b3829d118b85bdd704136f2843/Erdos1/Resolutions/Erdos1_219usd_38h.lean#L2419).

**Audit of the linked proof**

- The file contains no `sorry`, no `axiom`, no `native_decide`, no `unsafe`, no `implemented_by`.
- Its `IsSumDistinctSet` is **verbatim identical** to this file's, and the linked theorem is
  literally `¬ (∃ C > (0 : ℝ), ∀ (N : ℕ) (A : Finset ℕ) (_ : IsSumDistinctSet A N), N ≠ 0 →
  C * 2 ^ A.card < N)` — the exact negation of the statement here. Its own docstring says so.

**The real variant** is proved here rather than linked: map a natural counterexample into `ℝ` by
the cast embedding; distinct integer subset sums differ by at least `1`, which supplies the real
separation hypothesis, so a real bound would give an integer bound.

**Note on overlap.** Open PR #4382 also touches this file, but adds a `formal_proof` link to
`erdos_1.variants.lb_strong`, a different declaration. No semantic conflict, though the two may
need a trivial rebase.

*AI assistance: the match between this declaration and the external proof was found by an OpenAI Codex agent during a repository-wide sweep. The linked Lean proof itself was generated by **GPT-6 Astra** in the FrontierMath Erdős benchmark (Adamczewski and Bloom, 2026). The statement match and the `sorry`/axiom audit of the linked proof were carried out with Claude Opus 5.*

### DIFF
--- a/FormalConjectures/ErdosProblems/1.lean
+++ b/FormalConjectures/ErdosProblems/1.lean
@@ -41,9 +41,13 @@ distinct for all $S\subseteq A$ then
 $$
   N \gg 2 ^ n.
 $$
+
+This conjecture is false. A machine-checked disproof constructs sum-distinct sets for which
+$N / 2^n$ tends to zero.
 -/
-@[category research open, AMS 5 11]
-theorem erdos_1 : ∃ C > (0 : ℝ), ∀ (N : ℕ) (A : Finset ℕ) (_ : IsSumDistinctSet A N),
+@[category research solved, AMS 5 11, formal_proof using lean4 at
+  "https://github.com/tadamcz/erdos1/blob/0e395153306f34b3829d118b85bdd704136f2843/Erdos1/Resolutions/Erdos1_219usd_38h.lean#L2419"]
+theorem erdos_1 : ¬ ∃ C > (0 : ℝ), ∀ (N : ℕ) (A : Finset ℕ) (_ : IsSumDistinctSet A N),
     N ≠ 0 → C * 2 ^ A.card < N := by
   sorry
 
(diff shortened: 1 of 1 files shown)


tadamcz:
@mo271 this one (and probably several others from @j2d9w5xtjn-png about erdos problems) should not be tagged misformalization

j2d9w5xtjn-png:
@tadamcz agreed, and thank you — the label is wrong on this one and on most of the batch. It looks
auto-applied from the `fix(...)` title prefix. I don't have permission to change labels here, so
here is the breakdown in case it saves you or @mo271 the triage.

**Not misformalizations — the statement was fine, the answer became known** (23):

| | |
|---|---|
| status flip only, statement untouched | #5557, #5560, #5568, #5574, #5589, #5590, #5592, #5593, #5594, #5595, #5596, #5604 |
| disproof, recorded by restating as the negation | #5586, #5587, #5588, #5591 |
| answer slot filled (`answer(True)`/`answer(False)`) | #5556, #5561, #5571, #5572 |
| external `formal_proof` link added | #5581, #5582, #5583 |

**Genuinely misformalizations — the Lean did not represent the informal statement** (11):

- #5550 — `a, b` quantified inside the set-builder, so they may vary with `p`; makes it elementary
- #5551 — no openness condition, so the Bloch constant collapsed onto the Landau constant
- #5552 — the file asserted a proposition *and* its negation as two `research open` theorems
- #5553 — `exists_better_bound` is satisfied by the bound it was meant to improve on
- #5554 — parameterised by `n` and `C`, so it asserted the claim rather than asking it
- #5555, #5566 — "the only zero/square" stated as `→`, which is vacuous without existence
- #5559 — supremum taken over `Icc (1/2) 1`; the source uses `[-1/2, 1/2]`
- #5565 — inherited two errors from the OEIS comment, and searched from the wrong index
- #5567 — `answer( )` under a binder, plus a degenerate `x = 0` start from totalisation
- #5573 — the `monomial` variant was parameterised by `k` and false at even `k ≥ 6`

**Neither** (1): #5603 is purely an attribution fix (the `k ≥ 6` case is due to Luo, not "Liu")
plus the missing citation.

Several in the first group would fit `solution found`, which a few of them already carry.

If it would reduce noise, I am happy to close any subset and resubmit them differently — or to stop
opening new ones until the current batch has been triaged. I opened rather a lot at once, and I'd
rather not create work.

FormalConjectures/ErdosProblems/1.lean
```diff
@@ -107,14 +111,107 @@ abbrev IsSumDistinctRealSet (A : Finset ℝ) (N : ℕ) : Prop :=
 A generalisation of the problem to sets $A \subseteq (0, N]$ of real numbers, such that the subset
 sums all differ by at least $1$ is proposed in [Er73] and [ErGr80].
 
+The positive statement is false: every natural-number counterexample to `erdos_1` embeds into
+`ℝ`, and distinct integer subset sums differ by at least one.
```
mo271:
```suggestion
$\mathbb{R}$, and distinct integer subset sums differ by at least one.
```

mo271 (approved):
I think a good follow-up might be to add a link to https://epoch.ai/latest/announcing-frontiermath-erdos 
is there an arxiv-id for this paper?
Perhaps @tadamcz could say the way they prefer to get cited for the Erdosproblems that got solved when creating that dataset? And while add it check if we mark those 5 that GPT-6 Astra could solve.

tadamcz:
I'm still waiting for it to appear on the arxiv. Will update when I have it.

PR #5586, opened 2026-09-11
fix(OEIS/1223): prime-gap blocks need not recur infinitely often

j2d9w5xtjn-png:
**What changed**

- `conjecture` is restated as the negation of the original universal claim and marked

**Why**

The OEIS comment asks whether every gap block `a(n..n+m)` with `n > 2` recurs infinitely often
at other starting points. **It does not.** The five-term block beginning at `n = 3` is
`(2, 4, 2, 4, 2)`, and a congruence argument modulo `5` shows it occurs only at `n = 3`, so the
set of matching starting points is a singleton rather than infinite.

**Audit of the linked proof**

- Its `a` and `gap_subsequence` are verbatim identical to this file's `a` and `gapSubsequence`.

*AI assistance: the match between this declaration and the external proof was found by an OpenAI Codex agent during a repository-wide sweep. The linked Lean proof itself was generated by **GPT-5.5** in the OEIS Open benchmark ([Adamczewski, 2026](https://arxiv.org/pdf/2608.11941)); its `scores.json` records the verifier verdict `C` (accepted). The statement match and the `sorry`/axiom audit of the linked proof were carried out with Claude Opus 5.*

### DIFF
--- a/FormalConjectures/OEIS/1223.lean
+++ b/FormalConjectures/OEIS/1223.lean
@@ -61,10 +61,15 @@ theorem a_3 : a 3 = 2 := by
 /--
 Any subsequence a(n .. n+m) with n > 2 (as to exclude the
 untypical primes 2 and 3) should occur infinitely many times at other starting points k.
+
+This is false. The five-term block starting at $n = 3$ is $(2,4,2,4,2)$, and a congruence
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #5588, opened 2026-09-11
fix(OEIS/46969): conjecture2 fails at n = 236791

j2d9w5xtjn-png:
**What changed**

- `conjecture2` is restated as the negation of the original universal claim and marked

**Why**

Conjecture II asserts a trio of divisibility-by-6 congruences whenever the stated hypotheses
hold. **It fails at `n = 236791`**, where the hypotheses hold but the middle congruence does not.

**Audit of the linked proof**

- Its `a` is verbatim identical to this file's Bernoulli-denominator definition. The
  external statement writes the hypotheses as `a n % 12 = 0` where this file writes `12 ∣ a n`;
  these are the same condition on naturals.


mo271 (approved):
Thanks!

PR #5589, opened 2026-09-11
fix(OEIS/48153): link a proof of the (n^2-1)/2 bound

j2d9w5xtjn-png:
**What changed**

- `conjecture` is marked `research solved` with a `formal_proof using lean4` attribute. The
  statement is unchanged — the conjectured bound holds.

**Why**

Aspen A.M. Meissner's conjecture that `a(n) ≤ (n² - 1)/2` is **true**, and the linked development
proves it for all `n ≥ 1`.

**Audit of the linked proof**

- The external `A048153` sums over `Finset.range n` where this file sums over `Finset.Icc 1 n`.
  These agree: the `k = 0` term is `0`, and the `k = n` term is `n² mod n = 0`, so the two ranges
  differ only by terms that vanish.


mo271 (approved):
Thanks!

PR #5591, opened 2026-09-11
fix(OEIS/64169): the prime criterion fails at 16843^2

j2d9w5xtjn-png:
**What changed**

- `conjecture` is restated as the negation of the original universal claim and marked

**Why**

Eldar and Ordowski conjectured that for `n > 2`, `n ∣ a(n-2)` exactly when `n` is prime, checked
to 20000. **It fails at `n = 16843²`**, the square of a Wolstenholme prime, which is composite but
divides the corresponding term.

**Audit of the linked proof**

- This file's `a` returns `ℤ` (`(H n).num - (H n).den`); the external `A064169` returns `ℕ` via
  `Int.natAbs` of the same difference, using Mathlib's `harmonic`, which is this file's `H`.
  Divisibility by `n` is unaffected by the absolute value.

### DIFF
--- a/FormalConjectures/OEIS/64169.lean
+++ b/FormalConjectures/OEIS/64169.lean
@@ -57,9 +57,12 @@ theorem a_5 : a 5 = 77 := by
 
 /--
 "Conjecture: for $n > 2$, $n$ divides $a(n-2)$ if and only if $n$ is a prime.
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5592, opened 2026-09-11
fix(OEIS/93456): infinitely many terms are not divisible by the previous one

j2d9w5xtjn-png:
**What changed**

- `conjecture` is restated as the negation of the original finiteness claim and marked

**Why**

The conjecture is that only finitely many `n` satisfy `a(n-1) ∤ a(n)`. **There are infinitely
many.** The linked proof produces, above any bound, a suitable `n` via a large prime.

**Audit of the linked proof**

- Its `a` is verbatim identical to this file's product of composites in `Icc (C(n,2)+1) (C(n+1,2))`.


mo271 (approved):
Thanks!

PR #5596, opened 2026-09-11
fix(OEIS/185895): sign changes occur exactly at triangular numbers

j2d9w5xtjn-png:
**What changed**

- `conjecture1` is marked `research solved` with a `formal_proof using lean4` attribute. The
  statement is unchanged.

**Why**

Peter Bala's conjecture is **true**: `a(n)` differs in sign from `a(n-1)` exactly when `n` is a
triangular number. It had been checked only to `n = 1225`.

**Audit of the linked proof**

- Its `A185895` is this file's `a`: the `n!`-scaled coefficient of `xⁿ` in
  `∏_{k=1}^{n} (1 - xᵏ/k!)`, taken as an integer (the coefficients are integers, so the external
  `.num` and this file's `.floor` agree).

*AI assistance: the match between this declaration and the external proof was found by an OpenAI Codex agent during a repository-wide sweep. The linked Lean proof itself was generated by **GPT-5.6 Sol** in the OEIS Open benchmark ([Adamczewski, 2026](https://arxiv.org/pdf/2608.11941)); its `scores.json` records the verifier verdict `C` (accepted). The statement match and the `sorry`/axiom audit of the linked proof were carried out with Claude Opus 5.*

### DIFF
--- a/FormalConjectures/OEIS/185895.lean
+++ b/FormalConjectures/OEIS/185895.lean
@@ -103,7 +103,8 @@ $a(n)$ differs in sign from $a(n-1)$ if and only if $n$ is a triangular number
 (checked up to $n = 1225 = (50 \cdot 51)/2$).
 - _Peter Bala_, Mar 17 2022
 -/
-@[category research open, AMS 11]
+@[category research solved, AMS 11,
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5597, opened 2026-09-11
feat(Arxiv/2010.00102): state the modular Schanuel conjecture

SmwYin:
States the modular Schanuel conjecture, with and without derivatives. Adds to FormalConjecturesForMathlib JFunction.lean, because this will also be needed for a future PR on stating the modular Zilber-Pink conjecture, which I will make once this PR is merged.


SmwYin:
Oh unfortunately Claude hasn't signed a CLA so I'll have to make a new PR by hand.

PR #5721, opened 2026-09-11
feat(Kourovka): formalize Problem 8.8

homeowmorphism:
Formalizes part b) of Kourovka Notebook Problem 8.8 (Anosov, via Grigorchuk, 8th edition, 1982).

Does there exist a non-cyclic finitely presented group $G$ with an element $a$ such that every element of $G$ is conjugate to some power of $a$?

Part a), the finitely generated version, was answered by Guba (1986) and is recorded in the docstring. Part b) is open as of Kourovka Notebook v46 (September 2026).

Finite presentability is Mathlib's `Group.IsFinitelyPresented` (available since the v4.33.1 bump), so the PR adds a single file, `FormalConjectures/Kourovka/8_8.lean`. An earlier revision carried a local copy of that definition under `FormalConjecturesForMathlib`; it has been removed.

### DIFF
--- a/FormalConjectures/Kourovka/8_8.lean
+++ b/FormalConjectures/Kourovka/8_8.lean
@@ -43,7 +43,7 @@ Here a power of $a$ means $a^n$ for some $n \in \mathbb{Z}$.
 -/
 @[category research solved, AMS 20]
 theorem kourovka.«8.8».parts.b : answer(sorry) ↔
-    ∃ (G : Type) (_ : Group G), Group.IsFinitelyPresented G ∧ ¬ IsCyclic G ∧
+    ∃ (G : Type) (_ : Group G), ¬ IsCyclic G ∧ Group.IsFinitelyPresented G
       ∃ a : G, ∀ g : G, ∃ n : ℤ, IsConj g (a ^ n) := by
   sorry
 
--- a/FormalConjectures/Kourovka/8_8.lean
+++ b/FormalConjectures/Kourovka/8_8.lean
@@ -1,0 +1,53 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


FormalConjectures/Kourovka/8_8.lean
```diff
@@ -0,0 +1,58 @@
...
+containing an element $a$ such that every element of $G$ is conjugate to a
+power of $a$. It was answered affirmatively by V. S. Guba
+(Math. USSR-Izv. 29 (1986), 233–277) and now sits in the Archive of the
+notebook.
+
+Part b) asks the same question for finitely presented groups:
+
+Does there exist a non-cyclic finitely presented group $G$ which contains an
+element $a$ such that each element of $G$ is conjugate to some power of $a$?
+
```
mo271:
```suggestion
```

not need to duplicate that, we already have this below

mo271 (approved):
Thanks, LGTM!

If not much more effort also having "Part a)" as a `research solved` problem would also be great, but I'm happy to merge with only part b being there as well.

mo271:
The CLA bot should get green once the commit that has claude as co-author is amended to remove the Co-authored-by line.

FormalConjectures/Kourovka/8_8.lean
homeowmorphism: Thanks I have removed it.

homeowmorphism:
@mo271 Thanks for looking at this! I should've gotten rid of the CLA and added 8a.

mo271 (approved):
Thanks, Looks great!

A possible follow up would be a the implication b => a, which would be a nice test statement. 
I think it would need smething like
```
instance (priority := 100) Group.fg_of_isFinitelyPresented
    [h : Group.IsFinitelyPresented G] : Group.FG G := by
  obtain ⟨n, φ, hφ, -⟩ := h.out; exact Group.fg_of_surjective hφ
```
which seems to be missing from (For)Mathlib

PR #5723, opened 2026-09-11
fix(Wikipedia/RegularPrimes): restrict Kummer's criterion to even Bernoulli indices

### LINKED ISSUE #5711: Wikipedia/RegularPrimes: `isRegularPrime_iff_Bernoulli` includes odd Bernoulli indices, so every p ≥ 7 is irregular
`RegularPrimes.isRegularPrime_iff_Bernoulli` requires `¬ (p : ℤ) ∣ (bernoulli' k).num` for every `k ∈ Finset.Icc 2 (p - 3)`, including odd `k` such as `k = 3` where `bernoulli' 3 = 0`, whereas the source defines a regular prime by non-divisibility of the numerators of the even-indexed Bernoulli numbers B₂, B₄, …, Bₚ₋₃ only, so the stated condition fails for every prime p ≥ 7.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/Wikipedia/RegularPrimes

- 1 misformalization
- 1 equivalent reformulation
- trivial proof compiles
- fix compiles

smmercuri:
`isRegularPrime_iff_Bernoulli` quantified over every index `k ∈ [2, p - 3]`. Mathlib's `bernoulli' k` is `0` for odd `k ≥ 3`, so `(p : ℤ) ∣ (bernoulli' 3).num` holds for every `p`; for `p ≥ 7` the index `3` lies in the range and the right-hand side was false, i.e. the statement declared every prime `p ≥ 7` irregular (contradicting `small_regular_primes`).

**Change:** Kummer's criterion only involves the even-indexed Bernoulli numbers `B_2, B_4, …, B_{p-3}`, so add `Even k →` inside the quantifier and state the criterion explicitly in the docstring. Attributes unchanged.

**Checked:** `lake --wfail build FormalConjectures.Wikipedia.RegularPrimes` passes with no warnings.

Fixes #5711

### DIFF
--- a/FormalConjectures/Wikipedia/RegularPrimes.lean
+++ b/FormalConjectures/Wikipedia/RegularPrimes.lean
@@ -54,11 +54,12 @@ lemma small_regular_primes :
     { 2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31 } ⊆ regularPrimes := by
   sorry
 
-/-- An equivalent definition of a regular prime `p` is that it does not divide the numerator of the
-first `p-3` Bernoulli numbers. Not in Mathlib. -/
+/-- **Kummer's criterion.** An equivalent definition of a regular prime `p` is that it does not
+divide the numerator of any of the Bernoulli numbers $B_k$ for $k = 2, 4, 6, \dots, p - 3$.
+Not in Mathlib. -/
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5724, opened 2026-09-11
fix(GreensOpenProblems/16): forbid all nontrivial Zhao solutions

### LINKED ISSUE #5679: GreensOpenProblems/16: `ZhaoSolutionFree` admits nontrivial solutions; the conjectured lower bound is unsourced
`Green16.ZhaoSolutionFree` forbids only solutions of `x + 2y + 3z = x' + 2y' + 3z'` with six pairwise distinct values, whereas the source forbids every nontrivial solution (one whose coefficients do not cancel within each equal-value class), so `Green16.g` and `Green16.zhao_question` concern the wrong avoidance condition, and `Green16.green_16_conjectured_lower_bound` asserts f(N) ≫ N exp(-c(log N)^(1/7)), which Green states only as an upper bound.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/GreensOpenProblems/16

- 2 misformalizations
- 2 status issues
- 1 equivalent reformulation
- trivial proof compiles
- fix compiles

smmercuri:
`Green16.ZhaoSolutionFree` forbade only solutions of $x + 2y + 3z = x' + 2y' + 3z'$ with six pairwise distinct values. Green's Problem 16 comments ask for a subset of $[N]$ of size $N^{1/3 - o(1)}$ with no *nontrivial* solutions, where (following Ruzsa) a solution is trivial when the coefficients of the variables taking each value sum to the same amount on both sides. For example $\{1, 2, 3\}$ satisfied the old predicate despite the nontrivial solution $(1, 1, 3, 2, 2, 2)$. The file also asserted a lower bound $f(N) \gg N e^{-c(\log N)^{1/7}}$ that Green states only as the Schoen–Sisask upper bound.

**Change**
- Add `IsTrivialZhaoSolution` (coefficient sums agree on every equal-value class) and redefine `ZhaoSolutionFree` to require every solution in the set to be trivial; `g` and `zhao_question` now use the intended avoidance condition.
- Remove `green_16_conjectured_lower_bound` (no source poses it).
- Retag `green_16_lower_bound` (Ruzsa) and `green_16_upper_bound` (Schoen–Sisask) as `research solved`, as Green calls them the best known bounds.

**Checked**: `lake --wfail build 'FormalConjectures.GreensOpenProblems.«16»'` passes with no warnings.

Fixes #5679

### DIFF
--- a/FormalConjectures/GreensOpenProblems/16.lean
+++ b/FormalConjectures/GreensOpenProblems/16.lean
@@ -49,28 +49,27 @@ theorem green_16 (N : ℕ) :
   sorry
 
 /-- From [Ruzsa] $f(N) \gg N^{1/2}$. -/
-@[category research open, AMS 5 11]
+@[category research solved, AMS 5 11]
 theorem green_16_lower_bound :
     (fun N ↦ (N : ℝ) ^ (1 / 2 : ℝ)) ≪ fun N ↦ (f N : ℝ) := by
   sorry
 
 /-- From [Schoen and Sisask] $f(N) \ll N \cdot e^{-c(\log N)^{1/7}}$. -/
-@[category research open, AMS 5 11]
+@[category research solved, AMS 5 11]
 theorem green_16_upper_bound :
     ∃ c > (0 : ℝ), (fun N ↦ (f N : ℝ)) ≪ fun N ↦ (N : ℝ) * exp (-c * (log N) ^ (1 / 7 : ℝ)) := by
   sorry
 
-/-- $f(N) \gg N \cdot e^{-c(\log N)^{1/7}}$. -/
-@[category research open, AMS 5 11]
-theorem green_16_conjectured_lower_bound :
-    ∃ c > (0 : ℝ), (fun N ↦ (N : ℝ) * exp (-c * (log N) ^ (1 / 7 : ℝ))) ≪ fun N ↦ (f N : ℝ) := by
-  sorry
+/-- A solution $(x, y, z, x', y', z')$ of $x + 2y + 3z = x' + 2y' + 3z'$ is *trivial* if for every
+value $a$ the coefficients of the variables equal to $a$ have the same sum on both sides. -/
+def IsTrivialZhaoSolution (x y z x' y' z' : ℕ) : Prop :=
+  ∀ a : ℕ, (if x = a then 1 else 0) + (if y = a then 2 else 0) + (if z = a then 3 else 0) =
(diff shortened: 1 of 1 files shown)


smmercuri:

I compared this fix against issue #5679 and the audit's evidence at https://tadamcz.com/fc-review-results/c90271f0fa/#/f/GreensOpenProblems/16, read the primary sources (Green's Problem 16 comments; Ruzsa's paper), and tried to break the new statement. Verdict: **the fix is correct and fully addresses the misformalization.**

Why it is the right fix:

- **The defect is gone.** The old `ZhaoSolutionFree` forbade only solutions with six pairwise distinct values, so it accepted `{1, 2, 3}` despite the nontrivial solution `(1, 1, 3, 2, 2, 2)`. I verified in Lean that the new predicate rejects `{1, 2, 3}`.
- **Equivalent to the audit's proposal.** I proved `Green16.ZhaoSolutionFree A ↔ <audit's ZhaoSolutionFree> A` in Lean against this branch; the PR only extracts `IsTrivialZhaoSolution`, uses `∀ x ∈ A` binders matching `SolutionFree`, and also applies the audit's recommended status retags.
- **Faithful to the source.** Green asks for "no *nontrivial* solutions" (reserving "distinct integers" for Problem 16 itself), and Ruzsa's Definition 1.1 makes a solution trivial exactly when the signed coefficients sum to zero on each equal-value class, which is what `IsTrivialZhaoSolution` says. I also proved it equivalent in Lean to `({x, y, y, z, z, z} : Multiset ℕ) = {x', y', y', z', z', z'}`.
- **No junk values or vacuity.** In Lean: `∅`, singletons and `{1, 2}` remain admissible, `(a, a, b, b, b, a)` is trivial, and `g N ≤ N`; only ℕ-addition occurs. `zhao_question` is not trivialisable (known constructions give only `N^{1/6}` against the `N^{1/3}` counting bound).
- **Every change is justified, and it builds.** Green states `N e^{−c(log N)^{1/7}}` only as the Schoen–Sisask *upper* bound, so removing `green_16_conjectured_lower_bound` is right (nothing else referenced it); the `research solved` retags follow Green's "the best bounds known", with Ruzsa's Theorem 7.3 giving the lower bound. `lake --wfail build` of the module succeeds with no warnings.

Optional nits only (not blocking): the upper-bound `research solved` tag rests on Green's attribution (Schoen–Sisask §9 only sketches the four-variable case); the multiset form above would be a more self-explanatory `IsTrivialZhaoSolution`.

mo271 (approved):
Thanks!

Issue #5719, opened 2026-09-11
WrittenOnTheWallII/GraphConjecture32: `conjecture32` averages vertex-to-`A` distances, not the distances between members of `A`


tadamcz:
`WrittenOnTheWallII.GraphConjecture32.conjecture32` uses `distavg G A`, defined as `(1 / |V|) * ∑ v, min_{a ∈ A} d(v, a)`, the average distance from all vertices to `A`, whereas the source averages the nonzero distances between members of `A`, so on the source's counterexample `P₅` with `A = {0, 4}` the Lean inequality `32/15 ≤ 5` holds while the source's `16/3 ≤ 5` fails.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/WrittenOnTheWallII/GraphConjecture32

- 1 misformalization
- 1 minor
- trivial proof compiles
- fix compiles

<sub>Found by a language-model audit of the repository at c90271f0fa. Not reviewed by a human. This issue's title and description were written by a language model as well.</sub>

PR #5736, opened 2026-09-11
fix(WrittenOnTheWallII/GraphConjecture23): average pairwise distances in M

### LINKED ISSUE #5718: WrittenOnTheWallII/GraphConjecture23: `conjecture23` averages vertex-to-`M` distances, not pairwise distances within `M`
`WrittenOnTheWallII.GraphConjecture23.conjecture23` uses `distavg G M`, the average over all vertices `v` of the distance from `v` to the set `M`, whereas the source averages the distances between distinct vertices both belonging to `M`, and the two inequalities differ: on an 11-vertex spine with four attached triangles and pendant leaves the source's inequality fails while the Lean one holds.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/WrittenOnTheWallII/GraphConjecture23

- 1 misformalization
- trivial proof compiles
- fix compiles

smmercuri:
`conjecture23` used `distavg G M`, which is the average over *all* vertices `v` of `G` of the distance from `v` to the set `M` of maximum-degree vertices. WOWII's `dist_avg(M)` is instead the average distance between distinct vertices both lying in `M`. The two inequalities are genuinely different (the audit gives a graph where the source's inequality fails but the Lean one holds).

**Change:** add `SimpleGraph.distAvgSet G S : ℝ` to `FormalConjecturesForMathlib/Combinatorics/SimpleGraph/VertexDistance.lean` — the mean of `G.dist u v` over ordered pairs `(u, v)` of distinct vertices of `S`, defaulting to `0` when `S` has fewer than two vertices — and restate `conjecture23` with `distAvgSet G M`; docstrings updated. The existing `distavg` (vertex-to-set average) is left untouched since other conjectures use it.

**Checked:** `lake --wfail build FormalConjectures.WrittenOnTheWallII.GraphConjecture23` passes with no warnings.

Note for reviewers: the same "average distance within a set" notion is defined by a local `let` in the fix for #5719; if both land, that one could be switched to `distAvgSet`.

Fixes #5718

### DIFF
--- a/FormalConjectures/WrittenOnTheWallII/GraphConjecture23.lean
+++ b/FormalConjectures/WrittenOnTheWallII/GraphConjecture23.lean
@@ -32,20 +32,20 @@ variable {α : Type*} [Fintype α] [DecidableEq α] [Nontrivial α]
 /--
 WOWII [Conjecture 23](http://cms.dt.uh.edu/faculty/delavinae/research/wowII/)
 
-For a simple connected graph `G`, `b(G) ≥ ⌊α(G) + dist_avg(M, V) / 2⌋`, where `b(G)` is
+For a simple connected graph `G`, `b(G) ≥ ⌊α(G) + dist_avg(M) / 2⌋`, where `b(G)` is
 the size of a largest induced bipartite subgraph, `α(G)` is the independence number,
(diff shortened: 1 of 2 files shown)


mo271 (approved):
Thanks!

PR #5739, opened 2026-09-11
fix(Wikipedia/Fuglede): dim_3_or_higher fails in every dimension n ≥ 3

### LINKED ISSUE #5703: Wikipedia/Fuglede: `variants.dim_3_or_higher` asserts failure in some dimension ≥ 3, not in every dimension
`Fuglede.FugledeConjecture.variants.dim_3_or_higher` states `answer(False) ↔ ∀ n : ℕ, 3 ≤ n → FugledeConjectureFor n`, which is equivalent to `∃ n, 3 ≤ n ∧ ¬ FugledeConjectureFor n` and is proved by a counterexample in one dimension alone, whereas the source states that Fuglede's conjecture fails in every dimension `n ≥ 3`.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/Wikipedia/Fuglede

- 1 misformalization
- 1 questionable
- 2 equivalent reformulations
- trivial proof: bailed out
- fix compiles

smmercuri:
`FugledeConjecture.variants.dim_3_or_higher` stated `answer(False) ↔ ∀ n : ℕ, 3 ≤ n → FugledeConjectureFor n`. Unfolding the `answer(False)`, this is `∃ n ≥ 3, ¬ FugledeConjectureFor n`, i.e. failure in *some* dimension, witnessed by a single counterexample. The source (Tao's counterexample in dimension 5, later pushed down to dimension 3, with counterexamples lifting to higher dimensions) says the conjecture fails in *every* dimension `n ≥ 3`.

**Change:** restate as `theorem ... (n : ℕ) (hn : 3 ≤ n) : ¬ FugledeConjectureFor n`, and expand the docstring to spell out what "fails" means (a bounded positive-measure set that is spectral but not a translational tile, or vice versa).

**Checked:** `lake --wfail build FormalConjectures.Wikipedia.Fuglede` passes with no warnings.

Fixes #5703

### DIFF
--- a/FormalConjectures/Wikipedia/Fuglede.lean
+++ b/FormalConjectures/Wikipedia/Fuglede.lean
@@ -52,12 +52,14 @@ theorem FugledeConjecture.variants.dim_2 :
   sorry
 
 /--
-**Fuglede's conjecture** in three or higher dimensions has been disproven.
+**Fuglede's conjecture** fails in every dimension `n ≥ 3`: for each such `n` there is a bounded
+subset of ℝ^n with positive Lebesgue measure that is spectral but does not tile ℝ^n by translation,
+or that tiles ℝ^n by translation but is not spectral.
 (Note that counterexamples in lower dimensions would also disprove the conjecture in higher dimensions.)
 -/
 @[category research solved, AMS 42 46 47]
-theorem FugledeConjecture.variants.dim_3_or_higher :
-    answer(False) ↔ ∀ n : ℕ, 3 ≤ n → FugledeConjectureFor n := by
+theorem FugledeConjecture.variants.dim_3_or_higher (n : ℕ) (hn : 3 ≤ n) :
+    ¬ FugledeConjectureFor n := by
   sorry
 
 end Fuglede


FormalConjectures/Wikipedia/Fuglede.lean
```diff
@@ -52,12 +52,14 @@ theorem FugledeConjecture.variants.dim_2 :
   sorry
 
 /--
-**Fuglede's conjecture** in three or higher dimensions has been disproven.
+**Fuglede's conjecture** fails in every dimension `n ≥ 3`: for each such `n` there is a bounded
+subset of ℝ^n with positive Lebesgue measure that is spectral but does not tile ℝ^n by translation,
+or that tiles ℝ^n by translation but is not spectral.
```
mo271: can we use latex in the docstring here?

mo271 (approved):
Thanks, just one comment about docstrings

FormalConjectures/Wikipedia/Fuglede.lean
smmercuri: Done, switched the new docstring to LaTeX.

smmercuri:

I compared this fix against issue #5703 and the audit's evidence/proposed fix at https://tadamcz.com/fc-review-results/c90271f0fa/#/f/Wikipedia/Fuglede, and tried to break the new statement. Verdict: **the fix is correct and fully addresses the misformalization.**

Why it is the right fix:

- **The defect is gone.** The old `answer(False) ↔ ∀ n ≥ 3, FugledeConjectureFor n` unfolds to `∃ n ≥ 3, ¬ FugledeConjectureFor n` (proved in Lean), so one counterexample in dimension 5 would have discharged it. The new `(n : ℕ) (hn : 3 ≤ n) : ¬ FugledeConjectureFor n` asserts failure in each dimension separately.
- **Equivalent to the audit's proposal.** The audit suggested `∀ n : ℕ, 3 ≤ n → ¬ FugledeConjectureFor n`; I proved `PR ↔ audit` in Lean against this branch (it is `Iff.rfl`), and the new statement strictly implies the old one. Plain hypotheses without an `answer` marker match other `research solved` results in the repository.
- **Faithful to the source.** Wikipedia: the conjecture "is false on ℝ^d for d ≥ 5" and "is also false for d = 3 and 4", i.e. in every dimension n ≥ 3. Those counterexamples are finite unions of unit cubes, hence bounded, so the negation holds under this file's bounded-set formulation.
- **No vacuity, no junk values.** `3 ≤ n` excludes `n = 0`, where `FugledeConjectureFor 0` is actually true (the only positive-measure set is the point, which is both spectral and a tile). The unit cube is both spectral and a tile under the underlying definitions, so the statement is neither trivially provable nor trivially false.
- **Every change justified; builds.** Only `dim_3_or_higher` changes; `research solved`/AMS tags are rightly kept, and the new docstring is an accurate unfolding of `¬ FugledeConjectureFor n`. `lake --wfail build FormalConjectures.Wikipedia.Fuglede` succeeds with only the expected `sorry` warnings.

Optional nits only (pre-existing, not blocking): the module title "Fuglede's conjecture in dimensions 1 and 2" now undersells the file, and lines 31/39/47 exceed 100 characters.

PR #5741, opened 2026-09-11
fix(Wikipedia/MeanValueProblem): put Tischler hypotheses on p(X+z)-p(z)

### LINKED ISSUE #5708: Wikipedia/MeanValueProblem: `…_of_real_roots` and `…_of_roots_same_norm` restrict roots of `p`, not of `p(X+z)-p(z)`
`MeanValueProblem.mean_value_problem_of_real_roots` and `mean_value_problem_of_roots_same_norm` place the real-root and equal-norm hypotheses on the roots of `p` while concluding at every `z`, whereas Tischler's special cases place them on the nonzero roots of the normalized polynomial `p(X+z)-p(z)`, so `X^3 - X` meets the first hypothesis yet violates Tischler's condition at `z = 2i`, and fails the second hypothesis at `z = 0` although its nonzero roots `±1` have equal norm.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/Wikipedia/MeanValueProblem

- 2 misformalizations
- 1 status issue
- 2 equivalent reformulations
- trivial proof compiles
- fix compiles

smmercuri:
`MeanValueProblem.mean_value_problem_of_real_roots` and `mean_value_problem_of_roots_same_norm` placed the real-root and equal-norm hypotheses on the roots of `p` itself, while concluding the `(d-1)/d` bound at every point `z`. Tischler's special cases instead place these conditions on the (nonzero) roots of the normalised polynomial `p(X + z) - p(z)`. For example `X^3 - X` has only real roots but violates Tischler's condition at `z = 2i`, and fails the old equal-norm hypothesis at `z = 0` although its nonzero roots `±1` have equal norm.

**Change:** transport both hypotheses to `p(X + z) - p(z)`, phrased in terms of the points `x` with `p.eval x = p.eval z`: the real-roots lemma now assumes every such `x` has `x.im = z.im`, and the same-norm lemma assumes all such `x ≠ z` have the same `‖x - z‖`. `z` is moved before the hypothesis so it can be referenced. The module docstring and both lemma docstrings are updated to describe the corrected special cases.

**Checked:** `lake --wfail build FormalConjectures.Wikipedia.MeanValueProblem` passes with no warnings.

Fixes #5708

### DIFF
--- a/FormalConjectures/Wikipedia/MeanValueProblem.lean
+++ b/FormalConjectures/Wikipedia/MeanValueProblem.lean
@@ -32,7 +32,8 @@ The conjecture has been proven for:
 * `K = 4`
   [The fundamental theorem of algebra and complexity theory](https://www.ams.org/journals/bull/1981-04-01/S0273-0979-1981-14858-8/)
   by *Steve Smale*
-* `K = (d-1)/d` if $p$ has real roots or all the roots of $p$ have the same norm.
+* `K = (d-1)/d` at a point $z$ if the normalised polynomial $p(X + z) - p(z)$ has only real
+  roots, or if all its nonzero roots have the same norm.
   [Critical points and values of complex polynomials](https://doi.org/10.1016/0885-064X(89)90019-8)
   by *David Tischler*
 -/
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5748, opened 2026-09-11
fix(OEIS/109905): align the answer polarity with the source question

### LINKED ISSUE #5689: OEIS/109905: `conjecture` reverses the yes/no polarity: `answer(True)` denies any further zero
`OeisA109905.conjecture` makes `answer(True)` equivalent to the set of positive indices `n` with `a n = 0` being exactly `{1, 6, 30, 54}`, that is, to the non-existence of any further zero, whereas the source asks whether another zero exists, so the Lean right-hand side is the negation of the source's proposition and the two answers are swapped.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/OEIS/109905

- 1 misformalization
- trivial proof: bailed out
- fix compiles

smmercuri:
`OeisA109905.conjecture` stated `answer(sorry) ↔ {n | n > 0 ∧ a n = 0} = {1, 6, 30, 54}`, so `answer(True)` would mean there is *no* further zero. The OEIS entry asks "a(n) = 0 for n = 1, 6, 30 and 54. Are there any others?", i.e. `answer(True)` should mean another zero *exists*. The Lean right-hand side was the negation of the source proposition.

**Change:** restate the right-hand side as `∃ n : ℕ, n > 0 ∧ a n = 0 ∧ n ∉ ({1, 6, 30, 54} : Finset ℕ)` so the yes/no polarity matches the question.

**Checked:** `lake --wfail build FormalConjectures.OEIS.«109905»` passes with no warnings.

Fixes #5689

### DIFF
--- a/FormalConjectures/OEIS/109905.lean
+++ b/FormalConjectures/OEIS/109905.lean
@@ -58,7 +58,8 @@ theorem a_5 : a 5 = 7 := by decide
 $a(n) = 0$ for $n = 1$, $6$, $30$ and $54$. Are there any others?
 -/
 @[category research open, AMS 11]
-theorem conjecture : answer(sorry) ↔ {n : ℕ | n > 0 ∧ a n = 0} = {1, 6, 30, 54} := by
+theorem conjecture :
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #5749, opened 2026-09-11
fix(OptimizationConstants/1a): sup over [-1/2,1/2] and restrict C1a to L² f

### LINKED ISSUE #5696: OptimizationConstants/1a: `C1a` sups over `Icc (1/2) 1`, and the real Bochner integral and `sSup` force `C1a = 0`
`Constant1a.C1a` takes the inner supremum over `Icc (1/2) 1` instead of the source's [-1/2, 1/2], and its real-valued Bochner convolution integral and real `sSup` both return 0 on nonintegrable or unbounded values, so `f ≡ 1` forces every admissible `C` to be nonpositive and the definition evaluates to `C1a = 0` rather than the source's autocorrelation constant.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/OptimizationConstants/1a

- 3 misformalizations
- 1 minor
- 2 status issues
- trivial proof compiles
- fix compiles

smmercuri:
`Constant1a.C1a` took the inner supremum over `t ∈ Icc (1/2) 1`, whereas the source (Tao's optimization constant 1a / Matolcsi–Vinuesa) maximises the autoconvolution over `t ∈ [-1/2, 1/2]`. Moreover it quantified over *all* nonnegative `f : ℝ → ℝ`: because the real Bochner integral and real `sSup` return `0` on nonintegrable or unbounded input, `f ≡ 1` forced every admissible `C` to be `≤ 0`, so the definition evaluated to `C1a = 0` instead of the intended constant.

**Change:** take the supremum over `Icc (-1/2) (1/2)` and add the hypothesis `MemLp f 2` (square-integrable `f`), which guarantees every autoconvolution integral exists and the supremum is finite; this restriction does not change the constant ([M2010, §2]). The docstring now states the defining inequality explicitly. The open-problem frontier bounds are updated to the currently best known values: `mem_Ico_c1a` upper bound `1.5029 → 1.502862` [T2026] and `mem_Ioc_c1a` lower bound `1.2748 → 1.292` [PBV2026]; the older `c1a_lower_bound`/`c1a_upper_bound` theorems are kept with their docstrings relabelled "A lower/upper bound" rather than "best known".

**Checked:** `lake --wfail build FormalConjectures.OptimizationConstants.«1a»` passes with no warnings.

Reviewer note: the two new frontier references [T2026] (Together AI, EinsteinArena-new-SOTA) and [PBV2026] (Piterbarg–Bajaj–Vincent) are GitHub repositories rather than papers; please check these are acceptable as sources for the open-problem bounds. This PR overlaps with #5559, which also touches this file.

Fixes #5696

### DIFF
--- a/FormalConjectures/OptimizationConstants/1a.lean
+++ b/FormalConjectures/OptimizationConstants/1a.lean
@@ -23,38 +23,50 @@ import FormalConjecturesUtil
 - [M2010] Matolcsi, Máté, and Carlos Vinuesa. "Improved bounds on the supremum of autoconvolutions."
   Journal of mathematical analysis and applications 372.2 (2010): 439-447. [arXiv:0907.1379](https://arxiv.org/abs/0907.1379)
 - [Y2026] Yuksekgonul, Mert et al., "Learning to Discover at Test Time," 2026, [arXiv:2601.16175](https://arxiv.org/abs/2601.16175)
+- [T2026] Together AI, "EinsteinArena-new-SOTA: State-of-the-art results on open math problems,"
+  2026, [GitHub](https://github.com/togethercomputer/EinsteinArena-new-SOTA)
+- [PBV2026] Piterbarg, Andrei, Jai Bajaj, and Derrick Vincent. "A multi-scale arcsine lower bound
+  for the Sidon autocorrelation constant $C_{1a}$," 2026,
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks, LGTM!

PR #5767, opened 2026-09-11
fix(ErdosProblems/90): add the size bound and shortest-vector gap to Lemma 2

### LINKED ISSUE #5673: ErdosProblems/90: `sawin_lattice_reduction` omits the size bound on `U.card`, leaving a density-only claim
`Erdos90.sawin_lattice_reduction` concludes only that the unit-distance density of `U` exceeds a bound, with no upper bound on `U.card` and no shortest-vector parameter, whereas Sawin's Lemma 2 bounds the size of `U` by `(2R/ρ+1)^(2d)` at the same time as its unit-distance density, so the Lean conclusion holds for a suitable planar point set without any of the lattice hypotheses.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/ErdosProblems/90

- 1 misformalization
- 1 minor
- 3 equivalent reformulations
- trivial proof: bailed out
- fix compiles

smmercuri:
`Erdos90.sawin_lattice_reduction` concluded only that some finite planar set `U` has unit-distance density at least `(1 - 1/R)^(2d) * #S`, with no bound on `U.card` and no shortest-vector parameter. Sawin's Lemma 2 ([arXiv:2605.20579](https://arxiv.org/abs/2605.20579)) proves the size bound `#U ≤ (2R/ρ + 1)^(2d)` at the same time, where `ρ` is the minimal norm of a nonzero lattice vector, and its density counts ordered unit-distance pairs. Without the size bound the Lean conclusion holds for a suitable planar set independently of the lattice hypotheses.

**Change**
- Add `ρ > 0` with the hypothesis that every nonzero vector of `Λ` has norm at least `ρ` (Sawin's shortest-vector length; this also makes `Λ` a lattice).
- Add the conclusion `(U.card : ℝ) ≤ (2 * R / ρ + 1) ^ (2*d)`.
- Convert the ordered-pair count: the density inequality now has `2 * unitDistNum U` on the right, since `unitDistNum` counts unordered pairs.
- Relax `2 ≤ R` to the source's `1 < R`, and update the docstring accordingly.

**Checked**: `lake --wfail build 'FormalConjectures.ErdosProblems.«90»'` succeeds.

Fixes #5673

### DIFF
--- a/FormalConjectures/ErdosProblems/90.lean
+++ b/FormalConjectures/ErdosProblems/90.lean
@@ -148,25 +148,30 @@ theorem erdos_90.variants.polynomial_lower_bound_implies_erdos_90 :
 /--
 **Sawin's Lemma 2: lattice geometry of unit distances** (Sawin, [arXiv:2605.20579](https://arxiv.org/abs/2605.20579)).
 
-Let $d \ge 1$, $R \ge 2$, and suppose $\Lambda \subset \mathbb{R}^{2d}$ is a lattice equipped with
-an additive embedding $\pi : \Lambda \to \mathbb{R}^2$. Suppose $S \subseteq \Lambda$ is a finite
-set of "matching" vectors satisfying $\|v\| \le 1$ and $\|\pi v\| = 1$ for every $v \in S$. Then
-there is a finite point set $U \subset \mathbb{R}^2$ with unit-distance density at least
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #5769, opened 2026-09-11
fix(ErdosProblems/786): Selfridge's set has density greater than 1/e - ε

### LINKED ISSUE #5667: ErdosProblems/786: `parts.i.selfridge` asserts density exactly 1/e − ε for a fixed set, not greater than it
`Erdos786.erdos_786.parts.i.selfridge` concludes `A.HasDensity (1 / rexp 1 - ε)` for the set `A` fixed by the `let`, whereas Selfridge's construction gives `A` density greater than 1/e − ε, and since the same `A` cannot have two distinct densities for different ε the statement as written is false.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/ErdosProblems/786

- 1 misformalization
- 2 equivalent reformulations
- trivial proof compiles
- fix compiles

smmercuri:
`erdos_786.parts.i.selfridge` concluded `A.HasDensity (1 / rexp 1 - ε)` for the set `A` fixed by `p` and `k`. The density of that set does not depend on `ε`, so the same `A` would have to have two distinct densities for two values of `ε`, and the statement was false. The source ([erdosproblems.com/786](https://www.erdosproblems.com/786)) says Selfridge's construction gives a set of density $1/e - \epsilon$ for any $\epsilon > 0$, i.e. density greater than $1/e - \epsilon$.

**Change**: the conclusion is now `∃ δ, 1 / rexp 1 - ε < δ ∧ A.HasDensity δ ∧ A.IsMulCardSet`, mirroring the form used in `erdos_786.parts.i`. The docstring now says the density is $> \frac{1}{e} - \epsilon$.

**Checked**: `lake --wfail build 'FormalConjectures.ErdosProblems.«786»'` completes successfully with no warnings.

Fixes #5667

### DIFF
--- a/FormalConjectures/ErdosProblems/786.lean
+++ b/FormalConjectures/ErdosProblems/786.lean
@@ -177,7 +177,7 @@ $$
 \sum_{i=1}^k \frac{1}{p_i} < 1 < \sum_{i=1}^{k + 1} \frac{1}{p_i},
 $$
 and let $A$ be the set of all naturals divisible by exactly one of $p_1, ..., p_k$ (with
-multiplicity $1$). Then $A$ has density $\frac{1}{e} - \epsilon$ and has the property
+multiplicity $1$). Then $A$ has density $> \frac{1}{e} - \epsilon$ and has the property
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #5773, opened 2026-09-11
fix(ErdosProblems/785): take A, B positive; Chen–Fang bound uses min

### LINKED ISSUE #5666: ErdosProblems/785: Summands may be 0 but `counting` starts at 1; `chen_fang` uses `counting A`, not the min
`erdos_785`, `variants.ruzsa_lower_bound`, `variants.chen_fang_limsup` and `variants.chen_conjecture` let `A` and `B` contain 0 as a summand while `counting` counts only `A ∩ Set.Icc 1 x`, whereas the source takes both sets positive, so a shifted Ruzsa pair gives discrepancy tending to −∞; and `variants.chen_fang` bounds the discrepancy by a power of `counting A` with no hypothesis making `A` the sparse set, whereas Chen–Fang bound it by the smaller counting function.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/ErdosProblems/785

- 5 misformalizations
- 3 equivalent reformulations
- trivial proof: bailed out
- fix compiles

smmercuri:
The source (and the cited formal proof by van Doorn) takes $A, B$ to be sets of positive integers, but the Lean statements allowed $0 \in A$ or $0 \in B$ while `counting A x = |A ∩ [1, x]|` ignores `0`. A zero summand makes `A + B` easier to cover without changing the counting functions, so the statements did not match the source. Separately, `chen_fang` bounded $A(x)B(x)-x$ by `counting A x ^ c` with nothing making $A$ the sparse set; for the denser set and $c = 2$ the $O$-bound holds trivially, so that statement was false. Chen–Fang [ChFa15] actually prove $A(x)B(x)-x \ge \min(A(x),B(x))^M$ for every $M > 1$.

**Change**: add `0 ∉ A` / `0 ∉ B` to every statement in the file (hypotheses for the universal ones, conjuncts for the existential ones), state the main problem's docstring with "sets of positive integers", and replace `counting A x ^ c` by `(min (counting A x) (counting B x) : ℝ) ^ c` in `chen_fang`, with the docstring updated accordingly.

**Checked**: `lake --wfail build FormalConjectures.ErdosProblems.«785»` builds successfully.

Fixes #5666

### DIFF
--- a/FormalConjectures/ErdosProblems/785.lean
+++ b/FormalConjectures/ErdosProblems/785.lean
@@ -62,8 +62,8 @@ def IsExactAdditiveComplement (A B : Set ℕ) : Prop :=
     Tendsto (fun x : ℕ => (counting A x * counting B x : ℝ) / (x : ℝ)) atTop (𝓝 1)
 
 /--
-Let $A,B\subseteq \mathbb{N}$ be infinite sets such that $A+B$ contains all large integers.
-Let $A(x)=\lvert A\cap [1,x]\rvert$ and similarly for $B(x)$. Is it true that if
+Let $A,B\subseteq \mathbb{N}$ be infinite sets of positive integers such that $A+B$ contains all
+large integers. Let $A(x)=\lvert A\cap [1,x]\rvert$ and similarly for $B(x)$. Is it true that if
 $A(x)B(x)\sim x$ then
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #5788, opened 2026-09-11
fix(ErdosProblems/486): only forbid residues modulo n for 0 < n < m

### LINKED ISSUE #5654: ErdosProblems/486: `erdos_486` tests every modulus `n`, including `0` and `n ≥ m`, instead of `0 < n < m`
`Erdos486.erdos_486` requires `(m : ZMod n) ∉ X n` for every modulus `n`, including `n = 0` (where `ZMod 0` is `ℤ`, so `X 0` alone can carve out an arbitrary set of `m`) and every `n ≥ m`, whereas the source only forbids residues modulo `n` for `0 < n < m`, so the stated set need not have a logarithmic density.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/ErdosProblems/486

- 2 misformalizations
- 2 equivalent reformulations
- trivial proof: bailed out
- fix compiles

smmercuri:
`Erdos486.erdos_486` required `(m : ZMod n) ∉ X n` for every modulus `n`. This includes `n = 0` (where `ZMod 0 = ℤ`, so `X 0` alone makes `B` an arbitrary set of naturals with no logarithmic density) and every `n ≥ m`. The source (erdosproblems.com/486, updated after Terence Tao's comment pointing to the activation threshold `b ≥ a_i` in [Er61, p.236]) only forbids `m` from the residues in `X_n` for `n ∈ A` with `m > n`; without that threshold there is a cheap negative answer.

**Change**: the set is now `{m : ℕ | ∀ n, 0 < n → n < m → (m : ZMod n) ∉ X n}`. The docstring is updated to the source wording and explains that `A` is encoded by `X_n = ∅` for `n ∉ A` and that only positive moduli are considered.

**Checked**: `lake --wfail build 'FormalConjectures.ErdosProblems.«486»'` — Build completed successfully, no warnings.

Fixes #5654

### DIFF
--- a/FormalConjectures/ErdosProblems/486.lean
+++ b/FormalConjectures/ErdosProblems/486.lean
@@ -25,13 +25,19 @@ import FormalConjecturesUtil
 namespace Erdos486
 
 /--
-For each $n \in \mathbb{N}$ choose some $X_n \subseteq \mathbb{Z}/n\mathbb{Z}$.
-Let $B = \{m \in \mathbb{N} : \forall n, m \not\equiv x \pmod{n} \text{ for all } x \in X_n\}$.
+Let $A \subseteq \mathbb{N}$, and for each $n \in A$ choose some
+$X_n \subseteq \mathbb{Z}/n\mathbb{Z}$. Let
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #5791, opened 2026-09-11
fix(ErdosProblems/422): specify Hofstadter's Q-sequence by a predicate

### LINKED ISSUE #5649: ErdosProblems/422: `f` is a `partial def`, an opaque constant with no equation lemmas giving the recurrence
`Erdos422.f` is a `partial def`, which Lean elaborates as an opaque constant `ℕ+ → ℕ+` with no equation lemmas, so the logic does not impose `f 2 = 1` or the recurrence `f n = f (n - f (n-1)) + f (n - f (n-2))`, whereas the source concerns Hofstadter's Q-sequence and the four statements in the file refer to a function with no logical specification identifying it.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/ErdosProblems/422

- 1 misformalization
- 2 questionable
- trivial proof: bailed out
- fix compiles

smmercuri:
`Erdos422.f` was a `partial def`. Lean elaborates this as an opaque constant `ℕ+ → ℕ+` with no equation lemmas, so the four statements in the file referred to a function that the logic never identified with Hofstadter's $Q$-sequence: neither $f(1) = f(2) = 1$ nor the recurrence $f(n) = f(n - f(n-1)) + f(n - f(n-2))$ was available.

**Change**

- Replace `partial def f` by the predicate `IsHofstadterQ (f : ℕ+ → ℕ+)`: $f(1) = f(2) = 1$ and, for every $n > 2$, $f(n-1) < n$, $f(n-2) < n$ and $f(n) = f(n - f(n-1)) + f(n - f(n-2))$. The two inequalities are the conditions that make both arguments positive integers; at most one function satisfies the predicate, and one does exactly when the sequence is well-defined for all $n$ (the source notes this is not known).
- Quantify `erdos_422`, `erdos_422.variants.surjective`, `erdos_422.variants.growth_rate` and `erdos_422.variants.eventually_const` over `f` with `IsHofstadterQ f`, keeping `answer(sorry)` before the quantifier.
- Add test lemmas `erdos_422.test.f3` ($f(3) = 2$) and `erdos_422.test.f4` ($f(4) = 3$), and cite OEIS A005185 in the module docstring.

**Checked**: `lake --wfail build FormalConjectures.ErdosProblems.«422»` builds successfully.

Fixes #5649

### DIFF
--- a/FormalConjectures/ErdosProblems/422.lean
+++ b/FormalConjectures/ErdosProblems/422.lean
@@ -19,7 +19,9 @@ import FormalConjecturesUtil
 /-!
 # Erdős Problem 422
 
-*Reference:* [erdosproblems.com/422](https://www.erdosproblems.com/422)
+*References:*
+- [erdosproblems.com/422](https://www.erdosproblems.com/422)
+- [OEIS A005185](https://oeis.org/A005185)
 -/
 
 namespace Erdos422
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #5793, opened 2026-09-11
fix(ErdosProblems/522): use counting measure as Kac coefficient reference

### LINKED ISSUE #5658: ErdosProblems/522: `KacCoefficients` uses Lebesgue measure, excluding every two-point uniform coefficient
`Erdos522.KacCoefficients` takes `volume` as its default reference measure, and `erdos_522`, `zero_one`, `number_real_roots` and `yakir_solution` all use that default on the two-point sets `{-1, 1}` or `{0, 1}`, which have Lebesgue measure zero, so `h_unif` forces the law of each coefficient to be the zero measure rather than the fair Rademacher or Bernoulli distribution the source requires.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/ErdosProblems/522

- 1 misformalization
- 2 equivalent reformulations
- trivial proof compiles
- fix compiles

smmercuri:
`Erdos522.KacCoefficients` took `volume` as the default reference measure `μ`, and `erdos_522`, `variants.zero_one`, `variants.number_real_roots` and `variants.yakir_solution` all used this default on the two-point sets `{-1, 1}` or `{0, 1}`. These sets have Lebesgue measure zero, so `h_unif : pdf.IsUniform (toFun i) S ℙ μ` (which unfolds to `map (toFun i) ℙ = cond μ S`) forced the law of each coefficient to be the zero measure, not the fair `±1` (resp. `{0, 1}`) coefficients of [erdosproblems.com/522](https://www.erdosproblems.com/522).

**Change**: the default reference measure of `KacCoefficients` is now `Measure.count`, and the docstring explains that for a finite `S` each coefficient is then uniform on `S`. With this default, `cond Measure.count {-1, 1} = 2⁻¹ • (dirac (-1) + dirac 1)` is exactly the intended Rademacher law (checked in a scratch file). No theorem statements needed to change since they all use the default.

**Checked**: `lake --wfail build 'FormalConjectures.ErdosProblems.«522»'` builds successfully with no warnings.

Fixes #5658

### DIFF
--- a/FormalConjectures/ErdosProblems/522.lean
+++ b/FormalConjectures/ErdosProblems/522.lean
@@ -29,7 +29,9 @@ namespace Erdos522
 
 /--
 A sequence of *Kac coefficients* over a subset `S` of a field `k` is a countably infinite sequence
-of independent random variables, each uniformly distributed over `S`.
+of independent random variables, each uniformly distributed over `S` with respect to the reference
+measure `μ`. The default reference measure is the counting measure, so that for a finite set `S`
+each coefficient takes every value of `S` with probability `1 / |S|`.
 
 Such a sequence determines a *Kac polynomial* of degree `n` for each `n`, which is the random
 polynomial given by `KacCoefficients.polynomial`.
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #5803, opened 2026-09-11
fix(ErdosProblems/289): require the intervals to be non-adjacent

### LINKED ISSUE #5636: ErdosProblems/289: `erdos_289` requires the intervals only to be disjoint, not separated
`Erdos289.erdos_289` asks only that the `k` integer intervals be pairwise disjoint, so adjacent intervals such as `[2,3]` and `[4,5]` count separately, whereas the source requires `k` separated intervals with at least one integer between any two of them.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/ErdosProblems/289

- 1 misformalization
- 1 minor
- 1 equivalent reformulation
- trivial proof compiles
- fix compiles

smmercuri:
[erdosproblems.com/289](https://www.erdosproblems.com/289) asks for finite intervals $I_1, \dotsc, I_k \subset \mathbb{N}$ that are "distinct, not overlapping or adjacent" (the page notes that without this restriction the problem is easy, by Kovač's argument). `erdos_289` only required the intervals to be pairwise disjoint, so adjacent intervals such as $[2,3]$ and $[4,5]$ counted as two separate intervals.

**Change:** replace the disjointness condition `(I i).2 < (I j).1 ∨ (I j).2 < (I i).1` by the separation condition `(I i).2 + 1 < (I j).1 ∨ (I j).2 + 1 < (I i).1`, so that any two intervals are separated by at least one integer, and state the restriction in the docstring.

**Checked:** `lake --wfail build FormalConjectures.ErdosProblems.«289»` builds successfully.

Fixes #5636

### DIFF
--- a/FormalConjectures/ErdosProblems/289.lean
+++ b/FormalConjectures/ErdosProblems/289.lean
@@ -25,17 +25,20 @@ open Asymptotics Filter Finset
 
 namespace Erdos289
 
-/-- Is it true that, for all sufficiently large $k$, there exists finite intervals
-$I_1, \dotsc, I_k \subset \mathbb{N}$ with $|I_i| \geq 2$ for $1 \leq i \leq k$ such that
+/-- Is it true that, for all sufficiently large $k$, there exist finite intervals
+$I_1, \dotsc, I_k \subset \mathbb{N}$, distinct, not overlapping or adjacent, with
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks

PR #5333, opened 2026-09-07
fix(MovingSofa): apply translation before rotation

dawidmtrela-dotcom:
Refs #5270.

This is a **draft**, pending confirmation of the intended composition order.
It proposes following the existing docstring: translate first, then rotate.
If rotation before translation is intended instead, this code change should
not be merged; the documentation and path conventions need to be considered
instead.

Changes (one file only):

- Reverse the composition in `MovingSofa.rotateTranslate`.
- Add `MovingSofa.rotateTranslate_apply`, a pointwise regression theorem for
  `rotateTranslate alpha p q = rotation alpha (q + p)`, proved by `rfl` and
  tagged `@[category test, AMS 49]`.

The integral path formulas and existing conjecture placeholders are unchanged.
This is not a formal proof of Gerver sofa optimality or of the full
representation/motion bridge in this repository.

Validation already performed locally on commit
`8323e878b83fcd7f4a448256069352a265460d75` with this exact patched file:

- Lean 4.33.1 (`leanprover/lean4:v4.33.1`).
- `lake --wfail build FormalConjectures.Wikipedia.MovingSofa`: exit 0.
- `#print axioms MovingSofa.rotateTranslate_apply`: only `propext`,
  `Classical.choice`, and `Quot.sound`; no `sorryAx` for this theorem.

The submission branch starts at `2c817e975be7a95478b72a8429155ca568e1a3de`. Before applying the change,
the launcher verified that `MovingSofa.lean` there was byte-identical to the
tested original. The full build was not rerun for this submission base; CI still
needs to validate the submitted branch. The axiom check above concerns the
new regression theorem, not all existing declarations in the file.

Prepared with AI assistance; the build and axiom-check results above come
from local Lean runs.

PR #5827, opened 2026-09-11
fix(Wikipedia/MovingSofa): apply the translation before the rotation

### LINKED ISSUE #5270: MovingSofa: rotateTranslate implements the opposite composition order from its documented convention

In `FormalConjectures/Wikipedia/MovingSofa.lean`, the documentation for
`rotateTranslate` says:

> The rigid motion that translates by `p` and then rotates counterclockwise by `α`.
> Note that [Ge92] used this definition while [Ro18] used rotation first and then translation.

However, the current definition is

```lean
def rotateTranslate (α : Real.Angle) (p : ℝ²) : E(2) :=
  (EuclideanGeometry.o.rotation α).toAffineIsometryEquiv
    |>.trans (AffineIsometryEquiv.vaddConst ℝ p)

Mathlib's composition convention is
@[simp]
theorem coe_trans
    (e₁ : P ≃ᵃⁱ[𝕜] P₂) (e₂ : P₂ ≃ᵃⁱ[𝕜] P₃) :
    ⇑(e₁.trans e₂) = e₂ ∘ e₁ := rfl

Therefore the current definition acts on a point q as

$$ q \mapsto R_\alpha q + p, $$

i.e. it rotates first and translates afterwards.

By contrast, “translate by p and then rotate” acts as

$$ q \mapsto R_\alpha(q+p) = R_\alpha q + R_\alpha p. $$

Thus the implementation realizes the opposite composition order from the convention stated in its documentation.

Pinned source:

https://github.com/google-deepmind/formal-conjectures/blob/8323e878b83fcd7f4a448256069352a265460d75/FormalConjectures/Wikipedia/MovingSofa.lean

Trivial proof/disproof

The discrepancy follows directly from the semantics of
AffineIsometryEquiv.trans.

The current expression
rotation.trans (vaddConst p)
is definitionally

q ↦ vaddConst p (rotation q)

and hence

$$ q \mapsto R_\alpha q + p, $$

not

$$ q \mapsto R_\alpha(q+p). $$

I encountered this while constructing a rigorous interoperability bridge
between the integral path defined in this file and Romik's five-phase
rotation path.

For the exact path currently defined in the pinned MovingSofa.lean
source, I have now derived phase-by-phase the identity

$$ \boxed{x_R(t)=R_t p_G(t)} \qquad (0\le t\le \pi/2), $$

with no additional translation, reflection, fixed rotation, sign change,
or parameter reversal.

Consequently, the hallway family corresponding to the documented
translate-then-rotate convention is

$$ R_t(L+p_G(t)) = R_tL+R_tp_G(t) = x_R(t)+R_tL. $$

The current implementation instead produces

$$ R_tL+p_G(t), $$

so for the path presently defined in this file it does not give the
Romik hallway family associated with the documented convention.

Suggested fix

If the intended convention is the documented translate-then-rotate
convention, the composition order can be reversed:
def rotateTranslate (α : Real.Angle) (p : ℝ²) : E(2) :=
  (AffineIsometryEquiv.vaddConst ℝ p).trans
    (EuclideanGeometry.o.rotation α).toAffineIsometryEquiv
which acts as

$$ q \mapsto R_\alpha(q+p). $$

Equivalently, the current composition order can be retained if the
world-frame translation R_α p is supplied instead. For the concrete
Gerver path considered here,

$$ R_\alpha p_G(\alpha)=x_R(\alpha). $$

I would prefer confirmation of the intended convention before changing
the downstream definition of gerversSofa.

Additional context

I recently submitted a separate proof of
GerversSofa.ABφθSpec.existsUnique as PR #5264.

The present issue arose while working on the next interoperability step:
relating the concrete MovingSofa.gerversSofa to an already certified
Romik moving-sofa construction.

This issue is independent of PR #5264.

One source qualification is worth making explicit. The path-equivalence
calculation above is complete for the formulas in the pinned DeepMind
source. I have not yet independently compared every one of those
integral formulas against the original printed Gerver 1992 article.
Accordingly, I am not claiming here that the DeepMind transcription
itself is wrong. The issue reported here is the composition semantics of
rotateTranslate relative to its documentation and to the exact path
currently defined in this file.

Choose either option
 I plan on working on this issue

smmercuri:
`MovingSofa.rotateTranslate α p` is documented as the rigid motion that translates by `p` and then rotates by `α` (Gerver's convention, `q ↦ R_α (q + p)`), with the note that [Ro18] uses the opposite order. But `AffineIsometryEquiv.trans` composes left-to-right, so `rotation.trans (vaddConst p)` computes `q ↦ R_α q + p`, i.e. Romik's order. The Gerver path `GerversSofa.p` in this file satisfies `x_Romik(t) = R_t p(t)` (checked numerically phase by phase against Romik's Table 1), so only the documented translate-then-rotate order makes `sofaOfRotateTranslatePath p` equal Romik's hallway family `⋂ (x(t) + R_t L)`; the current order places the hallways wrongly (e.g. at `t = π/2` it translates by `(0, 1.23)` instead of `(-1.23, 0)`).

**Change**: reverse the composition in `rotateTranslate` to `(vaddConst ℝ p).trans (rotation α)`, and add the `rfl` regression theorem `rotateTranslate_apply : rotateTranslate α p q = rotation α (q + p)` (`category test`) pinning the convention. Docstrings and the path formulas are unchanged.

**Checked**: `lake --wfail build FormalConjectures.Wikipedia.MovingSofa` (Build completed successfully, no warnings).

Note: this is the same change as the draft PR #5333 by the issue reporter, which was awaiting confirmation of the intended convention; the numerical check above confirms it.

Fixes #5270

### DIFF
--- a/FormalConjectures/Wikipedia/MovingSofa.lean
+++ b/FormalConjectures/Wikipedia/MovingSofa.lean
@@ -115,8 +115,13 @@ The rigid motion that translates by $p$ and then rotates counterclockwise by $\a
 Note that [Ge92] used this definition while [Ro18] used rotation first and then translation.
 -/
 def rotateTranslate (α : Real.Angle) (p : ℝ²) : E(2) :=
-  (EuclideanGeometry.o.rotation α).toAffineIsometryEquiv
-    |>.trans (AffineIsometryEquiv.vaddConst ℝ p)
+  (AffineIsometryEquiv.vaddConst ℝ p).trans
+    (EuclideanGeometry.o.rotation α).toAffineIsometryEquiv
+
+/-- `rotateTranslate α p` sends $q$ to $R_\alpha(q + p)$: the translation is applied first. -/
+@[category test, AMS 49]
+theorem rotateTranslate_apply (α : Real.Angle) (p q : ℝ²) :
+    rotateTranslate α p q = EuclideanGeometry.o.rotation α (q + p) := rfl
 
 /--
 The sofa according to a rotation path $p : [0, \pi/2] \to \mathbb{R}^2$ as in [Ge92] is the


mo271 (approved):
Thanks!

PR #5832, opened 2026-09-11
fix(ErdosProblems/363): fix the number of blocks and their sizes

### LINKED ISSUE #5643: ErdosProblems/363: `erdos_363` asks finiteness of the union over all block counts and lengths, not fixed ones
`Erdos363.erdos_363` applies `.Finite` to the set of all valid collections `S` with the number of blocks and the block lengths left free, so `answer(False)` asserts that the union over every parameter choice is infinite, whereas the source asks whether, for one fixed number of blocks with fixed lengths, there are infinitely many solutions.

https://tadamcz.com/fc-review-results/c90271f0fa/#/f/ErdosProblems/363

- 1 misformalization
- 1 equivalent reformulation
- trivial proof: bailed out
- fix compiles

smmercuri:
`erdos_363` applied `.Finite` to the set of all valid collections `S` with the number of blocks and the block sizes left free, so `answer(False)` only asserted that the union over every parameter choice is infinite. The sources keep the parameters fixed: Ulas [Ul05, p. 332] states the conjecture for "each fixed k" and separately notes Skałba's observation that "if we allow k to vary, the above conjecture is not true"; Bennett–Van Luijk [BeVL12, §1] ask for finitely many solutions "for fixed r ≥ 1 and {k₁, …, k_r} with k_j ≥ 4". So the previous free-parameter statement was true for a cheap reason and did not capture the problem.

**Change:** the statement now reads `answer(False) ↔ ∀ ks : List ℕ, {S | IsValidCollection S ∧ S.map Finset.card = ks}.Finite`, so the list `ks` fixes both the number of blocks and their sizes. The docstring says so and records Ulas's counterexample family (`n = 4`, all sizes `4`). The `formal_proof` link is kept: the linked construction uses four blocks of length `4` (`Icc (x + 1) (x + 4)`) with an injective parametrisation, which gives infinitely many solutions for the fixed `ks = [4, 4, 4, 4]`.

**Checked:** `lake --wfail build 'FormalConjectures.ErdosProblems.«363»'` — Build completed successfully.

Fixes #5643

### DIFF
--- a/FormalConjectures/ErdosProblems/363.lean
+++ b/FormalConjectures/ErdosProblems/363.lean
@@ -49,13 +49,20 @@ def IsValidCollection (S : List (Finset ℕ)) : Prop :=
   IsSquare ((S.map (fun I => ∏ m ∈ I, m)).prod)
 
 /--
-Is it true that there are only finitely many collections of disjoint intervals $I_1,\ldots,I_n$ of size $\lvert I_i\rvert \geq 4$ for $1\leq i\leq n$ such that$$\prod_{1\leq i\leq n}\prod_{m\in I_i}m$$is a square?
+Is it true that, for each fixed $n$ and fixed sizes $k_1,\ldots,k_n \geq 4$, there are only finitely
+many collections of disjoint intervals $I_1,\ldots,I_n$ of size $\lvert I_i\rvert = k_i$ for
+$1\leq i\leq n$ such that$$\prod_{1\leq i\leq n}\prod_{m\in I_i}m$$is a square?
 
-This is false: Ulas [Ul05] constructed infinitely many such collections.
+The number of intervals and their sizes are fixed: the list `ks` of sizes determines both. Without
+this restriction, finiteness fails for a trivial reason (Skałba, see [Ul05]).
+
+This is false: Ulas [Ul05] constructed infinitely many such collections with $n = 4$ and
+$k_1 = \cdots = k_4 = 4$.
 -/
 @[category research solved, AMS 11,
   formal_proof using lean4 at "https://github.com/plby/lean-proofs/blob/main/src/v4.29.1/ErdosProblems/Erdos363.lean"]
-theorem erdos_363 : answer(False) ↔ {S : List (Finset ℕ) | IsValidCollection S}.Finite := by
+theorem erdos_363 : answer(False) ↔ ∀ ks : List ℕ,
+    {S : List (Finset ℕ) | IsValidCollection S ∧ S.map Finset.card = ks}.Finite := by
   sorry
 
 end Erdos363


smmercuri:

I compared this fix against issue #5643 and the audit's evidence page (https://tadamcz.com/fc-review-results/c90271f0fa/#/f/ErdosProblems/363), read the cited papers and tried to break the new statement of `erdos_363`. **The fix is correct and fully addresses the misformalization, with one small wording change requested.**

Why it is the right fix:

- **The defect is gone.** The old `.Finite` ranged over all block counts and sizes, so Skałba's varying-`k` infinitude (Ulas [Ul05], p. 332) already refuted it; now `ks` fixes both the number of blocks and their sizes, and `answer(False)` asserts that some fixed fibre is infinite.
- **Equivalent to the audit's proposal.** The RHS is byte-identical to the audit's fix (`Iff.rfl` against this branch), and I also proved in Lean that it is `↔` the Fin-indexed phrasing `∀ n (k : Fin n → ℕ), …`.
- **Faithful to the source.** Ulas's abstract states the conjecture "for fixed k ≥ 2 and l ≥ 4", and his Theorem 2 (k = l = 4) gives infinitely many solutions, so `answer(False)` is right; Bennett–Van Luijk likewise fix r and {k₁,…,k_r}.
- **No junk.** Proved in Lean: the `ks = []` fibre is exactly `{[]}`, any `ks` with an entry `< 4` has an empty fibre, and `S.map card = ks` fixes `S.length`; so no side-conditions on `ks` are needed.
- **Every change is justified and it builds.** Only the theorem RHS and docstring change; tags are untouched, the `formal_proof` link's construction (four `Icc (x+1) (x+4)` blocks, injective in the parameter) still witnesses the `[4, 4, 4, 4]` fibre, and `lake --wfail build` succeeds with only the expected `sorry` warning.

Optional nits only (not blocking): line 54, `such that$$\prod` could use spaces around `$$`.

---

### Changes requested

1. `FormalConjectures/ErdosProblems/363.lean:56-57` — "finiteness fails for a trivial reason (Skałba, see [Ul05])" is unsourced: Ulas says Skałba *proved* this in a cited paper. Replace the sentence with:
   ```
   Without this restriction finiteness fails, by a result of Skałba (see [Ul05]).
   ```

smmercuri:
🤖 *This comment was written by an automated agent (Claude, run by the PR author), not by a human.*

Applied the changes requested in https://github.com/google-deepmind/formal-conjectures/pull/5832#issuecomment-5654773603 in adea4691:
1. Docstring line 57: replaced "finiteness fails for a trivial reason (Skałba, see [Ul05])" with "finiteness fails, by a result of Skałba (see [Ul05])".

`lake --wfail build FormalConjectures.ErdosProblems.«363»` passes.

mo271 (approved):
Thanks, LGTM!

PR #4646, opened 2026-07-27
feat(Paper): mark Bézier–Bernstein Voronovskaja problem solved

KitaKen1:
This PR marks `voronovskaja_theorem.bezier_bernstein_operators` as `research solved` and adds a `formal_proof` link to the complete Lean 4 proof.

I, [KitaKen1 (Kenta Kitamura)](https://github.com/KitaKen1), completed a Lean 4 formalization of this theorem and published the proof in the following GitHub repository.
Formal proof: [immutable commit](https://github.com/KitaKen1/bezier-bernstein-voronovskaja-lean/blob/3f35c631d215b3841242275bf3ed2c59ea153a2d/Voronovskaja.lean)
Lean4Web: [Open the single-file proof in your browser](https://live.lean-lang.org/#project=MathlibDemo&url=https%3A%2F%2Fraw.githubusercontent.com%2FKitaKen1%2Fbezier-bernstein-voronovskaja-lean%2F3f35c631d215b3841242275bf3ed2c59ea153a2d%2Flean4web%2FLean4Web.lean)

The proof compiles with Lean 4.27.0, contains no `sorry` or `admit`, and `#print axioms` reports only `[propext, Classical.choice, Quot.sound]`.

AI usage disclosure: This formalization was assisted by ChatGPT 5.6 sol and Codex GPT 5.6 sol with xhigh reasoning.

PR #5833, opened 2026-09-11
fix(Paper/VoronovskajaTypeFormula): make the answer depend on α, f, x

### LINKED ISSUE #5008: Paper/VoronovskajaTypeFormula: the `answer` slot in `bezier_bernstein_operators` is scoped under `α`, `f`, `x`, so the constant function forces it to 0
## Summary

`voronovskaja_theorem.bezier_bernstein_operators` puts its `answer( )` slot **inside** the
binders `α`, `f` and `x`, so a closed answer is one real number that must be the limit for every
admissible `α`, `f` and `x`. The constant function is admissible, and on it the Bézier–Bernstein
weights telescope so the sequence is identically `0`. By uniqueness of limits the answer is
therefore **forced to `0`** — at which point the declaration asserts that
`√n (B_{n,α} f(x) − f(x)) → 0` for *every* `C²` function, contradicting the file's own "Known
Results" note and the answer both open PRs on this file propose.

https://github.com/google-deepmind/formal-conjectures/blob/638da20efd8eeeed2993fc2550fc596dc90c1ce8/FormalConjectures/Paper/VoronovskajaTypeFormula.lean#L119-L132

```lean
@[category research open, AMS 26 40 47]
theorem voronovskaja_theorem.bezier_bernstein_operators
    (α : ℝ) (hα_pos : 0 < α) (hα : α ≠ 1)
    (f : ℝ → ℝ) (x : ℝ) (hx : x ∈ I)
    (hf : ContDiffOn ℝ 2 f I) :
    Tendsto (fun n : ℕ => Real.sqrt n * (bezierBernstein n α f x - f x)) atTop
      (𝓝 answer(sorry)) := by
  sorry
```

The file's own statement of the problem is *"determine an explicit expression for it in terms of
`f`, `x`, and `α`"* — precisely the dependence the slot cannot express.

## Proof that the answer is forced to `0`

For constant `f ≡ c`,

```
B_{n,α} f(x) = c · Σ_{k=0}^{n} (J_{n,k}(x)^α − J_{n,k+1}(x)^α) = c · (J_{n,0}(x)^α − J_{n,n+1}(x)^α)
```

by telescoping, and `J_{n,0} = 1` (binomial theorem) while `J_{n,n+1} = 0` (the index set
`Finset.Icc (n+1) n` is empty), so with `α > 0` the bracket is `1 − 0 = 1` and
`B_{n,α} f(x) = c`. The scaled difference is then identically `0`.

Machine-checked with the file's definitions copied verbatim from `main` @ `638da20e`
(`lean -DautoImplicit=false -DrelaxedAutoImplicit=false -DwarningAsError=true`, Mathlib
v4.27.0, exit 0, no `sorry`):

```lean
noncomputable def bernsteinTail (n k : ℕ) : Polynomial ℝ :=
  ∑ j ∈ Finset.Icc k n, bernsteinPolynomial ℝ n j

noncomputable def bezierBernstein (n : ℕ) (α : ℝ) (f : ℝ → ℝ) (x : ℝ) : ℝ :=
  ∑ k ∈ Finset.range (n + 1),
    f (k / n) * ((bernsteinTail n k).eval x ^ α - (bernsteinTail n (k + 1)).eval x ^ α)

theorem bernsteinTail_zero (n : ℕ) : bernsteinTail n 0 = 1 := by
  rw [bernsteinTail, ← Nat.range_succ_eq_Icc_zero]
  exact bernsteinPolynomial.sum ℝ n

theorem bernsteinTail_succ_self (n : ℕ) : bernsteinTail n (n + 1) = 0 := by
  simp [bernsteinTail]

theorem bezierBernstein_const (n : ℕ) (α c x : ℝ) (hα : α ≠ 0) :
    bezierBernstein n α (fun _ ↦ c) x = c := by
  rw [bezierBernstein, ← Finset.mul_sum, Finset.sum_range_sub']
  rw [bernsteinTail_zero, bernsteinTail_succ_self]
  simp [hα]

/-- Any closed `L : ℝ` satisfying the upstream statement for all admissible `α`, `f`, `x`
is `0`. -/
theorem answer_forced_zero (L : ℝ)
    (h : ∀ (α : ℝ), 0 < α → α ≠ 1 → ∀ (f : ℝ → ℝ) (x : ℝ), x ∈ I → ContDiffOn ℝ 2 f I →
      Tendsto (fun n : ℕ => Real.sqrt n * (bezierBernstein n α f x - f x)) atTop (𝓝 L)) :
    L = 0 := by
  have hx : (0 : ℝ) ∈ I := ⟨le_refl 0, zero_le_one⟩
  have hf : ContDiffOn ℝ 2 (fun _ : ℝ ↦ (0 : ℝ)) I := contDiffOn_const
  have key := h 2 two_pos (by norm_num) (fun _ ↦ (0 : ℝ)) 0 hx hf
  have hzero : (fun n : ℕ => Real.sqrt n * (bezierBernstein n 2 (fun _ : ℝ ↦ (0 : ℝ)) 0 - 0))
      = fun _ : ℕ => (0 : ℝ) := by
    funext n
    rw [bezierBernstein_const n 2 0 0 two_ne_zero]
    ring
  rw [hzero] at key
  exact (tendsto_nhds_unique tendsto_const_nhds key).symm
```

(`bezierBernstein_const` is also proved, independently, in #4432 — its lemma of the same name.)

## Why `answer := 0` then makes the declaration false

With `L = 0` the declaration says the limit is `0` for **every** `C²` `f`, every `x ∈ [0,1]` and
every `α > 0`, `α ≠ 1`. That contradicts:

* the file's own **Known Results** section — *"Numerical experiments indicate that for
  `α ≠ 1` the quantity `√n (B_{n,α} f(x) − f(x))` may converge to a **non-zero** limit"*;
* the answer proposed in **#4646**, `μ(α)·√(x(1−x))·f′(x)` with
  `μ(α) = ∫₀^∞ ((1−Φ(t))^α − (1−Φ(t)^α)) dt`, which is non-zero for e.g. `f = id`, `x = 1/2`,
  `α ≠ 1`;
* the answer proposed in **#4432**, which is likewise a formula in `α`, `f` and `x`.

So the declaration is either false (if a closed answer is required, as in `with_auxiliary` mode,
where the answer becomes the top-level definition `<decl>._answer`) or answerable only by
capturing the binders, i.e. trivially. Either way it does not ask the source's question.

**This is load-bearing for the two open PRs on this file.** #4646 marks the declaration
`research solved` and attaches a formal proof of the `μ(α)√(x(1−x))f′(x)` formula, but leaves
`answer(sorry)` in place — and that formula cannot be written into the slot as it stands, since
it mentions `α`, `f` and `x`. #4432 records the same formula in the docstring for the same
reason. Fixing the scoping first would let either PR say what it means. I am not disputing the
mathematics in either PR; the point is only that the current statement shape cannot hold their
answer.

## The same scoping recurs twice more in the file

* `voronovskaja_theorem.bezier_bernstein_operators.variants.eventually_smooth` binds
  `let limitFormula : (ℝ → ℝ) → ℝ → ℝ := answer(sorry)` under `(α : ℝ)`, so the formula cannot
  depend on `α` — while `μ(α)` does.
* `...variants.answer_smoothness` binds `let p : ℕ × ((ℝ → ℝ) → ℝ → ℝ) := answer(sorry)` under
  `(α : ℝ)`, likewise.
* `...variants.eventually_smooth.limit_exists` is **correctly** formed: it uses `∃ L` *inside*
  all the binders. It is not part of this report.

## Suggested fix

Make the answer a function of the quantities the source says it depends on:

```lean
    Tendsto (fun n : ℕ => Real.sqrt n * (bezierBernstein n α f x - f x)) atTop
      (𝓝 ((answer(sorry) : ℝ → (ℝ → ℝ) → ℝ → ℝ) α f x))
```

and correspondingly `(answer(sorry) : ℝ → (ℝ → ℝ) → ℝ → ℝ) α` for `variants.eventually_smooth`,
`(answer(sorry) : ℝ → ℕ × ((ℝ → ℝ) → ℝ → ℝ)) α` for `variants.answer_smoothness`. The idiom is
already used correctly elsewhere in the corpus — `OpenQuantumProblems/13.lean` writes
`IsMaxMUBCount d ((answer(sorry) : ℕ → ℕ) d)`.

I have not opened a PR for this one because #4646 and #4432 are both open against the same
declaration and either would conflict; whoever lands first should probably carry the scoping
change.

## Relation to the repository's own tooling

The `AnswerLinter` message names exactly this defect — *"Move the quantifiers outward … likely
the intention was `theorem foo : answer(sorry) ↔ ∀ᵉ (bar : ℕ) …"* — but its
`stars_with_answer_sorry_iff` check fires only on statements of the form `answer(sorry) ↔ …`,
and this one has a different shape. Compare #33, #1407, and #5005, which reports five instances
of the same defect in `GreensOpenProblems`.

## What is *not* claimed

Abel's problem is untouched: whether the limit exists for `α ≠ 1`, and what it is, is exactly as
open as before. This is a report about what the Lean declaration asserts.

## Duplicate check

Offline sweep of all 1747 issues and 3237 PRs (with per-PR file lists), matched on
`Voronovskaja`, `bezier`, `bernstein`, `VoronovskajaTypeFormula` and the path
`FormalConjectures/Paper/VoronovskajaTypeFormula.lean`. Hits: #1396 (the original formalization
request), #1444 (the implementing PR), #4526 (LaTeX), #4034/#4042 (chores), #4525 (closed), and
the two open PRs #4646 and #4432 discussed above. None reports the scoping defect. Not in
#4896, #4923 or #4927.

Found while auditing `research open` declarations in the non-Wikipedia collections:

## AI assistance disclosure

Prepared with AI assistance (Claude, Anthropic) for the audit, the Lean witnesses and drafting.
The submitter reviewed the declaration, the proof, and the suggested repair, and takes

### Choose either option

smmercuri:
In `voronovskaja_theorem.bezier_bernstein_operators` the `answer(sorry)` slot sat inside the binders `α`, `f`, `x`, so a closed answer had to be a single real number valid for every admissible `α`, `f` and `x`. For a constant function the Bézier–Bernstein weights telescope to `1`, so the scaled sequence is identically `0` and the answer is forced to `0`, which then asserts a zero limit for every `C²` function. The source (Abel, Sozopol 2010) asks for the limit as an explicit expression in `f`, `x` and `α`. The `variants.eventually_smooth` and `variants.answer_smoothness` slots had the same defect with respect to `α`.

**Change**: the three answer slots are now function-typed and applied to the binders: `(answer(sorry) : ℝ → (ℝ → ℝ) → ℝ → ℝ) α f x` in the main statement, `(answer(sorry) : ℝ → (ℝ → ℝ) → ℝ → ℝ) α` in `eventually_smooth`, and `(answer(sorry) : ℝ → ℕ × ((ℝ → ℝ) → ℝ → ℝ)) α` in `answer_smoothness`. Docstrings updated accordingly. Categories and AMS tags are unchanged; `limit_exists` was already correctly formed and is untouched.

**Checked**: `lake --wfail build FormalConjectures.Paper.VoronovskajaTypeFormula` (build completed successfully, no warnings).

Note: open PRs #4646 and #4432 touch the same declaration; whichever lands second will need a trivial rebase on the answer slot.

Fixes #5008

### DIFF
--- a/FormalConjectures/Paper/VoronovskajaTypeFormula.lean
+++ b/FormalConjectures/Paper/VoronovskajaTypeFormula.lean
@@ -120,8 +120,9 @@ with shape parameter $\alpha > 0$, $\alpha \neq 1$.
 The source asks for sufficiently smooth functions. This concrete version uses
 `ContDiffOn ℝ 2 f I` as a readable baseline regularity assumption; since the
 domain is the compact interval $[0,1]$, this also explains why no separate
-boundedness assumption is included here. The variants below record the unknown
-smoothness threshold more explicitly.
+boundedness assumption is included here. The limit is sought as an explicit
+expression in $\alpha$, $f$ and $x$, so the answer is a function of these.
+The variants below record the unknown smoothness threshold more explicitly.
 -/
 @[category research open, AMS 26 40 47]
 theorem voronovskaja_theorem.bezier_bernstein_operators
@@ -129,7 +130,7 @@ theorem voronovskaja_theorem.bezier_bernstein_operators
     (f : ℝ → ℝ) (x : ℝ) (hx : x ∈ I)
     (hf : ContDiffOn ℝ 2 f I) :
     Tendsto (fun n : ℕ => Real.sqrt n * (bezierBernstein n α f x - f x)) atTop
-      (𝓝 answer(sorry)) := by
+      (𝓝 ((answer(sorry) : ℝ → (ℝ → ℝ) → ℝ → ℝ) α f x)) := by
   sorry
 
 /--
@@ -140,7 +141,7 @@ $C^m$ function on $[0,1]$ should have the asserted asymptotic formula.
 @[category research open, AMS 26 40 47]
 theorem voronovskaja_theorem.bezier_bernstein_operators.variants.eventually_smooth
     (α : ℝ) (hα_pos : 0 < α) (hα : α ≠ 1) :
-    let limitFormula : (ℝ → ℝ) → ℝ → ℝ := answer(sorry)
+    let limitFormula : (ℝ → ℝ) → ℝ → ℝ := (answer(sorry) : ℝ → (ℝ → ℝ) → ℝ → ℝ) α
     ∀ᶠ m : ℕ in atTop,
       ∀ (f : ℝ → ℝ) (x : ℝ), x ∈ I → ContDiffOn ℝ m f I →
         Tendsto (fun n : ℕ => Real.sqrt n * (bezierBernstein n α f x - f x)) atTop
@@ -164,13 +165,13 @@ theorem voronovskaja_theorem.bezier_bernstein_operators.variants.eventually_smoo
 
 /--
 Variant of the Bézier-Bernstein Voronovskaja problem with the required smoothness order itself
-left as an answer. Replacing `(answer(sorry) : ℕ × ((ℝ → ℝ) → ℝ → ℝ))` by a concrete value lets one
-state the conjecture for a chosen regularity threshold.
+left as an answer. Replacing `(answer(sorry) : ℝ → ℕ × ((ℝ → ℝ) → ℝ → ℝ))` by a concrete function
+of $\alpha$ lets one state the conjecture for a chosen regularity threshold.
 -/
 @[category research open, AMS 26 40 47]
 theorem voronovskaja_theorem.bezier_bernstein_operators.variants.answer_smoothness
     (α : ℝ) (hα_pos : 0 < α) (hα : α ≠ 1) :
-    let p : ℕ × ((ℝ → ℝ) → ℝ → ℝ) := answer(sorry)
+    let p : ℕ × ((ℝ → ℝ) → ℝ → ℝ) := (answer(sorry) : ℝ → ℕ × ((ℝ → ℝ) → ℝ → ℝ)) α
     let m := p.1
     let limitFormula := p.2
     ∀ (f : ℝ → ℝ) (x : ℝ), x ∈ I → ContDiffOn ℝ m f I →


smmercuri:

I compared this fix against issue #5008 and the proposed fix in issue #5008 (no audit page exists for this file), re-checked the primary source (Abel, *Constructive Theory of Functions, Sozopol 2010*, pp. 401–402), and tried to break the new statements of `voronovskaja_theorem.bezier_bernstein_operators` and its two `variants` in Lean. **The fix is correct and fully addresses the misformalization.**

Why it is the right fix:

- **The defect is gone.** I re-proved in Lean on this branch that the old shape forced the answer to `0` (for constant `f` the Bézier–Bernstein weights telescope to `1`, so the scaled sequence is identically `0`), whereas `#check` now shows a closed `sorry : ℝ → (ℝ → ℝ) → ℝ → ℝ` applied to `α f x`, exactly what `with_auxiliary` mode needs.
- **Equivalent to the issue's proposal.** The three new statements are the issue's suggested text; with the answer slot abstracted, I proved `PR ↔ issue` by `Iff.rfl` for all three.
- **Faithful to the source.** Abel asks "to prove its existence ... and to find an explicit expression" for `lim √n (B_{n,α}f(x) − f(x))`, which is necessarily a function of `α`, `f`, `x`.
- **No junk, no vacuity.** Under the new shape a constant `f` only pins `A α (fun _ ↦ c) x = 0` (proved, consistent), and I proved that the existential reading of the main statement is precisely the existence half of the problem for `C²` functions; `hα : α ≠ 1` and all other hypotheses are retained, and `limit_exists` is rightly untouched.
- **Every change is justified and it builds.** Only the three slots and two docstrings change; the applied-`answer` idiom is already used in `OpenQuantumProblems/13` and `GreensOpenProblems/37`, and `lake --wfail build FormalConjectures.Paper.VoronovskajaTypeFormula` succeeds with no warnings.

Optional nits only (not blocking): the `variants.eventually_smooth` docstring (L136–139) could add "the limit formula may depend on α", as the other two docstrings now do.

mo271 (approved):
Thanks, LGTM!

PR #5828, opened 2026-09-11
fix(Wikipedia/BatemanHorn): define the constant via ordered partial products

smmercuri:
`BatemanHornConstant` was defined with `∏'` (Mathlib's unconditional `tprod`) over all primes. The Bateman–Horn Euler product `∏_p (1 - 1/p)^{-k} (1 - ω_p/p)` is only conditionally convergent (in increasing order of primes): for a single irreducible quadratic the log of the local factor is `±1/p + O(1/p²)` and `∑ 1/p` diverges, so the family is not `Multipliable` and the `tprod` returns the junk value `1`. This made the statement disprovable (`X² + 1` vs `4X² + 1` get the same constant `1/2` but their counts satisfy `Count{f}(2N) = Count{g}(N) + 1`). [Wikipedia](https://en.wikipedia.org/wiki/Bateman%E2%80%93Horn_conjecture) defines `C` as the ordered product over primes and states `P(x) ~ (C/D) ∫₂ˣ dt/(log t)^m`.

**Change**: replace `BatemanHornConstant` by `BatemanHornPartialProduct polys N`, the product over the primes `p < N` (`Nat.primesBelow N`), and state the conjecture as: there is `C > 0` such that the partial products tend to `C` and `CountSimultaneousPrimes polys x ~ C / D · x / (log x)^k`. Docstrings updated to match (the `1/D` factor is now written explicitly in the asymptotic rather than folded into `C`, as in the source).

**Checked**: `lake --wfail build FormalConjectures.Wikipedia.BatemanHornConjecture` (build completed successfully, no warnings).

Fixes #5255

PR #5835, opened 2026-09-11
fix(Wikipedia/HardyLittlewood): state the asymptotic, require distinct offsets

### LINKED ISSUE #4996: Wikipedia/HardyLittlewood: `first_hardy_littlewood_conjecture` states `=O` where the source states `∼`, and does not require the offsets distinct
## Summary

`HardyLittlewood.first_hardy_littlewood_conjecture` has two independent defects.

1. It states `=O[atTop]` where its own docstring — and Wikipedia, and Hardy–Littlewood —
   state `∼`. For admissible tuples the surviving one-sided bound is a **classical
   sieve theorem**, so the declaration no longer expresses an open problem.
2. The offsets `m` are not required to be distinct. At `m = ![0, 1, 1]` the Euler
   product diverges, Mathlib's `tprod` falls back to its junk value `1`, the constant
   collapses to `C = 4`, and the resulting bound is **false** if the twin-prime count
   has its conjectured order.

https://github.com/google-deepmind/formal-conjectures/blob/638da20efd8eeeed2993fc2550fc596dc90c1ce8/FormalConjectures/Wikipedia/HardyLittlewood.lean#L61-L85

```lean
def FirstHardyLittlewoodConjectureFor {k : ℕ} (m : Fin k.succ → ℕ) : Prop :=
  let C : ℝ :=
      2 ^ k * ∏' (q : { q : ℕ // q.Prime ∧ 3 ≤ q}),
        (1 - (Nat.numResidues q m : ℝ) / q) / (1 - 1 / q) ^ k.succ
    let π_P : ℕ → ℝ := fun n => (Nat.primeTupleCounting m n : ℝ)
    π_P =O[atTop] fun n => C * ∫ t in (2)..n, 1 / t.log ^ k.succ

/-- … Then
$$
  \pi_P(n)\sim C_P\int_2^n\frac{dt}{\log^{k+1}t}.
$$
-/
theorem first_hardy_littlewood_conjecture {k : ℕ} (m : Fin k.succ → ℕ) :
    FirstHardyLittlewoodConjectureFor m := by
  sorry
```

## Defect 1 — `∼` weakened to `=O`

`f ∼ g` is `f / g → 1`; `f =O[atTop] g` is a one-sided bound with an arbitrary implied
constant. Replacing the first by the second discards the entire main-term content of the
conjecture — the constant `C_P` is computed with care in the file and then made
irrelevant, since `=O` tolerates any multiple of it.

Worse, what survives is provable. For a fixed admissible constellation with distinct
offsets, the Brun/Selberg sieve gives

$$\pi_P(n) \;\ll_k\; \mathfrak{S}(P)\,\frac{n}{(\log n)^{k+1}}$$

— the standard upper-bound-sieve estimate for prime `k`-tuples (Halberstam–Richert,
*Sieve Methods*, Ch. 5) — and `∫₂ⁿ dt/(log t)^{k+1} ≍ n/(log n)^{k+1}`. So the
declaration's right-hand side dominates `π_P` up to a constant by a classical theorem;
no open question is being asked. The degenerate cases are consistent with this: if the
offsets cover every residue class mod some prime `q ≥ 3` then `C = 0`, but then no `p`
is admissible either and `π_P ≡ 0`, so `=O 0` still holds.

## The constant itself is correct

Worth recording, because it is easy to suspect the `2 ^ k` prefactor. For the twin tuple
`m = ![0, 1]` (`k = 1`, `w(q) = 2` for `q ≥ 3`) the Lean's `C` evaluates to

```
2 · ∏_{3 ≤ q < 2·10⁶} (1 − 2/q)/(1 − 1/q)²  =  1.320323674
2·C₂  (twin prime constant)                 =  1.320323632…
```

The `2 ^ k` prefactor is exactly the missing `q = 2` Euler factor: all offsets `2·m i`
are even, so `w(2) = 1` and `(1 − 1/2)/(1 − 1/2)^{k+1} = 2^k`. Using residues of `m i`
rather than of `2·m i` is harmless because doubling is a bijection modulo every odd `q`.
**No defect in `C`.**

## Defect 2 — offsets are not required distinct

The theorem is `∀ m`, and nothing forbids repeats. Take `k = 2`,
**`m = ![0, 1, 1]`**. It satisfies the file's own shape conditions: `m 0 = 0` and
`∀ i ≠ 0, 0 < m i`.

* `IsPrimeConstellation m p` unfolds to "`p` and `p + 2` are both prime", since the
  index set `{p, p+2, p+2}` has only two distinct entries. Admissibility holds for every
  prime `q`: for `q ≥ 3` the image has at most 2 < q elements, and for `q = 2` both
  entries are congruent mod 2. So `Nat.primeTupleCounting m n = π₂(n)`, the count of
  twin-prime starts `p ≤ n`.
* `Nat.numResidues q m = #{0, 1} = 2` for every `q ≥ 3`, while `k.succ = 3`. Each Euler
  factor is therefore `(1 − 2/q)/(1 − 1/q)³ = 1 + 1/q + O(1/q²) > 1`, and the partial
  products **diverge** (`log` of the product is `∑ 1/q + O(1)`):

  | `Q` | `∏_{3 ≤ q ≤ Q}` |
  |---|---|
  | `10` | 1.4954 |
  | `10²` | 2.7485 |
  | `10³` | 4.0773 |
  | `10⁴` | 5.4215 |
  | `10⁵` | 6.7705 |
  | `10⁶` | 8.1224 |
  | `10⁷` | 9.4759 |

* So the family is not `Multipliable` over `ℝ`, Mathlib's `∏'` returns its junk value
  `1`, and `C = 2² · 1 = 4`.
* The asserted bound is then `π₂(n) =O[atTop] 4·∫₂ⁿ dt/(log t)³ ≍ n/(log n)³`.

This instance is **false as soon as `π₂(n) ≫ n/(log n)³`**, which is exactly what the
conjecture the declaration is named after predicts: at `m = ![0, 1]` the docstring's
`∼` gives `π₂(n) ∼ 2C₂ · n/(log n)²`. In other words the declaration, read over all `m`,
contradicts its own intended content — if the First Hardy–Littlewood conjecture is true
then `first_hardy_littlewood_conjecture` is false.

I state this as a hypothesis gap rather than a counterexample, because the falsity is
conditional: no unconditional lower bound on `π₂` is known, so `m = ![0, 1, 1]` cannot
be *proved* to refute the declaration today.

The source's `m_i` are the offsets of a constellation and are implicitly distinct
(`P = (m_1, …, m_k)` with the `p + m_i` pairwise different primes), so a hypothesis such
as `Function.Injective m` — or `StrictMono m` — belongs in `FirstHardyLittlewoodConjectureFor`.

**Both defects are about the formalization, not about primes.** Nothing here bears on
the First Hardy–Littlewood conjecture, which remains open exactly as stated in the
source. In the four-coordinate sense: (1) the underlying conjecture is untouched;
(2) no formal solution exists; (3) the declaration is **unfaithful** to the source in
two separate ways; (4) what it literally asserts is, for distinct admissible offsets, a
classical sieve bound, and for repeated offsets a statement that the conjecture itself
would refute. Only (3) and (4) are claimed.

Note that `not_first_and_secondHardyLittlewoodConjecture` (Richards) is stated in terms
of `FirstHardyLittlewoodConjectureFor` as an *antecedent*, so weakening that predicate
makes the Richards implication correspondingly harder to prove — it is stated with the
`=O` reading and would need revisiting alongside any fix to defect 1.

## Suggested fixes

* Defect 1: replace `π_P =O[atTop] …` with the asymptotic the docstring states, e.g.
  `atTop.Tendsto (fun n => π_P n / (C * ∫ t in (2)..n, 1 / t.log ^ k.succ)) (𝓝 1)`, or
  `Asymptotics.IsEquivalent atTop π_P (fun n => C * ∫ …)`.
* Defect 2: add an injectivity/strict-monotonicity hypothesis on `m` in
  `FirstHardyLittlewoodConjectureFor`.

I have not opened a PR for this one: fixing defect 1 changes the statement that
`not_first_and_secondHardyLittlewoodConjecture` quantifies over, so the right shape for
that theorem should be agreed first.

faithfulness to their cited sources. Full derivation, the Euler-product computations,
and the duplicate search:

## Source/status note

Wikipedia, *First Hardy–Littlewood conjecture*, states the asymptotic relation
`π_P(n) ∼ C_P ∫₂ⁿ dt/(log t)^{k+1}` with
`C_P = 2^k ∏_{q ≥ 3 prime} [1 − w(q; m₁,…,m_k)/q] / [(1 − 1/q)^{k+1}]`, for `P` a tuple
of positive even offsets — i.e. the docstring in this file is a faithful transcription
of the source, and it is the Lean that diverges from both.

The Euler-product numbers above were computed with exact prime enumeration to `10⁷`; the
`Q = 10²` … `10⁶` rows agree with an earlier independent run of the same computation.
`git log` on this file shows no statement-level edit since creation (only docstring,
whitespace and classical-reasoning chores).

## AI assistance disclosure

This report was prepared with AI assistance (Claude, Anthropic) for the source audit, the
numerical checks, and drafting. The submitter reviewed the declaration, the argument, and
the suggested repairs, and takes responsibility for the submission.

### Choose either option

smmercuri:
`FirstHardyLittlewoodConjectureFor m` stated `π_P =O[atTop] fun n => C_P * ∫ t in (2)..n, 1 / t.log ^ k.succ`, a one-sided bound with an arbitrary implied constant. For distinct admissible offsets this bound is a classical sieve theorem (Brun–Selberg), so the declaration did not express an open problem, while Wikipedia (and the file's own docstring) state the asymptotic `π_P(n) ∼ C_P ∫_2^n dt / log^{k+1} t`. Independently, nothing required the offsets `m i` to be distinct: at `m = ![0, 1, 1]` the Euler product diverges, Mathlib's `∏'` returns its junk value `1`, and the resulting bound `π₂(n) =O n / (log n)^3` contradicts the conjecture itself.

**Change**
- `FirstHardyLittlewoodConjectureFor` now assumes `m 0 = 0` and `Function.Injective m` (so the offsets `2 * m i`, `i ≠ 0`, are distinct positive even integers) and states `π_P ~[atTop] fun n => C * ∫ t in (2)..n, 1 / t.log ^ k.succ` using `Asymptotics.IsEquivalent`. The hypotheses live inside the predicate so that `not_first_and_secondHardyLittlewoodConjecture` (Richards) still quantifies over the full conjecture; with the `∼` reading its antecedent now implies the prime `k`-tuples conjecture, as Richards' argument needs.
- Docstrings updated: the offsets are distinct, and the predicate documents the asymptotic it states.

**Reviewer notes**
- The Euler product is kept as `∏'`: with distinct offsets `w(q) = k + 1` for every prime `q` larger than all `m i`, so the factor is `1 + O(1/q²)` and the product converges absolutely (unlike the Bateman–Horn constant of #5828, which is only conditionally convergent). If some odd prime `q` has `w(q) = q` then the product is `0` and `π_P ≡ 0`, so the statement is trivially `0 ~ 0`, matching the source's restriction to admissible tuples.
- `m 0 = 0` is needed alongside injectivity: without it `IsPrimeConstellation m p` never holds, `π_P ≡ 0`, and `0 ~ C_P ∫ …` would be false for `C_P ≠ 0`.

**Checked**: `lake --wfail build FormalConjectures.Wikipedia.HardyLittlewood` — Build completed successfully.

Fixes #4996

### DIFF
--- a/FormalConjectures/Wikipedia/HardyLittlewood.lean
+++ b/FormalConjectures/Wikipedia/HardyLittlewood.lean
@@ -22,7 +22,7 @@ import FormalConjecturesUtil
 *Reference:* [Wikipedia](https://en.wikipedia.org/wiki/First_Hardy%E2%80%93Littlewood_conjecture)
 -/
 
-open Filter
+open Asymptotics Filter
 
 open scoped Nat.Prime
 
@@ -58,15 +58,23 @@ noncomputable def Nat.primeTupleCounting {k : ℕ} (m : Fin k.succ → ℕ) (n :
   open scoped Classical in
   Nat.count (IsAdmissiblePrimeConstellation m) n.succ
 
+/--
+The first Hardy–Littlewood conjecture for the tuple $(m_0, m_1, \dots, m_k)$: if $m_0 = 0$ and
+the $m_i$ are pairwise distinct, then
+$$
+  \pi_P(n)\sim C_P\int_2^n\frac{dt}{\log^{k+1}t}.
+$$
+-/
 def FirstHardyLittlewoodConjectureFor {k : ℕ} (m : Fin k.succ → ℕ) : Prop :=
-  let C : ℝ :=
+  m 0 = 0 → Function.Injective m →
+    let C : ℝ :=
       2 ^ k * ∏' (q : { q : ℕ // q.Prime ∧ 3 ≤ q}),
         (1 - (Nat.numResidues q m : ℝ) / q) / (1 - 1 / q) ^ k.succ
     let π_P : ℕ → ℝ := fun n => (Nat.primeTupleCounting m n : ℝ)
-    π_P =O[atTop] fun n => C * ∫ t in (2)..n, 1 / t.log ^ k.succ
+    π_P ~[atTop] fun n => C * ∫ t in (2)..n, 1 / t.log ^ k.succ
 
 /--
-Let $P = (m_1, \dots, m_k)$ be a tuple of positive even integers. Let
+Let $P = (m_1, \dots, m_k)$ be a tuple of distinct positive even integers. Let
 $\pi_P(n)$ denote the number of primes $p\leq n$ such that $(p, p + m_1, \dots, p + m_k)$
 forms an admissible prime constellation. Let $w(q; m_1, \dots, m_k)$ denote the
 number of distinct residues of $0, m_1, \dots, m_k$ modulo $q$, and let


smmercuri:

I compared this fix against the two proposed fixes in issue #4996 and the Wikipedia statement, and tried to break `FirstHardyLittlewoodConjectureFor` on degenerate tuples. **The fix is correct and fully addresses the misformalization.**

Why it is the right fix:

- **The defect is gone.** `=O[atTop]` (a classical sieve bound) becomes `~[atTop]`, and `Function.Injective m` rules out repeated offsets like `![0, 1, 1]`, where `∏'` returned its junk value. The added `m 0 = 0` is needed once `~` is used: for `![1, 2]`, `π_P ≡ 0` but `C = 2C₂ > 0`, so the `∀ m` theorem would be refutable.
- **Relation to the issue's proposals.** Proved in Lean against this branch that the PR's predicate is `Iff.rfl`-equal to the issue's `IsEquivalent` option; the issue's `Tendsto (π_P / (C·∫)) (𝓝 1)` option is strictly stronger, and I proved it *false* at the non-admissible tuple `![0, 1, 2]` (`0 / 0 = 0 ↛ 1`), so the PR rightly chose `IsEquivalent`.
- **Faithful to the source.** Wikipedia states `π_P(n) ∼ C_P ∫_2^n dt / log^{k+1} t` for positive even offsets that do "not form a complete residue class with respect to any prime"; admissibility is built into both `π_P` (via `IsAdmissiblePrimeConstellation`) and `C` (`w(q) = q` gives a zero factor), and the constant is unchanged and correct.
- **No junk values or vacuity.** Non-admissible tuples give the trivial `0 ~ 0` (proved in Lean for `![0, 1, 2]`); admissible tuples give factors `1 + O(1/q²)`, so `∏'` genuinely converges; `k = 0` is the prime number theorem, but the theorem over all `k` remains open.
- **Every change is justified and it builds.** `open Asymptotics` is needed for `~[l]`, the docstrings match, and hypotheses inside the predicate let `not_first_and_secondHardyLittlewoodConjecture` still quantify over the full conjecture; `lake --wfail build FormalConjectures.Wikipedia.HardyLittlewood` succeeds.

Optional nits only (not blocking): the docstring of `FirstHardyLittlewoodConjectureFor` could note that non-admissible tuples make the statement vacuous (`0 ~ 0`).

mo271 (approved):
Thanks, LGTM!

PR #5836, opened 2026-09-11
fix(Wikipedia/Koethe): drop unused nil-ideal binders, fix docstring

### LINKED ISSUE #5002: Wikipedia/Koethe: `matrixOver_KotherRadical` carries an unused `{I}`/`hI`, and its docstring describes a different statement from the (correct) code
## Summary

Low-severity, two small things in
`Koethe.KotherConjecture.variants.matrixOver_KotherRadical`:

1. the hypotheses `{I : TwoSidedIdeal R} (hI : IsNil I)` do **not** occur in the
   conclusion — they are spurious, unlike in the two sibling declarations where the same
   `I` is the subject;
2. the docstring describes something different from (and weaker than) the code, which is
   itself correct.

https://github.com/google-deepmind/formal-conjectures/blob/638da20efd8eeeed2993fc2550fc596dc90c1ce8/FormalConjectures/Wikipedia/Koethe.lean#L74-L81

```lean
open scoped Classical in
/-- The **Köthe conjecture**: for any positive integer `n`, the Köthe radical of `R` is the matrix ideal `M_2(Nil*(R))`. -/
@[category research open, AMS 16]
theorem KotherConjecture.variants.matrixOver_KotherRadical
    {I : TwoSidedIdeal R} (hI : IsNil I) (n : Type*) [Fintype n] :
    matrix n (Nil* R) = Nil* (Matrix n n R) := by
  sorry
```

## Spurious hypotheses

`I` and `hI` appear nowhere in `matrix n (Nil* R) = Nil* (Matrix n n R)`. `I` is
implicit and otherwise unconstrained, so it can never be inferred by unification and can
only ever be supplied by hand at a use site.

This is **not** a vacuity problem: `IsNil I` is satisfiable (`I = ⊥` is nil), so the
declaration is logically equivalent to the same statement with the two binders removed —
it is not accidentally trivially true, it is just carrying dead weight. Hence "low
severity". Still worth fixing, because a reader comparing the three matrix variants sees
`(hI : IsNil I)` in all of them and reasonably assumes it plays the same role, and
because the two siblings

```lean
theorem KotherConjecture.variants.general_matrix {I : TwoSidedIdeal R} (hI : IsNil I)
    (n : Type*) [Fintype n] : IsNil (matrix n I)

theorem KotherConjecture.variants.two_by_two_matrix {I : TwoSidedIdeal R} (hI : IsNil I) :
    IsNil (matrix (Fin 2) I)
```

*do* use them.

## Docstring does not match the code

The docstring says "for any positive integer `n`, the **Köthe radical of `R`** is the
matrix ideal `M_2(Nil*(R))`". The code says `matrix n (Nil* R) = Nil* (Matrix n n R)`,
i.e. the Köthe radical of the **matrix ring** is the matrix ideal over the Köthe radical
of `R`. Two mismatches:

* `Nil*(R)` vs `Nil*(M_n(R))` on the right of the equation — the docstring names the
  wrong ring;
* `M_2` vs an arbitrary `n : Type*` `[Fintype n]` — the docstring fixes the size while
  the code quantifies over it (`M_2` is the sibling `two_by_two_matrix`).

**The code is the correct one.** Wikipedia's equivalent formulation 6 of the Köthe
conjecture reads

> For any ring `R`, the upper nilradical of `M_n(R)` is the set of matrices with entries
> from the upper nilradical of `R` for every positive integer `n`

which is exactly `matrix n (Nil* R) = Nil* (Matrix n n R)`. Only the docstring needs
changing.

## Not a defect

`n : Type*` `[Fintype n]` admits `n = Empty`; both sides then live in the trivial ring
and are equal, so the empty case is harmless and no degeneracy hides there.

## Suggested fix

```lean
open scoped Classical in
/-- The **Köthe conjecture**: for any finite index type `n`, the Köthe radical of the matrix
ring `M_n(R)` is the ideal of matrices with entries in the Köthe radical of `R`. -/
@[category research open, AMS 16]
theorem KotherConjecture.variants.matrixOver_KotherRadical (n : Type*) [Fintype n] :
    matrix n (Nil* R) = Nil* (Matrix n n R) := by
  sorry
```

I have not opened a PR: removing binders from a declaration is the kind of change a
maintainer may prefer to bundle with the file's other pending tidy-up (the namespace is
spelled `KotherConjecture` throughout, apparently a typo for `KotheConjecture`, and the
first declaration is `KotheConjecture` with no `variants` prefix). Happy to send one if
that is preferred.

**This is a defect in the formalization only, and a cosmetic one.** The Köthe conjecture
is untouched; the declaration is not false and not vacuous, and it is a correct
statement of formulation 6. In the four-coordinate sense: (1) the underlying conjecture
is untouched and open; (2) no formal solution exists; (3) the *docstring* is unfaithful
while the code is faithful; (4) what the declaration literally asserts is right, just
with two unused binders attached. Only (3) is at issue.

faithfulness to their cited sources. Full derivation and the duplicate search:

## Source/status note

Wikipedia, *Köthe conjecture*, "Equivalent formulations", item 6, quoted above. The
Köthe conjecture is open. `git log` on this file shows only formatting and
classical-reasoning chores since the file was created, so the docstring has said this
since the beginning; `matrixOver_KotherRadical` returns no issue or PR hits.

## AI assistance disclosure

This report was prepared with AI assistance (Claude, Anthropic) for the source audit and
drafting. The submitter reviewed the declaration, the source text, and the suggested
repair, and takes responsibility for the submission.

### Choose either option

smmercuri:
`Koethe.KotherConjecture.variants.matrixOver_KotherRadical` took hypotheses `{I : TwoSidedIdeal R} (hI : IsNil I)` that do not occur in its conclusion `matrix n (Nil* R) = Nil* (Matrix n n R)`, and its docstring said "the Köthe radical of `R` is the matrix ideal `M_2(Nil*(R))`". The Lean statement itself is Wikipedia's equivalent formulation 6 (the upper nilradical of `M_n(R)` is the set of matrices with entries in the upper nilradical of `R`, for every `n`), so the docstring named the wrong ring and fixed the size.

**Change**: remove the two unused binders and reword the docstring to describe the statement as written (Köthe radical of `M_n(R)` equals `M_n(Nil*(R))` for any finite index type `n`). No change to the statement's meaning or attributes.

**Checked**: `lake --wfail build FormalConjectures.Wikipedia.Koethe` — Build completed successfully.

Fixes #5002

### DIFF
--- a/FormalConjectures/Wikipedia/Koethe.lean
+++ b/FormalConjectures/Wikipedia/Koethe.lean
@@ -72,10 +72,10 @@ theorem KotherConjecture.variants.two_by_two_matrix {I : TwoSidedIdeal R} (hI : 
   sorry
 
 open scoped Classical in
-/-- The **Köthe conjecture**: for any positive integer `n`, the Köthe radical of `R` is the matrix ideal `M_2(Nil*(R))`. -/
+/-- The **Köthe conjecture**: for any finite index type `n`, the Köthe radical of the matrix
+ring `M_n(R)` is the ideal `M_n(Nil*(R))` of matrices with entries in the Köthe radical of `R`. -/
 @[category research open, AMS 16]
-theorem KotherConjecture.variants.matrixOver_KotherRadical
-    {I : TwoSidedIdeal R} (hI : IsNil I) (n : Type*) [Fintype n] :
+theorem KotherConjecture.variants.matrixOver_KotherRadical (n : Type*) [Fintype n] :
     matrix n (Nil* R) = Nil* (Matrix n n R) := by
   sorry
 


smmercuri:

I compared this fix against the proposed fix in issue #5002 and the cited Wikipedia page, and tried to break the new statement of `KotherConjecture.variants.matrixOver_KotherRadical` in `FormalConjectures/Wikipedia/Koethe.lean`. **The fix is correct and fully addresses the misformalization.**

Why it is the right fix:

- **The defect is gone.** The dead binders `{I : TwoSidedIdeal R} (hI : IsNil I)`, which never occurred in the conclusion, are removed, and the docstring now describes what the code states: `Nil*(M_n(R)) = M_n(Nil*(R))` for any finite index type `n`, rather than "the Köthe radical of `R` is `M_2(Nil*(R))`".
- **Equivalent to the issue's proposal.** The new Lean statement is character-identical to the one suggested in issue #5002; I also proved `Original ↔ Proposed` in Lean against this branch (the old hypothesis was always dischargeable since `IsNil ⊥`), confirming the binders were dead weight rather than a vacuity bug.
- **Faithful to the source.** Wikipedia's equivalent formulation 6 reads "the upper nilradical of M_n(R) is the set of matrices with entries from the upper nilradical of R for every positive integer n", which is exactly `matrix n (Nil* R) = Nil* (Matrix n n R)`; the statement elaborates to `TwoSidedIdeal.matrix`, whose `mem_matrix` is "every entry lies in `I`".
- **No junk values or vacuity.** For empty `n` both sides are ideals of a subsingleton ring and the equation is trivially true, which is the correct reading of "for every n"; for general `n` the statement is not `rfl`- or `simp`-trivial, and no `ℕ`-subtraction or `answer(sorry)` scoping is involved.
- **Every change is justified and it builds.** The only changes are the two binder removals and the docstring rewording (the sibling `general_matrix`/`two_by_two_matrix` genuinely use their `I`, so they are rightly untouched); no attribute or status change, and `lake --wfail build FormalConjectures.Wikipedia.Koethe` succeeds with no warnings.

mo271:
perhaps we can land #5268 first?

mo271 (approved):

PR #5839, opened 2026-09-11
fix(Wikipedia/UnionClosed): state tightness as a maximum, not for every element

### LINKED ISSUE #5001: Wikipedia/UnionClosed: `cardinality_even_of_union_closed_tight` assumes every element hits half the family, which is stronger than "tight" (and forces `n` finite)
## Summary

`UnionClosed.union_closed.variants.cardinality_even_of_union_closed_tight` states its
hypothesis as `∀ i, #{x ∈ A | i ∈ x} = #A / 2` — *every* element of the ambient type
lies in exactly half the members. That is strictly stronger than "the union-closed
conjecture is tight for `A`", which the docstring claims it is, and it silently forces
the ambient type to be finite and exactly covered by `⋃ A`. The binder `hA` is also used
twice, shadowing the first hypothesis.

https://github.com/google-deepmind/formal-conjectures/blob/638da20efd8eeeed2993fc2550fc596dc90c1ce8/FormalConjectures/Wikipedia/UnionClosed.lean#L200-L211

```lean
/--
If the UC conjecture is tight for some family `A` then $|A| = 2^k$ for some $k$.

Reference: Conjecture 3 in https://www.nieuwarchief.nl/serie5/pdf/naw5-2023-24-4-225.pdf.
-/
theorem union_closed.variants.cardinality_even_of_union_closed_tight
    [Nonempty n] (hA : A ≠ {∅} ∧ A ≠ ∅) (hA : IsUnionClosed A)
    (UCC_tight : ∀ i, #{x ∈ A | i ∈ x} = (1 / 2 : ℝ) * #A) :
    ∃ k, #A = 2 ^ k := by
  sorry
```

## Defect 1 — `UCC_tight` is not "tight"

The conjecture, as this very file states it two hundred lines above, is

```lean
theorem union_closed … : ∃ i : n, (1 / 2 : ℚ) * #A ≤ #{x ∈ A | i ∈ x}
```

so "the conjecture is tight for `A`" means the bound is *attained and not exceeded*:

```
max_{i} #{x ∈ A | i ∈ x} = #A / 2
```

The Lean hypothesis is `∀ i`, not `max_i`. It demands the exact half-count from every
single element of `n`, including elements that appear in no member of `A`. Since
`∀ i, … = #A/2` implies `max_i … = #A/2` but not conversely, the Lean theorem is
**implied by** the theorem the docstring describes, i.e. it is strictly weaker than
Conjecture 3 of the cited NAW article.

The gap is not empty. Exhaustive enumeration over all `2^(2^n)` subfamilies of `P([n])`
(union-closed, `A ≠ {∅}`, `A ≠ ∅`):

| `n` | families satisfying the Lean's `∀ i` | families satisfying `max_i = #A/2` |
|---|---|---|
| 1 | 1 | 1 |
| 2 | 2 | 4 |
| 3 | 5 | 14 |
| 4 | 15 | 51 |

So at `n = 4` the Lean hypothesis captures 15 of the 51 tight families and the
declaration says nothing about the other 36.

For the record, the conclusion holds in that whole range under **both** readings —
`#A` is a power of 2 for every one of the 51 as well as every one of the 15 — so this
is a fidelity report, not a counterexample, and no counterexample can come from a
divergence in this direction.

## Defect 2 — the hypothesis silently forces `n` finite

Suppose some `i : n` lies in no member of `A`. Then `#{x ∈ A | i ∈ x} = 0`, so
`UCC_tight i` gives `0 = #A / 2`, hence `#A = 0`, i.e. `A = ∅` — contradicting the first
hypothesis. Therefore `UCC_tight` forces every element of `n` to lie in some member of
`A`, so `n` injects into the (finite) union of the members of `A` and is finite.

That is a real constraint arriving through the back door: the declaration carries only
`[Nonempty n]`, while the sibling `union_closed.variants.univ_card` asks for `[Fintype n]`
explicitly. A reader comparing the two would reasonably conclude that this one applies to
infinite ambient types, and it does not (over an infinite `n` the hypotheses are simply
unsatisfiable, making the declaration vacuous there).

## Defect 3 — shadowed binder

```lean
    [Nonempty n] (hA : A ≠ {∅} ∧ A ≠ ∅) (hA : IsUnionClosed A)
```

The second `hA` shadows the first, which becomes inaccessible by name in any proof
attempt. Both should be named, e.g. `h_ne` and `h_union_closed`, as the other five
declarations in this file already do
(`h_ne_singleton_empty` / `h_union_closed`). This predates PR #1255, which changed the
first hypothesis from `A ≠ {∅}` to `A ≠ {∅} ∧ A ≠ ∅` but kept the name.

## Suggested fix

```lean
theorem union_closed.variants.cardinality_even_of_union_closed_tight
    [Nonempty n] (h_ne_singleton_empty : A ≠ {∅}) (h_ne_empty : A ≠ ∅)
    (h_union_closed : IsUnionClosed A)
    (h_tight : IsGreatest {c : ℝ | ∃ i : n, c = #{x ∈ A | i ∈ x}} ((1 / 2 : ℝ) * #A)) :
    ∃ k, #A = 2 ^ k := by
  sorry
```

or any other rendering of "`#A / 2` is the maximum of `i ↦ #{x ∈ A | i ∈ x}`". If the
current `∀ i` form is wanted as well, it deserves its own name and a docstring that says
it is stronger than tightness. I have not opened a PR because the right encoding of
"tight" is a judgement call for the maintainers — `IsGreatest`, an explicit
`∃ i, … = … ∧ ∀ j, … ≤ …`, or a `Finset.sup'` — and because the fix should be checked
against Conjecture 3 in the NAW source, which I could not read directly (see below).

**This is a defect in the formalization only.** The union-closed sets conjecture and
Conjecture 3 of the cited article are both untouched by this report, and no
counterexample to either is claimed — the exhaustive check above found none, in either
reading. In the four-coordinate sense: (1) the underlying conjecture is untouched and
open; (2) no formal solution exists; (3) the declaration is **unfaithful** to the
docstring and to the cited Conjecture 3; (4) what it literally asserts is a strictly
weaker open statement, not a false one. Only (3) is at issue, plus the naming defect.

faithfulness to their cited sources. Full derivation, the exhaustive enumeration, and the
duplicate search:

https://github.com/Kuberwastaken/c5-k4/blob/d8a0c47da83203e1cd2c739b383f7a436435549f/results/expansion/live-search-2026-08-15/wikipedia-scan-M-Z.md

## Source/status note

The reading of "tight" used above is taken from this file's own statement of the
conjecture (`∃ i, (1/2) * #A ≤ #{x ∈ A | i ∈ x}`), not from the NAW article, which I did
not retrieve — so the precise form of its Conjecture 3 should be checked before any fix
lands. What is independent of that check is Defect 2 (the implicit finiteness) and
Defect 3 (the shadowed binder), and the fact that the `∀ i` and `max_i` readings are
genuinely different hypotheses, which the table above demonstrates.

Also recorded, and **not** a defect: the headline `union_closed` omits `A ≠ ∅`, but that
is harmless — at `A = ∅` both sides of the inequality are `0` and the conclusion holds
for any `i`, which exists by `[Nonempty n]`.

## AI assistance disclosure

This report was prepared with AI assistance (Claude, Anthropic) for the audit, the
exhaustive enumeration, and drafting. The submitter reviewed the declaration, the
argument, and the suggested repair, and takes responsibility for the submission.

### Choose either option

smmercuri:
`union_closed.variants.cardinality_even_of_union_closed_tight` assumed `∀ i, #{x ∈ A | i ∈ x} = #A / 2`, i.e. every element of the ambient type lies in exactly half of the members of `A`. Conjecture 3 of the cited NAW article (Roos, 2023) only assumes that Frankl's conjecture is *tight* for `A`, which the article uses to mean that the largest number of members containing a common element equals `#A / 2` (Theorem 2: "the largest number of ones per column equals …, which is tight for the UCC if and only if …"). The `∀ i` form is strictly stronger (it also forces every element of `n` to lie in some member, hence `n` finite), so the Lean theorem was strictly weaker than the source. The binder `hA` was also used twice.

**Change**
- Replace the hypothesis by `IsGreatest (Set.range fun i : n => (#{x ∈ A | i ∈ x} : ℚ)) ((1 / 2 : ℚ) * #A)`, using `ℚ` as in the file's statement of the conjecture.
- Split `hA : A ≠ {∅} ∧ A ≠ ∅` / `hA : IsUnionClosed A` into `h_ne_singleton_empty`, `h_ne_empty`, `h_union_closed`, matching the sibling declarations. `A ≠ ∅` is kept since the empty family is tight in this sense but has cardinality `0`.
- Docstring now defines tightness.

**Checked**: `lake --wfail build FormalConjectures.Wikipedia.UnionClosed` — Build completed successfully.

Fixes #5001

### DIFF
--- a/FormalConjectures/Wikipedia/UnionClosed.lean
+++ b/FormalConjectures/Wikipedia/UnionClosed.lean
@@ -199,14 +199,21 @@ theorem union_closed.variants.sharpness [Fintype n] (c : ℝ) (hc : 1 / 2 < c) :
   simp at this
 
 /--
-If the UC conjecture is tight for some family `A` then $|A| = 2^k$ for some $k$.
+The union-closed sets conjecture is tight for a union-closed family `A` if the largest number of
+sets in `A` containing a common element is exactly half of the sets in `A`, that is,
+$\max_i |\{X \in A : i \in X\}| = |A| / 2$.
+
+If the union-closed sets conjecture is tight for some family `A` then $|A| = 2^k$ for some $k$.
 
 Reference: Conjecture 3 in https://www.nieuwarchief.nl/serie5/pdf/naw5-2023-24-4-225.pdf.
 -/
 @[category research open, AMS 5]
 theorem union_closed.variants.cardinality_even_of_union_closed_tight
-    [Nonempty n] (hA : A ≠ {∅} ∧ A ≠ ∅) (hA : IsUnionClosed A)
-    (UCC_tight : ∀ i, #{x ∈ A | i ∈ x} = (1 / 2 : ℝ) * #A) :
+    [Nonempty n]
+    (h_ne_singleton_empty : A ≠ {∅})
+    (h_ne_empty : A ≠ ∅)
+    (h_union_closed : IsUnionClosed A)
+    (h_tight : IsGreatest (Set.range fun i : n => (#{x ∈ A | i ∈ x} : ℚ)) ((1 / 2 : ℚ) * #A)) :
     ∃ k, #A = 2 ^ k := by
   sorry
 


smmercuri:

I compared this fix against the proposed fix in issue #5001 and Conjecture 3 of the cited NAW article (Roos, 2023), and tried to break the new `union_closed.variants.cardinality_even_of_union_closed_tight` in Lean. **The fix is correct and fully addresses the misformalization, with one small docstring suggestion.**

Why it is the right fix:

- **The defect is gone.** The old `∀ i, #{x ∈ A | i ∈ x} = #A / 2` forced *every* element of `n` to lie in exactly half the members, refuted by the paper's own Theorem 1 example (power set of `[k]` inside `[n]`, zero columns, called "tight"); the new `IsGreatest` hypothesis says exactly "the maximum is `#A / 2`".
- **Equivalent to the proposal in issue #5001.** Proved in Lean against this branch: PR's `ℚ`/`Set.range` form ↔ the issue's `ℝ`/set-builder form ↔ the explicit `(∃ i, 2 * #{…} = #A) ∧ ∀ i, 2 * #{…} ≤ #A`; also old ⟹ new, and the converse fails on `{∅, {0}} ⊆ Fin 2`.
- **Faithful to the source.** Roos's Theorem 2: "the largest number of ones per column equals min(m, N/2), which is tight for the UCC if and only if m = N" — tightness concerns the maximum, as encoded here.
- **No junk values.** `∅` is tight in the new sense with `#∅ = 0` not a power of two, so `A ≠ ∅` is needed; `{∅}` is not tight, so `A ≠ {∅}` is harmless; `univ : Finset (Finset (Fin 3))` satisfies every hypothesis (not vacuous).
- **Every change is justified and it builds.** The binder split matches the sibling declarations, `research open` is rightly kept, and `lake --wfail build FormalConjectures.Wikipedia.UnionClosed` succeeds.

Optional nits only (not blocking): `cardinality_even_of_…` is a pre-existing misnomer (power of two).

---

### Changes requested

1. `FormalConjectures/Wikipedia/UnionClosed.lean:206-208` — the docstring cites Conjecture 3 whole, but the theorem formalizes only its first half (the "`A_F` contains `A_k` as a submatrix" clause is omitted). Replace the last two paragraphs with:
```
If the union-closed sets conjecture is tight for some family `A` then $|A| = 2^k$ for some $k$.
This is the first part of Conjecture 3 in the reference below; its second part (that the
family then contains the power set of a $k$-element set as a "submatrix") is not formalized here.

Reference: Conjecture 3 in https://www.nieuwarchief.nl/serie5/pdf/naw5-2023-24-4-225.pdf.
```

smmercuri:
🤖 *This comment was written by an automated agent (Claude, run by the PR author), not by a human.*

Applied the changes requested in https://github.com/google-deepmind/formal-conjectures/pull/5839#issuecomment-5654817933 in d00d2c0c:
1. Docstring of `cardinality_even_of_union_closed_tight` now states that only the first part of Conjecture 3 is formalized (the "power set as a submatrix" clause is not).

`lake --wfail build FormalConjectures.Wikipedia.UnionClosed` passes.

mo271 (approved):
Thanks, LGTM!

PR #5364, opened 2026-09-09
fix: lower density in Erdős 851

Sukumarsawant:
Fixes #4973

PR #5841, opened 2026-09-11
fix(ErdosProblems/851): bound the lower density instead of the density

### LINKED ISSUE #4973: Erdős 851: "the density ... is at least"
The statement at https://www.erdosproblems.com/851 says
> Let $\epsilon>0$. Is there some $r\ll_\epsilon 1$ such that the density of integers of the form $2^k+n$, where $k\geq 0$ and $n$ has at most $r$ prime divisors, is at least $1-\epsilon$?

It's not 100% clear what is meant, but I think "the density ... is at least" is meant to be a bound on the lower density.

### Suggested fix
Bound the lower density instead of the (natural) density.

### Additional context
It would be good to scan the remaining problems for the correct interpretations of "density is at least/most".

### Choose either option

smmercuri:
[erdosproblems.com/851](https://www.erdosproblems.com/851) asks whether, for every $\epsilon > 0$, there is some $r \ll_\epsilon 1$ such that the density of integers of the form $2^k + n$ ($n$ with at most $r$ prime divisors) is at least $1 - \epsilon$. The Lean statement `erdos_851` required the natural density to exist and be at least $1 - \epsilon$ (`∃ r d, (TwoPowAddSet r).HasDensity d ∧ 1 - ε ≤ d`), which is stronger than the source: "density is at least" is a bound on the lower density, and the existence of the natural density is not part of the problem or its proof.

**Change**: state the conclusion as `∃ r, 1 - ε ≤ (TwoPowAddSet r).lowerDensity`, matching the reading used for the same phrase elsewhere in the repository (Erdős 10, 26). Docstring, attributes and the Romanoff variant are unchanged.

**Checked**: `lake --wfail build 'FormalConjectures.ErdosProblems.«851»'` — Build completed successfully, no warnings.

Same fix as the open PR #5364.

Fixes #4973

### DIFF
--- a/FormalConjectures/ErdosProblems/851.lean
+++ b/FormalConjectures/ErdosProblems/851.lean
@@ -49,8 +49,8 @@ form $2^k+n$, where $k \geq 0$ and $n$ has at most $r$ prime divisors, is at lea
 This was proved affirmatively by Price and GPT-5.2 Pro [Pr26].
 -/
 @[category research solved, AMS 11]
-theorem erdos_851 (ε : ℝ) (hε : ε ∈ Set.Ioo 0 1) : ∃ r d,
-    (TwoPowAddSet r).HasDensity d ∧ 1 - ε ≤ d := by
+theorem erdos_851 (ε : ℝ) (hε : ε ∈ Set.Ioo 0 1) : ∃ r,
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #4890, opened 2026-08-12
feat(ErdosProblems/319): prove the linear upper bound

KitaKen1:
This PR changes `Erdos319.erdos_319.variants.isBigO` from `answer(sorry)` to
`answer(fun N : ℕ ↦ (N : ℝ))` and supplies a kernel-checked Lean proof.

It leaves `erdos_319`, `isTheta`, `isLittleO`, and `lb` unchanged.

The Lean formalization was prepared by [KitaKen1 (Kenta Kitamura)](https://github.com/KitaKen1).
The proof establishes the exact target theorem:

```lean
theorem Erdos319.erdos_319.variants.isBigO (_N : ℕ) (c : ℕ → ℝ)
    (h : ∀ N, IsGreatest
      { (#A : ℝ) | (A) (_ : A ⊆ Finset.Icc 1 N)
        (_ : ∃ δ : ℕ → ℤˣ, ∑ n ∈ A, (δ n : ℚ) / n = 0 ∧
          ∀ A' ⊂ A, A'.Nonempty →
            ∑ n ∈ A', (δ n : ℚ) / n ≠ 0) } (c N)) :
    c =O[atTop] (fun N : ℕ ↦ (N : ℝ))
```

Proof: https://github.com/KitaKen1/erdos-319-linear-upper-bound/blob/3e2cb405aa7c5e2adedcce2c804d704b62c7998a/lean/Erdos319LinearUpperBound.lean#L13-L31

Repository: https://github.com/KitaKen1/erdos-319-linear-upper-bound

Lean4Web: https://live.lean-lang.org/#url=https%3A%2F%2Fraw.githubusercontent.com%2FKitaKen1%2Ferdos-319-linear-upper-bound%2Frefs%2Fheads%2Fmain%2Flean4web%2FErdos319LinearUpperBoundLean4Web.lean

`#print axioms` reports only `propext`, `Classical.choice`, and `Quot.sound`;
there is no `sorryAx`.

AI Usage Disclosure: This formalization was assisted by OpenAI Codex, using GPT-5.6 sol.

**Note:** This answer may not match the intended question: it proves only the upper bound `c = O(N)`, whereas the intended asymptotic statement appears to require matching upper and lower bounds.

PR #5843, opened 2026-09-11
fix(ErdosProblems/319): remove the tautological isBigO and isLittleO variants

### LINKED ISSUE #4892: Erdős 319 `isBigO` variant admits a tautological answer
`FormalConjectures/ErdosProblems/319.lean` contains the variant `Erdos319.erdos_319.variants.isBigO`.

There is a gap between its docstring and its Lean statement.
The docstring asks for the simplest `g(N)` such that `c = O(g)`, while the Lean statement checks only that one supplied function is an upper bound.
It does not encode that the answer is the simplest or optimal upper bound.

### Trivial proof/disproof
Taking `g(N) = N` works immediately: every admissible set satisfies `A ⊆ Finset.Icc 1 N`, hence `A.card ≤ N`. The signed unit-fraction condition is not used.

PR: https://github.com/google-deepmind/formal-conjectures/pull/4890

Kernel-checked proof: https://github.com/KitaKen1/erdos-319-linear-upper-bound/blob/3e2cb405aa7c5e2adedcce2c804d704b62c7998a/lean/Erdos319LinearUpperBound.lean#L13-L31

### Suggested fix
Require a matching lower bound or another explicit optimality condition, or reclassify the variant so that an arbitrary upper bound is not treated as a solution.

### Additional context
This proves a valid elementary upper bound, but not Erdős Problem 319.
The same formulation concern was raised during review of PR #4890.

### Choose either option

smmercuri:
`erdos_319.variants.isBigO` asked for "the simplest $g(N)$ such that $c(N) = O(g(N))$", but the Lean statement only required the answer to be some upper bound of $c$. Since every admissible $A$ satisfies $A \subseteq \{1, \dots, N\}$, the answer $g(N) = N$ closes it without using the unit-fraction condition (see #4890). The source only asks for the size of the largest $A$; it does not ask for an $O$- or $o$-bound.

**Change**

Remove `erdos_319.variants.isBigO` and `erdos_319.variants.isLittleO`, and shorten the formalisation note accordingly. Adding a matching lower bound to `isBigO` makes it identical to the existing `erdos_319.variants.isTheta`, which already pins the growth rate up to constants. The `isLittleO` variant has the same defect ($c(N) = o(N^2)$ trivially) and admits no optimal answer, since $c(N) \neq o(N)$ by the Croot lower bound. `erdos_319`, `isTheta` and `lb` are unchanged.

**Checked**: `lake --wfail build FormalConjectures.ErdosProblems.«319»`

Fixes #4892

### DIFF
--- a/FormalConjectures/ErdosProblems/319.lean
+++ b/FormalConjectures/ErdosProblems/319.lean
@@ -47,11 +47,7 @@ theorem erdos_319 (N : ℕ) : IsGreatest
   sorry
 
 -- Formalisation note: it's possible that solution to `erdos_319` needs to be
--- expressed asymptotically. To handle this we include `IsTheta`, `IsBigO`
--- and `IsLittleO` variants below. Since a solution is not known this necessitates
--- the use of an `answer(sorry)` placeholder. Trivial or sub-optimal solutions
--- will therefore exist to the asymptotic formalisations. A true solution to
--- the asymptotic variants should have a degree of optimality or non-triviality to it.
+-- expressed asymptotically. To handle this we include an `IsTheta` variant below.
 /-- Let $c(N)$ be the size of the largest $A\subseteq\{1, \dots, N\}$ such that there is a function
 $\delta : A \to \{-1, 1\}$ such that
 $$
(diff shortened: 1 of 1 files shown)


mo271 (approved):
Thanks!

PR #3422, opened 2026-03-05
feat(ErdosProblems): All 1179 Erdős conjectures formalized

ryantuck:
## All 1179 Erdős conjectures are now formalized

I have utilized Claude (primarily Opus 4.6) to produce formal conjectures for **all 805 remaining Erdős problems**, completing the formalization of the entire set of 1179 conjectures, and achieving [Milestone 1: All open Erdős problems formalized](https://github.com/google-deepmind/formal-conjectures/milestone/1) in this repo. 

I believe all conjectures adhere to the style guidelines in this repo and `lake build` does pass, making this a valid and nontrivial contribution to this project.

After [contributing the conjecture for problem 13](https://github.com/google-deepmind/formal-conjectures/pull/1793) I realized that formalizing the few hundred remaining conjectures basically boiled down to running a for-loop (and enough tokens to pay for it), so set out to accomplish just that.

Note that this PR contains no _proofs_, only formalizations of the conjectures.

## Question for maintainers - how to proceed?

This is a massive PR generated exclusively by frontier AI and appears to be fairly new territory for this project. That being said, I've put significant effort into making it as ready as possible by focusing on strict adherence to style guidelines, believe it's worth the time to consider, and I'd be happy to assist in whatever way makes sense to do so.

**Maintainers** - Is this PR okay as-is? Should I open up 800+ PRs? Is there a more rigorous `PR_REVIEW.md` that Claude / Gemini / GPT can use to further review the conjectures to give the maintainers confidence that this isn't a deluge of AI slop? Does it make more sense to put that sort of proof-of-work in _prior_ to merging, or do these appear correct-enough and would it make more sense to continue to apply a similar batch processing on all 1179 conjectures each time a new frontier model drops? Should I just go kick rocks and maintain my own repo?

## Methodology

There were 374 formalized `.lean` conjectures in `FormalConjectures/ErdosProblems/` when I began, leaving 805 of the problems in the set remaining to work on.

I utilized Claude in four separate stages for each problem in the batch:

1. **Formalization** - formalize conjecture - fetch from erdosproblems.com, implement, ensure `lake build` passes (see [`FORMALIZE_CONJECTURE.md`](https://github.com/ryantuck/erdos-ai/blob/master/FORMALIZE_CONJECTURE.md))
[...]

PR #5876, opened 2026-09-12
feat(ErdosProblems): formalise Erdős Problem 1066

### LINKED ISSUE #1108: Erdős Problem 1066
### What is the conjecture

https://www.erdosproblems.com/1066

  Let $G$ be a graph given by $n$ points in $\mathbb{R}^2$, where any two distinct points are at least distance $1$ apart, and we draw an edge between two points if they are distance $1$ apart.

Let $g(n)$ be maximal such that any such graph always has an independent set on at least $g(n)$ vertices. Estimate $g(n)$, or perhaps $\lim \frac{g(n)}{n}$.

Status:  open

### Choose either option

stantheman0128:
## Summary
Statement-only Lean 4 formalisation of [Erdős Problem 1066](https://www.erdosproblems.com/1066).

Fixes #1108.

## Check
`lake --wfail build 'FormalConjectures.ErdosProblems.«1066»'`

### DIFF
--- a/FormalConjectures/ErdosProblems/1066.lean
+++ b/FormalConjectures/ErdosProblems/1066.lean
@@ -1,0 +1,129 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


stantheman0128:

PR #5927, opened 2026-09-12
feat(ErdosProblems): formalise Erdős Problem 1123

stantheman0128:
## Summary
Statement-only Lean 4 formalisation of [Erdős Problem 1123](https://www.erdosproblems.com/1123): the Boolean algebras of sets modulo density zero and modulo logarithmic density zero are not isomorphic.

The claim is independent of ZFC, so the headline uses `answer(sorry)` (as in Erdős Problem 1119). Just–Krawczyk under CH is recorded as a variant.

Integers are taken as `ℕ`, matching $\mathcal{P}(\omega)/I$. The quotients are the Boolean algebras of the Boolean rings `Set ℕ ⧸ I`.

## Check
`lake --wfail build 'FormalConjectures.ErdosProblems.«1123»'`

### DIFF
--- a/FormalConjectures/ErdosProblems/1123.lean
+++ b/FormalConjectures/ErdosProblems/1123.lean
@@ -1,0 +1,218 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


stantheman0128:

PR #5930, opened 2026-09-12
feat(ErdosProblems): formalise Erdős Problem 1042

stantheman0128:
## Summary
Statement-only Lean 4 formalisation of [Erdős Problem 1042](https://www.erdosproblems.com/1042): if $F\subset\mathbb{C}$ is closed of transfinite diameter $1$ and not contained in a closed disc of radius $1$, can $\{z:|f(z)|<1\}$ have $n$ connected components for monic $f$ with roots in $F$? If the transfinite diameter is $<1$, is the number of components at most $(1-c)n$?

Ghosh–Ramachandran [GhRa24] solved both parts, so the headlines use `answer(True)`. Transfinite diameter (logarithmic capacity) is not in Mathlib and is defined in the problem file via the usual $n$-point Vandermonde limit (`Filter.limsup`). Display math uses `$$ $$`.

## Check
`lake --wfail build 'FormalConjectures.ErdosProblems.«1042»'`

### DIFF
--- a/FormalConjectures/ErdosProblems/1042.lean
+++ b/FormalConjectures/ErdosProblems/1042.lean
@@ -1,0 +1,144 @@
+/-
+Copyright 2026 The Formal Conjectures Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


stantheman0128:

PR #5933, opened 2026-09-12
feat(ErdosProblems): formalise Erdős Problem 1166

stantheman0128:
## Summary
Statement-only Lean 4 formalisation of [Erdős Problem 1166](https://www.erdosproblems.com/1166).

The source asks whether a simple random walk on $\mathbb{Z}^2$ has only $(\log n)^{O(1)}$ distinct favourite values up to time $n$, almost surely for all but finitely many $n$. The problem is solved in the affirmative.

## Formalisation notes
- The walk is the simple symmetric random walk on $\mathbb{Z}^2$: i.i.d. uniform nearest-neighbour steps, realised as `Measure.infinitePi` of the uniform measure on `Fin 4`, with position given by partial sums.
- $(\log n)^{O(1)}$ is the existence of a deterministic $C>0$ such that the bound holds eventually, almost surely.
- The stronger bound $\ll (\log n)^2$ from the source is recorded as `erdos_1166.variants.log_sq`.

## Check
`lake --wfail build 'FormalConjectures.ErdosProblems.«1166»'`


stantheman0128:

PR #5951, opened 2026-09-13
chore: fix universe issue in poincare conjecture

felixpernegger:
I dont think it makes sense to restrict to `Type`. This is equivalent however, since connected compact manifolds cant have cardinality > c anyways (thus one can lift the manifolds between universes). So this is just a nicer version IMO.


mo271 (approved):
Thanks, LGTM!

PR #5938, opened 2026-09-12
feat(ErdosProblems): formalise Erdős Problem 1099

stantheman0128:
## Summary
Statement-only Lean 4 formalisation of [Erdős Problem 1099](https://www.erdosproblems.com/1099).

The boxed question is solved: Vose [Vo84] constructed a sequence with $h_\alpha$ bounded, so this is `research solved` with `answer(True)`. Still open: boundedness along $n!$ and $\mathrm{lcm}[1,n]$.

## Formalisation notes
- Consecutive divisor ratios live in ForMathlib as `Nat.consecutiveDivisorRatio`, next to the existing `Nat.nth (· ∈ n.divisors)` enumeration. $h_\alpha$ is a Finset sum over those ratios.
- `liminf h_α ≪_α 1` is encoded as `∃ C, liminf ≤ C` (a bounded subsequence, not a uniform bound on all $n$).
- $h_\alpha(1)=0$ (empty sum).

## Check
`lake --wfail build FormalConjecturesForMathlib.NumberTheory.Divisors`
`lake --wfail build 'FormalConjectures.ErdosProblems.«1099»'`

## AI / harness
- **Model**: Grok 4.6 via **Grok Build** CLI
- **Review**: math + style pass against erdosproblems.com/1099 (LaTeX source) and repo `AGENTS.md` / `QUALITY_PIPELINE`

Not on the #5495 list.

PR #5957, opened 2026-09-13
feat(ErdosProblems): formalise Erdős Problem 1182

stantheman0128:
## Summary
Statement-only Lean 4 formalisation of [Erdős Problem 1182](https://www.erdosproblems.com/1182): estimate $f(n)$ and $F(n)$ for connected graphs with $R(K_3,G)=2n-1$, and whether $F(n)/n\to\infty$.

Ramsey numbers are written with `SimpleGraph.graphRamsey` (the repository API; there is no `ramseyNumber`). $F(n)$ is capped at $\binom{n}{2}$ so the `sSup` on `ℕ` stays well-defined when every connected $n$-vertex graph has this Ramsey number. The particular question $F(n)/n\to\infty$ is recorded as solved in the negative, following Brandt's bound $F(n)\le 84n$.

## Check
`lake --wfail build 'FormalConjectures.ErdosProblems.«1182»'`


stantheman0128:

PR #5960, opened 2026-09-13
feat(ErdosProblems): formalise Erdős Problem 984

stantheman0128:
## Summary
Statement-only Lean 4 formalisation of [Erdős Problem 984](https://www.erdosproblems.com/984).

Can $\mathbb{N}$ be $2$-coloured so that every monochromatic $k$-term arithmetic progression $\{a,a+d,\ldots,a+(k-1)d\}$ satisfies $k\ll_\varepsilon a^\varepsilon$ for all $\varepsilon>0$? This is recorded as solved (`answer(True)`): Zach Hunter proved that the answer is yes.

- `erdos_984` (solved): a $2$-colouring with $k\ll_\varepsilon a^\varepsilon$
- `erdos_984.variants.spencer` (solved): Spencer's $3$-colouring with a bound growing more slowly than every positive power
- `erdos_984.variants.erdos` (solved): Erdős's $2$-colouring with $k\ll a^{1-c}$ for some $c>0$

The first term $a$ and difference $d$ are taken to be positive, so that the progression has $k$ distinct terms and $a^\varepsilon$ is positive. Display math uses `$$ $$`.

## Check
`lake --wfail build 'FormalConjectures.ErdosProblems.«984»'`


stantheman0128:

PR #5961, opened 2026-09-13
feat(ErdosProblems): formalise Erdős Problem 947

stantheman0128:
## Summary
Statement-only Lean 4 formalisation of [Erdős Problem 947](https://www.erdosproblems.com/947): non-existence of an exact covering system with distinct moduli.

## Check
`lake --wfail build 'FormalConjectures.ErdosProblems.«947»'`


stantheman0128:

PR #5963, opened 2026-09-13
feat(ErdosProblems): formalise Erdős Problem 989

stantheman0128:
## Summary
Statement-only Lean 4 formalisation of [Erdős Problem 989](https://www.erdosproblems.com/989): disk discrepancy of an infinite discrete point set in $\mathbb{R}^2$.

Beck [Be87] proved $f(r)\gg r^{1/2}$ for every such set, and constructed a set with $f(r)\ll (r\log r)^{1/2}$.

## Formalisation notes
- The comparison of $|A\cap C|$ with the area $\pi r^2$ means $C$ is a closed disk of radius $r$, not a circumference.
- $A$ is required to be discrete (locally finite) so that $A\cap C$ is finite; otherwise the count is not a natural number.
- Unboundedness and the lower bound quantify over centres rather than using `sSup` on $\mathbb{R}$, which is $0$ on unbounded sets.

## Check
`lake --wfail build 'FormalConjectures.ErdosProblems.«989»'`


stantheman0128:

PR #5966, opened 2026-09-13
feat(ErdosProblems): formalise Erdős Problem 925

stantheman0128:
## Summary
Statement-only Lean 4 formalisation of [Erdős Problem 925](https://www.erdosproblems.com/925).

Is there a constant $\delta>0$ such that, for all large $n$, if $G$ is a graph on $n$ vertices which is not Ramsey for $K_3$ then $G$ contains an independent set of size $\gg n^{1/3+\delta}$? This is recorded as solved (`answer(False)`): Alon and Rödl [AlRo05] disproved the claim.

- `IsRamseyForK3`: every 2-edge-colouring of $G$ has a monochromatic triangle
- `erdos_925` (solved): no such $\delta>0$ exists
- `erdos_925.variants.cube_root` (solved): the easy bound $\alpha(G)\gg n^{1/3}$

A graph is Ramsey for $K_3$ in the sense of the source: there is no 2-colouring of its edges without a monochromatic triangle. The implied constant in $\gg$ is independent of $n$ and $G$. Display math uses `$$ $$`.

## Check
`lake --wfail build 'FormalConjectures.ErdosProblems.«925»'`


stantheman0128:

PR #5969, opened 2026-09-13
feat(ErdosProblems): formalise Erdős Problem 910

stantheman0128:
## Summary
Statement-only Lean 4 formalisation of [Erdős Problem 910](https://www.erdosproblems.com/910).

The boxed questions (topology, AMS 54) are:

- `erdos_910.parts.i` (disproved): does every connected set in $\mathbb{R}^n$ contain a connected subset which is not a point and not homeomorphic to the original set?
- `erdos_910.parts.ii` (disproved): if $n\geq 2$, does every connected set in $\mathbb{R}^n$ contain more than $2^{\aleph_0}$ many connected subsets?

Asked by Erdős in the 1940s, who thought the answer to both questions is yes. The answer to both is in fact no, as shown by Rudin [Ru58] (conditional on the continuum hypothesis). Recorded as `answer(False)`.

## Formalisation choices
- Connected sets live in `ℝ^n` as `EuclideanSpace ℝ (Fin n)`.
- Both statements restrict to `s.Nontrivial`, so a singleton is not a degenerate counterexample.
- “Not a point” is `t.Nontrivial`. Homeomorphism of subspaces is `IsEmpty (t ≃ₜ s)`.
- Part (ii) compares `Cardinal.mk` of `{t | t ⊆ s ∧ IsConnected t}` with `𝔠`.
- There is no GitHub issue titled Erdős Problem 910.

## Check
```bash
lake --wfail build 'FormalConjectures.ErdosProblems.«910»'
```


stantheman0128:

PR #5975, opened 2026-09-13
feat(ErdosProblems): formalise Erdős Problem 909

stantheman0128:
## Summary
Statement-only Lean 4 formalisation of [Erdős Problem 909](https://www.erdosproblems.com/909).

The boxed question (topology, AMS 54) is: let $n\geq 2$. Is there a space $S$ of dimension $n$ such that $S^2$ also has dimension $n$?

This has been solved in the affirmative by Anderson and Keisler [AnKe67]. Recorded as `answer(True)`.

## Formalisation choices
- Dimension is Mathlib `smallInductiveDimension` (small inductive / inductive topological dimension, as in [AnKe67]). This agrees with covering dimension on the separable metric examples they construct.
- $S^2$ is the product space `S × S`.
- The quantifier is `∀ n ≥ 2, ∃ S, …`, matching “let $n\geq 2$” and the remark that the result was proved for general $n$.
- There is no GitHub issue titled Erdős Problem 909.

## Check
```bash
lake --wfail build 'FormalConjectures.ErdosProblems.«909»'
```


stantheman0128:

Issue #5990, opened 2026-09-13
ErdosProblems/283: `formal_proof` link still anchors the `ℤ[X]` wrapper after #5812


zblore:
#5812 (merged 2026-09-12) changed `erdos_283` to quantify over integer-valued `p : ℚ[X]`, matching the source. The `formal_proof` link added in #4246 still points at

```lean
theorem erdos_283 : True ↔ ∀ p : ℤ[X], Condition p
```

in `Shashi456/erdos-formalizations`, with `Condition` defined on `ℤ[X]`. The PR body of #5812 notes that the linked development's core result `PolynomialEgyptianSums.theorem_1` is stated for integer-valued `ℚ[X]` and that only the wrapper is `ℤ[X]`. That note lives in the PR, not in the file.

As it stands, the attribute on a `ℚ[X]` statement anchors a `ℤ[X]` proof. Two options: anchor `theorem_1` (or a `ℚ[X]` wrapper if the proof author adds one), or record the gap in the docstring until the wrapper is updated.

Found by the same statement comparison as #5989.

Issue #5989, opened 2026-09-13
OEIS/211417: `formal_proof` anchor points at the form without `0 < D`


zblore:
`OEIS/211417.lean`, `general_divisibility`:

```lean
theorem general_divisibility (r : ℕ) (hr : 1 ≤ r) :
    ∃ D : ℤ, 0 < D ∧ ∀ n : ℕ, (divisorProduct n r) ∣ (D * (a n : ℤ))
```

The `formal_proof` attribute anchors line 651 of the linked file. The declaration at that line is

```lean
theorem general_divisibility (r : ℕ) (hr : 1 ≤ r) :
    ∃ D : ℤ, ∀ n : ℕ, (divisorProduct n r) ∣ (D * (a n : ℤ))
```

without `0 < D`. The same file proves `general_divisibility_is_vacuous` with `D = 0`, and separately proves the statement that matches main as `general_divisibility_strong` (witness `Dwit r = (r!)^(r*r)` with `Dwit_pos`).

The docstring in main says the permalink "proves the non-vacuous form". The mathematics is there; the anchor is on the weak form.

Suggested fix: point the link at `general_divisibility_strong`. Happy to open the PR if that is the preferred route.

Found by fetching every `formal_proof` target and comparing the statement at the anchor with the statement in main (script to follow as a PR).

mo271:
thanks!

Sounds indeed like pointing the link to `general_divisibilty_strong` is the way to go here, pelase go ahead creating that pr!

Issue #5988, opened 2026-09-13
Seven `formal_proof` links return 404


zblore:
Every `formal_proof` link in `FormalConjectures/` was fetched on 2026-09-13 (commit 18e4097). 487 links; 7 return HTTP 404. Four are attached to `research solved` statements, two to `test`, one to `textbook`. None of the seven has a reachable proof.

| File | Declaration | Category | Link |
|---|---|---|---|
| `ErdosProblems/1043.lean` | `erdos_1043` | research solved | https://github.com/XC0R/formal-conjectures/blob/7db17471701f15b125d1c36bc1fa5bb9b702d6be/FormalConjectures/ErdosProblems/1043.lean#L214 |
| `ErdosProblems/138.lean` | `monoAPNumber_two_one` | test | https://github.com/XC0R/formal-conjectures/blob/6c7a16e8998d1c597fa2a5c6329bc9301fcc56e2/FormalConjectures/ErdosProblems/138.lean#L79 |
| `ErdosProblems/138.lean` | `monoAPNumber_two_two` | test | https://github.com/XC0R/formal-conjectures/blob/6c7a16e8998d1c597fa2a5c6329bc9301fcc56e2/FormalConjectures/ErdosProblems/138.lean#L142 |
| `ErdosProblems/154.lean` | `erdos_154` | research solved | https://github.com/willblair0708/lean-proofs/blob/main/ErdosProblems/Erdos154Sumset.lean |
| `ErdosProblems/198.lean` | `erdos_198` | research solved | https://github.com/XC0R/formal-conjectures/blob/33045ec97b08a40c7ff91f1e2a112f5e3b4725f8/FormalConjectures/ErdosProblems/198.lean#L168 |
| `ErdosProblems/379.lean` | `erdos_379` | research solved | https://github.com/XC0R/formal-conjectures/blob/80a965e9a85d3f3dabd0a398a49adab6742ea6e0/FormalConjectures/ErdosProblems/379.lean#L123 |
| `ErdosProblems/828.lean` | `erdos_828.variants.phi_dvd_self_iff_pow2_pow3` | textbook | https://github.com/XC0R/formal-conjectures/blob/03e00cf8d44098d0fb06e891fca30c29769df619/FormalConjectures/ErdosProblems/828.lean#L49 |

Six point at commits in `XC0R/formal-conjectures` that no longer exist. One points at `willblair0708/lean-proofs`, which returns 404 at the repository level.

[...]

PR #5995, opened 2026-09-14
chore(scripts): check that every formal_proof link resolves

zblore:
`formal_proof` links are the only evidence behind `research solved` statements proved outside this repository, and nothing followed them. This adds `scripts/check_formal_proof_links.py`, which does.

For each link it:

1. fetches the target and reports HTTP errors;
2. if the link has a `#L` anchor, checks that a `theorem` or `lemma` starts near it;
3. if it has no anchor and is a `formal_conjectures` link, checks the annotated name exists in the target (a fork keeps our names; other kinds are not checked for this);
4. with `--compare`, reports statement differences between here and the target, as advice. It is textual and over-reports, so it never fails the run.

On `main` today: 487 links, 7 unreachable (#5988) and 3 stale anchors (`ErdosProblems/1082` part ii, `ErdosProblems/349` `exists_finsetSum_two_pow`, `WrittenOnTheWallII/GraphConjecture143`). `--compare` also surfaces #5989 and #5990.

Includes unit tests (no network) in the style of `test_check_erdos_status.py`, and a weekly workflow with `workflow_dispatch`. The workflow only reports; it does not open issues.

**Checked**: `python -m unittest discover -s scripts -p 'test_*.py'` passes; the script runs against `main` in about 15 s.

Closes nothing on its own; fixes for #5988–#5990 belong to the proof authors.


zblore:
Ok i've now consented to CLA.

felixpernegger:
good idea

zblore:
Friendly ping @mo271 on this one when you have a moment. CI is green and there are no conflicts.

PR #4657, opened 2026-07-28
disprove(Other): convex additive-VC₂ bound in ℝ³

KitaKen1:
This PR replaces `VCDimConvex.hasAddVCNDimAtMost_two_one_of_convex_r3` with its negation and links a kernel-checked Lean proof.

The Lean proof was prepared by [KitaKen1 (Kenta Kitamura)](https://github.com/KitaKen1).

The external proof establishes:

```lean
VCDimConvex.not_hasAddVCNDimAtMost_two_one_of_convex_r3 :
  ¬ (∀ {C : Set ℝ³} (hC : Convex ℝ C),
    HasAddVCNDimAtMost C 2 1)
```

Proof: https://github.com/KitaKen1/vcdim-convex-counterexample/blob/ad7ffff1514843c886633f5408c6456dfb2e2a49/formal-conjectures-v4.27.0/VCDimConvexCounterexample.lean

Repository: https://github.com/KitaKen1/vcdim-convex-counterexample

Lean4Web: https://live.lean-lang.org/#url=https%3A%2F%2Fraw.githubusercontent.com%2FKitaKen1%2Fvcdim-convex-counterexample%2Fmain%2Flean4web-v4.33.0-rc1%2FVCDimConvexCounterexampleLean4Web.lean

`#print axioms` reports only `propext`, `Classical.choice`, and `Quot.sound`; there is no `sorryAx`.

Note: A prior counterexample was posted in [PR #88](https://github.com/DomTheDeveloper/formal-conjectures/pull/88), but its Lean CI did not pass and the draft PR was closed without being merged.

PR #6018, opened 2026-09-15
feat(Other/VCDimConvex): make the universal statement interesting again

YaelDillies:
The universal upper bound of 1 was disproved in #4657, so conjecture a universal bound of 2 instead.


mo271 (approved):
Thanks!

If you don't hink that the n=1 is completely boring we could keep it and just add another one in addition to it

PR #6020, opened 2026-09-15
feat(ErdosProblems): formalize problem 1025 square-root bounds

xukp20:
This PR formalizes the known square-root lower and upper bounds for Erdős problem 1025.
The file defines distinct unordered pairs, admissible pair mappings, and free sets. It states two bounds: every admissible mapping has a free set of size at least a uniform positive multiple of √n, and for every sufficiently large n there exists an admissible mapping whose free sets all have size at most a uniform multiple of √n. The unordered-pair representation follows the set-mapping formulation of Conlon, Fox, and Sudakov.
The complete proofs remain in the external [Erdos1025 repository](https://github.com/iiis-lean/Erdos1025/tree/30a98247750d691ba76dda074b80b7465a229036). Each statement has a formal_proof attribute linking to its corresponding proved declaration at a fixed commit. This contribution formalizes established results; it does not claim a new mathematical solution.
Validation:
- The new module builds with lake --wfail build 'FormalConjectures.ErdosProblems.«1025»' on Lean/Mathlib 4.33.1.
- A separate check on the external proof’s pinned 4.32.0 toolchain proves both proposed statements directly using the existing root theorems.
- Both root theorems depend only on propext, Classical.choice, and Quot.sound.
- The file includes a proved empty-set sanity check.
AI disclosure: the external Lean development used Lean Constellation with Codex and Grok. A Sol agent assisted with the declaration and semantic audit, and Codex prepared this statement adaptation and its validation. These checks do not constitute independent human expert review.


mo271:
Thanks for the contribution!

merging #6180 first since its definitions look better. Perhaps you can amend your proof to provide a `formal_proof using formal_conjectures` after rebasing when that pr is merged.

PR #6027, opened 2026-09-15
feat(Paper): add the Dichotomy Theorem for finite simple groups

henrykmichalewski:
## What

Adds `FormalConjectures/Paper/DichotomyTheorem.lean`, stating the Dichotomy Theorem for finite
simple groups: a finite simple group in which some 2-subgroup cannot be generated by four
elements is of component type or of characteristic 2 type. The groups excluded by that
hypothesis are exactly those classified by the Gorenstein–Harada theorem.

## Why

The Dichotomy Theorem is one of the pillars of the classification of the finite simple groups:
it splits the simple groups into the two families that the classification then treats by
different methods. Several statements already in this repository depend on the classification,
so having its main structural step available seems useful.

It is also one of the few results in this area that can be stated without enumerating simple
groups. The classification statement of Gorenstein–Harada cannot: its content *is* the list, and
the list ranges over families of groups of Lie type and several sporadic groups, none of which
Mathlib has. The dichotomy avoids this because both alternatives are intrinsic properties of the
group. Being of component type is a condition on centralizers of involutions, and being of
characteristic 2 type is a condition on 2-local subgroups.

## Formalisation choices

- The hypothesis follows the title of Gorenstein–Harada, "2-subgroups are generated by at most
  4 elements", rather than sectional 2-rank. The two are equivalent, and this phrasing needs no
  theory of sections.
- `core` is the join of the normal subgroups satisfying a predicate. `pCore` and `oddCore` are
  its instances, giving `O_p(G)` and `O(G)`. For a finite group these are the largest normal
  `p`-subgroup and the largest normal subgroup of odd order; the definitions are stated for any
  group and the theorem assumes `Finite G`.
- A 2-component is a subnormal subgroup `K` with `K / O(K)` quasisimple, following
  Gorenstein–Lyons–Solomon. `Subgroup.IsSubnormal` and `Group.IsPerfect` come from Mathlib.
- The theorem is tagged `@[category research solved]`: it is established, not open. If solved
  results of this kind are out of scope here, I am happy to close this.

## Checks

- `lake --wfail build FormalConjectures.Paper.DichotomyTheorem` passes.
- The problem theorem has one `category` tag and an AMS tag; the two helper statements are
  tagged `API` and `test` and are proved.
- Module and theorem docstrings cite the sources.

https://claude.ai/code/session_019NUDgLUBdCBW6wYY47jXP8


FormalConjectures/Paper/DichotomyTheorem.lean
```diff
@@ -0,0 +1,121 @@
...
+  exact ⟨∅, le_rfl, Subsingleton.elim _ _⟩
+
+/--
+**The Dichotomy Theorem.** Let $G$ be a finite simple group in which some $2$-subgroup cannot be
+generated by four elements. Then $G$ is of component type or of characteristic $2$ type.
+-/
+@[category research solved, AMS 20]
+theorem isComponentType_or_isCharacteristicTwoType [Finite G] (hG : IsSimpleGroup G)
+    (h : ¬TwoSubgroupsGeneratedBy G 4) :
+    IsComponentType G ∨ IsCharacteristicTwoType G := by
```
felixpernegger, line 118: This is not an open problem (and presumably solved since the 90s). So maybe not really appropriate for this repo

felixpernegger:
Maybe this would be good for lean-eval though? Since it apparently forces you to do the classification

PR #6045, opened 2026-09-18
Prove erdos_23.variants.n1_tight

### LINKED ISSUE #6044: Erdős 23: prove the test n1_tight
`FormalConjectures/ErdosProblems/23.lean` has the `test` declaration `erdos_23.variants.n1_tight` as `by sorry`: there is a triangle-free graph on 5 vertices from which at least one edge must be removed to make it bipartite.

Plan: take `cycleGraph 5`. It is triangle-free by `decide +kernel`. If a bipartite `H ≤ cycleGraph 5` misses no edge, then `H = cycleGraph 5`, which is not bipartite since `chromaticNumber_cycleGraph_of_odd` gives chromatic number `3`. A pull request follows.

Konamiu:
Fixes #6044

## What changed

`FormalConjectures/ErdosProblems/23.lean`: replace the `sorry` in the `test` declaration `erdos_23.variants.n1_tight` with a 13-line proof.

## Why

The witness is `cycleGraph 5`. Triangle-freeness is a finite check (`decide +kernel` over the 32 subsets of `Fin 5`). If a bipartite `H ≤ cycleGraph 5` misses no edge, then `cycleGraph 5 ≤ H`, so `H = cycleGraph 5`; but `chromaticNumber_cycleGraph_of_odd` gives chromatic number `3`, contradicting `Colorable 2`.

## How checked

- `lake --wfail build 'FormalConjectures.ErdosProblems.«23»'` succeeds.
- `#print axioms Erdos23.erdos_23.variants.n1_tight`: `[propext, Classical.choice, Quot.sound]`.

## Reviewer notes

- The statement is elaborated under `open scoped Classical`, so `edgeFinset` for the witness carries the classical `DecidableRel` instance while `cycleGraph 5` also has a computable one. The proof converts the edge-subset hypothesis to `edgeSet` membership with `simp only [mem_edgeFinset]` to avoid the instance mismatch.
- `erdos_23.variants.n1` and `blowupC5_tight` in the same file are not touched.


mo271 (approved):
Thanks, LGTM!

PR #6055, opened 2026-09-18
Prove erdos_423.test.unbounded_iff_infinite_complement

### LINKED ISSUE #6054: Erdős 423: prove the test unbounded_iff_infinite_complement
`FormalConjectures/ErdosProblems/423.lean` has the `test` declaration `erdos_423.test.unbounded_iff_infinite_complement` as `by sorry`: for Hofstadter sequences, `a n - n → ∞` is equivalent to the sequence missing infinitely many integers.

Plan. A Hofstadter sequence is strictly increasing from index 1, so `n ≤ a n`. (→) If `a n ≥ N + 2 + n`, the `n` values `a 0, …, a (n - 1)` cannot cover the `n + 1` integers of `[N + 1, a n - 1]`, and later terms exceed `a n`; so some integer above `N` is missed, for every `N`. (←) Take `M + 1` missed integers; for `n` beyond all of them, `a 1, …, a n` and the positive missed integers are distinct elements of `[1, a n]`, giving `a n ≥ n + M`. The monotonicity facts go in as `API` lemmas on `IsHofstadterSeq`. A pull request follows.

Konamiu:
Fixes #6054

## What changed

`FormalConjectures/ErdosProblems/423.lean`:

- Five `API` lemmas on `IsHofstadterSeq`: `strictMono` (from index 1), `le_apply` (`n ≤ a n`), `lt_iff_lt`, `infinite_compl_of_unbounded`, `unbounded_of_infinite_compl`.
- The `test` `erdos_423.test.unbounded_iff_infinite_complement` is now the two-line combination of the last two lemmas.

## Why

Both directions are counting arguments on `[1, a n]`:

- (→) With `a n ≥ N + 2 + n`, the image of `range n` under `a` has at most `n` elements while `[N + 1, a n - 1]` has at least `n + 1`, so `Finset.exists_mem_notMem_of_card_lt_card` gives an integer in that interval outside the image; it is not `a k` for `k ≥ n` either, since those are at least `a n`.
- (←) From `Set.Infinite.exists_subset_card_eq` take `M + 1` missed integers with maximum `X`. For `n > X`, the sets `(Icc 1 n).image a` (of size `n`, by injectivity) and the positive missed integers (at least `M` of them) are disjoint subsets of `Icc 1 (a n)`.

## How checked

- `lake --wfail build 'FormalConjectures.ErdosProblems.«423»'` succeeds.
- `#print axioms Erdos423.erdos_423.test.unbounded_iff_infinite_complement`: `[propext, Classical.choice, Quot.sound]`.

## Reviewer notes

- `IsHofstadterSeq` does not constrain `a 0`, so the lemmas only use monotonicity from index 1 and treat `a 0` as an arbitrary extra value where it matters.
- The `research solved` variants `unbounded` and `infinite_complement` themselves are not touched.


mo271 (approved):
Thanks, LGTM!

PR #6061, opened 2026-09-18
Green 14: link formal proofs of W_3_3, W_3_4 and W_3_5

### LINKED ISSUE #6060: Green's problem 14: prove W_3_3 (W(3,3) = 9)
`FormalConjectures/GreensOpenProblems/14.lean` states the exact values `W 3 r` for `r = 3, …, 8` from [AKS14] as `research solved` with `by sorry`. The smallest, `W 3 3 = 9`, is a finite check: every `2`-colouring of `{1, …, 9}` has a monochromatic `3`-term progression, and `{1, …, 8}` has a colouring without one.

The obstacle is that `mixedMonoAPGuaranteeSet` uses `Set.IsAPOfLength`, which quantifies over unbounded `a d : ℕ` and an `ENat.card`, so it is not decidable. Plan: prove that for `k ≥ 2` a finite subset of `{1, …, N}` is a `k`-term progression iff it equals `{a + i d | i < k}` with `a ≥ 1`, `d ≥ 1`, `a + (k - 1) d ≤ N`; define a decidable predicate `HasMonoAP` with bounded quantifiers and prove `mem_mixedMonoAPGuaranteeSet_iff`. Then `W 3 3 = 9` follows by `decide +kernel` (512 colourings). The larger values need `2^18` and more colourings and hit the kernel's recursion limit, so they are left alone. A pull request follows.

Konamiu:
Fixes #6060

## What changed

`FormalConjectures/GreensOpenProblems/14.lean`: add `formal_proof using formal_conjectures at …` to `W_3_3`, `W_3_4` and `W_3_5`, pointing at [Konamiu/formal-conjectures@d6a68af](https://github.com/Konamiu/formal-conjectures/blob/d6a68af97f6ca7856d892569d88ab1ecc8e927bf/FormalConjectures/GreensOpenProblems/14.lean). The statements stay `by sorry`.

## The linked proofs

On that commit the file contains:

- `isAPOfLength_iff`: for `k ≥ 2`, a finite subset of `{1, …, N}` is a `k`-term progression iff it is `{a + i d | i < k}` with `a ≥ 1`, `d ≥ 1`, `a + (k - 1) d ≤ N`;
- `HasMonoAP` (decidable, bounded quantifiers) and `mem_mixedMonoAPGuaranteeSet_iff`;
- `W_3_3 : W 3 3 = 9` by `IsLeast.csInf_eq` and kernel `decide` (512 colourings);
- `W_3_4 : W 3 4 = 18` and `W_3_5 : W 3 5 = 22` the same way, but with `native_decide`, since a kernel `decide` on `2^18` colourings fails with "deep recursion detected".

`#print axioms` on that commit: `W_3_3` uses only `[propext, Classical.choice, Quot.sound]`; `W_3_4` and `W_3_5` additionally use the `native_decide` axioms. The file builds there with `lake build` (about 7 minutes, almost all in the native checks).

## Not included

`W_3_6` (`2^32` colourings) and beyond are out of reach of `native_decide` by brute force; a certificate-based approach would be needed.

## How checked

`lake --wfail build 'FormalConjectures.GreensOpenProblems.«14»'` succeeds on this branch.


mo271 (commented):
This proofs seems a little long, perhaps we could use the `formal_proof using formal_conjectures at [commit]` mechanism instead? In that case it might be good to include all theorems where it is possible to find a proof in this file while at it

Konamiu:
Restructured as suggested: this PR now only adds `formal_proof using formal_conjectures` links for `W_3_3`, `W_3_4` and `W_3_5`, with the statements left as `by sorry`. The proofs are on [Konamiu/formal-conjectures@d6a68af](https://github.com/Konamiu/formal-conjectures/blob/d6a68af97f6ca7856d892569d88ab1ecc8e927bf/FormalConjectures/GreensOpenProblems/14.lean). `W_3_3` is a kernel `decide`; `W_3_4` and `W_3_5` need `native_decide` (the kernel hits its recursion limit at `2^18` colourings), and their axiom lists show that. `W_3_6` and up are beyond brute force.

mo271 (approved):
Thanks, LGTM!

PR #6074, opened 2026-09-18
Erdős 486: the answer is no (Wang 2026), link the formal disproofs

### LINKED ISSUE #6073: Erdős 486: the answer is no (Wang 2026, formal proof in plby/lean-proofs)
`erdos_486` is `research open` with `answer(sorry)`, but the problem has been disproved.

Shouqiao Wang (with GPT-5.6) constructed a congruence system whose survivor set has lower logarithmic density ≤ 177/200 and upper logarithmic density ≥ 49/50: paper https://github.com/ShouqiaoW/erdos/blob/main/486/paper.pdf, proof claim https://www.erdosproblems.com/forum/thread/486/proof-claims. Wang's Lean proof (MIT) is ported in plby/lean-proofs as `Erdos486.not_erdos_486` (https://github.com/plby/lean-proofs/blob/8822f7ddef30fadbd92e1c6ab4ed897af356af5e/src/latest/ErdosProblems/Erdos486.lean#L20), Mathlib-only, standard axioms (rebuilt against this repo's Mathlib).

That statement uses a real-cutoff normalisation `(∑_{m < x, m ∈ B} 1/m) / log x` with moduli restricted to an explicit `A`, while ours uses `Set.HasLogDensity` (integer cutoff, all positive moduli, `X n = ∅` encoding). The two are equivalent: I have a ~120-line bridge deriving our statement from theirs, so `erdos_486` should become `answer(False)` / `research solved` with `formal_proof` links.

Konamiu:
Fixes #6073.

### What changed
`erdos_486`: `answer(sorry)` → `answer(False)`, `research open` → `research solved`, reference [Wa26], and two `formal_proof` links. Statement unchanged, still `by sorry`.

### Why
Wang (with GPT-5.6) disproved the problem; the Lean proof is complete and ported in plby/lean-proofs (claim on erdosproblems.com is listed as under review).

### The two links
1. `formal_proof using lean4` → plby `Erdos486.not_erdos_486`: real-cutoff normalisation `(∑_{m<x, m∈B} 1/m) / log x`, moduli restricted to an explicit `A` with `0 ∉ A`.
2. `formal_proof using formal_conjectures` → [Konamiu/formal-conjectures@c69df05](https://github.com/Konamiu/formal-conjectures/blob/c69df0584ca9767090f5a68c8f09f1ff3c93ab80/FormalConjectures/ErdosProblems/486.lean#L39), where `erdos_486` is proved outright: the branch contains Wang's proof (MIT, licence and attribution kept, only import paths changed) under `FormalConjectures/ErdosProblems/486/`, plus [`Bridge.lean`](https://github.com/Konamiu/formal-conjectures/blob/c69df0584ca9767090f5a68c8f09f1ff3c93ab80/FormalConjectures/ErdosProblems/486/Bridge.lean) which shows `Set.HasLogDensity B d → Erdos486.HasLogDensity B d` (the real-cutoff sum at `x` equals the integer-cutoff sum at `⌈x⌉₊ - 1`, and `log(⌈x⌉₊ - 1) / log x → 1`) and matches the survivor sets after extending `X` by `∅` outside `A`.

### How checked
- Wang's 27 files rebuilt against this repo's Mathlib; `#print axioms Erdos486.erdos_486` on the proof branch → `[propext, Classical.choice, Quot.sound]`.
- `lake --wfail build FormalConjectures.ErdosProblems.«486»` passes on this PR.


Konamiu:
Note on status: erdosproblems.com (and teorth's `problems.yaml`, which `scripts/check_erdos_status.py` compares against) still records this problem as `open` while the proof claim is under human review, so after merging the hourly status check would report `repo=formally solved, erdosproblems.com=open`. If you would rather keep `research open` until the site updates, I am happy to restructure this PR the way the site handles "open (Lean)": leave the main statement untouched and add a `research solved` variant stating exactly the linked theorem with the `formal_proof` link (plus a docstring pointer). Just say which you prefer.

mo271:
> Note on status: erdosproblems.com (and teorth's `problems.yaml`, which `scripts/check_erdos_status.py` compares against) still records this problem as `open` while the proof claim is under human review, so after merging the hourly status check would report `repo=formally solved, erdosproblems.com=open`. If you would rather keep `research open` until the site updates, I am happy to restructure this PR the way the site handles "open (Lean)": leave the main statement untouched and add a `research solved` variant stating exactly the linked theorem with the `formal_proof` link (plus a docstring pointer). Just say which you prefer.

I think it is fine. 
If it turns out to be a misformalisation (which is doens't look like) we can amend the problem/or add a variant

mo271 (approved):
Thanks, LGTM!

especially nice to also have a `formal_proof using formal_conjectures`

PR #6138, opened 2026-09-18
chore: require module system

felixpernegger:
This PR does two things:

1.  modulizes 5 more files that got overlooked in the last PR (due to being created later)

2. enables `requiresModuleSystem = true` in lakefile.toml such that this won't happen again, i.e. `lake build` throws a warning if one includes a non-module file. (in particular, CI will fail)


felixpernegger:
I have a few more comments at [#Formal conjectures > Modulization @ 💬](https://leanprover.zulipchat.com/#narrow/channel/524981-Formal-conjectures/topic/Modulization/near/625318833) what I think would also be very useful for build times

mo271 (approved):

mo271:
@felixpernegger the failure is the `All.lean` file generated in the CI. We generate this file in order to detect name clashes. Seems like an easy fix would be to add `module` to that file in the script that generates it?

PR #6156, opened 2026-09-18
Restore native_decide proofs dropped during modulization

mo271:
The modulization in #4688 replaced 27 proofs that used `native_decide` or `decide +native` with `sorry`, because under the module system a plain `public import FormalConjecturesUtil` no longer loads the compiled IR of Mathlib declarations into the elaboration environment. The interpreter then fails with e.g. "Could not find native implementation of external declaration 'compositionEquiv'" or "IR of declaration ... not available".

Adding `meta import FormalConjecturesUtil` next to the public import makes the IR available again, so the original proofs work unchanged. This restores all 27 proofs in

- FormalConjectures/ErdosProblems/340.lean
- FormalConjectures/GreensOpenProblems/39.lean
- FormalConjectures/OEIS/111291.lean
- FormalConjectures/OEIS/358684.lean
- FormalConjectures/OEIS/41.lean
- FormalConjectures/Paper/LatinSquare.lean
- FormalConjectures/WrittenOnTheWallII/GraphConjecture103.lean
- FormalConjectures/WrittenOnTheWallII/Test.lean

and removes the TODO comment in OEIS/41.lean that referred to the missing proofs.


FormalConjectures/ErdosProblems/340.lean
```diff
@@ -16,6 +16,7 @@ limitations under the License.
 module
 
 public import FormalConjecturesUtil
+meta import FormalConjecturesUtil
```
felixpernegger: Do we really want to meta import everything though?

mo271: the alternative would be to just import the bits from `FormalConjecturesForMathlib` that we need right, or what do you suggest?

If we go this way, then we need to adapter the linter forbidding imports from FormalConjecturesForMathlib to not trigger on `meta` imports

felixpernegger: That seems reasonable to me, yeah

felixpernegger: Btw, sorry if I am a bit annoying, but do you have any thoughts on
[#Formal conjectures > Modulization @ 💬](https://leanprover.zulipchat.com/#narrow/channel/524981-Formal-conjectures/topic/Modulization/near/625321079)? (even if you dislike it, it would be good to know so we can move on with the topic :) )

mo271: will think about this later, thanks for bringing it up again!

felixpernegger (approved):

FormalConjecturesUtil/Linters/ImportLinter.lean
```diff
@@ -28,6 +28,12 @@ This file implements a linter that enforces import conventions in `FormalConject
    `import FormalConjecturesUtil` instead.
 2. **Require `FormalConjecturesUtil`**: Problem files in `FormalConjectures` must
    import `FormalConjecturesUtil`.
+
+`meta import`s are exempt from the first rule. A `meta import` does not bring any declarations
+into scope; it only loads the compiled code of the imported module (and of its transitive imports),
+which is what `native_decide` and `#eval` need under the module system. Problem files that
+evaluate a definition from `Mathlib` or `FormalConjecturesForMathlib` natively should
+`meta import` just the module defining it, rather than all of `FormalConjecturesUtil`.
```
eric-wieser: Should this be linted for? That is, no `meta import FormalConjecturesUtil`?

eric-wieser (approved):

FormalConjecturesUtil/Linters/ImportLinter.lean
mo271: makes sense added if (and fixed one pre-existing finding)

mo271: FormalConjectures/Arxiv/2507.17780/4.lean

PR #6430, opened 2026-09-20
Erdős Problem 475: valid orderings with distinct partial sums

### LINKED ISSUE #749: Erdős Problem 475
### What is the conjecture
    
https://www.erdosproblems.com/475

  Let $p$ be a prime. Given any finite set $A \subseteq \mathbb{F}_p \setminus \\{ 0 \\}$, is there always a rearrangement 
$A=\\{a_1, \ldots,a_t\\}$ such that all partial sums $\sum^{1\leq k\leq m} a_k$ are distinct, for all $1\leq m\leq t$?
  

Status: open

### Choose either option

elio040208:
# Add formalisation of Erdős problem 475

Adds `FormalConjectures/ErdosProblems/475.lean`, the main question of
[erdosproblems.com/475](https://www.erdosproblems.com/475).

## What changed

The new file contains:

- `Erdos475.IsValidOrdering`, the "valid ordering" of the source: a rearrangement of `A` whose partial
  sums `∑_{1≤k≤m} a_k` are pairwise distinct for `1 ≤ m ≤ t`.
- `erdos_475 : answer(sorry) ↔ ∀ p, p.Prime → ∀ A : Finset (ZMod p), 0 ∉ A → ∃ l, IsValidOrdering A l`.

## Formalisation choices

- The source compares the partial sums only for `1 ≤ m ≤ t`, so the empty partial sum is **not** compared.
  An intermediate or the total partial sum may therefore be `0`, provided all positive-length partial sums
  remain pairwise distinct. `IsValidOrdering` follows the source literally. Some sequenceability conventions
  also compare against the empty partial sum, while allowing only a possible terminal return to `0`.
- `l.Nodup ∧ l.toFinset = A` says that `l` is a rearrangement of `A`, and it gives `l.length = A.card = t`.
- Primality is the explicit hypothesis `p.Prime` rather than `[Fact p.Prime]`, because no field structure is
  used. It also excludes `ZMod 0`, which is `ℤ` and not a finite field.
- Category `research open`: the source records the problem as proved for all sufficiently large primes and
  as resolved up to a finite check, so the universal statement is not proved. The AMS tags `5 11` follow the
  page's tags (combinatorics, number theory).

## How I checked

- `lake --wfail build 'FormalConjectures.ErdosProblems.«475»'` builds with no warnings (lean `v4.33.1`,
  mathlib `0df444a3`), from a clean state after deleting the `.olean`/`.ilean`.
- Statement semantics on instances (scratch file, not part of this PR):
  - `IsValidOrdering {1, 4} [1, 4]` over `ZMod 5` is accepted, with partial sums `1` and `0`.
  - `IsValidOrdering {1, 2, 4} [1, 4, 2]` over `ZMod 5` is accepted, with partial sums `1`, `0`, and `2`.
    This confirms that the source does not compare a zero intermediate partial sum with the empty sum.
  - `[1, 1]` for `A = {1}` is rejected (not a rearrangement).
  - `[1, 2, 3]` for `A = {1, 2, 3}` over `ZMod 5` is rejected (partial sums `1, 3, 1`).
  - `[2, 1, 3]` for the same `A` is accepted (partial sums `2, 3, 1`).
  - The empty set with the empty list is accepted.

## Not included

Formalisations of the known partial results (`t ≤ 12`; `p - 3 ≤ t ≤ p - 1`; all sufficiently large primes)
and the Alspach conjecture for general abelian groups. These could be added later as
`erdos_475.variants.*`.

Fixes #749


elio040208:
@googlebot rescan

PR #6437, opened 2026-09-20
feat(Book/Bugeaud): Problem 10.53 (solved)

rwst:
This adds problem +3 variants, all solved.


mo271 (commented):
Thanks!

Let's move blockComplexity, blockEntropy, partQuot, cfEntropy, baseEntropy to ForMathlib with the two tests as lemmas.

(Perhaps we already something in that direction?

rwst:
-awaiting-author

mo271 (approved):
Thanks!

PR #6438, opened 2026-09-20
feat(Book/Bugeaud): Problem 10.49 (solved)

rwst:
This adds problem +2 variants, all solved.


FormalConjectures/Books/BugeaudDistributionModuloOne/Problem10_49.lean
```diff
@@ -0,0 +1,140 @@
...
+/--
+A real number $x$ is *continued fraction normal* if for all $0 \le \alpha < \beta < 1$ the orbit
+of $x$ under the Gauss map visits $[\alpha, \beta)$ with asymptotic frequency
+$\mu_G([\alpha, \beta))$ [Sch17, (1.2)]. Equivalently, every block of partial quotients occurs
+with the frequency given by the Gauss measure.
+-/
+noncomputable def IsCFNormal (x : ℝ) : Prop :=
+  ∀ α β : ℝ, 0 ≤ α → α < β → β < 1 →
+    Tendsto (fun n : ℕ ↦ (((Finset.range n).filter
+      fun i ↦ gaussMap^[i] x ∈ Set.Ico α β).card : ℝ) / n) atTop (nhds (gaussKuzmin α β))
```
mo271, line 76: gaussMap, gaussKuzmin and IsCFNormal are reusable (continued-fraction normality will recur in this chapter); consider ForMathlib.

mo271 (commented):
Thanks!

one suggestion:
The PR defines IsNormalToBase b x := ∀ k ≥ 1, NormalNumber.IsNormalInBase (b^k) x and explains in two docstrings that NormalNumber.IsNormalInBase is only simple normality. That is wrong: in FormalConjecturesForMathlib/NumberTheory/NormalNumber.lean, IsNormalInBase b x already quantifies over every block length k and every word w with frequency 1/b^k, i.e. it is Borel normality; the simple notion is the separate IsSimplyNormalInBase. So IsNormalToBase b is equivalent to IsNormalInBase b and the Pillai detour is unnecessary. Use NormalNumber.IsNormalInBase b ξ directly, delete IsNormalToBase and the test lemma about it, and fix the module docstring. IsAbsolutelyNormal is used correctly.

PR #6446, opened 2026-09-21
fix(ErdosProblems): align problem 10 with 'every large integer'

AItoBit:
## Summary
- Align `erdos_10` with [erdosproblems.com/10](https://www.erdosproblems.com/10): the source asks whether some $k$ works for every **large** integer, not for all $n \notin \{0,1\}$.
- State this as `∃ k, ∀ᶠ n, n ∈ sumPrimeAndTwoPows k`.
- Add small tests for the definition (`2` is a prime, `1` is never representable).
- Record `[Ga75]` and `[GrSo98]` in the module references.

## Test plan
- [x] `lake --wfail build 'FormalConjectures.ErdosProblems.«10»'`
- [x] Category, AMS, and docstring match the source

Co-authored-by: Cursor <cursoragent@cursor.com>


mo271 (approved):
Thanks, good catch: seems like the word "large" was added upstream...

mo271:
The CLA check should pass when the commits are amended to exclude any non-human Co-authored-by fields

AItoBit:
> The CLA check should pass when the commits are amended to exclude any non-human Co-authored-by fields
I've just updated the cla but I dont know why I still having the same problem, can you help please? @mo271

mo271:
> I've just updated the cla but I dont know why I still having the same problem, can you help please? @mo271

[d4d49d5](https://github.com/google-deepmind/formal-conjectures/pull/6446/commits/d4d49d502fe78474471074fd546b1cc8c0037088)

contains a `Co-authored-by` filed by an entity that hasn't signed the CLA. One way to fix this is to amend that commit moving the `Co-authored-by` in the pr description. another might be so quash all commits and force push

PR #6450, opened 2026-09-21
feat(Wikipedia/KomlosConjecture): formalise the Karingula–Lovett proof

mo271:
Prove the Komlós conjecture with constant 36, following the elementary proof of Karingula and Lovett (arXiv:2609.20979). The proof lives in FormalConjecturesForMathlib/Combinatorics/Discrepancy/:

- Balancing.lean: finitely supported distributions, the bit-splitting operator, and the balancing lemma (Lemma 1.3 of the paper).
- TentDistribution.lean: a discrete tent distribution on the lattice cube with overlap at least 2/3 with its small translates (Lemma 1.4, done with sums in place of integrals).
- Komlos.lean: transport to ℝ^d, the rational case, and the general case by truncation and pigeonhole.

komlos_conjecture is now research solved with a proof, and the Banaszczyk variant is derived from it.


Paul-Lez (approved):
Nice! I guess this could also be a nice Palomar submission:)

mo271:
@henrykmichalewski thanks for the formalisation of the statement again!

PR #6456, opened 2026-09-22
Mark OEIS conjectures settled by OEIS Open as research solved

mo271:
Three more OEIS statements were settled in the OEIS Open project (T. Adamczewski, "OEIS Open: How many conjectures can language models turn into theorems?", arXiv:2608.11941), with accepted Lean proofs in
epoch-research/LeanOpenProblems-results. Update them:

- A176477: `conjecture1` (integrality and positivity) and `conjecture2` (a(n) odd iff n = 2^m) are proved. Mark both `research solved` and link the proof, which uses the same recurrence.
- A306477: Sun's 2-4-6-8 conjecture is false, n = 896315812331399 has no representation. Negate `conjecture`, mark it `research solved`, link the proof, and record the counterexample in the docstrings.
- A060841: split the conjunction into `conjecture1` (1/det(M) is an integer iff n is 1..34, 36, 38; still open) and `conjecture2` (all denominators are powers of two). The latter is false at n = 1807, which the OEIS entry now records. Mark it `research solved` with the negated statement. No `formal_proof` is attached because the Lean disproof works with the closed form prod k^2/phi(k) rather than the determinant.

Proof links are pinned to commit fd09021e of LeanOpenProblems-results.


mo271:
probably also good to report the 2-4-6-8 counter example to OEIS and https://mathoverflow.net/questions/323541/positive-integers-written-as-binomw2-binomx4-binomy6-binomz8-with

smmercuri (approved):

PR #6482, opened 2026-09-22
fix(Arxiv/2209.04540): make Problem 7.1 an existential over dimension

### LINKED ISSUE #6457: Arxiv/2209.04540/SpectralSetsAndWeakTiling: `exists_nowhereDense_spectralSet` asks for an example in every dimension `d ≥ 2`, not in some dimension
`NowhereDenseSpectralSet.exists_nowhereDense_spectralSet` puts the `answer(sorry)` slot opposite `∀ d ≥ 2, ∃ Ω : Set (Fin d → ℝ)` bounded, measurable, nowhere dense, of positive measure and spectral, whereas KLM Problem 7.1 asks whether such an Ω exists at all in dimensions two and higher; because a product with a spectral cube carries an example from dimension d to every larger dimension, the Lean proposition is equivalent to the two-dimensional case rather than to the source's question.

https://tadamcz.com/fc-review-results/782312817c/#/f/Arxiv/2209.04540/SpectralSetsAndWeakTiling

- 1 misformalization
- 2 equivalent reformulations
- trivial proof: bailed out
- fix compiles

smmercuri:
[KLM2023, Problem 7.1] asks: "Let $\Omega \subset \mathbb{R}^d$ be a bounded, nowhere dense set of positive measure. Can $\Omega$ be a spectral set? ... the answer is known to be negative in dimension one, but in dimensions two and higher the problem is open." The Lean statement instead put `answer(sorry)` opposite `∀ d ≥ 2, ∃ Ω ...`, i.e. the assertion that such a set exists in *every* dimension `d ≥ 2`. Since a product with a spectral cube preserves boundedness, nowhere density, positive measure and spectrality, existence in dimension `d` implies existence in every larger dimension, so the universal form is equivalent to the single case `d = 2` rather than to the source's question. Its negation ("some dimension has no such set") is also much weaker than the expected answer, which is that no such set exists in any dimension.

**Change:** the dimension quantifier becomes existential, `answer(sorry) ↔ ∃ d : ℕ, 2 ≤ d ∧ ∃ Ω : Set (Fin d → ℝ), ...`, so a negative answer says exactly that no bounded, nowhere dense set of positive measure in any dimension `d ≥ 2` is spectral. The docstring now states the question without a dimension quantifier and explains why `d ≥ 2` is imposed (the case `d = 1` is settled negatively). Category, AMS tags and the rest of the file are unchanged.

**Checked:** `lake --wfail build 'FormalConjectures.Arxiv.«2209.04540».SpectralSetsAndWeakTiling'` — completed successfully, no warnings.

Judgement call: `d` is bound after `answer(sorry)` rather than as a theorem argument, per the repository convention that all quantifiers follow the answer slot; this also keeps the whole open problem, not one dimension of it, inside the slot.

Fixes #6457


smmercuri:

I compared this against issue #6457, the audit page for this file, and the LaTeX source of arXiv:2209.04540, and stress-tested the new statement in a scratch Lean file importing the module; `lake --wfail build 'FormalConjectures.Arxiv.«2209.04540».SpectralSetsAndWeakTiling'` completes with no warnings. Verdict: **CORRECT**.

One nit: the docstring records the `d ≥ 2` restriction but not that `d` is existentially quantified.

Why it is the right fix:

- **The defect is gone.** The old `∀ᵉ (d : ℕ) (hd : 2 ≤ d)` form collapsed to the case `d = 2` by instantiation alone, whereas the new right-hand side negates to "no such set in any dimension `d ≥ 2`", the expected answer.
- **Equivalent to the audit's proposal.** `exists_nowhereDense_spectralSet.mpr h` typechecks on an `h` of the proposal's type, so the two propositions are definitionally the same.
- **Faithful to the source.** §7 asks whether a bounded, nowhere dense set of positive measure can be spectral, noting the answer is "known to be negative in dimension one, but in dimensions two and higher the problem is open".
- **No junk values or vacuity.** `isSpectral` demands positive finite measure and a total orthogonal family, and I proved the excluded `d = 0` carries no witness anyway, since every positive-measure set there is the whole nonempty space.

mo271 (approved):

PR #6483, opened 2026-09-22
fix(ErdosProblems/35): assume positive density in Plünnecke variant

### LINKED ISSUE #6464: ErdosProblems/35: `erdos_35.variants.plunnecke` omits Plünnecke's positive-density hypothesis and is false
`erdos_35.variants.plunnecke` asserts `schnirelmannDensity A ^ (1 - (k : ℝ)⁻¹) ≤ schnirelmannDensity (A + B)` for every `A` with no lower bound on `schnirelmannDensity A`, so it fails at `B = Set.univ`, `k = 1`, `A = ∅`, where the exponent is 0, `Real.rpow` gives `0 ^ (0 : ℝ) = 1` and the sumset has Schnirelmann density 0 (and again at `A = Set.Ici 2`), whereas Plünnecke's Satz 1.2 assumes a density parameter in `(0, 1]`.

https://tadamcz.com/fc-review-results/782312817c/#/f/ErdosProblems/35

- 1 misformalization
- 1 minor
- 1 equivalent reformulation
- trivial proof compiles
- fix compiles

smmercuri:
`erdos_35.variants.plunnecke` stated Plünnecke's bound `d_s(A)^(1-1/k) ≤ d_s(A+B)` for *every* `A ⊆ ℕ`, with no lower bound on `d_s(A)`. Plünnecke's Satz 1.2 (*Eine zahlentheoretische Anwendung der Graphentheorie*, 1970, p. 172) states it for a density parameter in `(0, 1]`. Without that hypothesis the Lean statement is false: `B = Set.univ`, `k = 1`, `A = ∅` satisfy `0 ∈ B` and `B.IsAddBasisOfOrder 1`, while the exponent is `0`, `Real.rpow` gives `(0 : ℝ) ^ (0 : ℝ) = 1`, and `d_s(∅ + univ) = 0`, so the conclusion reads `1 ≤ 0`. (`A = Set.Ici 2` breaks it the same way, so requiring `A` nonempty or infinite would not suffice.)

**Change:** add `(hA : 0 < schnirelmannDensity A)` to the variant and qualify both docstrings with `d_s(A) > 0`. The main theorem `erdos_35` is untouched: its polynomial bound is trivially true at density zero. Separately, the module bibliography's page range for [Er36c] is corrected to Acta Arith. 1 (1936), 197-200 (the article is four pages, with running heads 198-200).

**Checked:** `lake --wfail build 'FormalConjectures.ErdosProblems.«35»'` — `Build completed successfully`, no warnings.

Judgement call: the erdosproblems.com/35 remark quotes the power bound without the endpoint qualification, so the `erdos_35` docstring now adds "and for $\alpha>0$" to keep the quoted remark in step with what Plünnecke actually proved.

Fixes #6464


smmercuri:

I checked this against issue #6464, the audit page, erdosproblems.com/35 and the Rényi Institute's Erdős bibliography, and stress-tested the statement in Lean. **Verdict: CORRECT.**

Why it is the right fix:

- **The defect is gone.** The audit's disproof, re-run verbatim against this head, now leaves `schnirelmannDensity ∅ ≤ 0 ⊢ False` unsolved.
- **Equivalent to the audit's proposal.** Restating the suggested fix in a fresh namespace and comparing via `type_of%` closes by `Iff.rfl`, so the two are the same proposition.
- **Faithful to the source.** The docstrings now add a positivity qualifier that erdosproblems.com omits, which is right because the unqualified claim is false in Lean at density zero; Plünnecke's original could not be read here, so its "(0,1]" phrasing rests on the audit, while the necessity of `0 < d_s A` is established independently.
- **No junk values or vacuity.** `k = 0` is impossible and `k = 1` forces `B = univ`, where `0 < d_s A` gives `1 ∈ A` and the boundary case holds; both audited counterexamples are excluded, the hypotheses stay satisfiable, `erdos_35` is rightly untouched, and `lake --wfail build` succeeds cleanly.

mo271 (approved):

PR #6484, opened 2026-09-22
fix(ErdosProblems/308): shift Croot's bounds by one

### LINKED ISSUE #6463: ErdosProblems/308: `erdos_308.variants.croot` bounds `f N` where Croot's theorem bounds `f N - 1`
`Erdos308.erdos_308.variants.croot` places `f N`, defined in this file as the smallest non-representable positive integer, between the two floors of Croot's Main Theorem, but Croot bounds η(N), the largest integer up to which every integer is representable, which is `f N - 1`; since (log log N)²/log N → 0 the shift is not absorbed by the o(1) factors, and the stated upper bound eventually forces `f N ≤ m N`, contradicting Croot's lower bound, which gives `f N = m N + 1` for arbitrarily large N.

https://tadamcz.com/fc-review-results/782312817c/#/f/ErdosProblems/308

- 1 misformalization
- 2 equivalent reformulations
- trivial proof: bailed out
- fix compiles

smmercuri:
`erdos_308.variants.croot` placed `f N` — defined in this file as the smallest non-representable positive integer — between Croot's two floors. Croot's Main Theorem [Cr99] instead defines η(x) as *the largest* integer such that every integer in {1,…,η(x)} is representable, and it is η that the two displayed floors bound. Since η(N) = f(N) − 1 and the correction term (log log N)²/log N tends to 0, the shift is not absorbed by the `o(1)` factors: as stated, the upper bound gives `f N ≤ ⌊H_N⌋ = m N` for all large `N`, which rules out the `f N = m N + 1` alternative of `erdos_308.parts.i` and the `representable N = Set.Icc 1 (m N)` alternative of `erdos_308.variants.shape` — both of which Croot shows do occur (p. 2: `m` is representable exactly when the fractional part of `H_N` exceeds the correction).

**Change:** added `+ 1` outside each floor in `erdos_308.variants.croot`, keeping the definition of `f` as it is, and rewrote the docstring to say which quantity Croot bounds and why the `+ 1` appears. No other declaration touched; `category` and AMS tags unchanged; the proof stays `by sorry`.

**Checked:** `lake --wfail build 'FormalConjectures.ErdosProblems.«308»'` — `Build completed successfully`, no warnings.

Judgement call: the prose on erdosproblems.com/308 carries the same off-by-one ("if f(N) is the smallest integer not representable then …"). I followed the primary paper, which this docstring explicitly cites, rather than the secondary transcription; the alternative would have been to restate the bounds for `f N - 1`, which is equivalent but reads less directly.

Fixes #6463


smmercuri:

Compared against issue #6463, the audit page for `ErdosProblems/308`, and Croot's manuscript, then stress-tested in a Lean scratch file importing the module. Verdict: **correct**.

Why it is the right fix:

- **The defect is gone.** I proved sorry-free that the pre-PR statement forces `∀ᶠ N in atTop, f N ≤ m N`, since `o₂ → 0` makes the subtracted term eventually nonnegative, while the PR head yields only `f N ≤ m N + 1`, so both branches of `erdos_308.parts.i` survive.
- **Equivalent to the audit's proposal.** The new statement is character-identical to the audit's suggested fix, confirmed by `Iff.rfl` against a verbatim restatement that itself typechecks against the declaration.
- **Faithful to the source.** Croot's Main Theorem bounds "the largest integer such that, whenever 1 ≤ n ≤ n(x)" every such n is representable, which is `f(N) - 1`, so each floor needs `+ 1`; erdosproblems.com repeats the same off-by-one, so following the cited paper is right.
- **No junk values, and it builds.** The `+ 1` is ℕ-addition outside the floor, not the ℕ-subtraction alternative, which is strictly weaker at the `sInf ∅` junk value `f N = 0`; `lake --wfail build` is warning-free.

mo271 (approved):

PR #6486, opened 2026-09-22
fix(ErdosProblems/1180): bound summands for large primes only

### LINKED ISSUE #6461: ErdosProblems/1180: `variants.shparlinski` and `variants.glibichuk` bound the all-primes constant `C ε`, which is not polynomial in `ε⁻¹`
`Erdos1180.erdos_1180.variants.shparlinski` and `Erdos1180.erdos_1180.variants.glibichuk` assert `(C ε : ℝ) ≤ K * ε⁻¹ ^ 3` and `≤ K * ε⁻¹ ^ 2` for the file's `C ε`, the least summand bound valid for *all* primes, whereas Shparlinski's and Glibichuk's bounds are stated for primes above a threshold depending on ε; taking ε = log(3/2)/log p makes 1 the only admissible denominator, so `C ε ≥ p − 1` while the asserted bounds are only O((log p)^3) and O((log p)^2).

https://tadamcz.com/fc-review-results/782312817c/#/f/ErdosProblems/1180

- 2 misformalizations
- 1 questionable
- 1 equivalent reformulation
- trivial proof: bailed out
- fix compiles

smmercuri:
`variants.shparlinski` and `variants.glibichuk` asserted `(C ε : ℝ) ≤ K * ε⁻¹ ^ 3` and `≤ K * ε⁻¹ ^ 2` for the file's `C ε`, which is the least summand bound valid for *every* prime. No bound on that quantity can be polynomial in `ε⁻¹`: taking `ε = log(3/2)/log p` gives `p ^ ε = 3/2`, so `1` is the only admissible denominator and representing `-1` needs `p - 1` summands, giving `C ε ≥ p - 1` against claimed bounds of size `O((log p)^3)` and `O((log p)^2)`. The primary sources state the quantitative bounds with a prime threshold depending on `ε`: Glibichuk's abstract reads "Consider an arbitrary ε > 0 and a sufficiently large prime p > 2 ... N = 8([1/ε + 1/2] + 1)^2", and Shparlinski's `≪ ε⁻³` bound is likewise a large-prime statement (his all-primes abstract gives only a qualitative `k(ε)`).

**Change:** both variants now quantify a prime threshold `P` after `ε` and bound the number of summands directly, `∃ K, ∀ ε ∈ (0, 1], ∃ P, ∀ p ≥ P prime, ∀ a, ∃ s, (s.card : ℝ) ≤ K * ε⁻¹ ^ 3 ∧ Represents ε p a s` (and `^ 2` for Glibichuk). The matching sentence in the `erdos_1180` docstring now says the `ε⁻³` bound is for sufficiently large primes. The qualitative theorem and `lower_bound` are unchanged: the all-primes qualitative statement is fine, since for the finitely many primes with `p ^ ε < 2` one can use `a.val` copies of the denominator `1`.

**Checked:** `lake --wfail build 'FormalConjectures.ErdosProblems.«1180»'` — completes successfully with no warnings.

Judgement call: I bounded `s.card` inline rather than introducing a second `noncomputable def` for the large-prime constant, to keep the change local; `C` is still used by `lower_bound` and `optimal`. The audit also flags `variants.optimal` as questionable for the same reason (it uses the all-primes `C`), but the website reuses `C_ε` there without introducing a threshold, so that ambiguity is left alone and is not part of this issue.

Fixes #6461


smmercuri:

Compared against issue #6461, the audit page, Glibichuk's and Shparlinski's abstracts, and erdosproblems.com/1180, then stress-tested in Lean. **Verdict: correct.**

Why it is the right fix:

- **The defect is gone.** I reproved the mechanism sorry-free — `(p : ℝ) ^ ε < 2` forces denominator `1`, so `C ε ≥ p - 1` at `ε = log(3/2)/log p` — but it no longer bites: the threshold follows `ε`, and neither variant mentions `C`.
- **Equivalent to the audit's proposal.** With `Cbig ε := sInf {B | ∃ P, ∀ p ≥ P, …}`, the new statement provably gives `(Cbig ε : ℝ) ≤ K * ε⁻¹ ^ 2` for the same `K`; the converse needs nonemptiness since `Nat.sInf ∅ = 0`, so this form avoids the junk value.
- **Faithful to the source.** Glibichuk's abstract reads "Consider an arbitrary ε > 0 and a sufficiently large prime p > 2", giving `N = 8([1/ε + 1/2] + 1)^2`; Shparlinski's states only a qualitative all-primes constant, so the large-prime reading of `ε⁻³` is the only consistent one.
- **No vacuity, and it builds.** `∃ P` cannot empty the quantifier, categories, AMS tags and `C` are unchanged, nothing else imports the module, and `lake --wfail build` is clean.

Possible follow-up, not requested here: `lower_bound` and `optimal` still use the all-primes `C`, so `optimal`'s right-hand side is degenerate by the same family.

mo271 (approved):

PR #6487, opened 2026-09-22
fix(ErdosProblems/361): drop the elementary Theta asymptotic variant

### LINKED ISSUE #6465: ErdosProblems/361: `erdos_361.asymptotic` is answered by `fun n ↦ (n : ℝ)` for every `c > 0`, so the Θ question is elementary
`Erdos361.erdos_361.asymptotic` asks only for the `=Θ[atTop]` class of `fun n ↦ (subsetSumAvoidanceNumber c n : ℝ)`, which `answer(fun n : ℕ ↦ (n : ℝ))` satisfies for every `c > 0` (for an integer `k ≥ 1` with `1/k ≤ c`, the integers strictly between `n/(k+1)` and `n/k` form an admissible set of linear size, and the ambient `{1, …, ⌊cn⌋}` bounds it above), so the declaration does not pose the source's open problem, which asks for the exact size and its irregular dependence on `n` and which `erdos_361` already states faithfully.

https://tadamcz.com/fc-review-results/782312817c/#/f/ErdosProblems/361

- 1 misformalization
- trivial proof: bailed out
- fix compiles

smmercuri:
`Erdos361.erdos_361.asymptotic` asked only for the `=Θ[atTop]` class of
`fun n ↦ (subsetSumAvoidanceNumber c n : ℝ)`. That class is `fun n : ℕ ↦ (n : ℝ)` for every
`c > 0`: pick an integer `k ≥ 1` with `1/k ≤ c`; the integers strictly between `n/(k+1)` and
`n/k` avoid `n` as a subset sum (any `≤ k` of them sum below `n`, any `≥ k+1` of them sum above
`n`) and there are `n/(k(k+1)) + O(1)` of them, while the ambient `{1, …, ⌊cn⌋}` bounds the size
above by `cn`. So the declaration was an elementary linear-growth statement marked
`research open`, not the source's problem. [erdosproblems.com/361](https://www.erdosproblems.com/361)
asks for the exact size of the largest `A ⊆ {1, …, ⌊cn⌋}` with `n` not a subset sum of `A`, and
whether it depends on `n` in an irregular way; it poses no separate question about growth up to
constant factors.

**Change:** remove `erdos_361.asymptotic` and the `open Filter` it was the only user of.
`erdos_361`, which already states the source's question faithfully, and the definitions and test
are untouched.

**Checked:** `lake --wfail build 'FormalConjectures.ErdosProblems.«361»'` — build completed
successfully, no warnings.

Judgement call: I removed the variant rather than restating it with `~[atTop]`. An asymptotic
formula need not exist here — the source explicitly raises the possibility that the size depends
on `n` irregularly — so a `~[atTop]` version would risk being unsatisfiable as posed, and the
exact-size question is already covered by `erdos_361`.

Note for reviewers: open PR #5446 also edits this file, rewriting the `erdos_361` statement
(a different declaration), so the two may conflict textually.

Fixes #6465


smmercuri:

I compared this against issue #6465, the audit page, and [erdosproblems.com/361](https://www.erdosproblems.com/361), and stress-tested it by re-proving the deletion's premise in Lean rather than trusting the PR body. Verdict: **CORRECT**.

Why it is the right fix:

- **The defect is gone, and I machine-checked why it was one.** I proved `(fun n ↦ (subsetSumAvoidanceNumber c n : ℝ)) =Θ[atTop] (fun n : ℕ ↦ (n : ℝ))` for every `c > 0`, sorry-free — `#print axioms` gives only `propext, Classical.choice, Quot.sound`, no `sorryAx` — so the removed answer slot had a uniform elementary answer.
- **Faithful to the source.** The page asks only for the exact size and "Does this depend on $n$ in an irregular way?", so a `~[atTop]` restatement would be worse than deletion: an asymptotic formula need not exist, and `erdos_361` already states the exact-size question.
- **Every change is justified.** Only the variant and the `open Filter` it alone used go, `rg erdos_361` finds no other reference in the repo, and no category or pinned subset moves.
- **Builds.** `lake --wfail build 'FormalConjectures.ErdosProblems.«361»'` is clean; open PR #5446 edits line 68 of this file, a textual but not semantic conflict.

The Θ proof is ~120 lines of Lean (interval construction `Finset.Ioo (n/(k+1)) (n/k)` with `k = max ⌈1/c⌉₊ 1`, giving `≥ n/(k(k+1)) − 2` below and `≤ c·n` above); happy to paste it or push it as a scratch file if a maintainer wants to check it.

mo271 (approved):

PR #6488, opened 2026-09-22
fix(ErdosProblems/1108): restrict factorial indices to be positive

### LINKED ISSUE #6460: ErdosProblems/1108: `FactorialSums` allows index 0, so 0! and 1! are counted separately and the set is too large
`Erdos1108.FactorialSums` sums `n.factorial` over an arbitrary `Finset ℕ`, so a finset containing both 0 and 1 counts the value 1 twice and the set contains e.g. 4 = 0! + 1! + 2! and every n! + 4 with n ≥ 3, none of which is a sum of distinct positive-index factorials, whereas the source's sums ∑ εᵢ i! run over i ≥ 1 (Brindza–Erdős, eq. (2)), so both `erdos_1108.parts.i` and `erdos_1108.parts.ii` ask their finiteness questions about a strictly larger set A.

https://tadamcz.com/fc-review-results/782312817c/#/f/ErdosProblems/1108

- 1 misformalization
- trivial proof compiles
- fix compiles

smmercuri:
`Erdos1108.FactorialSums` summed `n.factorial` over an arbitrary `Finset ℕ`, so a finset
containing both `0` and `1` contributed the value `1` twice. The Lean set therefore contained
`4 = 0! + 1! + 2!`, and more generally every `n! + 4` with `n ≥ 3`, none of which is a sum of
distinct factorials. The source's sums are indexed from `1` (Brindza–Erdős, *On some Diophantine
problems involving powers and factorials*, §1 eq. (2): `∑ εᵢ i!` with `i ≥ 1`), which is also
what the docstring "sums of distinct factorials" already claimed. Both `erdos_1108.parts.i` and
`erdos_1108.parts.ii` were consequently asking their finiteness question about a strictly larger
set.

**Change:** require the indices to be positive in `FactorialSums`, and say so in the three
docstrings. The empty finset is still allowed; its contribution of `0` is harmless for both
finiteness questions. Statements, `@[category research open]` and the AMS tag are unchanged.

**Checked:** `lake --wfail build 'FormalConjectures.ErdosProblems.«1108»'` — `Build completed
successfully`, no warnings. A scratch file additionally confirms `0 ∈ FactorialSums`,
`9 = 1! + 2! + 3! ∈ FactorialSums` and `4 ∉ FactorialSums`.

Judgement call: the erdosproblems.com page writes `S ⊂ ℕ` without qualification, but on that page
`ℕ` must exclude `0` here for "distinct factorials" to be meaningful, and the cited primary
reference indexes from `1`. I kept the restriction as a side condition on `S : Finset ℕ` rather
than changing the index type, to keep the diff minimal.

Fixes #6460


smmercuri:

Checked against #6460, the audit page and the primary reference, and stress-tested in Lean; the audit's proposal restated in a fresh namespace satisfies `AuditFix = Erdos1108.FactorialSums := rfl`, so this is exactly the suggested fix. Verdict: **correct**.

Why it is the right fix:

- **Faithful to the source.** Brindza–Erdős, *On some Diophantine problems involving powers and factorials* (1991), p. 2 eq. (2) reads `∑_{i=1}^∞ εᵢ i! = x^z` with lower limit `i = 1`; erdosproblems.com writes `S ⊂ ℕ` unqualified, but since `0! = 1!` only the paper's indexing makes "sums of distinct factorials" meaningful.
- **The defect is gone.** The audit's `⟨{0, 1, 2}, by norm_num⟩` demo no longer compiles (its goal reduces to `False`), `(4 : ℕ) ∉ Erdos1108.FactorialSums` is now provable, and the old definition is strictly larger.
- **Nothing is trivialised.** `0` (empty sum), `1 = 1!` and `9 = 1! + 2! + 3!` stay in the set, `0` remains a `k`-th power in it for every `k ≥ 2` and `1` remains a powerful member, so both finiteness questions keep their content.
- **Builds.** `lake --wfail build 'FormalConjectures.ErdosProblems.«1108»'` is warning-free, and the statements, the `research open` category and the AMS tag are unchanged.

FormalConjectures/ErdosProblems/1108.lean
```diff
@@ -30,27 +30,28 @@ open Nat Filter BigOperators
...
 -/
 def FactorialSums : Set ℕ :=
-  {m : ℕ | ∃ S : Finset ℕ, m = ∑ n ∈ S, n.factorial}
+  {m : ℕ | ∃ S : Finset ℕ, (∀ n ∈ S, 0 < n) ∧ m = ∑ n ∈ S, n.factorial}
 
 /--
 A number is powerful if each prime factor appears with exponent at least 2.
 -/
 def IsPowerful (n : ℕ) : Prop :=
   ∀ p : ℕ, p.Prime → p ∣ n → p ^ 2 ∣ n
```
mo271, line 44: while at it, can we use the usual IsPowerful def (or make a follow-up fixing all instances where good api is not used...)

PR #6491, opened 2026-09-22
fix(ErdosProblems/664): restore the block-size bound in block_design

### LINKED ISSUE #6470: ErdosProblems/664: `erdos_664.variants.block_design` omits the block-size bound `|A i| > c√n` and the parameter `c`
`Erdos664.erdos_664.variants.block_design` asks for one absolute `K` for every family satisfying only the unique-pair-coverage hypothesis, dropping the parameter `c` and the hypothesis `c * √n < (A i).card` that [Er81] §VI keeps (only `c < 1` is meant to be omitted in this variant), so its right-hand side is false for trivial reasons: with n = m = 1 and `A 0 = ∅` the hypothesis holds vacuously but no `B` meets the empty block.

https://tadamcz.com/fc-review-results/782312817c/#/f/ErdosProblems/664

- 1 misformalization
- trivial proof compiles
- fix compiles

smmercuri:
`erdos_664.variants.block_design` asked for a single absolute `K` that works for every family
satisfying only unique pair coverage, dropping both the parameter `c` and the block-size
hypothesis `|A i| > c √n`. [Er81] §VI (first block-design problem) asks whether "to every `c₁`
there is a `c₂`" such that a pairwise balanced block design with `|A_i| > c₁ n^{1/2}` admits a
set `B` with `1 ≤ |B ∩ A_i| ≤ c₂`; Alon's note restates the same problem for block designs,
keeping the size bound. Only the extra restriction `c < 1` is omitted in this variant. As stated,
the right-hand side was false for trivial reasons: `n = m = 1` with `A 0 = ∅` satisfies the
hypothesis vacuously (no distinct points), yet no `B` meets the empty block.

**Change:** quantify `∀ c : ℝ, 0 < c →` before `∃ K`, so `K` may depend on `c`, and add the
hypothesis `(∀ i, c * √n < (A i).card)` before the unique-pair-coverage hypothesis. The docstring
now says the size bound is kept and only `c < 1` is dropped. Nothing else changed; the statement
stays `research open` with `AMS 5` and `by sorry`.

**Checked:** `lake --wfail build FormalConjectures.ErdosProblems.«664»` — completed successfully,
no warnings.

Judgement call: `c` ranges over all positive reals rather than `c < 1`, matching Erdős's "to every
`c₁`" and the existing docstring note that `c < 1` is what this variant omits.

Fixes #6470


smmercuri:

Checked against issue #6470, the audit page for `ErdosProblems/664`, and the primary source, then stress-tested the new statement in Lean. **Verdict: CORRECT.**

Why it is the right fix:

- **Faithful to the source.** [Er81], p. 12 asks whether "to every `c₁` there is a `c₂`" such that a pairwise balanced block design "satisfying `|A_i| > c₁ n^{1/2}`" admits `Q` with `1 ≦ |Q ∩ A_i| ≦ c₂`, so the size bound is kept and no `c < 1` appears; Alon's note, §3, confirms this block-design version is the one still open.
- **The defect is gone.** I proved `¬ OldRhs` for the old right-hand side (`n = m = 1`, `A 0 = ∅`), checked that the same witness now fails the restored size hypothesis for every `c > 0`, and the new right-hand side matches the audit's proposed fix by `Iff.rfl`.
- **Quantifying `c` over all positive reals is safe.** I proved in Lean that it is equivalent to restricting to `c < 1`, since the hypothesis is monotone in `c` and `K` may depend on it (pass to `min c (1/2)`).
- **No junk values, and it builds.** `0 < c` with `c * √n < (A i).card` forces every block nonempty and makes `n = 0` force `m = 0`, `m = 0` with `n ≥ 2` falsifies coverage, no other file mentions the declaration, and `lake --wfail build FormalConjectures.ErdosProblems.«664»` is warning-free.

mo271 (approved):

PR #6492, opened 2026-09-22
fix(ErdosProblems/673): require 1 < m in Tao's divisor bound

### LINKED ISSUE #6471: ErdosProblems/673: `variants.tao` allows `m = 1`, making Tao's divisor bound `τ(n/m)/m ≤ G n` false
`Erdos673.erdos_673.variants.tao` assumes only `hn : 0 < n` and `hm : m ∣ n`, so `m = 1` is admitted and the asserted `((n / m).divisors.card : ℝ) / m ≤ G n` fails — n = m = 1 gives 1 ≤ G 1 = 0 and n = 2, m = 1 gives 2 ≤ 1/2 — whereas Tao's observation and the file's own cited Lean proof require `1 < m`.

https://tadamcz.com/fc-review-results/782312817c/#/f/ErdosProblems/673

- 1 misformalization
- 1 equivalent reformulation
- trivial proof compiles
- fix compiles

smmercuri:
`erdos_673.variants.tao` assumed only `hn : 0 < n` and `hm : m ∣ n`, so `m = 1` was admitted and the asserted lower bound `((n / m).divisors.card : ℝ) / m ≤ G n` is false: `n = m = 1` gives `1 ≤ G 1`, while `G 1 = 0` (empty sum), and `n = 2, m = 1` gives `2 ≤ 1/2`. Tao's observation is about a nontrivial divisor — the site's own illustration `τ(n)/4 ≤ G(n) ≤ τ(n)` for even `n` comes from `m = 2` — and the file's cited external Lean proof states its `tao_lower_bound` with `(hm : 1 < m)`.

**Change:** add `hm : 1 < m` to `erdos_673.variants.tao` (renaming the divisibility hypothesis to `hmn`), and qualify the bound with `$m>1$` in this theorem's docstring and in the docstring of `erdos_673.parts.i`. The conclusion and the `@[category textbook, AMS 11]` attributes are unchanged.

**Checked:** `lake --wfail build 'FormalConjectures.ErdosProblems.«673»'` — `Build completed successfully`.

Judgement call: the erdosproblems.com prose writes "for any divisor $m \mid n$" without the restriction, but the inequality is simply false at `m = 1`, so the docstrings now say `m > 1` explicitly. The upper bound `G n ≤ τ(n)` holds for all `n` and is kept in the same conjunction as the source states it, so it is now stated only under `1 < m`.

Fixes #6471


smmercuri:

Checked against issue #6471, the audit page for `ErdosProblems/673`, erdosproblems.com/673 and the cited external Lean proof, and stress-tested in Lean. Verdict: **CORRECT**. A half-sentence in the docstring noting that the source omits `m > 1` would make the deviation self-documenting.

Why it is the right fix:

- **Faithful to the source.** The site writes "for any divisor $m\mid n$" unqualified, but its own illustration $\tau(n)/4\leq G(n)\leq\tau(n)$ for even $n$ is precisely the $m=2$ case, and the cited proof (plby/lean-proofs @8822f7d, L257-259) states `tao_lower_bound` with `(hm : 1 < m)` under the comment "with the necessary hypothesis `m > 1`".
- **The defect is gone, and this matches the audit's proposal.** Applying the new statement at `n = m = 1` fails with ``decide proved that the proposition 1 < 1 is false`` (captured under `#guard_msgs`, the old type refuted outright since `G 1 = 0`), and I proved both directions of the equivalence with the audit's version, which differs only in binder order and names.
- **No junk values or vacuity.** `G 0 = G 1 = 0` under ℕ-subtraction, and confining the upper bound to `1 < m` loses exactly the instance `G 1 ≤ τ(1)`, i.e. `0 ≤ 1`, since every `n ≥ 2` has a prime factor — so splitting the conjunction is not warranted.
- **Builds.** `lake --wfail build 'FormalConjectures.ErdosProblems.«673»'` completes successfully, and the head sha's checks are green.

mo271 (approved):

PR #6495, opened 2026-09-22
fix(ErdosProblems/748): use real exponent in the lower bound

### LINKED ISSUE #6473: ErdosProblems/748: `erdos_748.variants.lower_bound` uses natural division in `2 ^ (n / 2)`, weakening the sourced `2^(n/2)` bound
`Erdos748.erdos_748.variants.lower_bound` states `2 ^ (n / 2) ≤ f n` with `n / 2` computed in ℕ, so for odd `n = 2m+1` it requires only `2^m ≤ f n` rather than the `2^(n/2) = 2^m·√2` of the remark on erdosproblems.com/748 that its own docstring quotes (at `n = 1` it asks for `f 1 ≥ 1` instead of `f 1 ≥ √2`), and there is no asymptotic notation or constant here to absorb the rounding.

https://tadamcz.com/fc-review-results/782312817c/#/f/ErdosProblems/748

- 1 misformalization
- 3 equivalent reformulations
- trivial proof compiles
- fix compiles

smmercuri:
`erdos_748.variants.lower_bound` stated `2 ^ (n / 2) ≤ f n` entirely in `ℕ`, so the exponent was
`⌊n/2⌋`. For odd `n = 2m + 1` that asks only for `2 ^ m ≤ f n`, whereas erdosproblems.com/748 —
quoted verbatim by the declaration's own docstring — remarks that `f(n) ≥ 2^{n/2}`, i.e.
`2 ^ m * √2` at such `n` (at `n = 1`: `f 1 ≥ √2`, not `f 1 ≥ 1`). This is a pointwise inequality
with no asymptotic notation or constant to absorb the rounding, so the natural division was a
strict weakening.

**Change:** state the bound over `ℝ` with a real exponent, `(2 : ℝ) ^ ((n : ℝ) / 2) ≤ (f n : ℝ)`,
matching the way `erdos_748.variants.green_sapozhenko` in the same file already writes `2^(n/2)`.
The statement stays `by sorry`, and the category and AMS tags are unchanged.

**Checked:** `lake --wfail build 'FormalConjectures.ErdosProblems.«748»'` — completes successfully
with no warnings.

One judgement call: the docstring's "considering all subsets of $[n/2,n]$" is left as is. It
reproduces the source's wording, and the endpoint is slightly loose there (for even `n` one needs
the strict upper half `(n/2,n]`), but that is the source's own expository slip rather than a
Lean/source discrepancy, so it is out of scope for this fix.

Fixes #6473


smmercuri:

Compared against issue #6473, the audit page for `ErdosProblems/748` and erdosproblems.com/748 (live and 2026 wayback capture agree verbatim), with the new statement stress-tested in Lean. **The fix is correct.**

Why it is the right fix:

- **Equivalent to the audit's proposal.** `(∀ n : ℕ, (2 : ℝ) ^ ((n : ℝ) / 2) ≤ (f n : ℝ))` is `rfl`-equal to the new declaration's type, and the exponent really is `Real.rpow`.
- **The strengthening is true, not merely stronger.** I proved it outright, sorry-free, from `(Finset.Ioc (n / 2) n).powerset ⊆ (Finset.Icc 1 n).powerset.filter _` via `Finset.card_le_card`; the old form no longer follows, as at `n = 1` it asked only `1 ≤ f 1`.
- **Faithful to the source.** The page gives `f(n) ≥ 2^{n/2}` pointwise, with no asymptotic notation or constant to absorb a floor, so natural division genuinely weakened it.
- **Builds.** `lake --wfail build` of the module is warning-free, and nothing else references this declaration.

---

### Changes requested

1. `FormalConjectures/ErdosProblems/748.lean:64` — the docstring's `[n/2,n]` is a proof hint that does not work: under this file's `IsSumFree`, which permits repeated summands, `¬ IsSumFree ((Finset.Icc 2 4 : Finset ℕ) : Set ℕ)` holds since `2 + 2 = 4`. The source does write `[n/2,n]`, but it is loose there, and the construction proving the bound needs the strict upper half. Replace with:

```lean
/-- It is trivial to see that $f(n) \geq 2^{\frac{n}{2}}$, considering all subsets of
$(n/2,n]$. -/
```

smmercuri:

Applied the requested change in 06f8d53: the lower-bound hint now says subsets of $(n/2,n]$, with a note that the source writes the closed form, which is not sum-free for even $n$. The module still builds warning-free with `--wfail`.

mo271 (approved):

PR #6496, opened 2026-09-22
fix(ErdosProblems/772): restrict the n^{2/3} upper bound to k >= 4

### LINKED ISSUE #6474: ErdosProblems/772: `erdos_772.variants.upper_bound` asserts the `n^{2/3}` bound for all `k ≥ 1`, where it fails at k = 1, 2
`Erdos772.erdos_772.variants.upper_bound` claims `∃ C, ∀ k ≥ 1, ∀ᶠ n, (H k n : ℝ) ≤ C * n^(2/3)`, but with this file's ordered-pair counting convention the constraint `∀ m, sumRep A m ≤ 2` forces `A` itself to be Sidon, so `H 2 n = n` (and the documented truncation gives `H 1 n = n`), which no fixed real `C` can bound by `C * n^(2/3)`; Erdős's B₂^(2) construction supports the bound only from `k ≥ 4` under that convention.

https://tadamcz.com/fc-review-results/782312817c/#/f/ErdosProblems/772

- 1 misformalization
- trivial proof compiles
- fix compiles

smmercuri:
`erdos_772.variants.upper_bound` asserted `∃ C, ∀ k ≥ 1, ∀ᶠ n, (H k n : ℝ) ≤ C * n^(2/3)`, but
`sumRep` counts *ordered* pairs, so the hypothesis `∀ m, sumRep A m ≤ 2` already forces `A`
itself to be Sidon; hence `H 2 n = n`, and `H 1 n = n` as well (for `n ≥ 2` no admissible `A`
exists, so the documented truncation at `n` applies). No fixed real `C` bounds `n` by
`C * n^(2/3)` eventually, so the statement as written was false. Erdős [Er84d] derives the upper
bound from a `B₂^(2)` set (the numbers `4^u + 4^v` with `u`, `v` from two disjoint exponent
ranges), which has at most two *unordered* representations of each sum and therefore at most
four ordered ones; the bound is supported from `k ≥ 4` under this file's convention.

**Change:** replace `∀ k ≥ 1` by `∀ k ≥ 4` in `erdos_772.variants.upper_bound`, and extend its
docstring to record why the restriction is needed. Nothing else is touched: the lower-bound
declarations are monotone in the right direction and remain at `k ≥ 1`.

**Checked:** `lake --wfail build 'FormalConjectures.ErdosProblems.«772»'` — completed
successfully with no warnings.

Judgement call: the erdosproblems.com page states the upper bound for all `k ≥ 1`, which is
loose; `k ≥ 4` is the weakest restriction actually supported by the cited construction. Whether
`k = 3` also admits such a construction is not settled by the sources, so it is excluded.

Fixes #6474


smmercuri:

Checked against #6474, the audit page, and erdosproblems.com/772, with the old and new statements stress-tested in Lean. **The fix is correct**; one docstring request below.

Why it is the right fix:

- **The defect is gone.** Sorry-free in Lean, `∀ m, sumRep A m ≤ 2` forces `A` Sidon, so `H k n = n` for `k ≤ 2`, which no `C·n^(2/3)` bounds; `sumRep` is ordered (`FormalConjecturesForMathlib/Combinatorics/Additive/Convolution.lean:61`), and it reaches no admissible `k`.
- **Equivalent to the audit's proposal.** It is that proposal verbatim (`Iff.rfl`), and `H` is antitone in `k`, so `∀ k ≥ 4` carries exactly the `k = 4` content.
- **Faithful to the source, and 4 is exactly right.** The page states the bound for all `k ≥ 1`, false under the ordered convention; the Erdős scan would not OCR, so I re-derived the construction (`4^u + 4^v`, disjoint exponent ranges: two unordered, four ordered representations). `k = 3` fails too — its only collisions are `a + b = 2c`, so Sidon subsets are the 3-AP-free ones, of size `n^{1-o(1)} ≫ n^{2/3}`.
- **Builds.** `lake --wfail build` on the module is clean; nothing else imports it.

---

### Changes requested

1. `FormalConjectures/ErdosProblems/772.lean:80-84` — the docstring explains `k ≥ 4` and `k ≤ 2` but is silent on `k = 3`. Replace it with (builds with `--wfail`):

```lean
/-- Erdős [Er84d] proved that $H_k(n) \ll n^{2/3}$, where the implied constant is absolute.

Since $1_A\ast 1_A$ counts ordered pairs, the bound needs $k\geq 4$: the construction of
Erdős has at most two unordered representations of each sum, hence at most four ordered ones,
while $k\leq 2$ forces $A$ itself to be Sidon and so $H_k(n)=n$. For $k=3$ the only collisions
are $a+b=2c$, so the Sidon subsets of $A$ are its $3$-term-progression-free subsets, which have
size $n^{1-o(1)}$, and the bound fails there too. -/
```

smmercuri:

Applied review item 1 verbatim in 2748152: the `upper_bound` docstring now explains that $k=3$ fails too, since its only collisions are $a+b=2c$, making the Sidon subsets the $3$-term-progression-free ones of size $n^{1-o(1)}$. The module still builds clean with `--wfail`.

mo271 (approved):

PR #6008, opened 2026-09-14
Mark five problems solved with links to Lean 4 proofs

niketp03:
Marks as `research solved`, each with a `formal_proof using lean4` link to a proof in niketp03/atlas-fc-verified derived with minor modifications from facebookresearch/atlas-lean:

- ErdosProblems/138: erdos_138.variants.dvd_two_pow (answer(True))
- ErdosProblems/337: erdos_337.variants.ruzsa_turjanyi
- OEIS/108081: count_words_in_x_is_a_shifted
- OEIS/211417: supercongruence
- OpenQuantumProblems/35: ame_9_10_open (answer(True))

Each linked proof imports the formal-conjectures problem file, uses its definitions directly, restates the statement (with the answer filled in where applicable), and depends only on propext, Classical.choice and Quot.sound.

Docstrings also cite independent concurrent work: Campos-Fox-Schildkraut (arXiv:2608.20824, proof generated with ChatGPT) for Erdos 138, noting that the separate W(k+1) - W(k) variant was solved in arXiv:2605.22763; and Peter Bala's Jan 2020 OEIS proof sketch via Mestrovic (arXiv:1111.3057) for the A211417 supercongruence.

PR #6497, opened 2026-09-22
fix(OpenQuantumProblems/35): restrict ame_7_exists_iff to d ≥ 2

### LINKED ISSUE #6477: OpenQuantumProblems/35: `ame_7_exists_iff` is stated for every `d : ℕ`, but `ExistsAME 7 1` holds while `3 ≤ 1` fails
`OpenQuantumProblem35.ame_7_exists_iff (d : ℕ) : ExistsAME 7 d ↔ 3 ≤ d` quantifies over all local dimensions including d = 1, where the file's definitions make `ExistsAME 7 1` true (the unique computational-basis configuration with amplitude 1 has 1×1 maximally mixed reductions) while the right-hand side is false, so the classification is refutable as written; the cited seven-party classification excludes the trivial dimension d = 1.

https://tadamcz.com/fc-review-results/782312817c/#/f/OpenQuantumProblems/35

- 1 misformalization
- 1 minor
- 1 equivalent reformulation
- trivial proof compiles
- fix compiles

smmercuri:
`ame_7_exists_iff` stated the seven-party classification for every `d : ℕ`, with no lower bound on the local dimension. Under this file's definitions `ExistsAME 7 1` is true: `Config 7 1` is a singleton, so the state with amplitude `1` on the unique computational-basis configuration is normalized and every reduced density matrix is the `1×1` identity, which is exactly `maximallyMixed 3 1`. Since `3 ≤ 1` is false, the stated iff was refutable. Shi-Zhang-Zhao-Li ([arXiv:2608.01011](https://arxiv.org/abs/2608.01011)) prove that an AME state of seven *qudits* exists iff `d ≥ 3`; the trivial dimension `d = 1` is outside the AME convention.

**Change:** add the hypothesis `(hd : 2 ≤ d)` to `ame_7_exists_iff`, explain the restriction in its docstring, and supply the bound at the two call sites `ame_7_6_open` and `ame_7_10_open`. Statement, category and AMS tags are otherwise unchanged and the proof stays `by sorry`.

**Checked:** `lake --wfail build 'FormalConjectures.OpenQuantumProblems.«35»'` — `Build completed successfully`.

Judgement call: `2 ≤ d` is used rather than rewriting the right-hand side as `d = 1 ∨ 3 ≤ d`, since `d = 1` is a degenerate artefact of the encoding rather than part of the cited classification, and `2 ≤ d` matches the convention already used in this file by `ame_2_exists` and `ame_3_exists`.

Note for reviewers: open PR #6008 also edits this file, but a different declaration (`ame_9_10_open`); the two changes are independent, though they may need a trivial textual merge.

Fixes #6477


smmercuri:

Checked against issue #6477, the audit page for `OpenQuantumProblems/35`, and Shi–Zhang–Zhao–Li (arXiv:2608.01011), then stress-tested the statement in Lean. **Verdict: correct.**

Why it is the right fix:

- **The defect is gone.** Reproving the degeneracy from the file's own definitions — `Config 7 1` is a singleton and `maximallyMixed m 1` is the 1×1 identity — gives a sorry-free `ExistsAME n 1` and hence a sorry-free refutation of the old unrestricted iff, which `2 ≤ d` now blocks.
- **`2 ≤ d` is the better shape.** It matches the file's own convention in `ame_2_exists`, `ame_3_exists` and `oqp_35`, and I proved it interderivable in both directions with the audit's alternative `d = 1 ∨ 3 ≤ d`, using the existing `not_existsAME_zero_dim` for `d = 0`.
- **Faithful to the source.** The paper (abstract only — the full text was not reachable) states that "an absolutely maximally entangled state of seven qudits exists if and only if the local dimension satisfies $d \geq 3$", and "qudits" is the $d \geq 2$ convention.
- **Builds.** The two updated call sites are the only users, `ExistsAME 7 6 ↔ True` and `ExistsAME 7 2 ↔ False` still follow, and `lake --wfail build` is clean.

mo271:
@MarioKrenn6240 can you have a look?

PR #6498, opened 2026-09-22
fix(ErdosProblems/1051): take the growth liminf in EReal

### LINKED ISSUE #6458: ErdosProblems/1051: `GrowthCondition`'s real-valued `liminf` excludes sequences with `aₙ^(1/2ⁿ) → ∞`
`Erdos1051.GrowthCondition` takes `Filter.liminf (fun n => ((a n : ℝ) ^ (1 / 2 ^ n : ℝ))) Filter.atTop > 1` in ℝ, where the liminf of an unbounded family is the supremum of its eventual lower bounds and hence defaults to 0, so the condition is false for sequences such as `a n = (n + 2) ^ (2 ^ n)` whose roots are `n + 2` and tend to +∞, and `erdos_1051` therefore omits the infinite-liminf case that Erdős–Graham and [BKKKZ26] explicitly include.

https://tadamcz.com/fc-review-results/782312817c/#/f/ErdosProblems/1051

- 1 misformalization
- 1 equivalent reformulation
- trivial proof compiles
- fix compiles

smmercuri:
`Erdos1051.GrowthCondition` took `Filter.liminf (fun n => (a n : ℝ) ^ (1 / 2 ^ n : ℝ)) atTop > 1` in `ℝ`. Mathlib's real `liminf` is the supremum of the eventual lower bounds, and that supremum defaults to `0` when the bounds are unbounded above, so the condition was *false* for sequences whose roots tend to `+∞` — for instance the strictly increasing `a n = (n + 2) ^ (2 ^ n)`, whose roots are exactly `n + 2`. Erdős–Graham (Old and new problems, p. 64) require only `a n ^ (1 / 2 ^ n) > 1 + ε` with no upper bound, and [BKKKZ26] explicitly discusses `lim a n ^ (1 / 2 ^ n) = ∞` as a covered case, so `erdos_1051` was silently omitting the infinite-liminf case.

**Change:** cast the real roots into `EReal` before taking the `liminf`, so `+∞` is a legitimate value, and record this in the definition's docstring. Nothing else is touched; both theorems keep their statements, attributes and `sorry`.

**Checked:** `lake --wfail build 'FormalConjectures.ErdosProblems.«1051»'` — `Build completed successfully`, no warnings. Separately verified in a scratch file that `GrowthCondition ((· + 2) ^ (2 ^ ·))` (the previously excluded `+∞` case) now holds, and that the finite case `a n = 2 ^ (2 ^ n)`, with `liminf = 2`, still holds, so the fix is not vacuous.

Judgement call: the `EReal` `liminf` was preferred over the equivalent eventual-lower-bound form `∃ c > 1, ∀ᶠ n in atTop, c ≤ (a n : ℝ) ^ (1 / 2 ^ n : ℝ)` because it keeps the Lean text literally matching the `liminf > 1` of the source.

Fixes #6458


smmercuri:

Checked against issue #6458, the audit page for `ErdosProblems/1051`, and the arXiv source of [BKKKZ26], then stress-tested the definition in Lean. Verdict: **correct**.

Why it is the right fix:

- **The defect is gone.** The strictly increasing `a n = (n + 2) ^ (2 ^ n)` has roots exactly `n + 2 → ∞`, yet the old real `liminf` rejected it — its eventual lower bounds are unbounded and `Real.sSup` of such a set is `0` — and the new definition accepts it.
- **Equivalent to the audit's proposal.** I proved the full iff, both directions and `sorry`-free, between this `EReal` `liminf` and `∃ c : ℝ, 1 < c ∧ ∀ᶠ n in atTop, c ≤ (a n : ℝ) ^ (1 / 2 ^ n : ℝ)`.
- **Faithful to the source.** [BKKKZ26] quotes the question with `liminf aₙ^(1/2ⁿ) > 1` and reads Erdős–Graham as meaning "for example, that `lim aₙ^(1/2ⁿ) = ∞`", so `+∞` sits inside the condition.
- **Contained and non-vacuous.** `GrowthCondition` has exactly one use, `2 ^ (2 ^ n)` is still accepted while constant `1` and `0` are rejected, and the module builds with `--wfail`.

`ErdosProblems/263.lean` met an analogous defect with a separate variant, but there the real `liminf` sits in a sufficient condition that stays true when narrowed; here it is the question's own hypothesis.

mo271 (approved):

PR #6500, opened 2026-09-22
fix(ErdosProblems/868): count unordered representations, fix constant

### LINKED ISSUE #6475: ErdosProblems/868: `ncard_add_repr` counts ordered representations and `variants.fixed_ε` uses the endpoint constant `(log(4/3))⁻¹`
`Erdos868.ncard_add_repr` counts arbitrary tuples `a : Fin o → ℕ` with `∑ a i = n`, i.e. ordered representations, while Erdős–Nathanson and Larsen–Larsen count nondecreasing ones (so the Lean count is about twice the source's r_A(n)), and `Erdos868.erdos_868.variants.fixed_ε` hypothesises only `∀ᶠ n, (Real.log (4/3))⁻¹ * Real.log n < f n`, whereas Erdős–Nathanson's Theorem 2 assumes `c * log n < f n` for some fixed `c > (log(4/3))⁻¹`, so the Lean statement asserts the endpoint case with an inflated representation count.

https://tadamcz.com/fc-review-results/782312817c/#/f/ErdosProblems/868

- 2 misformalizations
- 3 equivalent reformulations
- trivial proof compiles
- fix compiles

smmercuri:
`Erdos868.ncard_add_repr` counted arbitrary tuples `a : Fin o → ℕ` with `∑ a i = n`, i.e. ordered representations, so for order 2 it was about twice the representation function of the sources: Erdős–Nathanson (1979, p. 89) define `r(n)` as the number of representations `n = a_j + a_k` with `a_j, a_k ∈ A` and `a_j ≤ a_k`, and Larsen–Larsen count pairs `(a, b)` with `a ≤ b` likewise. Separately, `erdos_868.variants.fixed_ε` hypothesised the endpoint bound `(log(4/3))⁻¹ * log n < f n`, whereas Erdős–Nathanson's theorem assumes `r(n) > c log n` for a fixed constant `c > log⁻¹(4/3)`; they explicitly remark that the constant may be best possible, so the endpoint case is not what was proved.

**Change:** `ncard_add_repr` now restricts to nondecreasing tuples (`Monotone a`), matching `a_1 ≤ ⋯ ≤ a_h`. `erdos_868.variants.fixed_ε` now quantifies over `c > (Real.log (4 / 3))⁻¹` and uses `c * Real.log n` in the eventual bound. Both docstrings were updated to match. Categories and AMS tags are unchanged.

**Checked:** `lake --wfail build 'FormalConjectures.ErdosProblems.«868»'` completes successfully with no warnings. A scratch check confirms `ncard_add_repr {1, 2} 2 3 = 1` under the new definition, which is the sources' count (it was `2` before).

Changing the definition also changes `parts.i` and `parts.ii`, where erdosproblems.com phrases the question with the ordered convolution `1_A * 1_A`. This is harmless: `r ≤ F ≤ 2r`, so `f n → ∞` and "there is a positive logarithmic coefficient" are equivalent for the two conventions, and a single definition keeps the file consistent with the original Erdős–Nathanson and Larsen–Larsen counts.

Fixes #6475


smmercuri:

Checked against issue #6475, the audit page for `ErdosProblems/868`, and the cited Larsen–Larsen paper, with the statements stress-tested in Lean. Verdict: **CORRECT**.

Why it is the right fix:

- **Faithful to the source.** Larsen–Larsen §1 defines `r_A(m)` as the number of pairs `(a,b) ∈ A²` with `a+b=m` **and `a ≤ b`**, and states Erdős–Nathanson proved the result "if `c > log(4/3)⁻¹` and `r_A(m) > c log m` for all sufficiently large `m`" — exactly the two things this PR changes; erdosproblems.com instead pairs the ordered convolution `1_A*1_A` with the unordered constant, and following the paper is the better choice.
- **The defect is gone.** The audit's demo `ncard_add_repr {1,2} 2 3 = 2` no longer compiles; under the new definition I proved the value is `1`, the sources' count.
- **The knock-on is proved, not assumed.** I re-declared the old ordered count, proved `r ≤ F ≤ 2r`, and derived that the part (i) `Tendsto` hypotheses and the part (ii) `∀ ε > 0` statements are interchangeable between the two conventions, so both `answer(False)` claims are untouched.
- **Equivalent to the audit's proposal.** `∀ᵉ (A) (c > (Real.log (4 / 3))⁻¹), …` is `Iff.rfl` to the audit's version, and `lake --wfail build` of the module succeeds.

mo271 (approved):
Thanks!

PR #6502, opened 2026-09-22
fix(WrittenOnTheWallII/GraphConjecture141): drop the floor on girth/2

### LINKED ISSUE #6479: WrittenOnTheWallII/GraphConjecture141: `conjecture141` uses `(G.girth / 2 : ℤ)`, flooring the source's unrounded `girth/2` and losing a vertex at odd girth
`WrittenOnTheWallII.GraphConjecture141.conjecture141` states the bound as `(G.girth / 2 : ℤ) - 1 + max_v λ(v) ≤ largestInducedTreeSize G`, where `G.girth / 2` is natural-number division, so for odd girth `2k+1` it asserts only `k - 1 + λ ≤ tree(G)` instead of DeLaViña's unrounded `tree(G) ≥ (1/2)·girth - 1 + max_v λ(v)` (equivalently `⌈girth/2⌉ - 1 + λ`), and the docstring's `⌊girth(G) / 2⌋` repeats the rounding.

https://tadamcz.com/fc-review-results/782312817c/#/f/WrittenOnTheWallII/GraphConjecture141

- 1 misformalization
- trivial proof: bailed out
- fix compiles

smmercuri:
`conjecture141` stated the bound as `(G.girth / 2 : ℤ) - 1 + max_v λ(v) ≤ tree(G)`. That `/` is integer division, so the Lean floored half the girth: for odd girth `2k+1` it asserted only `k - 1 + λ ≤ tree(G)` where the source asserts `k + λ ≤ tree(G)`. The archived WOWII list states Conjecture 141 as "If G is a simple connected graph, then tree(G) ≥ (1/2)*girth - 1 + maximum of l(v)" — unrounded (`all.html`, July 19 2005 entry; WOWII writes `FLOOR[...]`/`CEIL[...]` explicitly when it means rounding). The docstring repeated the floor as `⌊girth(G) / 2⌋`.

**Change:** state the inequality over `ℝ` with genuine division, `(G.girth : ℝ) / 2 - 1 + ↑(⨆ v, λ(v)) ≤ ↑(largestInducedTreeSize G)`, matching how the neighbouring conjectures 142 and 143 state their fractional bounds, and correct the docstring to the source's unrounded form. The theorem stays `research solved`: the proof still goes through after routing the girth-4 and girth-5 cases through the counting bound `Δ(G) + girth(G) ≤ tree(G) + 3` (previously used only from girth 6, which no longer suffices at girth 5), with girth 3 and the acyclic case still coming from `λ(v) + 1 ≤ tree(G)`. The real goal is reduced to the equivalent integer inequality `girth + 2λ ≤ 2·tree(G) + 2` so `omega` can close the cases.

**Checked:** `lake --wfail build FormalConjectures.WrittenOnTheWallII.GraphConjecture141` — `Build completed successfully`, no warnings.

Judgement call: `ℝ` rather than `⌈(G.girth : ℝ) / 2⌉` over `ℤ`, since it is literally the source's `(1/2)*girth` and is the form used by the sibling WOWII girth conjectures; the two are equivalent here because `tree(G)` and `λ(v)` are integers.

Fixes #6479


smmercuri:

Checked against issue #6479, the audit page for `WrittenOnTheWallII/GraphConjecture141`, and the archived WOWII list, then stress-tested the statement and the adapted proof in Lean. Verdict: **correct**.

Why it is the right fix:

- **Faithful to the source.** The archived `all.html` (July 19, 2005 entry) reads `141. If G is a simple connected graph, then tree(G) ≥ (1/2)*girth - 1+ maximum of l (v)`, and the same page writes `FLOOR(...)`/`CEIL(...)` explicitly 72 and 70 times respectively, so the bare `(1/2)*girth` is genuinely unrounded.
- **The defect is gone, as a strict strengthening.** From the new statement I derived both the audit's ceiling variant over `ℤ` and the old floored statement, and exhibited the closed gap at girth 5 with `l = 0`, `tree = 1`, which satisfies the floored form but not the real one.
- **The proof still stands.** `#print axioms conjecture141` gives `[propext, Classical.choice, Quot.sound]`, no `sorryAx`; widening the `maxDegree_add_girth_le_largestInducedTreeSize_add_three` branch from girth ≥ 6 to girth ≥ 4 is necessary, since girth 5 needs `2*l + 3 ≤ 2*tree`, which `l + 1 ≤ tree` alone cannot give.
- **Builds.** `lake --wfail build FormalConjectures.WrittenOnTheWallII.GraphConjecture141` completes successfully with no warnings.

mo271 (approved):

PR #5733, opened 2026-09-11
fix(WrittenOnTheWallII/GraphConjecture34): average nonzero pair distances

smmercuri:
`WrittenOnTheWallII.GraphConjecture34.conjecture34` bounded `path(G)` using `SimpleGraph.distavg G S`, the mean over all vertices `v` of the minimum distance from `v` to the set `S`, including the zero terms for `v ∈ S`. The source's `dist_avg(S, V)` is instead the average of all nonzero pairwise distances `d(s, v)` with `s ∈ S` and `v ∈ V` (the double distance sum divided by `|S|(|V| - 1)`).

**Change:** replace `distavg` by an inline `distAvg S` computing the average of the nonzero `G.dist s v` over `S × V`, for both the center set `C` and the maximum-degree set `M`. With the corrected quantity the statement is false: the tree on 39 vertices given by a vertex with three pendant leaves, joined by a path of five edges to the root of a perfect binary tree of depth four, has `path(G) = 11` while `C` and `M` are singletons with distance sums `154` and `266`, giving bound `⌈420/38⌉ = 12`. The theorem now records `answer(False)` (dropping the linter-suppressed `answer(sorry)`), with the counterexample documented in the docstring.

**Checked:** `lake --wfail build FormalConjectures.WrittenOnTheWallII.«GraphConjecture34»` passes with no warnings.

Judgement call: the WOWII source still lists this conjecture as open, but the counterexample above was verified against the corrected definition, so `answer(False)` is recorded. `GraphConjecture33` uses `distavg` with the same defect; that is left for a separate issue.

Fixes #5720

PR #6503, opened 2026-09-22
fix(WrittenOnTheWallII/GraphConjecture33): use dist_avg(M, V)

### LINKED ISSUE #6480: WrittenOnTheWallII/GraphConjecture33: `conjecture33` uses `distavg`, the mean of nearest-`M` distances, not `dist_avg(M, V)` over nonzero pairs
`WrittenOnTheWallII.GraphConjecture33.conjecture33` bounds `path G` by `⌈2 * distavg G M⌉`, where `distavg` averages over all vertices the minimum distance to `M`, including the zero terms for v ∈ M, whereas WOWII's glossary defines `dist_avg(M, V)` as the average of all nonzero `dist(u, v)` with u ∈ M and v ∈ V; on the perfect binary tree of depth 5 the two quantities are 11/21 and 1797/310, giving the true bound 2 ≤ 11 rather than the false 12 ≤ 11, so the recorded `answer(False)` concerns a different inequality.

https://tadamcz.com/fc-review-results/782312817c/#/f/WrittenOnTheWallII/GraphConjecture33

- 1 misformalization
- 1 minor
- trivial proof compiles
- fix compiles

smmercuri:
`conjecture33` bounded `path G` by `⌈2 * distavg G M⌉`, but `SimpleGraph.distavg G M` is the average over *all* vertices `v` of `min_{m ∈ M} dist(v, m)`, which includes a zero term for every vertex of `M`. WOWII glossary entry 37 (linked from Conjecture 33) instead defines `dist_avg(M, V)` as "the average of all nonzero `dist_G(u, v)` such that `u` is in `M` and `v` is in `V`" — a mean over ordered pairs, with no minimum taken, and with denominator the number of nonzero-distance pairs. The two disagree on almost every graph (on `K₂` with `M = V` the Lean value is `0` and the source value is `1`), so the recorded `answer(False)` was attached to a different inequality. The docstring also described `path(G)` as "the floor of the average distance of `G`", whereas glossary entry 35 — and the imported `SimpleGraph.path` — give the number of vertices of a largest induced path.

**Change:** replace `distavg G M` by a local `let` computing the mean of `G.dist m v` over the pairs of `M ×ˢ univ` at nonzero distance, following the same shape used in `GraphConjecture34` (PR #5733, the sibling conjecture using the same two-argument `dist_avg(S, V)` invariant); rewrite the docstring to state both invariants correctly and to record the source's counterexample. The quantifiers, the maximum-degree set, `answer(False)`, `@[category research solved, AMS 5]` and the sanity checks are unchanged.

**Checked:** `lake --wfail build FormalConjectures.WrittenOnTheWallII.GraphConjecture33` — `Build completed successfully`, no warnings.

Two `dist_avg` invariants appear in WOWII: the one-argument `dist_avg(S)` (mean over pairs *within* `S`, available as `SimpleGraph.distAvgSet` from PR #5736) and the two-argument `dist_avg(S, V)`. Conjecture 33 cites entry 37, the two-argument one, so `distAvgSet` does not apply here and the inline `let` of `GraphConjecture34` is reused instead. Status and polarity are confirmed by the source, which marks Conjecture 33 `F` and records an October 2005 counterexample with `path(G) = 7` and `dist_avg(M, V) = 3.56` (so `⌈7.12⌉ = 8 > 7`).

Fixes #6480


smmercuri:

Compared against issue #6480, the audit page, and the archived WOWII source (`all.html`, `wowIIdefs.js`), then stress-tested in Lean. **Verdict: correct.**

Why it is the right fix:

- **Faithful to the source.** The `all.html` row for Conjecture 33 links `printDefinitions(35,37,...)`, so the invariant is glossary entry 37 — "the average of all nonzero dist_G(u,v) such that u is in M and and v is in V" — not the one-argument `dist_avg(S)` that `SimpleGraph.distAvgSet` models; entry 35 gives `path(G)` as the order of a largest induced path, and the row's `F` status with "Path number is 7, ... avg dist = 3.56" supports `answer(False)`.
- **The defect is gone, and the fix matches the audit's.** Re-ran the audit's K₂ demonstration — `distavg G M = 0` where the new expression is `1` — and closed the iff against a fresh transcription of the audit's suggested fix by `Iff.rfl`.
- **No junk denominator.** Proved the filtered pair set is nonempty for every connected `Nontrivial` graph, so `/ pairs.card` never falls into `x / 0 = 0`.
- **Builds.** `lake --wfail build FormalConjectures.WrittenOnTheWallII.GraphConjecture33`, no warnings.

A follow-up, not a change request: this `let` is now byte-identical with GraphConjecture34's and would be better shared in `FormalConjecturesForMathlib`, which would also give `SimpleGraph.distavg` — now with no user left in the repo — a successor.

mo271 (approved):
Thanks!

PR #6504, opened 2026-09-22
fix(ErdosProblems/682): count exceptional gaps with p_n in [X, 2X]

### LINKED ISSUE #6472: ErdosProblems/682: `exceptional_count` bounds an index-scale count by `X / (log X) ^ 2`, one log factor too strong
`erdos_682.variants.exceptional_count` counts exceptional indices `n ∈ Finset.Icc 1 X` and bounds that count by `(X : ℝ) / Real.log X ^ 2`, but Gafni–Tao's `≪ X / (log X)^2` counts gaps whose lower prime satisfies `p_n ∈ [X, 2X]`, which over the first `N` indices yields only `O(N / log N)`, so the Lean statement (and its docstring) is stronger than the cited theorem by a logarithmic factor and is incompatible with the paper's conditional asymptotic `c N / log N` for the index count.

https://tadamcz.com/fc-review-results/782312817c/#/f/ErdosProblems/682

- 1 misformalization
- 1 equivalent reformulation
- trivial proof: bailed out
- fix compiles

smmercuri:
`erdos_682.variants.exceptional_count` counted exceptional *indices* `n ∈ Finset.Icc 1 X` and bounded that count by `X / (log X) ^ 2`. Gafni–Tao (arXiv:2508.06463, Theorem 1.1) instead bound `N(X)`, the number of exceptional gaps whose lower prime satisfies `p_n ∈ [X, 2X]`, by `≪ X / (log X)^2`. Transferred to the index scale this only gives `O(N / log N)` exceptional `n ≤ N`, and the paper's conditional asymptotic on that scale is `c N / log N`, so the Lean statement was stronger than the cited theorem by a logarithmic factor. The docstring (taken from the erdosproblems.com summary, which repeats the same slip) had the same error.

**Change:** restate the count over the prime window of the paper: `{n | p n ∈ Set.Icc X (2 * X) ∧ ∀ m ∈ Set.Ioo (p n) (p (n + 1)), m.minFac < p (n + 1) - p n}.ncard`, keeping the sharp `X / (log X) ^ 2` bound, and update the docstring to match. `open scoped Classical in` is no longer needed since `Set.ncard` requires no decidability.

**Checked:** `lake --wfail build 'FormalConjectures.ErdosProblems.«682»'` — `Build completed successfully`, no warnings.

Judgement call: the audit suggested either weakening the bound to `X / log X` on the index scale or moving to the prime window; I took the second, since it reproduces Theorem 1.1 verbatim and keeps the full strength of the result. I checked separately that the counting set is finite (`p` is strictly monotone, so `n ≤ p n ≤ 2 * X`), so `Set.ncard` is an honest count rather than a junk value.

Fixes #6472


smmercuri:

Checked against issue #6472, the audit page for `ErdosProblems/682`, and the arXiv e-print source of Gafni–Tao (2508.06463), and stress-tested the statement in Lean. Verdict: **correct**.

Why it is the right fix:

- **Faithful to the source.** Theorem 1.1 counts "prime gaps $(p_n,p_{n+1})$ with $p_n \in [X,2X]$ that do not contain a rough number", bounded by $\ll X/\log^2 X$; this PR renders that literally, whereas the audit's applied patch states only the weaker corollary derived in the following paragraph.
- **The defect is gone.** Counting `p n ∈ Set.Icc X (2 * X)` removes the index-scale/prime-scale mismatch; no counterexample demo is possible, since the defect is an asymptotic strengthening rather than a computable boundary failure, which is why the audit bailed out.
- **No junk values.** I proved in Lean that `n ≤ p n`, so the counting set sits inside `Set.Icc 0 (2 * X)` and is finite, making `Set.ncard` an honest count rather than the `0` it gives infinite sets — the point on which dropping `open scoped Classical in` turns.
- **A deliberate choice, not a discrepancy.** The audit's branches are inequivalent: the prime-window bound gives the index-scale one only via the prime number theorem, not conversely.
- **Builds.** `lake --wfail build` on the module succeeds with no warnings.

mo271 (approved):
Thanks, LGTM!

